PR #564, opened 2016-12-13
return all selected images in the API for multilingual apps

### LINKED ISSUE #460: Include all multilingual images in json
The world read json does not include all urls for the images, it only contains the images for a single language. This makes it very hard to create apps that support multiple languages.

Include the urls of the image urls for the other languages as well.

hangy:
fixes #460 
depends on/includes #563


lib/ProductOpener/Display.pm
```diff
@@ -2413,7 +2419,7 @@ HTML
 		
 		}
 		else {
-			$html .= "<p>$html_count" . lang("sep") . ":</p>";
+			$html .= "<p>$html_count " . lang("sep") . ":</p>";
```
hangy: Probably a result of a bad merge? I'll have look.

hangy: Actually the same as in [master](https://github.com/openfoodfacts/openfoodfacts-server/blame/b1a3ad30e03d2b900dea393bbbb952d5542c0175/lib/ProductOpener/Display.pm#L2451).

PR #565, opened 2016-12-13
Redirect static content from other subdomains

hangy:
Some websites hotlink to product images using fr.openfoodfacts.org or similar URLs. As we want to serve all static content from the static subdomain, we could just redirect those request to the static subdomain. This should be better for caching and possibly even SEO (no duplicate images being indexed).

Previously https://bitbucket.org/openfoodfacts/product-opener/pull-requests/52


teolemon:
@hangy Conflicting files

PR #582, opened 2016-12-20
Fix perlcritic issues

### LINKED ISSUE #580: Fix all errors raised by Perl::Critic
http://annuel.framapad.org/p/off-critique

hangy:
Closes #580 - at least for now.


hangy:
> I'm hoping the Use Modern::Perl only adds new features but does not cause any regression.

I had a quick look at [the source](http://cpansearch.perl.org/src/CHROMATIC/Modern-Perl-1.20161229/lib/Modern/Perl.pm) and could not find anything that looked fishy to me.

PR #583, opened 2016-12-20
Re-add the asterisk for non-taxonomized tags in lists

### LINKED ISSUE #417: Re-add the asterisk for non-taxonomized tags in lists (or add some other way to sort by taxonomizedness).
The asterisk was removed in #353, but it has a use, even if non-obvious (maybe add a headnote?).

hangy:
fixes #417


teolemon:
more legibility.  I'm Team asterisk :-)

PR #588, opened 2016-12-22
Implemented water hardness

### LINKED ISSUE #386: Carbonated water: Dureté de l'eau / Жёсткость воды
Some carbonated waters specify "[Жёсткость воды](https://ru.wikipedia.org/wiki/%D0%96%D1%91%D1%81%D1%82%D0%BA%D0%BE%D1%81%D1%82%D1%8C_%D0%B2%D0%BE%D0%B4%D1%8B)" ([Dureté de l'eau](https://fr.wikipedia.org/wiki/Duret%C3%A9_de_l%27eau)) in мг-экв/л.

hangy:
Fixes #386


cgi/product_multilingual.pl
```diff
@@ -453,7 +453,7 @@
 				$value = $value / 100 * $Nutriments{$nid}{dv} ;
 				$unit = $Nutriments{$nid}{unit};
 			}
-			$product_ref->{nutriments}{$nid} = unit_to_g($value, $unit);
+			$product_ref->{nutriments}{$nid} = unit_to_mmoll(unit_to_g($value, $unit), $unit);
```
stephanegigandet: It looks a bit weird to chain the conversions even though at least one of them will actually do nothing.
I'm thinking it may be easier to understand if we do something like if ($nid eq "water-hardness") -> use unit_to_mmol, else unit_to_g

hangy: Agreed, I'll fix that this weekend. Thanks for the feedback!

cgi/product_multilingual.pl
```diff
@@ -1445,7 +1450,7 @@ ($$$$$$)
 		elsif ((exists $Nutriments{$nid}) and (exists $Nutriments{$nid}{unit})) {
 			$unit = $Nutriments{$nid}{unit};
 		}
-		my $value = g_to_unit($product_ref->{nutriments}{$nid}, $unit);
+		my $value = mmoll_to_unit(g_to_unit($product_ref->{nutriments}{$nid}, $unit), $unit);
```
stephanegigandet: Can you add the if ($nid eq 'water-hardness') {  here as well?

hangy: Missed this, sorry. Fixed now. :)

PR #591, opened 2016-12-28
Web Share API

hangy:
Add Web Share API for products, tags, and searches.

### DIFF
--- a/cgi/search.pl
+++ b/cgi/search.pl
@@ -690,6 +690,8 @@ elsif ($action eq 'process') {
 	
 	
 	
+	my $share = lang('share');
+
 	# Graph, map, export or search
 
 	if (param("generate_map")) {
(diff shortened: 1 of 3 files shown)


teolemon:
Do you have some screenshot ?

hangy:
> Do you have some screenshot ?

Not right now. From the UI point of view, it is just an additional "[SHARE ICON] Share" button where the "Edit Product" button is located, which is hidden, unless the Web Share API is available. When you click it, the native browser UI should pop up, but Chrome Canary doesn't support in on desktop right now. We can possibly test it with Chrome Canary on Android though ...

hangy:
I managed it to run with Chrome Canary. The screenshot is a bit broken, because the Foundation Fonts currently don't work on my server, but you get the idea. It's the new button on the right hand side.

screenshot_20161229-102704
screenshot_20161229-102708

PR #600, opened 2016-12-31
Made "Création du produit" translatable

### LINKED ISSUE #515: "Création du produit" in French in History log
http://ru.preprod.openfoodfacts.org/cgi/product.pl?type=edit&code=4690228009297

image

hangy:
Use translatable string to for product creation log entry.

Fixes #515


stephanegigandet:
The comment field also contains free text entries entered by contributors (in any language), I think those lines will give bad results (probably perl complaining that $Lang{"some random comment"} is not a hash reference:

my $comment = $change_ref->{comment};
 +			$comment = lang($comment) if $comment;

I have not tested it though.

hangy:
You're right. It didn't actually log an error, but just left some comments empty. Fixed (pushing with some other changes).

Issue #549, opened 2016-11-20
Progressive Web App report


teolemon:
[pwa-report.pdf](https://github.com/openfoodfacts/openfoodfacts-server/files/602028/pwa-report.pdf)

hangy:
Looks like the [Manifest](https://www.w3.org/TR/appmanifest/) is not on Git. Should we generate it dynamically instead, so that a user could add `world.$server_domain`, `ru.$server_domain` etc. as separate web apps?

hangy:
I currently don't see how we could fulfill [`Progressive web apps work offline.`](https://developers.google.com/web/tools/lighthouse/audits/http-200-when-offline) usefully. But here are some of the TODOs apart from the manifest:

- [ ] [Has a registered Service Worker](https://developers.google.com/web/tools/lighthouse/audits/registered-service-worker)
- [ ] [URL responds with a 200 when offline](https://developers.google.com/web/tools/lighthouse/audits/http-200-when-offline)
- [ ] [Site uses HTTP/2 for its own resources](https://developers.google.com/web/tools/lighthouse/audits/http2)
- [ ] [Site does not use <link> that delay first paint](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources)
  - https://static.openfoodfacts.org/foundation/css/app.css (delayed first paint by 548ms)
  - https://static.openfoodfacts.org/foundation/foundation-icons/foundation-icons.css (delayed first paint by 267ms)
  - https://static.openfoodfacts.org/js/jquery-ui-1.11.4/jquery-ui.min.css (delayed first paint by 387ms)
  - https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.3/css/select2.min.css (delayed first paint by 163ms)
- [ ] [Site does not use <script> in head that delays first paint](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources)
  - https://platform.twitter.com/widgets.js (delayed first paint by 1903ms)
  - https://static.openfoodfacts.org/foundation/js/vendor/modernizr.js (delayed first paint by 309ms)
  - https://static.openfoodfacts.org/foundation/js/vendor/jquery.js (delayed first paint by 653ms)
  - https://static.openfoodfacts.org/js/jquery-ui-1.11.4/jquery-ui.min.js (delayed first paint by 839ms)
[...]

PR #610, opened 2017-01-13
Dynamic web manifest file

hangy:
Generate Web App manifest dynamically with icons etc. loaded from ProductOpener::Config. Somewhat related to #549

### DIFF
--- a/lib/ProductOpener/Config_off.pm
+++ b/lib/ProductOpener/Config_off.pm
@@ -124,27 +124,41 @@ $google_analytics = <<HTML
 HTML
 ;
 
+my @icons = (
+	{ "platform" => "ios", "sizes" => "57x57", "src" => "https://static.$server_domain/images/favicon/apple-touch-icon-57x57.png" },
+	{ "platform" => "ios", "sizes" => "60x60", "src" => "https://static.$server_domain/images/favicon/apple-touch-icon-60x60.png" },
+	{ "platform" => "ios", "sizes" => "72x72", "src" => "https://static.$server_domain/images/favicon/apple-touch-icon-72x72.png" },
+	{ "platform" => "ios", "sizes" => "76x76", "src" => "https://static.$server_domain/images/favicon/apple-touch-icon-76x76.png" },
+	{ "platform" => "ios", "sizes" => "114x114", "src" => "https://static.$server_domain/images/favicon/apple-touch-icon-114x114.png" },
+	{ "platform" => "ios", "sizes" => "120x120", "src" => "https://static.$server_domain/images/favicon/apple-touch-icon-120x120.png" },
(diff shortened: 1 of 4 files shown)


teolemon:
https://github.com/openfoodfacts/openfoodfacts-server/compare/cc933f3bb13a...b7f37c069045#diff-112c4c36357894c1e9f9b58a906866d1R144

lib/ProductOpener/Config_off.pm
```diff
@@ -124,27 +124,41 @@ $google_analytics = <<HTML
...
-<link rel="apple-touch-icon" sizes="120x120" href="/images/favicon/apple-touch-icon-120x120.png">
-<link rel="apple-touch-icon" sizes="144x144" href="/images/favicon/apple-touch-icon-144x144.png">
-<link rel="apple-touch-icon" sizes="152x152" href="/images/favicon/apple-touch-icon-152x152.png">
-<link rel="apple-touch-icon" sizes="180x180" href="/images/favicon/apple-touch-icon-180x180.png">
-<link rel="icon" type="image/png" href="/images/favicon/favicon-32x32.png" sizes="32x32">
-<link rel="icon" type="image/png" href="/images/favicon/android-chrome-192x192.png" sizes="192x192">
-<link rel="icon" type="image/png" href="/images/favicon/favicon-96x96.png" sizes="96x96">
-<link rel="icon" type="image/png" href="/images/favicon/favicon-16x16.png" sizes="16x16">
-<link rel="manifest" href="/images/favicon/manifest.json">
+<link rel="manifest" href="/cgi/manifest.pl">
```
stephanegigandet: So both iOS and Android  and most browsers will be able to find their icons from the <link rel="manifest" href="/cgi/manifest.pl"> line, without keeping all the <link rel="*icon" lines in the generated HTML?

stephanegigandet: https://diessi.ca/blog/a-guide-to-favicons-and-touch-icons/#Bulletproof_favicons

stephanegigandet: Also /cgi requests are expensive, as they will go to Product Opener, go through some initialization etc.
I think it would be better to generate manifest.json files dynamically at startup, and save them as static files.

hangy: Yes, cgi requests are definitely more expensive than static files. I intentionally added a long cache attribute to migitate that a bit. The problem is that we'd need to generate manifests for all thousands subdomains on startup, because we cannot know if anything that's related to the subdomains has changed since the last start. That doesn't sound too efficient either.

I don't know about the bulletproof favicons. Some stuff could have changed since the blog entry has been written 1.5 years ago, and I can't say if this would make anything bad on relevant browsers. If you don't like it, we can easily add the favicon stuff back to the config. 👍

teolemon: @hangy @stephanegigandet This one is important to finish and merge for native mobile apps suggestions (Chrome will offer app installs if the person keeps coming back on the web version)

PR #618, opened 2017-01-17
Finnish

teolemon:


hangy:
IMHO just test the translations in your test instance (see if some translation is shown) and then commit it to master. As these have been reviewed in Launchpad and automatically exported from there, I don't think pull requests are the best way to get this into master ..? I'd just merge #618, #619, #620, and #621 based on that. 👍

teolemon:
Let's avoid merging for now. We haven't synced the latest strings yet.

PR #626, opened 2017-01-18
Translations from Launchpad

teolemon:
Translations from Launchpad


teolemon:
What about merging this into Translations ?

PR #630, opened 2017-01-19
Generate po from lang

stephanegigandet:


lib/ProductOpener/Lang.pm
```diff
@@ -11571,6 +11544,121 @@ sub lang($) {
...
+"PO-Revision-Date: \\n"
+"Language-Team: \\n"
+"Last-Translator: \\n"
+
+msgctxt ":langtag"
+msgid   ":langtag"
+msgstr  "$l"
+
+msgctxt ":langname"
+msgid   ":langname"
```
hangy: Shouldn't be required after 62ef30c5b7410311030fd679d973d09c129c9b8a, but I guess it doesn't hurt.

stephanegigandet: Right, this was to generate the po files from the current %Lang.pm to get the exact format from the i18n-po branch. We can do some cleanup afterwards.

PR #632, opened 2017-01-19
I18n po stephane

stephanegigandet:
Merging the new .po branch into preprod so that we can deploy it on preprod... (and on prod)


teolemon:
merging the lp-branch into this one first ?

PR #640, opened 2017-01-22
Hide confirmed alert box

hangy:
Only display the edit-mode alert-boxes if they have not been confirmed, yet.


teolemon:
The cookie based solution is probably the least intrusive one. +1

stephanegigandet:
merged to master, will merge to i18n-po-stephane-2 and deploy on preprod and net

PR #661, opened 2017-01-24
merge master into preprod

stephanegigandet:


stephanegigandet:
preprod is not in good state, will delete the branch and start a clean one again later from master when needed

Issue #271, opened 2016-05-09
Add input fields for language and or country in the query part of the API HTTP requests


stephanegigandet:
Related to bug #270 : allow to specify country code and/or language code in the fields of the HTTP API request instead of using the subdomain.

stephanegigandet:
Maybe adding ?lc=fr&cc=be to the GET urls (or as POST parameters)

Nymerea:
and add the possibility to get lc and cc throughout the api .

For example off.org/api/list_countrycode.xml

and return something like : 
<countries>
<country code="be" langue="fr"> Belgium</country>
<country code="fr" langue="fr">France</country>
...
</countries>

stephanegigandet:
Currently tested on preprod: http://world.preprod.openfoodfacts.org/cgi/search.pl?search_terms=coke&search_simple=&cc=fr&lc=it

stephanegigandet:
Live on production. You can now add cc and lc query parameters to the url. They will override the subdomain.

stephanegigandet:
The way it works:

1. By default, the country code (cc) and language of the interface (lc) are set through the subdomain.
e.g. 
1.1. world.openfoodfacts.org --> cc = world, lc = en (default language for the "country")
1.2 fr.openfoodfacts.org --> cc = fr (France), lc = fr (French)
1.3 jp.openfoodfacts.org --> cc = jp (Japan), lc = ja (Japanese)
1.4 ch.openfoodfacts.org --> cc = ch (Switzerland), lc = de (German, set as main language for Switzerland)
1.5 ch-fr.openfoodfacts.org --> cc = ch, lc = fr (French)

2. If you pass ?cc=us and/or &lc=it, it will set the country to US and the language to Italian

-> if you just pass cc, the lc will be the one set in the subdomain.
e.g. world.openfoodfacts.org + cc = fr --> cc = fr, lc = en (default language of world)

3. The cc affects which results are returned. cc = fr -> you get only products with France listed in the countries where sold field.

4. The lc affects the language in which results will come back.

5. If the results include links to products, those links will use a new subdomain that corresponds to the overrides.

PR #693, opened 2017-01-29
issue #271 : add cc + lc requests parameters overrides

stephanegigandet:


lib/ProductOpener/Display.pm
```diff
@@ -328,13 +352,22 @@ sub analyze_request($)
...
 	
 	if ($request_ref->{query_string} =~ /(\&|\?)(utm_|ref=)/) {
 		$request_ref->{query_string} = $`;
-	}	
+	}
 	
+	# cc and lc query overrides have already been consumed by init()
+	# FIXME / TODO: just remove those 2 parameters, leave others.
+	if ($request_ref->{query_string} =~ /(\&|\?)(cc|lc)/) {
+		$request_ref->{query_string} = $`;
```
hangy: Couldn't they just be kept in so that other parts of the server can detect the override as well?

lib/ProductOpener/Display.pm
```diff
@@ -264,6 +273,21 @@ sub init()
 		return 301;
 	}
 	
+	
+	# Allow cc and lc overrides as query parameters
```
hangy: I guess this works, but we probably need to add some docs that ie. https://de.openfoodfacts.org/produkt/123?lc=fr will generate links in the website like https://fr.openfoodfacts.org/produit/123, because that could be unexpected.

stephanegigandet: The intended use is for the API requests, we'll probably document it only for the API (even if it does work on other requests too)

stephanegigandet: In fact it generates broken links like de.openfoodfacts.org/produit/123, I'm going to make the cc and lc overrides change the $subdomain too

PR #696, opened 2017-01-29
Issue/313 unselect images

stephanegigandet:


cgi/product_image_unselect.pl
```diff
@@ -0,0 +1,51 @@
...
+	exit(0);
+}
+
+my $product_ref = process_image_unselect($code, $id);
+
+my $data = encode_json({ status_code => 0, status => 'status ok', imagefield=>$id });
+
+print STDERR "product_image_unselect - JSON data output: $data\n";
+
+print header() . $data;
```
hangy: Would probably output this as HTML. I think this should use `header( -type => 'application/json', -charset => 'utf-8' )` as in 234b5c201e62007619c688ece9e63e824a958add.

lib/ProductOpener/Images.pm
```diff
@@ -961,6 +962,28 @@ sub process_image_crop($$$$$$$$$$) {
...
+	print STDERR "Images.pm - process_image_unselect - id: $id\n";
+			
+	# Update the product image data
+	my $product_ref = retrieve_product($code);
+	defined $product_ref->{images} or $product_ref->{images} = {};
+	if (defined $product_ref->{images}{$id}) {
+		delete $product_ref->{images}{$id};
+	}
+
+	store_product($product_ref, "unselected image $id");
```
hangy: What about making this log entry translatable like 0ebb55c633f6c8cceab2334a692e5a8ae1736104?

cgi/product_image_unselect.pl
stephanegigandet: Thanks, I'll change it.

lib/ProductOpener/Images.pm
stephanegigandet: We could, but then we need to do something else than the $comment = lang($comment) if $comment; as there's the id of the image in the comment. Not sure if it's worth it.

PR #709, opened 2017-02-05
Fix Wiki SSO

hangy:
Just puts the cookie values into an associative array after splitting the cookie string. Worked well when I tested the snippet on http://sandbox.onlinephpfunctions.com/.


mediawiki/extensions/CASAuth/CASAuth.php
```diff
@@ -83,18 +83,20 @@ function casLogin($user, &$result) {
...
 							&	stephane 			*/
 						$valeur_cookie = $cookie[$name_cookie];
 						
-						//Parser user_session&4414725571&user_id&manu1400
-						$tab_valeurs = split("&", $valeur_cookie); //Pas de limite
-						$numero_session = $tab_valeurs[1]; //num�ro de la session
-						$username = $tab_valeurs[3]; //nom de l'utilisateur
+						//Parser user_session&4414725571&user_id&manu1400 using http://stackoverflow.com/a/32768029/11963
+						$chunks = array_chunk(preg_split('/&/', $valeur_cookie), 2);
+						$tab_valeurs = array_combine(array_column($chunks, 0), array_column($chunks, 1));
```
hangy: Note that `array_column` requires PHP 5.5+. (PHP 5.6 is EOL for about half a year right now.)

hangy:
While the automatic login fails, it doesn't allow me to edit any wiki pages for some unknown reason. We'll probably need to change to [this](https://github.com/noris-network/mediawiki-extensions-sessionprovider-remoteuser/blob/master/AuthRemoteuser.body.php) as the base.

PR #710, opened 2017-02-05
Add and use package.json

nicolasleger:


nicolasleger:
@hangy Hi, you might review and integrate that in your frontend branch.

PR #725, opened 2017-02-11
Image optimization

hangy:
I ran ran most if not all SVGs through `svgo --multipass`. This yielded 75%-90% file size reduction for some images with no noticeable loss in quality.


stephanegigandet:
Good idea. Can we store the original SVGs somewhere though? The original files are useful if we want to edit them again. I know they would be in the github history, but it's less convenient.

Or another approach could be to keep the unoptimized SVG files in github, and just to run svgo in a deployment script or something. i.e. we keep the source, not the "binaries". What do you think?

hangy:
Sure, with the Foundation SASS branch, I added gulp anyways, so we could try integrating an image optimizer in that step easily. Just wasn't sure if everyone would want that. 😃

teolemon:
conflict

teolemon:
:kissing_heart:

PR #734, opened 2017-02-12
Import emb_codes to MongoDB

hangy:
Allows us to import `emb_codes` coordinates to MongoDB and then use a spatial query to create better results for MadeNearMe.


stephanegigandet:
Could we just change get_packager_code_coordinates($) to return an array_ref with [$lat, $lng] and use it, instead of duplicating the code?

hangy:
Good idea, I'll do that and update the PR.

hangy:
Merci! 👍

PR #756, opened 2017-02-19
Show incomplete products after edit

### LINKED ISSUE #4: OFF should show some random uncomplete product on "Edit Successful page"
OFF should show some random uncomplete product on "Edit Successful page" to foster more contributions.

hangy:
Closes #4


stephanegigandet:
Apparently $sample is only supported in Mongo 3.2 and onward, which is fine on production, but not on the old server I use for .net... I'm going to add an option to turn it off.

PR #830, opened 2017-05-29
Add Open Pet Food Facts specific strings

teolemon:


po/openpetfoodfacts/openfoodfacts.pot
```diff
@@ -0,0 +1,33 @@
+msgid  ""
```
hangy: File should not be in the openpetfoodfacts dir.

po/openpetfoodfacts/openbeautyfacts.pot
```diff
@@ -0,0 +1,45 @@
+msgid  ""
```
hangy: File should not be in the openpetfoodfacts dir.

teolemon:
File deleted for one, name changed for the other. @hangy @stephanegigandet

PR #844, opened 2017-06-28
L10n master

teolemon:


teolemon:
Yes, I'm on it, as fast as I can. The thing is I don't know why they happened in the first place.

PR #900, opened 2017-10-02
Issue/897 tests po files --> Make the build break if the English or French .po files have empty entries.

stephanegigandet:
I don't know why the English .po files have ended up with missing values, but it can cause a lot of problems, so I'm proposing that we make the build break if there are missing values.


teolemon:
How about storing and executing as a test ?
https://github.com/openfoodfacts/openfoodfacts-server/tree/ee77c5680f2c44c108ea9046184d418bb522a8ec/t

stephanegigandet:
I can try, I'm not sure if I can find out and access properly the location of the po files from the Perl tests executed through Travis.

.travis.yml
```diff
@@ -31,6 +31,8 @@ install:
   - sed -i -e 's|\$server_domain = "openfoodfacts.org";|\$server_domain = "off.travis-ci.org";|g' $TRAVIS_BUILD_DIR/lib/ProductOpener/Config2.pm
   - sed -i -e 's|\/home\/off|'$TRAVIS_BUILD_DIR'|g' $TRAVIS_BUILD_DIR/lib/ProductOpener/Config2.pm
 script:
+  - scripts/check_po_file.pl $TRAVIS_BUILD_DIR/po/common/en.po
```
hangy: Do you think this could be added to `package.json` instead, so that everyone could run it with `npm test` independently of Travis CI?

stephanegigandet: Probably, I'm really not familiar with travias and npm test, if you know how to do it, can you do it?

hangy: `npm test` also just runs some scripts, but those are configured in package.json, so that everyone could run the complete "test suite" (Perl syntax, prove -l, check_po_file.pl, etc.) by just running `npm test`. I'll have a look. 😄

PR #936, opened 2017-11-09
Fix for datatables bug

### LINKED ISSUE #935: Table sorting and filtering is broken for facets that have a taxonomy
Maybe related to issue #873 , and probably because there is an extra column for facets that have a taxonomy like categories and labels.

https://ru.openfoodfacts.org/packaging --> works
https://ru.openfoodfacts.org/labels --> does not work, JS Error:

Uncaught TypeError: Cannot read property 'mData' of undefined
    at HTMLTableCellElement.<anonymous> (datatables.min.js:102)

hangy:
This should fix #935. I don't have a running test instance of OFF right now, but the Perl syntax seems to be OK. Could you test this, please, @stephanegigandet? 😃

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -1549,6 +1549,11 @@ HTML
 		
 		my $tagtype_p = $Lang{$tagtype . "_p"}{$lang};
 		
+		my $extra_column_searchable = "";
+		if (defined $taxonomy_fields{$tagtype}) {
+			$extra_column_searchable .= ', { "searchable": false }';
+		}
+		
 		$initjs .= <<JS
 oTable = \$('#tagstable').DataTable({
 	language: {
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -1560,7 +1565,7 @@ oTable = \$('#tagstable').DataTable({
 	order: [[ 1, "desc" ]],
 	columns: [
 		null,
-		{ "searchable": false }
+		{ "searchable": false } $extra_column_searchable
```
nicolasleger: You might have forgotten to remove the content now inside the variable.

stephanegigandet: looks good to me

hangy: Thanks!

@nicolasleger The idea was that `$extra_column_searchable` is either empty or contains the entry for the third column. Not the cleanest way to implement this - I'll admit 😄 - but it should work. 😉

nicolasleger: Ok no problem so ! I was in doubt, sorry. Good 👍🙂

hangy: No problem. Thanks for checking! :)

PR #951, opened 2017-11-15
MongoDB connection string

hangy:
Use a connection string to connect to MongoDB. This allows us to use a MongoDB server that's located elsewhere than on localhost.


hangy:
Thanks 👍

PR #958, opened 2017-11-17
Pet-specific examples in the product editor

aleksejrs:
ingredients_text_example is from https://github.com/openfoodfacts/openfoodfacts-server/blob/master/po/common/common-web.pot , but I removed cocoa.

### DIFF
--- a/po/openpetfoodfacts/openpetfoodfacts.pot
+++ b/po/openpetfoodfacts/openpetfoodfacts.pot
@@ -57,7 +57,6 @@ msgctxt "labels_example"
 msgid "No sugar, No colorings"
 msgstr ""
 
-
 msgctxt "ingredients_text_example"
 msgid ""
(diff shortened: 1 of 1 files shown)


po/openpetfoodfacts/openpetfoodfacts.pot
```diff
@@ -36,3 +36,30 @@ msgstr ""
...
+
+msgctxt "categories_example"
+msgid "Dry adult cat food"
+msgstr ""
+
+msgctxt "labels_example"
+msgid "No sugar, No colorings"
+msgstr ""
+
+
```
teolemon: removing extra line

aleksejrs: Are you removing it, or did you want to say I should?

teolemon: suggestion to remove before I merge

Issue #941, opened 2017-11-11
Unicode issues at the graph page


aleksejrs:
https://world-ru.openfoodfacts.org/cgi/search.pl?action=process&tagtype_0=categories&tag_contains_0=contains&tag_0=en%3ACondensed%20milks&sort_by=unique_scans_n&page_size=20&axis_x=proteins&axis_y=carbohydrates&graph=1

Russian product names are displayed in broken Unicode, and links to products (with Russian for "product") are broken.

hangy:
Apparently something in https://github.com/openfoodfacts/openfoodfacts-server/blob/483dba98960aa529cb2bfb3f3efb488640af885a/lib/ProductOpener/Display.pm#L3326 does not mix well. Generally, `encode_json` works well with UTF-8 (ie. in https://github.com/openfoodfacts/openfoodfacts-server/blob/80089c5c194ecbb9d12fe5d8b233900f06179439/cgi/product_multilingual.pl#L155), but something in there messes up the encoding. We should probably aim to phase out manual string concatenation for the Highcharts data object, but the first step is to see what's wrong here. FWIW, not a translation issue.

PR #960, opened 2017-11-21
Unicode problem

hangy:
Apparently, the problem described in #941 was, that the `product_name` and `url` properties already were in UTF8 (not the other internal Perl format), `encode_json` also applied `encode_utf8` (or similar) internally, and thus the encoding got messed up.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -3305,7 +3305,7 @@ sub display_scatter_plot($$$) {
 				}
 				
 				defined $series{$seriesid} or $series{$seriesid} = '';
-				
+
 				# print STDERR "Display::search_and_graph_products: i: $i - axis_x: $graph_ref->{axis_x} - axis_y: $graph_ref->{axis_y}\n";
 					
 				my %data;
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -3325,8 +3323,7 @@ sub display_scatter_plot($$$) {
 				$data{url} = $url;
 				$data{img} = display_image_thumb($product_ref, 'front');
 				
-				defined $series{$seriesid} or $series{$seriesid} = '';
-				$series{$seriesid} .= encode_json(\%data) . ',';
+				$series{$seriesid} = \%data;
```
stephanegigandet: I haven't tested the code, but it seems that $series{$seriesid} will only contain data for 1 product, instead of data for all products in the graphs, this is in a while loop that goes over all products returned by the search query. If you want to make only one call to JSON::PP->new->encode at the end, then I'm guessing we need to push the data values in an array, instead of using a scalar for $series{$seriesid}.

hangy: Good point. I'll fix that later. Merci!

PR #968, opened 2017-11-29
Map for tags with Wikidata

### LINKED ISSUE #390: Display a map on facet page when an OpenStreetMap relationship exists
### Categories

http://world.openfoodfacts.org/category/fr:fitou
is linked to:
https://www.wikidata.org/wiki/Q470974
which is linked in turn to:
https://www.openstreetmap.org/relation/2727716

hangy:
Will display/add a map layer for taxonomized tag entries that have a Wikidata ID and whose Wikidata entry has a relation id.

This implementation is done by querying Wikidata and Overpass Turbo (OSM) dynamically to ensure we display the latest data from Wikidata and OSM.

Closes #390


CloCkWeRX:
Got any screenshots of the maps?

hangy:
> Got any screenshots of the maps?

Not right now. I'll see if I can take some later. It's basically the same map (object) as for packager codes with coordinates (ie. https://world.openfoodfacts.org/packager-code/fr-42-156-006-ec), but with a GeoJSON shape for the OSM object that's linked from Wikidata.

hangy:
image

teolemon:
👏 👏 👏

teolemon:
@stephanegigandet @offvince

PR #986, opened 2017-12-10
Disable barcode change for normal users

### LINKED ISSUE #979: Add feedback when user has no permission to change a barcode, or hide the field

hangy:
fixes #979


teolemon:
LGTM.

PR #1037, opened 2018-01-17
Add lang to multilingual input fields

hangy:
The actual language of the input/textarea will often be different from the HTML page's language, so that screen readers and spell checking may assume an incorrect language for multilingual products. What do you think?


CloCkWeRX:
That seems to make sense. Are there any cases where that could render blank or invalid content?

(https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang suggests that would upset things a little)

hangy:
> Are there any cases where that could render blank or invalid content?

I don't think so. As far as I can say, no such issue arose when I tested this. Adding a check or defaulting `$display_lc` to the current language should be easy enough, though. Thanks for the hint!

PR #1041, opened 2018-01-28
Detect quantities with "e"

hangy:
Tags quantities that contain stuff like "200 mg e", "300 ml e", and so on, but not "℮" as "quantity-contains-e".

I recently noticed that some products contain quantities like "500 g e", even though the packaging reads "500 g ℮". This new quality rule helps us find products like that, so that this issue can be fixed.

### DIFF
--- a/lib/ProductOpener/SiteQuality_off.pm
+++ b/lib/ProductOpener/SiteQuality_off.pm
@@ -1,7 +1,7 @@
 ﻿# This file is part of Product Opener.
 # 
 # Product Opener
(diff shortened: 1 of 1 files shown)


teolemon:
Nice addition. We need as many quality rules as we can. 
Also perhaps make the quantity parser able to ignore e ?

hangy:
> Also perhaps make the quantity parser able to ignore e ?

Is ℮ currently parsed in ingredients? I have thought about adding a label like "en:Quantity estimaed" automatically, if ℮ is present in the quantity field.

PR #1043, opened 2018-01-28
Remove dead links to /reuses content, which is 404

### LINKED ISSUE #859: 404 Failed to load resource Open https://world.openfoodfacts.org/reuses from https://world.openfoodfacts.org/discover
404 Failed to load resource: the server responded with a status of 404 (Not Found) https://world.openfoodfacts.org/reuses - 1 time - click to see details
Steps to reproduce
Open https://world.openfoodfacts.org/discover
Open https://world.openfoodfacts.org/reuses

CloCkWeRX:
Fix #859

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,11 @@ https://github.com/openfoodfacts/openfoodfacts-server/issues
 You can help translate Open Food Facts and the app at (no technical knowledge required, takes a minute to signup): <br>
 https://crowdin.com/project/openfoodfacts
 
+#### Updating translations for static content
+A number of the translations are maintained externally, see `scripts/update_texts_from_wiki.pl` to sync translation data.
+
+Specific pages are defined in the relevant `Config.pm` `wiki_texts` variables.
+
 ### Who do I talk to? ###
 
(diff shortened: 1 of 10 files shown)


hangy:
@stephanegigandet Shouldn't some of these texts (some of the languages don't seem to be linked) be loaded from the Wiki? Maybe they should be removed from git and added to `.gitignore` to make it more obvious? tbh, I regularly get confused by this. 😃

https://github.com/openfoodfacts/openfoodfacts-server/blob/8a2adef2adba29c1be644013307a2db498736806/lib/ProductOpener/Config_off.pm#L247-L259

CloCkWeRX:
I've made the corresponding wiki page changes as well that mirror the github commits. For most of the languages, it was obvious, but Greek and Japanese probably need a native speaker to double check.

There seems to be a bit of drift between wiki and git; so I'll update the PR with sync'd text shortly.

stephanegigandet:
Indeed we should remove the texts from git as they are overwritten by the script that gets them from the wiki.

PR #1044, opened 2018-01-28
Add nutrients

### LINKED ISSUE #64: Vitamin K1 (Phylloquinone)  and Choline not understood
Choline (http://en.wikipedia.org/wiki/Choline)
Vitamin K1 (http://en.wikipedia.org/wiki/Phylloquinone)

…aren't well detected by the nutrition facts table, as you can see here : 

http://fr.openfoodfacts.org/produit/3041090796562/bledilait-croissance-3-bledina

CloCkWeRX:
Fixes #64 

Not sure if this is all that is required, or if there are other places in the code that need new values.

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -316,6 +316,11 @@ nutrition-score-fr-
 nutrition-score-uk-
 glycemic-index-
 water-hardness-
+choline-
+phylloquinone-
+beta-glucan-
+inositol-
+carnitine-
 )
 ],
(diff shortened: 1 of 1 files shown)


CloCkWeRX:
Ah, I did see the various structures for us/ru/etc; but wasn't sure what the right syntax was for that, thanks. Will update shortly.

CloCkWeRX:
@hangy, this might be good to go

hangy:
Thank you :)

PR #1057, opened 2018-02-09
Add OCR links

hangy:
For admin users, adds links to OCR/Cloud Vision results, if they exist.


teolemon:
Very useful. 
- Being able to generate them and store the results if they don't exist
- Being able to leverage them for edit suggestions

Issue #185, opened 2016-01-07
Product with no created_t field


stephanegigandet:
http://fr.openfoodfacts.org/produit/3017760038409/lulu-la-barquette-fraise-kraft-foods

created_t field was removed between revision 11 and 12, cause unknown.

need to check if other products are affected.

hangy:
There are some products that don't contain the `created_t` field, at all. They don't contain a barcod either.
```
> db.products.find( { created_t: { $exists: false } } ).count();
28
> db.products.find( { created_t: { $exists: false } }, { code: 1 } );
{ "_id" : ObjectId("57d139e83296350bfc5a9271") }
{ "_id" : ObjectId("57d13a123296350bfc5a9272") }
{ "_id" : ObjectId("57d13ad43296350bfc5a9273") }
{ "_id" : ObjectId("57d13b293296350bfc5a9274") }
{ "_id" : ObjectId("57d13b343296350bfc5a9275") }
{ "_id" : ObjectId("57d13b5c3296350bfc5a9276") }
{ "_id" : ObjectId("57d13b5c3296350bfc5a9277") }
{ "_id" : ObjectId("57d13b5c3296350bfc5a9278") }
{ "_id" : ObjectId("57d13b7b3296350bfc5a927a") }
{ "_id" : ObjectId("57d6728b3296352f837c8411") }
{ "_id" : ObjectId("57d672d73296352f837c8412") }
{ "_id" : ObjectId("57d674553296352f837c8413") }
{ "_id" : ObjectId("57d6750a3296352f837c8414") }
{ "_id" : ObjectId("57d6751f3296352f837c8415") }
{ "_id" : ObjectId("57d676353296352f837c8416") }
{ "_id" : ObjectId("57d676453296352f837c8417") }
{ "_id" : ObjectId("57d676503296352f837c8418") }
{ "_id" : ObjectId("57d6795e3296352f837c841a") }
{ "_id" : ObjectId("57d7be8f32963548284f25c1") }
{ "_id" : ObjectId("57d7bea33296354885634da1") }
{ "_id" : ObjectId("57d7becf3296354885634da2") }
{ "_id" : ObjectId("57d7bfbc3296354885634da3") }
{ "_id" : ObjectId("57d7c0a93296354885634da4") }
{ "_id" : ObjectId("57d7c0cc3296354885634da5") }
{ "_id" : ObjectId("57d7c2973296354885634da6") }
[...]

PR #1061, opened 2018-02-10
Check for missing/empty code or created_t

hangy:
This could help in detecting bugged products for #185.


hangy:
> Looks good! Note if the code is missing, we won't be able to get this product from Product Opener, we probably need to wipe them in MongoDB directly.

Yup - I thought this might be helpful to see whether we have any messed up .sto files, though?

PR #1081, opened 2018-02-15
Structured Logging

hangy:
I'd like to propose moving to using [Log::Any](https://metacpan.org/pod/Log::Any) for structured logging instead of the current `print STDERR` statements.

# Advantages
## Automatic logging of calling module
Instead of having to log `print STDERR "some.pl Doing $whatever\n";`, we can log `$log->debug("Doing", { whatever => $whatever });`. While this might seem a bit more verbose, the advantage of the fact that the logging libraries can automatically log the calling module does outweigh the more verbose syntax, in my opinion.

## Timestamps
The loggers can automatically timestamp log entries.

## Multiple output streams
Different log levels can be sent to different/multiple targets. Even if we send everything to a file (rolling files possible with [Log::Dispatch::FileRotate](https://metacpan.org/pod/Log::Dispatch::FileRotate)), errors could also be dispatched to Slack with [Log::Dispatch::Slack](https://metacpan.org/pod/Log::Dispatch::Slack) and the structured data could be collected in [Graylog](https://metacpan.org/pod/Log::Log4perl::Appender::Graylog) or an alternative.

## Controllable log levels
Log levels can be switch on and off using a config file, so that it would be possible to have trace/debug log calls in the code and enable it when necessary.

# Disadvantages
## More complex logging code
```
print STDERR "some.pl Doing $whatever\n";
```
vs
```
$log->debug("Doing", { whatever => $whatever }) if $log->is_debug();
```
(Note that the call to `is_debug` avoids doing complex calls and object initialization if a log level is disabled completely.)

This is a bit more complex, but I think it's still easy to read.

## Performance degradation
It is very likely that using a logging library and abstraction layer will cost some performance. However, logging libraries are usually developed and tested in a way that they won't destroy process performance unless you have ie. really active trace logging.

# Example log output
Example output using `Log4perl` file appender
```
[308419] /opt/product-opener/lib/ProductOpener/Store.pm 124 ProductOpener.Store - get_urlid {in => "To-be-completed",out => "to-be-completed"}
[310008] /opt/product-opener/cgi/display.pl 47 ModPerl.ROOT.ModPerl.Registry.opt_product_2dopener_cgi_display_2epl - before analyze_request {query_string => ""}
[310010] /opt/product-opener/cgi/display.pl 51 ModPerl.ROOT.ModPerl.Registry.opt_product_2dopener_cgi_display_2epl - after analyze_request {blogid => undef,query => undef,tagid => undef,urlid => undef,urlsdate => undef,user => undef}
[310017] /opt/product-opener/lib/ProductOpener/Store.pm 124 ProductOpener.Store - get_urlid {in => "To-be-completed",out => "to-be-completed"}
```

# Final remarks
I'm writing this lenghty PR, because my experience tells me that better logging and the ability to control what is logged are worth it, but this is a paradigm shift for ProductOpener, and this change touches a lot of files. Because of that, I'd love to get your input on the topic. 😄


hangy:
@TravisBuddy Thanks, mate! 😉

Issue #1028, opened 2018-01-03
Usability: Support faster, mission oriented attribute editing (ie: product names)


CloCkWeRX:
I'm trying to name a lot of things in rapid succession:
https://world.openfoodfacts.org/state/product-name-to-be-completed/country/australia/4

The current steps are:
 * Ctrl + click on a bunch of likely candidates to view
 * Wait for various tabs to load
 * Swap through each and whack 'edit'
 * Wait for edit UIs to load
 * Key in name, brand, plain description OR close because it's not answerable
 * Scroll all of the way down, save

This takes a reasonable amount of time; and the usability could be improved.

Ideas to do so:
 * On the view product page, prompt users with mission oriented single answer questions and a "Submit" or "I don't know/Skip" control. Example of a system that does this well: Maproulette
 * Make it easier to select multiple states in the search UI (Right now it isn't easy to work out how to do state:has-photos AND product-name-missing, country: XYZ)
 * On the edit UI, style the save control as position: fixed; bottom: 0 so that it is always accessible.

CloCkWeRX:
Similar: #1005

CloCkWeRX:
Mockup of what "Affixed save controls on edit form" could look like:
```
<div style="position: fixed; bottom: 0; width: 100%; border-top: 1px solid #eee; background-color: white; z-index: 100; padding-top: 10px;"><input id="comment" name="comment" placeholder="Changes summary" value="" type="text" class="text" style="width: 70%; float: left; margin-right: 5px"><input type="submit" name=".submit" value="Save" class="button small" style="float: left"></div>
```
<img width="1680" alt="screen shot 2018-01-04 at 12 51 47 am" src="https://user-images.githubusercontent.com/365751/34523932-7e127694-f0e9-11e7-81d1-fb199c2e6bde.png">

PR #1099, opened 2018-02-18
Sticky edit comment

### LINKED ISSUE #609: Keep the edit and save button always visible in the edit page (responsive mode)
Keep the edit and the save buttons always visible in the product and edit pages. 
Currently you have to scroll down to the bottom of the page to save.

image

This is really important for the new mobile app that lets you edit products.

hangy:
Code by @CloCkWeRX https://github.com/openfoodfacts/openfoodfacts-server/issues/1028#issuecomment-355023253

Closes #609

### DIFF
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -1786,15 +1786,24 @@ HTML
 	. hidden(-name=>'code', -value=>$code, -override=>1)
 	. hidden(-name=>'action', -value=>'process', -override=>1);
 	
+	$html .= '<div style="position: fixed; bottom: 0; width: 100%; border-top: 1px solid #eee; background-color: white; z-index: 100; padding-top: 10px;">';
+
 	if ($type eq 'edit') {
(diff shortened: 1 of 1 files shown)


stephanegigandet:
It looks good on mobile, but on desktop it's a bit odd, it may be just a habit issue, I'm not used to fixed stuff on desktop.

image

Otherwise code looks good.

hangy:
Maybe browser dependent? For me, it looked exactly like @CloCkWeRX screenshot on desktop (https://github.com/openfoodfacts/openfoodfacts-server/issues/1028#issuecomment-355023253). Let's do some cross-browser tests first. :)

stephanegigandet:
I didn't run the exact code in the fix so that's why the Save button looks a bit different. What's troubling me (a bit, I can get used to it) is the fact that if you scroll all the way down, the save button goes in the page footer. But don't worry about it.

hangy:
Closing old non-reviewed PRs.

PR #1110, opened 2018-02-23
add optional fields parameter to JSON API

### LINKED ISSUE #554: Create a JSON version of the JQM compact search
Rationale: 10X improvement in performance for the Android Search
URL: http://world.openfoodfacts.org/cgi/search.pl?search_terms=nutella&search_simple=1&action=process&jqm=1

### LINKED ISSUE #623: Create a compact version of the product JSON
We need a compact version of the product JSON to ensure the mobile app works in poor network conditions and in intensive scanning scenario.
Same output as https://github.com/openfoodfacts/openfoodfacts-server/issues/554 (name, brand, weight, very small image resolution url) but as a plus, would it be possible to add the nutriscore to the output.
We have a new ```Quick view```mode that lets you lookup many products very quickly in a row, showing the Nutriscores, and the result should be instant.

image

stephanegigandet:
add optional fields parameter to JSON API to retrieve and search products. closes #623, closes #554

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -395,28 +395,34 @@ sub analyze_request($)
 		
 	print STDERR "analyze_request : query_string 1 : $request_ref->{query_string} \n";
 	
+	# Process API parameters: fields, formats, revision
+	
 	# API calls may request JSON, JSONP or XML by appending .json, .jsonp or .xml at the end of the query string
 	# .jqm returns results in HTML specifically formated for the OFF mobile app (which uses jquerymobile)
 	# for calls to /cgi/ actions (e.g. search.pl), the format can also be indicated with a parameter &json=1 &jsonp=1 &xml=1 &jqm=1
 	# (or ?json=1 if it's the first parameter)
 	
-	# first check for the rev parameter (revision of a product)
+	# first check parameters in the query string
+
+	foreach my $parameter ('fields', 'rev', 'json', 'jsonp', 'jqm','xml') {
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -395,28 +395,34 @@ sub analyze_request($)
...
-	foreach my $parameter ('rev', 'json', 'jsonp', 'jqm','xml') {
+	# first check parameters in the query string
+
+	foreach my $parameter ('fields', 'rev', 'json', 'jsonp', 'jqm','xml') {
 	
-		if ($request_ref->{query_string} =~ /(\&|\?)$parameter=(\d+)/) {
-			$request_ref->{query_string} =~ s/(\&|\?)$parameter=(\d+)//;
+		if ($request_ref->{query_string} =~ /(\&|\?)$parameter=([^\&]+)/) {
+			$request_ref->{query_string} =~ s/(\&|\?)$parameter=([^\&]+)//;
 			$request_ref->{$parameter} = $2;
```
hangy: Could this cause any problem if `$request_ref->{json}` were set to a non-numeric value? The check is in https://github.com/openfoodfacts/openfoodfacts-server/blob/16d18d199414fc430ce44acc51784cc54e7cb4c1/lib/ProductOpener/Display.pm#L4563 and could have to try evaluate the string as boolean value?

stephanegigandet: I think any non-numeric value would be treated as a true value

PR #1115, opened 2018-02-25
plant milks/dairies

### LINKED ISSUE #73: All vegetal milks are part of the dairy products (in fr)
Les laits vagétaux font partie des Laits ainsi que des Produits laitiers. Faut-il également garder l'inclusion à la catégorie Lait ?

hangy:
This quality asset allows users to find plant milks that are also categorized as dairies, because of the former categories taxonomy. In conjunction with the updated taxonomy, I think this should be enough to close #73.


teolemon:
Nice one, @hangy

PR #1118, opened 2018-02-25
Unit Test for translated logos

### LINKED ISSUE #1116: Add a test for missing logos
We deployed the new translations, and some logos are missing (example:https://hu.openfoodfacts.org/)
Check the file exists if the logo string is translated.

hangy:
Closes #1116


hangy:
> So do they all exist?

Actually no, that will need to be fixed in Crowdin as an authoritative source for translations.

PR #1121, opened 2018-02-26
add cigarette brands

teolemon:
add a list for cigarettes. the facet needs to be written.


teolemon:
@hangy can you add a quality facet based on this?

teolemon:
Can we merge ?

PR #1153, opened 2018-03-15
Delta Product Export

### LINKED ISSUE #153: Add daily database to download
Hello guys,

I would like get your database, for avoiding to download all database every day, is it possible for you to provide a daily "delta" database ?

hangy:
Add delta export for the products collection. Closes #153

Dumps all products with a `last_modified_t` in the last day or since the last run of the script into the /data directory. Using [`mongoimport --mode=upsert`](https://stackoverflow.com/a/29776292/11963), this could be used to refresh a previous import.

### DIFF
--- a/scripts/mongodb_dump.sh
+++ b/scripts/mongodb_dump.sh
@@ -1,9 +1,24 @@
 #!/bin/sh
 
 cd /home/off/html
+
 mongodump --collection products --db off
 tar cvfz data/openfoodfacts-mongodbdump.tar.gz dump
 pushd data/ > /dev/null
 sha256sum openfoodfacts-mongodbdump.tar.gz > sha256sum
 md5sum openfoodfacts-mongodbdump.tar.gz > md5sum
+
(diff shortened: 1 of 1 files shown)


stephanegigandet:
So I'm not used to write to shell script, but that's going to create one file  products_delta.json.gz that is going to be overwritten every day, right?
That would work for things that are actually going to request the delta every day, but maybe we could keep exports for more days.

hangy:
> So I'm not used to write to shell script

Me neither :grin: 

> but that's going to create one file products_delta.json.gz that is going to be overwritten every day, right?
> That would work for things that are actually going to request the delta every day, but maybe we could keep exports for more days.

Yes, I thought about that, too. We'd need to have some kind of index for those files, though.

hangy:
@stephanegigandet I modified the PR to create a `data/delta` directory, keep the last 14 delta files, and build an `index.txt` which contains the currently existing delta files. What do you think? 😄

hangy:
In that case, I'll merge this. 👍

PR #1182, opened 2018-04-04
[WIP] Webpack

hangy:
Use webpack to bundle JavaScript/CSS etc. instead of gulp. Also eleminates Bower.


hangy:
> > On the homepage:
> > 
> > Is missing:
> > 
> > css/dist/app.css
> > select2.min.css
> 
> I think that's from the latest merge, where a lot of stuff was changed in `master` unexpectedly (to me). Will fix. :)

Actually, I cannot find this referenced in the code right now. Where did you find it?

NerOcrO:
> > 2 JS errors:
> > ReferenceError: $ is not defined
> 
> Yes, there's probably some stuff in the dynamically generated JavaScript that still references jQuery via `$`. That's an open TODO.

Can you resolve it?

> > All the JavaScript are in devDependencies (package.json), why?
> 
> We're not really splitting dev/prod right now, but to quote (emphasis mine):
> 
> > `devDependencies` should contain packages which are used during development or which are used to **build your bundle**
> 
> As everything - including JavaScript libraries - is simply bundled, I think it should make sense to list them as `devDependencies`.

`dependencies` are required to run, `devDependencies` only to develop, e.g.: unit tests, CoffeeScript to JavaScript transpilation, minification, ...
https://docs.npmjs.com/files/package.json#dependencies
https://docs.npmjs.com/files/package.json#devdependencies

Please, move the JavaScript to the right place.
Examples:
- https://github.com/expressjs/express/blob/master/package.json#L29
- https://github.com/hapijs/hapi/blob/master/package.json#L17

hangy:
> > > `devDependencies` should contain packages which are used during development or which are used to **build your bundle**
> > 
> > As everything - including JavaScript libraries - is simply bundled, I think it should make sense to list them as `devDependencies`.
> 
> `dependencies` are required to run, `devDependencies` only to develop, e.g.: unit tests, CoffeeScript to JavaScript transpilation, minification, ...
> https://docs.npmjs.com/files/package.json#dependencies
> https://docs.npmjs.com/files/package.json#devdependencies

That's the point. ProductOpener is not a runnable node app and by bundling all JS and CSS we do not need them in ` dependencies`.

> Please, move the JavaScript to the right place.
> Examples:
> 
> https://github.com/expressjs/express/blob/master/package.json#L29
> https://github.com/hapijs/hapi/blob/master/package.json#L17

They already are in the right place. `express` and `hapi` are used to build Node.js based web apps and their dependencies are JS libraries required to run their binaries. We don't have that.

> > > > On the homepage:
> > > > Is missing:
> > > > css/dist/app.css
> > > > select2.min.css
> > > 
> > > 
> > > I think that's from the latest merge, where a lot of stuff was changed in `master` unexpectedly (to me). Will fix. :)
> > 
> > 
> > Actually, I cannot find this referenced in the code right now. Where did you find it?
> 
> https://github.com/openfoodfacts/openfoodfacts-server/blob/master/lib/ProductOpener/Display.pm#L5366
> 
> * app.css
> * select2.min.css
> * opensearch.pl

That's on the `master` branch, though. 😄

NerOcrO:
Yes the website works but it is not usable!
We don't care if it's a node app or a simple website, what is important is the semantics:
- dependencies: we need for production
- devDependencies: we need just for test/lint...

And please, be careful when you are merging and testing before tell me to reviewing, thanks.
Like I said two times, is missing this three lines: https://github.com/openfoodfacts/openfoodfacts-server/blob/master/lib/ProductOpener/Display.pm#L5366
Fix it please.

hangy:
> We don't care if it's a node app or a simple website, what is important is the semantics:
> 
> dependencies: we need for production
> devDependencies: we need just for test/lint...

I don't quite see the purpose of this, but will create a PR for this. 👍 

> Like I said two times, is missing this three lines: https://github.com/openfoodfacts/openfoodfacts-server/blob/master/lib/ProductOpener/Display.pm#L5366

Now I get what you mean. 😆 Will re-add OpenSearch, but `app.css` is not used any more, because CSS is loaded through the webpack module. select2 is actually here: https://github.com/openfoodfacts/openfoodfacts-server/blob/2bd2123cb1de049189a8ba6216aa1ca56333a08e/html/js/src/display.js#L34

NerOcrO:
> Now I get what you mean. laughing Will re-add OpenSearch, but `app.css` is not used any more, because CSS is loaded through the webpack module. select2 is actually here:
> 
> [openfoodfacts-server/html/js/src/display.js](https://github.com/openfoodfacts/openfoodfacts-server/blob/2bd2123cb1de049189a8ba6216aa1ca56333a08e/html/js/src/display.js#L34)
> 
> Line 34 in [2bd2123](/openfoodfacts/openfoodfacts-server/commit/2bd2123cb1de049189a8ba6216aa1ca56333a08e)
>  import('select2').then(() => {

Ok, I understand but I don't have any css:
https://screenshots.firefox.com/ngS6d4iuv4EghhOR/world.openfoodfacts.localhost

This is my step:
- yarn install
- yarn run dev
- F5
What am I missing?

hangy:
Can you try
```
npm install -g pnpm
pnpm install
pnpm run dev
```

NerOcrO:
The same thing.
I don't see `html/css/dist` for example.
In HTML code, I haven't `<link href="...css" rel="stylesheet">` too.
Is that normal?

NerOcrO:
@hangy?

hangy:
I'm out oft town. Will have a look when I get back.

VaiTon:
Ehm @hangy, what do we want to do with this PR? 😅

hangy:
> Ehm @hangy, what do we want to do with this PR? 😅

I was thinking of re-starting this with a simpler approach. I initially tried to move all JS to actual JavaScript files and out of the HTML source, but the current OFF source is just too messy with dynamically generated JavaScript to do it in one go ... 🤔

VaiTon:
> 
> 
> > Ehm @hangy, what do we want to do with this PR? 😅
> 
> I was thinking of re-starting this with a simpler approach. I initially tried to move all JS to actual JavaScript files and out of the HTML source, but the current OFF source is just too messy with dynamically generated JavaScript to do it in one go ... 🤔

Yeah you're right. Maybe for now we should just use webpack as build system for the things that are on gulp.

PR #1207, opened 2018-05-11
Update MongoDB connection handling

hangy:
I updated the referenced MongoDB version and moved connection handling to a new `ProductOpener::Data` module, so that we can take advantage of the built-in connection pooling. This could solve the MongoDB server dying repeatedly.

## Related issues and discussion
#1081 was used as base branch because of the included cleanup
#666 deprecated methods were replaced with alternatives
#939 allowDiskUse was enabled

Issue #939, opened 2017-11-10
Allow MongoDB to sort on disk


hangy:
https://world.openfoodfacts.org/ingredients fails ("no products"), even though there are products

MongoDB log tells us the following

> 2017-11-10T23:14:27.785+0100 I COMMAND  [conn54625] command off.products command: aggregate { aggregate: "products", pipeline: [ { $match: {} }, { $unwind: "$ingredients_tags" }, { $group: { count: { $sum: 1 }, _id: "$ingredients_tags" } }, { $sort: { count: -1 } } ], cursor: {} } keyUpdates:0 writeConflicts:0 exception: Sort exceeded memory limit of 104857600 bytes, but did not opt in to external sorting. Aborting operation. Pass allowDiskUse:true to opt in. code:16819 numYields:2960 reslen:200 locks:{ Global: { acquireCount: { r: 6026 } }, Database: { acquireCount: { r: 3013 } }, Collection: { acquireCount: { r: 3013 } } } protocol:op_query 3747ms

https://www.mkyong.com/mongodb/mongodb-sort-exceeded-memory-limit-of-104857600-bytes/

hangy:
Do we want MongoDB to allow this? A quick search didn't show up any obvious downsides.

PR #1208, opened 2018-05-11
Update MongoDB connection handling (master)

hangy:
I updated the referenced MongoDB version and moved connection handling to a new `ProductOpener::Data` module, so that we can take advantage of the built-in connection pooling. This could solve the MongoDB server dying repeatedly.

Backport of #1207, so it can be applied to master without `Log::Any`
#666 deprecated methods were replaced with alternatives
#939 allowDiskUse was enabled


hangy:
Dropped in favour of #1207, because #1081 updated the master branch.

PR #1215, opened 2018-05-27
UNSPSC as a category system for Open Products Facts

teolemon:
## Description

Add the raw UNSPSC taxonomy of e-commerce in its opendata, English version. 
It is hierarchized, includes categories for food, petfood and cosmetics (for the future).
It would need to be translated in our format.

The legal status of translation in unclear, and they are not easily accessible. We could go for our own translations, or try to get them released in machine legible format, with a clear open data licence overtime.
Meanwhile, that would be a very good first step for Open Products Facts.


hangy:
> The legal status of translation in unclear, and they are not easily accessible.

[This site](https://catalog.data.gov/dataset/unspsc-codes) provides the file under the terms of the ODbL.

> It would need to be translated in our format.

I don't see an advantage of committing this to the OFF code repository as-is. It will need to be converted to another format, but we would be having this 11 MiB BLOB in the code repository for all time. Maybe adding it to [outside-data](https://github.com/openfoodfacts/outside-data) would be a good intermediary step?

teolemon:
It would need to be converted to GiganFormat ;-)
Having autosuggestions on taxonomies would be quite nice.

PR #1221, opened 2018-05-31
Update MongoDB retry handling (master)

hangy:
Instead of having a custom retry logic for every query, I moved it to the `ProductOpener::Data` module, so that it can be reused properly. I also added `Action::Retry` to allow for a slight delay between retries, and implemented `Action::CircuitBreaker` on top of it, so that we don't query an already unresponsive MongoDB server.

Backport of #1220, so it can be applied to master without Log::Any


hangy:
@stephanegigandet The idea is that `Action::Retry` will retry the given anonymous `sub` upon failure with a backoff that can be configured (in code). 

`Action::CircuitBreaker` "opens" the circuit after a certain amount of failures for a configurable amount of time. During that timeframe, all other tries to access MongoDB will err /without hitting MongoDB/. If the database is overloaded and produces a lot of timeouts, that can be quite often be solved by not querying the DB for some time. So by employing the [circuit breaker pattern](https://martinfowler.com/bliki/CircuitBreaker.html), we basically try to give it a chance to recover by not passing any new queries to the DB server (while it is unresponsive anyways).

> And then if it still fails, the circuit becomes open? For how long?

10 seconds by default, but that can be changed.

> How does this work with mod_perl? The circuit being open or not will live for several client requests, as long as they are sent to the same mod_perl process?

Yes, those seem to be global/static variables, if I understood it correctly.

By wrapping all of that into `execute_query`, we can reuse the same retry-logic for all MongoDB queries.

stephanegigandet:
Thanks for the details, that sounds great. 10 seconds sounds fine. Have you tested that it works as expected?

hangy:
I wasn't able to fully test it due to my Docker container coming back too soon. :grin: I could, howerver, check that the queries still worked, that the retry works, and the circuit breaker works on it's own.

PR #1222, opened 2018-06-02
Add delta files to data docs

hangy:
## Description

Add hint on where to find the delta files and how to process them.

This is a follow-up to #1153, because we have not announced how to process the delta files, yet.


teolemon:
Looks good to me

PR #1223, opened 2018-06-02
Update data URLs in data.en.html

hangy:
The current syntax of "Links :" with spaces around the separator looks weird, and having a separate paragraph for each link is superfluous. Therefore, I would like to propose using a definition list instead. This should also help translators in Crowdin.

I also replaced the domain `world.openfoodfacts.org` with `static.openfoodfacts.org` in this page, because the content should be identical for all regions/languages, and thus does not need translation.


teolemon:
LGTM

Issue #1204, opened 2018-04-30
Enable a token based auth for the mobile app


teolemon:
### What
- Enable a token based auth for the mobile app so that we don't have to store the password on the mobile devices and send them everytime.

### Part of
- #8919

hangy:
As you may have noticed, I had started something like that in the [OpenID Connect branch](https://github.com/openfoodfacts/openfoodfacts-server/tree/oidc). For some reason, though, Perl always SEGFAULTs in one of the used Perl modules. It should be easy to finish this if anyone has an idea how to fix those Perl modules. 👍

hangy:
I revisited this idea last weekend. Instead of re-implementing OIDC (as there seems to no good OIDC Identity Provider library for Perl - and is out of scope for this project), we could integrate [hydra](https://github.com/ory/hydra) with ProductOpener to get a verified OIDC IdP, and see whether the APIs users can be [verified](https://auth0.com/docs/api-auth/tutorials/verify-access-token) using [Crypt::JWT](https://metacpan.org/pod/Crypt::JWT).

hangy:
Closing as a duplicate of #1596

PR #1230, opened 2018-06-05
[WIP] OpenID Connect

hangy:
Add an OpenID Connect server, based on OIDC::Lite.

Can be used to add token based authorization to the API, see #1204


hangy:
I'm ambandoning this approach and the `oidc` branch. It still segfaults on me (even on several different Linux VMs, Docker images, ...), and the OIDC::Lite module is meant to be integrated as a PSGI middleware, and thus cannot be integrated into the current ProductOpener structure easily.

PR #1235, opened 2018-06-10
Taxonomy IDs

### LINKED ISSUE #405: Provide stable ids for Facets
As we're linking facets to Wikidata, slugs might change overtime.
Having the equivalent of Qids would prevent issues

hangy:
A proposal for adding unique stable IDs to taxonomy tags.

- When reading taxonomies from the `$tagtype.txt` file, import UUIDs from lines beginning with `_uuid`.
- If a taxonomy tag does not have any UUID after the import, generate a new UUID.
- Output all UUIDs into `$tagtype.result.txt`, so they can be exported to the Wiki.
- Maps UUIDs to the `$canon_tagid` and vice-versa.
- Updated `canonicalize_taxonomy_tag` to support lookup by the tag UUIDs.
- Multiple UUIDs for a tag are supported, so that tags can be merged.

Closes #405

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -48,6 +48,7 @@ requires 'Email::Stuffer';
 requires 'Text::CSV', '>= 1.99, < 2.0';
 requires 'Text::Fuzzy';
 requires 'File::Copy::Recursive';
+requires 'Data::GUID';
 
 # Logging
 requires 'Log::Any', '>= 1.705';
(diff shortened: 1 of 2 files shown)


teolemon:
Wikidata prefers stable IDs to changing ones based on slug text

teolemon:
@stephanegigandet UUIDs should not be the primary way in Product Opener due to the behaviours you mentionned, but having them is good practise for external linking.

VaiTon:
@stephanegigandet any thoughts on this?

PR #1318, opened 2018-08-11
Add Up to date apk

teolemon:
Add Up to date apk (2.8.7)


hangy:
These are now outdated. I think it would be better to link to an external source instead. Adding the binaries/apks to this repository in order to upload them to off/obf is not a good (but rather slow) way of handling releases of the app.

PR #1326, opened 2018-08-29
Nutrition table with CLDR::Number

hangy:
Format (most) nutrition value entries using CLDR::Number

I noticed that the nutrition table when showing a single product does not display the data in a localized way. This is an attempt to improve it by displaying decimals and percentages using the current locale.


hangy:
> Aren't we losing the specification of the number of digits?

The `decimal_formatter` defaults to 2 digits, but that can be adjusted with `my $decf = $cldr->decimal_formatter( maximum_fraction_digits => 10 );`. However, the default of two digits is roughly the same (and arguably easier to read) as `sprintf("%.2e", ..)`, so I figured it might be ok.

stephanegigandet:
Hi @hangy ,
I tried on the .net server and I'm getting weird results.
e.g. in French:

image

and in English:

image

There's something weird with the % -> they are multiplied by 100. 

And the non % values don't seem changed, I would have expected a comma for French : 3,08 instead of 3.08.

Here is German:

image

hangy:
That's weird. I'm pretty sure I successfully tested the decimal point formatting in a Perl REPL, but I'll need to retest to verify. Thanks for testing!

As for the percentage values, I guess I will just need to divide them by 100, because it probably needs fractions for formatting.

hangy:
@stephanegigandet I honestly don't quite understand what might be wrong regarding the decimal point. When I run the following script in `perl -CS`, the output looks fine.

```
use CLDR::Number;
use CLDR::Number::Format::Decimal;
use CLDR::Number::Format::Percent;

my $cldr = CLDR::Number->new(locale => 'fr');
my $decf = $cldr->decimal_formatter;
my $perf = $cldr->percent_formatter( maximum_fraction_digits => 2 );

print $decf->format(10.25);
print "\n";
print $perf->format(0.1337);
```

Output:
```
10,25
13,37 %
```

I don't have a running copy of ProductOpener right now, so I can't test the integrated solution. :disappointed:

stephanegigandet:
I'm making a few changes, there are more places where $decf->format() needs to be applied so that we get a consistent formatting in all the values in the nutrition table.

PR #1362, opened 2018-09-19
Use SameSite cookie for CSRF defense

### LINKED ISSUE #1147: invalid CSRF token on account creation from an iphone
reported by an user: "Bonjour, j’essaye de créer un compte open food facts à partir de mon iphone. Mais j’ai toujours un message d’erreur : invalid CSRF token."

hangy:
## Description
Removed all references to `WWW::CSRF` and replaced it with `SameSite=Lax` cookies.

Due to browser enhancements, manual token checks are not the single most useful defense against CSRF [any more](https://scotthelme.co.uk/csrf-is-dead/). Also see [OWASP](https://www.owasp.org/index.php?title=Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet&oldid=242455#Alternate_CSRF_Defense:_SameSite_cookie_attribute).

I believe that this fixes #1147 while retaining security precautions.

### DIFF
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -51,8 +51,6 @@ use Encode;
 use JSON::PP;
 use Log::Any qw($log);
 
-use WWW::CSRF qw(CSRF_OK);
-
 ProductOpener::Display::init();
 
 if ($User_id eq 'unwanted-user-french') { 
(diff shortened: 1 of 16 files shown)


stephanegigandet:
I'm not sure I understand all the implications.

In production, I actually removed the CSRF token check as it made creating accounts on the Android app impossible.

if (not ($csrf_token_status eq CSRF_OK)) {
                # disabling as it is causing issue on android
                #display_error('Invalid CSRF token', 403);
        }

Would this create a similar issue?

Also if there is samesite=lax, then GET requests on forms will have the cookie right? All the forms on OFF currently accept GET requests.

hangy:
No, this is to avoid the account creation bug while still avoiding CSRF. According to https://www.owasp.org/index.php/SameSite it will avoid the cookie to be sent when `POST`ed from an outside domain, but it will send it for `GET` requests. This should not create the account creation bug. I actually made this PR to fix #1147.

PR #1363, opened 2018-09-19
Fix HTML syntax error in translations

hangy:
`<h2>` heading was not closed correctly


teolemon:
You can directly edit the source file and crowdin should fix all the translations

PR #1382, opened 2018-09-25
Month and Weekday translations from CLDR

### LINKED ISSUE #1378: Don't store months and days as arrays in translations
Would it be possible to use standard librairies to get months and days already translated in all languages ?
Can we kill the arrays that are currently there to do the job ?

hangy:
Fixes #1378


hangy:
> But does it work for all locales? Do they all need to be installed on the server first somehow?

We're already using the same locale module to display the last edit date and the editing history. As far as I can tell, it should [bundle](https://github.com/houseabsolute/DateTime-Locale/tree/v1.22/share) all locales that are part of Unicode CLDR 33.0.0.0. There might be a slight regression, because exotic languages like LOLCAT or Klingon would not have an option to have localized months/weekdays, though. 😄

PR #1396, opened 2018-09-29
add .editorconfig

NerOcrO:
## Description

EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs. The EditorConfig project consists of a file format for defining coding styles and a collection of text editor plugins that enable editors to read the file format and adhere to defined styles.

Please read this website: https://editorconfig.org/ and configure your IDE.


hangy:
Thanks for the suggestion! I do like the idea of `.editorconfig` in general and have used it in other projects (especially using C#, where an extended syntax that is more similar to a lot of linter configurations is supported). We definitely should set up `.editorconfig` to make it easier for IDEs to get indention right, but your suggestion does not take the current formatting of the Perl and JavaScript code into account. Could you have a look at that and update it accordingly?

Additionally, we already use [stylelint](https://stylelint.io/user-guide/rules/) for CSS and [ESLint](https://eslint.org/docs/rules/) for JS, so that adding [Perltidy](http://perltidy.sourceforge.net/stylekey.html) for Perl files might be a good next step. 🙂

hangy:
Closing PR as the source repository has been deleted.

PR #1397, opened 2018-09-29
Update backend dockerfile

juliengk:
## Description

Update Dockerfile according to Docker best practices https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -1,20 +1,19 @@
 FROM hangy/httpd-alpine-perl-modperl AS builder
 
-# Remove APK cache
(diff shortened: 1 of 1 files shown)


docker/backend/Dockerfile
```diff
@@ -1,65 +1,55 @@
 FROM hangy/httpd-alpine-perl-modperl AS builder
 
-# Remove APK cache
```
hangy: I had experienced some issue when this was not in the Dockerfile before, but now it seems to work with `apk -U --no-cache add`. 👍

PR #1398, opened 2018-09-30
Lazy load product images/thumbnails

### LINKED ISSUE #1198: Consider lazy loading images
Currently, all images are loaded directly. Especially for mobile users, lazy loading product images could improve perceived load time.

https://www.sitepoint.com/five-techniques-lazy-load-images-website-performance/

hangy:
## Description
Uses an Interaction Observer based lazy loading library (+ polyfill) to defer loading of product images and thumbnails.

I believe this is actually very important, because I noticed that Chrome seems to load full product front/ingredients/nutrient images, even though the modal popup is never opened. Considering the full product image can be > 1 MiB in size, that is kind of wasteful for metered connections. It also negatively impacts loading speed and Google's speed rating. Lazily loading those images fixes the problem.

Fixes #1198


stephanegigandet:
I just tested it, this works very well, but the downside is that we don't get any images if javascript is turned off.
I'm not sure how important it is to support javascript being off.
This is what it looks like currently if javascript is off (without lazy loading):

image

hangy:
Merci :)

PR #1446, opened 2018-10-27
Update html

NerOcrO:
Improve the front.


stephanegigandet:
Currently on .net   Can you take a look at the various pages @teolemon @hangy ? Thanks!

teolemon:
I've made a few checks on Chrome. No visible breakage.
We perhaps need to ensure there's a mechanism to autogenerate SCSS.

PR #1472, opened 2018-11-08
Update alcohol names for Food.pm

teolemon:


lib/ProductOpener/Food.pm
```diff
@@ -780,36 +780,44 @@ carnitine-
...
 	ro => "Alcool",
-	bg => "Алкохол",
+	rs => "Alkohol",
+	ru => "Алкоголь",
+	sk => "Alkohol",
+	sl => "Alkohol",
+	sv => "Alkohol",
+	tr => "Alkol",
 	zh => "酒精度",
+	zh-CN => "酒精度",
```
stephanegigandet: zh-CN is not valid perl and it will break, put either zh_cn or "zh-CN". Those won't be used at this time.

teolemon:
```suggestion
	zh_CN => "酒精度",
```

PR #1488, opened 2018-11-10
Added translations of lecithin

aleene:


taxonomies/ingredients.txt
```diff
@@ -27238,6 +27233,57 @@ fr:e950
...
+sr:Лецитин
+su:Lésitin
+sv:Lecitin
+th:เลซิทิน
+tr:Lesitin
+uk:Лецитин
+uz:Letsinlar
+vi:Lecithin
+zh:卵磷脂
+wikidata:Q241595
```
teolemon:
```suggestion
wikidata:en:Q241595
```

taxonomies/ingredients.txt
```diff
@@ -27238,6 +27233,57 @@ fr:e950
...
+wikipedia:https://en.wikipedia.org/wiki/Lecithin
```
teolemon:
```suggestion
wikipedia:en:https://en.wikipedia.org/wiki/Lecithin
```

taxonomies/ingredients.txt
aleene: for wikidata en: is not necessary. wikidata is language independent.

stephanegigandet: It is necessary for OFF to put wikidata:en: , the property system assumes that properties have a a language. For properties that are language independent, we use :en

PR #1500, opened 2018-11-12
Added translations for E301

aleene:
Is this the way to work? I compare the OFF-fork with mine. And I want to push my edits to off.

I am a bit lost here.


aleene:
I see there is a conflict as the off version is ahead of my version. Can I resolve that?

PR #1503, opened 2018-11-13
Added bread crumbs & E471(#1)

aleene:
* Added bread crumbs

* Added E471


aleene:
I do not see what the conflict is.

PR #1506, opened 2018-11-13
replace ~~ by grep

NerOcrO:
1/ Avoid a warning message when I launch startup_apache2.pl: `Smartmatch is experimental at /home/off/lib/ProductOpener/Lang.pm line 462.`
2/ Clean Apache's logs
3/ https://metacpan.org/pod/release/RJBS/perl-5.18.0/pod/perldelta.pod#The_smartmatch_family_of_features_are_now_experimental
4/ "The smartmatch operator is experimental and its behavior is subject to change.": https://perldoc.perl.org/perlop.html#Smartmatch-Operator
5/ https://perlmaven.com/smart-matching-in-perl-5.10 (Conclusion)

Don't use this operator anymore please.


NerOcrO:
Sure!
Is there a function like [array_search](http://fr2.php.net/manual/en/function.array-search.php) in PHP?

lib/ProductOpener/Ingredients.pm
```diff
@@ -1399,7 +1399,7 @@ INFO
 		$product_ref->{$tagtype . '_tags'} = [];		
 				
 		# skip palm oil classes if there is a palm oil free label
-		if (($class =~ /palm/) and ($product_ref->{labels_tags} ~~ 'en:palm-oil-free')) {
+		if (($class =~ /palm/) and ($product_ref->{labels_tags} == 'en:palm-oil-free')) {
```
hangy: I think this probably also needs to be an array search, because `labels_tags` is an array? @stephanegigandet

stephanegigandet: yes, labels_tags is an array

We have a has_tag function:  https://github.com/openfoodfacts/openfoodfacts-server/blob/master/lib/ProductOpener/Tags.pm#L179 

has_tag($product_ref, "labels", 'en:palm-oil-free')

lib/ProductOpener/URL.pm
```diff
@@ -64,8 +64,8 @@ sub subdomain_supports_https {
 	my ($sd) = @_;
 	
 	return $sd unless $sd;
-	return 1 if '*' ~~ @ssl_subdomains;
-	return $sd ~~ @ssl_subdomains;
+	return 1 if grep $_ eq '*', @ssl_subdomains;
+	return grep $_ eq $sd, @ssl_subdomains;
```
stephanegigandet: This function is called a lot of times when generating a page, instead of doing a grep, it would be much more efficient to create a hash with the @ssl_subdomains values as keys during initialization of the module, and then use the hash to quickly check it.

That being said, we now have wildcards, so @ssl_subdomains contains only "*"  (at least in production)

PR #1527, opened 2018-11-19
Various E-numbers around acid

aleene:
E300, E120, E200, E260, E262, E270, E296, E330, E334

B9


aleene:
This is still needed.

aleene:
This one seems to be merged.

Issue #1068, opened 2018-02-12
Broken links to /data /terms-of-use /legal in footer (translated without corresponding text)


stephanegigandet:
The footer contains those links:
 <li><a href="$Lang{footer_legal_link}{$lc}">$Lang{footer_legal}{$lc}</a></li>
 <li><a href="$Lang{footer_terms_link}{$lc}">$Lang{footer_terms}{$lc}</a></li>
 <li><a href="$Lang{footer_data_link}{$lc}">$Lang{footer_data}{$lc}</a></li>

if the footer_[..]_link are translated in the po files, then a corresponding file needs to exist in /lang/[lc]/[text id].html

e.g. on production, https://fr.openfoodfacts.org links to https://fr.openfoodfacts.org/donn%C3%A9es which does not exist.

I'm going to add test so that the build fails when those links are broken.

teolemon:
@stephanegigandet I think we can close this one

PR #1532, opened 2018-11-21
Fix menu and footer broken links in all languages (checked all .po files in /po/common/)

lan33:
**Description:**

This PR tries to fix broken links in menu and footer in all languages
Links should not be translated.
Alls `.po` files in `/po/common/` are checked, translated links are removed.

This fix may be related to the bug #1068


po/common/mr.po
```diff
@@ -23,7 +23,7 @@ msgstr "mr"
 #. put here the language name ("French", "Japanese", etc)
 msgctxt ":langname"
 msgid ":langname"
-msgstr "भाषानाव"
+msgstr "Marathi"
```
hangy: I am not quite sure which language `:langname` should be in. In production, we have some cases where the name is in the native language of the file, and others where it is in English. More importantly, I do not see `:langname` being used outside of loading the translations. Am I missing something, @stephanegigandet @maddingue?

teolemon: No use for Crowdin. @maddingue  ?

stephanegigandet: We could remove it then. It was generated when we transformed I18N.pm to .po files but it's not used.

stephanegigandet: Filed https://github.com/openfoodfacts/openfoodfacts-server/issues/1553

PR #1546, opened 2018-12-03
Move JavaScript from devDependencies

hangy:
Move JavaScript from devDependencies to dependencies. Follow up to the discussion in #1182.


NerOcrO:
I don't know pnpm, so let's try this!

PR #1549, opened 2018-12-03
Update Config_off.pm

### LINKED ISSUE #1548: Fix manifest.pl
We need to add a new parameter to manifest.pl

https://developers.google.com/web/fundamentals/app-install-banners/native

`prefer_related_applications >needs to be> true`

hangy:
Fixes #1548


hangy:
I wonder if this works with a `'true'` string instead of  `true`.

teolemon:
Looks good to me

hangy:
I think #1551 is better.

PR #1575, opened 2018-12-14
Fix CI build

hangy:
Gulp (or rather one of it's dependencies in the yarn.lock file) stopped running on Node.js 11.


hangy:
Merging, because `yarn run build` was successful.

PR #1581, opened 2018-12-19
Compile cgi/*.pl in Travis CI

hangy:
Previously, we only built `cgi/product_multilingual.pl` during CI. Obviously, any change in the CGI directory should be tested, because all of it is user-facing. This is implemented in this CR. To then get this build going, I actually needed update some `use` statements from `JSON` to `JSON:PP`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,8 @@ module.exports = {
     "env": {
         "browser": true,
         "es6": true,
-        "jquery": true
+        "jquery": true,
+        "amd": true
     },
     "extends": "eslint:recommended",
     "parserOptions": {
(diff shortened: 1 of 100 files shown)


hangy:
GitHub currently doesn't let me edit the PR text (except for the checkboxes), so I'll just add it as a comment for now:

Previously, we only built `cgi/product_multilingual.pl` during CI. Obviously, any change in the CGI directory should be tested, because all of it is user-facing. This is implemented in this CR. To then get this build going, I actually needed update some `use` statements from `JSON` to `JSON:PP`.

stephanegigandet:
Shouldn't we try to use JSON::XS instead? Or is it something that is not supported by Travis for some reason? Hopefully the output is the same for both.

hangy:
> Shouldn't we try to use JSON::XS instead? Or is it something that is not supported by Travis for some reason? Hopefully the output is the same for both.

I have not looked up the difference between those modules at all. The only reason I updated the scripts to `JSON::PP` is that those were apparently left out in 920d13efc4ffd4c77c3bfdd0d1ece409eb230f41.

PR #1598, opened 2018-12-30
Use JSON module for #world-map data

### LINKED ISSUE #196: Processing places map bug
http://fr.openfoodfacts.org/lieux-de-fabrication

Hover on some countries:

Example of buggy countries : 
Fra (for France)
E (for Spain)
Uk

hangy:
Apparently, one problem in the world map is that some country codes are used multiple times. By using a hash map and the JSON module to encode the hash to JavaScript, duplicate keys should be avoided.

Fixes #196

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -1265,12 +1265,12 @@ sub display_list_of_tags($$) {
 		}
 
 		$log->debug("MongoDB query done", { error => $@ }) if $log->is_debug();
-		
+
 		$log->trace("aggregate query done") if $log->is_trace();
 
 		if ($admin) {
-			$log->debug("aggregate query results", { results => $results }) if $log->is_debug();	
-		}		
+			$log->debug("aggregate query results", { results => $results }) if $log->is_debug();
+		}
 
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -1570,12 +1570,19 @@ sub display_list_of_tags($$) {
 					if ($region eq 'UK') {
 						$region = 'GB';
 					}
-					$countries_map_links .=  '"' . $region . '": "' . $product_link . "\",\n";
-					$countries_map_data .= '"' . $region . '": ' . $products . ",\n";
+
+					$countries_map_links->{$region} = $product_link;
```
stephanegigandet: Shouldn't we keep only the first $product_link (which should be the most common one as the list is in descending order). In that case maybe we should only keep the first count and not add the count for other variations. What do you think?

hangy: Yes, we could probably just keep the first link. I summed up the different links for a region, because it didn't look like they represent the same products. If all of them are contained in the first figure, summing them up again would obviously be redundant.

stephanegigandet: Let's try it as it is now.

Issue #1610, opened 2019-01-09
Improve aggregations performance


syl10100:
**Summary:**
Aggregations can take a while, especially when none country is selected (e.g. https://world.openfoodfacts.org/labels )

**Steps to reproduce:**
Go to this URL https://fr-en.openfoodfacts.org/labels
It will execute in Mongodb :
```
db.products.aggregate( [
{"$match" : {countries_tags : "en:france"}},
{"$unwind" : "$labels_tags"},
{"$group" : {_id : "$labels_tags",count : {"$sum" : 1}}},
{"$sort" : {count : -1}}
])
```

**Proposition:**

For aggregation queries, we can assume that real-time is not mandatory. I propose to 
create a copy of products collection with only tags fields used in aggregation. This will allow to have a collection with smaller documents. And fetching small documents is faster.  
This new collection should be refreshed every night.

The following mongo shell command will copy in a collection called «**products_tags**» all tags field
used in drilldown menu.

```
db.products.aggregate( [
{"$project" : 
{
countries_tags:1,
brands_tags:1, 
categories_tags:1, 
labels_tags:1, 
packaging_tags:1, 
origins_tags:1, 
manufacturing_places_tags:1,
emb_codes_tags:1, 
ingredients_tags:1,
additives_tags:1,
vitamins_tags:1,
minerals_tags:1,
amino_acids_tags:1,
nucleotides_tags:1,
allergens_tags:1,
traces_tags:1,
nova_groups_tags:1,
nutrition_grades_tags:1,
languages_tags:1,
creator_tags:1,
editors_tags:1, 
states_tags:1,
entry_dates_tags:1,
last_edit_dates_tags:1,
codes_tags:1
}},
{"$out": "products_tags"}
])
```

All aggregate commands will run on products_tags collection instead of products collection.

stephanegigandet:
I'm testing it on my test server (world.openfoodfacts.net) which has an old mongodb:

tail -f refresh_products_tags.log
MongoDB shell version: 2.6.6
connecting to: off
Thu Jan 31 2019 16:56:30 GMT+0100 (CET) : Refresh products_tags collection...
2019-01-31T16:57:14.528+0100 TypeError: Property 'estimatedDocumentCount' of object off.products_tags is not a function at refresh_products_tags.js:43
failed to load: refresh_products_tags.js

In production we have mongo 4.0.2, and it seems db.collection.estimatedDocumentCount() is new in 4.0.3.

[...]

PR #1612, opened 2019-01-09
improve aggregations perf

syl10100:
**Description:**

1. create products_tags collection with script refresh_products_tags.js
2. modify files Data.pm and Display.pm to use products_tags collection in aggregate.
3. schedule refresh of products_tags collection

**Related issues and discussion:** #1610

### DIFF
--- a/scripts/refresh_products_tags.js
+++ b/scripts/refresh_products_tags.js
@@ -1,13 +1,11 @@
 /* 
   Name        : refresh_products_tags.js 
   Description : Refresh products_tags collection by copying *_tags fields from products collection 
-  Version     : 1.0
-  Usage       : mongo refresh_products_tags.js
+  Version     : 1.1
+  Usage       : mongo dbname refresh_products_tags.js
   Add in crontab for daily refresh : 
-	00 02 * * * /usr/bin/mongo /etc/mongo/refresh_products_tags.js  >> /var/log/mongodb/refresh_products_tags.log
+	00 02 * * * /usr/bin/mongo off /etc/mongo/refresh_products_tags.js  >> /var/log/mongodb/refresh_products_tags.log
 */
-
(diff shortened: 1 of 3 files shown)


scripts/refresh_products_tags.js
```diff
@@ -0,0 +1,55 @@
+/* 
+  Name        : refresh_products_tags.js 
+  Description : Refresh products_tags collection by copying *_tags fields from products collection 
+  Version     : 1.0
+  Usage       : mongo refresh_products_tags.js
+  Add in crontab for daily refresh : 
+	00 02 * * * /usr/bin/mongo /etc/mongo/refresh_products_tags.js  >> /var/log/mongodb/refresh_products_tags.log
+*/
+
+db = db.getSiblingDB('off');
```
stephanegigandet: is there a way to make the db name a parameter? we have separate names for OBF, OPF, OPFF

scripts/refresh_products_tags.js
```diff
@@ -0,0 +1,55 @@
...
+	codes_tags:1
+	}},
+{"$out": "products_tags"}
+]);
+
+print(Date() + ' : ' + db.products_tags.estimatedDocumentCount() + ' products refreshed in products_tags collection.');
+
+
+print(Date() + ' : Create indexes for products_tags collection (if not already existing)...');
+db.products_tags.createIndex({countries_tags: 1});
```
stephanegigandet: isn't that going to block the database without the background : true ?

stephanegigandet:
It looks good to me, thank you. I'll probably change it a little to make it optional, so that we can test it more easily.

syl10100:
> 
> 
> It looks good to me, thank you. I'll probably change it a little to make it optional, so that we can test it more easily.

Yes good idea

PR #1615, opened 2019-01-10
sodium carbonates

aleene:


stephanegigandet:
It should be <en:sodium carbonates instead of <en:sodium bicarbonates I think.
I fixed it in another branch.
This parent category is tricky, we have it in the additives taxonomy for E500, but the parent name is the same as one of the children.

aleene:
I do not think so: https://en.wikipedia.org/wiki/E_number#E500.E2.80.93E599_.28acidity_regulators.2C_anti-caking_agents.29

PR #1625, opened 2019-01-23
Add local databases

CharlesNepote:
Some files are local databases created by the app on each environment: *.sto, top_translators.csv. They have to be ignored by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,9 @@ logs/modperl_error_log
 
 # Local databases
 data/mongodb
+Lang.openfoodfacts.*
+users_emails.sto
+html/data/top_translators.csv
 
 # Libraries
 node_modules
--- a/.gitignore
+++ b/.gitignore
@@ -20,9 +20,9 @@ logs/modperl_error_log
 
 # Local databases
 data/mongodb
-Lang.openfoodfacts.*
+Lang.open*
 users_emails.sto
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -20,6 +20,9 @@ logs/modperl_error_log
 
 # Local databases
 data/mongodb
+Lang.openfoodfacts.*
+users_emails.sto
+html/data/top_translators.csv
```
hangy: Could we ignore html/data/ as a whole? There's other scripts that export MongoDB, RDF, etc. there.

.gitignore
```diff
@@ -20,6 +20,9 @@ logs/modperl_error_log
 
 # Local databases
 data/mongodb
+Lang.openfoodfacts.*
```
hangy: Is `openfoodfacts` static or could it also be `openpetfoodfacts`?

.gitignore
CharlesNepote: Ok, done.

.gitignore
CharlesNepote: I have put `Lang.open*`.

I think I will open a new bug to suggest moving this files from root directory.

CharlesNepote:
As it's very anecdotic changes, I'll merge tomorrow if there are no more comments.

PR #1633, opened 2019-02-01
Issue 1630/long lists of tags

stephanegigandet:
A relatively small performance improvement to call format_subdomain only once. In lists of tags, it was called for each tag, which in some cases like pages like /codes /ingredients etc. can be more than 100k times.
Each tag still calls other functions to canonicalize tags to get links, to get the right translation from the taxonomy etc. We could try to factorize those calls so that we have one unique call to:
- see if the tag exists in the taxonomy
- get the link for the tag
- get the display form of the tag

Also reduced the number of tags in codes_tags.


lib/ProductOpener/Display.pm
```diff
@@ -3478,7 +3495,7 @@ HTML
 		if (defined $request_ref->{jqm}) {
 			if (defined $next_page_url) {
 				my $loadmore = lang("loadmore");
-				my $loadmore_domain = format_subdomain($subdomain);
+				my $loadmore_domain = $format_subdomain;
```
hangy: `$loadmore_domain` is now redundant

lib/ProductOpener/Display.pm
```diff
@@ -78,6 +78,7 @@ BEGIN
 
 					$original_subdomain
 					$subdomain
+					$format_subdomain
```
hangy: I think, `$formatted_subdomain` might be a better name, because the formatting has already been done.

lib/ProductOpener/Display.pm
```diff
@@ -3504,7 +3521,7 @@ HTML
 
 	if ($subdomain ne $original_subdomain) {
 		$log->debug("subdomain not equal to original_subdomain, converting relative paths to absolute paths", { subdomain => $subdomain, original_subdomain => $original_subdomain }) if $log->is_debug();
-		my $formated_subdomain = format_subdomain($subdomain);
+		my $formated_subdomain = $format_subdomain;
```
hangy: `$formated_subdomain` is now redundant

lib/ProductOpener/Display.pm
```diff
@@ -8533,7 +8551,7 @@ sub display_structured_response_opensearch_rss {
 
 	$long_name = $xs->escape_value(encode_utf8($long_name));
 	$short_name = $xs->escape_value(encode_utf8($short_name));
-	my $dom = format_subdomain($subdomain);
+	my $dom = $format_subdomain;
```
hangy: `$dom` is now redundant

lib/ProductOpener/Display.pm
```diff
@@ -1624,7 +1638,7 @@ sub display_list_of_tags($$) {
                 text: '$request_ref->{title}'
             },
             subtitle: {
-                text: '$Lang{data_source}{$lc}$sep: @{[ format_subdomain($subdomain) ]}'
+                text: '$Lang{data_source}{$lc}$sep: @{[ $format_subdomain ]}'
```
hangy: I /think/ the `@{[...}]` syntax was only required, because it called the sub. Shouldn't this work (and be more readable)?
```suggestion
                text: '$Lang{data_source}{$lc}$sep: $format_subdomain'
```

lib/ProductOpener/Display.pm
```diff
@@ -1717,7 +1731,7 @@ HTML
   },
   onRegionClick: function(e, code, region){
 	if (countries_map_links[code]) {
-		window.location.href = "@{[ format_subdomain($subdomain) ]}" + countries_map_links[code];
+		window.location.href = "@{[ $format_subdomain ]}" + countries_map_links[code];
```
hangy:
```suggestion
		window.location.href = "$format_subdomain" + countries_map_links[code];
```

lib/ProductOpener/Display.pm
```diff
@@ -4105,7 +4122,7 @@ JS
                 text: '$graph_ref->{graph_title}'
             },
             subtitle: {
-                text: '$Lang{data_source}{$lc}$sep: @{[ format_subdomain($subdomain) ]}'
+                text: '$Lang{data_source}{$lc}$sep: @{[ $format_subdomain ]}'
```
hangy:
```suggestion
                text: '$Lang{data_source}{$lc}$sep: $format_subdomain'
```

lib/ProductOpener/Display.pm
```diff
@@ -4456,7 +4473,7 @@ JS
...
```
hangy:
```suggestion
                text: '$Lang{data_source}{$lc}$sep: $format_subdomain'
```

lib/ProductOpener/Display.pm
```diff
@@ -5288,7 +5305,7 @@ $options{favicons}
 <link rel="stylesheet" href="/css/dist/app.css?v=$file_timestamps{"css/dist/app.css"}">
 <link rel="stylesheet" href="@{[ format_subdomain('static') ]}/bower_components/jquery-ui/themes/base/jquery-ui.min.css">
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.3/css/select2.min.css" integrity="sha384-HIipfSYbpCkh5/1V87AWAeR5SUrNiewznrUrtNz1ux4uneLhsAKzv/0FnMbj3m6g" crossorigin="anonymous">
-<link rel="search" href="@{[ format_subdomain($subdomain) ]}/cgi/opensearch.pl" type="application/opensearchdescription+xml" title="$Lang{site_name}{$lang}">
+<link rel="search" href="@{[ $format_subdomain ]}/cgi/opensearch.pl" type="application/opensearchdescription+xml" title="$Lang{site_name}{$lang}">
```
hangy:
```suggestion
<link rel="search" href="$format_subdomain/cgi/opensearch.pl" type="application/opensearchdescription+xml" title="$Lang{site_name}{$lang}">
```

lib/ProductOpener/Display.pm
```diff
@@ -5737,17 +5754,17 @@ window.addEventListener('load', onLoad);
 	"\@context" : "https://schema.org",
 	"\@type" : "WebSite",
 	"name" : "$Lang{site_name}{$lc}",
-	"url" : "@{[ format_subdomain($subdomain) ]}",
+	"url" : "@{[ $format_subdomain ]}",
```
hangy:
```suggestion
	"url" : "$format_subdomain",
```

lib/ProductOpener/Display.pm
```diff
@@ -5737,17 +5754,17 @@ window.addEventListener('load', onLoad);
...
 	"potentialAction": {
 		"\@type": "SearchAction",
-		"target": "@{[ format_subdomain($subdomain) ]}/cgi/search.pl?search_terms=?{search_term_string}",
+		"target": "@{[ $format_subdomain ]}/cgi/search.pl?search_terms=?{search_term_string}",
```
hangy:
```suggestion
		"target": "$format_subdomain/cgi/search.pl?search_terms=?{search_term_string}",
```

lib/ProductOpener/Display.pm
```diff
@@ -5737,17 +5754,17 @@ window.addEventListener('load', onLoad);
...
 		"query-input": "required name=search_term_string"
 	}
 }
 {
 	"\@context": "https://schema.org/",
 	"\@type": "Organization",
-	"url": "@{[ format_subdomain($subdomain) ]}",
+	"url": "@{[ $format_subdomain ]}",
```
hangy:
```suggestion
	"url": "$format_subdomain",
```

lib/ProductOpener/Display.pm
```diff
@@ -1401,9 +1405,17 @@ sub display_list_of_tags($$) {
...
+		$log->debug("going through all tags", {}) if $log->is_debug();
+		
+		my $i = 0;
 
 		foreach my $tagcount_ref (@tags) {
-
+	
+			$i++;
+			
+			($i % 10000 == 0) and $log->debug("going through all tags", {i => $i}) if $log->is_debug();
```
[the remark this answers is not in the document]
stephanegigandet: I'm not really sure either, but the debug should not be called if either is false. ;) I can add a if ($i % 10000 == 0) { .. }

lib/ProductOpener/Display.pm
stephanegigandet: Agreed.

PR #1639, opened 2019-02-03
Trim trailing whitespaces (see OFF .editorconfig)

CharlesNepote:
Trim trailing whitespaces to conform to .editorconfig and increase future diff readability.

Code doesn't change at all.


CharlesNepote:
There no change to the code. Can I merge pull request myself tin this case?

PR #1643, opened 2019-02-03
Fix <<site_name>> in German translations

hangy:
image

Not being replaced correctly right now.


stephanegigandet:
Could you add a test to catch those?

e.g. can we make a test fail if we get an output from this:
grep site_name po/*/*.po  |grep -v "<<site_name>>" |grep -v msgctxt

It catches many more issues in other languages in fact.

hangy:
> Could you add a test to catch those?
> 
> e.g. can we make a test fail if we get an output from this:
> grep site_name po/_/_.po |grep -v "<<site_name>>" |grep -v msgctxt
> 
> It catches many more issues in other languages in fact.

I created a new issue for this: #1648

Issue #811, opened 2017-04-27
RDF Data Export seems broken


teolemon:
```I was wondering if there was any contributors currently working on improving and correcting the RDF database of OpenFoodFacts.
If yes, is there a way to be in contact with them or a discussion about what and how they are working on it?```

**Comment by rogargon** [comment]
RDF data is broken. The EN and FR versions include URL with whitespaces. The ES version is missing.

For instance, trying to load the data into Virtuoso:

```
dumps/en.openfoodfacts.org.products.rdf | https://world.openfoodfacts.org/data | 2 | 2019-01-29 12:17:14.528896 | 2019-01-29 12:17:14.607020 | 0 | <DB NULL> | 22007 XM033: XML parser detected an error: ERROR : Syntax error in the attribute list (no whitespace) at line 14074 column 30 of source text <food:food rdf:resource="http://fr.openfoodfacts.org/ingredient/fr:Viter l' ----------------------------^
dumps/fr.openfoodfacts.org.products.rdf | https://world.openfoodfacts.org/data | 2 | 2019-01-29 12:50:39.451954 | 2019-01-29 12:50:39.525361 | 0 | <DB NULL> | 22007 XM033: XML parser detected an error: ERROR : Syntax error in the attribute list (no whitespace) at line 14074 column 30 of source text <food:food rdf:resource="http://fr.openfoodfacts.org/ingredient/fr:sin aler ----------------------------^
```

CharlesNepote:
Thanks @rogargon. The RDF is an old export not very well maintained. Olivier Rossel on twitter said he wrote an good exporter based on the CSV, see: https://twitter.com/datao/status/1086213152327516160 (in french).

This bug is not a high priority for us. I'll try to have a look but I can't garanty when it's done. In case you know a bit of Perl, the export script is here:
https://github.com/openfoodfacts/openfoodfacts-server/blob/1d4063feaedc759543419227403c66cfa6658029/scripts/export_database.pl

Ready to use solutions welcome ;-) (Pull request or whatever.)

CharlesNepote:
@rogargon: @lolive has just published his scripts here: https://github.com/lolive/openfoodfacts-rdf

For the CSV alternative there is a good article here related to injecting the CSV in PostgreSQL: http://blog-postgresql.verite.pro/2018/12/21/import-openfoodfacts.html (french but should be ok for Google Trad).

lolive:
[...]

PR #1659, opened 2019-02-08
Correct rdf data export (#811) and prevent broken CSV (#1585)

### LINKED ISSUE #1585: Dataset shouldn't contain specials chars between \000-\007 and \013-\037
**Summary:**

Chars between \000-\007 and \013-\037 decimal codes are not used anymore -- they come from the old times where they were useful to control printers and so on. They break the CSV file which can't be processed by common tools such as csvkit.

**Steps to reproduce:**

A file listing every line containing bad chars can be produced with a simple command line (see [bad-chars.txt](https://github.com/openfoodfacts/openfoodfacts-server/files/2698784/bad-chars.txt)):
```
$ perl -ne 'print if /[\000-\007\013-\037]/' fr.openfoodfacts.org.products.csv > bad-chars.csv
```

**Expected behavior:**

Every data entered in the database should be controlled.

CharlesNepote:
**Description:**

1. Encode URIs for RDF export (see #811).
2. Filter bad chars which break CSV and log it in a file (prevent but doesn't resolve #1585).

It seems to work with a test dataset of 177 products, but you might try it before deploying in production. I don't know how to test on the whole database.

You should review commit by commit if you want to see clearly what I've done. Commit 9fb9f6f is a mistake and I'm not very confident on commit deletion.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -369,7 +369,7 @@ CSS
 CSS
 ;
 	}
-	
+
 	# call format_subdomain($subdomain) only once
 	$formatted_subdomain = format_subdomain($subdomain);
 }
@@ -1405,19 +1405,19 @@ sub display_list_of_tags($$) {
 		}
 
 		my %products = ();	# number of products by tag, used for histogram of nutrition grades colors
-		
+
 		$log->debug("going through all tags", {}) if $log->is_debug();
-		
+
 		my $i = 0;
 
 		foreach my $tagcount_ref (@tags) {
-	
+
 			$i++;
-			
+
 			if (($i % 10000 == 0) and ($log->is_debug())) {
 				$log->debug("going through all tags", {i => $i});
 			}
-		
+
 			my $tagid = $tagcount_ref->{_id};
 			my $count = $tagcount_ref->{count};
 
@@ -1609,7 +1609,7 @@ sub display_list_of_tags($$) {
 		$html .= "</tbody></table></div>";
 
 		$log->debug("going through all tags - done", {}) if $log->is_debug();
-		
+
 
 		# nutrition grades colors histogram
 
@@ -1803,7 +1803,7 @@ HEADER
 
 	# datatables clears both
 	$request_ref->{full_width} = 1;
-	
+
 	$log->debug("end", {}) if $log->is_debug();
 
 
@@ -3627,19 +3627,19 @@ sub search_and_export_products($$$$$) {
 	}
 
 
-	
+
 
 	if ($count > 0) {
-	
+
 		# Send the CSV file line by line
-		
+
 		use Apache2::RequestRec ();
 		my $r = Apache2::RequestUtil->request();
 		$r->headers_out->set("Content-type" => "text/csv; charset=UTF-8");
 		$r->headers_out->set("Content-disposition" => "attachment;filename=openfoodfacts_search.csv");
 		binmode(STDOUT, ":encoding(UTF-8)");
-		print "Content-Type: text/csv; charset=UTF-8\r\n\r\n";	
-	
+		print "Content-Type: text/csv; charset=UTF-8\r\n\r\n";
+
 		my $categories_nutriments_ref = retrieve("$data_root/index/categories_nutriments_per_country.$cc.sto");
 
 		# First pass needed if we flatten results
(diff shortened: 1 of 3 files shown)


CharlesNepote:
i made some tests with the whole database. There is still database content issues which break the CSV export. Strange database issue: there is a carriage return (\n) at the end of the GPS coordinates related to villecomtal-sur-arros-gers-france: 43.400279,0.199525.

So we have to sanitize the content of more fields than I expected.

I suggest to create a `sanitize_field_value()` function, probably in `ProductOpener::Data`, with two arguments:

- `$field_value`: the content to be sanitized;
- `$log`: optionally the output where to log each problem (example:  a filehandler, STDOUT, etc.)

The function would return the sanitized content. Non visible ASCII chars would be replaced by a space, except trailing whitespace or "\n", which would be trimed. Example:

`sanitize_field_value("something\nanother\n", $LOG)` => "something another"

The function would be used both in input (web app, imports) and in ouput (CSV exports).

@hangy @stephanegigandet ?

hangy:
> i made some tests with the whole database. There is still database content issues which break the CSV export. Strange database issue: there is a carriage return (\n) at the end of the GPS coordinates related to villecomtal-sur-arros-gers-france: 43.400279,0.199525.

If there's just one case of bad source data, cleaning the source file might be easier?

> The function would return the sanitized content. Non visible ASCII chars would be replaced by a space, except trailing whitespace or "\n", which would be trimed.

Sounds good! If you want to replace all control characters, you could also use `\p{Cc}`.

> There might be an issue related to \n in some fields: do we have to delete it all the time?

I don't think so. For CSV, [RFC 4180](https://tools.ietf.org/html/rfc4180#section-2) clearly states that fields containing CRLF should be possible (but be quoted with double quotes). In XML, line breaks should be fine in attributes and values, even though they might need to be used with `<![CDATA[` to avoid incorrect with some XML parsers. Therefore, multi-line text should be valid field/attribute/column content.

CharlesNepote:
Ok I made more testings and enhance the exporter:
- prevent perl warnings
- avoid control chars in rdf file

All seems to be fine. CSV file is ok. RDF file is valid (tested with `xmllint --stream en.openfoodfacts.localhost.products.rdf` doesn't provide any more issues).

For the moment I keep `sanitize_field_content()` in the `./scripts/export_database.pl`. I'll see later if it's interesting to move it to a perl OFF lib.

> If there's just one case of bad source data, cleaning the source file might be easier?

Yes we will have to clean to the source, but my experience is that the source might be corrupted again by the time, so I prefer to also clean the output.

> Sounds good! If you want to replace all control characters, you could also use `\p{Cc}`.

I never tested `\p{Cc}` so I keep char codes for the moment. Maybe I'll try later if it's quicker or if it consume less memory.

>> There might be an issue related to \n in some fields: do we have to delete it all the time?

> I don't think so. For CSV, RFC 4180 clearly states that fields containing CRLF should be possible (but be quoted with double quotes). In XML, line breaks should be fine in attributes and values, even though they might need to be used with `<![CDATA[` to avoid incorrect with some XML parsers. Therefore, multi-line text should be valid field/attribute/column content.

Yes RFC states CRLF is possible if fields are quoted. I think it's ok for small files, with few columns. But I think it's dangerous for huge files with dozens of columns as Open Food Facts dataset. In theory CSV tools should deal with CRLF in quoted fields but in fact some of them doesn't play well with that. Also you can't use regular unix text tools when your CSV contain CRLF inside fields: `sort`, `wc -l`, `head`, `grep`, etc., will not work well. It makes CSV manipulation more complex.

So I think we should avoid CRLF inside fields in the CSV export. RDF file might be another discussion.

My question was more related to the usages. Is there some OFF usages where CRLF are important inside a field? Maybe we could replace CRLF inside some field by a string: "---" or "\\\\" for example.

By the way we can see later and deploy this pull request which is a first good step.

hangy:
> > Sounds good! If you want to replace all control characters, you could also use `\p{Cc}`.
> 
> I never tested `\p{Cc}` so I keep char codes for the moment. Maybe I'll try later if it's quicker or if it consume less memory.

It's just the Unicode category [Other, Control](https://www.fileformat.info/info/unicode/category/Cc/list.htm). It includes the range you already mentioned, plus different control characters outside of that range.

> My question was more related to the usages. Is there some OFF usages where CRLF are important inside a field? Maybe we could replace CRLF inside some field by a string: "---" or "\\" for example.

Fair point. I don't know about that. 🙂

CharlesNepote:
> > > Sounds good! If you want to replace all control characters, you could also use `\p{Cc}`.
> > 
> > I never tested `\p{Cc}` so I keep char codes for the moment. Maybe I'll try later if it's quicker or if it consume less memory.
> 
> It's just the Unicode category [Other, Control](https://www.fileformat.info/info/unicode/category/Cc/list.htm). It includes the range you already mentioned, plus different control characters outside of that range.

Thanks for the link, I'll make some tests to see if there are issues with control chars outside of the range I focused on.

At first, I think we can merge and deploy my pull request as it solves a real issue. @stephanegigandet ?

PR #1662, opened 2019-02-12
Patch 17 - ingredients taxonomy

aleene:
coeur de boeuf to start with


teolemon:
@aleene There's a merge conflict. Can we solve it, and perhaps merge @stephanegigandet ?

aleene:
There are to many changes, which is to much for the conflicts resolver.

stephanegigandet:
Hi @aleene, the conflict resolver does not work because the file has Windows CRLF line endings, so all the lines have changed compared to master. Can change the line ending to Unix LF and commit it? We should be able to resolve the conflit after (which is just 1 extra line in Spanish in fact). Thanks!

Issue #1671, opened 2019-02-18
Issue with a made near me product


teolemon:
``` 
{ product_name:'Dos de la mer', brands:'Vici', url: 'https://world.openfoodfacts.org/product/4770190084161', img:'<img src="https://static.openfoodfacts.org/images/products/477/019/008/4161/front_fr.16.100.jpg" width="70" height="100" srcset="https://static.openfoodfacts.org/images/products/477/019/008/4161/front_fr.16.200.jpg 2x" alt="Dos de la mer - Product" /> ', origins:'Manufacturing or processing places: Lituanie<br>', geo:[Birutės g. 50 . Plungės raj. . Lithuania,55.9172451]},
```

teolemon:
@stephanegigandet @hangy not sure what's going on
https://world.openfoodfacts.org/product/4770190084161/dos-de-la-mer-vici

hangy:
https://github.com/openfoodfacts/openfoodfacts-server/blob/a96d311a29f77671db885e69aeec7abc1c412850/scripts/update_packager_codes.pl#L102 splits the files by `;` and then by `\t` https://github.com/openfoodfacts/openfoodfacts-server/blob/7f53be6d953a04ef6330dddaae3f36b18d673d90/packager-codes/LT-merge-UTF-8.csv#L1765 contains both - `\t` as field separator and `;` in one column value. This causes the geo code column values to be off by one or more columns. MadeNearMe then fails, because the coordinates aren't valid numeric values.

hangy:
Could you rebuild packager codes with the new script and test if PR #1675 helps? @teolemon @stephanegigandet

teolemon:
is this fixed ? how do we check it's fixed ? I'm not sure what the issue was. @hangy

hangy:
I think this should be done. The issue was bad geo codes in the taxonomy.

PR #1675, opened 2019-02-19
Support TSV for packager codes

hangy:
**Description:**

Instead of using both separators (`;` and `\t`) when loading packager codes, only use the one that matches the file extension.

**Related issues and discussion:** #1671


stephanegigandet:
Perfect, thank you!

teolemon:
@stephanegigandet We need to run this script

stephanegigandet:
I just ran it and restarted the server.

stephanegigandet:
Everything looks good. :) Thanks @hangy !

PR #1678, opened 2019-02-22
Fix "prefer_related_applications" being ignored

hangy:
Resolves a warning from Chrome. which warns us that `prefer_related_applications` is ignored, because it's not an actual boolean value:
> manifest.pl:1 Manifest: property 'prefer_related_applications' ignored, type boolean expected.


teolemon:
LGTM

PR #1681, opened 2019-02-27
Update ingredients.txt

JBelcoco:
Adding Azúcar Moreno <> Brown sugar translation

Nova calculation is not being done properly when azúcar moreno is in the list of ingredients.


teolemon:
@stephanegigandet @JBelcoco is using the API for El Coco in Spain.
He will be doing fixes on the taxonomy to ensure Nova computation is accurate for Spanish

PR #1682, opened 2019-02-27
Disable Nutri-Score for products waiting to be moved to other projects

teolemon:
Disable Nutri-Score for products waiting to be moved to other projects

### DIFF
--- a/lib/ProductOpener/Config_off.pm
+++ b/lib/ProductOpener/Config_off.pm
@@ -406,8 +406,8 @@ fr:levure
 fr:levures
 en:honeys
 en:vinegars
-en:pet food
-en:non food products
+en:pet-food
+en:non-food-products
 
 )];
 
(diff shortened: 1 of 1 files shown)


stephanegigandet:
The qw (quote word) operator will turn this:
en:pet food
into : "en:pet", "food", which is probably not what you want. en:pet-food will work.

On a side note, the Nutri-Score won't be computed if we don't have nutrition facts with sugars, fat etc.

lib/ProductOpener/Config_off.pm
```diff
@@ -406,6 +406,9 @@ fr:levure
 fr:levures
 en:honeys
 en:vinegars
+en:pet food
+en:non food products
```
hangy: As Stephane said, please use hyphens instead of spaces.

teolemon: done

PR #1685, opened 2019-02-28
Ensure that <<site_name>> is not translated

hangy:
Fixes #1643


teolemon:
Cool test. I was expecting Travis to kick in. Can you make it a regular PR so that we can see the test failing ? @hangy

hangy:
> Cool test. I was expecting Travis to kick in. Can you make it a regular PR so that we can see the test failing ? @hangy

Looks like it does not run if the branch is still a work in progress. :)

Unfortunately, the test does not work, because `<<site_name>>` is already replaced when it gets run, so I'll have to figure out a different approach.

hangy:
Fixed. See https://github.com/hangy/openfoodfacts-server/runs/71686874 for a failing test and https://github.com/hangy/openfoodfacts-server/runs/71690755 for the one fixed `.po` files.

Example error:
```
t/i18n.t .................. 1/? 
#   Failed test ''app_please_take_pictures' in 'ku' should contain '<<site_name>>''
#   at t/i18n.t line 22.
Wide character in print at /home/travis/perl5/lib/perl5/Test2/Formatter/TAP.pm line 113.
#                   '<p>Ev berhema e yet di ne<<site_nav>> di nav Htmlê da ne. To lê zêde bike li ser products, barcodes, lîsteya maddeyên û maddeyên bi we hinek pictures <a href="https://world.openfoodfacts.org" target="_blank"><<site_nav>></a>?</p>
# <p>Spas dikim di pêş!</p>
# '
#     doesn't match '(?^:<<site_name>>)'
```

PR #1695, opened 2019-03-07
Don't overwrite `no_nutrition_data` if a value was given in the request

### LINKED ISSUE #1191: no_nutrition_data is removed with the ingredient editing
Hello,

When saving an ingredient list after you edited it on the fly with the "edit ingredient" button on the product page, the no_nutrition_data box is unchecked.
It's very troublesome on OpenBeautyFacts because then the products is no longer Complete (it gets Nutrition facts to be completed).

The issue might happen on OFF too, didn't check

hangy:
**Description:** We should only overwrite `no_nutrition_data` if it is present in the POST/GET request. This avoids that this value is changed accidentally, because it was omitted from the request.
**Related issues and discussion:** Fixes #1191

### DIFF
--- a/cgi/product_jqm_multilingual.pl
+++ b/cgi/product_jqm_multilingual.pl
@@ -1,22 +1,22 @@
 #!/usr/bin/perl -w
 
 # This file is part of Product Opener.
-# 
+#
 # Product Opener
-# Copyright (C) 2011-2018 Association Open Food Facts
+# Copyright (C) 2011-2019 Association Open Food Facts
 # Contact: contact@openfoodfacts.org
 # Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
-# 
+#
 # Product Opener is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Affero General Public License as
 # published by the Free Software Foundation, either version 3 of the
 # License, or (at your option) any later version.
-# 
+#
 # This program is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


stephanegigandet:
If I remember correctly, the issue with the checkbox is that we cannot know if the box was unchecked, or if it was not there in the request: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox
"Note: If a checkbox is unchecked when its form is submitted, there is no value submitted to the server to represent its unchecked state (e.g. value=unchecked); the value is not submitted to the server at all."

But since the calls to product_jqm_multilingual.pl cannot ommit fields, then your change is ok I think. We won't be able to unset the checkbox, but that's better than always unsetting it.

hangy:
> If I remember correctly, the issue with the checkbox is that we cannot know if the box was unchecked, or if it was not there in the request: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox
"Note: If a checkbox is unchecked when its form is submitted, there is no value submitted to the server to represent its unchecked state (e.g. value=unchecked); the value is not submitted to the server at all."

Yup, but `product_jqm_multilingual.pl` is only used for the [write API](https://en.wiki.openfoodfacts.org/API/Write), not the full form, though. That should be `product_multilingual.pl`, no?

> But since the calls to product_jqm_multilingual.pl cannot ommit fields, then your change is ok I think. We won't be able to unset the checkbox, but that's better than always unsetting it.

I thought, calls to `product_jqm_multilingual.pl` can omit fields. That how the ad-hoc editing of the ingredients text works. If `no_nutrition_data` is omitted in production, then the checkbox would be affected. However, it should still be possible to unset the value through `product_jqm_multilingual.pl` by passing something like `?no_nutrition_data=off` in the query parameter.

stephanegigandet:
Right, sounds good!

Issue #455, opened 2016-09-07
Search in Russian: consider "е" and "ё" interchangable in Russian


aleksejrs:
Find "е" when searching for "ё", and "ё" when searching for "е".

stephanegigandet:
Are those the only interchangeable characters?

And can you think of any conflicts where a word/name with "e" means something different than the same word with "ë"?

aleksejrs:
> Are those the only interchangeable characters?

Yes, for letters.  "ё"/"Ё" has usually been written as "е"/"Е". https://en.wikipedia.org/wiki/Yo_(Cyrillic)#Russian
As for other characters, there is, of course, punctuation like "«»", "„“" etc.

> And can you think of any conflicts where a word/name with "e" means something different than the same word with "ë"?

That is a problem, but probably not one we can do anything about. It will probably mostly affect place names.

teolemon:
fr:pâtés
fr:pates

aleksejrs:
I am talking only about search (where you enter search keywords and get results, not even graphs).

There has already been a problem which resulted in some labels or categories to have "и" instead of "й".

aleksejrs:
Although categories might have no problem if е/ё synonyms are auto-generated.  I don't know.

hangy:
We need to see if this is something that needs to be implemented locale-aware. We have had bug reports where "de:Bürger" was automatically deaccentuated to "de:Burger", even though those are not interchangeable.

aleksejrs:
"Ё" is a separate letter, not a "е" with an umlaut.

hangy:
True, in German they are just vowels with umlauts, but they canned be used interchangeably either. In Finnish or Swedish, "Ö" is considered a separate letter, too, but deaccentuated by OFF all the same. Not saying we should not do it, I just want to raise caution. 👍

aleksejrs:
You could instead enforce the use of "ё" where the label says "е" instead of "ё". ;)

PR #1696, opened 2019-03-07
Don't always remove accents from characters

### LINKED ISSUE #458: Pâtés et Pates are merged together
fr:pâtés
fr:pates
are amalgamated when typing the category from world.off

hangy:
**Description:** This removes the unconditional removal of accents and conversions like œ to oe.
**Related issues and discussion:** Fixes #458


stephanegigandet:
get_fileid() is used in many different places, I'm a bit afraid of unwanted / unforeseen consequences. In which cases would we not remove the accent? By language? Unaccenting in French is a good thing in 99% of the cases (with a few conflicts like pâte and pâté).

hangy:
> get_fileid() is used in many different places, I'm a bit afraid of unwanted / unforeseen consequences.

Me too. I added an optional parameter that allows the previous behaviour to be used.  I intend to use that in places where actual file access done (ie. user's `.sto` file).

> In which cases would we not remove the accent? By language? Unaccenting in French is a good thing in 99% of the cases (with a few conflicts like pâte and pâté).

I can (mostly) only speak for German, where it is almost never a good idea to replace ie. `ö` by `o` (see https://github.com/openfoodfacts/openfoodfacts-server/issues/455#issuecomment-246099234, too). Also, it is needlessly lossy to replace `ß` by `ss` (which can also lead to collisions) and not consequent if Unicode characters in other scripts (ie. cyrillic) are kept intact.

Could you explain why unaccenting is a good idea for French? Is it because the characters are mostly similar, so users might be lazy when entering them? At least for German, Umlauts aren't usually treated the same as non-Umlauts. I can't speak for Swedish, Estonian, Finnish, Hungarian, or Icelandic.

Maybe we need a `unac_string_perl_light` version that does not replace Umlauts?

stephanegigandet:
For French, it is very common for French users to type search queries etc. without accents. For oe, almost everyone types oeuf or boeuf (egg and beef) because it's too complex to enter the oe character. It also has the benefit of creating ascii file names and urls, with no escaping needed. In French, capital letters are also very often not accented.

hangy:
> It also has the benefit of creating ascii file names and urls, with no escaping needed.

Yup, but as URLs like https://ru.openfoodfacts.org/%D1%83%D0%BF%D0%B0%D0%BA%D0%BE%D0%B2%D0%BA%D0%B0/%D0%BA%D0%B0%D1%80%D1%82%D0%BE%D0%BD (and the tags this is based on) are present in the projects, anyone who uses the data or API will need to support non-ASCII filenames anyways. 😉

CharlesNepote:
Not sure to understand the debate but here are my 2 cents.

> For French, it is very common for French users to type search queries etc. without accents. For oe, almost everyone types oeuf or boeuf (egg and beef) because it's too complex to enter the oe character. It also has the benefit of creating ascii file names and urls, with no escaping needed. In French, capital letters are also very often not accented.

Yes it's true for oeuf or boeuf. But accents are actually used in search queries because accents **make sense**: pâte and pâté doesn't have the same meaning. Someone listed more than 12,000 (!) french terms with a different meaning based on the accents: https://french.stackexchange.com/questions/24380/mots-avec-et-sans-accent-avec-un-sens-diff%c3%a9rent
Many words are related to food which lead to confusion. When I read "mais" it doesn't "sound" *maïs* for me but the common word "mais".

Search *queries* should hold "pâté" and "pâte" as the same word but printed information (title, url, category, etc.) should make the difference.

hangy:
Thanks for your input @CharlesNepote! As you have probably figured out already, this PR is about the issue that ie.  "fr:Pâté" and "fr:Pate" are treated identically by ProductOpener, because accents are removed before the tag is actually stored.

> Search queries should hold "pâté" and "pâte" as the same word but printed information (title, url, category, etc.) should make the difference.

To me, this sounds like merging this PR should make sense for most languages. Maybe we could implement your suggestion of ignoring the accents for search by using the old logic in when writing a product's `_keywords` and when normalizing the search terms https://github.com/openfoodfacts/openfoodfacts-server/blob/0bbc249b86e38f9110a5643325e9e0f93969c409/cgi/search.pl#L574

stephanegigandet:
We need to evaluate the benefits and risks of doing this. At this point I see very little benefits and a lot of risks... 

Benefits:

1. We can differentiate words that have the same unaccented form but have different meanings. (e.g. "pâtes" and "pâtés" in French).
This is only useful when we have 2 words in the same category (e.g. ingredients or food categories like "pâtes" and "pâtés"). Other conflicts like "mais" and "maïs" in French are not issues as "mais" means "but" and is not an ingredient, so no risk of conflict.

2. Better looking urls in things that display UTF8 urls correctly. (but worse looking urls in things that don't).

Questions:
- besides "pâtes" and "pâtés", how many other actual conflicts do we have (in all languages)?
- are there other ways to deal with those conflicts?

Risks:
1. We will lose correct mappings when entries are not accented (e.g. all caps ingredients without accents, and words typed without accents (e.g. the very common "oeuf" and "boeuf" in French).
To some extent, this can be avoided by adding synonyms in the taxonomies.
2. No results in search if user types without accents (or with wrong accents). An option is to deaccent for indexing and search (but then what is the point?)
3. This is going to change the canonical forms and urls of all existing entries, for all taxonomies. This it what makes me very unconfortable as it will be very difficult to make sure that it's not going to break something. We will have to recompute all taxonomies and recompute all the tags of all products at the same time.

hangy:
First of all, I think we may need to treat umlauts and accent separately for the sake of this discussion. In German, umlauts should not be removed for canonical forms of tags, because they're simply not used that way. This is an unexpected behaviour for most native speakers. For a keyword search, a fuzzy or stemmed search would obviously be acceptable. Apparently, this is the same in Russian, as asked in #455 by @aleksejrs. For languages that use a lot of accents, I don't have the knowledge to judge.

> 1. We can differentiate words that have the same unaccented form but have different meanings. (e.g. "pâtes" and "pâtés" in French).
>    This is only useful when we have 2 words in the same category (e.g. ingredients or food categories like "pâtes" and "pâtés"). Other conflicts like "mais" and "maïs" in French are not issues as "mais" means "but" and is not an ingredient, so no risk of conflict.

Keep in mind that the issue is not limited to highly taxonomized tags like categories or ingredients, but also applies to currently free text like brands or packaging.

> * besides "pâtes" and "pâtés", how many other actual conflicts do we have (in all languages)?

For example, [Bürger](https://de.openfoodfacts.org/marke/bürger) and [Burger](https://de.openfoodfacts.org/marke/burger) should be separate brands. As the `get_urlid` is also used to build the `titleid` in the `product_url`, this also creates incorrect product URLs like https://de.openfoodfacts.org/produkt/3083680003841/erbsen-mohrchen-zart-und-fein-bonduelle  which should really be https://de.openfoodfacts.org/produkt/3083680003841/erbsen-möhrchen-zart-und-fein-bonduelle for two reasons: Firstly, "Mohrchen" isn't a German word, unless you'd use it as a diminutive for the racist "Mohr" … I hope this makes it obvious how it's impossible to create a list for all languages and fields where the current behaviour is incorrect or problematic.

> * are there other ways to deal with those conflicts?

I don't know. From my point of view, the current `get_urlid` produces actually incorrect data.

> 1. We will lose correct mappings when entries are not accented (e.g. all caps ingredients without accents, and words typed without accents (e.g. the very common "oeuf" and "boeuf" in French).
>    To some extent, this can be avoided by adding synonyms in the taxonomies.

As I haven't experienced this in German, I have no strong opinion on those cases. Here, "O" and "Ö" are different characters (not the same character with a hint about it's pronunciation), and the diacritic wouldn't be removed from an an upper-case word.

> 2. No results in search if user types without accents (or with wrong accents). An option is to deaccent for indexing and search (but then what is the point?)

For `_keyword` search, I think it is totally fine to have fuzzy matches. That's probably something that users expect anyways.

> 3. This is going to change the canonical forms and urls of all existing entries, for all taxonomies. This it what makes me very unconfortable as it will be very difficult to make sure that it's not going to break something. We will have to recompute all taxonomies and recompute all the tags of all products at the same time.

Agreed that this needs to be done carefully. However, if aliases were added where the deaccented version is identical to the non-deaccented version, then this wouldn't break any queries or links.

stephanegigandet:
I'm thinking that it could be a good idea to make get_fileid / get_urlid language specific. That would allow us to make changes on a language by language basis. And I would feel more confortable if we test those changes in one language first (German seems to be a good candidate as our current behavior is incorrect).

It would also allow us to make language specific exceptions to language specific rules (e.g. in French, we could keep the current behaviour, except for the few cases where we have issues like pâte and pâté).

hangy:
> I'm thinking that it could be a good idea to make get_fileid / get_urlid language specific. That would allow us to make changes on a language by language basis. And I would feel more confortable if we test those changes in one language first (German seems to be a good candidate as our current behavior is incorrect).

Both methods are used in a lot of places, so that it could be difficult to fix the current buggy behavior in languages other than French in a timely manner. I just skimmed the code and I'm not even sure if every relevant usage of the methods does have a `$lc` available to differentiate based on the language.

Lunique:
I can only agree with hangy that as a native german speaker it is highly uncommon to have the ö, ä, ü just removed instead of turning them into oe, ae, ue. German words just don't work this way and it looks cheap to see it on pages like the category list. The turning ß to ss is quite common in the digital age although it still can change the meaning of a word ("weiß" is a colour "weiss" means knowing something).
Could you turn "ü" to "ue" instead of "u" but when searching for "huhn" it would look for both "huhn" and "huehner" ? So "u" and "ü" would always work to look for both "u" and "ue" within a word (vice versa)? Or is that too complicated? (am I even making sense here?)

hangy:
The typical ü=>ue conversion is really useful - albeit lossy - in German for cases where the platform/software does not know any better. However, removing the trema isn't applicable to all languages. To quote [Wikipedia](https://en.wikipedia.org/w/index.php?title=Diaeresis_(diacritic)&oldid=900441591#Borrowing_of_German_umlaut_notation):
> As the borrowed diacritic has lost its relationship to Germanic i-mutation, they are in some languages considered **independent graphemes**, and cannot be replaced with ⟨ae⟩, ⟨oe⟩, or ⟨ue⟩ as in German. In Estonian and Finnish, for example, these latter diphthongs have independent meanings. […] The transformations ä → ae and ö → oe can, therefore, be considered less appropriate for these languages, […]

Therefore, we'd need to make these methods locale-specific, as suggested by [stephane](https://github.com/openfoodfacts/openfoodfacts-server/pull/1696#issuecomment-473686041). Unfortunately, that's most likely pretty complex in the current source code.

> Could you turn "ü" to "ue" instead of "u" but when searching for "huhn" it would look for both "huhn" and "huehner" ? So "u" and "ü" would always work to look for both "u" and "ue" within a word (vice versa)? Or is that too complicated? (am I even making sense here?)

Yes, I believe that makes a lot of sense and is common practice when full-text search is performed. Actually, you can see this in some browsers. If you search (<kbd>CTRL</kbd>+<kbd>F</kbd> in a lot of browsers) for "ue" on this page, Google Chrome will (at least on my PC 😉) also highlight matches of "ü". It would have the advantage that "de:bürger" and "de:burger" in the URL could still lead to different pages.

hangy:
@stephanegigandet I have reworked this PR to consider the tag's `lc` when performing `get_fileid`. All of this is rebased on current `master`, so that the UUID userids etc. are used now. There are a few fixed unit tests, and additional unit tests for `get_fileid`. Since the compiled taxonomies depend on the `get_fileid` sub (at least indirectly), all taxonomies had to be rebuilt.

lib/ProductOpener/Store.pm
```diff
@@ -68,43 +68,44 @@ sub unac_string_perl($) {
...
-# 1. deaccent: é -> è, + German umlauts: ä -> ae
-# 2. lowercase
+# 1. lowercase
+# 2. deaccent: é -> è, + German umlauts: ä -> ae if $unaccent is 1 OR $lc is 'fr'
 # 3. turn ascii characters that are not letters / numbers to -
 # 4. keep other UTF-8 characters (e.g. Chinese, Japanese, Korean, Arabic, Hebrew etc.) untouched
 # 5. remove leading and trailing -, turn multiple - to -
 
-sub get_fileid($) {
+sub get_fileid {
```
stephanegigandet: I would prefer to keep the ($$$) to indicate that the arguments are mandatory, so that we get errors at compile time. Otherwise we might forget one argument by error.

Maybe we could change the name of the function (I named it at the very beginning of the project when it was only used for filenames, but now it's mostly used for tag names). How about something like this?

sub get_string_id_for_language($$) {

my $string = shift;
my $string_lc = shift; # undef to get the default normalization for strings that are not in a specific language like user names

Then accenting / deaccenting is decided from the $string_lc parameter.

hangy: https://stackoverflow.com/a/297265/11963 claims that
> Perl's prototypes have a singular and very different purpose:
> Prototypes allow you to define functions that behave like built-in functions.

Because of that, I'm a bit torn about keeping them. I was actually considering creating an issue about removing them altogether, but that's off-topic for this PR. :)

I'll have a look at renaming the sub and will integrate it into this PR if it goes well.

hangy: > I'll have a look at renaming the sub and will integrate it into this PR if it goes well.

Doesn't work without creating a duplicate or misusing the `$string_lc` parameter, because we need forced unaccenting for actual file IO  (ie. reset_password.pl finds a `"$data_root/users/id.sto"` file with that).

hangy:
I'll try to rebase and fix the newly broken unit tests if there's a chance to merge this in a timely manner. Otherwise, it's getting too annoying to solve new merge conflicts all the time.

stephanegigandet:
@hangy : I just got back to vacations, I can test it and merge it in the next few days.

lib/ProductOpener/Store.pm
stephanegigandet: For things that in fact don't have a language (like usernames), we can just pass undef as the language, and we apply the current default unaccenting. The decision to unaccent a specific language can be put in the configuration file (and left to the people who administer a particular instance), instead of being set in the code in the calls to the normalization function.

stephanegigandet:
I deployed this branch on the dev server, can everyone check that everything works as expected?
https://world.dev.openfoodfacts.org (login and pass : off / off )

Issue #423, opened 2016-08-20
Uploading a picture sometimes overwrites an existing one


aleksejrs:
The old thumbnail stays in the browser's cache.

hangy:
Could be related to #304. I wonder if it might be better to use a different identifier for the Images instead of an auto-incremented number. Even though the software should increment the number _and_ check that the Image does not exist, this seems to fail sometimes. (Wel, that's my best guess right now.)

hangy:
I think we should use a digest of the file content as the filename, ie. ` my $digest = Digest::SHA3->new(224)->addfile("$www_root/images/products/$path/$imgid.$extension", "b")->hexdigest;`. That should avoid most collisions that could be caused by automatically incremented integer imageids. By using the same hash as the filename, duplicate checks should be pretty inexpensive.

aleksejrs:
This has happened recently a couple of times just to the thumbnails, which were OK after saving and reopening the edit page. Not saying the rest cannot be affected: I avoid uploading multiple images at once.

PR #1699, opened 2019-03-09
Update Image Upload

hangy:
1. Perform image validation and existence checks before copying images to the product's image directory.
2. Validate the image's file type using PerlMagick.
3. Use the SHA3 hash of the image content as the `$imgid`, so that we do not need to have a separate counter (which seems to be problematic, because images sometimes get overwritten, see #423).

### DIFF
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -490,11 +490,12 @@ sub process_image_upload($$$$$$) {
 			# create the the directories for the product
 			make_path("$image_base_path/", { chmod => 0755 });
 
-			$imgid = Digest::SHA3->new(224)->addfile($file, 'b')->b64digest;
+			my $hash = Digest::SHA3->new(224)->addfile($file, 'b')->b64digest;
+			$imgid = time() . "." . $hash;
 			local $log->context->{imgid} = $imgid;
 			my $img_path = "$image_base_path/$imgid.$extension";
 			# Check that we don't already have the image with the same name
-			if (-e $img_path) {
+			if (glob("$image_base_path/*.$hash.$extension")) {
 				return -3;
 			}
 
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -455,6 +455,9 @@ sub process_image_upload($$$$$$) {
 		$log->trace("validating file type") if $log->is_trace();
 		my $validate = Image::Magick->new();
 		my ($width, $height, $size, $format) = $validate->Ping($file);
+		local $log->context->{width} = $width;
+		local $log->context->{height} = $height;
+		local $log->context->{size} = $size;
 		local $log->context->{format} = $format;
 		if (not (defined $format)) {
 			$log->info("file type invalid") if $log->is_info();
@@ -462,6 +465,13 @@ sub process_image_upload($$$$$$) {
 		}
 		else {
 			$log->debug("file type validated") if $log->is_debug();
+
+			# Check the image is big enough so that we do not get thumbnails from other sites
+			if (  (($width < 640) and ($height < 160))
+				and ((not defined $options{users_who_can_upload_small_images})
+					or (not defined $options{users_who_can_upload_small_images}{$userid}))){
+				return -4;
+			}
 
 			my $extension = 'jpg';
 			if ($file =~ /\.(gif|jpeg|jpg|png)$/i) {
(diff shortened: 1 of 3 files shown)


hangy:
This works well in my container until it tries to access `Image::Magick`, which just segfaults. 😕

stephanegigandet:
# Keep original in case we need it later  -> I think that's an old comment, we don't store it anymore (except for the gif / png that don't get overwritten as you remarked).

I would prefer to keep numeric ids in sequence, so that the order they have been sent is obvious.

Using the hash instead of the size is a good idea, but then that means we need to compute it every time for all images in the directory. The size was kind of a quick way to find duplicates, even if it's imperfect.

hangy:
> Using the hash instead of the size is a good idea, but then that means we need to compute it every time for all images in the directory.

The hash does not need to recomputed if it is used as the filename. :wink:

> I would prefer to keep numeric ids in sequence, so that the order they have been sent is obvious.

What advantage is there is storing the files in numeric order? In most cases, the upload order should be visible in the product history or at least the file's creation date in the filesystem. We could also combine the numeric id and hash in the filename, if required. A check to see if `*.[hash].jpg` exists should work, too.

>> Keep original in case we need it later

> I think that's an old comment, we don't store it anymore (except for the gif / png that don't get overwritten as you remarked).

OK, so how about we just delete all source images to keep consistent behaviour?

hangy:
> > I would prefer to keep numeric ids in sequence, so that the order they have been sent is obvious.
> 
> What advantage is there is storing the files in numeric order? In most cases, the upload order should be visible in the product history or at least the file's creation date in the filesystem. We could also combine the numeric id and hash in the filename, if required. A check to see if `*.[hash].jpg` exists should work, too.

Actually, I think the numeric id being re-used because of some kind of race-condition might be something that caused images to be overridden. What about "time().$hash.$extension` instead? It would allow images to be in upload order and by checking with `if (glob("*.$hash.$extension")) {` we could check for dupes ...

lib/ProductOpener/Images.pm
```diff
@@ -467,32 +483,51 @@ sub process_image_upload($$$$$$) {
...
 
-			# Create the directories for the product
-			foreach my $current_dir  ($www_root . "/images/products") {
-				(-e "$current_dir") or mkdir($current_dir, 0755);
-				foreach my $component (split("/", $path)) {
-					$current_dir .= "/$component";
-					(-e "$current_dir") or mkdir($current_dir, 0755);
-				}
+			my $hash = Digest::SHA3->new(224)->addfile($file, 'b')->b64digest;
+			$imgid = time() . "." . $hash;
```
stephanegigandet: The imgid is used in the id field on input elements in the product_multilingual.pl form, it may be better to use '_' instead of '.' as a separator. 

The hash base 64 digest is also very long and can containt plus signs, slashes etc. I don't think that's a good idea for file names:

rOJhUs3nd21jgPdk7ag2fTA7/Q0/AWuAcAVwuQ.rOJhUs3nd21jgPdk7ag2fTA7/Q0/AWuAcAVwuQ

It is also very long, those imgids will be in file names, in urls, in mongo etc. I think we should put something much shorter. e.g. CRC32

lib/ProductOpener/Images.pm
```diff
@@ -541,47 +576,9 @@ sub process_image_upload($$$$$$) {
 			}
 
 			$source->Set('quality',95);
-			$x = $source->Write("jpeg:$www_root/images/products/$path/$imgid.jpg");
-
-			# Check that we don't already have the image
```
stephanegigandet: We still need to keep the file size check so that we check collisions with the existing files that have no hash in the filename. That's in particular useful for imports of data from other databases, producers etc. It may not be efficient, but currently we import again existing images and they are discarded when uploaded.

lib/ProductOpener/Images.pm
```diff
@@ -467,32 +483,51 @@ sub process_image_upload($$$$$$) {
...
+					else {
+						$log->warn("image missing in product.sto");
+					}
+				}
 			}
 
-			local $log->context->{imgid} = $imgid;
 			$log->debug("new imgid determined") if $log->is_debug();
 
+			my $lock_path = "$image_base_path/$imgid.lock";
```
stephanegigandet: The aim of the lock was to prevent 2 images with the same imgid to be created (when an user sends multiple images and 2 different perl processes run product_image_upload.pl, if we use a hash, we don't need the lock.

stephanegigandet:
The imgid field is used in many different places: file names for images, HTML ids, in Mongo, in the JSON returned in the API, possibly in apps. I can see the value of using a hash, but I'm concerned about unwanted and possibly unforeseen effects:
1. issues with characters in file / url / html / json field names. e.g. "/", "." etc. returned by the base 64 encoding
2. very long file names (timestamp + hash) / fields -> makes everything bigger in the db, html, json, CSV file etc.
3. things that expect imgids to be numbers in sequences.
e.g. in product_image_upload. pl : "		if ($imgid_returncode < 0) { "
in Images.pm : "for (my $imgid = 1; $imgid <= ($object_ref->{max_imgid} + 5); $imgid++) {"

We can try to resolve those using another checksum like a simple CRC32, do a grep for imgid in all the perl and javascript code etc. to make sure everything works.

I am not sure if the reward is worth the risk of breaking something though. We don't really know what goes wrong in bug #423 (or if it still happens) . There is currently a lock which should prevent the same imgid to be assigned to different images. The race condition might be elsewhere. In fact there is probably another race condition when multiple files are uploaded, regardless of the imgid : the product is loaded, the images field are updated, and the product is saved. Even if we make sure we have different imgids, one save can still overwrite the other.

hangy:
I can see that my attempt doesn't help the problem, so I'll close this PR in favour of other fixes.

> In fact there is probably another race condition when multiple files are uploaded, regardless of the imgid : the product is loaded, the images field are updated, and the product is saved. Even if we make sure we have different imgids, one save can still overwrite the other.

Yes, I'm also pretty sure that there are race conditions on those scenarios,

PR #1714, opened 2019-03-17
Add ORY Hydra based login

### LINKED ISSUE #1204: Enable a token based auth for the mobile app

hangy:
This enables OAuth 2.0 to be used with the Write API by using ProductOpener as a [ORY Hydra Consent App](https://github.com/ory/hydra-login-consent-node).

This means that Hydra manages the nitty-gritty details of OIDC/OAuth, and we just provide the login form. That way, we do not have to encryption keys, token expiration, etc. in ProductOpener.

This currently does not contain any client management (for end-users), but that can be done afterwards. I believe it would be good to start with as-is.

**Related issues and discussion:** Fixes #1204


stephanegigandet:
I really don't know much about authentication, so I can't really review this feature. My concern is more on the need to add another server for hydra, which makes one more dependency. And if something goes wrong with it, I don't know anything about it... We do not use docker in production, so it means setting up and running an hydra server separately.
Maybe others can have a look and comment. @offvince @cquest

hangy:
Running ORY Hydra in Docker is just one possible deployment option, but as there already seems to be some production Docker running Robotoff, it might be fast.

If you're not comfortable running hydra in Docker, it's always possible to compile it from source and run it as a systemd script, proxied by nginx. It does need some kind of [SQL](https://www.ory.sh/docs/next/hydra/dependencies-environment#sql) backend, though.

The reason why I opted to try hydra is that there's no good library to implement a working OIDC server in Perl. Therefore, using an externally hosted server for this task sounds like the best option at the moment. Obviously, that's also better for consumers than us trying to implement a custom token authorization.

stephanegigandet:
We are going to try to continue @hangy's work and deploy ORY Hydra based login in OFF, @landonpattison is a Google Summer of Code 2022 intern who will be working on it. The project page is here: https://wiki.openfoodfacts.org/GSOC_2022_-_External_Authentication_Service

PR #1725, opened 2019-03-27
pt

teolemon:


po/common/pt.po
```diff
@@ -650,557 +639,573 @@ msgid "<a href=\"https://world.openfoodfacts.org/legal\">Legal</a> -\n"
...
-msgstr ""
+msgstr "/termos-de-utilizacao"
 
 msgctxt "footer_who_we_are"
 msgid "Who we are"
-msgstr ""
+msgstr "Quem somos"
 
+# Do not translate without having the same exact string in the Tags template. Do not use spaces, special characters, only alphanumeric characters separated by hyphens
 msgctxt "footer_who_we_are_link"
```
stephanegigandet: Those will make links break, but we can put symlinks to make them work again.

teolemon:
@stephanegigandet This is a copy paste of pt_PT.po to pt.po so that ProductOpener can display the updated text

PR #1726, opened 2019-03-27
nl

teolemon:
nl


teolemon:
@aleene @stephanegigandet This is a copy paste of nl_NL.po to nl.po so that ProductOpener can display the updated text

PR #1744, opened 2019-04-22
Ingredients patch 20 aleene

aleene:
The 20th set of ingredients patches.


teolemon:
@stephanegigandet what do you make of the Nutrient taxo ? simple to integrate ?
@aleene can you perhaps make a separate PR for it ?

aleene:
I should have created a separate pull for the nutriments file, but as it is not used yet, it does not matter much.

PR #1761, opened 2019-05-09
Quality tag for max energy

hangy:
As discussed on Slack: https://openfoodfacts.slack.com/archives/C7FPRUXGB/p1557247920048200

### DIFF
--- a/lib/ProductOpener/SiteQuality_off.pm
+++ b/lib/ProductOpener/SiteQuality_off.pm
@@ -540,7 +540,7 @@ sub check_nutrition_data($) {
 
 		# Fat has the highest nutrient energy density of 37 kJ/g. With some leeway/inaccuracies,
 		# it is safe to assume that no product can have a higher energy of 3800 kJ.
-		if ((defined $product_ref->{nutriments}{energy}) and ($product_ref->{nutriments}{energy} gt 3800)) {
+		if ((defined $product_ref->{nutriments}{energy}) and ($product_ref->{nutriments}{energy} > 3800)) {
 			push @{$product_ref->{quality_tags}}, 'illogically-high-energy-value';
 		}
 	}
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/SiteQuality_off.pm
```diff
@@ -401,145 +401,149 @@ sub detect_categories ($) {
...
-	
-	
+
+				push @{$product_ref->{quality_tags}}, "nutrition-saturated-fat-greater-than-fat";
+
+		}
+
+		# Fat has the highest nutrient energy density of 37 kJ/g. With some leeway/inaccuracies,
+		# it is safe to assume that no product can have a higher energy of 3800 kJ.
+		if ((defined $product_ref->{nutriments}{energy}) and ($product_ref->{nutriments}{energy} gt 3800)) {
```
stephanegigandet: In Perl, gt is used for string comparisons
```suggestion
		if ((defined $product_ref->{nutriments}{energy}) and ($product_ref->{nutriments}{energy} > 3800)) {
```

t/sitequality.t
```diff
@@ -0,0 +1,66 @@
...
+ok( !has_tag($product_ref_with_low_energy_value, 'quality', 'illogically-high-energy-value'), 'product does not have illogically-high-energy-value tag as it has an energy_value below 3800' );
+
+# illogically-high-energy-value - does not add tag, if energy_value is equal 3800
+my $product_ref_with_lowish_energy_value = {
+	lc => "de",
+	nutriments => {
+		energy => 3800
+	}
+};
+
```
stephanegigandet: I think that if you add a test with energy => 40, it will fail:

perl -e "print 40 gt 3800;"
1

hangy: Thanks for the input! I added the test, and at least after applying your suggestion, `40` seems to work fine. 🙂

Issue #1707, opened 2019-03-13
Notify Robotoff when a product is created/updated/deleted


raphael0202:
Notify Robotoff when a product is created/updated/deleted, so we can:
- generate insights (ex: category)
- delete insights associated with deleted products
- delete insights if the attribute has been manually provided (ex: a user manually set the category, the category insight is not necessarily relevant anymore)

For `created` actions, the best would be to wait for the form to be sent before triggering a notification (in other words, don't send a notification as soon as the images are uploaded)

Send an application/x-www-form-urlencoded `POST` request to `/api/v1/webhook/product` with the following parameters:
- barcode (string, required)
- action (string, required) : one of `updated`, `deleted`, `created`
- updated_fields (str, comma separated list): list of updated fields (only for `updated` action)

stephanegigandet:
Currently the form does not keep track of what has been updated, can I just notify that it has been updated? Then you can delete all current insights and recompute new ones. Is that ok?

hangy:
> Currently the form does not keep track of what has been updated, can I just notify that it has been updated? Then you can delete all current insights and recompute new ones. Is that ok?

Don't we have the changed fields available when the changelog entry is created?

raphael0202:
@stephanegigandet Yes just notifying that the product has been updated is enough, no need to provide a detailed changelog

raphael0202:
@stephanegigandet I believe this issue can be closed?

PR #1765, opened 2019-05-10
Use $robotoff_url

hangy:
**Description:** There were a few cases where the OFF development server pushed it's images to the productive Robotoff instance. This should only be done if the `$robotoff_url` was configured correctly.

**Related issues and discussion:** #1707

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -53,7 +53,7 @@ BEGIN
 		&add_back_field_values_removed_by_user
 
 		&process_product_edit_rules
-		
+
 		&make_sure_numbers_are_stored_as_numbers
 		&change_product_server_or_code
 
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Products.pm
```diff
@@ -110,9 +110,9 @@ sub make_sure_numbers_are_stored_as_numbers($) {
 			# fields like "salt", "salt_value"
 			# -> used internally, should not be used by apps
 			# store as numbers
-			elsif (looks_like_number($product_ref->{nutriments}{$field}))  {
```
stephanegigandet: Was this removed intentionaly?

hangy: I don't think so. Actually, I don't remember editing this line in this PR. The only intentional change was regarding the `$robotoff_url`. The rest should only be changed because of `.editorconfig`. Let me check and fix this. 🙂

hangy: Fixed. Thanks!

PR #1768, opened 2019-05-11
Use Text::CSV for CSV output

### LINKED ISSUE #161: CSV Generator not send all columns for all products
# Steps Reproduction
1. Access to http://world.openfoodfacts.org/cgi/search.pl
2. Search for Criteria: "Country" Value: "Chile"
3. Export Result in CSV
# Example error data

Product "7802800567034" column "image_url"
- In Cell "AU3" using libreoffice the value is "Soja"
- In Cell "BW3" the value is "http://world.openfoodfacts.org/images/products/780/280/056/7034/front.4.400.jpg"

hangy:
**Description:** I replaced the custom CSV output with the `Text::CSV` package that is currently being used in `ProductOpener::Import`. Because the format is identically, the output is compatible with the previous export and should not cause any problems with consumers.

**Related issues and discussion:** I believe this fixes #161, because `Text::CSV` will take care of escaping CSV values. Additionally, output will be sped up, because less strings will be built to produce the output.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -121,6 +121,7 @@ use experimental 'smartmatch';
 use MongoDB;
 use Tie::IxHash;
 use JSON::PP;
+use Text::CSV;
 use XML::Simple;
 use CLDR::Number;
 use CLDR::Number::Format::Decimal;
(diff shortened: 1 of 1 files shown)


stephanegigandet:
Hi @hangy,
Have you tried to do a diff of the CSV files generated before and after this change?
If you did, are there differences in the output?

hangy:
For my test dataset, the output was identical. The actual advantage of this module should be that it allocates less strings, and has built-in support for quoting/escaping values according to the RFC.

stephanegigandet:
Perfect, thank you.

PR #1781, opened 2019-05-17
Other Nutri translations

aleene:
I merge the same translations from ingredients into other nutritional values.


VaiTon:
@stephanegigandet @teolemon can we merge?

PR #1787, opened 2019-05-19
Categories translations

aleene:
Merged translations from the ingredients and categories taxonomy.


stephanegigandet:
The categories are in plural though, whereas the ingredients are in singular. I don't know if there's a way to get the plurals from wikidata?

aleene:
No, singular-plural difference is not in the wikidata. 
In dutch plurals for categories does not work, should be mostly in singular.
I wonder if the users think of categories in plural or singular.

aleene:
I put the plurals in front in the categories and viceversa in the ingredients. So should be okay.

aleene:
If there are plurals available, I put them first in category taxonomy. But the wikidata does not always have those. Are you ok with this approach?

stephanegigandet:
OK, sounds good.

Issue #1810, opened 2019-05-27
Prevent all apps (OFF + 3rd party apps) to modify images and data sent by manufacturers


stephanegigandet:
Apps should still be able to send new pictures, but only to select them if the image is missing (e.g. no ingredient image sent by the manufacturer).

stephanegigandet:
There are some edit rules in place to prevent that already, but they are not always working.

CharlesNepote:
Many manufacturers have recently put the nutricore on their products. Will we have to wait for their new pictures to see it? Even if users publish updated photos?

Wouldn't it be better to provide a validating process for special products?
- Regular user modifies the product (whatever modification it is).
- If the product is flagged as "protected" then the modification needs moderator/editor validation.
- The user is told that his data will be published after a validation.

We could then protect:
- automatically, products from manufacturers
- automatically, verified products
- automatically, most scanned complete products
- manually (or automatically), sensitive products such as Coca-Cola, Nutella...

"Editors" could be new role between regular users and admins.

hangy:
How do you know which fields were added by a manufacturer? If the manufacturer only sends data in for one language, it should still be possible to update other fields. Also, if a manufacturers provides a producer code, other producer codes could also be valid...

stephanegigandet:
Apps will still be able to upload new pictures, and they will trigger robotoff changes (e.g. if there's a photo with the Nutri-Score label). Same for automatically recognized packaging codes.

We don't know exactly which fields were added by a manufacturer (there's some info about that in the sources field, but it's possible it was changed later).

My plan is that when a manufacturer has sent us data (with a bare minimum of ingredients in one language and nutrition facts) to:

1. Allow all editions on the Web site.
2. Restrict editions through the apps so that they can only add fields, but not edit or delete them.

aleene:
Guess there is a relation with the checked status. Ideally behaviour should be similar.

stephanegigandet:
For checked products, we don't know when the products change though. When we have data from manufacturers, we can somewhat assume that they have sent us the last version.

stephanegigandet:
Deployed in production.

PR #1687, opened 2019-03-03
HAL

hangy:
**Description:**

This is currently an experimental branch that tries to re-implement the API as REST API with HAL+JSON and HATEOAS in mind.

**Related issues and discussion:**
- Important? #280, #462, #510, #780, #976
- Maybe also: #501, #525, #704, #722

PR #1811, opened 2019-05-27
disallow edits from apps to product data and images sent by manufacturers

stephanegigandet:
https://github.com/openfoodfacts/openfoodfacts-server/issues/1810

### DIFF
--- a/cgi/product_image_crop.pl
+++ b/cgi/product_image_crop.pl
@@ -61,7 +61,17 @@ if (not defined $code) {
 	exit(0);
 }
 
-my $product_ref = process_image_crop($code, $id, $imgid, $angle, $normalize, $white_magic, $x1, $y1, $x2, $y2);
+# Check if we have a picture from the manufacturer
+
+my $product_ref = retrieve_product($code);
+
+if ((defined $product_ref) and (has_tag($product_ref,"data_sources","producers")) and (defined $product_ref->{images}) and (defined $product_ref->{images}{$id})
+	and (referer() !~ /\/cgi\/product.pl/)) {
+	print STDERR "product_image_crop.pl - skip image id $id for product code $code (data from producer) - referer: " . referer() . "\n";
+}
+else {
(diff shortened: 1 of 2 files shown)


VaiTon:
> It would be better to at least communicate the problem with the caller, so that they know that their data has _not_ been saved.

I strongly agree with that

stephanegigandet:
The thing is that the current API does not have a feedback mechanism to indicate that some fields were saved and some were not. And we do not want apps to resubmit the data. Returning something else than the current status: ok response is likely to have unwanted behavior in some apps (like not being able to submit a product).

Let me provide some context: we have 12 000 products (and growing) with data like ingredients and nutrition facts that has been provided by producers (and that is almost always correct). And we have bogus data that overwrite this data. In particular there are some bugs in the offline mode of the OFF apps that make ingredients and nutritions facts being erased.

So while I agree that the silent ignore is not ideal, it's bad data that is being ignored. I think it's time that we define a new version of the API, and this time we should incorporate a better feedback mechanism (like which fields were saved, or rejected and why).

hangy:
> So while I agree that the silent ignore is not ideal, it's bad data that is being ignored.

Oh, that's a bad situation ... and rejecting the whole edit might also be not too good for contributors ...

> I think it's time that we define a new version of the API, and this time we should incorporate a better feedback mechanism (like which fields were saved, or rejected and why).

Feel free to jump in on #1687 😉

hangy:
> > So while I agree that the silent ignore is not ideal, it's bad data that is being ignored.
> 
> Oh, that's a bad situation ... and rejecting the whole edit might also be not too good for contributors ...

I just stumbled upon this again while skimming the source, and was confused why it was done this way, again. 😆 To be honest, it's *potentially* bad data being ignored. The apps that write data using the API might very well be providing corrections/improvements to the data provided by a producer. Silently discarding parts of the edits might drive well-behaved, motivated contributors away.

PR #1814, opened 2019-05-27
Fix 1422

### LINKED ISSUE #1422: Use localStorage to add suggestions for most recently added tags in the autocomplete box
We are currently using tagsInput for tags like categories, labels, brands etc.  
A few of those fields like categories also use jquery.autocomplete. This is feature request to also store locally with localStorage the most recent tag (or tags) added by the current user for each type of field so that they can be suggested when the user clicks on the input field but has not typed in characters yet.  (feature request also listed in #1401)

One way to do it may be to try to piggy back on jquery.autocomplete so that we retrieve suggestions from localstorage when users have not typed anything yet.

https://github.com/xoxco/jQuery-Tags-Input

http://api.jqueryui.com/autocomplete/

$('#brands').tagsInput({
--
  | 'height':'3rem',
  | 'width':'100%',
  | 'interactive':true,
  | 'minInputWidth':130,
  | 'delimiter': [','],
  | 'defaultText':"add a brand"
  | });
  | $('#categories').tagsInput({
  | 'height':'3rem',
  | 'width':'100%',
  | 'interactive':true,
  | 'minInputWidth':130,
  | 'delimiter': [','],
  | 'defaultText':"add a category",
  | 'autocomplete_url': 'https://world.openfoodfacts.org/cgi/suggest.pl?lc=en&tagtype=categories&'

VaiTon:
Fixes #1422


stephanegigandet:
Hi @VaiTon , thanks a lot for the pull request! It seems there's a lot of cleanup in addition to the tagsInput change. I'm going to test all that on the dev server.
In the mean time, could you try to resolve the conflicts?
Thanks a lot!

stephanegigandet:
Hi @VaiTon , I tried to install your branch on the dev server, but I wasn't able to get suggestions from localstorage as described in bug #1422 . Does it work for you? Could you take some pictures to make sure that we are talking about the same thing? Thank you!

VaiTon:
How many tags should we store?
Also should they be divided for field?

VaiTon:
This is ready to be reviewed @stephanegigandet @hangy @teolemon

VaiTon:
@hangy done

VaiTon:
Bump

stephanegigandet:
@VaiTon very cool, thank you very much!

Would it be possible to:

1. Do not save the same suggestions more than once.
2. Do not suggest a tag that is already there.

?

image

stephanegigandet:
For the fields that do not have a taxonomy, suggest.pl won't return suggestions, but it seems that there is now a (broken) call to product.pl :

image

Issue #1813, opened 2019-05-27
Broken footer for 404 not found pages, and remove exit() calls


stephanegigandet:
404 pages have an extra header at the bottom of the generated page:

https://world.openfoodfacts.org/some-bogus-url

</body>
--
  | </html>
  | <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
  | <html><head>
  | <title>404 Not Found</title>
  | </head><body>
  | <h1>Not Found</h1>
  | <p>The requested URL /cgi/display.pl was not found on this server.</p>
  | <hr>
  | <address>Apache/2.4.25 (Debian) Server at world.openfoodfacts.org Port 80</address>
  | </body></html>

hangy:
Sounds like we need to send the non-success status-code and tell Apache that evertyhing is OK.

https://stackoverflow.com/a/8288955/11963
https://metacpan.org/pod/release/GOZER/mod_perl-1.31/faq/mod_perl_faq.pod#So-how-do-I-use-mod_perl-in-conjunction-with-ErrorDocument?

stephanegigandet:
Fixed by @hangy, thank you!

PR #1815, opened 2019-05-27
Flush HTTP headers and tell Apache that everything is OK

hangy:
**Description:** Try to avoid Apache's built-in error page.
**Related issues and discussion:** #1813

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -6256,6 +6256,9 @@ HTML
 	my $status = $request_ref->{status};
 	if (defined $status) {
 		print header ( -status => $status );
+		my $r = Apache2::RequestUtil->request();
+		$r->rflush;
+		$r->status(200);
 	}
 
 	binmode(STDOUT, ":encoding(UTF-8)");


hangy:
@stephanegigandet This seems to work fine in my Docker container. Can you see if it's the same on dev? 🙂

stephanegigandet:
@hangy : I tried on the dev server but it does not seem to change anything. You do not get the extra header at the bottom in docker?

stephanegigandet:
Maybe it's because of this ModRegistry thing we use. Or the exit() call in display_error... It would be great to remove it and more gracefully exit (which may be contributing to the memory leaks we have)

hangy:
> @hangy : I tried on the dev server but it does not seem to change anything. You do not get the extra header at the bottom in docker?

Not with this change applied ...

> Maybe it's because of this ModRegistry thing we use. Or the exit() call in display_error... It would be great to remove it and more gracefully exit (which may be contributing to the memory leaks we have)

I use the same in Docker https://github.com/openfoodfacts/openfoodfacts-server/blob/7022b7cbe3fa6aad6546f7776efdcc57b7cfa2ca/docker/backend-dev/conf/apache.conf, just because it was used in https://github.com/openfoodfacts/openfoodfacts-server/blob/015154723f19fa5ef0d7b631dc7514738fc6a3c7/conf/apache-2.4/sites-available/off.conf. I literally have no idea about how the CGI process model even works. :slightly_smiling_face:

lib/ProductOpener/Display.pm
```diff
@@ -6256,6 +6256,9 @@ HTML
 	my $status = $request_ref->{status};
 	if (defined $status) {
 		print header ( -status => $status );
```
stephanegigandet: There's already a print header that has been applied in the if elsif else loop above, that seems strange. Am I missing something?

hangy: Those only set cookies or cache policies. The status header is a separate header line, so it should not be (and doesn't appear to be) a problem to sends multiple `print header(...);` calls.

stephanegigandet: It generates an extra empty line, but it seems mod_perl parses the headers anyway, so I guess it's fine.

perl test_header.pl
Expires: Wed, 29 May 2019 09:25:24 GMT
Date: Thu, 30 May 2019 09:25:24 GMT
Content-Type: text/html; charset=UTF-8

Status: 200
Content-Type: text/html; charset=UTF-8

stephanegigandet:
I tested it again, it is working, I must have made a mistake when I tested it earlier on the test server. Thanks @hangy !

PR #1821, opened 2019-05-29
more-cutting-edge-words

teolemon:
more-cutting-edge-words


teolemon:
``` 
Bad name after INGREDIENTES' at /home/travis/build/openfoodfacts/openfoodfacts-server/lib/ProductOpener/Ingredients.pm line 775.
```

lib/ProductOpener/Ingredients.pm
```diff
@@ -741,23 +741,23 @@ fr => [
 ],
 
 es => [
-'ingredientes'(\s*)(\s|-|:|\r|\n)+
+'ingredientes'(\s*)(\s|-|:|\r|\n)+,
```
stephanegigandet: This is a string, the delimiters ' ... ' need to be at the beginning and end :

'ingredientes'(\s*)(\s|-|:|\r|\n)+, --> 'ingredientes(\s*)(\s|-|:|\r|\n)+',

--> same for all changes below

Otherwise it won't compile.

teolemon: fixed

aleene:
You can add the other stopwords as well, as defined in the ingredients taxonomy.

PR #1824, opened 2019-05-29
More cutting words

teolemon:
More cutting words


VaiTon:
Why are they commented?

PR #1827, opened 2019-05-29
fix tests

stephanegigandet:
fix the tags test that are failing


teolemon:
The PR to fix test that are failing is failing. I'm reviewing it to solve this conendruum.

PR #1830, opened 2019-05-30
merge italian, danish and german updates

teolemon:


lang/it/texts/nutriscore.html
```diff
@@ -1,7 +1,7 @@
 <!-- no side column -->
 
 <styles>
-@media only screen and (max-width: 40em) { h2 { font-weight:bold; font-size:1.2em;} }
+@media solo schermo e (larghezza massima: 40em) {h2 {font-weight: bold; font-size: 1.2em;}}
```
hangy: This shouldn't be translatable and must not be translated.
```suggestion
@media only screen and (max-width: 40em) { h2 { font-weight:bold; font-size:1.2em;} }
```

lang/nl/texts/data.en.html
```diff
@@ -26,59 +26,59 @@ <h3>MongoDB dump</h3>
 <dl>
  <dt>Link</dt>
  <dd><a href="https://static.openfoodfacts.org/data/openfoodfacts-mongodbdump.tar.gz">https://static.openfoodfacts.org/data/openfoodfacts-mongodbdump.tar.gz</a></dd>
- <dt>sha256sum</dt>
+ <dt>sha256som</dt>
```
hangy: This probably should not be translated unless we phrase the source differently.
```suggestion
 <dt>sha256sum</dt>
```

lang/nl/texts/data.en.html
```diff
@@ -87,13 +87,13 @@ <h2>Android and iPhone mobile app</h2>
 <h2>Wrappers and SDKs</h2>
 <p>We probably have a wrapper for your favorite programming language. If we do, you can use it and improve it. If we don't, you can help create it.</p>
 
-<h2>Mailing list for data, API and exports</h2>
+<h2>Mailinglijst voor gegevens, API en export</h2>
 
-<p>You can also join the mailing list to discuss the API and exports, tell us about your re-uses, and be notified when improvements or changes are made to the API. Send an empty e-mail to
-<a href="mailto:api-subscribe@openfoodfacts.org">api-subscribe@openfoodfacts.org</a> to subscribe.</p>
+<p>U kunt ook deelnemen aan de mailinglijst om de API en export te bespreken, ons te vertellen over uw hergebruik en u kunt op de hoogte worden gesteld wanneer er verbeteringen of wijzigingen worden aangebracht in de API. Stuur een lege e-mail naar
+<a href="mailto:api-subscribe@openfoodfacts.org">api-abonnees@openfoodfacts.org</a> om te abonneren.</p>
```
hangy: That mail address probably does not exist? The `mailto` link isn't affected, but copying this address would not work.
```suggestion
<a href="mailto:api-subscribe@openfoodfacts.org">api-subscribe@openfoodfacts.org</a> om te abonneren.</p>
```

lang/it/texts/nutriscore.html
stephanegigandet: I moved the CSS styles out of the text file (the English one, hopefully Crowdin will then remove it from all the other ones) into the scss

lang/nl/texts/data.en.html
stephanegigandet: Right, we need to change that to tell them to come on Slack #api instead.

stephanegigandet:
@hangy @teolemon : we need the new translated nutriscore pages online asap, so I manually copied the de / es / it / da pages live: https://de.openfoodfacts.org/nutriscore

teolemon:
@hangy I did the changes. Can you manually review positively since it seems blocked.

stephanegigandet:
@hangy: any reason why this pull request is blocked? It says there are changes requested, but I don't see how to resolve it.

scss/_off.scss
```diff
@@ -325,3 +325,5 @@ p.details {
   color: gray;
   font-size: 0.8em;
 }
+
+@media only screen and (max-width: 40em) { h2 { font-weight:bold; font-size:1.2em;} }
```
hangy: Did you test if this messes up `h2` tags on other OFF pages?

teolemon: @stephanegigandet

teolemon: @hangy short answer, don't think was tested

stephanegigandet: Well if it should actually fix the h2 tags in other pages. On small screens, they currently do not differ from normal text at all.

PR #1838, opened 2019-05-31
[WIP] Create autocorrect.txt

teolemon:
Small autocorrect file. Might be useful


teolemon:
@stephanegigandet does such a list make sense ?

CloCkWeRX:
That somewhat makes sense, but do we have stats on the most frequent mistakes? Brand names for example having an autocomplete in the web UI/mobile UIs would potentially cut down a lot of minor typos.

stephanegigandet:
I don't really think we should auto-correct OCR errors like that: it's very likely that the ingredients lists contain many other typos, that we won't be able to correct entirely. So it's probably more useful to leave the typos, and then try to remove completely those lists.

PR #1839, opened 2019-05-31
Fix mod_perl not working in Docker

hangy:
- Use fixed image versions to avoid unintentional changes.
- Build apache2-mod-perl from source, because the @testing version in edge is compiled against a newer version than Alpine 3.9.4 (stable).


hangy:
cc @VaiTon

PR #1846, opened 2019-06-01
try to set language for Spanish products uploaded by Yuka, bug #1837

### LINKED ISSUE #1837: Redirect _fr API calls to the right language for kiliweb
Redirect _fr API calls to the right language for kiliweb.

kiliweb always sends images and data values in French. If the country passed is Spain, assign the images to _es instead, and store the product name in _es
Same for Germany and Italy.

This only applies to new products, of course

stephanegigandet:
fixes #1837

### DIFF
--- a/cgi/product_jqm_multilingual.pl
+++ b/cgi/product_jqm_multilingual.pl
@@ -147,13 +147,17 @@ else {
 		$param_cc =~ s/^en://;
 		
 		my %lc_overrides = (
-			es => "es",
-			it => "it",
-			de => "de",
-			uk => "en",
-			gb => "en",
-			pt => "pt",
-			nl => "nl",
+				au => "en",
+				es => "es",
+				it => "it",
+				de => "de",
+				uk => "en",
+				gb => "en",
+				pt => "pt",
+				nl => "nl",
+				us => "en",
+				ie => "en",
+				nz => "en",
 		);
 		
 		if (defined $lc_overrides{$param_cc}) {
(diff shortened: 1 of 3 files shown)


cgi/product_jqm_multilingual.pl
```diff
@@ -138,8 +138,32 @@
...
 	}
 	my @param_langs = keys %param_langs;
+	
+	# 01/06/2019 --> Yuka always sends fr fields even for Spanish products, try to correct it
+
+	if ((defined $User_id) and ($User_id eq 'kiliweb') and (defined param('cc')) and (param('cc') ne "fr")) {
+	
+		my $param_cc = param('cc');
+		
+		my %lc_overrides = (
```
hangy: You could put the whole `lc_override` logic into some `sub` in Products.pm to avoid code duplication in here. 🙂

lib/ProductOpener/Products.pm
```diff
@@ -220,6 +216,21 @@ sub init_product($) {
 	if ($creator eq 'kiliweb') {
 		if (defined param('cc')) {
 			$country = param('cc');
+			
+			# 2019 06 01: also try to change the language according to country
```
hangy: Is there a specific reason why this comment is different to this? https://github.com/openfoodfacts/openfoodfacts-server/blob/5430048b2d138e123098bbaed8b29b660d66859a/cgi/product_jqm_multilingual.pl#L142

Also, product_jqm_multilingual.pl checks for `param('cc') ne "fr"`. This most likely won't make a difference, because no override for `fr` is defined, but the logic might be easier to understand if it's the same for both files.

PR #1849, opened 2019-06-02
make allergens/traces parser more resilient for broken ingredient lis…

stephanegigandet:
…ts, bug #1504

Will fix https://fr.openfoodfacts.org/produit/3256221111774/spaghetti-qualite-superieure-u


teolemon:
lgtm. eventually, we'll need to massively make it multilingual

Issue #1788, opened 2019-05-20
"Yes" doesn't shown in robotoff question (for some resolutions/zoom levels)


CharlesNepote:
Every page where a robotoff question is displayed at the end of the screen, the "yes" button is not shown: 
![image](https://user-images.githubusercontent.com/562352/58013428-e1c77700-7af6-11e9-9dbc-a705f41bb65f.png)

A quick and dirty fix is to change: `width: 80%;` in `:host { background-color: #274477; color: white; position: fixed; bottom: 0; width: 100%; border-top: 1px solid #eee; z-index: 100; padding-top: 10px; }`
(File: https://github.com/openfoodfacts/openfoodfacts-server/blob/a8dc903cb32c15288ff80aef079c92e3ee6f71d5/html/js/display-product.js )

@hangy?

aleene:
I do not even see the no! The width of the window does not seem to be taken into account.

hangy:
Can you give me some input on used OS, browser, and screen size? It worked successfully for me (with Chrome @ 1920x1080) on the weekend.

CharlesNepote:
I use both Firefox 66 (last version) and Chromium Version 73.0.3683.86 (Build officiel) under Linux Ubuntu 18.04 @ screen resolution 1920x1080. The "yes" button *never* appear.

Here is a screenshot under Chromium:
![Capture d’écran_2019-05-20_16-38-00](https://user-images.githubusercontent.com/562352/58029829-b0ae6d00-7b1d-11e9-9aba-2162d89885ce.png)

If I use 90% zoom, the "yes" button appear.

aleene:
This is the same as #1287, I guess. I use Safari on iMac and iPad.

hangy:
> I use both Firefox 66 (last version) and Chromium Version 73.0.3683.86 (Build officiel) under Linux Ubuntu 18.04 @ screen resolution 1920x1080. The "yes" button _never_ appear.
> 
> Here is a screenshot under Chromium:
> ![Capture d’écran_2019-05-20_16-38-00](https://user-images.githubusercontent.com/562352/58029829-b0ae6d00-7b1d-11e9-9aba-2162d89885ce.png)
> 
> If I use 90% zoom, the "yes" button appear.

Do you possibly use some high DPI setting/text scaling in the OS? I'm currently unable to reproduce the error in current Firefox ESR or Chrome stable - independently of the window size.
[...]

PR #1871, opened 2019-06-04
display the robotoff buttons to the left, fixes bug #1788

stephanegigandet:
Put the buttons at the left, remove the product thumbnail.

### DIFF
--- a/html/js/display-product.js
+++ b/html/js/display-product.js
@@ -35,22 +35,22 @@ class RobotoffAsker extends HTMLElement {
         .hidden { display: none; }
       </style>
       <div class="row hidden zoomrow" id="zoomrow">
-        <div class="medium-12 columns">
+        <div class="small-12 columns">
           <img src="" id="zoom" data-zoom-in-src="" data-zoom-out-src="">
         </div>
       </div>
       <div class="row">
-        <div class="medium-2 columns">
+        <div class="small-2 columns hidden">
           <img src="" id="thumbnail">
         </div>
-        <div class="medium-10 large-6 columns">
+        <div class="small-12 columns">
           <span id="question">Question</span>
           <span id="value">Value</span>
         </div>
-        <ul class="medium-12 large-4 columns button-group">
-          <li><a href="#" id="no" class="button alert annotate" data-annotation="0">No</a></li>
-          <li><a href="#" id="skip" class="button secondary annotate" data-annotation="-1">Not sure</a></li>
-          <li><a href="#" id="yes" class="button success annotate" data-annotation="1">Yes</a></li>
+        <ul class="small-12 columns button-group" style="margin-top:0.5rem;">
+          <li><a href="#" id="no" class="small button alert annotate" data-annotation="0">No</a></li>
+          <li><a href="#" id="skip" class="small button secondary annotate" data-annotation="-1">Not sure</a></li>
+          <li><a href="#" id="yes" class="small button success annotate" data-annotation="1">Yes</a></li>
         </ul>
       </div>
       <a href="#" id="close">&times;</button>
--- a/html/js/display-product.js
+++ b/html/js/display-product.js
@@ -40,7 +40,7 @@ class RobotoffAsker extends HTMLElement {
         </div>
       </div>
       <div class="row">
-        <div class="small-2 columns hidden">
+        <div class="small-2 columns">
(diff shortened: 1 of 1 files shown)


teolemon:
👍

html/js/display-product.js
```diff
@@ -35,22 +35,22 @@ class RobotoffAsker extends HTMLElement {
...
       </style>
       <div class="row hidden zoomrow" id="zoomrow">
-        <div class="medium-12 columns">
+        <div class="small-12 columns">
           <img src="" id="zoom" data-zoom-in-src="" data-zoom-out-src="">
         </div>
       </div>
       <div class="row">
-        <div class="medium-2 columns">
+        <div class="small-2 columns hidden">
```
hangy: Wouldn't this *always* hide the thumbnail? I don't see any change removing the hidden class.

hangy: No thumbnail: 
image

html/js/display-product.js
```diff
@@ -35,22 +35,22 @@ class RobotoffAsker extends HTMLElement {
         .hidden { display: none; }
       </style>
       <div class="row hidden zoomrow" id="zoomrow">
-        <div class="medium-12 columns">
+        <div class="small-12 columns">
```
hangy:
```suggestion
        <div class="small-10 columns">
```
Re-enable the thumbnail. 10/12 columns seems to be enough for the text.

html/js/display-product.js
```diff
@@ -35,22 +35,22 @@ class RobotoffAsker extends HTMLElement {
...
           <img src="" id="zoom" data-zoom-in-src="" data-zoom-out-src="">
         </div>
       </div>
       <div class="row">
-        <div class="medium-2 columns">
+        <div class="small-2 columns hidden">
           <img src="" id="thumbnail">
         </div>
-        <div class="medium-10 large-6 columns">
+        <div class="small-12 columns">
```
hangy:
```suggestion
        <div class="small-10 columns">
```
Re-enable the thumbnail. 10/12 columns seems to be enough for the texts.

html/js/display-product.js
```diff
@@ -35,22 +35,22 @@ class RobotoffAsker extends HTMLElement {
...
           <span id="question">Question</span>
           <span id="value">Value</span>
         </div>
-        <ul class="medium-12 large-4 columns button-group">
-          <li><a href="#" id="no" class="button alert annotate" data-annotation="0">No</a></li>
-          <li><a href="#" id="skip" class="button secondary annotate" data-annotation="-1">Not sure</a></li>
-          <li><a href="#" id="yes" class="button success annotate" data-annotation="1">Yes</a></li>
+        <ul class="small-12 columns button-group" style="margin-top:0.5rem;">
```
hangy:
```suggestion
        <ul class="small-10 columns button-group" style="margin-top:0.5rem;">
```
Re-enable the thumbnail. 10/12 columns seems to be enough for the texts.

html/js/display-product.js
hangy:
```suggestion
        <div class="small-2 columns">
```

stephanegigandet: In fact I do want to always hide the thumbnail, I don't see the point of showing a thumbnail here. I added the hidden class in case we want to put it back at some time for some reason.

image

html/js/display-product.js
stephanegigandet: I prefer to put the text and buttons at the very left. There is the same 100% width on a position:fixed element that makes the actual available space much less than 10/12.

e.g. it can look like this with the small-2 + small-10 columns:

image

html/js/display-product.js
stephanegigandet: I forgot about the zoom when clicking on the thumbnail, I guess the thumbnail can be useful in some cases, if the selected picture is different.

hangy: Yes, the thumbnail was added to enable users to see the source of Robotoff's suggestion. 🙂 Even the small image can be too small to identify the text, which is why I added the optional (rudimentary) zoom levels.

stephanegigandet:
Sounds good, thanks @hangy!

stephanegigandet:
@hangy : I still see "Merging can be performed automatically once the requested changes are addressed.", is that something I can do (if so how?) or that you have to do?

PR #1883, opened 2019-06-06
Nutrition OCR on nutrition facts images

stephanegigandet:
This is to run OCR of nutrition facts images, in order to try later to extract the nutrition facts from them.

### DIFF
--- a/cgi/nutrition.pl
+++ b/cgi/nutrition.pl
@@ -1,0 +1,85 @@
+#!/usr/bin/perl -w
+
+# This file is part of Product Opener.
+# 
+# Product Opener
+# Copyright (C) 2011-2018 Association Open Food Facts
+# Contact: contact@openfoodfacts.org
+# Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
+# 
+# Product Opener is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+# 
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+# 
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use Modern::Perl '2012';
+use utf8;
+
+use CGI::Carp qw(fatalsToBrowser);
+
+use ProductOpener::Config qw/:all/;
+use ProductOpener::Store qw/:all/;
+use ProductOpener::Index qw/:all/;
+use ProductOpener::Display qw/:all/;
+use ProductOpener::Tags qw/:all/;
+use ProductOpener::Users qw/:all/;
+use ProductOpener::Images qw/:all/;
+use ProductOpener::Products qw/:all/;
+use ProductOpener::Food qw/:all/;
+use ProductOpener::Images qw/:all/;
+
+use CGI qw/:cgi :form escapeHTML/;
+use URI::Escape::XS;
+use Storable qw/dclone/;
+use Encode;
+use JSON::PP;
+use Log::Any qw($log);
+
+ProductOpener::Display::init();
+
+my $code = normalize_code(param('code'));
+my $id = param('id');
+my $ocr_engine = param('ocr_engine');
+my $annotations = param('annotations') | 0;
+
+if (not defined $ocr_engine) {
+	$ocr_engine = "tesseract";
+	# $ocr_engine = "google_cloud_vision";
+}
+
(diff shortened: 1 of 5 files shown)


teolemon:
This is very important and needs to be run ASAP. 
Also @stephanegigandet can we run it as well on the ingredient lists, so that we can have a consistency check before tuesday ?

lib/ProductOpener/Images.pm
```diff
@@ -1349,4 +1349,132 @@ HTML
...
+			}
+		}
+		else {
+			$log->warn("no available tesseract dictionary", { lc => $lc, lan => $lan, id => $id }) if $log->is_warn();
+		}
+
+	}
+	elsif ($ocr_engine eq 'google_cloud_vision') {
+
+		my $url = "https://alpha-vision.googleapis.com/v1/images:annotate?key=" . $ProductOpener::Config::google_cloud_vision_api_key;
```
hangy: Why `alpha-vision`? The official docs don't mention that subdomain: https://cloud.google.com/vision/docs/reference/rest/

lib/ProductOpener/Images.pm
```diff
@@ -1349,4 +1349,132 @@ HTML
...
+
+			if ((defined $cloudvision_ref->{responses}) and (defined $cloudvision_ref->{responses}[0])
+				and (defined $cloudvision_ref->{responses}[0]{fullTextAnnotation})
+				and (defined $cloudvision_ref->{responses}[0]{fullTextAnnotation}{text})) {
+
+				$log->debug("text found in google cloud vision response") if $log->is_debug();
+
+
+				$product_ref->{$field} = $cloudvision_ref->{responses}[0]{fullTextAnnotation}{text};
+				$product_ref->{$field . "_annotations"} = $cloudvision_ref;
```
hangy: This saves the complete cloudvision response in the `product.sto`, right?

1. This could be a lot of data to be stored in the database and to be returned in our API (large API responses)?
2. By saving it in our database, we would basically publish the cloudvision data under the ODbL, and mixing ODbL with other sources is weird. Do the terms of service of the Cloud Vision API allow this?

cgi/nutrition.pl
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/perl -w
```
hangy: Where is `cgi/nutrition.pl` called?

lib/ProductOpener/Images.pm
```diff
@@ -1349,4 +1349,132 @@ HTML
...
+
+	if ($id =~ /_(\w\w)$/) {
+		$lc = $1;
+	}
+
+	my $size = 'full';
+	if ((defined $product_ref->{images}) and (defined $product_ref->{images}{$id})
+		and (defined $product_ref->{images}{$id}{sizes}) and (defined $product_ref->{images}{$id}{sizes}{$size})) {
+		$filename = $id . '.' . $product_ref->{images}{$id}{rev} ;
+	}
```
hangy: Shouldn't we have an `else` clause for the case that `$filename` cannot be set? The following code that uses `$image` or `$image_url` would be useless in that case, because the URLs would only point to non-existing images.

lib/ProductOpener/Images.pm
teolemon: Nope, it saves on disk.

lib/ProductOpener/Images.pm
teolemon: We're using the alpha endpoint because I have early and free access. Behaves like the regular endpoint, with more features.

hangy: Gotcha!

lib/ProductOpener/Images.pm
hangy: > Nope, it saves on disk.

Not necessarily. If `extract_text_from_image` was called with a product reference and that reference was then stored on disk as usual, it would (at some point) be stored in MongoDB, too. If that's not intentional, then it would be weird to use the `$product_ref` only as a container for the resulting data, because that's not how the reference is used throughout ProductOpener.

stephanegigandet: It's only called from ingredients.pl, which does not save the product. I did not intend the annotations to be saved in product.sto or mongodb. It won't do it at this point, but we could pass an extra parameter for the resulting data (both the text and the annotations).

lib/ProductOpener/Images.pm
stephanegigandet: Agreed, we should bail out.

cgi/nutrition.pl
stephanegigandet: It's to be used in things like robotoff or later in the product edit form, it's not called yet.

lib/ProductOpener/Images.pm
stephanegigandet: Now using a separate $results_ref structure for results.

lib/ProductOpener/Images.pm
stephanegigandet: Added else clause.

PR #1888, opened 2019-06-07
fix manifest.cgi that did not compile

stephanegigandet:
replaced ( ) by { } to get a reference to the manifest options


hangy:
Interesting ... is there a difference in compilation with `perl -c` and the one that's used when the script is run in Apache?

https://github.com/openfoodfacts/openfoodfacts-server/blob/2631ff3719e7f0799a91472892cf263918bb753b/compile_cgi.js#L15
https://travis-ci.org/openfoodfacts/openfoodfacts-server/builds/542436479#L2733

stephanegigandet:
@hangy : I said a mistake, the script does compile, it's at runtime that we expect a reference, so manifest.pl crashes.

PR #1891, opened 2019-06-07
Issue 1887/ingredient preparsing

stephanegigandet:
I'm making the ingredient parsing more generic, with a preparsing function that can be applied for both ingredients extraction and additives extraction (previously it was only used for additives extraction).

There are still a lot of improvements that can be made to make it more generic, to use the taxonomies etc. instead of hardcoded values. But it is an improvement, and this is something very sensitive so we should do it step by step.

I'm adding tests in the process to make sure we don't create regressions.

Regarding the hardcoded strings: at this point, I'm really not confortable to use strings from the po files or the taxonomies, at every change should be tested, with specific test cases added etc.


lib/ProductOpener/Ingredients.pm
```diff
@@ -389,9 +395,13 @@ sub extract_ingredients_from_text($) {
 	return if not defined $product_ref->{ingredients_text};
 
 	my $text = $product_ref->{ingredients_text};
-	$text =~ s/\&quot;/"/g;
+	my $lc = $product_ref->{lc};
```
hangy: This hides `$ProductOpener::Lang::lc`, doesn't it? Could be a bit confusing.

lib/ProductOpener/Ingredients.pm
```diff
@@ -677,7 +685,57 @@ sub normalize_fr_a_de_enumeration {
...
+
+sub normalize_enumeration($$$) {
+
+	my $lc = shift;
+	my $type = shift;
+	my $enumeration = shift;
+	
+	$log->debug("normalize_enumeration", { type => $type, enumeration => $enumeration }) if $log->is_debug();
+	
+	my $separators = '\(|\)|\/| \/ | - |, |,|'  . $Lang{_and_}{$lc};
```
hangy: Could it be useful to use `$cbrackets, `$dashes`, `$commas`, or `$stops` here? We already have a `$separators` declared at https://github.com/openfoodfacts/openfoodfacts-server/blob/8732b1f90de8bb21ef65d279fb162f88f8845cfc/lib/ProductOpener/Ingredients.pm#L94, which is hidden by this local variable.

t/ingredients2.t
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/perl -w
```
hangy: Why is this t/ingredients2.t? The "2" isn't very descriptive. 🙂

stephanegigandet: It's exactly like ingredients.t, just more test cases. I split it to make it easier / faster to test specific tests or add new test cases.

lib/ProductOpener/Ingredients.pm
stephanegigandet: I removed that one. There are other instances of $lc hiding in Tags.pm especially, but then I think the right thing will be to rename $ProductOpener::Lang::lc to $ProductOpener::Lang::site_lc 
It's a much bigger change though, let's do it separately.

lib/ProductOpener/Ingredients.pm
stephanegigandet: I changed to use the $dashes etc.

t/ingredients2.t
stephanegigandet: Renamed to ingredients_tags.t , it's a bit different from ingredients.t in fact, as it just looks at the ingredients_original_tags field.

PR #1913, opened 2019-06-10
Fix for #1889

### LINKED ISSUE #1889: "Nutrition facts are not specified on the product." is hard to see
Sometimes "Nutrition facts are not specified on the product." ([example](https://world.openfoodfacts.org/product/8423018110172/sesame-dore-hennedrog))

It's an important information. It could be highlighted a little bit. At least this information should have an id and or a class to let designers easily highlight it.

I suggest:
- `<p id="nutrition-facts-ns">Nutrition facts are not specified on the product.</p>`
- instead of `<p id="nutrition-facts-ns">Nutrition facts are not specified on the product.</p>`

That way we will be able to make some tests with the Open Food Facts power user extension script.

VaiTon:
Fixes #1889

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -7250,9 +7250,8 @@ HTML
 	if ((defined $product_ref->{no_nutrition_data}) and ($product_ref->{no_nutrition_data} eq 'on')) {
 		$html .= "<div class='panel callout'>$Lang{no_nutrition_data}{$lang}</div>";
 	}
-	else {
-		$html .= display_nutrition_table($product_ref, \@comparisons);
-	}
+
+	$html .= display_nutrition_table($product_ref, \@comparisons);
 
 	$html .= <<HTML
 </div>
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -7248,11 +7248,11 @@ HTML
 
 
 	if ((defined $product_ref->{no_nutrition_data}) and ($product_ref->{no_nutrition_data} eq 'on')) {
-		$html .= "<p>$Lang{no_nutrition_data}{$lang}</p>";
+		$html .= "<div class='panel callout'>$Lang{no_nutrition_data}{$lang}</div>";
+	}
+	else {
```
stephanegigandet: It's still useful to display the nutrition table even if the product has no nutrition data, as it is possible to see the average nutrition values for products in the same category.

VaiTon: Aren't average values shown as a percentage?

hangy: > Aren't average values shown as a percentage?

You can do both https://world.openfoodfacts.org/product/5450168551800/raibi-saveur-grenade-luxlait
image

stephanegigandet: Even if a product has no nutrition data, we show the average:

image

CharlesNepote: Also, as this information make sense I think it's important to use an id in the HTML code, and not only CSS classes. That way, for example, I can check with my Open Food Facts power user extension if a product hasn't nutrition data on its cover.

Issue #1915, opened 2019-06-10
API returns a CO2 level when no computation is done


teolemon:
https://world.openfoodfacts.org/api/v0/product/20211394.json?fields=environment_impact_level_tags

https://world.openfoodfacts.org/product/20211394/lasagne-alla-bolognese-italiamo
We get a CO2 value where there should be none

VaiTon:
@stephanegigandet is this fixed?

teolemon:
fixed @VaiTon @stephanegigandet

PR #1916, opened 2019-06-11
remove old invalid co2 computations

stephanegigandet:
fixes bug #1915

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -4549,6 +4549,11 @@ sub compute_carbon_footprint_infocard($) {
 	
 	if (not ((has_tag($product_ref, "countries", "en:france")) and (defined $product_ref->{ingredients}) 
 		and (length($product_ref->{ingredients}) > 5))) {
+		delete $product_ref->{environment_impact_level};
+		delete $product_ref->{environment_impact_level_tags};
+		delete $product_ref->{environment_infocard};
+		delete $product_ref->{environment_infocard_en};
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Food.pm
```diff
@@ -4547,16 +4547,21 @@ sub compute_carbon_footprint_infocard($) {
 	
 	# Limit to France, as the carbon values from ADEME are intended for France
```
VaiTon: Could you add a comment explaining why are we removing those?

lib/ProductOpener/Ingredients.pm
```diff
@@ -154,6 +154,7 @@ sub compute_carbon_footprint_from_ingredients($) {
 		delete $product_ref->{nutriments}{"carbon-footprint-from-meat-or-fish_product"};
 	}
```
VaiTon: Could you add a comment here too?

lib/ProductOpener/Food.pm
stephanegigandet: This function computes those values, and at the start of the function we remove previous values if they exist. New values are computed below.

lib/ProductOpener/Ingredients.pm
stephanegigandet: Same thing, we delete old values before computing new ones.

lib/ProductOpener/Food.pm
VaiTon: @stephanegigandet I mean in the source code :-)

PR #1924, opened 2019-06-14
Ingredients patch 24

aleene:
More translations


stephanegigandet:
I think the tests will pass when we can merge this in: https://github.com/openfoodfacts/openfoodfacts-server/commit/3a6607fe34f3f313d68a751d65f087852676479c

PR #1931, opened 2019-06-16
palm oil repair

aleene:
Repaired the palm oil entry for better detection.


aleene:
I'll start on the vegan/vegetarian next.

taxonomies/ingredients.txt
```diff
@@ -20410,19 +20407,37 @@ it:oli vegetali non idrogenati
 fr:huile végétale hydrogénée
 de:pflanzliches Öl gehärtet
 
+<<<<<<< HEAD
```
hangy: Can you fix this merge conflict, please?

stephanegigandet:
@aleene : for palm oil, we are going to use the from_palm:en:yes property instead of making en:palm a parent of everything.

There shouldn't be an ingredient "palm" in fact, we can keep it for now to make sure that we get all formulations of "some form of oil (palm)". So let's keep it under <en:palm oil

aleene:
I moved **palm** as a parent of **palm oil**, as I did not want to make any implicit assumptions. **palm** might be **palm oil** (and probably is), but you never know. I wanted to take the conservative approach.

teolemon:
``` 
#   Failed test at t/ingredients.t line 338.
#          $got->{ingredients_tags}[3] = 'en:palm'
#     $expected->{ingredients_tags}[3] = 'en:palm-kernel-oil-and-fat'
```

stephanegigandet:
Can we close this pull request? We have another solution working for palm oil, and we should not have "XYZ" as a parent of "XYZ oil".

aleene:
I guess we can close it.

PR #1941, opened 2019-06-17
Send an actual 404 status if a product does not exist

### LINKED ISSUE #976: Read API status code
Reported by @pilu on Slack:

> READ APis are always returning a 200 status code, even when the product is not found. is there a plan to upgrade the api to support the right HTTP statuses?

hangy:
**Description:** Instead of sending 200, actually send 404, so that the status can be parsed using common RESTful practices. Same status trick as in #1815.
**Related issues and discussion:** Fixes #976


stephanegigandet:
Looks good to me, we just need to check that it's not going to break one of the apps. cc @teolemon

Issue #1943, opened 2019-06-18
New popularity facet with top 10 / 100 / 1000 etc. popular products based on unique scans


stephanegigandet:
In order to be able to quickly review the quality / compute stats etc. for popular products, we are going to add a new popularity facet.

The facet will initially have values related to the number of scans by unique ips made with the OFF official app:

- top-10-scans-2019 (same for top 50, 100, 500, 1000, 5000 ... 100 000)

We could possibly use the same facet for other measures of popularity. (e.g. scans by other apps, sales figures from some stores, stock data from food banks, searches and views on the OFF web site etc.)

When we have a proper scans database (as described in #1501), we will populate the /popularity facet through the scans database.

In the short term, we will use the scanbot.pl log processing script to populate the facet.

teolemon:
quick and 80% efficient for 20% effort 👍

teolemon:
outside the scope, but can you add another facet: 
`countryscan/scanned-3-times-in-{country}`

CharlesNepote:
@teolemon 3 times is few isn't it? I would say 5, 7 or 10 times. And you must precise X times by different people/IP on a different day, isn't it?

@stephane is it possible to add:
- /scans/never # products never scaned except the day they where entered in the database; will allow to identify false products, wrong barcode recongintion, etc.
- /scans/never-since-1-year # (or 2 years?) will allow to identify products that are not sold anymore
- /scans/less-than-7 # will allow see products that could be false or not sold anymore
- /scans/less-than-7-since-1-year
- /scans/less-than-0-dot-01-percent (or a thing like that) # will allow to answer to "trolls" based on rarely scanned products
 
We should discuss a little bit the interface, but if you prefer to go quick, we can discuss later.

The good thing should be to document **before** coding ;-)

stephanegigandet:
At this point, we have 250k scans from 2019 (only scans from the new Android app), out of 900k products. So there will be 650k products without scan data at all. So the "never" scans values may not be very useful.

We have some scan data from previous years, but it's not directly comparable (it contains all API calls, including bots, searches etc.). We could load it to populate different values (e.g. top-1000-api-2018), but it would need cleaning to make sure we remove bots etc.

This bug is a quick and dirty fix to get a useful popularity facet, it is not a replacement for the scan database that will make many more metrics etc. possible.

teolemon:
[...]

PR #1946, opened 2019-06-18
log processing to populate a new popularity facet, bug #1943

stephanegigandet:
This is a new popularity facet based on scan logs, to be used in particular for quality review.

Details are on bug #1943

It is not perfect, but it does the job. When we have a scans database, we can make a better system.

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -3427,3 +3427,12 @@ msgstr ""
 msgctxt "button_caption_skip"
 msgid "Skip"
 msgstr ""
+
+msgctxt "popularity_s"
+msgid "popularity"
+msgstr ""
+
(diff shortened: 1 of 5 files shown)


stephanegigandet:
The results can be seen on the dev server: https://world.dev.openfoodfacts.org/popularity

teolemon:
Seems to work well in dev.

teolemon:
I've noticed edits by scanbot in dev, but couldn't access them because the login is broken for me. What are they ?
Is the scan count updated to rank products by popularity in advanced search ?

stephanegigandet:
The edits are added countries, when we have at least 3 ips from one country that have scanned a product.

PR #1947, opened 2019-06-18
Minerals cleanup

aleene:
All minerals have been moved from ingredients.txt to minerals.txt


teolemon:
``` 

Waiting for spawned processed to exit.
[additives_classes] child process exited with code 0
[minerals] child process exited with code 255
[allergens] child process exited with code 0
1 processes still running
[ingredients] child process exited with code 0
Done!
The command "node refresh_taxonomies.js" failed and exited with 255 during .

```

stephanegigandet:
The tests fail because all minerals now have a "en:mineral" parent. We will need to update the tests. (and make sure that it doesn't break something for mineral / additives detection)

aleene:
"mineral" and "minerals" also appear as words in ingredients.

PR #1948, opened 2019-06-18
Import scamark (Leclerc) + mousquetaires (Intermarché)

stephanegigandet:
Import scripts for the French stores chains Leclerc and Intermarché.


teolemon:
I've started looking at Intermarché products. Can't spot Scamark ones yet.

Issue #1734, opened 2019-04-06
Category autocompletion should provide results *containing* the search term


CharlesNepote:
The category autocompletion shows categories **beginning** by the term entered.

Category autocompletion should provide results **containing** the term entered. The Android app works like that.

When you start entering "poulet" you should have:
- Brochettes de poulet
- Cuisses de poulet
- Poulets au curry
- Curry au poulet
- etc.

Example: I wasn't able to find a better category than "Poulet" for https://fr.openfoodfacts.org/produit/3095759147011/des-de-poulet-halal-fleury-michon

**Howto**
When entering "po" in "Categories" field, the [following script](https://github.com/openfoodfacts/openfoodfacts-server/blob/master/cgi/suggest.pl) is called: https://world.openfoodfacts.org/cgi/suggest.pl?lc=en&tagtype=categories&&term=po

Is it possible to change the behavior of this script?

**Stopwords**
When searching for "de" in french we find more than 300 categories (ex. "Blanc de poulet"). Category searching should exclude stopwords depending on the language.

**Search start**
Starting the search with a single letter is not relevant because it provides two many results. Searching should start, at least, with two letters.

**Sort**
Results should be sorted in alphabetical order. BUT results might begin with results "starting with" the search term. Example for "tom":
```
Tomate jams
Tomato juices
Tomato mustards
Tomato pastes
Tomato purées
Tomato sauces
Tomato soups
Tomato vinegars
Tomatoes
Tomatoes and tomato products
Tomme
Baked beans in tomato sauce
Canned tomatoe
Cherry tomatoes
Concentrated tomato juices
Cream of tomato soups
Dried tomato powder
Dried tomatoes
Dried tomatoes in oil
Fried tomato sauces
Mackerels fillets with tomato and basilisk
Natural dried tomatoes
Natural grated tomato
Peeled tomatoes
Pumpkin tomatoes
Rice stuffed tomatoes
Sardines in tomato sauce
Strained tomatoes
Stuffed tomatoes
Tabbouleh with tomatoes
```

VaiTon:
Not sure about the stopwords, they would be too many don't you think? What about making the search only start with three or more letters?

VaiTon:
Fixed with #1956

PR #1956, opened 2019-06-20
Partial fix for issue 1734

VaiTon:
Make suggest.pl return strings that contains the word.

Related to #1734

### DIFF
--- a/cgi/suggest.pl
+++ b/cgi/suggest.pl
@@ -87,8 +87,12 @@ else {
 		if (not ($search_lc eq $original_lc)) {
 			$tag = $search_lc . ":" . $tag;
 		}
-
-		push @suggestions, $tag;
+		if ($tag =~ /^$stringid/) {
+			push @suggestions, $tag;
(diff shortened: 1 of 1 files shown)


cgi/suggest.pl
```diff
@@ -82,13 +82,17 @@
...
 		if (not ($search_lc eq $original_lc)) {
 			$tag = $search_lc . ":" . $tag;
 		}
-
-		push @suggestions, $tag;
+		if ($tag =~ /^$stringid/) {
+			push @suggestions, $tag;
+		}
+		else {
+			unshift @suggestions, $tag;
```
stephanegigandet: Isn't that going to put suggestions that do not begin with the string first, in reverse alphabetical order? It may be easier to have 2 arrays and to concatenate them after.

VaiTon: Yeah that's right. I tried to keep just one array for performance but guess I'll implement your solution tomorrow.

PR #1958, opened 2019-06-20
Carbon footprint

stephanegigandet:
A few small changes to @RonanMorgan 's implementation of carbon footprint computation from all known ingredients found in FoodGES, in addition to meat and fish.


RonanMorgan:
no comment :)

PR #1959, opened 2019-06-20
Replace standard icon fonts with custom build

hangy:
We currently use Foundation Icons, which hasn't been updated since 2013. Additionally, we also use the whole FontAwesome brands font to use just two icons. In total, this is >100 KiB, of which most is wasted.

This PR replaces both libraries with a build process in Gulp, which builds a custom font with icons from https://icofont.com with the few icons that we actually use.

I also replaces several calls to `format_subdomain('static')` with the `$static_subdomain` variable, and updated the `gulpfile.js` to match the more recent guidelines on https://gulpjs.com.


hangy:
> Looks good to me.

Thanks for the quick review!

> I assume the new icons have a very similar look to the old icons?

Yes, I tried to keep it very similar. Obviously, this also gives us the opportunity of easily switching out icons for better alternatives. 🙂

stephanegigandet:
I installed it on https://world.dev.openfoodfacts.org , I finally found why we didn't get the right headers to allow the font to be loaded.

I don't like the new icons very much though. @teolemon suggested the Material icons which look better (it's all a matter of personal taste of course): http://google.github.io/material-design-icons/

One option is to use the google font directly, or we could build a custom font with just the icons we need. The whole font is 59.4 Kb. I'm not sure that the extra complexity of generating a custom font is worth it...

stephanegigandet:
There is also a slight visual glitch in the top bar that I have not been able to remove:

image

stephanegigandet:
The height is 49px instead of 47px

VaiTon:
> 
> 
> I installed it on https://world.dev.openfoodfacts.org , I finally found why we didn't get the right headers to allow the font to be loaded.
> 
> I don't like the new icons very much though. @teolemon suggested the Material icons which look better (it's all a matter of personal taste of course): http://google.github.io/material-design-icons/
> 
> One option is to use the google font directly, or we could build a custom font with just the icons we need. The whole font is 59.4 Kb. I'm not sure that the extra complexity of generating a custom font is worth it...

I would prefer the entire material font

PR #1962, opened 2019-06-20
Migrate lazy-loading of images to HTML5 `loading` attribute.

### LINKED ISSUE #1198: Consider lazy loading images

### LINKED ISSUE #1853: Bottom pagination not available in some cases
image
Bottom pagination not available in some cases. It looks like the same bug as the "Save" button bug ?

hangy:
**Description:** Swapped out IOLazy in favour of the new `loading="lazy"` image attribute.
**Related issues and discussion:** Fixes #1198, and probably also fixes #1853

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -50,8 +50,7 @@ function js() {
       './node_modules/papaparse/papaparse.min.js',
       './node_modules/osmtogeojson/osmtogeojson.js',
       './node_modules/leaflet/dist/**/*.*',
-      './node_modules/leaflet.markercluster/dist/**/*.*',
-      './node_modules/iolazyload/dist/js/iolazy.min.js'
+      './node_modules/leaflet.markercluster/dist/**/*.*'
     ], { sourcemaps: true })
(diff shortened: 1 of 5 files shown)


VaiTon:
Are we sure that we want every image to be lazy loaded? Couldn't we just maintain the "lazyload" parameter in the sub and if true lazyload the image?

hangy:
> Are we sure that we want every image to be lazy loaded? Couldn't we just maintain the "lazyload" parameter in the sub and if true lazyload the image?

Sure, but the only places that lazyloading is not used right now, is for the tooltips on the search results for the scatter plot and the maps. As far as I remember, this was added, because of some incompatibility with the JavaScript used in there. Therefore, I see no harm in using `loading="lazy"` in these cases.

hangy:
Will set ready for review when Chrome 76 is release with https://www.chromestatus.com/feature/5645767347798016 enabled.

PR #1590, opened 2018-12-25
Poland packaging

simonghrt:
**Description:**

This PR contains the packager codes from Poland in a CSV file and the script that helped me to retrieve it. 
This PR is for the carbon footprint project.
Let me know if you see any change that could be made.

PR #1988, opened 2019-06-24
Poland packaging

hangy:
#1590
> **Description:**
> 
> This PR contains the packager codes from Poland in a CSV file and the script that helped me to retrieve it.
> This PR is for the carbon footprint project.
> Let me know if you see any change that could be made.

With simonghrt/openfoodfacts-server#1 merged.
> Updates the import script and data structure, so that the Polish codes can be imported correctly.


stephanegigandet:
Thank you!

Issue #1991, opened 2019-06-24
Auto rotate ingredient and nutrition images


stephanegigandet:
Yuka and other apps are sending a lot of images that are badly rotated.
This feature is to run all selected ingredients and nutrition images through Google Cloud Vision, and then to analyze the text blocks coordinates to compute the orientation of each image.

We can then autorotate the images that have an incorrect orientation.

It may be too slow to do this in real time when images are uploaded, but we can do it in batch mode on already uploaded and selected images.

teolemon:
@raphael0202 has been working on that.

teolemon:
Google announced this morning that it will be in their API by the end of August

teolemon:
https://github.com/openfoodfacts/openfoodfacts-ai/issues/31

stephanegigandet:
Deployed, applied on the products edited by kiliweb.

PR #1992, opened 2019-06-24
compute image orientation, autorotate, bug #1991

stephanegigandet:
Initial implementation to:
1. run ingredients and nutrition images through google cloud vision
2. compute the image orientation
3. autorotate the images
in batch mode, through update_all_products.pl


teolemon:
Perhaps unselect ingredient images where we have nothing but nutrition hints

stephanegigandet:
@teolemon "Perhaps unselect French ingredient images where we detect 0 ingredients (and have nutrition hints)" --> We can try that separately, it's not trivial, and depending on the confidence we can achieve, it may be better to do in robotoff questions.

PR #2008, opened 2019-06-25
import scripts, biscuiterie Sainte-Victoire

stephanegigandet:


stephanegigandet:
@hangy I was thinking that one thing that would be nice to do is to add tests for each of the conversion scripts, maybe with a sample of products (e.g. 100 products max), and to compare the resulting csv output.

stephanegigandet:
Filed https://github.com/openfoodfacts/openfoodfacts-server/issues/2014

Issue #1973, opened 2019-06-22
In edit mode, provide a link to get back to product view without saving


CharlesNepote:
When we edit a product, we don't know how to get back to view mode. It's possible with the "back" button of the browser but not everybody knows that.

To get back, some people save without any modification (which save an empty entry in the changelog).

**Suggestions**
- First of all, we shouldn't save a modification without changes.
- We could have something in the interface to let the user get back to the product view; ideally, if the user made some changes, this link would prevent to leave the page without saving

VaiTon:
The back button has been implemented however it doesn't check for any modifications before returning to the product page.

stephanegigandet:
This has been fixed by @VaiTon and deployed, closing.

PR #2018, opened 2019-06-27
Added back button and made save button match bar height

VaiTon:
Partial fix for issue #1973


cgi/product_multilingual.pl
```diff
@@ -2152,14 +2152,17 @@ ($$$$$$)
 ;
 
 	if ($type eq 'edit') {
-		$html .= <<HTML
+		$html .= <<"HTML"
```
hangy: What do the quotes around the Heredoc cause?

stephanegigandet: Why the "HTML" instead of HTML?

cgi/product_multilingual.pl
```diff
@@ -2152,14 +2152,17 @@ ($$$$$$)
...
 <div class="row">
-<div class="small-12 medium-12 large-8 xlarge-10 columns">
-<input id="comment" name="comment" placeholder="$Lang{edit_comment}{$lang}" value="" type="text" class="text">
-</div>
-<div class="small-12 medium-12 large-4 xlarge-2 columns">
-<input type="submit" name=".submit" value="$Lang{save}{$lc}" class="button small">
-</div>
+	<div class="small-12 medium-12 large-8 xlarge-8 columns">
+		<input id="comment" name="comment" placeholder="$Lang{edit_comment}{$lang}" value="" type="text" class="text" />
```
hangy: There's actually no need to close the `<input>` element with HTML5. 🙂

stephanegigandet:
I'd prefer "Cancel" instead of "Back".

stephanegigandet:
Could you post a screenshot? Have you tried resizing the window to different sizes to make sure that we can always see the 2 buttons? Thank you!

One thing we could add is the "check" and "cancel" icons from Material, I have included them in the custom font. : <i class="icon-check"></i> should add it.

cgi/product_multilingual.pl
VaiTon: Because without quotes it is less clear if the heredoc should be interpolated or not.
```perl
my $message = <<END_MESSAGE;
...
END_MESSAGE
```
>It works and it behaves the same as if you put double-quotes around the END_MESSAGE like in the next example, but it is less clear. I'd recommend you do not use here-document without quotes around the definition of the end-string. 
~ https://perlmaven.com/here-documents

VaiTon: See https://github.com/openfoodfacts/openfoodfacts-server/pull/2018#discussion_r298182422

VaiTon:
> 
> 
> Could you post a screenshot? Have you tried resizing the window to different sizes to make sure that we can always see the 2 buttons? Thank you!
> 
> One thing we could add is the "check" and "cancel" icons from Material, I have included them in the custom font. :
> 
> ```
> <i class="icon-check"></i> should add it.
> ```

Yes, I've tried and it works :D
immagine
immagine
immagine

I'm going to add the material icons and post some screenshots.

cgi/product_multilingual.pl
stephanegigandet: Interesting, I have never though of that.

VaiTon:
immagine
immagine
@stephanegigandet

Issue #2028, opened 2019-06-29
Compute and display stats of known / unknown tags in each taxonomy and how many products are associated with them


stephanegigandet:
I'm planning to compute stats when we display lists of tags like https://fr.dev.openfoodfacts.org/editeur/scamark/ingredients `

Those queries already have the data we need: all the different tags, whether they exist in the taxonomy or not, and how many products have them.

		my %stats = (
			all_tags => 0,
			all_tags_products => 0,
			known_tags => 0,
			known_tags_products => 0,
			unknown_tags => 0,
			unknown_tags_products => 0,
		);

Then we can have a special ?stats=1 parameter to display the stats (and ?only_stats=1 ) to display just the stats and not the huge data table with all the entries.

VaiTon:
Fixed by #2029

PR #2029, opened 2019-06-29
new stats feature for tags, bug #2028

stephanegigandet:
This is a new feature to compute and display stats for tags: how many unique ones exist, and how many occurences they represent, for known tags and unknown tags (not recognized in the taxonomy).

https://fr.dev.openfoodfacts.org/editeur/scamark/ingredients&stats=1

This feature is very useful for ingredients, but it's available for all types of tags. And it can be used on subsets of products.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -435,7 +435,7 @@ sub analyze_request($)
 
 	# first check parameters in the query string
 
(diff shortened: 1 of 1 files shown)


stephanegigandet:
@hangy : at this point it is a progress tracking feature for developers, contributors working on the taxonomies etc. We can certainly make it better, support a json output etc. but the priority is much less than having the feature in place to get the stats. That's why it's also not translated etc.

stephanegigandet:
@hangy: now checking that ?stats= is defined and positive.

Issue #1912, opened 2019-06-10
Create UUID contribution pages


teolemon:
Create contribution pages for each anonymous UUID, so that we can easily track non-quality on a per-user basis.

/uuid/345675GDFESD4678546453237487RFXB

VaiTon:
And how do we create that UUID?
From the IP?
From the session?

hangy:
Isn't it just some text content of the comment provided by a few apps?

teolemon:
yes it is @hangy

teolemon:
Normally, all major apps now send this uuid, including the official one

stephanegigandet:
We discussed this with @teolemon, a simpler alternative is to use the uuid as the user name when we compute the edit history and the fields editors_tags etc.

stephanegigandet:
@teolemon : which apps send a UUID besides Yuka, and in which format?

stephanegigandet:
Not very pretty, for the display I'm just going to put "yuka - user".

![image](https://user-images.githubusercontent.com/8158668/59964742-115a0c80-9505-11e9-870a-6a8da6e355ec.png)

stephanegigandet:
That looks better I think. The actual UUID is in the URL.

![image](https://user-images.githubusercontent.com/8158668/59966175-a9adbc80-9518-11e9-8a42-874057e240e0.png)

teolemon:
This is solved. Thanks @stephanegigandet

PR #2035, opened 2019-06-30
add elcoco, ethic-advisor + new Apps values in data sources, bug #1912

stephanegigandet:
+ new Apps and App - nameofapp values for data_sources facet


teolemon:
I can't see the regex for ethic adivisor

stephanegigandet:
It has user in it already so it should be caught by the catchall, I added a specific regexp just in case.

Issue #2037, opened 2019-06-30
Quality facet - "quantity-contains-e" has many false positives


CharlesNepote:
### What
See: https://world.openfoodfacts.org/quality/quantity-contains-e

Example: "1 litre" in https://world.openfoodfacts.org/product/3464660002434/jus-d-aloe-vera-pur-aloe

The regexp recognize too much things: https://github.com/openfoodfacts/openfoodfacts-server/blob/aa25a7eb248399d5d61e6e90a221f74d35ff32ce/lib/ProductOpener/SiteQuality_off.pm#L747

I replaced:
- `/(?:.*e$)|(?:[0-9]+\s*[kmc]?[gl]?\s*e)/i` by
- `/(?:^e\s)|(?:\se[^a-z])|(?:\se$)/i`

And it deleted more than 1500 false positives (see file attached).

[4.txt](https://github.com/openfoodfacts/openfoodfacts-server/files/3343075/4.txt)

### Part of
- #5538

stephanegigandet:
I agree that the regexp is broken, but the proposed change also delete true positives.

What we really want to identify is cases like this:

500 ge

VaiTon:
In the first place, why do we want to identify quantities containing "e"? @stephanegigandet

benbenben2:
See: (https://world.openfoodfacts.org/quality/quantity-contains-e) -> https://world.openfoodfacts.org/data-quality-info/en:quantity-contains-e

For the aforementioned example, see also from the API: https://world.openfoodfacts.org/api/v2/search?code=3464660002434

The aforementioned line of code can now be found in /lib/ProductOpener/SiteQuality_off.pm

Finally, to find more examples: 
https://mirabelle.openfoodfacts.org/products?sql=select+quantity+from+%5Ball%5D%0D%0AWHERE+quantity+%21%3D+%22%22+AND+quantity+LIKE+%22%25e%25%22%0D%0ALIMIT+100

stephanegigandet:
> In the first place, why do we want to identify quantities containing "e"? @stephanegigandet

That's a  good question. :) (sorry @VaiTon I missed it 4 years ago).

The symbol is present on most products, I didn't see the point in including it in the quantity field.

benbenben2:
Should we close this issue @CharlesNepote?

PR #2038, opened 2019-06-30
fix: solve quantity false positives issue #2037

CharlesNepote:
**Description:**

Change regexp to avoid many false positives.

**Related issues and discussion:** #2037


VaiTon:
@CharlesNepote could you provide some cases that the old regex matched and this one does not? Thank you! :-)

lib/ProductOpener/SiteQuality_off.pm
```diff
@@ -738,13 +738,15 @@ sub check_ingredients($) {
...
+# Check quantity values. See https://en.wiki.openfoodfacts.org/Products_quantities
 sub check_quantity($) {
 
 	my $product_ref = shift;
 
 	# quantity contains "e" - might be an indicator that the user might have wanted to use "℮" \N{U+212E}
+	# example: 650 g e
 	if ((defined $product_ref->{quantity})
-		and ($product_ref->{quantity} =~ /(?:.*e$)|(?:[0-9]+\s*[kmc]?[gl]?\s*e)/i)
+		and ($product_ref->{quantity} =~ /(?:^e\s)|(?:\se[^a-z])|(?:\se$)/i)
```
stephanegigandet: The purpose of the more complex regexp was to flag things like this : "500 ge"

stephanegigandet:
I think @VaiTon 's suggestion is a very good one: we should add test cases (both positives like "500 ge" and negatives like "1 litre") so that we don't break this later. There is already a test file : t/sitequality.t

CharlesNepote:
Yes, you're both right. I'm not very comfortable with test cases but I know it's important, so I'll do it.

teolemon:
rebased. @CharlesNepote can you write the test so we can merge

teolemon:
Can we close @CharlesNepote ?

benbenben2:
@CharlesNepote , @stephanegigandet , @VaiTon , @teolemon 

Tests have been added.
The regex (initially `/(?:.*e$)|(?:[0-9]+\s*[kmc]?[gl]?\s*e)/i`) have been updated. 
Because it seems to be covered already by the second part of the regex, the first part of the regex (`/(?:.*e$)|`) has been removed. 
Because this "e" letter is always (? could not find exception) preceded by "g" or "kg" or "l" or "ml", we can make `[gl]` as mandatory in the regex (by removing the question mark which makes it optional). This leads to: `/(?:[0-9]+\s*[kmc]?[gl]\s*e)/i`.

False positives such as "litre" or "verre" (ending by letter e) are not catch anymore with this regex.

stephanegigandet (approved):
Perfect, thank you!

PR #2047, opened 2019-07-03
Bring the French file in the official order of Crowdin.

teolemon:
Bring the French file in the official order of Crowdin. Please don't force edit the French po in the future.

### DIFF
--- a/po/common/fr.po
+++ b/po/common/fr.po
@@ -2376,7 +2376,7 @@ msgstr "Trier par"
 
 msgctxt "sort_completeness"
 msgid "Completeness"
-msgstr "Entièreté"
+msgstr "Complétude"
 
 msgctxt "sort_created_t"
 msgid "Add date"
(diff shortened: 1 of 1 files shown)


po/common/fr.po
```diff
@@ -2376,6 +2374,10 @@ msgctxt "sort_by"
 msgid "Sort by"
 msgstr "Trier par"
 
+msgctxt "sort_completeness"
+msgid "Completeness"
+msgstr "Entièreté"
```
stephanegigandet: Complétude

stephanegigandet:
"Bring the French file in the official order of Crowdin. Please don't force edit the French po in the future."

@teolemon : what does that mean?

Usually when I add entries to the po file, I put them in common.pot, en.po and fr.po. Can I do that?

po/common/fr.po
teolemon:
```suggestion
msgstr "Complétude"
```

VaiTon:
> 
> 
> "Bring the French file in the official order of Crowdin. Please don't force edit the French po in the future."
> 
> @teolemon : what does that mean?
> 
> Usually when I add entries to the po file, I put them in common.pot, en.po and fr.po. Can I do that?

I think it's better to add them just to the common file. 

I, too, used to add them to en.po but I will fix my bad habit :-)

PR #2050, opened 2019-07-04
Add a number of country defaults

CloCkWeRX:
**Checklist:**

* [x] Code is well documented
* [x] Include unit tests for new functionality
* [x] Code passes Travis builds in your branch
* [x] If you have multiple commits please combine them into one commit by squashing them.
* [x] Read and understood the [contribution guidelines](https://github.com/openfoodfacts/openfoodfacts-server/blob/master/CONTRIBUTING.md)

**Description:**
This adds a few country entries to specific labels. 

Ecocert Canada -> Assumption is they would only certify canadian goods
Bio Austria -> As above
French brands -> Assuming they are products of France
Champion beers -> assuming they are only open to beers produced in that country.

**Related issues and discussion:** Closes https://github.com/openfoodfacts/openfoodfacts-ai/issues/45

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -261,16 +261,12 @@ en:Australian International Beer Awards
 en:Brazilian Beer Awards
 
 en:Champion Beer of Britain
-country:United Kingdom
 
 en:Champion Beer of Scotland
-country:United Kingdom
 
 en:Champion Beer of Wales
-country:United Kingdom
 
 en:Champion Winter Beer of Britain
-country:United Kingdom
 
 en:European Beer Star
 ca:Cervesa Estrella Europea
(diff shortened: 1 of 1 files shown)


stephanegigandet:
We need to think carefully about what we exactly want to achieve here. We currently do not have a "country of origin" field (as in the product was made in that country). We could add one. Currently we use labels for that: "Made in Belgium".

We have a "origins of ingredients" field where we put countries, but that's something else entirely.

Also we should not make any assumptions without verifying them. It's very possible that Bio Austria etc. certify products that are imported. That does not mean they were produced in Austria.

CloCkWeRX:
Happy to reduce it to just the "Made in XYZ" labels for now; others later

CloCkWeRX:
> I think they need "country:en:" as the prefix.

I wasn't sure on that, as other examples didn't use that approach and it seemed more prevalent. Will swap those over however.

PR #2052, opened 2019-07-04
Updated store icons

VaiTon:
Updated store icons to reflect legal changes. And they're cooler too :)


teolemon:
I have pinged to you files on Slack that are localized, and already in prod.

Issue #1976, opened 2019-06-22
Merge allergens and ingredients taxonomies


aleene:
**Is your feature request related to a problem? Please describe.**
The same ingredients are now appearing in both taxonomies. Double work, double error chance.

**Describe the solution you'd like**
Add a variable to relevant parents, like allergen:en:milk. 

**Describe alternatives you've considered**
None

**Additional context**
We need to check (and if required merge) the existing allergens into the ingredients.

teolemon:
@aleene it has been merged. Can I assume this fixed ?

aleene:
@stephanegigandet recently we spoke about your plans to do this. I do not know if you continued it, and what still needs to be done.

PR #2058, opened 2019-07-05
Allergens & Compliants

aleene:
**Description:**
I went through the allergens.txt and added them to the ingredients.txt.

I created a separate "taxonomy"-file called compliant.text, which lists which ingredients match a certain allergy. This should cover much more synonyms and languages than additives.txt.

As the approach is the as allergens, I also added palm- and 7 vegetarian diets to this document. (the vegetarian diets are incomplete, as the additives still need to be added.).

Issue: #1976


aleene:
Added halal and fodmap diets

taxonomies/additives.txt
```diff
@@ -1592,6 +1598,7 @@ efsa_evaluation:en: Scientific Opinion on the reconsideration of the ADI and a r
 vegan:en:maybe
 vegetarian:en:maybe
 
+<en:E160
```
stephanegigandet: For the a/b/c/etc. additives, like E160a, we do not want to add E160 as a parent, it doesn't really exist and it will mess the additives detection.

taxonomies/ingredients.txt
```diff
@@ -10731,21 +10792,19 @@ es:lactosa y minerales de la leche
...
-
-fr:lait de montagne
-
-fr:lait entier standardisé
-
-fr:lait de bufflonne pasteurisé
 
 ################################## Yogurt related stuff ####################
 
+<en:dairy
```
stephanegigandet: Please don't add parents to show the origin. Instead you can put a property:
from_milk:en:yes

taxonomies/ingredients.txt
```diff
@@ -18093,6 +18232,7 @@ nl:peperolie
 # FISH OIL is oil derived from the tissues of oily fish
 
 <en:oil
+<en:fish
```
stephanegigandet: from_fish:en:yes

stephanegigandet:
For parents, we should only add parents that have a "is a" relationship to the child. "dairy" could be ok for "butter", but "fish oil" is not "fish". If the relation is different, then we need to use a property like "from_fish:en:yes". 

For flavourings, we need to be careful too. "crab flavouring" is not "crab". It may be "from crab" (especially if it's listed as "natural crab flavouring", otherwise it could be something else entirely).

taxonomies/ingredients.txt
aleene: "dairy" occurs as allergen, that is the reason why it has been added. For me it is an isa parent. At least in dutch it works like that.

aleene:
I agree with the fish (etc). Can we introduce from:en:fish or origin:en:fish?

I have to review the whole flavoring section, it has become a bit of a mess now.

PR #2059, opened 2019-07-05
add photo licensing information

hangy:
Based on #1962, so it will stay a draft until that one is completed (once Chrome 76 with lazy loading is out).

Some UI elements are added to allow re-users to attribute the photographers of products on OFF and it's sister sites correctly.

### Licensing info page
Loosely based on the attribution information of [wikimedia commons](https://commons.wikimedia.org/w/index.php?title=File:Prinzipalmarkt_(M%C3%BCnster).jpg&oldid=345551668). Shows the image, a link to the full image, as well as licensing info and the photographer/contributor. This should help with attribution, because [a good CC-BY-SA 3.0 attribution requires the author](https://wiki.creativecommons.org/index.php?title=Best_practices_for_attribution&oldid=116817#This_is_an_ideal_attribution).
image

### Link on product page
To help users find the licensing info page, each image on the product page got a new link with the Creative Commons icon. The link leads directly to the individual product image's licensing info page.

#### Small
image

#### Large
image


stephanegigandet:
Looks good to me. The CC icon is good candidate to put it the icons. I'll make one that looks like the Material icons.

hangy:
> Looks good to me. The CC icon is good candidate to put it the icons. I'll make one that looks like the Material icons.

I don't think we can modify the icon in that way: https://creativecommons.org/policies
> You are not authorized to use any modified versions of our trademarks, except that you may use a different color for the CC logo and its background so long as the two colors chosen have a contrast ratio of at least 3:1.

stephanegigandet:
@hangy : I put the same SVG icon, only sized like the Material icon so that I could put it in the font. Try icon-cc

VaiTon:
@hangy any news?

hangy:
I'm just waiting for the lazy loading attribute to be released in Chrome (probably July 30).

PR #2060, opened 2019-07-05
Don't save empty edits

### LINKED ISSUE #374: Null edits can be saved
Saving without changing anything creates an empty edit in the history.

hangy:
**Description:** If the computed `%diff` of the product is `undef` or empty, do not actually store the product changes on disk (neither `.sto` files, nor MongoDB). To enable reporting this to end users, `store_product` now returns `1` if the save was successful; otherwise it returns `0`.
**Related issues and discussion:** Fixes #374

### DIFF
--- a/cgi/product_jqm_multilingual.pl
+++ b/cgi/product_jqm_multilingual.pl
@@ -278,7 +278,7 @@ else {
 	detect_allergens_from_text($product_ref);
 	compute_carbon_footprint_from_ingredients($product_ref);
 	compute_carbon_footprint_from_meat_or_fish($product_ref);
-	
+
 	# Nutrition data
 
 	# Do not allow nutrition edits through API for data provided by producers
@@ -438,13 +438,17 @@ else {
 
 	my $time = time();
 	$comment = $comment . remove_tags_and_quote(decode utf8=>param('comment'));
-	store_product($product_ref, $comment);
+	if (store_product($product_ref, $comment)) {
+		# Notify robotoff
+		send_notification_for_product_change($product_ref, "updated");
 
-	# Notify robotoff
-	send_notification_for_product_change($product_ref, "updated");
-
-	$response{status} = 1;
-	$response{status_verbose} = 'fields saved';
+		$response{status} = 1;
+		$response{status_verbose} = 'fields saved';
+	}
+	else {
+		$response{status} = 0;
+		$response{status_verbose} = 'not modified';
+	}
 }
 
 my $data =  encode_json(\%response);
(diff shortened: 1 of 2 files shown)


stephanegigandet:
The changes look good, but it is very likely that the current diff does not encompass all fields, so we'll need to make sure that it does first. The diff was only intended to display changes in the edit history, so I never tried to make it 100% complete.

e.g. the fields are currently hardcoded in the diff:

	my @fields = qw(lang product_name generic_name quantity packaging brands categories origins manufacturing_places labels emb_codes expiration_date purchase_places stores countries ingredients_text traces no_nutrition_data serving_size nutrition_data_per);

product_multilingual.pm and product_jqm_multilingual.pl might have a few other fields.

And scripts/import_csv_file.pl has fields that are currently not supported in the API (they probably should) but that can be passed by manufacturers. They can be retrieved from Config.pm like this:

````
	foreach my $field ('lc', 'product_name', 'generic_name',
		@ProductOpener::Config::product_fields, @ProductOpener::Config::product_other_fields,
		'no_nutrition_data', 'nutrition_data_per', 'nutrition_data_prepared_per', 'serving_size', 'allergens', 'traces', 'ingredients_text','lang', 'data_sources') {

		if (defined $language_fields{$field}) {
			foreach my $display_lc (@param_sorted_langs) {
				push @param_fields, $field . "_" . $display_lc;
			}
		}
		else {
			push @param_fields, $field;
		}
	}

stephanegigandet:
@hangy : I added the new fields in compute_history() in this commit:
https://github.com/openfoodfacts/openfoodfacts-server/pull/2082/commits/1fd44aa30db2a43c6ecf09626bdc6d1a727aaf2f

VaiTon:
@stephanegigandet so now does the function compute for all the fields?

teolemon:
@stephanegigandet @hangy I use empty edits to force ingredient list recomputation and cutting using the Mass Edit extension. This will prevent such mass fixes that can be usefuls.

hangy:
> @stephanegigandet @hangy I use empty edits to force ingredient list recomputation and cutting using the Mass Edit extension. This will prevent such mass fixes that can be usefuls.

What do you mean by cutting? If the ingredients list changes, a new revision should be created. Otherwise, allowing empty edits nedlessly spams the server with empty product revision files. In the case of any API misbehaving (ie. accidentally sending the same edit multiple times), that could lead to problems and could be detected too late. Keep in mind that there's no option to delete old product revisions.

If some taxonomy lists need to be recomputed without the source data, I belive that should be done in a different way (ie. Stephane's server-side update scripts, that don't create new product versions).

stephanegigandet:
" If the ingredients list changes, a new revision should be created. " --> That's right.

For recomputation of tags (new taxonomies or parsing code), quality checks and all that, the cleanest solution would probably be to create a new API function for that.

VaiTon:
@stephanegigandet do you think we can merge this?

VaiTon:
I am gonna review this PR one more time and them I will merge it.

Issue #1567, opened 2018-12-08
Some images seem deindexed


teolemon:
https://static.openfoodfacts.org/images/products/326/385/221/4415/3.jpg
https://world.openfoodfacts.org/product/3263852214415/chocolat-lait-dessert-leader-price

stephanegigandet:
Same as #1866

PR #2075, opened 2019-07-09
Add scripts/add_missing_photos.pl

hangy:
**Description:** Adds a script that adds any `\d+\.jpe?g` file in the product image directory to the image.

**Related issues and discussion:** Could help in finding images missing because of #1567

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -47,6 +47,7 @@ requires 'XML::Rules';
 requires 'Email::Stuffer';
 requires 'Text::CSV', '>= 1.99, < 2.0';
 requires 'Text::Fuzzy';
+requires 'Try::Tiny', '>= 0.30';
 requires 'File::Copy::Recursive';
 
 # Logging
--- a/scripts/add_missing_photos.pl
+++ b/scripts/add_missing_photos.pl
@@ -1,0 +1,136 @@
+#!/usr/bin/perl -w
+
+# This file is part of Product Opener.
+#
+# Product Opener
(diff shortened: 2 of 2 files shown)


stephanegigandet:
Instead of reuploading the image, I think it would be best to add the existing missing imgid back in the images field directly. That way the imgid would stay the same, and references to this image would exist again. (e.g. for https://world.openfoodfacts.org/api/v0/product/3263852214415/chocolat-lait-dessert-leader-price the selected ingredients image comes from imgid 3).

This way we can also keep the upload date. And I think we could also assign the uploader when the date of the file is very close before its cropping / selection.

We just need to add a structure like this one for imgid 3:

````
2: {
uploader: "kiliweb",
uploaded_t: "1509706429",
sizes: {
100: {
h: 100,
w: 75
},
400: {
w: 300,
h: 400
},
full: {
h: 1360,
w: 1021
}
}
},

hangy:
> That way the imgid would stay the same, and references to this image would exist again. (e.g. for https://world.openfoodfacts.org/api/v0/product/3263852214415/chocolat-lait-dessert-leader-price the selected ingredients image comes from imgid 3).

I didn't know there were images where selected images could reference a non-"indexed" image. How is that possible in the first place? In the examples that I found, the products did not have any selected/cropped images. (ie. robotoff had suggestions based on uploaded images that weren't added to a product)

stephanegigandet:
At one point there was a bug where the image upload script with a division by zero error, that may be it. I haven't been able to completely understand how this bug happens.

Issue #315, opened 2016-06-02
Lots of Perl warnings when initializing Product Opener


stephanegigandet:
Initialization of the different modules called in by Product Opener perl scripts (including startup.pl which loads all of them when mod_perl starts) results in a lot of repeated Perl warnings:

Step to reproduce:  perl product.pl

[..]
SiteLang{android_apk_app_link}
SiteLang{android_apk_app_link}{en}
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value in substitution (s///) at Blogs/Lang.pm line 11538.
[..]
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value $file in substitution (s///) at Blogs/Store.pm line 79.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value $file in substitution (s///) at Blogs/Store.pm line 81.
[Thu Jun  2 14:27:04 2016] Tags.pm: Use of uninitialized value $file in substitution (s///) at Blogs/Store.pm line 83.
[..]
[...]

PR #2078, opened 2019-07-10
Replace Email::IsEmail with Email::Valid

hangy:
**Description:** Replace the email validation with a different module. See tests by @CharlesNepote [on slack](https://openfoodfacts.slack.com/archives/C045H3K57/p1549476230033000) for more background information.
**Related issues and discussion:** Removes some warnings upon startup, see #315

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -37,7 +37,7 @@ requires 'DateTime::Locale', '>= 1.22';
 requires 'Math::Random::Secure';
 requires 'Crypt::ScryptKDF';
 requires 'Locale::Maketext::Lexicon::Getcontext', '>= 0.05';
-requires 'Email::IsEmail', '>= 3.04.8';
+requires 'Email::Valid', '>= 1.202, < 2.0';
 requires 'CLDR::Number::Format::Decimal';
 requires 'CLDR::Number::Format::Percent';
 requires 'CLDR::Number';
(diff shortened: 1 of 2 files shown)


hangy:
I tested this locally on my server, and everything seemed to work fine.

cpanfile
```diff
@@ -37,7 +37,7 @@ requires 'DateTime::Locale', '>= 1.22';
 requires 'Math::Random::Secure';
 requires 'Crypt::ScryptKDF';
 requires 'Locale::Maketext::Lexicon::Getcontext', '>= 0.05';
-requires 'Email::IsEmail', '>= 3.04.8';
+requires 'Email::Valid', '>= 1.202, < 2.0';
```
CharlesNepote: `Email::Valid` exists as a Debian package: `libemail-valid-perl`. Isn't it better to use it?

hangy: We need the module to be listed in cpanfile, so that Travis CI and Docker can build successfully. If libemail-valid-perl is installed on Debian (stable is 1.202), then cpanm won't re-installed it either way. 🙂

CharlesNepote: I wrote this comment because you put `Email::Valid` under `# Probably not available as Debian packages` section (but that's not very important).

Thanks a lot @hangy!

PR #2084, opened 2019-07-12
added Catalan translation

moon-rabbitOFF:
Some Catalan words I added have an interpunct, like "Cel·lulosa". Will it causes issues?


aleene:
Usually the dot is used to help pronunciation, so I guess tit can be removed.

moon-rabbitOFF:
I thought it was really part of the writing in catalan because I found it in all dictionnaries

VaiTon:
@moon-rabbitOFF could you resolve conflicts so that we can merge?

Issue #2091, opened 2019-07-12
nginx caches and returns 304 even if there is a session cookie


stephanegigandet:
This causes opening a session to appear to fail:

1. open a session on https://world.openfoodfacts.org
2. session.pl is called, returns a session cookie and a redirect to the previous page https://world.openfoodfacts.org
3. nginx serves a 304 not modified which make it seem that the login was not successful (nothing changed on the page)

![image](https://user-images.githubusercontent.com/8158668/61144605-9073ac80-a4d5-11e9-872e-6d9b0a2e90d3.png)

hangy:
I cannot reproduce this.
1. Open https://world.openfoodfacts.org/ => 200
2. Login (POST to /cgi/session.pl)
3. 301 Redirect to /
4. https://world.openfoodfacts.org/ => 200 & logged in successfully

As far as I can tell, there's no setting for nginx to even [cache proxied results](https://www.nginx.com/resources/wiki/start/topics/examples/reverseproxycachingexample/) from Apache: https://github.com/openfoodfacts/openfoodfacts-server/blob/bcf6d255be58769c97c99699df369b18246e3e16/conf/nginx/sites-available/off#L171-L184

hangy:
> This causes opening a session to appear to fail:
> 
> 1. open a session on https://world.openfoodfacts.org
> 2. session.pl is called, returns a session cookie and a redirect to the previous page https://world.openfoodfacts.org
> 3. nginx serves a 304 not modified which make it seem that the login was not successful (nothing changed on the page)
> 
> ![image](https://user-images.githubusercontent.com/8158668/61144605-9073ac80-a4d5-11e9-872e-6d9b0a2e90d3.png)

Actually, it is possible to reproduce this. Not on the usual subdomains that I work on (which would be  de.openfoodfacts.org or world.openfoodfacts.org), but on **fr**.openfoodfacts.org. Apparently, the config at https://github.com/openfoodfacts/openfoodfacts-server/blob/bcf6d255be58769c97c99699df369b18246e3e16/conf/nginx/sites-available/off is not actually applied, because this has to mean that there's different config being used for that subdomain, even though all domains should actually be equal.

hangy:
https://www.webfoobar.com/node/28

VaiTon:
Any news on this?

stephanegigandet:
Probably not current, closing

PR #2092, opened 2019-07-12
update nginx config from production

stephanegigandet:
There may be updates that were done on github that we want to put in production.

off-fr is a special config we use so that we can cache things etc. when OFF goes on TV in France.


hangy:
> off-fr is a special config we use so that we can cache things etc. when OFF goes on TV in France.

Sorry for being that annoying guy, but why treat France differently? Wouldn't it be better to apply that same caching algorithm to all subdomains, so that there's no difference in application behaviour between subdomains? What's the downside? This is pretty confusing, as present by #2091.

stephanegigandet:
Well off-fr is something we have done while the server was unresponsive under heavy load, trying to make it go back to life as quickly as possible. It hasn't been planned. It's also a way to protect the rest of OFF when one country gets overloaded: we turn off certain features like search, and we do very aggressive caching (e.g. serving static files from nginx).

VaiTon:
We could keep the configuration and just rename and change it so that its effect is available side-wide.

stephanegigandet:
Can we merge this?

PR #2097, opened 2019-07-13
Reviewed  creams

aleene:
**Description:**
Performed an update of the creams section

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -10572,9 +10572,11 @@ fr:crème en poudre rehydratée
 # ingredient/rehydrated-cream-powder has 2 products @2019-07-12
 
 <en:cream
-en:Cream from Normandy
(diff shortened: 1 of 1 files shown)


moon-rabbitOFF:
"Crème UHT" and "Crème stérilisée UHT" are the same thing. Ultra-High Temperature processing is particular type of sterilisation.

moon-rabbitOFF:
Pasteurization is not considered as a sterilisation technique. It is carried out under 10°C, it kills bacteria but doesn't detroy bacterial spores (Wikipedien EN). So crème fraîche can be pasteurizated.

aleene:
OK. Note that these are the words used in ingredient lists. So I did not make it up. I understood that there are multiple sterilisation methods, so I did not make UHT and sterilised synonyms.

aleene:
I will removing many translations taken from wiikidata, as they seem to be wrong.

PR #2093, opened 2019-07-12
Issue 1930/ingredients classes

stephanegigandet:
Add support for parsing ingredients like this:

Fraise origine France, Cassis (origine Afrique du Sud), Framboise (origine : Belgique), Pamplemousse bio, Orange (bio), Citron (issue de l'agriculture biologique), cacao et beurre de cacao (commerce équitable)

PR #2099, opened 2019-07-13
Update Config.pm in Docker to match current Config_off.pm fields.

hangy:
Adds taxonomy fields etc. from Config_off.pm to the Config.pm for docker/helm.

### DIFF
--- a/docker/backend-dev/conf/Config.pm
+++ b/docker/backend-dev/conf/Config.pm
@@ -269,90 +269,142 @@ XML
 
 # fields for which we will load taxonomies
 
-@taxonomy_fields = qw(states countries languages labels categories additives additives_classes allergens traces nutrient_levels misc);
+@taxonomy_fields = qw(states countries languages labels categories additives additives_classes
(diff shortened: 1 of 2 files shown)


docker/backend-dev/conf/Config.pm
```diff
@@ -269,90 +269,142 @@ XML
...
+ emb_codes link expiration_date purchase_places stores countries  );
 
 # fields currently not shown in the default edit form, can be used in imports or advanced edit forms
 
 @product_other_fields = qw(
-producer_version_id
-net_weight_value net_weight_unit drained_weight_value drained_weight_unit volume_value volume_unit
-other_information conservation_conditions recycling_instructions_to_recycle recycling_instructions_to_discard
-nutrition_grade_fr_producer
+	producter_product_id
```
stephanegigandet: I made a typo there, it's producer_product_id . It's fixed in https://github.com/openfoodfacts/openfoodfacts-server/pull/2093

hangy: Fixed, thanks.

Issue #2106, opened 2019-07-14
Missing mapping for XK country code


teolemon:
Map XK to Kosovo. We receive data like, and there's no mapping.

lc=fr&cc=XK&code=5319991131019
--
lc=fr&cc=XK&code=9000159714507

stephanegigandet:
This is fixed I think.

Issue #2102, opened 2019-07-13
country/language taxonomy do not build


hangy:
I currently (as of 97bdb87e11599086c44616b684f632608197c158) cannot build the language or country taxonomy. 
```
Errors in the languages taxonomy definition:
ERROR - en:en already is a synonym of cornish () - cannot make a synonym of english for en:english
ERROR - en:en already is a synonym of cornish () - cannot make a synonym of english for en:english
ERROR - hi:तुलु-भाषा already is a synonym of चुवाश-भाषा () - cannot make a synonym of एबी-भाषा for en:ewe
ERROR - hi:तुलु-भाषा already is a synonym of चुवाश-भाषा () - cannot make a synonym of गुआरानी-भाषा for en:guarani
ERROR - be:мова already is a synonym of гуарані () - cannot make a synonym of хаўса for en:hausa
ERROR - hi:तुलु-भाषा already is a synonym of चुवाश-भाषा () - cannot make a synonym of हौसा-भाषा for en:hausa
ERROR - hi:तुलु-भाषा already is a synonym of चुवाश-भाषा () - cannot make a synonym of हंगेरियाई-भाषा for en:hungarian
ERROR - be:мова already is a synonym of гуарані () - cannot make a synonym of ігба for en:igbo-language
ERROR - de:gronlandische-sprache already is a synonym of gronlandisch () - cannot make a synonym of inuktitut for en:inuktitut
ERROR - hi:तुलु-भाषा already is a synonym of चुवाश-भाषा () - cannot make a synonym of जापानी-भाषा for en:japanese
ERROR - hi:तुलु-भाषा already is a synonym of चुवाश-भाषा () - cannot make a synonym of कन्नड़-भाषा for en:kannada
ERROR - hi:तुलु-भाषा already is a synonym of चुवाश-भाषा () - cannot make a synonym of कनूरी-भाषा for en:kanuri
ERROR - hi:तुलु-भाषा already is a synonym of चुवाश-भाषा () - cannot make a synonym of मलयालम-भाषा for en:malayalam
ERROR - be:мова already is a synonym of гуарані () - cannot make a synonym of дывехі for en:maldivian-language
ERROR - en:gaelic already is a synonym of irish () - cannot make a synonym of manx for en:manx
ERROR - be:мова already is a synonym of гуарані () - cannot make a synonym of маратхі for en:marathi
[...]

PR #2111, opened 2019-07-16
Remove synonyms used for multiple languages or countries 

stephanegigandet:
bug #2102, add country code for Kosovo, bug #2106


stephanegigandet:
@teolemon @hangy : can one of you have a look?

PR #2112, opened 2019-07-16
add en:dairy which was not defined

stephanegigandet:


aleene:
It is in one of my pull requests

PR #2114, opened 2019-07-16
Fix decomposition of ingredients, add ?regexp=pattern parameter

stephanegigandet:


stephanegigandet:
I fixed the bug reported by @aleene (#2119), and I needed to easily filter the ingredients lists, so I added a regexp parameter that works like this:

https://fr.dev.openfoodfacts.org/ingredients?regexp=arome

hangy:
Could the regexp parameter be subject to code injection? http://www.pal-blog.de/entwicklung/perl/sql-injection-2-0-regular-expression-injection.html

stephanegigandet:
I tried it and I get an error: "Eval-group not allowed at runtime, use re 'eval' in regex", it seems we would need to put "use re 'eval';" in the code for that to work.

But it's needlessly dangerous, I'll escape the regexp.

stephanegigandet:
@hangy : it's now fixed.

PR #2124, opened 2019-07-21
Recreated "Android apk" badge as SVG

hangy:
I recreated the "Android apk" badge as an SVG, so it scales better and doesn't require different versions.


teolemon:
@VaiTon I gave translated logo urls to @VaiTon @hangy 
But this is a very good first step

hangy:
> I gave translated logo urls to @VaiTon @hangy

This is for the direct APK download, not the Google Play Store.

PR #2158, opened 2019-08-08
Swedish OFF logos

Karljoones:


teolemon:
thanks @Karljoones How did you create it ? Would you be open to creating some more ?

Karljoones:
@teolemon I used Inkscape! Can do, if you want to send me over the translations. I was planning on doing some with white text for apps with dark mode - partly for Recipe Keep, as I use the icon to credit OFF and the text fades into the background for dark mode

PR #2160, opened 2019-08-09
butterfat cleanup

aleene:
**Description:**
Cleanup of the butterfat ingredients


stephanegigandet:
@aleene : there are 2 en:butterfat sections, this is making loops in the taxonomy, can you remove one of them?

aleene:
I have to get into the failed testing thing, so I can discover what is wrong.

stephanegigandet:
@aleene : if you click on the "Details" link next to "The Travis CI build failed", you can see the log of the tests: https://travis-ci.org/openfoodfacts/openfoodfacts-server/builds/579438663?utm_source=github_status&utm_medium=notification

Then you can search for "fail" to see which test failed and why:

#   Failed test at t/ingredients.t line 338.
#          $got->{ingredients_tags}[1] = 'en:oil-and-fat'
#     $expected->{ingredients_tags}[1] = 'en:oil'

Here it's because you added new parents to all oil and fat, which causes new tags to be generated for many products, including some products that are used in the tests. So the tests need to be updated to indicate that the new output is correct. The corresponding tests are here: https://github.com/openfoodfacts/openfoodfacts-server/blob/master/t/ingredients.t

You can update the test files, but it's easier if you have a Product Opener instance running so that you can run the tests directly on your machine.

For ingredients, I think it's best to add parents only when they are needed for an actual use (like computing NOVA, vegan / vegetarian / palm oil etc.). It's different from categories where it's useful to have many different levels. For ingredients, the fewer levels, the better.

aleene:
If I update the test, do I it together with the taxonomy of first a test pull request?

stephanegigandet:
If you can, update the taxonomy and the test together. If it's done in sequence, no pb either, but it won't be mergeable until the build passes.

aleene:
The hierarchy is very useful for the search function and it helps keeping an overview of the ingredients (better than just sorting).

aleene:
Hierarchy also allows to define more complex diets (not just vegan/vegetarian).

aleene:
Unclear what is failing now.

stephanegigandet:
There's a syntax error in ingredients.t, so the tests are not running and failing.

String found where operator expected at t/ingredients.t line 1175, near "'from_palm_oil'"
	(Missing semicolon on previous line?)
syntax error at t/ingredients.t line 1175, near "'from_palm_oil'"
Execution of t/ingredients.t aborted due to compilation errors.
t/ingredients.t ..................

PR #2174, opened 2019-08-13
Nutriscore exempt edit

aleene:
All food additives are exempt from Nutriscore. This will include yeast, gelatin and flavourings, which are mentioned explicitly the the NutriScore articles.


lib/ProductOpener/Config_off.pm
```diff
@@ -417,10 +417,11 @@ $options{categories_exempted_from_nutriscore} = [qw(
 	en:coffees
 	en:teas
 	en:herbal-teas
-	fr:levure
-	fr:levures
+	en:food additives
```
stephanegigandet: In this list we need the canonical name of the entries.
```suggestion
	en:food-additives
```

Issue #2177, opened 2019-08-15
Many food additions are not recognized.


Slawek234:
Many food additions are not recognized.
Examples
1. https://pl.openfoodfacts.org/product/5904277900566/kiełbasa-wieprzowa-z-dodatkiem-mięsa-wołowego-salami-z-pieprzem-konecke  - R̶e̶c̶o̶g̶n̶i̶z̶e̶d̶:̶ ̶E̶2̶0̶2̶,̶ ̶n̶o̶t̶ ̶r̶e̶c̶o̶g̶n̶i̶z̶e̶d̶:̶ ̶E̶3̶0̶1̶,̶ ̶E̶1̶2̶0̶,̶ ̶E̶2̶5̶0̶
2. https://pl.openfoodfacts.org/product/5900562001147/wieprzowina-w-sosie-wlasnym-sokołow - 
R̶e̶c̶o̶g̶n̶i̶z̶e̶d̶:̶ ̶E̶4̶5̶0̶,̶ ̶n̶o̶t̶ ̶r̶e̶c̶o̶g̶n̶i̶z̶e̶d̶:̶ ̶g̶l̶u̶t̶a̶m̶i̶n̶i̶a̶n̶ ̶m̶o̶n̶o̶s̶o̶d̶o̶w̶y̶ ̶(̶ ̶M̶S̶G̶,̶ ̶E̶6̶2̶1̶)̶

3. https://pl.openfoodfacts.org/product/5900344102444/płaty-śledziowe-w-zalewie-octowej-lisner - Recognized: E260, E262, not recognized: Sachary (E954),

4. https://pl.openfoodfacts.org/product/5908260252168/napoj-gazowany-o-smaku-jagod-acai-miechunki-peruwiańskiej-oshee - not recognized: Kwas cytrynowy (E330) Cytryniany sodu (E331)

5. https://pl.openfoodfacts.org/product/7622300644574/herbatniki-dekorowane-cukrem-san - not recognized: węglany amonu (E502), węglany sodu (E500),

aleene:
These synonyms are not yet in the additives taxonomy.
I submitted a pull request: https://github.com/openfoodfacts/openfoodfacts-server/pull/2178

aleene:
`węglany sodu` should have been recognised, it is in the file, but capitalised.

stephanegigandet:
Thanks for the report @Slawek234

[...]

PR #2182, opened 2019-08-15
added plurals for Polish additives classes, bug #2177

stephanegigandet:
Additives classes are used for additives and ingredients parsing, so we need both singular and plurals, as well as synonyms used in ingredients lists.

### DIFF
--- a/taxonomies/additives_classes.txt
+++ b/taxonomies/additives_classes.txt
@@ -21,7 +21,7 @@ lv: Skābe
 cs: Kyselina
 et: Hape
 hu: Étkezési sav
-pl: Kwas
+pl: Kwas, Kwasy
 sl: Kislina
 lt: Rūgštis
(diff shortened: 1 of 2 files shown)


stephanegigandet:
Right now it's only used as strings before additives, with or without a separator (e.g. "colour: caramel", "colour caramel"). If there are languages where it goes after ingredients, we could try to support it too.

taxonomies/additives_classes.txt
```diff
@@ -775,7 +775,7 @@ lv: Konservants
 cs: Konzervant
 et: Säilitusaine
 hu: Tartósítószer
-pl: Substancja konserwująca
+pl: Substancja konserwująca, Substancja niekonserwująca, Substancje konserwujące
```
Slawek234: Please delete the entry Substancja niekonserwująca,

PR #2186, opened 2019-08-16
Update yarn.lock

teolemon:
I have no idea what I'm doing
image


hangy:
The bad dependency is pulled in by the gulp-iconfonts package. I've created a few PRs upstream, but didn't receive a response.

Tbh, I don't think this dependency actually affects us, because it's only run during build. But let's merge this, if it builds. 👍

PR #2188, opened 2019-08-16
Nutriscore exempts

aleene:
**Description:**

Added chewing gums, aromatic herbs and sugar substitutes

### DIFF
--- a/lib/ProductOpener/Config_off.pm
+++ b/lib/ProductOpener/Config_off.pm
@@ -410,18 +410,21 @@ $options{categories_considered_as_beverages_for_nutriscore} = [qw(
 )];
 
 $options{categories_exempted_from_nutriscore} = [qw(
+	en:alcoholic-beverages
+	en:aromatic-herbs
 	en:baby-foods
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Config_off.pm
```diff
@@ -410,18 +410,21 @@ $options{categories_considered_as_beverages_for_nutriscore} = [qw(
...
 	en:food-additives
+	en:herbal-teas
 	en:honeys
-	en:vinegars
+	en:meal-replacements
 	en:salts
 	en:spices
+	en:sugar-substitutes
+	en:vinegars
+	en:teas
```
stephanegigandet: We have too many ice teas, tea beverages that are (badly) categorized in en:teas, so we should not add it for now. Teas that don't have nutrition facts won't have nutriscore computed anyway.

aleene: What is the right category for these?

VaiTon:
```suggestion
```

PR #2208, opened 2019-08-20
polish fix

teolemon:
Fix in GitHub since the translation wasn't available in the web interface

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -644,7 +644,7 @@ he:נטול גלוטן
 hu:Gluténmentes
 it:Senza glutine
 nl:Glutenvrij, Gluten vrij, Vrij van gluten, Zonder gluten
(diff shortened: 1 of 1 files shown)


taxonomies/labels.txt
```diff
@@ -644,7 +644,7 @@ he:נטול גלוטן
 hu:Gluténmentes
 it:Senza glutine
 nl:Glutenvrij, Gluten vrij, Vrij van gluten, Zonder gluten
-pl: Bezglutenomy
+pl: Bezglutenowy
```
VaiTon: Why is there a leading space?

stephanegigandet: The leading space will be removed, but it's better without one just for consistency.

stephanegigandet:
```suggestion
pl:Bezglutenowy
```

Issue #1466, opened 2018-11-07
Add quality facet for dehydratated products that are missing prepared values


teolemon:
Add quality facet for dehydratated products that are missing prepared values

teolemon:
This one is important because very popular products are dehydrated, and they need to be well completed.

zigouras:
I can work on this one if you want to assign it to me.
Can someone show me in the UI where the quality facets are collected?  I assume they are stored in the `"quality_tags" : [ ]` array element of the product record.

CharlesNepote:
Thanks @zigouras.

All quality facets can be found here: https://world.openfoodfacts.org/quality/

The script producing quality facets can be found here: https://github.com/openfoodfacts/openfoodfacts-server/blob/master/lib/ProductOpener/SiteQuality_off.pm

It's not very well documented but it should be understandable.

teolemon:
The right category to use is probably "https://world.openfoodfacts.org/category/dried-products-to-be-rehydrated"

zigouras:
How can I tell if a product is "missing prepared values"?

zigouras:
I am ready to create a pull request for this from my fork I just need answers to the above questions.  Thanks.

stephanegigandet:
Hi @zigouras, the "nutriments" hash contains the nutrition values, you need to check that we have values suffixed with "_prepared_100g". e.g. "energy_prepared_100g" should be there. You can have a look at this product for instance: https://world.openfoodfacts.org/api/v0/product/7613034897930/puree-maggi

It will be better than just checking the nutrition_data_prepared field.

"missing-nutrition-data-prepared-with-category-dried-products-to-be-rehydrated" sounds good to me.

zigouras:
Here is my proposed pseudo-logic:
`if not checked nutrition_data_prepared
or
nutriments hash contains at least one key suffixed with _prepared_100g with a defined value > 0
then
add quality tag missing-nutrition-data-prepared-with-category-dried-products-to-be-rehydrated
end if
`
or do I specifically have to look for the nutriment key `energy_prepared_100g`?
Thanks.

stephanegigandet:
Hi @zigouras, the proposed logic is good. If they are specifying nutriments per serving, we will compute the per 100g values (provided they also specified the serving size).

stephanegigandet:
Fixed by @zigouras and deployed. Thank you!

PR #2210, opened 2019-08-20
Issue/1466

zigouras:
**Description:**

Code changes for issue 1466.

**Related issues and discussion:** #[ISSUE NUMBER]

I found a number of inconsistencies in the existing logic related to other quality tags.  They looked like bugs to me but I tried to change as little existing logic as possible except where I thought it was an obvious bug.

Below are my notes from what I encountered in the assignment of quality tags in SiteQuality_off.pm:

 

- I moved check for nutriments hash below unrelated quality logic because I thought that logic should be applied applied even if there is not hash.  I am not sure if that hash is ever undefined, even if no nutriments are filled out.
-    fixed issue that could both tag no-nutrition-data and nutrition-data-prepared* and nutrition-data-per-serving*.
-  nutrition-all-values-zero logic wrong, I didn't change anything here
-  no-nutrition-data does not clear the nutriment values in the UI, I didn't change anything here
-  per serving or prepared does not clear the other's nutrition values from the UI, I didn't change anything here

### DIFF
--- a/lib/ProductOpener/SiteQuality_off.pm
+++ b/lib/ProductOpener/SiteQuality_off.pm
@@ -464,24 +464,24 @@ sub check_nutrition_data($) {
 			push @{$product_ref->{quality_tags}}, "not-comparable-nutrition-data";
 		}
 	}
-	my $is_dried_product = has_tag($product_ref, "categories", "en:dried-products-to-be-rehydrated");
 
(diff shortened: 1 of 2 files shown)


zigouras:
I think you will want to squash the three commits into one.

zigouras:
Fixes issue #1466

Issue #1798, opened 2019-05-23
Travis CI builds fail when installing perl and perl modules


stephanegigandet:
The "system" perl has been removed:
https://storage.googleapis.com/travis-ci-language-archives/perl/binaries/ubuntu/14.04/x86_64/perl-system.tar.bz2 now returns a 404.

Replacing "system" by "5.24" in .travis.yml makes perl install work, but then some modules fail to install:

https://travis-ci.org/openfoodfacts/openfoodfacts-server/builds/536138563

! Installing Image::Magick failed. See /home/travis/.cpanm/work/1558596523.8981/build.log for details. Retry with --force to force install it.
! Installing the dependencies failed: Module 'Barcode::ZBar' is not installed, Module 'Apache2::Request' is not installed, Module 'Image::Magick' is not installed
! Bailing out the installation for ..
170 distributions installed
The command "cpanm --quiet --installdeps --notest --skip-satisfied ." failed and exited with 1 during .

stephanegigandet:
Last passing build was https://travis-ci.org/openfoodfacts/openfoodfacts-server/builds/503290348
Perl "system" was already not found, but the build kept running, using Perl 5.18.2

cpanm --quiet --installdeps --notest --skip-satisfied .   --> it seems there was no output at all, so the build continued.

In the current build, there's an output:

241.98s$ cpanm --quiet --installdeps --notest --skip-satisfied .
! Configure failed for Apache2::Request. See /home/travis/.cpanm/work/1558596523.8981/build.log for details.
Successfully installed DateTime-Locale-1.24 (upgraded from 1.17)

stephanegigandet:
"Fixed" by putting perl 5.18.4.  There are cpan module install errors with greater versions of Perl on Travis.

in .travis.yml:

perl:
  - "5.18.4"

PR #2217, opened 2019-08-23
Update Modern::Perl

hangy:
Use the current version of the `Modern::Perl` module with the options available with Perl 5.24.x.. I also updated the copyright year, as it was severely outdated in some cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,7 @@
+dist: bionic
 branches:
   except:
   - l10n_master
-language: perl
-perl:
-  - "5.24"
 env: COVERAGE=1
 cache:
   yarn: true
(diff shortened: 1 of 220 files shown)


stephanegigandet:
@hangy : I wasn't able to make Travis use Perl 5.24 a few months ago, 3 modules did not build:
https://github.com/openfoodfacts/openfoodfacts-server/issues/1798

hangy:
Yes, we probably need to install some headers, so that ZBar and other modules can be compiled. I'll have another look at it later this weekend.

hangy:
> I hope you didn't have to do this by hand :)

Search & replace worked quite well. 😄

Issue #2214, opened 2019-08-22
Some products have lc set to null


stephanegigandet:
> db.products.count({lc: null})
60

example: https://world.openfoodfacts.org/api/v0/product/00002523
in this case, we have the lang field set to "en".

This makes some processing fail (e.g. updating ingredients processing for all products).

stephanegigandet:
Fixed:

> db.products.find({lc: null}, {code : 1})
>

PR #2218, opened 2019-08-23
add more options to update_all_products.pl + fix for products without lc, bug #2214

stephanegigandet:


scripts/update_all_products.pl
```diff
@@ -20,7 +20,7 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use Modern::Perl '2012';
+use Modern::Perl '2017';
```
VaiTon: Do we want to mix up Modern 2012 & 2017?

scripts/update_all_products.pl
```diff
@@ -217,6 +245,22 @@
 			}
 		}
 
+		# Fix products that were created without the lc field, but that have a lang field
+		if ($fix_missing_lc) {
```
hangy: Wouldn't it be cleaner to use `if ($fix_missing_lc and (not defined $product_ref->{lc})) {` here and simplify the following `if` statements? There's probably not a huge advantage in printing `$product_ref->{lang}` for every product?

scripts/update_all_products.pl
```diff
@@ -217,6 +245,22 @@
...
+			print STDERR "lang: " . $product_ref->{lang} . "\n";
+			if ((not defined $product_ref->{lc}) and (defined $product_ref->{lang}) and ($product_ref->{lang} =~ /^[a-z][a-z]$/)) {
+				print STDERR "fixing missing lc, using lang: " . $product_ref->{lang} . "\n";
+				$product_ref->{lc} = $product_ref->{lang};
+				$product_values_changed = 1;
+			}
+			elsif (not defined $product_ref->{lc}) {
+				print STDERR "fixing missing lc, lang also missing, assignin en";
```
hangy:
```suggestion
				print STDERR "fixing missing lc, lang also missing, assigning en\n";
```
typo

scripts/update_all_products.pl
VaiTon: #2217 has been merged so no longer a problem.

scripts/update_all_products.pl
stephanegigandet: Right, I will change it, thanks!

PR #2233, opened 2019-08-27
German ingredients

aleene:
**Description:**
Added missing german ingredients,

**Related issues and discussion:** #2231


taxonomies/ingredients.txt
```diff
@@ -482,7 +482,7 @@ en:emulsifier
 bg:Емулгатор
 cs:Emulgátor
 da:Emulgator
-de:Emulgator
+de:Emulgator, Emulgatoren
```
stephanegigandet: The ingredients that are additives classes can be removed from the ingredients taxonomy, the additives classes taxonomy is automatically added to the ingredients taxonomy.

PR #2244, opened 2019-08-28
added users translations to ingredients taxonomy

stephanegigandet:


taxonomies/ingredients.txt
```diff
@@ -240,6 +239,7 @@ stopwords:zh::配料
 
 
 fr:additifs alimentaires
+	'dont acides gras satur(e|é)s
```
VaiTon: Why is this here?

stephanegigandet: I don't know, some bug with my script to add the translations. I fixed it manually.

PR #2246, opened 2019-08-28
Ingredients german

stephanegigandet:
A few ingredients / additives fixes


lib/ProductOpener/Tags.pm
```diff
@@ -3559,7 +3559,15 @@ sub add_users_translations_to_taxonomy($) {
 
 				foreach my $l (keys %{$users_translations_ref}) {
 					if (defined $users_translations_ref->{$l}{$tagid}) {
-						$translations{$l} = $users_translations_ref->{$l}{$tagid}{to};
+
+						if (not defined $translations{$l}) {
```
VaiTon: And how do we edit them if we can't add them if already added?

stephanegigandet: At this point, only by making a pull request directly on github on the taxonomy files. It's rather safe to add translations, but editing translations needs to be reviewed as it can break things.

PR #2248, opened 2019-08-29
add plurals of additives (de)

Dwarfex:
**Description:**

Added plurals for some additives in german

**Related issues and discussion:** #2231

stephane in Slack:

Could someone add the German plurals of additives classes in this file https://github.com/openfoodfacts/openfoodfacts-server/blob/master/taxonomies/additives_classes.txt ? e.g. Säureregulator -> Säureregulator, Säureregulatoren
This helps a lot to correctly parse ingredients, to recognize additives, compute NOVA etc.


teolemon:
Looks good to me

PR #2249, opened 2019-08-29
commit app stores logos

stephanegigandet:
The logos are on production but not in github.


teolemon:
I've added missing logos on the server (Microsoft ones in some languages)

stephanegigandet:
> I've added missing logos on the server (Microsoft ones in some languages)

Can you add them to this branch before we merge it?

Issue #2254, opened 2019-08-31
Support for private products and coordinated Product Opener instances


stephanegigandet:
This is to setup a new instance of Product Opener to create an Import area for producers and power users, so that they can import their products on the Import instance, test everything is correct, and then move the products to the public instance of OFF.

Main change: products on the Import instance are private, users can only see the products they add themselves. The id of each product is not only the barcode, but the userid (or organization id) + barcode.

Some changes are also made so that logging in on one instance works on all other instances.

stephanegigandet:
Currently live on https://world.openfoodfacts.dev and https://world.import.openfoodfacts.dev

stephanegigandet:
First version in production.

Issue #622, opened 2017-01-17
Remove circular dependencies in ProductOpener modules


stephanegigandet:
### What
- e.g. in Food.pm :  use ProductOpener::Tags qw/:all/;
- in Tags.pm : use ProductOpener::Food qw/:all/;
- This can cause strange bugs depending on the order of the use clauses, modules initialized multiple times etc.
- We need to create a graph of the uses in Product Opener modules, and find ways to remove cycles.
### Part of
- #5529

hangy:
Meh, I tried graphing the dependencies with `App:PrereqGrapher`, but apparently it fails when I set `depth` to anything higher than 1 (when analyzing Display.pm) or if I try to analyze all of our *.pm files with a wildcard. I suppose it could just abort because of the circular dependency we want to avoid. 😆

hangy:
I built this graph based on 501837dae42fb3f92307384278c4efcb6b88cfc9 and the curent `use` statements, without testing whether any subs or variables are actually used: https://hangy.de/off/dependencies2.svg

<details><summary>GraphViz</summary>
<p>

```graphviz
digraph g{
  rankdir=LR;
  "ProductOpener::Cache" -> "ProductOpener::Store";
  "ProductOpener::Cache" -> "ProductOpener::Config";

  "ProductOpener::Config" -> "ProductOpener::Config2";

  "ProductOpener::Data" -> "ProductOpener::Config";

  "ProductOpener::Display" -> "ProductOpener::Store";
  "ProductOpener::Display" -> "ProductOpener::Config";
  "ProductOpener::Display" -> "ProductOpener::Tags";
  "ProductOpener::Display" -> "ProductOpener::TagsEntries";
  "ProductOpener::Display" -> "ProductOpener::Users";
  "ProductOpener::Display" -> "ProductOpener::Index";
  "ProductOpener::Display" -> "ProductOpener::Cache";
  "ProductOpener::Display" -> "ProductOpener::Lang";
  "ProductOpener::Display" -> "ProductOpener::Images";
  "ProductOpener::Display" -> "ProductOpener::Food";
  "ProductOpener::Display" -> "ProductOpener::Ingredients";
  "ProductOpener::Display" -> "ProductOpener::Products";
  "ProductOpener::Display" -> "ProductOpener::Missions";
  "ProductOpener::Display" -> "ProductOpener::MissionsConfig";
  "ProductOpener::Display" -> "ProductOpener::URL";
[...]

PR #2255, opened 2019-08-31
Initial support for private products and coordinated Product Opener instances

stephanegigandet:
bug #2254

-> some changes to share session cookies between different instances
-> changed the _id field to incorportate the owner of each product on the Import instance


lib/ProductOpener/Users.pm
```diff
@@ -367,19 +369,26 @@ sub init_user()
...
+	if (defined $server_options{cookie_domain}) {
+		$cookie_domain = "." . $server_options{cookie_domain};	# e.g. fr.import.openfoodfacts.org sets domain to .openfoodfacts.org
+	}
 
 	$cookie = undef;
 
 	$Visitor_id = undef;
 	$Facebook_id = undef;
 	$User_id = undef;
+	$Org_id = undef;
```
hangy: How/when is this initialized?

cgi/product_image_move.pl
```diff
@@ -138,55 +137,56 @@
 
 if ($move_to ne 'trash') {
 
-	my $new_path = product_path($move_to);
+	my $move_to_id = product_id_for_user($User_id, $Org_id, $move_to);
+	my $new_path = product_path($move_to_id);
```
hangy: Should this be `product_path_from_id`?

lib/ProductOpener/Config2_sample.pm
```diff
@@ -76,4 +78,12 @@ $crowdin_project_key = '';
 # enable an in-site robotoff-asker in the product page
 $robotoff_url = '';
 
+%server_options = (
```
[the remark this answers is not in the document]
stephanegigandet: Yes, same for Config.pm : overtime, I'd like to move all the non-required fields to the %options and %server_options hashes. The goal is to avoid having to change all Config files whenever we want to add a new option.

lib/ProductOpener/Users.pm
stephanegigandet: Right now it is a placeholder, we'll need to add forms options etc. to create orgs.

cgi/product_image_move.pl
stephanegigandet: Good catch, I'll change it.

stephanegigandet:
> None of these changes apply to MongoDB, right? At least not when the database is searched for a product via any of the API methods or facet search on the UI.

Not yet. All mongo queries will have an extra "owner" filter that will be set from the user session.

> I think all of this is a relatively complex change for a very specific edge case. Wouldn't it be much easier to host (small) distinct instances of OFF for each producer - protected by HTTP Basic Auth? The effort of hosting separate instances of OFF with Docker would be small when compared to re-writing a large portion of the server software for this use case.

We thought about it, but our goal is to have thousands of small to medium producers using the import platform. I don't think it would work to have separate apache, mongodb etc. instances for each of them. For France we already have around 100 producers that have contacted us proactively with questions on how to integrate their data. And we are hoping that we can have many more if we manage to create a simple and easy import process that they can manage themselves from end to end, without having us as a bottleneck.

po/common/common.pot
```diff
@@ -3515,3 +3515,151 @@ msgstr ""
 msgctxt "ingredients_analysis_disclaimer"
 msgid "The analysis is based solely on the ingredients listed and does not take into account processing methods."
 msgstr ""
+
+msgctxt "import_data_file_title"
+msgid "Import a product data file"
+msgstr "Import a product data file"
+
+msgctxt "import_data_file_description"
+msgid "Upload a spreadsheet file (Excel .xslx files or a tab separated CSV file) with products data."
```
```suggestion
msgid "Upload a spreadsheet file (Excel .xlsx files or a tab separated CSV file) with products data."
```

po/common/common.pot
```diff
@@ -3515,3 +3515,151 @@ msgstr ""
...
+msgstr "Upload a spreadsheet file (Excel .xslx files or a tab separated CSV file) with products data."
```
```suggestion
msgstr "Upload a spreadsheet file (Excel .xlsx files or a tab separated CSV file) with products data."
```

lib/ProductOpener/Products.pm
```diff
@@ -210,6 +268,17 @@ sub init_product($) {
 		rev=>0,
 	};
 
+	if ((defined $server_options{private_products}) and ($server_options{private_products})) {
+		my $owner = "user-" . $userid;
```
hangy: Potentially conflicts with `$ProductOpener::Display::owner`.

stephanegigandet: I think I can remove Display.pm being loaded in Products.pm, which would make sure $owner does not come from Display.pm

scripts/minion_import.pl
```diff
@@ -0,0 +1,69 @@
...
+use Storable qw/dclone/;
+use Encode;
+use JSON::PP;
+use Time::Local;
+use Data::Dumper;
+use Text::CSV;
+use Getopt::Long;
+
+
+use Mojolicious::Lite;
```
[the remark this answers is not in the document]
stephanegigandet: I will add them to cpanfile.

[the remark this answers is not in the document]
stephanegigandet: One more thing we now need is a postgresql database, for the Minion backend.

lib/ProductOpener/Products.pm
```diff
@@ -191,13 +245,17 @@ sub product_exists($) {
...
-sub init_product($) {
+sub init_product($$$) {
 
+	my $userid = shift;
+	my $orgid = shift;
 	my $code = shift;
 
+	$log->debug("init_product", { userid => $userid, orgid => $orgid, code => $code }) if $log->is_debug();
+
 	my $creator = $User_id;
```
hangy:
```suggestion
	my $creator = $userid;
```

lib/ProductOpener/Products.pm
hangy: Would be a good step towards #622. :)

hangy: Would be a good step towards #622. :)

lib/ProductOpener/Producers.pm
```diff
@@ -0,0 +1,531 @@
...
+
+			while (my $row = $csv->getline ($io)) {
+				push @$rows_ref, $row;
+			}
+		}
+		else {
+			$results_ref->{error} = "Could not open CSV $file: $!";
+		}
+	}
+	else {
```
hangy: Do we want to limit this to the formats that `Spreadsheet::CSV` currently supports? (xls, xlsx, ods, sxc, gnumeric, ksp)?

hangy: Actually, do we need both branches? According to https://metacpan.org/pod/Spreadsheet::CSV, the module is a drop-in replacement for `Text::CSV`, so it supports .csv, too.

lib/ProductOpener/Producers.pm
```diff
@@ -0,0 +1,531 @@
...
+	$extension =~ s/^(.*)\.//;
+	$extension = lc($extension);
+
+	if (($extension eq "csv") or ($extension eq "tsv") or ($extension eq "txt")) {
+
+		my $encoding = "UTF-8";
+
+		$log->debug("opening CSV file", { file => $file, extension => $extension }) if $log->is_debug();
+
+		my $csv_options_ref = { binary => 1 , sep_char => "\t" };	# should set binary attribute.
```
hangy: Do we want to assume that CSV and TSV all use TAB as the separator? As we want to allow producers to upload custom formats, this might vary.

lib/ProductOpener/Producers.pm
```diff
@@ -0,0 +1,531 @@
...
+	# -> use Spreadsheet::CSV only for Excel files
+	# -> use Text::CSV directly for CSV files
+
+	my $extension = $file;
+	$extension =~ s/^(.*)\.//;
+	$extension = lc($extension);
+
+	if (($extension eq "csv") or ($extension eq "tsv") or ($extension eq "txt")) {
+
+		my $encoding = "UTF-8";
```
hangy: As we want to allow producers to upload custom formats, this could vary.

lib/ProductOpener/Producers.pm
stephanegigandet: In practice most of them will send Excel files. So for CSV we can mandate tabs and Unicode. We want the platform to be as simple as possible.

lib/ProductOpener/Producers.pm
```diff
@@ -0,0 +1,531 @@
...
+				push @headers, $field;
+				$headers_cols{$field} = $col;
+			}
+		}
+
+		$col++;
+	}
+
+	$csv_out->print ($out, \@headers);
+	print $out . "\n";
```
hangy: Why the `.`? I would have expected:
```suggestion
	print $out "\n";
```

lib/ProductOpener/Producers.pm
```diff
@@ -0,0 +1,531 @@
...
+	# Output CSV product data
+
+	foreach my $row_ref (@$rows_ref) {
+		my @values = ();
+		foreach my $field (@headers) {
+			my $col = $headers_cols{$field};
+			push @values, $row_ref->[$col];
+		}
+		$csv_out->print ($out, \@values);
+		print $out . "\n";
```
hangy: Why the `.`? I would have expected:
```suggestion
	print $out "\n";
```

lib/ProductOpener/Producers.pm
hangy: OK

cgi/import_file_upload.pl
```diff
@@ -0,0 +1,240 @@
...
+			$import_files_ref = {};
+		}
+
+		$import_files_ref->{$file_id} = {
+			filename => $filename,
+			extension => $extension,
+			uploaded_t => $uploaded_t,
+		};
+
+		store("$data_root/import_files/$owner/import_files.sto", $import_files_ref);
```
hangy: Do we want to delete old uploads at some point in time?

lib/ProductOpener/Producers.pm
stephanegigandet: It's already done by import_csv_file.pl in fact:

if ($file =~ /\.(xlsx|csv|tsv)$/i) {

I was a bit aggressive, I guess I can add a few more extensions.

cgi/import_file_upload.pl
stephanegigandet: Probably at some point, but at the start at least, I prefer to keep everything, including temporary files etc. so that it's easier to debug issues. We can always have a batch process to clean things periodically.

lib/ProductOpener/Producers.pm
```diff
@@ -0,0 +1,564 @@
...
+		],
+		["nutrition"],
+		["nutrition_other"],
+		["other", ["conservation_conditions", "warning", "preparation", "recipe_idea", "recycling_instructions_to_recycle", "recycling_instructions_to_discard", "customer_service", "link"]
+		],
+	];
+
+	# Create an options array for select2
+
+	my $select2_options_sample = <<JSON
```
hangy: Why don't you just build this with `JSON:PP`/`JSON:XS`? I'd avoid building JSON manually, if possible.

stephanegigandet: This is just a sample JSON expected by select2, it's not actually used anywhere, it's more like a comment, to explain what kind of JSON structure is expected by select2. I can make it an actual comment instead.

Issue #2279, opened 2019-09-04
quantity-not-recognized for Russian quantities: мл, г, л


aleksejrs:
**Describe the bug**

When saving a product with a Russian quantity unit, "quantity-not-recognized" appears.

**To Reproduce**
1. Go to https://ru.openfoodfacts.org/cgi/product.pl?type=edit&code=4607030636597 or anyother product with a Russian quantity unit like мл, г, л, and possibly кг.
2. Click on "Сохранить" ("Save")
3. See "quantity-not-recognized" in "Quality".

**Expected behavior**
No error message, just like when "g" is used.

stephanegigandet:
Thanks for the report, I will add them.

VaiTon:
Closed by #2288

PR #2288, opened 2019-09-06
add Russian units, bug #2279

stephanegigandet:
fix issue #2279 reported by @aleksejrs


aleksejrs:
cl = кл? I guess that would be центилитр or сантилитр, though I don't think I've ever seen it.

Issue #2273, opened 2019-09-02
Negative percent display issue in Finnish nutrient comparison


stephanegigandet:
Reported by @teolemon:

https://fi.openfoodfacts.org/product/8712000052034/heineken-alcohol-free-beer-0-0

![image](https://user-images.githubusercontent.com/8158668/64128399-af483e00-cdb6-11e9-96e9-c24c0bddfbbc.png)

This is because the percent formater in Finnish uses a long − for the minus sign instead of the more common dash - 

Related code: https://github.com/openfoodfacts/openfoodfacts-server/blob/master/lib/ProductOpener/Display.pm#L8699

stephanegigandet:
It might be worth testing all languages to see if other signs are used for the minus sign.

hangy:
This needs to be localized: https://github.com/openfoodfacts/openfoodfacts-server/blob/9eda3d89ab05bfdc50bb780543c133fe2f0abbf2/lib/ProductOpener/Display.pm#L8700-L8702

The Finish CLDR uses the actual minus sign. Other languages may use additionally non-printable signs: https://github.com/patch/cldr-number-pm5/blob/249762a48c7daa1e66ad64f113c5dce807563824/lib/CLDR/Number/Data/Base.pm#L1158

We might want to check for `$cldr->minus_sign` and add `$cldr->plus_sign`, instead of manually using the values.

stephanegigandet:
In fact we have the numeric value just above in the code, so let's keep it, put the formatted version in $percent_string, and then check if the numeric value is > 0 to add the + sign.

stephanegigandet:
Assigned to @zigouras

PR #2303, opened 2019-09-10
fix for issue 2273

zigouras:
fix locale issues with minus sign in different characters for percent differences.  Use numeric compare instead of string.

**Related issues and discussion:** #2273


stephanegigandet:
Thanks!

PR #2309, opened 2019-09-11
feat: Cooking duration taxonomy generation script

teolemon:
### What
- A small script to permit generating a cooking duration taxonomy
### Part of
- #5507


stephanegigandet:
The script works, but what is the motivation for a cooking duration taxonomy? If we want to record cooking time (or in many cases cooking times as frozen products for instance have different times for microwave, oven etc.), then a numeric field would probably be more adapted.

aleene:
Can this be closed?

teolemon:
The intent was to provide a foundation for Robotoff detections.

alexgarel:
@teolemon I think we should close that PR. 
If robotoff needs to generate this vocabulary, it will do it.

PR #2310, opened 2019-09-11
Protected Designations of Origins as categories

teolemon:
Adding Protected Designations of Origins as categories


teolemon:
superseded by a newer PR

PR #2317, opened 2019-09-12
English fix

teolemon:
English fix


stephanegigandet:
at least 98 strings in en.po have been reformatted with lines breaks:

msgctxt "ingredients_analysis_disclaimer"
msgid "Cancel"	msgid ""
msgstr "Cancel"	"The analysis is based solely on the ingredients listed and does not take "
"into account processing methods."
msgstr ""
"The analysis is based solely on the ingredients listed and does not take "
"into account processing methods."

That makes the test fail...

Issue #2010, opened 2019-06-26
page_size not applied in pagination


hangy:
**Describe the bug**
The user-supplied `page_size` is not used for the pagination.

**To Reproduce**
1. Go to https://world.openfoodfacts.org/cgi/recent_changes.pl?page_size=1000
2. Click on the 2nd page
3. See that https://world.openfoodfacts.org/cgi/recent_changes.pl?&page=2 is loaded

**Expected behavior**
The page size should be used for the pagination. Therefore, https://world.openfoodfacts.org/cgi/recent_changes.pl?page_size=1000=&page=2 should be opened in this example.

AlokTakshak:
@hangy  i would like to contribute to it. Is that fine?

AlokTakshak:
@hangy  Can you point  me to the corresponding file, that would be great help

hangy:
Of course. 🙂 
The pagination is displayed in https://github.com/openfoodfacts/openfoodfacts-server/blob/b98e3419d886a5e9ef9b1e4b771133bfa7c78767/lib/ProductOpener/Display.pm#L3909
I believe the size should actually be available in the `$limit` parameter: https://github.com/openfoodfacts/openfoodfacts-server/blob/b98e3419d886a5e9ef9b1e4b771133bfa7c78767/lib/ProductOpener/Display.pm#L3912

stephanegigandet:
Hi @AlokTakshak, are you working on this?

zigouras:
@hangy you can assign this to me if you want.

zigouras:
@hangy @stephanegigandet I went to the link in the OP to the 'Recent Changes' page and I don't see any paging buttons at the bottom.  No matter what page_size I put in the URL, I don't see the navigation buttons at the bottom.  Can you please advise?

stephanegigandet:
Fixed by @zigouras and deployed a couple of weeks ago, closing.

PR #2320, opened 2019-09-12
fix for issue #2010: continue page size in recent change pagination

zigouras:
changes for issue 2010 ... page_size aka limit is not persisted through pagination method and navigation links on the recent changes page

also removed unused method canonicalize_taxonomy_2tag_link from Tags.pm

**Related issues and discussion:** #2010


stephanegigandet:
Thanks @zigouras !

PR #2337, opened 2019-09-16
Update common.pot

teolemon:
Donate to support our work instead of Embrace the mission

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -359,7 +359,7 @@ msgid "corrector"
 msgstr ""
 
 msgctxt "countries"
-msgid "Country"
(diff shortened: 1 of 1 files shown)


teolemon:
Can you reapprove to make sure that's what you really want (especially Corrections/Corrector thing)

teolemon:
@stephanegigandet

po/common/common.pot
```diff
@@ -359,7 +359,7 @@ msgid "corrector"
 msgstr ""
 
 msgctxt "countries"
-msgid "Countries where sold"
+msgid "Country"
```
```suggestion
msgid "Countries where sold"
```

Issue #2339, opened 2019-09-16
Improve ingredients parsing for vegetal oils


stephanegigandet:
"graisse végétale bio (colza)"  -> not converted to graisse végétale bio de colza by the ingredient preparsing

stephanegigandet:
https://fr.openfoodfacts.org/produit/3770002327869/galettes-de-riz-choco-noisettes-good-gout-kidz-120g

stephanegigandet:
This has been fixed and deployed.

PR #2340, opened 2019-09-16
better parsing for oils, bug #2339

stephanegigandet:
handle things like "graisse végétale bio (colza)"
bug #2339

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -2088,7 +2088,8 @@ sub preparse_ingredients_text($$) {
 "matières grasses",
 "matière grasse végétale",
 "matières grasses végétales",
-"graisses",
+"graisse",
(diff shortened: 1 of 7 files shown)


aleene:
What is marine oil?
Is this a good approach to sub-ingredients? Seems not very scalable and hidden in the code. I would prefer a solution in the ingredients taxonomy.

stephanegigandet:
@aleene : marine oil is similar to fish oil, except it may contain other things than fish i guess. I've seen it only on a few products.

Handling things like "organic vegetal oils (palm, sunflower)" and "organic vegetal oils (organic palm oil, organic sunflower oil)" is tricky and language dependent (e.g. in French "huiles végétales biologiques d'olive, de colza et palme" is quite complex), we can certainly improve it, to make it more generic, but it's not a simple task.

PR #2356, opened 2019-09-19
German translations, spelling corrections for additives

mahlzahn:
**Description:**
* add many German translations, spelling corrections for additives (mainly capitalisation)
* carotenoids, carotene, beta-carotene marked as maybe vegetarian and maybe vegan, see e.g. [here](https://www.vrg.org/blog/2012/02/15/beta-carotene-in-us-beverages-not-stabilized-with-gelatin-unlike-some-products-in-the-uk) and [there](https://www.coca-cola.co.uk/faq/ingredients/coca-cola-drinks-suitable-for-vegans-vegetarians)
* calcium carbonate marked as maybe vegetarian and maybe vegan, see e.g. [here](https://en.wikipedia.org/wiki/Calcium_carbonate#Biological_sources) and [there](https://www.vrg.org/blog/2011/12/12/calcium-carbonate-in-most-soy-rice-beverages-and-in-calcium-supplements-derived-from-a-mineral-source-not-oyster-shell-source-does-not-have-to-be-labeled)
* rename E160 in English from carotene to carotenoids, because carotene is for E160a, compare [here](https://en.wikipedia.org/wiki/E160)

### DIFF
--- a/taxonomies/additives.txt
+++ b/taxonomies/additives.txt
@@ -1938,7 +1938,7 @@ wikidata:en:Q191907
 efsa_evaluation_url:en: http://dx.doi.org/10.2903/j.efsa.2014.3492
 efsa_evaluation_date:en: 2014-01-22
 efsa_evaluation:en: Scientific Opinion on the reconsideration of the ADI and a refined exposure assessment of beta-apo-8'-carotenal (E 160e)
-# Stabiliser/carrier may not be vegetarian, can be e.g. fish gelatine, which does not need to be labeled in EU right
+# Stabilizer/carrier may not be vegetarian, can be e.g. fish gelatine, which does not need to be labeled in EU right
 # https://www.vrg.org/blog/2012/02/15/beta-carotene-in-us-beverages-not-stabilized-with-gelatin-unlike-some-products-in-the-uk
 # https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:32007L0068
 vegan:en:maybe
@@ -3011,7 +3011,7 @@ sk:E170, Uhličitan vápenatý, CI pigment biely 18
 sl:E170, Kalcijev karbonat, CI Pigment White 18
 sv:E170, Kalciumkarbonat, CI Pigment White 18
 e_number:en:170
-mandatory_additive_class:en: en:acidity-regulator, en:anti-caking-agent, en:stabiliser, en:firming-agent, en:flour-treatment-agent, en:glazing-agent, en:colour
+mandatory_additive_class:en: en:acidity-regulator, en:anti-caking-agent, en:stabilizer, en:firming-agent, en:flour-treatment-agent, en:glazing-agent, en:colour
 wikidata:en:Q23767
 efsa_evaluation_url:en:http://www.efsa.europa.eu/fr/efsajournal/doc/2318.pdf
 efsa_evaluation_date:en:2011/07/26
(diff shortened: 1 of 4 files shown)


mahlzahn:
I needed to change the test [ingredients_tags.t#L51](https://github.com/openfoodfacts/openfoodfacts-server/blob/master/t/ingredients_tags.t#L51), because the ingredient beta-carotene should actually get the tag ```en:e160a``` and not ```en:e160```.
Line in test conditions:
```
[ { lc => "fr", ingredients_text => "Céleri - rave, choux - fleurs, béta - carotène"}, [ "en:celeriac", "en:cauliflower", "en:e160" ], ],
```
Test output:
```
#   Failed test at t/ingredients_tags.t line 83.
#          $got->[2] = 'en:e160a'
#     $expected->[2] = 'en:e160'
…
```

taxonomies/additives.txt
```diff
@@ -15766,7 +15767,7 @@ bg:E901, Пчелен восък, бял и жълт, Бял восък
 ca:E901, Cera d'abelles
 cs:E901, Včelí vosk, bílý a žlutý, Bílý vosk, E 901
 da:E901, Bivoks, hvidt og gult, Hvidt voks, E-901
-de:E901, Bienenwachs; weiss und gelb, weißes Wachs, Bienenwachs, E 901, Myricin, Cera Flava
+de:E901, Bienenwachs; weiß und gelb, weißes Wachs, Bienenwachs, E 901, Myricin, Cera Flava, Bienenwachs; weiß und gelb
```
teolemon: ; vs ,

stephanegigandet: This doesn't work in fact, we cannot have commas in an additive name.

"Beeswax; white and yellow," --> Needs to be changed to "Beeswax, white beeswax, yellow beeswax, white and yellow beeswax, yellow and white beeswax".

Same for all languages.

Otherwise it makes "White and yellow" a synonym of "Beeswax".

stephanegigandet:
@mahlzahn Why are the vegan/vegetarian/palm oil properties being changed? For changes like that, we need to discuss it on Slack first. It's a complex topic, so it's best to discuss it first. Thanks!

stephanegigandet:
Ok found the links above for the changes to vegan/vegetarian properties. Could you add the links as comments in the taxonomy so that we can track easily why we put maybe instead of yes.

taxonomies/additives.txt
mahlzahn: Thanks, what about brackets? I saw things like this: `beeswax (white and yellow)`. Also sometimes things like `beeswax \, white and yellow`.

Btw, on the page for [`en:e901`](https://world.openfoodfacts.org/ingredient/e901) the link to  [`en:white-and-yellow-beeswax`](https://world.openfoodfacts.org/ingredient/white-and-yellow-beeswax) is unreachable, as it links back to `en:901`, probably because in additives.txt it is synonym to it?

stephanegigandet: same things for brackets, parenthesis etc. : they should not be used in the names and synonyms of additives, it won't work.

stephanegigandet: The E901 link is because there are also entries for E901 / beeswax in the ingredients taxonomy, we will need to remove them from the ingredients taxonomy, and add the extra translations to the additives taxonomy instead. cc @aleene

mahlzahn: 1.) Here it seems to work with the brackets: [https://world.openfoodfacts.org/additive/e285-sodium-tetraborate (borax)](https://world.openfoodfacts.org/additive/e285-sodium-tetraborate%20(borax)), but it seems it needs the proper single words still. And e.g. here it doesn't work so well: https://world.openfoodfacts.org/additive/e342-ammonium-phosphates-i-monoammonium-phosphate-ii-diammonium-phosphate
2.) About additives: is it in principle save to remove them from `ingredients.txt`? (including transferring the translations) if they have a 1:1 partner in the `additives.txt`?

Issue #2377, opened 2019-09-25
Errors in ingredients taxonomy


stephanegigandet:
There are a number of errors that are raised when building the ingredients taxonomy, mostly because some entries are synonyms of 2 different ingredients, or because ingredients are defined twice.

In particular:

- Additives, minerals and vitamins need to be removed from the ingredients taxonomy. There may be more translations and synonyms in the ingredients taxonomy, so ideally it would be good to add the new translations and synonyms in the additives, minerals and vitamins taxonomy. There will still be some overlap between additives, minerals and vitamins, but that should be more manageable.

- There are many fishes defined twice.

- There is a dozen of ingredients that specify a parent that does not exist (yet).

Full list of errors:
```
Errors in the ingredients taxonomy definition:
ERROR - es:fosfato-de-lactoflavina already is a synonym of e106 () - cannot make a synonym of e101ii for en:e101ii
ERROR - fr:riboflavine-5′-phosphate already is a synonym of e106 () - cannot make a synonym of e101ii for en:e101ii
ERROR - hu:riboflavin-5-foszfát already is a synonym of e106 () - cannot make a synonym of e101ii for en:e101ii
ERROR - nl:riboflavine-5-fosfaat already is a synonym of e106 () - cannot make a synonym of e101ii for en:e101ii
ERROR - hu:riboflavin-5-foszfát already is a synonym of e106 () - cannot make a synonym of e101a for en:e101a
ERROR - it:e101a already is a synonym of e101 (en:e101a) - cannot make a synonym of e101a for en:e101a
ERROR - fi:karmiinit already is a synonym of e100 () - cannot make a synonym of e120 for en:e120
ERROR - es:clorofilas already is a synonym of e140 () - cannot make a synonym of e140i for en:e140i
ERROR - fr:chlorophylles already is a synonym of e140 () - cannot make a synonym of e140i for en:e140i
ERROR - fr:pheophytine-au-magnesium already is a synonym of e140 () - cannot make a synonym of e140i for en:e140i
ERROR - fr:chlorophylle already is a synonym of e140 () - cannot make a synonym of e140i for en:e140i
ERROR - fr:chlorophylles-magnesium already is a synonym of e140 () - cannot make a synonym of e140i for en:e140i
ERROR - fr:ci-75810 already is a synonym of e140 () - cannot make a synonym of e140i for en:e140i
[...]

PR #2384, opened 2019-09-29
Merging additives

aleene:
**Description:**
removing additives from ingredients.txt and merging them into additives.txt

**Related issues and discussion:** #2377


taxonomies/additives.txt
```diff
@@ -2666,17 +2671,21 @@ de:E162, Betanin, Beetenrot, Betanoin, Betenrot
 el:E162, Ερυθρα χρωστικη της ριζας των τευτλων, βετανινη
 es:E162, Rojo de remolacha, Betanina, Betacianina, E 162, Rojo remolacha
 et:E162, Peedipunane, betaniin
+fa:E162, بتانین
 fi:E162, Punajuuriväri, betalaiini
-fr:E162, Rouge de betterave, bétanine, rouge betterave, E 162, 7659-95-2, C24H27N2O13
+fr:E162, Rouge de betterave, bétanine, rouge betterave, EBetterave rouge
```
stephanegigandet:
```suggestion
fr:E162, Rouge de betterave, bétanine, rouge betterave
```
"betterave rouge" means "red beetroot", it's not an additive like "rouge de betterave" (beetroot red), it's the actual vegetable.

taxonomies/additives.txt
```diff
@@ -3025,39 +3066,86 @@ organic_eu:en:authorized
...
+# additive/e171-titanium-dioxide has 4518 products in 15 languages @2019-03-27
 efsa_evaluation_url:en: http://dx.doi.org/10.2903/j.efsa.2016.4545
 efsa_evaluation_date:en: 2016-09-14
 efsa_evaluation:en: Re-evaluation of titanium dioxide (E 171) as a food additive
 additives_classes:en: en:colour
 vegan:en:yes
 vegetarian:en:yes
 
+<en:E171
+en:colorant E171
```
stephanegigandet: This should be removed, it will be handled by the parser.

taxonomies/additives.txt
```diff
@@ -3466,7 +3611,19 @@ additives_classes:en: en:preservative
 vegan:en:yes
 vegetarian:en:yes
 
+<en:E202
+en:preservative potassium sorbate, preservative E202
```
stephanegigandet: This should be removed, it will be handled by the parser.

taxonomies/additives.txt
```diff
@@ -4018,6 +4340,10 @@ organic_eu:en:authorized
 vegan:en:yes
 vegetarian:en:yes
 
+<en:E224
```
stephanegigandet: This should be removed, it will be handled by the parser.

taxonomies/additives.txt
```diff
@@ -4636,31 +5030,89 @@ organic_eu:en:authorized
 vegan:en:yes
 vegetarian:en:yes
 
+<en:E250
```
stephanegigandet: This should be removed, it will be handled by the parser.

taxonomies/additives.txt
aleene: Some additives have multiple purposes. Encoding it like this emphasises why the producer added an additive to the product. I do not know if you keep this information in the parser.

In addition it also helps if you use the taxonomy for translation, as this entry can be used for that.

taxonomies/additives.txt
aleene: These are now often found as synonyms as well.

aleene:
I added the changes. However I am not convinced by your approach. OK for recognition, but not for conveying meaning and translation.

taxonomies/additives.txt
stephanegigandet: Producers have to indicate the additive class before the additive, so there should always be something like coloring: E171, E102 and E120. During parsing, we keep track of the additive class, so that we can associate it not only with E171, but also the additives after.

PR #2414, opened 2019-10-09
Vegetarian icons + ingredients taxonomy update

stephanegigandet:
Needed for Android app.


hangy:
> Needed for Android app.

What's the reason you want to add the icons to this repository (and probably openfoodfacts.org), if they're required for the Android app? As far as I know, it would be a more common approach to [embed](https://stackoverflow.com/a/39266840/11963) the SVGs in the apk.

stephanegigandet:
@hangy : We woud like to be able to add new ingredient analysis related icons in the app without having to release a new version of the app. The app can periodically reload taxonomies and show new icons.

hangy:
OK, that's file. However, IMHO that doesn't require the icons to be in the server's repository. Anyone new who wants to do something on the server will see this here and wonder why there are unused images. From my understanding, the app just requires some website to host these static files - and that isn't really connected to the Perl server's logic?

stephanegigandet:
It's a bit connected: the files depend on the ingredient analysis and the taxonomies. They are more connected to the app than to the server, but I think it is useful that they are tracked in this github repository, so that we can link the taxonomy changes to the icons changes.

PR #2418, opened 2019-10-12
Vegetables

moon-rabbitOFF:
Huge changes in vegetables

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -3012,7 +3012,7 @@ it:Succo di barbabietola
 nl:Rode bietensappen, Rode bietensap
 
 <en:Vegetable juices
-<en:Tomatoes and tomato products
+<en:Tomatoes and their products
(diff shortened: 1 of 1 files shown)


taxonomies/categories.txt
```diff
@@ -3031,7 +3031,7 @@ es:Zumos de tomate procedente de concentrado
 fr:Jus de tomates à base de concentré, jus de tomate à base de concentré
 nl:Tomatensappen uit concentraat, Tomatensap uit concentraat
 
-<en:Tomatoes and tomato products
+<en:Tomatoes and their products
```
VaiTon: What about "tomato based products"?

moon-rabbitOFF: I change the name of this category to aligne it on the current usage. We could change it, but it will requiere to change the name of numerous categories. Maybe we should discuss about it with the people who create those categories, I don't know who it is

Issue #1738, opened 2019-04-15
Enable Crowdin in context translations


teolemon:
When the Crowdin In-Context tool is enabled, the archive with your project translation will include one additional language. It is required to store operational data about localization strings. Integrate this pseudo-language to your app along with other languages you translate to. You will have to switch your application to this language as one of the requirements to enable In-Context localization.
lol/common.po
``` 
<script type="text/javascript">
  var _jipt = [];
  _jipt.push(['project', 'openfoodfacts']);
</script>
<script type="text/javascript" src="//cdn.crowdin.com/jipt/jipt.js"></script>
``` 
It's important to add the Crowdin In-Context script before any other scripts, at the <head> section of your document. Make sure this script is included on every page with localizable text. Optionally, you can include this JavaScript snippet only when the pseudo-language is active.

teolemon:
#2421

PR #2421, opened 2019-10-12
feat: Add in-context translation (solves #1738)

teolemon:
### What
https://support.crowdin.com/in-context-localization/#optional-parameters
To test before deploying widely. This will only work with a special locale called `lol`
https://github.com/openfoodfacts/openfoodfacts-server/blob/master/po/common/lol.po

### Part of
- #5528


hangy:
I tried adding the JavaScript with Chrome DevTools, but even after logging in, it doesn't display any translations for me. image I suppose there would need some way to let Crowdin know which text is which, as the internal ID is different from the displayed text (which is why we have a custom version of gettext). Also, we shouldn't always include that script, since it immediately forces the Crowdin login dialog on the user: 
image

hangy (changes requested):
- See https://github.com/openfoodfacts/openfoodfacts-server/pull/2421#issuecomment-541399738

teolemon:
My understanding is that they have this pseudo language called `lolcat`. So the user would need to switch the website to lolcat, and then indicate the language.
Something like
`fr-lolcat.openfoodfacts.org` would possibly work.
`translate-fr` or `fr-translate` would be more intuitive

teolemon:
@hangy you can try renaming lol.po to de.po https://github.com/openfoodfacts/openfoodfacts-server/blob/master/po/common/lol.po

Issue #2419, opened 2019-10-12
Add data-quality taxonomy and split quality facet in data-quality-info/warnings/errors facets


stephanegigandet:
The /quality facet currently mixes different things:

1. info tags that do not indicate an error (e.g. Carbon-footprint-from-known-ingredients-but-not-from-meat-or-fish) or that indicate that there isn't an error (e.g. Nutrition-grade-fr-producer-same-ok)

2. warning tags that indicate an error may be possible (e.g. Ingredients-50-percent-unknown)

3. error tags that indicate an error is very likely (if not certain). (e.g. Nutrition-value-over-105-for-salt)

For the producers platform, but also for the public database, it would be very useful to distinguish between what is information, a warning, or an error.

e.g. on the producers platform, if there are errors for some products, we won't synchronize them to the public database.

Proposed approach:

1. create a data_quality.txt taxonomy

2. add properties to indicate what is an info, warning or error.
e.g.
severity:en:warning

3. populate 3 different facets, based on the severity:
- data-quality-info
- data-quality-warnings
- data-quality-errors

The taxonomy will be the same file for all 3 sub-facets. It will allow us to translate the info / warnings / errors, and also to provide descriptions to explain some of them.

There will also be a property to indicate which quality entries will be shown to producers on the producer platform (e.g. the "% of unknown ingredients" entries are more likely to indicate issues with the ingredients taxonomy or parsing than issues in the ingredients text provided by producers).

CharlesNepote:
This very interesting!

For the feature part, there could also a distinction between important warnings/errors and less important ones. For example:
* https://world.openfoodfacts.org/quality/quantity-contains-e is not very important
* https://world.openfoodfacts.org/quality/ingredients-ingredient-tag-length-greater-than-200 is more important as it impacts Nova, etc.

stephanegigandet:
2. add properties to indicate what is an info, warning or error.  e.g.  severity:en:warning

-> This isn't such a good idea, as many data quality values may not be in the taxonomy (e.g. all the combinations of ingredients issues * langue like ingredients_fr_* )

The severity will be set directly in the code.

zigouras:
Do you guys want me to work on this?
Why not create a mongodb table instead of a file.  It will make updating and querying much easier.
Will all of the taxonomy tags have to be extracted from the Perl code for this?

stephanegigandet:
[...]

PR #2425, opened 2019-10-13
Data quality reorganization, taxonomy and sub-facets

stephanegigandet:
This pull request has 3 changes:

1. SiteQuality_off.pm has been split in several modules: DataQuality.pm, DataQualityCommon.pm and DataQualityFood.pm. The new modules have some perlpod documentation. This also removes the need for a symlink from SiteQuality.pm to SiteQuality_off.pm

2. There is a start of a data quality taxonomy. This will be useful in particular to surface quality issues to users (e.g. in the product page for logged in users, or in the product edit form) and to manufacturers (on the producers platform).

3. There are now different levels of issues: bugs, info, warnings and errors. (see bug #2419 )


po/common/common.pot
```diff
@@ -3688,3 +3688,63 @@ msgstr "Cancel"
...
+msgid "data quality errors"
+msgstr "data quality errors"
+
+msgctxt "data_quality_errors_s"
+msgid "data quality error"
+msgstr "data quality error"
+
+msgctxt "data_quality_errors"
+msgid "data quality errors"
+msgstr "data quality erros"
```
CharlesNepote: Small typo: s/erros/errors/

po/common/en.po
```diff
@@ -3564,3 +3564,63 @@ msgstr "Export product data and photos to the public database"
...
+msgid "data quality errors"
+msgstr "data quality errors"
+
+msgctxt "data_quality_errors_s"
+msgid "data quality error"
+msgstr "data quality error"
+
+msgctxt "data_quality_errors"
+msgid "data quality errors"
+msgstr "data quality erros"
```
CharlesNepote: Small typo: s/erros/errors/

po/common/common.pot
stephanegigandet:
```suggestion
msgstr "data quality errors"
```

stephanegigandet: Fixed, thank you!
BTW did you know that you can make suggestions like that in github? Like in a google doc suggestion, they can be accepted in one click.
image

lib/ProductOpener/DataQuality.pm
```diff
@@ -0,0 +1,150 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+=head1 NAME
+
+ProductOpener::DataQuality - check the quality of data for products
```
CharlesNepote: It does not only check quality but it also stores quality tags inside product data isn't it? Why not separate checking and storing data?

po/common/common.pot
CharlesNepote: Oh my!... thanks @stephanegigandet !

lib/ProductOpener/DataQuality.pm
stephanegigandet: Yes it does store the results of the quality checks as quality tags. But that's the best way to do it I think, having an extra layer between results and storing the results would only add code and complexity. It is also the way all other processing on product data work: we pass a reference to the product object, and it is enriched with new analysis etc.

PR #2431, opened 2019-10-14
ci: Automatically rebuild taxonomies

hangy:
**Description:** If a push to `master` (ie. a merged pull request) updates `taxonomies/*.txt*`, try to run `refresh_taxonomies.js` to check if the taxonomies changed content-wise from the previous state, without the `.sto` files being updated in the same push.

**Related issues and discussion:** N/A

### DIFF
--- a/.github/workflows/update_taxonomies.yml
+++ b/.github/workflows/update_taxonomies.yml
@@ -1,0 +1,66 @@
+name: Update Taxonomies
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'taxonomies/*.txt'
+      - '!taxonomies/*.result.txt'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-node@v1
+      with:
+        node-version: '10.x'
+    - name: prepare dependencies and rebuild taxonomies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y --no-install-recommends cpanminus tesseract-ocr
+        # apt packages from cpanfile
+        sudo apt-get install -y --no-install-recommends graphviz \
+        libwww-perl \
+        libimage-magick-perl \
+        libxml-encoding-perl \
+        libtext-unaccent-perl \
+        libmime-lite-perl \
+        libcache-memcached-fast-perl \
+        libjson-pp-perl \
+        libclone-perl \
+        libcrypt-passwdmd5-perl \
+        libencode-detect-perl \
+        libgraphics-color-perl \
+        libbarcode-zbar-perl \
+        libxml-feedpp-perl \
+        liburi-find-perl \
+        libxml-simple-perl \
+        libexperimental-perl \
+        libapache2-request-perl \
+        libdigest-md5-perl \
+        libtime-local-perl
+        # setup local::lib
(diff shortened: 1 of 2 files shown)


hangy:
This would avoid us having to manually update the taxonomies by automatically creating a pull request with regenerated `.sto` files.

We may want to move the `.sto` files to Git LFS before activating this, because high frequency updates of large(r) binary files is not very good when using git.

stephanegigandet:
> This would avoid us having to manually update the taxonomies by automatically creating a pull request with regenerated `.sto` files.

That would be very useful.

VaiTon:
@hangy is this ready for review? It would be very useful.

hangy:
> > This would avoid us having to manually update the taxonomies by automatically creating a pull request with regenerated `.sto` files.
> 
> That would be very useful.

> @hangy is this ready for review? It would be very useful.

What do you think about a potential issue with regularly changing binary files (possibly polluting this git repository too much)?

> We may want to move the `.sto` files to Git LFS before activating this, because high frequency updates of large(r) binary files is not very good when using git.

VaiTon:
In my opinion `.sto` files (or if we adopt `.json`) should be built only on the production/dev server. There is no reason to keep them in git if we can just run a script and rebuild them before deployment.

@hangy @stephanegigandet

hangy:
> In my opinion `.sto` files (or if we adopt `.json`) should be built only on the production/dev server. There is no reason to keep them in git if we can just run a script and rebuild them before deployment.

I generally agree. One problem stopping us from that at this moment, is that the compiled taxonomies are required for unit tests, and the unit tests do not (yet) use static, unchanging taxonomies.

stephanegigandet:
> unit tests do not (yet) use static, unchanging taxonomies.

Some of those unit tests are actually testing the taxonomy changes themselves, to make sure there aren't unintended changes in the taxonomies that affect many products. That's especially true for ingredients and additives.

VaiTon:
> 
> 
> > In my opinion `.sto` files (or if we adopt `.json`) should be built only on the production/dev server. There is no reason to keep them in git if we can just run a script and rebuild them before deployment.
> 
> I generally agree. One problem stopping us from that at this moment, is that the compiled taxonomies are required for unit tests, and the unit tests do not (yet) use static, unchanging taxonomies.

We could build them on unit testing, as we install npm modules. It would mean more testing time but if we optimize the build process it shouldn't be so much.

teolemon:
Is it still WIP ? It looks harmless if we merge @hangy ? @stephanegigandet

hangy:
> Is it still WIP ? It looks harmless if we merge @hangy ? @stephanegigandet 

Not necessarily harmless. Every build causes several MiBs of bloat, because of the binary format of the taxonomies. I'm inclined to close this and only actually do this if the taxonomies were stored elsewhere …

teolemon:
@hangy @stephanegigandet to solve the issue, we could only trigger it on a branch called `rebuild-taxonomies`, that we would merge master into, only on a schedule (each week, each month).

.github/workflows/update_taxonomies.yml
```diff
@@ -0,0 +1,65 @@
+name: Update Taxonomies
+
+on:
+  push:
+    branches:
+      - master
```
teolemon, line 6:
```suggestion
      - rebuild-taxonomies
```

teolemon: @stephanegigandet @hangy

Issue #2442, opened 2019-10-18
Keep obsolete products in db, put them last, grey them out, remove them from stats


stephanegigandet:
Issue #1592 introduced the possibility for admins to mark products as not being sold anymore ("obsolete"). The products were completely removed from MongoDB, but kept in the file system so that we could show them if they were scanned.

This issue is to improve the handling of obsolete products:

1. Brand owners will be able to mark products as obsolete
2. Obsolete products will be kept in MongoDB, but:
3. The sort order will put them last
4. They will be grayed out in the product list
5. Obsolete products won't be taken into account in category stats

stephanegigandet:
This has been deployed.

Issue #1592, opened 2018-12-26
Support to mark old products that are no longer sold obsolete, remove them from searches


stephanegigandet:
Add an "obsolete" check box in the product edit form, for admins, and an obsolete_since_date

Obsolete products will be removed from the database, but they can still be found by barcode.

hangy:
Done in production

PR #2443, opened 2019-10-18
better support for obsolete products, bug #2442

stephanegigandet:
Issue #1592 introduced the possibility for admins to mark products as not being sold anymore ("obsolete"). The products were completely removed from MongoDB, but kept in the file system so that we could show them if they were scanned.

This issue is to improve the handling of obsolete products:

1. Brand owners will be able to mark products as obsolete
2. Obsolete products will be kept in MongoDB, but:
3. The sort order will put them last
4. They will be grayed out in the product list
5. Obsolete products won't be taken into account in category stats

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -3258,7 +3258,11 @@ msgid "Withdrawal date"
 msgstr ""
 
 msgctxt "obsolete_since_date_note"
-msgid "Format: YYYY or YYYY-MM"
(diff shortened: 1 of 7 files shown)


po/common/common.pot
```diff
@@ -3747,4 +3751,4 @@ msgstr "data quality error"
 
 msgctxt "data_quality_errors"
 msgid "data quality errors"
-msgstr "data quality errors"
+msgstr "data quality erros"
```
```suggestion
msgstr "data quality errors"
```

PR #2444, opened 2019-10-18
Fix install of DBD::Pg during GitHub Actions CI

hangy:
**Description:**

Install the `libdbd-pg-perl` package during GitHub Actions PR validation, so that the following error does not occur any more

>  Installing the dependencies failed: Module 'Mojo::Pg' is not installed

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -73,7 +73,8 @@ jobs:
         libexperimental-perl \
         libapache2-request-perl \
         libdigest-md5-perl \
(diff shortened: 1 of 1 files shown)


hangy:
Apparently, I'm missing more libraries. Let's keep this PR open until I figure out what's missing for DBD:Pg to build in this case. 🙂

stephanegigandet:
Might be libpq-dev, trying: https://github.com/openfoodfacts/openfoodfacts-server/pull/2445

PR #2466, opened 2019-10-22
[WIP] donate-to-openfoodfacts.html

teolemon:
[WIP] donate-to-openfoodfacts.html


stephanegigandet:
@teolemon : the meta description etc. is in French, it needs to be updated.

stephanegigandet:
@teolemon The files in srv/off/html/donate/ should just be in html/donate

teolemon:
done @stephanegigandet

html/donate/en.html
```diff
@@ -0,0 +1,98 @@
+
+<!doctype html>
+<html class="no-js" lang="fr">
```
hangy: The `lang` is incorrect

html/donate/en.html
```diff
@@ -0,0 +1,98 @@
...
+<meta charset="utf-8">
+
+<title>Donate to Open Food Facts</title>
+
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+
+<meta property="fb:app_id" content="219331381518041">
+<meta property="og:type" content="food">
+<meta property="og:title" content="">
+<meta property="og:url" content="https://world.openfoodfacts.org/donate-to-openfoodfacts">
```
hangy: This URL is 404

html/donate/en.html
```diff
@@ -0,0 +1,98 @@
...
+<link rel="mask-icon" href="https://static.openfoodfacts.org/images/favicon/safari-pinned-tab.svg" color="#5bbad5">
+<link rel="shortcut icon" href="https://static.openfoodfacts.org/images/favicon/favicon.ico">
+<meta name="msapplication-TileColor" content="#da532c">
+<meta name="msapplication-TileImage" content="https://static.openfoodfacts.org/images/favicon/mstile-144x144.png">
+<meta name="msapplication-config" content="https://static.openfoodfacts.org/images/favicon/browserconfig.xml">
+<meta name="_globalsign-domain-verification" content="2ku73dDL0bAPTj_s1aylm6vxvrBZFK59SfbH_RdUya" />
+<meta name="apple-itunes-app" content="app-id=588797948">
+<meta name="flattr:id" content="dw637l">
+
+<link rel="stylesheet" href="https://static.openfoodfacts.org/css/dist/app.css?v=1562940554">
```
hangy: File shouldn't link the version, as it might be an incorrect assumption.

html/donate/en.html
```diff
@@ -0,0 +1,98 @@
...
+<link rel="stylesheet" href="https://static.openfoodfacts.org/css/dist/jqueryui/themes/base/jquery-ui.min.css">
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.3/css/select2.min.css" integrity="sha384-HIipfSYbpCkh5/1V87AWAeR5SUrNiewznrUrtNz1ux4uneLhsAKzv/0FnMbj3m6g" crossorigin="anonymous">
+<link rel="search" href="https://fr.openfoodfacts.org/cgi/opensearch.pl" type="application/opensearchdescription+xml" title="Open Food Facts">
```
hangy: `fr.openfoodfacts.org` is the wrong subdomain for the English file

html/donate/en.html
```diff
@@ -0,0 +1,98 @@
...
+<p>When we founded Open Food Facts, we made the clear choice to be totally non-profit to maximize our impact by giving away all our work. We also made the clear choice that we would be totally independent from the industry, and thus to never accept any funding that could put our mission at risk.</p>
+<p>As a result, we started the project on a shoestring, and we still keep our budget tight. From 2012 to 2017 we were all volunteers and our annual budget was less than 1000 euros! The project has grown a lot however, and we now need a bit more to continue our mission.</p>
+<p>We need your help to continue to operate and grow Open Food Facts.Please consider making a donation to protect and sustain Open Food Facts.</p>
+<p>Thank you,</p>
+<p><b>Anca, Florence, Léonore, Sylvia, Charles, Christian, Ludovic, Pierre, Sébastien, Stéphane, Vincent and the whole Open Food Facts team</b></p>
+</div>
+
+<div class="columns large-5 xlarge-4 xxlarge-3 text-center">
+  <center>
```
hangy: The `center` tag has been [deprecated](https://developer.mozilla.org/de/docs/Web/HTML/Element/center) long ago

VaiTon:
> 
> 
> Why does `html/donate/en.html` contain the whole header/footer HTML that surrounds the site? When embedded through `ProductOpener`, the software should generate all of that (correctly and up-to-date) through `ProductOpener::Display`.

Does the software automatically add the wrapping in current production for html pages?

html/donate/en.html
VaiTon: I think that's because it hasn't been deployed to production yet.

html/donate/en.html
VaiTon: See #2538

html/donate/en.html
VaiTon: See #2538

html/donate/en.html
VaiTon: See #2538

html/donate/en.html
VaiTon: See #2538

hangy:
> > Why does `html/donate/en.html` contain the whole header/footer HTML that surrounds the site? When embedded through `ProductOpener`, the software should generate all of that (correctly and up-to-date) through `ProductOpener::Display`.
> 
> Does the software automatically add the wrapping in current production for html pages?

Yes, see the layout of https://github.com/openfoodfacts/openfoodfacts-server/blob/19c1615a6fa21e3a497083914e5bbf897443ae04/lang/en/texts/donate.html. Adding all of the HTML+JS boilerplate in the translatable files is prone to errors. I'm not quite sure which different approach @teolemon intends to use here.

teolemon:
Crowdin allows us to translate HTML or PO files. 
With po files, invalidating a sentence means translations to do for that sentence.
With HTML, reformating line wrappings often means translations to start over for the whole file. 
What we could do is break it into a JS+CSS file, plus an include for the content.

teolemon:
<html>
[[title-include.html]]
hopefully some non language dependent JS and meta tags (but we all know that's not true)
[[content-include.html]]
</html>

VaiTon:
> 
> 
> ```
> <html>
> <head>
> hopefully some non language dependent JS and meta tags (but we all know that's not true)
> </head>
> <body>
> [[content-include.html]]
> </body>
> </html>
> ```
@teolemon if i understand correctly all of this is added by ProductOpener. It's basically useless to re-add all the basic scheme.

hangy:
> Crowdin allows us to translate HTML or PO files.
> With po files, invalidating a sentence means translations to do for that sentence.
> With HTML, reformating line wrappings often means translations to start over for the whole file.
> What we could do is break it into a JS+CSS file, plus an include for the content.

> ```
> <html>
> <head>
> hopefully some non language dependent JS and meta tags (but we all know that's not true)
> </head>
> <body>
> [[content-include.html]]
> </body>
> </html>
> ```

I just noticed that this `donate.html` is missing most of the basic OFF menus and functionality. Could you explain the use-case in a few words? From my (new) understanding, the idea would be to have a really stripped-down donation page that doesn't contain the usual header/side+top nav/footer? I actually wouldn't strip it down that much, because the donation page would need to link to some legal terms. Maybe we could just add an additional display mode, so that the side + top nav could be hidden to produce a full-width view for donations? Using that, `ProductOpener::Display` could still be responsible for generating correct links to subdomains, common CSS+JS (Foundation?), add analytics, and so on.

PR #2469, opened 2019-10-22
Creams (again)

moon-rabbitOFF:


taxonomies/categories.txt
```diff
@@ -37352,7 +37358,7 @@ fr:Tomme du Massif central, Tommes du Massif central
 fr:Tomme des Pyrénées, Tommes des Pyrénées
 country:en:France
 
-<en:Plant-based foods
+<en:Vegetables based foods
```
VaiTon: Are we sure we want to use the word "vegetables"? Does it contains all plant-based foods?

moon-rabbitOFF: As mushrooms are generally considered as a vegetable, I decided to move it from the category  "Plant-based food" to the category "Vegetables based foods" (which is included in the first one). I followed the categorisation we choose for Tomatoes, which are fruits from a botanical point of view but are generally considered as vegetables, and are therefore categorised in "Vegetables based foods". Does it answers to your question? Because I am not sure to have understand it well

moon-rabbitOFF: Mushrooms are also considered has vegetables for the calculation of the nutriscore

moon-rabbitOFF: As mushrooms are generally considered as a vegetable, I decided to move it from the category "Plant-based food" to the category "Vegetables based foods" (which is included in the first one). I followed the categorisation we choose for Tomatoes, which are fruits from a botanical point of view but are generally considered as vegetables, and are therefore categorised in "Vegetables based foods". Does it answers to your question? Because I am not sure to have understand it wel

moon-rabbitOFF: Mushrooms are also considered has vegetables for the calculation of the nutriscore

Issue #2474, opened 2019-10-23
Big numbers localized


CharlesNepote:
1009992 is hard to read (since we have more than one million products).

Big numbers should be split with locale separators. Here is an example for french locale (where groups of 3 digits are separated by a space):
![image](https://user-images.githubusercontent.com/562352/67378848-2fee1400-f588-11e9-97a6-2bbb8fc3d496.png)

The current HTML code is:
```
<div style="font-size:1.2rem;background-color:#eeeeee;padding:0.3rem 1rem;height:2.75rem;margin:0">1 009 392 products</div>
```
It should be improved: put CSS in a file, add ids, etc. Example:
```
<div id="xxxProducts"><span id="nbOfProducts" value="1009392">1 009 392</span> products</div>
```

VaiTon:
We could mark fields that need to be formatted with a class and then use the `Intl` class in javascript (available for most browsers).

Ex:
```js
var num = "125452.22"
Intl.NumberFormat('it-IT').format(num); // 125.452,22
Intl.NumberFormat('fr-FR').format(num) // 125 452,22
```

CharlesNepote:
Thanks a lot @VaiTon!

PR #2483, opened 2019-10-25
Made number of products localized

VaiTon:
Fixes issue #2474

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -3802,7 +3802,7 @@ sub search_and_display_products($$$$$) {
 			$html_count .= lang("1_product");
 		}
 		elsif ($count > 1) {
-			$html_count .= sprintf(lang("n_products"), $count) ;
+			$html_count .= sprintf(lang("n_products"), "<span class='formatable'>$count</span>") ;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -3851,11 +3851,21 @@ sub search_and_display_products($$$$$) {
...
+	</div>
+</li>
 <li>
 <button href="#" data-dropdown="drop1" aria-controls="drop1" aria-expanded="false" class="button dropdown small">$Lang{explore_products_by}{$lc}</button>
 <ul id="drop1" data-dropdown-content class="f-dropdown" aria-hidden="true">
 HTML
+;
+			$initjs .= <<JAVASCRIPT
+\$(".formatable").each(function(index, element) {
+	element.innerHTML = Intl.NumberFormat(navigator.language).format(element.innerHTML);
```
stephanegigandet: What happens if we are on an old browser without Intl.NumberFormat support?

stephanegigandet:
I wonder if it wouldn't be best to do it on the server side, as we already do for floating point numbers using https://metacpan.org/pod/CLDR::Number 
cc @hangy

lib/ProductOpener/Display.pm
VaiTon: Error thrown in console and then nothing changes. I could encapsulate that in a try statement.

PR #2486, opened 2019-10-26
Update sendinblue.html

teolemon:


lang/en/texts/sendinblue.html
```diff
@@ -129,7 +129,7 @@
           <div class="sib-form-block" style=" font-size:14px; text-align:center; font-family:&quot;Helvetica&quot;, sans-serif; color:#333; background-color:transparent; border-width:px;">
             <div class="sib-text-form-block">
               <p>
-                <a href="https://world.openfoodfacts.dev/legal" target="_blank">Terms and Privacy</a>
+                <a href="https://world.openfoodfacts.org/legal" target="_blank">Terms and Privacy</a>
```
hangy: Is this HTML available on other domains?
```suggestion
                <a href="/legal" target="_blank">Terms and Privacy</a>
```

lang/en/texts/sendinblue.html
```diff
@@ -169,4 +169,4 @@
 </script>
 <script src="https://www.google.com/recaptcha/api.js?hl=fr"></script>
```
hangy: We may not want to enforce the language in here. Google should be pretty good at using the best language for the user.
```suggestion
<script src="https://www.google.com/recaptcha/api.js"></script>
```

Issue #2510, opened 2019-10-31
Allow creating products without barcodes through the API


stephanegigandet:
On the OFF web site, there is currently a button to add products without a barcode. In that case we assign one, starting from 2000000000001.
We increment the number by 1 for each product (which means codes are not valid as the last digit is supposed to be the check digit), and check if there is already a product for that number.

This feature is to allow to create products without barcodes with the same mechanism, through the API.

It does not address the issues with the possible conflicts when using numbers in the 2 range. That would be another issue to think about and solve.

stephanegigandet:
Done.

PR #2511, opened 2019-10-31
allow creating products without barcodes through the API, bug #2510

stephanegigandet:
On the OFF web site, there is currently a button to add products without a barcode. In that case we assign one, starting from 2000000000001.
We increment the number by 1 for each product (which means codes are not valid as the last digit is supposed to be the check digit), and check if there is already a product for that number.

This feature is to allow to create products without barcodes with the same mechanism, through the API.

It does not address the issues with the possible conflicts when using numbers in the 2 range. That would be another issue to think about and solve.

 This feature is needed by another project that wants to put farm products without barcodes on OFF


cgi/product_jqm_multilingual.pl
```diff
@@ -59,11 +59,38 @@
...
+
+	$product_id = product_id_for_user($User_id, $Org_id, $code);
+
+	while (-e ("$data_root/products/" . product_path_from_id($product_id))) {
+
+		$code++;
+		$product_id = product_id_for_user($User_id, $Org_id, $code);
+	}
+
+	store("$data_root/products/internal_code.sto", \$code);
```
hangy: This seems to be the same as https://github.com/openfoodfacts/openfoodfacts-server/blob/0d871ca02bfe9aa93bcd6816ff85bb5aa090fd0f/cgi/product_multilingual.pl#L90-L106. Wouldn't it make sense to refactor it to a method in `ProductOpener::Products`?

stephanegigandet: @hangy : I just refactored it, thanks!

Issue #2506, opened 2019-10-30
Build failing because of rake


teolemon:
``` 
Build system information
0.02s0.01s0.38s0.25s0.06s0.00s0.05s0.00s0.01s0.01s0.01s0.01s0.01s0.00s0.00s0.03s0.00s0.01s0.33s0.00s0.00s0.00s0.01s0.00s0.12s0.01s0.91s0.00s0.12s3.07s0.00s2.79s0.00s2.34s
docker_mtu
resolvconf
git.checkout
36.27s$ git clone --depth=50 https://github.com/openfoodfacts/openfoodfacts-server.git openfoodfacts/openfoodfacts-server
git.submodule
0.04s$ git submodule update --init --recursive
Setting environment variables from repository settings
$ export FOSSA_API_TOKEN=[secure]
rvm
0.79s$ rvm use default
ruby.versions
$ ruby --version
No Gemfile found, skipping bundle install
0.17s$ rake
rake aborted!
No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)
/home/travis/.rvm/gems/ruby-2.5.3@global/gems/rake-12.3.2/exe/rake:27:in `<top (required)>'
/home/travis/.rvm/gems/ruby-2.5.3@global/bin/ruby_executable_hooks:24:in `eval'
/home/travis/.rvm/gems/ruby-2.5.3@global/bin/ruby_executable_hooks:24:in `<main>'
(See full trace by running task with --trace)
The command "rake" exited with 1.
Done. Your build exited with 1.
```

stephanegigandet:
I can't figure that one out. Any idea @hangy ?

stephanegigandet:
Probably a syntax error in .travis.yml in fact:
https://github.com/travis-ci/travis-ci/issues/9150

stephanegigandet:
the online validator https://config.travis-ci.com/explore does not seem to work...

hangy:
[...]

PR #2513, opened 2019-10-31
ci: Build Travis CI as a generic script

hangy:
**Description:**
If no language is defined, Travis CI falls back to Ruby. For some reason, it now tries to run `rake`, which fails because we have no Ruby files. Since we use Perl and node.js in our CI build, using the [`generic`](https://docs.travis-ci.com/user/languages/minimal-and-generic/) language seems like a good alternative.

**Related issues and discussion:** #2506

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+os: linux
+language: generic
 dist: bionic
 branches:
   except:


hangy:
The "generic" language seems to build fine [on my fork](https://travis-ci.com/hangy/openfoodfacts-server/builds/134465719).

The error
> ERROR: This build could not be run. Reason: app_error

is could thus be related to the "Partial System Outage" mentioned at https://www.traviscistatus.com/ right now.

PR #2520, opened 2019-11-01
removing lines that make travis-lint not respond

stephanegigandet:
Just checking. For some reason the slack notification makes https://config.travis-ci.com/explore not work.


stephanegigandet:
OK, so the tests are passing, I don't know why, but it does appear that the slack notifications are now making travis fail (not just the tests, but even the online travis-linet

notifications:	
  slack: openfoodfacts:Pre9ZXKFH1CYtix8DeJAaFi2

PR #2539, opened 2019-11-03
Pointers for newcomers

teolemon:
Adding proposed roadmap items and easy bugs so that new comers have a sense of priorities and feel less intimidated

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,13 @@ Data about food is of public interest and has to be open. The complete database 
 
 * <https://world.openfoodfacts.org>
 
+## Roadmap
+* Extracting HTML templates from Perl code
+* Creating an API product route that can also answer for cosmetics using the Open Beauty Facts database
+
+## Easy issues to get started
+* 5 minutes-fixes: https://github.com/openfoodfacts/openfoodfacts-server/labels/5%20minute%20fix
+
 ## Libraries used
 
 ### How do I get set up?


README.md
```diff
@@ -33,6 +33,13 @@ Data about food is of public interest and has to be open. The complete database
 
 * <https://world.openfoodfacts.org>
 
+## Roadmap
+* Extracting HTML templates from Perl code
```
stephanegigandet: This should not be done by a newcomer, this is something we need to discuss and we need to agree on which templating system to use. We don't want someone to just come in and start to use one.

README.md
```diff
@@ -33,6 +33,13 @@ Data about food is of public interest and has to be open. The complete database
...
+* Creating an API product route that can also answer for cosmetics using the Open Beauty Facts database
```
stephanegigandet: That probably should be done somewhere else than in Product Opener. Either in the app, or through a proxy.

teolemon: can you propose a roadmap for outsiders ?

stephanegigandet: Realistically, newcomers need to start with small things, small bugs etc. to get familiar with the code base. So things like the 5-minutes fix issues is probably a better way.

CloCkWeRX:
Thoughts on some quick HOWTOs like:
 * Add a new category (places in code, considerations, translations, wikidata links)
 * Add a new brand to recognize (minimum threshold, links to robotoff repo)
 * Report a data problem -> slack
?

CloCkWeRX:
For the actual roadmap, is it better to simply link to a specific milestone/project here?

ie: https://github.com/openfoodfacts/openfoodfacts-server/projects/17

That way, it doesn't have to be continually maintained in the README

teolemon:
OK. I'm closing this one for lack of popular support, but we need to make development more accessible in this repo.

PR #2541, opened 2019-11-03
turn .sto files to json

stephanegigandet:


cquest:
The idea is to store one product version in json format on each line (ndjson)
First line: first version... last line: last (current) version

Benefits:
- language agnostic storage
- much less files on disk (30M inodes on /srv !)
- much easier to rsync
- easy to get last version (last line, no more symlink)
- highly compressible (repeated json lines get a typical 80% compression using lz4 used for example by ZFS)
- can be store with less hierarchy: aaa/bbb/ccc/aaabbbcccdddd.json <- less inodes again

Drawbacks:
- high number of version generates large files (no so many products in that case)

PR #2544, opened 2019-11-04
Renamed compile_cgi.js to check_cgi.js

VaiTon:


stephanegigandet:
Looks good to me. Maybe a better name would be even check_perl_programs.js instead of check_cgi.js : the perl scripts in /scripts are not cgi, and a cgi can be in any language, not just perl.

VaiTon:
> 
> 
> I don't know. Compile should actually be correct: https://www.perl.com/pub/2004/08/09/commandline.html/
> 
> > The first of these is `-c`. This option compiles your program without running it. This is a great way to ensure that you haven’t introduced any syntax errors while you’ve been editing a program.

Yeah it compiles the file, but it doesn't create an output. That's why I think 'check' would be a more suited name.

hangy:
> > I don't know. Compile should actually be correct: https://www.perl.com/pub/2004/08/09/commandline.html/
> > > The first of these is `-c`. This option compiles your program without running it. This is a great way to ensure that you haven’t introduced any syntax errors while you’ve been editing a program.
> 
> Yeah it compiles the file, but it doesn't create an output. That's why I think 'check' would be a more suited name.

I like both. 😉 However, if we want to nitpick here, the file should possibly be renamed to `check_perl.js` instead, because the files in the `scripts` directory actually aren't available as CGI scripts.

PR #2547, opened 2019-11-04
Friendlier user interface for producers platform

stephanegigandet:


cgi/user.pl
```diff
@@ -140,8 +140,12 @@
 	$html .= lang($type . $dialog);
 
 	if (($type eq 'add') or ($type eq 'edit')) {
```
VaiTon: Can't we merge nested ifs?

stephanegigandet: We could, but it seemed more logic to do it that way, as we might want to display something else on the producer platform, and add an else. It doesn't change much I think.

Issue #2563, opened 2019-11-08
Big memory leak when adding / updating / importing products - Needs investigation


stephanegigandet:
There is a big memory leak when we update a product. It can be seen in particular when we launch the import_csv_file.pl script with a large (e.g. 1000) amount of products : the process keeps growing until it runs out of memory.

For a real product import of a big producer (e.g. 5000 products), the process grows to 10 Gb of memory.

There are many different things that happen when a product is updated: the ingredients lists are processed, allergens and additives detected, quality is measured etc. Each of those steps involve calling many different subroutines, in particular many functions from Tags.pm related to taxonomies.

At this point, we do not know what exact part of the pre-saving processing causes the memory leak (it is very likely that several parts contribute to it).

This needs further investigation, maybe by testing each part individually (e.g. running each function tens of thousands of times and measuring memory), and/or with tools like Devel::Gladiator etc.

Steps to see the memory leak:

1. Use a CSV file of product data that looks like this:

code | lc | product_name_en | ingredients_text_en |   |   |   |  
-- | -- | -- | -- | -- | -- | -- | --
7000000000000 | en | Test product | Flour, salt, sugar, cranberries, water, lemon, orange, strawberries, pears, grapes |   |   |   |  
7000000000001 | en | Test product | Flour, salt, sugar, cranberries, water, lemon, orange, strawberries, pears, grapes |   |   |   |  
7000000000002 | en | Test product | Flour, salt, sugar, cranberries, water, lemon, orange, strawberries, pears, grapes |   |   |   |  
[..] (with thousands of products)

in /srv/off/scripts/ , run:

./import_csv_file.pl --user_id debug --org debug --source_id debug --source_name debug --source_url debug --define lc=en --csv_file debug-code-product_name-ingredients2.csv

[debug-code-product_name-ingredients2.csv.zip](https://github.com/openfoodfacts/openfoodfacts-server/files/3823502/debug-code-product_name-ingredients2.csv.zip)

Related discussions/issues: #2053 #2054 

This issue is very important to resolve, because it also happens (more gradually) in production on the web site: whenever a product is updated, the Apache mod_perl process grows, and eventually it needs to be restarted.

zigouras:
I've spent some time trying to profile the import products Perl script.
I tried a number of memory profiling packages.
[...]

PR #2587, opened 2019-11-12
Cache CLDR and it's formatters

hangy:
**Description:**
I noticed that `ProductOpener::Text::normalize_percentages` took a lot of time during the CSV import. According to `Devel::NYTProf` that was caused by creating new instances of `CLDR::Number` and it's formatters in loops.

<table>
<tr><th scope="col">What?</th><th scope="col">Before</th><th scole="col">After</th></tr>
<tr><th scope="row">Subroutine list</th><td></td><td></td></tr>
<tr><th scope="row">Sub <code>normalize_percentages</code></th><td></td><td></td></tr>

Since the same will be done for display purposes on the web, this is likely to speed up web request, too.

**Related issues and discussion:** Related to the performance discussion in #2563

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -113,6 +113,7 @@ use ProductOpener::Missions qw(:all);
 use ProductOpener::MissionsConfig qw(:all);
 use ProductOpener::URL qw(:all);
 use ProductOpener::Data qw(:all);
+use ProductOpener::Text qw(:all);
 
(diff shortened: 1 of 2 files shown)


stephanegigandet:
One thing I wonder is whether we could remove all the calls to normalize_percentages. It seems to be used only in Tags.pm, and there are probably very few tags with percents in them. And even if there are, for matching, the % will be removed anyway.

Issue #2589, opened 2019-11-12
Consider removing normalize_percentage() calls from Tags.pm


stephanegigandet:
`normalize_percentages()` is called a lot when building taxonomies and inside `canonicalize_taxonomy_tag()`.

It is possible that it is not needed, or that we could find more efficient ways to deal with tags that contain the % sign.

Example categories tag that contain a % sign:
```
<en:Kefir
en:Kefir 2.5 %
de:Kefir 2.5 %
fr:Kéfir 2,5 %
ru:Кефир 2.5 %, Кефир 2.5%
```

aleene:
Why not disallow such categories (for the moment)? Not sure if such granularity is useful.

stephanegigandet:
@hangy : shall we close this issue now that you optimized normalize_percentages in https://github.com/openfoodfacts/openfoodfacts-server/pull/2593 ?

PR #2590, opened 2019-11-12
added tests with tags with percent signs

stephanegigandet:
to investigate bug #2589


stephanegigandet:
@hangy : can we merge this, if so can you approve it? I think it's useful to have some tests specific to % handling in canonicalization and displaying of tags.

PR #2593, opened 2019-11-13
Conditionally run normalize_percentages

hangy:
**Description:**
If the main body of normalize_percentages is only run if a known percent sign is present, then a lot of sub calls related to CLDR and formatting can be avoided. Performing the check using `index` is much faster than doing anything else.

In my benchmarks, this brought down the time for `normalize_percentages` from 2.25s exclusive/7.13 inclusive to 647ms exclusive/963ms inclusive.
image

**Related issues and discussion:** #2589

### DIFF
--- a/lib/ProductOpener/Text.pm
+++ b/lib/ProductOpener/Text.pm
@@ -33,7 +33,7 @@ BEGIN
 
 					&get_decimal_formatter
 					&get_percent_formatter
-					
+
 					);	# symbols to export on request
 	%EXPORT_TAGS = (all => [@EXPORT_OK]);
 }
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Text.pm
```diff
@@ -47,6 +47,18 @@ sub normalize_percentages($$) {
...
 
+	# Bail out of this function if no known percent sign is found.
+	# This is purely for performance reasons: CLDR functions are
+	# comparatively expensive to run.
+	if ((not (defined $text))
+		or (not ((index($text, "\N{U+0025}") > -1)
+		or (index($text, "\N{U+066A}") > -1)
+		or (index($text, "\N{U+FE6A}") > -1)
+		or (index($text, "\N{U+FF05}") > -1)
+		or (index($text, "\N{U+E0025}") > -1)))) {
```
hangy: This is a bit more difficult to read than
```
		or ((index($text, "\N{U+0025}") == -1)
		and (index($text, "\N{U+066A}") == -1)
		and (index($text, "\N{U+FE6A}") == -1)
		and (index($text, "\N{U+FF05}") == -1)
		and (index($text, "\N{U+E0025}") == -1))) {
```
but as it performs a bit fewer `index` calls, the avg in my test runs was 4µs vs. 3µs.

Issue #2053, opened 2019-07-04
Memory leak in update_all_products.pl


stephanegigandet:
There are memory leaks when updating products with update_all_products.pl (not through Apache and mod_perl).

When the process starts, after the taxonomies are loaded:

| PID USER   |   PR | NI  |  VIRT  |  RES  |  SHR S | %CPU |%MEM |    TIME+| COMMAND|
|---|---|---|---| --- | --- | --- | ---|---|---|
| 6835 off     |  30|  10 | 855956 |**665756** | 11352 R |100.0 | **2.0**  | 0:21.84| update_all+|

after 5 minutes:

| PID USER   |   PR | NI  |  VIRT  |  RES  |  SHR S | %CPU |%MEM |    TIME+| COMMAND|
|---|---|---|---| --- | --- | --- | ---|---|---|
 |6835 off  |     30 | 10| 1177888 |**987596**  |11352 R |100.0| **3.0** | 2:11.81| update_all+|

another 5 minutes:

| PID USER   |   PR | NI  |  VIRT  |  RES  |  SHR S | %CPU |%MEM |    TIME+| COMMAND|
|---|---|---|---| --- | --- | --- | ---|---|---|
 |6835 off  |     30  |10 |1531100| **1.279g** | 11352 R | 99.0|  **4.1** |  4:14.73 |update_all+|

The process eventually gets killed, or we run out of memory and swap.

stephanegigandet:
Trying Devel::MAT

https://tech.binary.com/tracing-perl-memory-leaks-with-devel-mat/

I'm testing update_all_products.pl, with 1 product updated, 10 products, and 100 products. The memory does grow:

-rw-r--r-- 1 off  off   477508290 Jul  4 17:47 file.pmat
-rw-r--r-- 1 off  off   485072433 Jul  4 18:00 file50.pmat
-rw-r--r-- 1 off  off   521910527 Jul  4 18:01 file100.pmat


with file50.pmat:

pmat [more]> count
  Kind         Count (blessed)        Bytes (blessed)
  ARRAY       215105        26     21.3 MiB   2.3 KiB
  CODE         16455        10      2.0 MiB   1.2 KiB
  GLOB         29074         2      4.2 MiB 304 bytes
  HASH        304464      4423    177.3 MiB   1.2 MiB
  INVLIST        435                1.1 MiB
  IO              33        33      5.2 KiB   5.2 KiB
  PAD          13950                4.5 MiB
[...]

PR #2594, opened 2019-11-13
Replace some regular expressions

hangy:
**Description:**
Some optimizations for `get_string_id_for_lang` and `get_fileid`.

In my test runs, it "only" reduced the inclusive time of `get_string_id_for_lang` from 235s to 202s. Replacing most `s/` calls with `tr/`, and the trimming of leading/trailing `-` sped up each statement by about 50%. Here's the latest output of `Devel::NYTProf` of this branch, which shows where most time is spent now: imageimage

**Related issues and discussion:** Related to performance discussion in #2053

### DIFF
--- a/lib/ProductOpener/Store.pm
+++ b/lib/ProductOpener/Store.pm
@@ -80,10 +80,13 @@ sub unac_string_perl($) {
 
 sub get_string_id_for_lang {
 
-	my $lc = shift;
-	my $string = shift;
+	my ($lc, $string) = @_;
 
 	defined $lc or die("Undef \$lc in call to get_string_id_for_lang (string: $string)\n");
+
+	if (not defined $string) {
+		return "";
+	}
 
 	my $unaccent = $string_normalization_for_lang{default}{unaccent};
 	my $lowercase = $string_normalization_for_lang{default}{lowercase};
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Store.pm
```diff
@@ -97,17 +100,37 @@ sub get_string_id_for_lang {
...
 
+	if ((defined $lc) and ($lc eq 'fr')) {
+		$unaccent = 1;
+	}
+
+	return _sanitize_string($file, 1, $unaccent);
+
+}
+
+sub _sanitize_string {
```
stephanegigandet: Given the huge amount of time this function is called, I think it could be worth to inline it in the 2 places it is used. It will double a small amount of code, but we'll save the overhead of calling one more function, copying arguments etc.

hangy: Is calling `sub`s that expensive in Perl? I know that `shift`ing the array is not the fastest way to access parameters, which is why I changed it to `my ($string, $lowercase, $unaccent) = @_;`, but I'm a bit surprised that an additional method/function is slow, since it's quite uncommon in other languages (ie. Java, C#, JavaScript might even inline short methods during JIT).

hangy: I've tried to read up on this, and apparently, Perl is actually pretty horrible with each additional sub call. I'll re-inline the method.

- [why are function calls in Perl loops so slow?](https://stackoverflow.com/a/3259215/11963)
- [Performance issues with subroutines/methods](https://www.perlmonks.org/bare/?node_id=621800)
- [Subroutine overhead in Perl](https://www.perlmonks.org/?node_id=649281)
- [Perl Performance - How to make Perl faster](http://makepp.sourceforge.net/2.0/perl_performance.html#functions)

zigouras:
Do we see a performance improvement with inlining sub _sanitize_string?

hangy:
> Do we see a performance improvement with inlining sub _sanitize_string?

I didn't test that after all. It could be interesting to do that, though. I'll have a look next week.

Issue #2396, opened 2019-10-03
Have 2 separate nutrient fields for energy in kJ and energy in kcal


stephanegigandet:
We currently have only 1 field for energy, which can be entered in kJ or in kcal (which is the same as the "big Calorie" Cal). Internally we convert the energy to kJ.

This is problematic because in Europe, the food labelling law stipulates that the energy should be computed from the other nutritional values, with conversion tables that have slightly differing kJ to kcal ratios for each nutrient.

https://eur-lex.europa.eu/LexUriServ/LexUriServ.do?uri=OJ:L:2011:304:0018:0063:EN:PDF

ANNEX XIV
CONVERSION FACTORS
CONVERSION FACTORS FOR THE CALCULATION OF ENERGY
The energy value to be declared shall be calculated using the following conversion factors:
— carbohydrate (except polyols), 17 kJ/g — 4 kcal/g
— polyols, 10 kJ/g — 2,4 kcal/g
— protein, 17 kJ/g — 4 kcal/g
— fat, 37 kJ/g — 9 kcal/g
— salatrims, 25 kJ/g — 6 kcal/g
— alcohol (ethanol), 29 kJ/g — 7 kcal/g
— organic acid, 13 kJ/g — 3 kcal/g
— fibre, 8 kJ/g — 2 kcal/g
— erythritol, 0 kJ/g — 0 kcal/g

That means that there isn't a single conversion ratio that we can apply to convert the energy from or to kJ and kcal to get exactly the same numbers as printed on food products.

e.g. current nutrient fields for https://fr.openfoodfacts.org/produit/3366321051594/saint-hubert-demi-sel-st-hubert : 

energy_value: 450,
energy_unit: "kcal",
energy_100g: 1883,
energy: 1883,

actual energy printed on label:  1850 / 450 kcal

Proposed solution:

1. Create 2 new nutrient fields: energy_kj and energy_kcal, with mandatory units: kj and kcal.

2. In the product edit form, remove the existing energy field that can be filled in kj or kcal, and display the 2 new fields. (for Europe ; in the US, just display the kcal field)

3. in the API, allow the values to be set through the 2 new fields, or through the existing energy field. If the energy is passed in kj, populate energy_kj. Same for kcal. 

4. When saving values, compute the energy field from energy_kj field if it exists, or from energy_kcal.

5. For existing products, depending on the unit of the energy field, populate energy_kj or energy_kcal.

As a result, we will have those nutrients:

[...]

PR #2596, opened 2019-11-14
Issue 2396/energy in kj and kcal -- huge change, please review

stephanegigandet:
This change is to introduce new energy-kj and energy-kcal nutrients. Please see bug #2396 for the explanation of why we need this.

We need to be careful as we need to handle several situations:

1. I will run the update_all_products.pl script to run compute_serving_size_data() that will populate either energy-kj or energy-kcal from the energy field. But it is going to take hours.

2. In the mean time, there will be the new product edit form on the web site, which will first run compute_serving_size_data() before displaying the form, so that we don't have empty values.

3. There is also the apps: they will still have access to the energy value, and when they update the "energy" field (and not "energy-kj" and "energy-kcal"), we will in fact use the unit they pass to set the energy-kj or -kcal field.

### DIFF
--- a/cgi/product_jqm_multilingual.pl
+++ b/cgi/product_jqm_multilingual.pl
@@ -373,6 +373,22 @@ else {
 
 			my $modifier = undef;
 
+			# energy: (see bug https://github.com/openfoodfacts/openfoodfacts-server/issues/2396 )
+			# 1. if energy-kcal or energy-kj is set, delete existing energy data
+			if (($nid eq "energy-kj") or ($nid eq "energy-kcal")) {
+				delete $product_ref->{nutriments}{"energy"};
+				delete $product_ref->{nutriments}{"energy_unit"};
+				delete $product_ref->{nutriments}{"energy_label"};
+				delete $product_ref->{nutriments}{"energy_value"};
+				delete $product_ref->{nutriments}{"energy_modifier"};
+				delete $product_ref->{nutriments}{"energy_100g"};
+			}
+			# 2. if the nid passed is just energy, set instead energy-kj or energy-kcal using the passed unit
+			elsif (($nid eq "energy") and (($unit eq "kJ") or ($unit eq "kcal"))) {
+				$nid = $nid . "-" . lc($unit);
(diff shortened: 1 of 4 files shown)


stephanegigandet:
The code is currently running on https://world.openfoodfacts.dev (login and pass : off )

I'm running the update script.

cgi/product_jqm_multilingual.pl
```diff
@@ -373,6 +373,22 @@
 
 			my $modifier = undef;
 
+			# energy: (see bug https://github.com/openfoodfacts/openfoodfacts-server/issues/2396 )
+			# 1. if energy-kcal or energy-kj is set, delete existing energy data
+			if (($nid eq "energy-kj") or ($nid eq "energy-kcal")) {
```
teolemon: - vs _ ?

stephanegigandet: All nutrient identifiers use a dash. e.g. saturated-fat. Then we add _things with underscore like _unit, _value, _100g, _serving etc.

lib/ProductOpener/Food.pm
```diff
@@ -4820,8 +4909,39 @@ sub compute_serving_size_data($) {
...
 	foreach my $product_type ("", "_prepared") {
 
+		# Energy
+		# Before November 2019, we only had one energy field with an input value in kJ or in kcal, and internally it was converted to kJ
+		# In Europe, the energy is indicated in both kJ and kcal, but there isn't a straightforward conversion between the 2: the energy is computed
+		# by summing some nutrients multiplied by an energy factor. That means we need to store both the kJ and kcal values.
+		# see bug https://github.com/openfoodfacts/openfoodfacts-server/issues/2396
+
+		# If we have a value for energy-kj, use it for energy
+		if (defined $product_ref->{nutriments}{"energy-kj" . $product_type}) {
```
VaiTon: So we give priority to kJ?

stephanegigandet: Yes, kJ is the official value in EU, kcal is historic. In the US we only have kcal. Also kJ is more precise.

VaiTon: Perfect

PR #2616, opened 2019-11-18
Producers - small fixes

stephanegigandet:


lib/ProductOpener/Images.pm
```diff
@@ -1253,24 +1253,11 @@ HTML
...
-ff00cc
-cc00ff
-);
-		my $color_id = $product_ref->{code} % (scalar @colors);
-		my $color = $colors[$color_id];
-
 		$html = <<HTML
-<div style="background-color:#$color">
-<img src="$static/images/misc/pacman.svg" width="$thumb_size" height="$thumb_size" alt="Please add pictures of the product if you have it!" $css/>
+<div style="background-color:#eeeeee" style="width:$thumb_size;height:$thumb_size">
```
VaiTon: Just an empty div? Wouldn't it be better with a placeholder?

minion_log.conf
```diff
@@ -0,0 +1,10 @@
+log4perl.rootLogger=DEBUG, LOGFILE
```
VaiTon: What is this file for?

stephanegigandet: It's the log file configuration for the Minion queue, so that we can turn on different levels of logs for debugging etc.

lib/ProductOpener/Images.pm
stephanegigandet: Well the empty div is simple and does the job, what kind of placeholder do you have in mind?

VaiTon: I was thinking of something like a question mark.

PR #2623, opened 2019-11-19
SVG icons

hangy:
Use SVG icons instead of a custom icon font, because that's what the cool kids do. 😉

Actually, the currently used `gulp-iconfont` package is causing regular [security warnings](https://github.com/openfoodfacts/openfoodfacts-server/network/alert/package-lock.json/lodash.template/open), and is (or rather it's  [depen](https://github.com/nfroidure/gulp-ttf2eot/pull/4)[dencies](https://github.com/nfroidure/gulp-ttf2woff/pull/7) are) not maintained any more. Additionally, Lighthouse complains about the icon font, because it can cause display issues. Most guides [nowadays](https://www.lambdatest.com/blog/its-2019-lets-end-the-debate-on-icon-fonts-vs-svg-icons/) [point](https://dev.to/marcel_cremer/the-icon-evolution-and-how-i-switched-from-font-based-icons-to-svg-11bm) out that SVG icons are the better option when compared to icon fonts. Apparently [inlining](https://css-tricks.com/pretty-good-svg-icon-system/) the icons is better than SVG sprites because of compatibility issues with the `<symbol>` approach, and [missing styling opportunities when using ``](https://css-tricks.com/the-many-ways-to-change-an-svg-fill-on-hover-and-when-to-use-them/#article-header-id-0).

In my tests, it [worked](https://caniuse.com/#feat=mdn-svg_elements_svg) well in Chrome, Edge, and even Internet Explorer. The Same approach is also used by [Google Search](https://www.google.com/), [GitHub](https://github.com/), [Twitter](https://twitter.com/), [Teams](https://teams.microsoft.com/), and most likely other big, well-performing web apps.


stephanegigandet:
I'll need to test this on the dev server, I'm a bit worried about the sizing and color of the icons, which was previously done through the text CSS attributes. If we inline SVG images, I don't see how those styles are kept right now.

hangy:
> I'll need to test this on the dev server

As we should. 👍 I actually found a minor when used on a small responsive device.

> I'm a bit worried about the sizing and color of the icons, which was previously done through the text CSS attributes. If we inline SVG images, I don't see how those styles are kept right now.

That's done via CSS, too:
https://github.com/openfoodfacts/openfoodfacts-server/blob/6767fa3127fe8534523298ca89e8072e4bf5a84a/scss/_off.scss#L336-L344 adds the necessary styling, and https://github.com/openfoodfacts/openfoodfacts-server/blob/6767fa3127fe8534523298ca89e8072e4bf5a84a/gulpfile.js#L18-L24 ensures that the minimized SVGs contain a `viewBox` (so that they can be resized relatively), and have the necessary CSS class.

hangy:
> > I'll need to test this on the dev server
> 
> As we should. 👍 I actually found a minor when used on a small responsive device.

Now fixed (was actually an anchor tag I mistakenly left opened).

hangy:
> @hangy how is the class "icon" added to the `<svg>`?

The `gulp-svgmin` tool does it when minifying the SVG files: https://github.com/openfoodfacts/openfoodfacts-server/blob/c0d762e1871e29bd1a9cca859db7a27e8aa6ed73/gulpfile.js#L22

stephanegigandet:
How do I generate this directory?
could not open /srv/off/html/images/icons/dist/settings.svg: No such file or directory at /srv/off-net/lib/ProductOpener/Display.pm line 9931.

I tried "npm install --unsafe-perm", I forgot if there is something else?

hangy:
> I tried "npm install --unsafe-perm", I forgot if there is something else?

Did you also run `npm run build` to perform the Gulp build? This process hasn't changed with this PR.

stephanegigandet:
Thanks @hangy, I was missing npm run build.

I'm trying it on .net

It looks good to me, there's just one change needed for the vegetarian / vegan icons, could you change the styles so that they have those styles? I could do it in the HTML but I'm guessing some place in the scss file may be better.

````
<span class="alert round label" style="background-color:#00aa00;color:white;font-size:1rem;padding-left:0.2rem;padding-right:1rem;">
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" class="icon" aria-hidden="true" focusable="false" style="width:32px;height:32px;vertical-align:middle"></svg>
Sans huile de palme</span>

**Comment by stephanegigandet** [comment]
I'm trying to get the look of the first button on the left:

image

**Comment by hangy** [comment]
> I'm trying to get the look of the first button on the left:
> 
> image

Fixed, see d3ab96fd1648b8ab1f5f920f05e93953469f9e15. 🙂

File: lib/ProductOpener/Display.pm [hunk 1]
```diff
@@ -7423,7 +7423,7 @@ JS
 				$icon = "<span style=\"font-size:32px;margin-right:0.2em;vertical-align:middle;line-height:24px;\">". display_icon($icon) ."</span>";
 			}
 
-			$html_analysis .= "<span class=\"alert round label\" style=\"background-color:$color;color:white;font-size:1rem;padding-right:1em;\">"
+			$html_analysis .= "<span class=\"alert round label ingredients_analysis\" style=\"background-color:$color;color:white;font-size:1rem;padding-left:0.2rem;padding-right:1rem;\">"
```
VaiTon: Can we put the `padding left`, `padding right`, `color` and `font-size` in the scss?

hangy: Of course. Good catch!

lib/ProductOpener/Display.pm
```diff
@@ -6212,7 +6213,8 @@ HTML
 <nav class="tab-bar show-for-small-only">
 	<div class="left-small" style="padding-top:4px;">
 		<a href="#idOfLeftMenu" role="button" aria-controls="idOfLeftMenu" aria-expanded="false" class="left-off-canvas-toggle button postfix">
-		<i class="icon-account_box" style="color:$torso_color;font-size:1.8rem"></i></a>
+		<span style="color:$torso_color;font-size:1.8rem">@{[ display_icon('account_box') ]}</span>
```
VaiTon: @hangy same here for `font-size`.

We could add:
```scss
#idOfLeftMenu {
  .icon {
    font-size: 1.8rem;
  }
}
```
to `_off.scss`

hangy: I agree that we could move this inline style to the SCSS file, but the element with id `idOfLeftMenu` doesn't appear to actually exist on the page. I wonder how the expand button actually works in this case.

PR #2627, opened 2019-11-20
Add a promo for Open Beauty Facts

teolemon:
Add a promo for Open Beauty Facts

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -810,6 +810,16 @@ msgctxt "footer_wiki_link"
 msgid "https://en.wiki.openfoodfacts.org"
 msgstr ""
 
+# Do not translate
+msgctxt "footer_obf"
+msgid "Open Beauty Facts - Personal care"
+msgstr ""
+
+# Do not translate
+msgctxt "footer_obf_link"
+msgid "https://world.openbeautyfacts.org"
+msgstr ""
+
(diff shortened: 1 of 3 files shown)


po/common/common.pot
```diff
@@ -810,6 +810,16 @@ msgctxt "footer_wiki_link"
 msgid "https://en.wiki.openfoodfacts.org"
 msgstr ""
 
+# Do not translate
```
stephanegigandet: Do not translate "Open Beauty Facts" but do translate "Personal care"... I think "Cosmetics" would be safer, personal care can be too easily mistranslated. "Soin personnel" wouldn't mean much in French for instance.

teolemon: Ok for cosmetics. Anectodal evidence shows me that cosmetics looses us 50% of the potential userbase.

po/common/common.pot
```diff
@@ -810,6 +810,16 @@ msgctxt "footer_wiki_link"
...
+msgctxt "footer_obf"
+msgid "Open Beauty Facts - Personal care"
```
```suggestion
msgid "Open Beauty Facts - Cosmetics"
```

po/common/common.pot
```suggestion
# Do not translate "Open Beauty Facts" but do translate "Cosmetics"
```

PR #2634, opened 2019-11-21
chore: logos - commit by teolemon

teolemon:


VaiTon:
@teolemon what are these logos used for?

teolemon:
Those logos are 
- the labels displayed on the product page
- the app stores logos renamed and normalized
- some logos used on the partner page.

They all are copyrighted, so we should attribute copyright to their rightful owners if we merge this

stephanegigandet:
Closing as we now have fresher versions of those logos

PR #2638, opened 2019-11-21
Update contribuer.html

teolemon:


VaiTon:
Woudn't it be better if we update the eng version? Crowdin would do the rest.

teolemon:
We have legacy files for French. Eventually, I want to kill them alltogether.

teolemon:
lang/fr/texts/contribuer.html (legacy)
lang/fr/texts/contribute.html (standardized)

VaiTon:
Merging this. Let's try to make eng default.

PR #2639, opened 2019-11-21
fix for pnns groups / sweetened beverages

stephanegigandet:


lib/ProductOpener/Food.pm
```diff
@@ -4053,6 +4061,12 @@ sub special_process_product($) {
 		$product_ref->{pnns_groups_1} = "unknown";
 		$product_ref->{pnns_groups_1_tags} = ["unknown", "missing-association"];
 	}
+
+	# Put back the original categories_tags so that they match what is in the taxonomy field
+	# if there is a mistmatch it can cause tags to be added multiple times (e.g. with imports)
+	if (defined $product_ref->{categories_tags}) {
```
hangy: Could the tags be `categories_tags` changed to `undefined` in the meanwhile? It _might_ be safer to set `@{$product_ref->{categories_tags}}` if  `@original_categories_tags` has any elements.

t/food.t
```diff
@@ -93,7 +93,7 @@ my $product_ref = {
 
 special_process_product($product_ref);
 
-ok( (not has_tag($product_ref, 'categories', 'en:unsweetened-beverages')), 'should not add en:unsweetened-beverages' ) || diag explain $product_ref;
+#ok( (not has_tag($product_ref, 'categories', 'en:unsweetened-beverages')), 'should not add en:unsweetened-beverages' ) || diag explain $product_ref;
```
hangy: Are these tests outdated?

stephanegigandet: They are outdated if we don't turn off and on automatically the categories for sweetened and unsweetened beverages using the analysis of the ingredients. We used to do it automatically, but it's causing issues, that's why I made the changes in Food.pm.  But if we do reenable it at some point, then the tests would be useful, that's why I commented them.

lib/ProductOpener/Food.pm
stephanegigandet: Good idea, i'll change it.

stephanegigandet:
@hangy : I uncommented the tests and adapted them to the new behaviour (which is now to not automatically add or remove categories for sweetened/unsweetened beverages).

t/food.t
hangy: You can always get the code back from the git history. A huge problem with commented code is, that no one knows (from the file itself) why the code is commented. Come back a few months later, and you probably won't know why those single lines are comments. Additionally, dead code makes the source files difficult to read.

VaiTon:
There are 13 failed tests and 1 non-zero exit status.

PR #2653, opened 2019-11-24
Made energy-kJ and energy-kcal have only one unit, removed select

VaiTon:
### Before
immagine
immagine
immagine

### After
immagine

### DIFF
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -1882,7 +1882,6 @@ HTML
 			if (($nid eq 'alcohol')) { $unit = '% vol / °' } # alcohol in % vol / °
 			elsif (($nid eq 'energy-kj')) { $unit = 'kJ' }
 			elsif (($nid eq 'energy-kcal')) { $unit = 'kcal'}
-
 			
 			$input .= <<"HTML"
 <td>
@@ -2335,4 +2334,3 @@ display_new( {
 
 
 exit(0);
-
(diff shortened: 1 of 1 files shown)


cgi/product_multilingual.pl
```diff
@@ -1876,82 +1876,79 @@ ($$$$$$)
...
 ;
 
-		if ($nid ne 'alcohol') {
+		if (($nid eq 'alcohol') or ($nid eq 'energy-kj') or ($nid eq 'energy-kcal')) {
+			my $unit = '';
+			
+			if (($nid eq 'alcohol')) { $unit = '% vol / °' } # alcohol in % vol / °
+			elsif (($nid eq 'energy-kj')) { $unit = 'kJ' }
+			elsif (($nid eq 'energy-kcal')) { $unit = 'kcal'}
+
```
VaiTon:
```suggestion
```

cgi/product_multilingual.pl
```diff
@@ -1876,82 +1876,78 @@ ($$$$$$)
 HTML
 ;
 
-		if ($nid ne 'alcohol') {
+		if (($nid eq 'alcohol') or ($nid eq 'energy-kj') or ($nid eq 'energy-kcal')) {
+			my $unit = '';
+			
+			if (($nid eq 'alcohol')) { $unit = '% vol / °' } # alcohol in % vol / °
+			elsif (($nid eq 'energy-kj')) { $unit = 'kJ' }
+			elsif (($nid eq 'energy-kcal')) { $unit = 'kcal'}
```
hangy: Does this work without semicolons terminating each line?

VaiTon: >Every simple statement must be terminated with a semicolon, unless it is the final statement in a block, in which case the semicolon is optional. 

But for the sake of convention i will add them. :)

hangy: Oh okay, that's news to me. 🙂 nvm then.

Issue #2085, opened 2019-07-12
New permissions / users groups system for confirmed users, moderators, admins and producers


stephanegigandet:
This is to replace the current list of hardcoded usernames for admins.

Proposal:
- Special groups / features / permissions are set by admins with checkboxes directly on the web site on the profile of each user.

Instead of having one checkbox for each individual feature (e.g. "allow to move products", "allow to delete products", "enable quick edit fields", "protect edits" (for producers), "allow edits of protected edits" (for experimented users), which would force us to go through all power users profiles whenever we add a new feature, we can try to instead have a small number of user groups or feature sets.

e.g.

- admins : can change user permissions
- moderators : can delete products and images, move products to obf / opf etc.
- expert users : access to quick edit features, can override producer data / images
- producers : mark edits as being  from producer (triggers data source)
- protected producers : protect edits from non expert users

For producers, it would be useful to be able to list the corresponding brands as well.

VaiTon:
We could use a permission system based on "nodes" (eg. "admin.image.move") and then groups have the main prefix (eg. "admin.*") and the software gives them all sub-permissions

hangy:
> We could use a permission system based on "nodes" (eg. "admin.image.move") and then groups have the main prefix (eg. "admin.*") and the software gives them all sub-permissions

I was thinking something similar. It might be interesting to have something similar to RBAC as implemented in [Apache Shiro](http://shiro.apache.org/permissions.html). Interestingly enough, there's [not that many RBAC modules on  CPAN](https://metacpan.org/search?size=500&q=rbac).

stephanegigandet:
We can hopefully find something simple. https://metacpan.org/pod/RBAC::Tiny seems ok, except I'd like to be able to give multiple roles to a single user.

stephanegigandet:
OK, this group system is badly needed, especially for the producers platform, and also to remove the hardcoded user names from the code. So I made something really simple but that works. At this point I don't think we need to over-engineer this.

Here what it looks like in the product edit form, which is available only for admins. There will be much fewer admins, and many more moderators (who will be able to do everything admins can do, except changing user groups).

![image](https://user-images.githubusercontent.com/8158668/69631152-e17fd980-104d-11ea-92c3-4abfedf7a121.png)

stephanegigandet:
Deployed.

PR #2654, opened 2019-11-25
Added and fixed tests for producers platform, small fixes

stephanegigandet:


cgi/export_products.pl
```diff
@@ -70,6 +70,10 @@
 	$html .= "<p>" . lang("producers_platform_licence") . "</p>";
 	$html .= "<p>" . lang("export_product_data_photos_please_check") . "</p>";
 
+	# Display button for admins only
+
+	if ($User_id =~ /-off$/) {
```
VaiTon: Why not use `isAdmin()`?

lib/ProductOpener/Producers.pm
```diff
@@ -413,6 +415,22 @@ sub init_fields_columns_names_for_lang($) {
...
+	"g" => "g",
+	"gr" => "g",
+	"grams" => "g",
+	"grammes" => "g",
+	"mg" => "mg",
+	"mcg" => "mcg",
+	"percent" => "percent",
+	"kj" => "kj",
+	"kcal" => "kcal",
+	"cal" => "kcal",
```
VaiTon: cal is 1/1000 of a kcal

stephanegigandet: This is not a conversion table, it is a list of synonyms used by producers when they transmit us data. In practice absolutely no producer uses cal (as in 1/1000 of kcal) as a unit for energy. When they have "cal" or "calories" in the header of a column, they always mean kcal. The units in this table are lowercased, so it's more for the "big Calories".

cgi/export_products.pl
stephanegigandet: Good point. I'll make things cleaner, I need to implement https://docs.google.com/document/d/154MN36FACbS3PqSaEcsoerSz0n6jo9RwaD3eLRChNgI/edit first though.

stephanegigandet: Good point, I'll make things cleaner. I need to implement https://github.com/openfoodfacts/openfoodfacts-server/issues/2085 first.

icons/arrow_upward.svg
```diff
@@ -0,0 +1 @@
+material/baseline-arrow_upward-24px.svg
```
VaiTon: Why don't we just rename the file?

lib/ProductOpener/Producers.pm
VaiTon: Ok, just to make it obvious we could add a comment explaining what you wrote here.

icons/arrow_upward.svg
stephanegigandet: To keep track that it's a material icon.

lib/ProductOpener/Producers.pm
stephanegigandet: Good idea, I'll add it.

PR #2655, opened 2019-11-25
Made NOVA score 1 for waters

### LINKED ISSUE #2362: Make water NOVA 1 if not flavoured, sugared etc.
The NOVA classification is currently done through the ingredients list, but for most spring water, mineral water etc. we don't have ingredients listed.

VaiTon:
Fixes #2362

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -5637,11 +5637,20 @@ sub compute_nova_group($) {
 
 	$product_ref->{nova_group_debug} = "";
 
-	# do not compute a score when we don't have ingredients
+	# If we don't have ingredients, only compute score for water
 	if ((not defined $product_ref->{ingredients_text}) or ($product_ref->{ingredients_text} eq '')) {
+
+		# Exclude flavored waters
+		if (has_tag($product_ref, 'categories', 'en:waters')
+		    and (not has_tag($product_ref, 'categories', 'en:flavored-waters'))) {
+			$product_ref->{nova_group} = 1;
+			return;
+
+		} else {
 			$product_ref->{nova_group_tags} = [ "not-applicable" ];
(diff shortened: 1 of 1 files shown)


stephanegigandet:
It looks good to me, but I wanted to check which products would be impacted, and it turns out this query does not work as intended:

https://world.openfoodfacts.dev/category/waters/category/-flavored-waters

The resulting query is

          '$and' => [
                      {
                        'categories_tags' => 'en:waters'
                      },
                      {
                        'categories_tags' => 'en:flavored-waters'
                      }
                    ]

stephanegigandet:
Filed https://github.com/openfoodfacts/openfoodfacts-server/issues/2657

stephanegigandet:
It does work in the reverse order: https://world.openfoodfacts.org/category/-flavored-waters/category/waters

stephanegigandet:
So one issue is that there seems to be around 5% of badly categorized products (e.g. not waters, or waters that should have the category flavoured water).

e.g. https://world.openfoodfacts.org/product/3329120002235/eau-minerale-gazeuse-aromatisee-mandarine-passion
https://world.openfoodfacts.org/product/20400248/lemon-lime-still-water-carrick-glen

Some of them we could probably fix by looking at nutrients: if energy is above 0, then it's not pure water.

VaiTon:
> 
> 
> So one issue is that there seems to be around 5% of badly categorized products (e.g. not waters, or waters that should have the category flavoured water).
> 
> e.g. https://world.openfoodfacts.org/product/3329120002235/eau-minerale-gazeuse-aromatisee-mandarine-passion
> https://world.openfoodfacts.org/product/20400248/lemon-lime-still-water-carrick-glen
> 
> Some of them we could probably fix by looking at nutrients: if energy is above 0, then it's not pure water.

That's another story. Could be added to SiteQuality_off though!

PR #2659, opened 2019-11-26
Added icon for product without front image

### LINKED ISSUE #2131: Add "no image" icon to products with no front image
Suggestion from Max Bredow, see: https://openfoodfacts.slack.com/archives/C02L4H55V/p1564144835002700

> I'm new to this group but I have a suggestion:
When browsing through the page, I was curious what all these colours mean and why pacman (or is it a pie?). Only after a while I realized that it's completely random. This completely takes away focus from the CTAs.I propose a less beautified but more logic design to help the user differentiate between important information (such as the NutriScore or NOVA rating) and information that the user can disregard (colours? pacman?)

image

VaiTon:
Fixes #2131


stephanegigandet:
I'm going to change the icon, otherwise looks good, thank you!

PR #2660, opened 2019-11-26
ingredient parsing: improve Finnish parsing

Ban3:
**Description:**

No functional changes, just adding to the regular expressions to improve the cutting of the ingredient list.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -101,6 +101,7 @@ my %traces_regexps = (
 	en => "traces|may contain",
 	de => "Kann Spuren|Spuren",
 	es => "puede contener|trazas|traza",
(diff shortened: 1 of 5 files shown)


stephanegigandet:
Hi @Ban3 , thanks a lot for the changes. Do you think that you could add a few tests in the /t/ directory? They help a lot with ingredients parsing, because very often when we change something in he parsing at one place, it affects things in other places, and it's really easy to break things that used to work before by fixing something else.

e.g. in ingredients.t ingredients_parsing.t etc. Just a few tests would be good I think.

Ban3:
@stephanegigandet I added a bunch of tests. The tests for traces and labels are a bit unnatural, but that will be fixed with changes to the taxonomies. I'll probably look into [Lemmatisation](https://en.wikipedia.org/wiki/Lemmatisation) of the ingredient list in the future, as that would remove the need for tens of synonyms for complex languages like Finnish.

stephanegigandet:
@Ban3 Thanks a lot for adding the tests!

PR #2676, opened 2019-11-29
add robots.txt

stephanegigandet:
I included the major SEO robots. We had one on OFF, but we were missing it on OBF etc.


hangy:
What about https://github.com/openfoodfacts/openfoodfacts-server/blob/829b38da7e87dc1b5830c23537f1c9e5b59203b7/html/files/robots.txt?

VaiTon:
Why are we disallowing them? noob question

stephanegigandet:
> What about https://github.com/openfoodfacts/openfoodfacts-server/blob/829b38da7e87dc1b5830c23537f1c9e5b59203b7/html/files/robots.txt?

I was under the impression we had one somewhere, but I didn't find it. I'll include yours in the one in /html/ as that's what's get served by nginx.

stephanegigandet:
> Why are we disallowing them? noob question

Most of those are robots for SEO, for sites that allow webmasters to see who links to them. They have absolutely no value to us, but very often they request tens of thousands of pages and overload our server.

PR #2677, opened 2019-11-29
Added taxonomy translations (ita)

VaiTon:


hangy:
@VaiTon Could you try building the taxonomies on your PC? Apparently, there is a problem with the labels: https://github.com/openfoodfacts/openfoodfacts-server/pull/2677/checks?check_run_id=325891751#step:5:16548
> [labels] child process exited with code 255

PR #2648, opened 2019-11-23
New Crowdin translations

teolemon:
Enjoy!

PR #2683, opened 2019-11-30
Use index.html instead of the Foundation version 

### LINKED ISSUE #84: [FOUNDATION] Create new index text
Text shown at the top of the page on the root of each country (e.g. fr.openfoodfacts.org).

It should be shorter than the current text (too much text, scary). Instead it should link to the Discover and Contribute pages.

hangy:
Reopening #2668 because #2648 was closed and the branch was deleted.

Probably also fixes #84


stephanegigandet:
@teolemon : are the crowdin changes ok for you?

teolemon:
2 conflicts to solve. Anything that standardizes those files is a good step forward.

Issue #2688, opened 2019-11-30
Rearchitect madenearme


stephanegigandet:
https://madenear.me , https://madenear.me.uk and https://cestfabriquepresdechezvous.fr are static pages that have been generated with CGI scripts that can Display.pm search_and_map() function. They are basically ProductOpener's search function with the option of displaying result on a map.

We have a nightly cron job to generate them:

*/45 1 * * *    wget -q -O /srv/off/html/cestemballepresdechezvous.html http://fr.openfoodfacts.org:8001/cgi/madenearyou.pl
*/50 2 * * *    wget -q -O /srv/off/html/madenearme-uk.html http://uk.openfoodfacts.org:8001/cgi/madenearme.pl
*/55 3 * * *    wget -q -O /srv/off/html/madenearme.html http://world.openfoodfacts.org:8001/cgi/madenearme.pl

The generation actually takes a lot of time, so much time (and memory) that it actually crashes sometime.

This bug is to:

1. Replace cgi/madenearme.pl and cgi/madenearyou/pl by script/generate_madenearme_page.pl so that we have a script that we can call outside of modperl to generate the HTML pages.

2. Investigate how to speed up the page generation.

stephanegigandet:
Note: this bug is not for solving the issue that generating a huge static html file should be instead replaced by some kind of map where elements are loaded dynamically. That would be another bug. :)

hangy:
> Note: this bug is not for solving the issue that generating a huge static html file should be instead replaced by some kind of map where elements are loaded dynamically. That would be another bug. :)

That's #1668 😉

stephanegigandet:
So it turns out that a huge chunk of the time is actually used by the MongoDB perl driver to retrieve and process the results:

Profile of ./generate_madenearme_page.pl for 151s (of 156s), executing 6122657 statements and 1977824 subroutine calls in 742 source files and 266 string evals.


Calls | P | F | ExclusiveTime | InclusiveTime | Subroutine
-- | -- | -- | -- | -- | --
847 | 2 | 1 | 131s | 131s | MongoDB::_Link::CORE:sselect (opcode)
101 | 1 | 1 | 5.67s | 5.67s | BSON::XS::_decode_bson (xsub)
284 | 1 | 1 | 2.09s | 2.09s | Storable::pretrieve (xsub)
63345 | 13 | 5 | 1.16s | 1.37s | ProductOpener::Store::get_string_id_for_lang
[...]

PR #2690, opened 2019-11-30
Issue 2688/madenearme - replace cgi by standalone script, remove a lot of unused code

stephanegigandet:
Rearchicted madenear.me : #2688


scripts/generate_madenearme_page.pl
```diff
@@ -0,0 +1,106 @@
...
+
+# We want products with emb codes
+$query_ref->{"emb_codes_tags"} = { '$exists' => 1 };
+
+$request_ref->{map_options} = $map_options{$cc} || "";
+
+my $map_html = search_and_map_products($request_ref, $query_ref, $graph_ref);
+
+
+$html =~ s/<HEADER>/$header/;
```
hangy: > [Sat Nov 30 18:12:25 2019] generate_madenearme_page.pl: Use of uninitialized value $header in substitution (s///) at generate_madenearme_page.pl line 99.

scripts/generate_madenearme_page.pl
```diff
@@ -0,0 +1,106 @@
...
+$html =~ s/<INITJS>/$initjs/;
```
hangy: > [Sat Nov 30 18:12:25 2019] generate_madenearme_page.pl: Use of uninitialized value $initjs in substitution (s///) at generate_madenearme_page.pl line 100.

stephanegigandet:
> Would it make sense to just set them to `''` at the top of this script?

Well in fact it looks like we don't need them, the generated files seem to work quite well without them :)

hangy:
> Well in fact it looks like we don't need them, the generated files seem to work quite well without them :)

ACK. We could, however, either initialize the variables or remove the replace (and placeholders) to get rid of the warnings.

stephanegigandet:
OK, restricting the fields returned by MongoDB results in a huge speed improvement.
Details here:
https://github.com/openfoodfacts/openfoodfacts-server/issues/2688#issuecomment-560012616

We probably need to do something similar for the graphs, and possibly for the lists of products (although for those, since we retrieve only 20 products, it's not as bad as for the maps and graphs where we can process hundreds of thousands of products...)

hangy:
> and possibly for the lists of products (although for those, since we retrieve only 20 products, it's not as bad as for the maps and graphs where we can process hundreds of thousands of products...)

Since we also return the whole result set without a projection from the JSON API, that could be a bad breaking change.

stephanegigandet:
> ACK. We could, however, either initialize the variables or remove the replace (and placeholders) to get rid of the warnings.

Sure, I initialized them. I wasn't getting the warnings, they only appeared when the map had no results I think.

stephanegigandet:
> Since we also return the whole result set without a projection from the JSON API, that could be a bad breaking change.

Right, we'll need to be careful, maybe we can do it when it's not an API request. We'll probably need to use separate memcached keys too.

PR #2696, opened 2019-12-01
Wikidata link

### LINKED ISSUE #449: Use Wikidata ids as stable ids
Feature Request: Add redirect thanks to Wikidata IDs.

For example : 

http://world.openfoodfacts.org/category/Q7802 would redirect to http://world.openfoodfacts.org/category/bread

http://world.openfoodfacts.org/country/Q38 would redirect to http://world.openfoodfacts.org/country/Italy

hangy:
**Description:**
Would allow links like http://world.openfoodfacts.org/category/wikidata:en:Q7802 to display http://world.openfoodfacts.org/category/breads, based on the present linked data.

**Related issues and discussion:** Fixes #449

### DIFF
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -2640,7 +2640,7 @@ sub canonicalize_taxonomy_tag_linkeddata {
 		or (not defined $tag)
 		or (not ($tag =~ /^(\w+:\w\w):(.+)/))
 		or (not defined $properties{$tagtype})) {
-			return undef;
+			return;
 	}
 
 	# Test for linked data, ie. wikidata:en:Q1234
@@ -2668,7 +2668,7 @@ sub canonicalize_taxonomy_tag_weblink {
 	if ((not defined $tagtype)
 		or (not defined $tag)
 		or (not ($tag =~ /^https?:\/\/.+/))) {
-			return undef;
+			return;
 	}
 
 	# Test for linked data URLs, ie. https://www.wikidata.org/wiki/Q1234
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -58,6 +58,8 @@ BEGIN
 					&build_tags_taxonomy
 
 					&canonicalize_taxonomy_tag
+					&canonicalize_taxonomy_tag_linkeddata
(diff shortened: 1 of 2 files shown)


hangy:
Note: Theoretically, it should also work for [http://world.openfoodfacts.org/category/https%3A%2F%2Fwww.wikidata.org%2Fwiki%2FQ1197400](http://world.openfoodfacts.org/category/https%3A%2F%2Fwww.wikidata.org%2Fwiki%2FQ1197400), but for some reason, the 2nd slash gets lost somewhere (maybe in Apache or the CGI perl module).

lib/ProductOpener/Tags.pm
```diff
@@ -2541,52 +2543,14 @@ sub canonicalize_taxonomy_tag($$$)
...
-				}
-
-				$matched_tagid = $canon_tagid;
-			}
-		}
-
-		if (defined $matched_tagid) {
-			return $matched_tagid;
-		}
+	my $linked_data_tag = canonicalize_taxonomy_tag_linkeddata($tagtype, $tag);
```
stephanegigandet: In what circumstances would canonicalize_taxonomy_tag() be called with a linked data or weblink identifier?

When they are used in URLs like in http://world.openfoodfacts.org/category/wikidata:en:Q7802 then the canonicalization should already have been done through analyze_request()

canonicalize_taxonomy_tag() is used quite extensively internally, with input values that I think are never wikidata ids, so I would much prefer that we remove any extra subroutine calls from it.

hangy: I'm not sure about the issue number, but the link detection was built in to `canonicalize_taxonomy_tag` to support users entering Wikidata ids or URLs on the website or the API when adding ie. category tags.

stephanegigandet: I think we should remove that support. I have not seen anyone (or anything) using this. And meanwhile, it adds processing / memory usage for tons of calls to that function, on all taxonomies. E.g. for ingredient parsing, we call canonicalize_taxonomy_tag for each ingredient in the ingredient list. We could keep the support for the URLs (although I'm not really sure who would use it and in what circumstances), but remove for input values to tags.

VaiTon:
@stephanegigandet what shall we do?

PR #2697, opened 2019-12-01
allergens: update Finnish translations

Ban3:
**Description:**

Is there a prize for the longest taxonomy line? :wink: 
More seriously, the lines got really long because not only are there multiple items per allergen, but all of the items appear in multiple cases in Finnish packaging. The inflections were generated using [voikko](https://voikko.puimula.org/) and few of the most commonly found cases were picked (nominative, partitive, elative). Lemmatisation is the way to go for the long term, but for the time being harcoding few of the most common cases seems to work reasonably well.


stephanegigandet:
> Lemmatisation is the way to go for the long term

Right, clearly our current language processing does not work well for Finnish.

Thanks for the changes!

PR #2709, opened 2019-12-02
Remove page that is not synced and duplicate

teolemon:
Remove page that is not synced and duplicate. Will need symlinks to nutrition-traffic-lights.html


stephanegigandet:
> Will need symlinks to nutrition-traffic-lights.html

Were those pages actually linked from somewhere?

teolemon:
I don't think so, just me being over-cautious

Issue #2675, opened 2019-11-28
Consider replacing jQuery UI


hangy:
I've been running a few website performance tests, and at least for mobile, Lighthouse points out that we load `jquery-ui.js` (587 KB pre-gzip) on all pages. Not only is this the largest single JS file we have, but it's stylesheet is also the largest file render-blocking resource after `app.css` (which is mainly Foundation 5, our main UI library).

- [x] The jQuery UI portion used by [blueimp-file-upload](https://github.com/blueimp/jQuery-File-Upload) can be replaced by [jquery.ui.widget](https://www.npmjs.com/package/jquery.ui.widget) which is around 15 KB (pre-gzip)
- [x] `autocomplete` is only really used for nutrients - this can probably replaced by [awesomplete](https://leaverou.github.io/awesomplete/) (or maybe even [select2](https://select2.org/), which we already load anyways)
- [x] `tagsInput` (which uses `autocomplete`) is [not maintained any more](https://github.com/xoxco/jQuery-Tags-Input), so it might be worthwhile to switch to a modern alternative like [tagify](https://github.com/yairEO/tagify)

BathoryPeter:
jqueryui.com allows building custom packages. Just uncheck Toggle All and then check only the needed modules. It will pay attention to dependencies automatically.
For example enabling only autocomplete module (and its dependencies), jquery-ui.min.js is only 32 kB.

hangy:
> jqueryui.com allows building custom packages. Just uncheck Toggle All and then check only the needed modules. It will pay attention to dependencies automatically.
> For example enabling only autocomplete module (and its dependencies), jquery-ui.min.js is only 32 kB.

That's a good suggestion. Thanks! I suppose we might also be able to achieve this by using just some source files form `jquery-ui` npm package instead of `jquery-ui-dist`.

PR #2710, opened 2019-12-02
Tagify

VaiTon:
Remove tagsInput (jQuery plugin) and use vanilla js [Tagify](https://github.com/yairEO/tagify).

Related to #2675


stephanegigandet:
Hi @VaiTon , does it keep our autocomplete with the taxonomy entries + recently used entries?

VaiTon:
> 
> 
> Hi @VaiTon , does it keep our autocomplete with the taxonomy entries + recently used entries?

For now it's just for taxonomy entries. Recently used ones are not implemented yet.

hangy:
Wow, you're fast! 👍

VaiTon:
> 
> 
> Hi @VaiTon , does it keep our autocomplete with the taxonomy entries + recently used entries?

Now implemented in ec6ac2f42. Not tested though as atm I can't seem to start docker.

hangy:
> I can add tags (also from autocomplete) just fine, but get the following error if I try to edit a product that contains tags:

I've fixed that issue, but the stored recently used entries aren't used, as there is no suggestions displayed before someone starts typing. (And when they start typing the 'input' event handler is run.)

hangy:
> I've fixed that issue, but the stored recently used entries aren't used, as there is no suggestions displayed before someone starts typing. (And when they start typing the 'input' event handler is run.)

I think this is fixed in f6f2cc0217ecad2607fff661703663efe65d45c1. I also took the liberty of moving the Tagify initialization to an external JavaScript file in 4b56a7b1bb67208bfad6236036fd8ddee986d968. Are you OK with these changes?

stephanegigandet:
I deployed it on dev, (e.g. https://uk.openfoodfacts.dev ) but I have an error in the JS which is not handled:

image

That results in the product edit form scripts not being loaded / working.

It's this line:

const obj = JSON.parse(window.localStorage.getItem("po_last_tags"));

stephanegigandet:
The suggestions are not working on the dev server, I think it's because of the authentication.
Can we change the host for the suggests request? Use the current host instead of world.
https://world.openfoodfacts.dev/cgi/suggest.pl?lc=en&tagtype=categories&term=c

stephanegigandet:
I like the way the new tags look. :)

hangy:
> I deployed it on dev, (e.g. https://uk.openfoodfacts.dev ) but I have an error in the JS which is not handled:
> 
> image
> 
> That results in the product edit form scripts not being loaded / working.
> 
> It's this line:
> 
> const obj = JSON.parse(window.localStorage.getItem("po_last_tags"));

Apparently, your browser's local storage might be corrupted: https://stackoverflow.com/q/18877643/11963 According to most sources I could find, `NS_ERROR_FILE_CORRUPTED` isn't related to the JavaScript source itself.

> The suggestions are not working on the dev server, I think it's because of the authentication.
> Can we change the host for the suggests request? Use the current host instead of world.
> https://world.openfoodfacts.dev/cgi/suggest.pl?lc=en&tagtype=categories&term=c

Sure. I guess that shouldn't change the functionality at all. I applied this in 1a7e05d085715bc9ac7f63e375239653b8746af9.

> I like the way the new tags look. :)

👍

stephanegigandet:
> Apparently, your browser's local storage might be corrupted: https://stackoverflow.com/q/18877643/11963 According to most sources I could find, `NS_ERROR_FILE_CORRUPTED` isn't related to the JavaScript source itself.

Right, my point is that we should handle it gracefully. Right now it completely breaks the product edit form for me, I can't use it.

The stackoverflow post you link to suggests something like that:

try {
    setLocalStorageItem(key, value);
} catch(e) {
    if(e.name == "NS_ERROR_FILE_CORRUPTED") {
        showMessageSomehow("Sorry, it looks like your browser storage has been corrupted. Please clear your storage by going to Tools -> Clear Recent History -> Cookies and set time range to 'Everything'. This will remove the corrupted browser storage across all sites.");
    }
}

-> Quite frankly in this case, I would just disable that feature completely, instead of showing a message to the user. It's not that important a feature. But it clearly shouldn't break the whole product edit form.

stephanegigandet:
@hangy @VaiTon : I made a fix to make the product edit form work on my apparently broken firefox: https://github.com/openfoodfacts/openfoodfacts-server/pull/2710/commits/ec4a7072aa0d245dd7f296c2a50478a97163bbf7
I'm clearly not experimented with javascript so please review. :)

cgi/product_multilingual.pl
```diff
@@ -670,16 +670,17 @@ ($$)
...
-	let obj = JSON.parse(window.localStorage.getItem("po_last_tags"));
+${field}.on('add', function(e) {
+	const obj = JSON.parse(window.localStorage.getItem("po_last_tags"));
 
-	if (obj == null) {
+	const tag = e.detail.data.value;
+	if (obj === null) {
 		obj = {"${field}": [tag]}
-	} else if (obj["${field}"] == null) {
+	} else if (obj["${field}"] === null) {
```
VaiTon: Here i would also like to check if it's = to "" or to undefined.
>There are only six falsy values in JavaScript you should be aware of:
    false — boolean false
    0 — number zero
    “” — empty string
    null
    undefined
    NaN — Not A Number

html/js/product-multilingual.js
```diff
@@ -494,8 +494,111 @@ function update_display(imagefield, first_display) {
...
+  let obj
+    try {
+      obj = JSON.parse(window.localStorage.getItem("po_last_tags"));
+    } catch(e) {
+      if(e.name == "NS_ERROR_FILE_CORRUPTED") {
+        obj = null;
+      }
+    }
+
+  if (obj === null) {
```
VaiTon:
```suggestion
```

html/js/product-multilingual.js
```diff
@@ -494,8 +494,111 @@ function update_display(imagefield, first_display) {
...
+    return [];
```
VaiTon:
```suggestion
```

html/js/product-multilingual.js
```diff
@@ -494,8 +494,111 @@ function update_display(imagefield, first_display) {
...
+  }
```
VaiTon:
```suggestion
```

html/js/product-multilingual.js
```diff
@@ -494,8 +494,111 @@ function update_display(imagefield, first_display) {
...
+
+  return obj[tagfield] === null ? [] : obj[tagfield];
```
VaiTon:
```suggestion
  return obj[tagfield] == null ? [] : obj[tagfield];
```

html/js/product-multilingual.js
```diff
@@ -494,8 +494,111 @@ function update_display(imagefield, first_display) {
...
+      }
+
+      obj[el.id].unshift(tag);
+    }
+
+
+    try {
+      window.localStorage.setItem("po_last_tags", JSON.stringify(obj));
+    } catch(e) {
+      if(e.name == "NS_ERROR_FILE_CORRUPTED") {
```
VaiTon:
```suggestion
      if(err.name == "NS_ERROR_FILE_CORRUPTED") {
        // TODO: Tell the user
```

html/js/product-multilingual.js
```diff
@@ -494,8 +494,111 @@ function update_display(imagefield, first_display) {
...
+        obj[el.id].pop();
+      }
+
+      obj[el.id].unshift(tag);
+    }
+
+
+    try {
+      window.localStorage.setItem("po_last_tags", JSON.stringify(obj));
+    } catch(e) {
```
VaiTon:
```suggestion
    } catch(err) {
```

html/js/product-multilingual.js
```diff
@@ -494,8 +494,111 @@ function update_display(imagefield, first_display) {
...
+    }
+  });
+
+  input.on('add', function(e) {
+    let obj;
+
+    try {
+      obj = JSON.parse(window.localStorage.getItem("po_last_tags"));
+    } catch(e) {
+      if(e.name == "NS_ERROR_FILE_CORRUPTED") {
```
VaiTon:
```suggestion
      if(err.name == "NS_ERROR_FILE_CORRUPTED") {
```

html/js/product-multilingual.js
```diff
@@ -494,8 +494,111 @@ function update_display(imagefield, first_display) {
...
+      });
+    }
+  });
+
+  input.on('add', function(e) {
+    let obj;
+
+    try {
+      obj = JSON.parse(window.localStorage.getItem("po_last_tags"));
+    } catch(e) {
```
VaiTon:
```suggestion
    } catch(err) {
```

html/js/product-multilingual.js
hangy: The current eslint ruleset (based on some standard that I don't remember 🤷‍♂) would complain about the missing third `=`: https://eslint.org/docs/rules/no-eq-null

html/js/product-multilingual.js
```diff
@@ -494,8 +494,111 @@ function update_display(imagefield, first_display) {
...
+    input.settings.whitelist = obj[el.id]; // reset the whitelist
+  });
+
+  document.getElementById('product_form').addEventListener('submit', function() {
+    el.value = input.value.map((obj) => obj.value).join(',');
+  });
+}
+
+function get_recents(tagfield) {
+  let obj
```
hangy:
```suggestion
  let obj;
```

html/js/product-multilingual.js
VaiTon: If we want to keep that rule just create an if with === null or === undefined or === ""

html/js/product-multilingual.js
hangy: Wouldn't `return obj[tagfield] == null ? [] : obj[tagfield];` fail if `obj` was null?

html/js/product-multilingual.js
```diff
@@ -494,8 +494,112 @@ function update_display(imagefield, first_display) {
...
+
+      obj[el.id].unshift(tag);
+    }
+
+
+    try {
+      window.localStorage.setItem("po_last_tags", JSON.stringify(obj));
+    } catch(err) {
+      if(err.name == "NS_ERROR_FILE_CORRUPTED") {
+        // TODO: Tell the user
```
hangy: This shouldn't be left as a `TODO`.

hangy:
> @hangy @VaiTon : I made a fix to make the product edit form work on my apparently broken firefox: [ec4a707](https://github.com/openfoodfacts/openfoodfacts-server/commit/ec4a7072aa0d245dd7f296c2a50478a97163bbf7)

Is it really useful to catch this and pollute the code with this error handling if it only affects a really small portion of users (where some Firefox profile got broken)?

VaiTon:
> 
> 
> > @hangy @VaiTon : I made a fix to make the product edit form work on my apparently broken firefox: [ec4a707](https://github.com/openfoodfacts/openfoodfacts-server/commit/ec4a7072aa0d245dd7f296c2a50478a97163bbf7)
> 
> Is it really useful to catch this and pollute the code with this error handling if it only affects a really small portion of users (where some Firefox profile got broken)?

We could just surround the function with a catcher. In that way the code would be more readable and the user would still be alerted.

stephanegigandet:
> Is it really useful to catch this and pollute the code with this error handling if it only affects a really small portion of users (where some Firefox profile got broken)?

Well we have no idea how many users are affected or could be affected in the future, and the consequence is very bad: it makes the product edit form not usable at all. So yes, I do believe we need a way to make sure the product edit form works.

About the alert to the user: I'm not sure if that's a good idea. This is a very small feature (saving the last tags used), I don't think it's a good idea to pop up an alert that could confuse users. Let's just turn the feature off if for some reason localstorage is not available.

html/js/product-multilingual.js
stephanegigandet: Let's just turn the feature off whenever there is an exception, without telling the user.

html/js/product-multilingual.js
VaiTon: You're right, I'll re-review this.

Issue #2712, opened 2019-12-02
Donate banners for December


stephanegigandet:
Display some donate banner images, with an option to turn them off, with a cookie to save the setting.

This is a temporary thing, with hardcoded banners to start. We don't have enough time to create a system that works well with translations / translated images etc. and we will probably focus on a few languages only.

stephanegigandet:
Deployed.

PR #2713, opened 2019-12-02
rotating donate banners for december

stephanegigandet:
bug #2712 

This is temporary for December, the banners are hardcoded, as we are in a hurry and we need to display translated images. We will soon add some more banners for OFF's most used languages.


stephanegigandet:
> We could just use the English version for every language. 👍

Yes, once we are done adding a few more languages, I'll add English as the default for the others.

PR #2714, opened 2019-12-03
Add 'en:Cassoulets' parent to category 'en:Cassoulets with duck confit

raphael0202:
See https://world.openfoodfacts.org/category/cassoulets-with-duck-confit


stephanegigandet:
@raphael0202 : could you merge master in your branch? that should fix the failing test

PR #2724, opened 2019-12-05
Changes to soda categories

moon-rabbitOFF:

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -23310,6 +23310,16 @@ country:en:Spain
 wikidata:en:Q72305
 
 <en:Spanish cheeses
+<en:Cow cheeses
+en:Tetilla cheese
+de:Queso Tetilla
+es:Queso Tetilla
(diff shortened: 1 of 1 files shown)


taxonomies/categories.txt
```diff
@@ -1964,44 +1964,44 @@ ru:Тоник
 zh:通宁水
 wikidata:en:Q237315
 
-<en:Diet beverages
+<en:Artificially sweetened beverages
 <en:Sodas
-en:Diet sodas, diet soft drinks
-de:Kalorienarme Erfrischungsgetränke, Kalorienreduzierte Erfrischungsgetränke, Diät-Erfrischungsgetränke
-es:Sodas light
+en:Artificially sweetened sodas, diet sodas, diet soft drinks
```
stephanegigandet: I think we should discuss this on Slack in #taxonomies. I'd prefer to keep "Diet sodas" as the main name of the category.

PR #2741, opened 2019-12-06
Initialize Select2 dynamically

hangy:
I noticed that we currently send the whole list of countries for the dropdown on every request. Since users will rarely change the country/domain, that is somewhat wasteful as the country list is about 10 KiB before gzip, which can be ~30% of the whole page on rather empty pages. This PR moves loading of this data via AJAX, so that it only happens if a user interacts with the dropdown.

As an aside, it does the same with the list of languages on `product-multilingual.pl`, because the list of languages is otherwise sent to the client once for every tab group (4 times right now).


gulpfile.js
```diff
@@ -52,12 +52,18 @@ function js() {
...
+  ], { sourcemaps: true })
+  .pipe(dest('./html/js/dist', { sourcemaps: true }))
+}
+
 function jQueryUiThemes() {
   return src('./node_modules/jquery-ui-dist/**/*.css', { sourcemaps: true })
     .pipe(dest('./html/css/dist/jqueryui/themes/base', { sourcemaps: true }))
 }
 
-exports.js = js
```
VaiTon: Why did you remove exports? Without them we cannot run tasks one by one.

hangy: True - I actually never did that, so I supposed they weren't required. I'll put them back in. :)

stephanegigandet:
It's a very good idea, thanks @hangy !

One thing I'm concerned about though is the extra request for lang.pl that happens even if there is no interaction on the menu.

Queries to cgi scripts are expensive because Apache + mod_perl have to serve them instead of nginx. When we get a spike (TV etc.) it is problematic.

For the small dropdown of languages for the country in the menu bar, I think we should just generate it in the main HTML page as we did before.

Related to that, I'd really like to remove the call to /cgi/manifest.pl too. We could just generate static manifest files for all languages, and they would be served by nginx.

stephanegigandet:
The dist/product-multilingual.js file has an extra blob at the end, could you tell me what it is?

< //# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIj
ozLCJuYW1lcyI6W10sIm1hcHBpbmdzIjoiIiwic291cmNlcyI6WyJwcm9kdWN0LW11bHRpbGluZ3VhbC
5qcyJdLCJzb3VyY2VzQ29udGVudCI6WyIvLyBUaGlzIGZpbGUgaXMgcGFydCBvZiBQcm9kdWN0IE9wZW
5lci5cbi8vXG4vLyBQcm9kdWN0IE9wZW5lclxuLy8gQ29weXJpZ2h0IChDKSAyMDExLTIwMTkgQXNzb2

hangy:
> For the small dropdown of languages for the country in the menu bar, I think we should just generate it in the main HTML page as we did before.

OK, I'll look into a way to fix this. Loading the whole language file for a single word is obviously overkill, anyways.

> Related to that, I'd really like to remove the call to /cgi/manifest.pl too. We could just generate static manifest files for all languages, and they would be served by nginx.

Personally, I'd just go the easy way and let nginx cache those responses, instead of having another mount shared between nginx and apache, but I'll see what I can do.

> The dist/product-multilingual.js file has an extra blob at the end, could you tell me what it is?
> 
> < //# sourceMappingURL=data:application/json;charset=utf-8;base64,eyJ2ZXJzaW9uIj
> ozLCJuYW1lcyI6W10sIm1hcHBpbmdzIjoiIiwic291cmNlcyI6WyJwcm9kdWN0LW11bHRpbGluZ3VhbC
> 5qcyJdLCJzb3VyY2VzQ29udGVudCI6WyIvLyBUaGlzIGZpbGUgaXMgcGFydCBvZiBQcm9kdWN0IE9wZW
> 5lci5cbi8vXG4vLyBQcm9kdWN0IE9wZW5lclxuLy8gQ29weXJpZ2h0IChDKSAyMDExLTIwMTkgQXNzb

That's a [Source Map](https://blog.rapid7.com/2017/05/24/what-are-javascript-source-maps/), but actually it's not required in this case, because we currently don't minify these JavaScript files. I'll disable it.

stephanegigandet:
I installed the new version on https://fr.openfoodfacts.dev/ (I think I did everything correctly).
I still get a call to lang.pl , is it expected?

hangy:
> I installed the new version on https://fr.openfoodfacts.dev/ (I think I did everything correctly).
> I still get a call to lang.pl , is it expected?

No, that should not be called any more. The call from the homepage was removed in 77df5a719fd1a5fdcbce28afec167d9908cd7498, and the file itself was removed in 7283b92add0e3b92c6d5c394e4bf660d3ee573e8. Maybe you need to clean the `/cgi` and `/html/js/dist` directories?

stephanegigandet:
@hangy : I deployed this in production. For some reason the display.js was not generated in the js/dist/ directory, I ended up copying it from the js/ dir. I think I copied everything and I ran npm install + npm run build

One thing I didn't notice: when a country is selected, we used to have a X sign to remove the country, and go to the world site. Would it be possible to put something like this back?

e.g. it's still on https://uk.openbeautyfacts.org/ -> You can see the currently selected country, with a X to remove it.

Ban3:
@stephanegigandet I'm still getting 404 on product-multilingual.js on production.

stephanegigandet:
@ban3 : right, I fixed it by deploying @hangy 's fix.

But I do have CORS issues that prevent the script to load. :-( 

````
jquery.js:26 Access to XMLHttpRequest at 'https://static.openfoodfacts.org/data/i18n/fr/lang.json' from origin 'https://fr.openfoodfacts.org' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
send @ jquery.js:26
ajax @ jquery.js:25
lang @ display.js:59
initLanguageAdding @ product-multilingual.js:843
(anonymous) @ product-multilingual.js:838
(anonymous) @ product-multilingual.js:840
product-multilingual.js:844 Uncaught TypeError: Cannot read property 'add_language' of undefined
    at initLanguageAdding (product-multilingual.js:844)
    at product-multilingual.js:838
    at product-multilingual.js:840
initLanguageAdding @ product-multilingual.js:844
(anonymous) @ product-multilingual.js:838
(anonymous) @ product-multilingual.js:840
product-multilingual.js:677 Uncaught ReferenceError: admin is not defined
    at Object.<anonymous> (product-multilingual.js:677)
    at Function.each (jquery.js:14)
    at HTMLDivElement.<anonymous> (product-multilingual.js:668)
    at Function.each (jquery.js:14)
    at _.fn.init.each (jquery.js:14)
    at _.fn.init.show (product-multilingual.js:661)
    at _.fn.init.$.fn.selectcrop (product-multilingual.js:825)
    at HTMLDocument.<anonymous> (product.pl?type=edit&code=7610100088056:1698)
    at k (jquery.js:24)
    at Object.fireWith [as resolveWith] (jquery.js:24)

stephanegigandet:
ok CORS fixed, but I still have an issue with an undefined admin var

				if ((stringStartsWith(id, 'manage')) && (admin)) {

stephanegigandet:
Fixed: https://github.com/openfoodfacts/openfoodfacts-server/pull/2770

Issue #1737, opened 2019-04-15
NOVA groups fields have inconsistent types (int or string)


stephanegigandet:
The nova_group field can be either an int (with values 1, 2, 3, 4) or a string (with values "1", "2", "3", "4"), which causes problems in apps etc.  (bug https://github.com/openfoodfacts/openfoodfacts-ios/issues/273 )

There are several fields related to NOVA:

$product_ref->{nova_group} can currently be either int and string, and other  fields are computed from it:

$product_ref->{nutriments}{"nova-group"} = $product_ref->{nova_group};
$product_ref->{nutriments}{"nova-group_100g"} = $product_ref->{nova_group};
$product_ref->{nutriments}{"nova-group_serving"} = $product_ref->{nova_group};

$product_ref->{nova_groups} = $product_ref->{nova_group};
$product_ref->{nova_groups_tags} = [ canonicalize_taxonomy_tag("en", "nova_groups", $product_ref->{nova_group}) ];

Proposed solution:

Make  $product_ref->{nova_group} always an int, same nova-group_100g and nova-group_serving.

Make $product_ref->{nutriments}{"nova-group"} and nova_groups strings.

This is consistent with the fields for the normal nutrients.

The Android app currently works with both int and string, but the current iOS app expects a string. We will wait for the new iOS app to accept both before deploying the change.

stephanegigandet:
Related to #772

stephanegigandet:
Deployed, and updating all products.

PR #2754, opened 2019-12-08
make nova_group a int, and nova_groups a string, bug #1737

stephanegigandet:
solves bug #1737 and makes ios app work (hopefully)


teolemon:
https://github.com/openfoodfacts/openfoodfacts-androidapp/search?q=nova_groups&unscoped_q=nova_groups

PR #2758, opened 2019-12-09
update German donation banners, display English as default

stephanegigandet:


teolemon:
common.pot is already good. So LGTM.

PR #2765, opened 2019-12-10
Run buildJs task in gulp

hangy:
The `buildJs` task needs to be run in the `default` task to produce `html/js/dist/display.js` and it's friends.


hangy:
I think I missed this in a merge. 😞

PR #2767, opened 2019-12-10
Diet for jQuery UI

### LINKED ISSUE #215: Minify Javascript files
`Minify JavaScript for the following resources to reduce their size by 105.7KiB (16% reduction).`
`Minifying http://static.openfoodfacts.org/foundation/js/foundation.min.js could save 67.2KiB (35% reduction).`
`Minifying http://static.openfoodfacts.org/js/jquery.fileupload.js could save 24.1KiB (65% reduction).`
`Minifying http://static.openfoodfacts.org/js/jquery.iframe-transport.js could save 5.4KiB (71% reduction).`
`Minifying http://static.openfoodfacts.org/js/jquery.fileupload-ip.js could save 3.8KiB (62% reduction).`
`Minifying http://slack.openfoodfacts.org/slackin.js could save 2.3KiB (36% reduction).`
`Minifying https://static.xx.fbcdn.net/…rc.php/v2iaFb3/yS/l/en_US/s9fFMJVTKYC.js could save 1.1KiB (2% reduction) after compression.`
`Minifying http://static.openfoodfacts.org/js/jquery-ui-1.11.4/jquery-ui.min.js could save 690B (1% reduction).`
`Minifying https://oauth.googleusercontent.com/…e:rpc:shindig.random:shindig.sha1.js?c=2 could save 661B (3% reduction) after compression.`
`Minifying http://static.openfoodfacts.org/foundation/js/vendor/jquery.js could save 518B (1% reduction).`

hangy:
- Instead of taking pre-minified versions of JS libs from npm, run them through the same minifier with Gulp (since they can be different)
- Create a partial jQuery UI version as suggested in https://github.com/openfoodfacts/openfoodfacts-server/issues/2675#issuecomment-563973740
- Also run our own JS scripts through the minification process (actually saves a bunch of KiB)
- (Re-)Enable Source Maps for everything, but force them to be written to external files (so browsers don't have load this huge comment chunk)

I believe this also fixes #215

### DIFF
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -728,7 +728,7 @@ HTML
 <script type="text/javascript" src="/js/dist/tagify.js"></script>
 <script type="text/javascript" src="/js/dist/jquery.iframe-transport.js"></script>
 <script type="text/javascript" src="/js/dist/jquery.fileupload.js"></script>
-<script type="text/javascript" src="/js/dist/load-image.all.js"></script>
+<script type="text/javascript" src="/js/dist/load-image.all.min.js"></script>
 <script type="text/javascript" src="/js/dist/canvas-to-blob.js"></script>
 <script type="text/javascript" src="/js/dist/product-multilingual.js"></script>
 HTML
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -725,12 +725,11 @@ HTML
 <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/cropper/2.3.4/cropper.min.js"></script>
 <script type="text/javascript" src="/js/jquery.tagsinput.20160520/jquery.tagsinput.min.js"></script>
 <script type="text/javascript" src="/js/jquery.form.js"></script>
-<script type="text/javascript" src="/js/dist/tagify.min.js"></script>
+<script type="text/javascript" src="/js/dist/tagify.js"></script>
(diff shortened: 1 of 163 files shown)


VaiTon:
>     * Instead of taking pre-minified versions of JS libs from npm, run them through the same minifier with Gulp (since they can be different)
Why? Do you want to create a bundle.js file?

hangy:
> > ```
> > * Instead of taking pre-minified versions of JS libs from npm, run them through the same minifier with Gulp (since they can be different)
> > ```
> 
> Why? Do you want to create a bundle.js file?

No - at least that's not what I had in mind. A bundle (or two, depending on the viewed sites) could make sense at some point in time, but I basically wanted to ensure that all scripts were minified in a good way. For example, some of the leaflet CSS files weren't minified before at all. I'm definitely not set on this decision, but I'm fairly certain it won't hurt (except for a few seconds of Gulp build time).

hangy:
Also, minifying already minified scripts could cause weird results (ie. unusable source maps). It's way easier to run all is/CSS through the same minifier than having several branches/gulp tasks …

stephanegigandet:
There are many changes, so I deployed it on https://fr.openfoodfacts.dev/ so that we can test everything.

hangy:
> There are many changes, so I deployed it on https://fr.openfoodfacts.dev/ so that we can test everything.

As far as I can tell, the touched jQuery UI stuff still works as expected. 🙂

PR #2770, opened 2019-12-11
fix issues with product edit script

stephanegigandet:
CORS issue + missing admin var.


stephanegigandet:
@hangy : I made more changes to the off nginx config, as there were more issues with CORS that prevent the app from downloading taxonomies.

It was this line that affected the /data dir:
location ~ ^/(.well-known|images|css|js|rss|data|files|resources|foundation|bower_components)/ {

I also removed the cache that @cquest had put one year ago, we had some issues with it so it had been removed from production but was still in github..

PR #2772, opened 2019-12-11
Normalize taxonomies

### LINKED ISSUE #2703: Normalize taxonomy directory structure
### Issue
Right now obf, opff and opf taxonomies are very messy. I guess they aren't even loaded.

`Tags.pm` is what deals with taxonomies in Product Opener and in `retrieve_tags_taxonomy()` **we only load open food facts taxonomies**:
https://github.com/openfoodfacts/openfoodfacts-server/blob/f4774d12cf833c3a94b408e91f4561d767f851de/lib/ProductOpener/Tags.pm#L1577

### Proposal
We should make rearrange the directory structure like this:
```
taxonomies
|-- openfoodfacts
|   |-- labels.txt
|   +-- etc...
|-- openbeautyfacts
|   |-- labels.txt
|   +-- etc...
|-- openpetfoodfacts
|   |-- labels.txt
|   +-- etc...
|-- openproductfacts
|   |-- labels.txt
|   +-- etc...
|-- labels.common.txt
|-- categories.common.txt
+-- ....common.txt
```
And then, when we build taxonomies, we should combine common taxonomies with site-specific ones and put them in a dir like `taxonomies/openfoodfacts/build`.

Then at software startup only relevant taxonomies would be loaded.

VaiTon:
Closes #2703.


teolemon:
This honestly looks a bit risky and a bit massive. Also there are now merge conflicts.

PR #2779, opened 2019-12-12
Update openpetfoodfacts.pot

teolemon:


po/openpetfoodfacts/openpetfoodfacts.pot
```diff
@@ -80,7 +80,7 @@ msgid   "OpenFoodFacts"
 msgstr  ""
 
 msgctxt "warning_not_complete"
-msgid "This product page is not complete. You can help to complete it by editing it and adding more data from the photos we have, or by taking more photos using the app for <a href=\"https://play.google.com/store/apps/details?id=org.openpetfoodfacts.scanner\">Android</a> or <a href=\"https://apps.apple.com/us/app/open-beauty-facts/id1122926380\">iPhone/iPad</a>. Thank you!"
```
VaiTon: @teolemon I don't think the site needs to know the language, it can just figure it out by itself.

teolemon:
@hangy not strictly required, but a good way to ensure those pages appear in the right language in corner cases, are indexable by Search Engines…
I would rather turn that into a variable, and extract it from the HTML and PO files to be honest

PR #2780, opened 2019-12-13
Don't add errored parsed results

### LINKED ISSUE #2777: Error on the Translators table
image
https://de.openfoodfacts.org/cgi/top_translators.pl

hangy:
**Description:**
Only add rows to the `dataSource` if there were no errors parsing the CSV row.

**Related issues and discussion:** Fixes #2777

### DIFF
--- a/cgi/top_translators.pl
+++ b/cgi/top_translators.pl
@@ -63,6 +63,12 @@ my $js = <<JS
 				if (results.errors.length === 0) {
 					dataSource = dataSource.concat(results.data);
 				}
+				else {
+					for (var i = 0; i < results.errors.length; ++i) {
+						var error = results.errors[0];
+						console.warn('%s error %s while parsing CSV row %d: %s', error.type, error.code, error.row, error.message);
+					}
+				}
 			},
 			complete: function () {
 				\$('#top_translators').DataTable({
--- a/cgi/top_translators.pl
+++ b/cgi/top_translators.pl
@@ -60,9 +60,9 @@ my $js = <<JS
 			skipEmptyLines: true,
 			withCredentials: false,
 			step: function(results, parser) {
-				dataSource = dataSource.concat(results.data);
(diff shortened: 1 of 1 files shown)


VaiTon:
@hangy what happens if there are errors?

hangy:
> @hangy what happens if there are errors?

The papaparse returns the last row as containing errors, because of missing columns. The row object `results.data` only contains the `Name` property with `null` as the value. All other properties/columns are missing, which in turn causes DataTables to trip over the data. This is only exposed by `results.errors` containing an error object.

VaiTon:
> 
> 
> > @hangy what happens if there are errors?
> 
> The papaparse returns the last row as containing errors, because of missing columns. The row object `results.data` only contains the `Name` property with `null` as the value. All other properties/columns are missing, which in turn causes DataTables to trip over the data. This is only exposed by `results.errors` containing an error object.

So, if I understand correctly, errored results aren't added to the table. Does it show an error message in the console? We could do that to inform the power user that something went wrong.

hangy:
> So, if I understand correctly, errored results aren't added to the table. Does it show an error message in the console? We could do that to inform the power user that something went wrong.

ACK. I've added a warning log message in dc0f4ecef0a65707a795d2dab78bf2fd547a5d9b.

cgi/top_translators.pl
```diff
@@ -63,6 +63,12 @@
 				if (results.errors.length === 0) {
 					dataSource = dataSource.concat(results.data);
 				}
+				else {
+					for (var i = 0; i < results.errors.length; ++i) {
```
VaiTon: I would use a forEach:
```js
results.errors.forEach(error => {
  console.warn("...");
}
```

hangy: There's no huge advantage over the `for` loop, and arrow functions don't work in a few browsers. I'll leave it as-is for now.

PR #2792, opened 2019-12-17
Producers - Import Système U + improvements

stephanegigandet:
Includes also:
- better normalization of quantity in normalize_quantity()


teolemon:
FYI @raphael0202 the function that converts Nutrition text into values

PR #2810, opened 2019-12-22
docs: Document API fields

teolemon:
### What
Document more API fields.
### Why
- Be more welcoming

### Part of
- #5559


stephanegigandet:
I don't understand what users are supposed to do with this list? Most of the fields in it are already documented in the sections above, and the new ones need to be incorporated in those sections, based on the type of field.

PR #2740, opened 2019-12-06
DO NOT MERGE - New Crowdin translations

teolemon:
Enjoy!

PR #2824, opened 2020-01-02
Revert "DO NOT MERGE - New Crowdin translations"

teolemon:
Reverts openfoodfacts/openfoodfacts-server#2740


teolemon:
The PR that was merged was not wrong per-se, but it wasn't squashed.

VaiTon:
> 
> 
> Reverting it now won't help. We can't undo this without rewriting Git history and force-pushing to master.

Quoting

teolemon:
It seems that this causes many issues on a public master with forks. Should I close this PR then ?

VaiTon:
> 
> 
> It seems that this causes many issues on a public master with forks. Should I close this PR then ?

Yup.

PR #2832, opened 2020-01-05
Remove %wiki_texts

### LINKED ISSUE #521: Expand automatic import from wiki to new languages
product-opener / lib / ProductOpener / Config_off.pm
lines 152 to 168 
add pages marked as done
http://en.wiki.openfoodfacts.org/Translations_-_Discover_page
http://en.wiki.openfoodfacts.org/Translations_-_Contribute_page
http://en.wiki.openfoodfacts.org/Translations_-_Press

hangy:
**Description:**
Removes the `%wiki_texts` variable and related scripts. All of these translations are now done through Crowdin, and the translations on the Wiki are massively outdated.

**Related issues and discussion:** Closes #521


hangy:
We should probably also remove the translations from the wikis to avoid confusion on where the translations should be done.

PR #2833, opened 2020-01-05
Initialize %emb_codes_* on demand

### LINKED ISSUE #2204: Tags.pm loads geofiles every time it is referenced.
**Describe the bug**
When this module is loaded it takes a few seconds and prints to STDERR that it loaded two geofiles.  This slows down development and I am guessing is not intended or optimal behavior.

**To Reproduce**

```
Nicos-MacBook-Pro:lib nico$ perl -c -I . ProductOpener/Tags.pm
Tags.pm - loading geofile villes-geo-france-galichon-20130208.csv
Tags.pm - loading geofile villes-geo-france-complement.csv
...
ProductOpener/Tags.pm syntax OK
```
**Expected behavior**
Module should be loaded without calling this routine.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Mac OSX Mojave

**Additional context**
Maybe this is necessary behavior, please let me know.

hangy:
**Description:**
Instead of loading the packager codes from CSV every time the `ProductOpener::Tags` module is loaded, load it only on demand. Also, scripts that use the geofiles (incl. the Apache startup script) eagerly load the data to avoid slowdowns during initial execution.

This should speed up all scripts that don't use the geodata, including the validation script compiles the Perl files during test.

**Related issues and discussion:** Fixes #2204


lib/ProductOpener/Tags.pm
```diff
@@ -1953,9 +1954,9 @@ sub display_tag_link($$) {
 	}
 
 	if ($tagtype eq 'emb_codes') {
-
 		my $city_code = get_city_code($tagid);
 
+		init_emb_codes();
```
stephanegigandet: I think we should remove the init_emb_codes() calls inside functions that may need %emb_codes, and assume it has been loaded by startup.pl our the scripts like update_all_products.pl etc. that need it. That will remove a huge number of function calls.

We could also initialize %emb_codes to an empty hash instead of being undefined, that way the functions would not die with errors if they are called without init_emb_codes() having been called.

VaiTon: > 
> 
> I think we should remove the init_emb_codes() calls inside functions that may need %emb_codes, and assume it has been loaded by startup.pl our the scripts like update_all_products.pl etc. that need it. That will remove a huge number of function calls.

I agree with you.

> We could also initialize %emb_codes to an empty hash instead of being undefined, that way the functions would not die with errors if they are called without init_emb_codes() having been called.

I would prefer if it threw an error. That's because we would know there is something wrong, instead of just having no `emb_codes`.

hangy: It's a few API calls, but (according to my test logs) not too many, since the sub is only called s few times. I'll revisit the PR later this week.

stephanegigandet: I changed my mind. :) It's actually breaking a lot of scripts (all the ones that update products in some way). I'll replace the die statements by a call to init_emb_codes() if it's not defined.

PR #2836, opened 2020-01-07
feat: Power User Script online version

teolemon:
### What
- copy pasting work done by @CharlesNepote today
- Start of a solution for #5561


hangy:
Do you want to maintain the user script in this repository from now on?

stephanegigandet:
@hangy : the idea is to make it available directly in Product Opener, for some users, without requiring installing things.

@CharlesNepote @teolemon : live on https://fr.openfoodfacts.dev/ for admins only right now, we may want to think of how to surface it, e.g. make it an option in the user profile or something.

CharlesNepote:
On the one hand the Power User Script is useful for some people but:
* it needs manual install for each user
* it needs manual deployment for each user when updated
* we have an enhanced version (in this PR) which allow mass editing, but we think it's not good to give it for everybody.

So we would like to try if Power User Script can be launched directly on the server side, but reserved for admins and moderators (due to mass edit). Also it would be a good thing to make it optional for users: an option should allow each users to activate it or not.

On the other hand:
* it will be more difficult for me to maintain the enhanced script because it needs far more work: launch docker, maintain a recent database of OFF products, etc. Nowadays it's very easy to me to code this script: just launching the browser that's all. Maybe I'll continue to code it as a Greasemonkey script, and publish new functionalities on Product Opener from time to time.
* the script could conflict with other things and deactivate it will be done by the script itself (not yet coded), which is not very clean.

IMHO what is sure is that we need easier ways to prototype user scripts / enhanced UI / test UIs / "labs" functionalities. Wikipedia allows users script. We should find a very simple way to do it: both easy for front-end devs and users.

teolemon:
I just rebased the branch

lib/ProductOpener/Display.pm
```diff
@@ -7193,6 +7201,10 @@ HTML
 	my $html;
 
 	# init javascript code
+  # Kept those 3 lines during the rebase. Not evident if needed @CharlesNepote
```
teolemon, line 7204: @CharlesNepote I rebased the Power User Script branch. but I was not sure about those 3 lines. Can I delete them ?

stephanegigandet:
@teolemon @CharlesNepote I'm going through old PRs. This one is 4 years old. I suggest that we close it, and that we decide on what we want to do first, as there are many different ways we could make power user script, or part of it, more easily available to some users.

teolemon:
Fine by me. I will prioritize the associated issue.

PR #2845, opened 2020-01-10
tagline-for-ios

teolemon:
Create a separate tagline so as to be able to send specific messages to iOS users (and assess the # of app opens)


stephanegigandet:
Can you put it in /html/files/app/tagline/ ?

teolemon:
done @stephanegigandet

PR #2846, opened 2020-01-10
fix allergen parsing bug #2719

### LINKED ISSUE #2719: Error from detect_allergens_from_text results in garbled ingredients list
```perl
ingredients_text_fr => "Filet de saumon sauvage certifié MSC, pêché en Pacifique Nord-est (100%)",
```
results in:
```perl
'ingredients_text_with_allergens_fr' => "Filet de saumon sauvage certifi\x{e9} MSC,estp\x{ea}ch\x{e9} en Pacifique Nord- (100%)",
```

stephanegigandet:
Fixes #2719

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -3233,11 +3233,12 @@ sub replace_allergen_between_separators($$$$$$) {
 
 	my $field = "allergens";
 
-	# print STDERR "replace_allergen_between_separators - allergen: $allergen\n";
+	#print STDERR "replace_allergen_between_separators - allergen: $allergen\n";
 
 	my $stopwords = $allergens_stopwords{$language};
 
 	my $before_allergen = "";
+	my $after_allergen = "";
 
 	# Remove stopwords at the beginning or end
 	if (defined $stopwords) {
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -3325,7 +3326,7 @@ sub detect_allergens_from_text($) {
 
 			# allergens between underscores
 
-			# print STDERR "current text 1: $text\n";
+			print STDERR "current text 1: $text\n";
```
VaiTon: We shouldn't use STDERR

lib/ProductOpener/Ingredients.pm
```diff
@@ -3344,7 +3345,7 @@ sub detect_allergens_from_text($) {
 
 			# allergens between separators
 
-			#print STDERR "current text 2: $text\n";
+			print STDERR "current text 2: $text\n";
```
VaiTon: We shouldn't use STDERR

lib/ProductOpener/Ingredients.pm
stephanegigandet: It's just for debugging/coding, that's why it's commented out. When debugging specific things in the code, I find it much easier than trying to get only the messages I want through log4perl.

VaiTon: I don't think leaving it in the production code would be a good idea. It clutters the console

stephanegigandet: Ah right I didn't intend to leave them, thanks for catching that, sorry I misunderstood the first time.

lib/ProductOpener/Ingredients.pm
stephanegigandet: I commented it out. Can you approve the pull request?

PR #2850, opened 2020-01-10
don't wait for robotoff if it is unresponsive

stephanegigandet:
set timeout to 2 sec.


teolemon:
best branch name ever @stephanegigandet

PR #2858, opened 2020-01-16
fixes for systeme u import

stephanegigandet:
Support parsing plain text nutrition facts with serving size. + test

### DIFF
--- a/lib/ProductOpener/Producers.pm
+++ b/lib/ProductOpener/Producers.pm
@@ -381,7 +381,7 @@ my %fields_synonyms = (
 
 en => {
 	lc => ["lang"],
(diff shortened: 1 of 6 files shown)


lib/ProductOpener/Producers.pm
```diff
@@ -421,6 +421,38 @@ fr => {
 
 );
 
+my %per_regexps = (
```
hangy: I'm not quite sure why this is named `per_regexps`, as it doesn't container regular expressinos right now?

stephanegigandet: Good point, I changed my mind and forgot to rename the field. :) I will rename it.

PR #2869, opened 2020-01-22
Enable producers platform moderators user group to view/edit products from different users and orgs

stephanegigandet:
This change is to decouple the $Owner_id of the product (org-someorg or user-someuser) from the actual $User_id, so that users in a new platform moderator group can see / edit products owned by any organization or user on the producers platform.


lib/ProductOpener/Users.pm
```diff
@@ -413,6 +412,41 @@ EMAIL
...
+sub check_edit_owner($$) {
+
+	my $user_ref = shift;
+	my $errors_ref = shift;
+
+	$user_ref->{pro_moderator_owner} = get_string_id_for_lang("no_language", remove_tags_and_quote(param('pro_moderator_owner')));
+
+	if ((not defined $user_ref->{pro_moderator_owner}) or ($user_ref->{pro_moderator_owner} eq "")) {
+		delete $user_ref->{pro_moderator_owner};
+		# Also edit the current user bbject so that we can display the current status directly on the form result page
```
teolemon:
```suggestion
		# Also edit the current user object so that we can display the current status directly on the form result page
```

PR #2870, opened 2020-01-23
call init_emb_codes() on demand, bug #2204

stephanegigandet:
Sorry @hangy , I changed my mind: it's too dangerous to expect all callers to have called init_emb_codes(), nearly all the scripts in /scripts actually need it. So let's put back an initialization on demand, but keeping the unless clause so that the function is called only once.


stephanegigandet:
Related discussion on https://github.com/openfoodfacts/openfoodfacts-server/pull/2833

Issue #2872, opened 2020-01-24
Handle traces specified in the allergens field


stephanegigandet:
If the allergens field contains something like "Mustard and celery. Possible traces of gluten, fish and soy", split the string to put "Mustard and celery" in the allergens field and "gluten, fish and soy" in the traces field.

In addition, the automatic split of bug #2871 will enable us to have "gluten", "fish", "soy" in traces instead of "fish and soy".

Some producers are sending us mixed data in the allergens field:

"Céleri, Œufs, Poisson, Gluten, Lait, Fruits à coque, Soja, GLUTEN. TRACES POTENTIELLES: CRUSTACÉS, MOUTARDE ET SULFITES."

stephanegigandet:
Deployed.

Issue #2871, opened 2020-01-24
Automatically split "X and Y" tags if X and Y exists in the taxonomy while "X and Y" does not


stephanegigandet:
There are some splits already done for ingredients (bug #1927), this is to generalize it for other taxonomies like allergens, labels etc. so that we can automatically split entries like "Organic and GMO free" for labels, "Mustard and crustaceans" for allergens etc.

The change will be made in Tags::gen_tags_hierarchy_taxonomy()

If the combined entry does exist in the taxonomy, then the split is not made. Similarly, the split is made only if the split entries exist.

stephanegigandet:
Deployed.

PR #2873, opened 2020-01-24
split "X and Y" tags #2871, handle traces in allergens #2872

stephanegigandet:
Enable #2871 and #2872


lib/ProductOpener/Tags.pm
```diff
@@ -1765,6 +1765,14 @@ sub gen_tags_hierarchy($$) {
 	return @sorted_list;
 }
 
+my %and = (
+	en => " and ",
+	de => " und ",
+	es => " y ",
+	fi => " ja ",
+	fr => " et ",
+	it => " e ",
```
teolemon:
```suggestion
	it => " e ",
	da => " og ",
	nl => " en ",
	cs => " a ",
	fi => " ja ",
	pt => " e ",
```

teolemon: https://www.indifferentlanguages.com/words/and

stephanegigandet: I'll add them, don't commit it.

Issue #442, opened 2016-08-30
Create a Sitemap for ProductOpener


teolemon:
Goals
- Increase indexation speed for new pages (eg new Packager codes that were fixed)
- Ease the burden on the server

VaibhavNarayann:
Hey there, is this issue still open? If not, could anyone please guide me to some other issues that I can work on?

teolemon:
<img width="1412" height="844" alt="Image" src="https://github.com/user-attachments/assets/b425a779-63be-4cbe-9c2b-25dd725d836d" />

<img width="1412" height="844" alt="Image" src="https://github.com/user-attachments/assets/d02802fb-9a40-4248-87b6-f13abdb6140a" />

PR #2878, opened 2020-01-26
feat: Add sitemap for taxonomized fields

hangy:
### What
Adds a script for generating sitemaps of taxonomized fields, packager codes, and products.

### Related issues and discussion
- #442


hangy:
Note that we don't link to this sitemap anywhere. It might be interesting to ping some search engines with the generated sitemap after building it, because we cannot really add the sitemaps to our shared `robots.txt` file (since it seems to require absolute URLs and needs to be hosted on the same subdomain as the content).

stephanegigandet:
The code looks good. How many urls does it generate? I'm a bit worried that we will generate a sitemap with millions of URLs. A large proportion of them may in fact be empty pages with 0 results, as we have few products for 100 countries.

hangy:
> How many urls does it generate?

One URl for each product and each taxonomy tag in each known country/language combination (only the country's official language).

> I'm a bit worried that we will generate a sitemap with millions of URLs.

The `Web::Sitemap` module takes care of the sitemap size. It creates a sitemap index for every cc/lc combo, and each sitemap index can contain up to 50.000 URLs. As the sitemap index can contain up to 50.000 sitemap URLs, sitemap size shouldn't be a problem.

> A large proportion of them may in fact be empty pages with 0 results, as we have few products for 100 countries.

This just creates a sitemap entry, regardless of any products using a taxonomy tag in a country or not. I'm not sure if this poses an issue, if the URL linked in the sitemap doesn't contain any products. At worst, it's that the "no products" pages will be redundant content in search engines, but is that problematic?

stephanegigandet:
The problem is the number of urls generated and referenced in the sitemaps. I just ran it, for one country and one language, it's already 79k urls. We have 333 combinations of countries x official languages. That's 26 millions urls. We don't really want any crawler to actually crawl all those urls. At a rate of one query per second, it would take 300 days... 

And that's just for the tags, not even counting the products sitemaps (which for some reason did not work for me, I don't get products sitemap generated).

I get some warnings during execution, not sure if they matter:

products ad
DB connection already exists
products ad - done
unlink0: /tmp/LYvT_Hw5Sx is gone already at /usr/local/share/perl/5.24.1/Web/Sitemap.pm line 284.
unlink0: /tmp/U0Qge3b5UX is gone already at /usr/local/share/perl/5.24.1/Web/Sitemap.pm line 284.
unlink0: /tmp/0rgjgtnZgH is gone already at /usr/local/share/perl/5.24.1/Web/Sitemap.pm line 284.
unlink0: /tmp/wfXhyiNxLN is gone already at /usr/local/share/perl/5.24.1/Web/Sitemap.pm line 284.
unlink0: /tmp/2Bcfpzydqa is gone already at ./generate_sitemaps.pl line 164.
unlink0: /tmp/O_kHJp1iLg is gone already at ./generate_sitemaps.pl line 164.
unlink0: /tmp/HdEind08G1 is gone already at ./generate_sitemaps.pl line 164.
unlink0: /tmp/MQj2dy1dSa is gone already at ./generate_sitemaps.pl line 164.
products kp

So we can definitely merge this, but I think we should not make those sitemaps publicly available.

VaiTon:
What about selecting only the labels/categories/etc with product count > 0 ?
cc @hangy @stephanegigandet

teolemon:
```So we can definitely merge this, but I think we should not make those sitemaps publicly available.``` 
Can we do so ? @stephanegigandet

teolemon:
One consensual thing would be to have the sitemap.xml for all products in all translated languages, or for languages available for that country + english ?

teolemon:
Perhaps not for facets, but being able to notify Google about added and updated product page would be great

Issue #2895, opened 2020-02-03
Unmatched parenthesis in regexp for ingredients parsing


stephanegigandet:
Products with ingredients in Russian cannot be saved because of a bug in one regexp:

Unmatched ( in regex; marked by <-- HERE in m/^(.*)\bÐ˜Ð½Ð³Ñ€ÐµÐ´Ð¸ÐµÐ½Ñ‚Ñ‹( <-- HERE (\s*)(-|:|\r|\n)+(\s*)/ at /srv/off/lib/ProductOpener/Ingredients.pm line 2234.

ru => [
'Состав',
'Ингредиенты(',
],

I will add tests to make sure we catch future errors like this one.

stephanegigandet:
Added a test:

not ok 154
#   Failed test at ProductOpener/Ingredients.pm line 2258.
Wide character in print at /usr/local/share/perl/5.24.1/Test2/Formatter/TAP.pm line 124.
#          got: 'Unmatched ( in regex; marked by <-- HERE in m/Ингредиенты( <-- HERE / at ProductOpener/Ingredients.pm line 2256.
# '
#     expected: ''

stephanegigandet:
Suggested by @hangy : use qr// to compile regular expressions.

PR #2896, opened 2020-02-03
fix regular expression and add tests, bug #2895

stephanegigandet:
This bug prevents products with ingredients in Russian to be saved.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -114,6 +114,7 @@ use LWP::UserAgent;
 use Encode;
 use JSON::PP;
 use Log::Any qw($log);
+use Test::More;
 
 # MIDDLE DOT with common substitutes (BULLET variants, BULLET OPERATOR and DOT OPERATOR (multiplication))
 my $middle_dot = qr/(?:\N{U+00B7}|\N{U+2022}|\N{U+2023}|\N{U+25E6}|\N{U+2043}|\N{U+204C}|\N{U+204D}|\N{U+2219}|\N{U+22C5})/i;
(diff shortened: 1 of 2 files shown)


hangy:
I always wondered why the regexes in that file are not stored as compiled regexes in the first place. Doing so should enable validation of the regex through the Perl compiler.

stephanegigandet:
> I always wondered why the regexes in that file are not stored as compiled regexes in the first place. Doing so should enable validation of the regex through the Perl compiler.

They are not the complete regexes though, only parts of them.
They are combined with other parts after:

e.G.
`````
if ((defined $phrases_before_ingredients_list{$language}) and (defined $product_ref->{"ingredients_text_$language"})) {

			foreach my $regexp (@{$phrases_before_ingredients_list{$language}}) {
				if ($product_ref->{"ingredients_text_$language"} =~ /(\s*)\b$regexp(\s*)(-|:|\r|\n)+(\s*)/is) {

hangy:
> > I always wondered why the regexes in that file are not stored as compiled regexes in the first place. Doing so should enable validation of the regex through the Perl compiler.
> 
> They are not the complete regexes though, only parts of them.
> They are combined with other parts after:
> 
> e.G.
> 
> ```
> if ((defined $phrases_before_ingredients_list{$language}) and (defined $product_ref->{"ingredients_text_$language"})) {
> 
> 			foreach my $regexp (@{$phrases_before_ingredients_list{$language}}) {
> 				if ($product_ref->{"ingredients_text_$language"} =~ /(\s*)\b$regexp(\s*)(-|:|\r|\n)+(\s*)/is) {
> ```

That should not be a problem, see https://github.com/openfoodfacts/openfoodfacts-server/blob/2df4bf58aca9cc62f7b02573d287053e7f328b86/lib/ProductOpener/Ingredients.pm#L137

stephanegigandet:
OK, it's something we can explore. I'll merge this as it fix crashes in production, we can try after to use qr//

PR #2897, opened 2020-02-03
Many improvements for traces and allergens

### LINKED ISSUE #2890: Handle "may contain traces of..." specified in the traces field
Related to #2872 , things like "Peut contenir des traces de lait et d'autres fruits à coques" in the traces field should be handled correctly, and also correctly split (bug #2871).

stephanegigandet:
Fixes #2890
Moved some strings from Ingredients.pm to the stopwords in the allergens taxonomy.
Added more tests.


aleene:
Maybe the allergens taxonomy should be moved to the ingredients taxonomy. The entries are now doubled (if correct). The ingredients taxonomy has much more synonyms than the allergens tax. At some stage I did a merge, i.e. moving all allergen entries into the ingredients taxonomy.

stephanegigandet:
> Maybe the allergens taxonomy should be moved to the ingredients taxonomy.

It's something we could try at some point, but it would make the parsing a lot more complex.

PR #2899, opened 2020-02-03
Producers - improved handling of image uploads

stephanegigandet:
- For mass upload of pictures, if barcode is not in the file name, try to scan it from the image
- Select uploaded "other" image as front image if no front image is selected


bredowmax:
Tested it today by uploading 400 images at once - it seemed like it was stopping once in a while but the feature to look for barcodes in the picture works great! Such an improvement, great work!

PR #2903, opened 2020-02-05
Unilever ingredients

aleene:
Down to 2 occurences.


stephanegigandet:
BTW rutabaga is different from navet, at least in France :) , we probably should create different entries

Issue #2851, opened 2020-01-13
Support .HEIC / HEIF Upload


bredowmax:
**Is your feature request related to a problem? Please describe.**
I wish OFF on desktop would support uploading .HEIC photos - this is the format my iPhone saves photos to save space

**Describe the solution you'd like**
I wish OFF would take care of those conversions so I don't have to.

**Describe alternatives you've considered**
It's possible to convert locally with a free app called HEIC Converter that is available on the MacOS appstore

**Additional context**
https://openfoodfacts.slack.com/archives/C02LDQDDD/p1578854003049700

stephanegigandet:
I looked a bit into that, imagemagick has support for HEIC in debian buster, but not in debian stretch (which we are running in production right now). It might be possible to build it from source, but I'm reluctant to do that as we had Apache + mod_perl crashing with the perlmagick modules and it was a lot of trouble to finally get a version of imagemagick that didn't crash in production.

So we definitely will be able to do it someday, but maybe not soon.

teolemon:
http://stuffjasondoes.com/2019/07/10/batch-convert-heic-to-jpg-in-linux/

teolemon:
it seems that there are alternative packages ?

stephanegigandet:
Uploading HEIC images is now supported. Like PNG images, we convert them to JPEG.

Issue #1793, opened 2019-05-22
Trimming of white background with ImageMagick Trim() during image crop is very slow 


stephanegigandet:
It can take up to 10 seconds to trim the white space of one single image... 

In process_image_crop.pl :

	# add auto trim to remove white borders (e.g. from some producers that send us images with white borders)

	$source->Trim();

Test image: https://fr.openfoodfacts.org/images/products/325/622/638/3503/9.jpg

This is with ImageMagick 7.0.8-34 Q16 x86_64 2019-03-18

stephanegigandet:
root@off1:/srv/off/html/images/misc# convert --version
Version: ImageMagick 7.0.8-34 Q16 x86_64 2019-03-18 https://imagemagick.org
Copyright: © 1999-2019 ImageMagick Studio LLC
License: https://imagemagick.org/script/license.php
Features: Cipher DPC HDRI OpenMP
Delegates (built-in): bzlib djvu fontconfig freetype gvc jbig jng jp2 jpeg lcms lqr lzma openexr png tiff wmf x xml zlib
root@off1:/srv/off/html/images/misc# time convert 3256226383671.jpg -trim 3256226383671_trim.jpg

real    0m38.503s
user    4m28.628s
sys     0m0.212s

hangy:
Would you consider re-compiling ImageMagick without HDRI? Apparently, that costs a lot of performance with ImageMagick 7: https://github.com/ImageMagick/ImageMagick6/issues/47#issuecomment-498311922

stephanegigandet:
We had huge problems with imagemagick making apache segfault, so I'd prefer not to touch it. ;)

See discussion in https://openfoodfacts.slack.com/messages/C1FPYCWM7/convo/GKB0C32TS-1559812818.045500/ 

I had a very bad night as I wasn't able to make imagemagick work correctly on the production server for several hours. We ended up installing imagemagick and perlmagick 7.0 from source, as the debian provided packages were making the server core..

hangy:
> We ended up installing imagemagick and perlmagick 7.0 from source, as the debian provided packages were making the server core..

I think that's part of the "problem". ImageMagick shold probably be re-compiled the same way as you did here, but without HDRI. If the dev server's specs&setup are identical to the production server, you could test there. 🙂

stephanegigandet:
This is causing big issues for the producers platform, when we import images: it takes a huge amount of cpu and time.

[...]

Issue #1663, opened 2019-02-13
Image colors shown as negative of the uploaded image when resized


stephanegigandet:
A very strange bug, a specific jpg image has its colors completely messed up after upload to OFF, when it is resized by OFF.

![karma-kombucha-citron-fr](https://user-images.githubusercontent.com/8158668/52710609-078d2f00-2f90-11e9-958f-0bb6105b9492.jpg)

Turns into:

![image](https://user-images.githubusercontent.com/8158668/52710678-3c00eb00-2f90-11e9-935a-fe2e507c206f.png)

teolemon:
Bad karma.

stephanegigandet:
It seems to be a bug when the JPG has the CMYK color profile instead of RGB. What's weird is that the old version of imagemagick / perlmagick on the old server does not have this issue.

stephanegigandet:
Probably due to this:

The behaviour of Set(colorspace) has changed:
https://www.imagemagick.org/discourse-server/viewtopic.php?p=143122#p143122
$magick->Set(colorspace => 'sRGB');

stephanegigandet:
Fixed.

PR #2914, opened 2020-02-07
HEIC / HEIX images support + fix colorspace issue

stephanegigandet:
This needs a specially compiled ImageMagick, see also bug #1793

Should solve #1663 and add support for HEIC #2851


teolemon:
3 birds. One stone 👍

Issue #2918, opened 2020-02-10
Add top scanned products by countries in popularity facet + products representing 90% and 95% of scans


stephanegigandet:
Add new values for each country in the /popularity facet, e.g. for Belgium:

"top-10-be-scans-2020",
"top-50-be-scans-2020",
"top-100-be-scans-2020",
"top-500-be-scans-2020",
"top-1000-be-scans-2020",
"top-5000-be-scans-2020",
"top-10000-be-scans-2020",
"top-50000-be-scans-2020",
"top-100000-be-scans-2020",

Also add global Top-90-percent-scans-2020 and Top-95-percent-scans-2020 facets on the top products that represent 90% or 95% of the scans worldwide.

teolemon:
can we close @stephanegigandet ?

stephanegigandet:
Deployed. One caveat is that we don't have enough products in the database to cover 90% of the scans, so those values are currently useless. (they are on all products that have been scanned)


Top-90-percent-scans-2019 | 344838
-- | --
Top-95-percent-scans-2019 | 344838
Top-95-percent-scans-2020 | 70622
Top-90-percent-scans-2020 | 70622

PR #2919, opened 2020-02-10
add scans by country + top 95 percent #2918, fixes #2397

### LINKED ISSUE #2397: Disable ScanBot on 20000000xxxxx product range
**Describe the bug**
Using IP adresses from in-app barcode scans, Scanbot adds countries to product pages. Some product pages were created with placeholder fake barcodes generated from the website (20000000xxxxx). 
Generally speaking, these products are either "freshly made" (i.e. their quantity or ingredients can chage) or without barcode. They can be considered unique and it's highly likely that their ingredients or nutrition values change very fast over time.
Recently, Scanbot added "Spain" to Burger King's Whopper burger product page. I don't think it's a good idea because we have no data to support the idea that this burger is exactly the same in France and in Spain. 

**To Reproduce**
1. Generate a new product page using the "Product without barcode" button.
2. Fill it with real information
3. ???
4. Scanbot will flag it someday and might add some countries.

**Expected behavior**
Stop scanbot patrol on the 20000000xxxxx EAN product range

**Screenshots**
See changelog at https://fr.openfoodfacts.org/cgi/product.pl?type=edit&code=2000000028596

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

Thank you.

stephanegigandet:
see #2918 for details

### DIFF
--- a/scripts/scanbot.pl
+++ b/scripts/scanbot.pl
@@ -78,9 +78,11 @@ while (<STDIN>)
 		my $code = $1;
 
 		# Skip bogus codes
+		
 		($code eq "1") and next;
+		($code eq "15600703") and next;
 
-		defined $codes{$code} or $codes{$code} = {n => 0, ips => {}};
+		(defined $codes{$code}) or $codes{$code} = {n => 0, ips => {}};
 
 		$codes{$code}{n}++;
 		$codes{$code}{ips}{$ip}++;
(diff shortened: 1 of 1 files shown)


teolemon:
I would need a facet where (80% - 90% )of scans are from a given country to be able to move easily products that are in the wrong country (but do not have ingredients yet) @stephanegigandet

PR #2926, opened 2020-02-18
use barcode scanned in previous photo / scan party mode - WIP

stephanegigandet:
This is to support "scan party mode" for image uploads:

The first photo contains the barcode, we scan it and recognize it.
The next photo is the front of the product: we don't have a barcode for it, so we use the previous barcode.

But it does not work. I spent hours trying to debug this, but for some reason I can't explain, when I upload my second photo, the "previous_code" is not set.

I think the issue is in there:

``````
  \$('#fileupload').fileupload(
    {
    sequentialUploads: true,
    // Uncomment the following to send cross-domain cookies:
    xhrFields: {withCredentials: true},
	formData : function () {
		var previous_code1 = previous_code;
		return ( [
					{ name : "random", value : Math.random()},
					{ name : "previous_code", value : previous_code + "-"},
					{ name : "previous_code1", value : previous_code1 + "-" + Math.random()},
					{ name : "previous_code_val", value : \$('#previous_code').val()},
					{ name : "previous_imgid", value : previous_imgid}
				]);
		}
	}
  );
``````

If I put 2 alert("previous_code: " + previous_code) before the return, then the first one does not show the previous_code, but the second one does.

So it works if I put an alert (even an empty one), but if I remove it, it fails.

Maybe some race condition or some caching somewhere. I don't understand it.

### DIFF
--- a/cgi/import_photos_upload.pl
+++ b/cgi/import_photos_upload.pl
@@ -78,14 +78,92 @@ else {
 	. "<li>" . lang("import_photos_format_nutrition") . "</li>"
 	. "</ul>";
 
-	$html .=<<HTML
+	$html .= <<HTML
       <form
         id="fileupload"
         action="/cgi/product_image_upload.pl"
         method="POST"
         enctype="multipart/form-data"
       >
+HTML
+;
 
+	# Enable adding field values for photos uploaded
+
+	my @add_fields = qw(brands categories packaging labels origins manufacturing_places emb_codes purchase_places stores countries);
+	my %add_fields_labels = ();
+	foreach my $field (@add_fields) {
+		$add_fields_labels{$field} = ucfirst(lang($field . "_p"));
+	}
+	$add_fields_labels{add_tag} = lang("add_tag_field");
+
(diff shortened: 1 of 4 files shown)


stephanegigandet:
It's currently on the dev server:

https://world.pro.openfoodfacts.dev/cgi/import_photos_upload.pl

stephanegigandet:
image

stephanegigandet:
I think the problem comes from this:

  \$('#fileupload').bind('fileuploadsubmit', function (e, data) {
    data.formData = function (form) {
		var d = new Date();
		var n = d.getTime();
		console.log(n);
		console.log(previous_code);
		return ( [
					{ name : "random", value : Math.random()},
					{ name : "time", value : n},
					{ name : "previous_code", value : previous_code},
					{ name : "previous_code_val", value : \$('#previous_code').val()},
					{ name : "previous_imgid", value : previous_imgid}
				]);
		};
});

This event is triggered when the files are added to the list of files to upload.

We should try to somehow wait for the first upload to complete before submitting the next one.

PR #2967, opened 2020-03-02
German Jams

moon-rabbitOFF:


stephanegigandet:
@moon-rabbitOFF  there is an error here:

ERROR - de:aprikosen-konfitüren already is a synonym of aprikosen-konfitüren (en:apricot-jams) - cannot make a synonym of aprikosenmarmelade for en:apricot-marmalades

I will separate apricot marmelade from apricot jam so that the taxonomy can build, we can revisit later if they should be merged.

Issue #3015, opened 2020-03-11
Apply vegan / vegetarian / palm oil labels overrides to ingredients analysis results


stephanegigandet:
We currently use ingredients analysis to populate the ingredients_analysis_tags field which contains a determination of whether the product is vegan, vegetarian, contains palm oil etc. based exclusively on the list of ingredients.

On the web site, we apply overrides so that labels like "Vegan", "Non-Vegan" etc. can take precedence over the results of the analysis (which can be "unknown" if we haven't been able to analyze all ingredients, or "maybe" if some ingredients may or may not be vegan for instance).

This will also allow to show the vegan status if we have a vegan label but no ingredients.

This has the drawback that it will no longer be possible to compare the results of ingredients analysis with the labels. So we will have to introduce another way to generate quality warnings if there is a mismatch between what labels and the ingredient analysis say.

stephanegigandet:
Fixed.

PR #3016, opened 2020-03-11
labels overrides for ingredients analysis, bug #3015

stephanegigandet:
Bug #3015 . Needed for apps.


stephanegigandet:
Keeping the branch open as I'll need to remove some code from Display.pm later.

PR #3021, opened 2020-03-12
Update ingredients_analysis.txt

teolemon:
I am not sure how it is formatted into something the app can still understand

### DIFF
--- a/taxonomies/ingredients_analysis.txt
+++ b/taxonomies/ingredients_analysis.txt
@@ -1,5 +1,3 @@
-
-
 en:Palm oil
 ca:Oli de palma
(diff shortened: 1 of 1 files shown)


stephanegigandet:
Yes, that doesn't work indeed, it would change things on the web site as well. :(

stephanegigandet:
Here we do want to keep our current message: https://fr.openfoodfacts.org/ingredients-analysis

stephanegigandet:
Leave the current message I guess. We'll see if app users understand it correctly.

teolemon:
I'm closing this draft.

Issue #1618, opened 2019-01-13
Support chinese serving size


aleene:
**Checklist:**

* [ ] Bug Report - The issue you are reporting might be there already. Please search on the [issues tracker](https://github.com/openfoodfacts/openfoodfacts-server/issues) before creating or reporting a issue.

* [ ] Before reporting, please ensure (if you can) that you have tested it with the latest (dev) version of the software.

**Summary:**

Summarize your issue in one sentence (what goes wrong, what did you expect to happen).

**Steps to reproduce:**

How can we reproduce the issue?

**Add system logs:**

Add logcat files here (if you are able to add).

**Expected behavior:**

What did you expect the app to do?

**Observed behavior:**

What did you see instead?  Describe your issue in detail here.

hangy:
I added some Chinese units in 3f4be298dd8c44860a4de708b82f0baab8e31698. However, the source that I found, [claims](http://www.new-chinese.org/lernwortschatz-chinesisch-masseinheiten.html) that "克" is "g", so I'm not quite sure what's actually correct. Maybe a native speaker could weigh in. 🙂

aleene:
You are correct. They seem to add 公, to indicate metric gram. Anyway they are both found on packages

aleene:
Install google translate on your phone and you can also add chinese products to off.

aleene:
By the way support for other scripts might be required as well (see wikidata item).

PR #3028, opened 2020-03-13
Add 公克 for metric grams in chinese, see #1618

CharlesNepote:
(please test before)
**Related issues and discussion:** #1618

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -255,6 +255,14 @@ sub get_nutrient_label {
 	}
 }
 
+=head2 unit_to_g($$)
+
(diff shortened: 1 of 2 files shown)


stephanegigandet:
New units also need to be added to unit_to_g and g_to_unit functions in Food.pm.

"(please test before)" --> Can you add unit tests? :) 

we have tests for other units in t/food.t

CharlesNepote:
You're right! I didn't know there is such unit test (and that's good!).

I'll also document the necessity to write tests, and tell the best practice is to write tests before coding.

CharlesNepote:
Unit tests added and passed :-)

Issue #3031, opened 2020-03-14
Streamline image cropping / 'select interesting zone' for ingredients & nutrient table by using higher-quality images


bredowmax:
**Describe the bug**
When cutting out the image for ingredients & nutrient table, figuring out where the area of interest is is tough because the image is too much compressed. I wish these preview photos were higher quality

**Discussion**
https://openfoodfacts.slack.com/archives/C06A7LENM/p1584120407088800

**Number of products impacted**
Any product with an image of lists of ingredients

**Time loss per product**
It's an extra few clicks per product. It doesn't stop me editing products, but it definitely makes it less efficient than it could be

stephanegigandet:
I'll try to make the crop area bigger, with a variable size depending on screen size, instead of the fixed 400x400 px.

I'll also try to use the new version of cropper https://github.com/fengyuanchen/jquery-cropper and see if we can load the full image and allow to zoom.

bredowmax:
My assumption is that the issue isn't the size of the crop area but the high compression of the image. The image looks quite blurry and could be much sharper even in the same size

stephanegigandet:
The image is 400x400 px and it's shown at 400x400 px, so it shouldn't be blurry, unless your browser displays it at another size. Or maybe on some retina display it's display 1 pixel over a 2x2 pixel block.

bredowmax:
I'm on a 2018 MacBook Pro 13" on Chrome. Typically, I'd expect the smaller version to appear sharper than the zoomed-in version, but the opposite is the case.

<img width="839" alt="Screenshot 2020-03-14 at 11 06 36" src="https://user-images.githubusercontent.com/30573532/76684630-ee680f00-65e3-11ea-8fec-1095e91f8ca1.png">

https://world.openfoodfacts.org/product/0073872746147/artisola-kale-garlic-cheese-ravioli-valley-fine-foods-company

bredowmax:
This exemplifies what I mean: Above image is the compressed image as shown in OFF when editing. On the bottom is the full version but forced to be displayed smaller because I made that Chrome window smaller

<img width="358" alt="Screenshot 2020-03-14 at 11 09 22" src="https://user-images.githubusercontent.com/30573532/76684678-53bc0000-65e4-11ea-8269-89f0ec026dfc.png">

bredowmax:
[...]

PR #3033, opened 2020-03-14
updated cropper to make it easier to use, bug #3031

stephanegigandet:
See bug #3031 and discussion in #ingredients and #productopener for details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,30 +37,30 @@
     "win32"
   ],
   "dependencies": {
-    "@yaireo/tagify": "^2.31.6",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -37,29 +37,30 @@
     "win32"
   ],
   "dependencies": {
-    "@yaireo/tagify": "^3.6.2",
```
hangy: Why all the package downgrades?

stephanegigandet: My bad, I had a stale copy of package.json. Fixed.

PR #3034, opened 2020-03-14
Ingredients analysis improvements + test web interface

stephanegigandet:
- support for 'ingredient (46%) (A, B, C)'
- better handling of ingredients with more than 1 label, processing, percent etc.
- and a new web interface to test ingredient analysis: https://fr.openfoodfacts.dev/cgi/test_ingredients_analysis.pl

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -696,7 +696,7 @@ sub parse_ingredients_text($) {
 		my $vegetarian = undef;
 		my $processing = '';
 
-		$debug_ingredients and print STDERR "s: $s\n";
+		$debug_ingredients and $log->debug("analyze_ingredients_function", { string => $s }) if $log->is_debug();
 
 		# find the first separator or ( or [ or :
 		if ($s =~ $separators) {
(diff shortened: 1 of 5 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -694,7 +696,7 @@ sub parse_ingredients_text($) {
 		my $vegetarian = undef;
 		my $processing = '';
 
-		#print STDERR "s: $s\n";
+		$debug_ingredients and print STDERR "s: $s\n";
```
VaiTon: Why don't we use log?

stephanegigandet: I need a way to easily turn on and off debugging messages for a given function, or at least module, for instance for tests in t.

In my .t file I can put something like:
use Log::Any::Adapter 'TAP', filter => 'debug';

But then it turns on debug error messages for all modules and functions, while I just want to see the function I'm working on.

There are probably better ways to do it than what I did. Would you know any?

stephanegigandet:
@VaiTon @hangy : I changed the print STDERR to statements like this:

$debug_ingredients and $log->debug("analyze_ingredients_function", { string => $s }) if $log->is_debug();

I kept the $debug_ingredients variable to turn on and off debugging for this function (it is very verbose and is very useful for ingredients parsing debugging, but otherwise it clutters the logs a lot).

There is probably a better way to filter logs function by function or module by module, we can improve on that after.

lib/ProductOpener/Ingredients.pm
stephanegigandet: Changed to statements like 
$debug_ingredients and $log->debug("analyze_ingredients_function", { string => $s }) if $log->is_debug();

Issue #3044, opened 2020-03-16
Document the ProductOpener::URL.pm module in perlpod format


stephanegigandet:
The /lib/ProductOpener/URL.pm is a small and rather independent module. It currently has no documentation in perlpod / POD format. (see https://perldoc.perl.org/perlpod.html )

It would be good to add perlpod documentation for the module and its functions.

See the /lib/ProductOpener/Nutriscore.pm module for an example of how to document a module with perlpod.

areeshatariq:
Working on this issue

stephanegigandet:
Fixed by @areeshatariq , thank you!

https://world.openfoodfacts.org/files/doc/perl/ProductOpener/URL.html

PR #3046, opened 2020-03-16
Document the ProductOpener::URL.pm module

areeshatariq:
**Description:**

Documentation of the ProductOpener::URL.pm module in perlpod format

**Related issues and discussion:** #3044

### DIFF
--- a/lib/ProductOpener/URL.pm
+++ b/lib/ProductOpener/URL.pm
@@ -18,6 +18,30 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+
+=head1 NAME
+
+ProductOpener::URL - generates the URL of the product
+
+=head1 SYNOPSIS
+
+C<ProductOpener::URL> is used to generate a URL of the product according to the subdomain .
+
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/URL.pm
```diff
@@ -42,6 +66,22 @@ use experimental 'smartmatch';
...
 
+=head1 FUNCTIONS
+
+=head2 format_subdomain( SUBDOMAIN )
+
+C<format_subdomain()> returns URL on the basis of subdomain and scheme (http/https)
+
+=head3 Arguments
+
+A scalar variable needs to be passed as an argument.
```
stephanegigandet: In addition to its type, we need to indicate was the argument is.

```suggestion
A scalar variable to indicate the subdomain (e.g. "us") needs to be passed as an argument. 
```

stephanegigandet: Same for the other functions.

areeshatariq:
@stephanegigandet I've made the suggested changes. Does it look fine now?

lib/ProductOpener/URL.pm
areeshatariq: Done

PR #3049, opened 2020-03-16
reorganize css styles, set max-height for ingredients full image, bug…

stephanegigandet:
… #3031

I moved a lot of CSS styles from the .pl to the .css. But I don't know how I can make the css styles be reloaded by browsers that have cached them. I remember you made something for that to add a version number in a ? parameter @hangy ?


hangy:
> I moved a lot of CSS styles from the .pl to the .css. But I don't know how I can make the css styles be reloaded by browsers that have cached them. I remember you made something for that to add a version number in a ? parameter @hangy ?

You added that in 5e60b2a68730b9d0fe9635165120c45334c8c056. 🙂 From my understanding, it should be automatically applied after a restart of the server.

html/css/product-multilingual.css
```diff
@@ -39,3 +39,125 @@ div.tagsinput input:focus {
...
+	margin-top:0.5rem;
+}
+
+#label_new_code, #new_code { display: inline; margin-top: 0px; width:200px; }
+
+th {
+	font-weight:bold;
+}
+
+// add borders to Foundation 5 tabs
```
hangy: I don't think this is a valid CSS comment? https://www.w3.org/TR/CSS21/syndata.html#comments

stephanegigandet: I never know how to make CSS comments. :) I'll fix it.

stephanegigandet:
> You added that in [5e60b2a](https://github.com/openfoodfacts/openfoodfacts-server/commit/5e60b2a68730b9d0fe9635165120c45334c8c056). 🙂 From my understanding, it should be automatically applied after a restart of the server.

I did? Wow, I completely forgot. OK that should work.

One more question: I noticed there is some processing through npm or gulpfile to make linted versions of js in the /dist/ directory. Should I also do that for the product-multilingual.css file? (e.g. does it get minified or something?)

hangy:
> One more question: I noticed there is some processing through npm or gulpfile to make linted versions of js in the /dist/ directory. Should I also do that for the product-multilingual.css file? (e.g. does it get minified or something?)

Linting is just syntax checking. That should already be done during PR builds.

We do some CSS minification with Gulp, but apparently not for `product-multilingual.css`. I guess, it shouldn't hurt to add it to https://github.com/openfoodfacts/openfoodfacts-server/blob/c0bdf053c416fd2a178ae76329f4aaf94e0ba3d9/gulpfile.js#L112-L124

html/css/product-multilingual.css
VaiTon: Also is the whole block commented out or the media query only? @stephanegigandet

stephanegigandet: just the media query I think.

stephanegigandet:
Thanks @hangy, I added the timestamp for product-multilingual.css & .js, and the css file to gulp.

Issue #2203, opened 2019-08-20
Document ProductOpener Perl modules.


zigouras:
**Is your feature request related to a problem? Please describe.**
There is little to no documentation in the code.  Perl POD (plain old documentation) can be used to generate a nice document page.

**Describe the solution you'd like**
Document all `ProductOpener/*.pm` modules with Perl POD including functions, classes, variables etc.

**Describe alternatives you've considered**
Moving blindly through the code.

**Additional context**
Since these Perl modules are the core of the business logic of the project they should be well documented.

### Part of 
- #5559

stephanegigandet:
Agreed, we are lacking a lot in terms of documentation of the code. A lot of it could also certainly be simplified, reorganized etc. Maybe we can start with some of the simpler / most independent modules (and then also see if we could break / reorganize the more complex / big modules).

I could try with the Imports.pm module first. I have never used POD, is there a good example of POD best practices for documenting a module?

zigouras:
Here is the official doc page for pod
https://perldoc.perl.org/perlpod.html

It lets you write inline markup in your code for headers, function documentation etc.  You then run
`perldoc myfile.pm`
to see a nicely formatted page.

Thanks for following up.

VaiTon:
> **Describe alternatives you've considered**
> Moving blindly through the code.

Love you <3 lol

CharlesNepote:
I appreciate recommendations from Damian Conway, in Perl Best Practices (2005). We could start with [comment templates](
https://books.google.fr/books?id=gJf9tI2mytIC&lpg=PT164&ots=Hn3hgNwthq&dq=perl%20%22comment%20templates%22&hl=fr&pg=PT163#v=onepage&q&f=false) for functions and methods.

Structured/templated comments are easier to write (you can't forget anything and it's easier to write) and also to read. Here is what he writes about it:
_Create comment templates that are suitable for your team. For example, to inter-
nally document a subroutine or method, you might use something like:_
```
############################################
# Usage      : ????
# Purpose    : ????
# Returns    : ????
# Parameters : ????
# Throws     : no exceptions
# Comments   : none
# See Also   : n/a
```
_which might be filled in like so:_
```
############################################
# Usage      : Config::Auto->get_defaults( )
# Purpose    : Defaults for 'new'
# Returns    : A hash of defaults
# Parameters : none
# Throws     : no exceptions
# Comments   : No corresponding attribute,
#            : gathers data from each
[...]

PR #3054, opened 2020-03-17
script to call pod2html to generate perlpod html doc, issue #2203

stephanegigandet:
Now that we have quite a few modules with at least some POD documentation, I'm looking into ways to present it. This is the result of pod2html:

https://world.openfoodfacts.org/files/doc/perl/index.html

It doesn't look very nice yet, but it's probably possible to make it nicer with some CSS styles etc.


stephanegigandet:
@hangy: I was hoping the default html rendered would look a bit better. I'm not sure about the auto-generation, in practice I need to generate the html on the prod server. Unless there's a way to host the resulting html files on github somehow?

PR #3056, opened 2020-03-17
[WIP] Added: new tests for Text.pm

### LINKED ISSUE #3039: \x{a0} space not handled in Text::_format_percentage / normalize_percentages()
When updating products I get a bunch of warnings like this one:

Argument "100\x{a0}" isn't numeric in division (/) at ProductOpener/Text.pm line 175.

muskangarg21:
Fixes #3039 
New test cases added.

### DIFF
--- a/lib/ProductOpener/Text.pm
+++ b/lib/ProductOpener/Text.pm
@@ -146,11 +146,11 @@ sub _get_locale_percent_regex {
 	# [+-]?(?:\d{3}\.)*\d+(?:,\d+)*\h*% where . is the group sign from the locale, and , is the decimal point - or other way around for tr etc.
 	my $regex;
 	if (index($perf->pattern, $perf->percent_sign) == 0) {
-		$regex = qr/(%\h*[$p$m]?(?:\d{3}$g)*\d+(?:$d\d+)*)/;
-	}
-	else {
-		$regex = qr/([$p$m]?(?:\d{3}$g)*\d+(?:$d\d+)*\h*%)/;
-	}
+		$regex = qr/(%\h*[$p$m]?(?:\d{1,3}$g)*\d+(?:($d|\.)\d+)*)/;
+    }
+    else {
+        $regex = qr/([$p$m]?(?:\d{1,3}$g)*\d+(?:($d|\.)\d+)*\h*%)/;
+    }
 
 	$ProductOpener::Text::regexes{$locale} = $regex;
 	return $regex;
@@ -168,9 +168,19 @@ sub _format_percentage($$$) {
 	# 1 make the string float parseable by Perl
 	# 1.1 remove % and group sign
 	$value =~ tr/%//d;
-	$value =~ s/$g//g;
-	# 1.2 replace decimal sign with a decimal dot
-	$value =~ s/$d/\./g;
+	# if the last group sign is not followed by 3 digits,
+    # assume it is in fact a decimal sign.
+    # e.g. in French 2,50 is the right form, but 2.50 is very common.
+    if ($value !~ /$g(\d{1,2}|\d{4,10})$/) {
+        $value =~ s/$g//g;
+    }
+    else {
+        $value =~ s/$g/\./;
+    }
+	# 1.2 remove nbsp
+	$value =~ tr/[\x{a0}]//d;
+	# 1.3 replace decimal sign with a decimal dot
+	$value =~ s/($d|,)/\./g;
 	# 2 make percent
 	$value = $value / 100.0;
 	# 3 format with given locale and return
--- a/t/text.t
+++ b/t/text.t
@@ -24,21 +24,40 @@ use ProductOpener::Text qw/:all/;
 # ur      #,##,##0%
 is( normalize_percentages('test 1234% hi there', 'ur'), 'test 1,234% hi there' );
 is( normalize_percentages('test 123,456.78% hi there', 'ur'), 'test 1,23,456.78% hi there' );
-
+is( normalize_percentages('test 0,12,345.67% hi there', 'ur'), 'test 12,345.67% hi there' );
+is( normalize_percentages('test 1,002.34% hi there', 'ur'), 'test 1,002.34% hi there' );
 # de	#,##0\N{U+00A0}%
 is( normalize_percentages('test 1234% hi there', 'de'), "test 1.234\N{U+00A0}% hi there" );
 is( normalize_percentages('test 123.456,78% hi there', 'de'), "test 123.456,78\N{U+00A0}% hi there" );
+is( normalize_percentages('test 1.023,45% hi there', 'de'), "test 1.023,45\N{U+00A0}% hi there" );
+is( normalize_percentages('test 1.23.045,67% hi there', 'de'), "test 123.045,67\N{U+00A0}% hi there" );
+is( normalize_percentages("test 1.23.045,67\N{U+00A0}% hi there", 'de'), "test 123.045,67\N{U+00A0}% hi there" );
+is( normalize_percentages("test 1.23.045,67 \N{U+00A0} % hi there", 'de'), "test 123.045,67\N{U+00A0}% hi there" );
 
 # eu	% #,##0
 is( normalize_percentages('test % 1234 hi there', 'eu'), "test %\N{U+00A0}1.234 hi there" );
 is( normalize_percentages('test %1234 hi there', 'eu'), "test %\N{U+00A0}1.234 hi there" );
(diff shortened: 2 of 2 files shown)


muskangarg21:
@stephanegigandet 
Could you please look into the Travis checks that are failing, they seem to be related to `Tags.pm` .

stephanegigandet:
that's because Tags.pm uses those functions from Text.pm to canonicalize strings:

in t/tags.t 
is(canonicalize_taxonomy_tag("en", "test", "kefir 2.5%"), "en:kefir-2-5");
is(canonicalize_taxonomy_tag("en", "test", "kefir 2,5%"), "en:kefir-2-5");
is(canonicalize_taxonomy_tag("fr", "test", "kefir 2,5%"), "en:kefir-2-5");
is(canonicalize_taxonomy_tag("fr", "test", "kéfir 2.5%"), "en:kefir-2-5");

The 2nd line is the one that now fails.   2,5% is now converted to 25%.

lib/ProductOpener/Text.pm
```diff
@@ -146,10 +146,10 @@ sub _get_locale_percent_regex {
 	# [+-]?(?:\d{3}\.)*\d+(?:,\d+)*\h*% where . is the group sign from the locale, and , is the decimal point - or other way around for tr etc.
 	my $regex;
 	if (index($perf->pattern, $perf->percent_sign) == 0) {
-		$regex = qr/(%\h*[$p$m]?(?:\d{3}$g)*\d+(?:$d\d+)*)/;
+		$regex = qr/(%\h*[$p$m]?(?:\d{1,3}$g)*\d+(?:$d\d+)*)/;
```
stephanegigandet: Why this change? For which strings is it needed?

muskangarg21: `is( normalize_percentages('test 0,12,345.67% hey', 'ur'), 'test 12,345.67% hey' );`

`is( normalize_percentages('test 1.23.045,67% hey', 'de'), "test 123.045,67\N{U+00A0}% hey" );`

For these test cases, where the number of digits before the `group_sign` can be 2 or 1 .

muskangarg21:
> that's because Tags.pm uses those functions from Text.pm to canonicalize strings:
> 
> in t/tags.t
> is(canonicalize_taxonomy_tag("en", "test", "kefir 2.5%"), "en:kefir-2-5");
> is(canonicalize_taxonomy_tag("en", "test", "kefir 2,5%"), "en:kefir-2-5");
> is(canonicalize_taxonomy_tag("fr", "test", "kefir 2,5%"), "en:kefir-2-5");
> is(canonicalize_taxonomy_tag("fr", "test", "kéfir 2.5%"), "en:kefir-2-5");
> 
> The 2nd line is the one that now fails. 2,5% is now converted to 25%.

@stephanegigandet I don't understand how is this an expected behaviour , when normalize_percentages called on `en` as `locale` it should be expecting the decimal_sign as `.` as opposed to `,` mentioned in testcase 2 above. 
I also ran this snippet of code to see the expected value.
<pre>
use CLDR::Number;
my $cldr = CLDR::Number->new(locale => 'en');
my $perf = $cldr->percent_formatter( maximum_fraction_digits => 2);
say $perf->format(0,1212);
say $perf->format(0.1212);
</pre>
Output: 
0%
12.12%
Now I am even more confused to what should the expected behaviour should be.
Please Help.

stephanegigandet:
@muskangarg21 : the input we deal with is far from perfect. We have ingredients lists in French that use , or . as the decimal sign, and same for other languages. (it can be from the package itself, or from what users enter).

So we try to be very liberal in what we accept, and to do what seems to be the right thing for each different case.

e.g. if we have 2,343 as input for English, we will convert it to 2343. But if we have 2,50 in English, it is probably 2.50, and there is very little chance that it is 250.

lib/ProductOpener/Text.pm
stephanegigandet: I thought digits were only grouped by groups of 3, but from reading https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping it seems that in India it is first a group of 3 then groups of 2.... 
In practice we won't get many percent values that are more to 100%, but for comparisons it could happen.
We don't want 2,50 to be converted to 250 (even in English).
Maybe we can change the regexp so that the last group of digits is \d{3}.

muskangarg21:
> @muskangarg21 : the input we deal with is far from perfect. We have ingredients lists in French that use , or . as the decimal sign, and same for other languages. (it can be from the package itself, or from what users enter).
> 
> So we try to be very liberal in what we accept, and to do what seems to be the right thing for each different case.
> 
> e.g. if we have 2,343 as input for English, we will convert it to 2343. But if we have 2,50 in English, it is probably 2.50, and there is very little chance that it is 250.

@stephanegigandet then what do you suggest to how we can accomodate the test cases I have added. 
- May I write another regex to accommodate them or you have some better approach in mind ?

- Also the use of spaces which was the main issue here , does the changes in this PR work for that ?

stephanegigandet:
Hi @muskangarg21 

> Also the use of spaces which was the main issue here , does the changes in this PR work for that ?

Yes, it does. We just need to be careful that fixing the new test cases won't break anything else, that's why we try to put all those tests cases, including the one that now is breaking.

> May I write another regex to accommodate them or you have some better approach in mind ?

Sure, please do. I think @hangy wrote this function, maybe he has ideas.

muskangarg21:
I am pretty much stuck in a loop.
If we want to add all these new test cases, we will need to change the function `normalize_percentages` to accommodate cases where 12,3% and 12.3 % becomes 12.3% but simultaneously cases like 1.004% and 1,004%  are ambiguous, should they be interpreted as 1.004% or 1004% and this is indeed ambiguous as in the link [`https://www.healthline.com/nutrition/11-most-nutrient-dense-foods-on-the-planet#TOC_TITLE_HDR_3`]   
the number 2 item `Kale` has a 1000% Vitamin K1 mentioned so we might run into a problem if not fixed.
@stephanegigandet @hangy your views ?
Thanks

stephanegigandet:
OK, here is my suggestion:

in t/text.t add those tests:

#fr
is( normalize_percentages('2,50%', 'fr'), "2,5\N{U+00A0}%" );
# 2.50 should be 2,50 in French, but the form with the . is very common too
is( normalize_percentages('2.50%', 'fr'), "2,5\N{U+00A0}%" );
is( normalize_percentages('2.5%', 'fr'), "2,5\N{U+00A0}%" );
is( normalize_percentages('2.500%', 'fr'), "2,5\N{U+00A0}%" );
is( normalize_percentages('2500%', 'fr'), "2\N{U+00A0}500\N{U+00A0}%" );

#en
is( normalize_percentages('2,50%', 'en'), "2.5%" );
is( normalize_percentages('2.50%', 'en'), "2.5%" );

in Text.pm:

 sub _get_locale_percent_regex {
        my $d = quotemeta($cldr->decimal_sign);

        # [+-]?(?:\d{3}\.)*\d+(?:,\d+)*\h*% where . is the group sign from the locale, and , is the decimal point - or other way around for tr etc.
+       # make the decimal sign also include . as it is common in languages
+       # where the correct sign is ,
        my $regex;
        if (index($perf->pattern, $perf->percent_sign) == 0) {
-               $regex = qr/(%\h*[$p$m]?(?:\d{1,3}$g)*\d+(?:$d\d+)*)/;
+               $regex = qr/(%\h*[$p$m]?(?:\d{1,3}$g)*\d+(?:($d|\.)\d+)*)/;
        }
        else {
-               $regex = qr/([$p$m]?(?:\d{1,3}$g)*\d+(?:$d\d+)*\h*%)/;
+               $regex = qr/([$p$m]?(?:\d{1,3}$g)*\d+(?:($d|\.)\d+)*\h*%)/;
        }

and:

sub _format_percentage($$$) {
        # 1 make the string float parseable by Perl
        # 1.1 remove % and group sign
        $value =~ tr/%//d;
-       $value =~ s/$g//g;
+       # if the last group sign is not followed by 3 digits,
+       # assume it is in fact a decimal sign.
+       # e.g. in French 2,50 is the right form, but 2.50 is very common.
+       if ($value !~ /$g(\d{1,2}|\d{4,10})$/) {
+               $value =~ s/$g//g;
+       }
+       else {
+               $value =~ s/$g/\./;
+       }
        # 1.2 remove nbsp
        $value =~ tr/[\x{a0}]//d;
        # 1.3 replace decimal sign with a decimal dot
-       $value =~ s/$d/\./g;
+       $value =~ s/($d|,)/\./g;
        # 2 make percent

Then all tests will pass, including the ones you added too.

PR #3057, opened 2020-03-17
NL poeder / DE pulver processing parsing + tests

stephanegigandet:


stephanegigandet:
> # Dutch: match before or after, do not require a space
> or (($product_lc =~ /^(nl)$/) and ($new_ingredient =~ /(^($regexp)|($regexp)$)/i)))
> 
> For dutch it should be an optional space, sometimes it is there, sometimes not.

@aleene : yes, that's what this does. The match does not require a space, and it will match if there is one.

PR #3092, opened 2020-03-23
Process gepelde

aleene:
**Description:**
Added process `gepelde`, tests and built taxonomy files.


taxonomies/ingredients_processing.result.txt
```diff
@@ -24,6 +24,8 @@ en:dehydrated
 fr:déshydraté, déshydratée, déshydratés, déshydratées, lyophilisé, lyophilisée, lyophilisés, lyophilisées
 it:disidratato, disidratata, disidratati, disidratate
 
+en:description:With the outermost layer or skin removed
```
stephanegigandet: This should be a comment, that format is not recognized in the taxonomy.

PR #3090, opened 2020-03-23
Association CIQUAL AGB : lines 997 to 1336

bernardlemoullec:
Previous one was https://github.com/openfoodfacts/openfoodfacts-server/pull/3075

PR #3096, opened 2020-03-24
Association CIQUAL AGB : lines 997 to 1400

bernardlemoullec:
Followup to https://github.com/openfoodfacts/openfoodfacts-server/pull/3090

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -17372,12 +17372,22 @@ fr:Biscuits feuilletés, biscuit feuilleté
 nl:Bladerdeegkoekjes, Bladerdeegkoekje
 
 <en:Flaky biscuits
-en:Palmiers
+en:Palmiers, Palmier, Puff pastry biscuit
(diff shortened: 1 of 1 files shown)


taxonomies/categories.txt
```diff
@@ -17708,6 +17865,14 @@ nl:Platte graanbroden, Platte graanbrood
 sv:Halvtjockkakor, Hällakaka
 wikidata:en:Q10516678
 
+<en:Wheat breads
+<en:Flatbreads
+en:Fajitas
```
stephanegigandet: Fajitas is a meal wrapped in a tortilla (which is the flat bread). Fajitas should not be under flatbreads.

bernardlemoullec: OK I will remove <en:Flatbreads in the next PR. Thanks for the heads up!

bernardlemoullec:
Please refer to https://github.com/openfoodfacts/openfoodfacts-server/pull/3110

Issue #3047, opened 2020-03-16
Dropdowns in Edit mode without background color


bredowmax:
## Discussion:
https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=5760466896481

<img width="315" alt="Screenshot 2020-03-16 at 08 55 50" src="https://user-images.githubusercontent.com/30573532/76781604-5d309e00-6785-11ea-93b0-922ba0f6e1be.png">

stephanegigandet:
I don't think the related code has changed, but we may have installed newer versions of the javascript library through npm. cc @hangy

stephanegigandet:
There's an error in the console:

![image](https://user-images.githubusercontent.com/8158668/76788498-56bd1900-67bb-11ea-9e72-d97bb8a39d5c.png)

nazimboudeffa:
I get this error when I tape enter after filling the name of a country
![error](https://user-images.githubusercontent.com/1375686/76795141-c8e82a80-67c8-11ea-8076-293c25389f8a.png)

stephanegigandet:
OK, this is probably because the scss / sass stuff is not taken into account.

PR #3099, opened 2020-03-24
trying to fix tagify missing background, bug #3047

stephanegigandet:
This is to try to fix the missing background of the tagify dropdown (bug #3047 ). (which is  quite difficult to debug in a browser, because that dropdown disappears as soon as I click in the inspector to try to select it)

I tried, but I didn't manage to get npm to correctly generate the tagify.css file with the default scss values for the variables. My html/dist/tagify.css file has things like border:1px solid var(--tags-border-color) in it, I'm guessing that's not right.

Could you take a look @hangy or @VaiTon ?

### DIFF
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -836,7 +836,6 @@ HTML
 
 	$scripts .= <<HTML
 <script type="text/javascript" src="/js/dist/cropper.js"></script>
-<script type="text/javascript" src="/js/jquery.tagsinput.20160520/jquery.tagsinput.min.js"></script>
 <script type="text/javascript" src="/js/jquery.form.js"></script>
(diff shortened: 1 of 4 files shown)


teolemon:
is it something that also happens without the power-user-script ?

stephanegigandet:
> is it something that also happens without the power-user-script ?

Yes, it's not related.

hangy:
We actually include the precompiled tagify.css, but the wrong version. I think the fix in #3102 has got right.

stephanegigandet:
> We actually include the precompiled tagify.css, but the wrong version. I think the fix in #3102 has got right.

Perfect, I'll close this bug. I'll make another PR to remove the loading of taginput.js which I think is useless now that we use tagify.

PR #3106, opened 2020-03-25
Process nl:gehakte added

aleene:
**Description:**
Added process nl:gehakte, adapted tests and built taxonomies.


taxonomies/ingredients.txt
```diff
@@ -43201,6 +43201,10 @@ pl:krojane pomidory
 sk:krájané rajčiny
 # ingredient/en:chopped-tomatoes has 15 products in english @2018-12-24
 
+<en:peeled tomatoes
+<en:chopped tomatoes
+nl:gehakte gepelde tomaten
```
stephanegigandet: That's the kind of ingredient we don't want to put in the ingredients taxonomy. Instead of matching nl:gehakte gepelde tomaten we want to match en:tomatoes with the processing en:chopped and en:peeled.

aleene: Yes, later I will prune the ingredinets for the processed ones.

aleene: I might add a test for this case though. But I am going step by step.

aleene: Can I add a test for these cases?

PR #3108, opened 2020-03-25
Ingredients analysis: built taxonomies + small parsing improvement for organic ingredients

stephanegigandet:


stephanegigandet:
> I can see the test, but not the improvement. Besides, there's a merge conflict.

The parsing improvement results of the taxonomy change.

PR #3120, opened 2020-03-26
Some german processes

aleene:
**Description:**
Some german processes

I am giving up on the tests. I do not understand how this works, errors are to cryptic, takes to much time.


stephanegigandet:
@aleene : the tests show you the differences between the results of the parsing, and what you expected:

We got this:

not ok 6
#   Failed test at ../t/ingredients_processing.t line 201.
#          $got->[0]{processing} = Does not exist
#     $expected->[0]{processing} = 'en:powdered'
[
  {
    'id' => 'de:Sauerkrautpuder',
    'text' => 'Sauerkrautpuder'
  }
]

And in the t/ingredients_processing.t test file, you were expecting this:

        [ { lc => "de", ingredients_text => "Sauerkrautpuder" },
                [
                        {
                        'id' => 'en:sauerkraut',
                        'processing' => 'en:powdered',
                        'text' => 'Sauerkraut'
                        }
                ]
        ],

Then the question is: why didn't we get the expected result. I'll have a look.

stephanegigandet:
@aleene in ingredients_processing.txt you have this:

en:powdered, powder
de:pulver

"pulver" but not "puder" as you put in the test file : "Sauerkrautpuder"

I don't speak German so I don't know which one is correct, or if both are correct.

lib/ProductOpener/Ingredients.pm
```diff
@@ -1012,7 +1012,7 @@ sub parse_ingredients_text($) {
 									# English, French etc. match before or after the ingredient, require a space
 									(($product_lc =~ /^(en|es|it|fr)$/) and ($new_ingredient =~ /(^($regexp)\b|\b($regexp)$)/i))
 									#  German: match after, do not require a space
-									or	(($product_lc =~ /^(de)$/) and ($new_ingredient =~ /($regexp)$/i))
+									or	(($product_lc =~ /^(de)$/) and ($new_ingredient =~ /(^($regexp)|($regexp)$)/i))
```
stephanegigandet: Why have you changed this? Do you want to match "Pulversaurkraut" in addition to "Saurkrautpulver"?

aleene: Yes and both are correct

aleene: https://de.openfoodfacts.org/ingredients?filter=puder
https://de.openfoodfacts.org/ingredients?filter=pulver

aleene:
Are all these tests not to much? Is it not sufficient to test the principles? i.e. the regexp in ingredients.pm.

stephanegigandet:
Well no, the fact that the tests fail actually shows that there is something wrong: is it pulver or puder? or both?

And it is very easy to break something that used to work when we introduce something. For instance the next time someone enhance one of the regexp, he or she might not be aware of why it was that way, and break a test. It happens to me all the time, the tests are an insurance that we won't break in the future something that works now.

For instance here, you changed the German regexp so that it matches boths "pulversomething" or "somethingpulver". We should have tests that exercize both of those. If not, then someone else may change that regexp, and assume it should only match "somethingpulver". If there's no test case, then he or she won't know that something got broken.

The parsing code is run on all products, there are tons of different ways to write things in ingredients, so we really need an extensive set of test cases, otherwise we will make mistakes, and we might not notice them for very long.

stephanegigandet:
So yes, tests take time now, but not having tests would cost much more time later.

aleene:
Well I was/am getting annoyed, because I am totally lost. It seems that I do everything double for every process. I agree with testing, but I am missing what is being tested.
Per language there is only one regexp, so that should be tested. In this case `puder$` or `$puder`. And synonyms should work, i.e. `puder` and `pulver`. But there should be no need to t this also for gehackt, gehobelt, etc.

stephanegigandet:
I'll fix this one.

I don't see any German ingredient that begins with "puder" or "pulver", it's always at the end, so I will change the regexp to match only at the end. (see https://de.openfoodfacts.org/ingredients?filter=pu )

stephanegigandet:
The only exception seems to be "Puderzucker", which is in fact the only ingredient with "Puder" in it.

stephanegigandet:
Working on it.

aleene:
The same in dutch, we have poedersuiker. And otherwise at the end.

aleene:
But there are others: https://de.openfoodfacts.org/ingredients?filter=gehackt

aleene:
But dies that explain the failing test?

stephanegigandet:
I'm getting close, I had forgotten to rebuild the new ingredients taxonomy.

stephanegigandet:
I'm adding a distinction for "icing sugar" and "granulated sugar / powdered sugar".

Also I'm keeping "powdered sugar" in the taxonomy for now, especially as we want to keep a distinction between "icing sugar" and "granulated sugar".

aleene:
Great, and what was the issue(s) now?

PR #3126, opened 2020-03-27
More logs in case of failure

teolemon:
I found this on the old travis config file we disabled
https://github.com/openfoodfacts/openfoodfacts-server/blob/master/.travis.yml


stephanegigandet:
I have no idea how this work. Would that still be executed?

teolemon:
I'm still debugging it, the syntax is not the same as travis. Let's see if it works with my latest commit

PR #3133, opened 2020-03-27
Process de:püree

aleene:
****Description:**
- Added process de:püree and synonyms
- Removed püree entries from ingredients taxonomy
- Created two tests for `ingredientpüree` and `ingredient püree`.

**Issues
Are more tests required? 
- Synonym tests? 
- `püreeingredient` tests

`de:Apfel-Püree konzentriert` has been left in taxonomy

### DIFF
--- a/refresh_taxonomies.js
+++ b/refresh_taxonomies.js
@@ -26,6 +26,7 @@ async function main() {
     'countries',
     'ingredients',
     'ingredients_analysis',
+    'ingredients_processing',
     'labels',
     'languages',
     'minerals',
--- a/t/ingredients_processing.t
+++ b/t/ingredients_processing.t
@@ -162,14 +162,35 @@ my @tests = (
 		]
 	],
 	
-		[ { lc => "de", ingredients_text => "gehackter Dickmilch" }, 
-	[
-	  {
-	    'id' => 'en:soured-milk',
-	    'processing' => 'en:chopped',
-	    'text' => 'Dickmilch'
-	  }
-	]
+	[ { lc => "de", ingredients_text => "gehackter Dickmilch" }, 
+		[
+	  		{
+	    		'id' => 'en:soured-milk',
+	    		'processing' => 'en:chopped',
+	    		'text' => 'Dickmilch'
+	  		}
+		]
+	],
+
+# Test for de:püree without space
+	[ { lc => "de", ingredients_text => "Schalottepüree" }, 
+		[
+	  		{
+	    		'id' => 'en:shallot',
+	    		'processing' => 'en:puree',
+	    		'text' => 'Schalotte'
+	  		}
+		]
+	],
+# Test for de:püree with space
+	[ { lc => "de", ingredients_text => "Schalotte püree" }, 
+		[
+		  	{
+				'id' => 'en:shallot',
+				'processing' => 'en:puree',
+				'text' => 'Schalotte'
+				}
+			]
 		],
 
 	[ { lc => "de", ingredients_text => "hartkäse gehobelt, haselnüsse gehackt, haselnüsse gehackt und geröstet, gehackte und geröstete haselnusskerne, gehobelte und gehackte mandeln" },
--- a/t/ingredients_processing.t
+++ b/t/ingredients_processing.t
@@ -239,6 +239,33 @@ my @tests = (
 
 ]
 	],
+
+[ { lc => "de", ingredients_text => "Schalottepüree, zwiebel püree, spinat-püree, gurkenmark" },
+[
+  {
+    'id' => 'en:shallot',
+    'processing' => 'en:pureed',
+    'text' => 'Schalotte'
+  },
+  {
+    'id' => 'en:onion',
+    'processing' => 'en:pureed',
+    'text' => 'zwiebel'
+  },
+  {
+    'id' => 'en:spinach',
+    'processing' => 'en:pureed',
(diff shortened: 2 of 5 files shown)


stephanegigandet:
> * `püreeingredient` tests

It doesn't look like there are any ingredients that start with püree : https://de.openfoodfacts.org/ingredients?filter=p%C3%BCre

aleene:
There is one in the ingredients taxonomy

aleene:
But question is mainly about the tests that should be added. Which ones to add? What combinations? What should I look at for?

aleene:
Added a few false positives to the test and ingredients taxonomy.

stephanegigandet:
> But question is mainly about the tests that should be added. Which ones to add? What combinations? What should I look at for?

At least one basic test for the general case, plus cases for things to watch for. Adding tests for false positives like you did is a very good idea.

aleene:
> At least one basic test for the general case, plus cases for things to watch for. Adding tests for false positives like you did is a very good idea.

What do you mean with the general case, the process? What about synonyms? 

Bythe way I tried to get the false positives working. I rebuilt the ingredients taxonomy. But the test still fails. So I am missing something.

stephanegigandet:
For the next processing methods:

I think it would be easier not to change the ingredients taxonomy at this point (unless we need to add entries that would be false positives otherwise).

e.g. don't remove the existing "something processing" that we have in the taxonomy today. We can keep them for now, and later we can remove them in block.

And in the tests, that means you should use another entry, one that is not yet covered by the existing taxonomy.

The reason is that it's much easier to make the tests etc. when we change only one taxonomy at a time. Especially as building the ingredients taxonomy takes 30 minutes, whereas building the processing taxonomy is almost instant.

stephanegigandet:
It doesn't cost much to add a test that covers each synonym, you can group them like that:
````
[ { lc => "de", ingredients_text => "Schalottepüree, brombeer püree, spinat-püree, gurkenmark" },
[]
],
````
Notice the empty [] for the expected result. Just run the test like that. It will fail, and in the fail report, you will get the actual result returned:

````
not ok 15
#   Failed test at ../t/ingredients_processing.t line 264.
#          $got->[0] = HASH(0x5613a89974d0)
#     $expected->[0] = Does not exist
[
  {
    'id' => 'en:shallot',
    'processing' => 'en:pureed',
    'text' => 'Schalotte'
  },
  {
    'id' => "de:brombeer p\x{fc}ree",
    'text' => "brombeer p\x{fc}ree"
  },
  {
    'id' => 'en:spinach',
    'processing' => 'en:pureed',
    'text' => 'spinat'
  },
  {
    'id' => 'en:gherkin',
    'processing' => 'en:pureed',
    'text' => 'gurken'
  }
]
````
-> so in this case, we got what we expected, except for brombeer püree (which is because we have brombeeren in the taxonomy)

So we change the test to something else:
````
[ { lc => "de", ingredients_text => "Schalottepüree, zwiebel püree, spinat-püree, gurkenmark" },
[]
],
````
We re run the test:
````
not ok 15
#   Failed test at ../t/ingredients_processing.t line 264.
#          $got->[0] = HASH(0x5563f373e358)
#     $expected->[0] = Does not exist
[
  {
    'id' => 'en:shallot',
    'processing' => 'en:pureed',
    'text' => 'Schalotte'
  },
  {
    'id' => 'en:onion',
    'processing' => 'en:pureed',
    'text' => 'zwiebel'
  },
  {
    'id' => 'en:spinach',
    'processing' => 'en:pureed',
    'text' => 'spinat'
  },
  {
    'id' => 'en:gherkin',
    'processing' => 'en:pureed',
    'text' => 'gurken'
  }
]
1..15
````
And here everything is as we expect. So we can just copy and paste this array into the test, to replace the [] with the good result.

stephanegigandet:
And we should make a note to add plurals for berries in German in the ingredients taxonomy, but we should do it after, in another pull request, so that it's easier.

(unless there is a way in German to automatically guess the plural of a word? e.g. anything that ends with beer becomes beeren? cc @hangy)

stephanegigandet:
@aleene : I think I know why it was failing:  the ingredients_processing taxonomy was not automatically rebuilt by the github test when ingredients_processing.txt changed. So it worked only when the ingredients_processing.result.sto/txt files were commited as well. Fixing it.

stephanegigandet:
Fixed a formatting error in the test result. That's why it's best to just run the test to get the results before copying it into the test, that way it's much faster to add a test in 2 minutes :

1. add the input, with a dummy output
2. run the test locally on the dev server, the test fails and shows the output
3. if the output is correct, copy the output inside the test file

aleene:
> @aleene : I think I know why it was failing: the ingredients_processing taxonomy was not automatically rebuilt by the github test when ingredients_processing.txt changed. So it worked only when the ingredients_processing.result.sto/txt files were commited as well. Fixing it.

Yes I know. I did the testing first on the server. The process tests passed, but the false positives  test did not pass (and I even did rebuild of the ingredients taxonomy). But then I stopped there (and did not upload ingredients_processing.result.sto).

> I think it would be easier not to change the ingredients taxonomy at this point (unless we need to add entries that would be false positives otherwise).

OK. I can add the false positives in a later stage.

> And in the tests, that means you should use another entry, one that is not yet covered by the existing taxonomy.
So we'll add tests for:
- all regexp rules for a language;
- all processes, including synonyms for each language;
- all false positives for each language;

PR #3143, opened 2020-03-28
repaired de:mark

aleene:
de:mark is not the same as de:püree. Created en:pulp as synonym of de:mark.


stephanegigandet:
I think "pulp" is more a part than a processing, like "skin", "seeds", "kernel", "leaves" etc.

e.g. "orange pulp", "coconut pulp" etc.

So I would prefer that we remove it from the processing taxonomy.

But it's a very interesting point. Maybe we could create a "parts" taxonomy, and use it so that we can drastically reduce the number of needed entries in the taxonomy.

aleene:
That is right. Let's continue this and try with some other parts, and see if it works for other parts as well. We can always split later into another taxonomy.

PR #3144, opened 2020-03-28
de:passiert

aleene:
**Description:**
de:passiert added


stephanegigandet:
I tried Google Translation on the few examples of ingredients with "passiert" we have, and I have no idea what it means:

image
image

?

aleene:
Indeed it is also happens. Check this, which hangy sent me: https://www.lebensmittellexikon.de/f0000920.php

You can buy it in France as passata (passierte Tomaten). Which makes me wonder whether de:mark should be under en:filtered.

stephanegigandet:
Isn't it more an ingredient by itself, instead of a processing? If it's mainly tomatoes, then it's easier to just have an entry for it in the taxonomy.

aleene:
Is possible. As it is a synonym of de:mark, I think it better to put into processing, otherwise we are still left with a line in the ingredients taxonomy. The tomato example was just to explain the word passiert (passata).

aleene:
I will move de:passiert to en:sieved, fr:tamisé

stephanegigandet:
> I will move `de:passiert` to `en:sieved`, `fr:tamisé`.

Sounds good.

PR #3145, opened 2020-03-28
de:entsteint

aleene:
**Description:**
de:entsteint added


t/ingredients_processing.t
```diff
@@ -197,20 +197,16 @@ my @tests = (
 				}
 			]
 		],
-# Test for de: ingredients, that should NOT be detected through processing
```
stephanegigandet: Was this test removed by error?

aleene: The same

PR #3148, opened 2020-03-28
de:eingelegt

aleene:
**Description:**
de:eingelegt


t/ingredients_processing.t
```diff
@@ -197,20 +197,16 @@ my @tests = (
 				}
 			]
 		],
-# Test for de: ingredients, that should NOT be detected through processing
```
stephanegigandet: was this test removed by error?

aleene: No, I removed the false positives tests, until we have a taxonomy update.

PR #3150, opened 2020-03-29
Added: Brand facet to edit page

### LINKED ISSUE #3087: New data quality facet: brand repeated in product name
Many people fill the brand in the product name. Example:
https://world.openfoodfacts.org/product/5060358171024/lucia-popperley-s-cappucino-popcorn-ten-acre?rev=11

When the "brand" field is completed, we should be able to identify if the product name contains the given brand.

(I understand it's not a big issue.)

muskangarg21:
Fixes #3087

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -1841,6 +1841,13 @@ sub product_name_brand($) {
 
 	if (defined $ref->{brands}) {
 		my $brand = $ref->{brands};
+		my @brand_all = split(',', $brand);
+		foreach my $i (@brand_all)
+		{
+			if (index($full_name,$i) != -1){
+				$full_name =~ s/\Q$i//ig;
+			}
+		}
 		$brand =~ s/,.*//;	# take the first brand
 		my $brandid = '-' . get_string_id_for_lang($lc, $brand) . '-';
 		my $full_name_id = '-' . get_string_id_for_lang($lc, $full_name) . '-';


lib/ProductOpener/Products.pm
```diff
@@ -1841,6 +1841,13 @@ sub product_name_brand($) {
 
 	if (defined $ref->{brands}) {
 		my $brand = $ref->{brands};
+		my @brand_all = split(',', $brand);
```
stephanegigandet: Hi @muskangarg21 ,
The product_name_brand() function is the function we use when we want to display a "name" for the product: we combine product name and brand.
But the issue #3087 is different: we want to add a data quality warning when we detect one of the brands from the brands field inside the product_name field.
So your code should go in the DataQualityCommon.pm module instead.

What we want is to be able to see those products when we access the https://in.openfoodfacts.org/data-quality url for instance.

We don't want to automatically remove the brands from the product name, as for some cases, the product name is really the brand name. e.g. for "Coca Cola", we consider it's both the product name and the brand name. But if it is something like "ABCD Chocolate Chip Cookies with Almonds", then we would prefer to remove ABCD from the product name.

Sorry if it's not very clear.

muskangarg21: Hi @stephanegigandet !
I am still not very clear to what needs to be done.
for eg: `ingredients-percent-analysis-ok` is the first category in the list on the link [https://in.openfoodfacts.org/data-quality](url) , so do we want to add a `brand repeated in product name` to that list ?
and could you suggest how do i add a new one, I can't find where the list is defined/referenced. 
Thank you

VaiTon: > so do we want to add a `brand repeated in product name` to that list ?

Exactly

> and could you suggest how do i add a new one, I can't find where the list is defined/referenced.

You have to create a sub in DataQualityCommon.pm or add this check to an already existent sub.

stephanegigandet: Hi @muskangarg21 : @vaiton is right, you can just create a new "brand-repeated-in-product-name" value, that you can push to the @data_quality_tags array, in DataQualityCommon.pm

stephanegigandet:
Closing this PR as it is 3 years old.

PR #3152, opened 2020-03-29
Divide status between: TO DO and DONE

### LINKED ISSUE #2866: Divide status between: TODO and DONE
image
Note: the sorting is bogus, as you might notice ;-)

areeshatariq:
Signed-off-by: Areesha Tariq <areeshatariq02@gmail.com>

**Description:**
The status is separated as 'To Do' and 'Done', based on whether the status indicates something is done ("Nutrition facts completed" etc.) or not done ("Characteristics to be completed").

**Output:**

Status
When Done does not contain any entry, only To Do is displayed:

To-Do

When To Do does not contain any entry, only Done is displayed:

Done

fixes #2866

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -6958,9 +6958,30 @@ sub display_field($$) {
 			}
 		}
 		my $lang_field = lang($field);
-		if ($lang_field eq '') {
+		if ($lang_field eq '' and $field ne 'states') {
 			$lang_field = ucfirst(lang($field . "_p"));
 		}
+		
+		# Separate To-Do and Done Status
+		if ($field eq 'states') {
+			my $done_status = '';
+			my $to_do_status = '';
+			my @status_split = split(',', $value);
+			foreach my $val (@status_split) {
+				if (index($val, "to-be") != -1) {
+		 			$to_do_status .=$val . ",";
+  		 		}
+		 		else {
+		 		$done_status .=$val . ",";
+		 		}
+			}
+			$to_do_status =~ s/,$//;
+			$done_status =~ s/,$//;
+			$lang_field = "To Do";
+			$value = $to_do_status;
+			$html .= '<p><span class="field">' . "Done" . separator_before_colon($lc)  . ":</span>" . $done_status . "</p>";
+		}
+		
 		$html .= '<p><span class="field">' . $lang_field . separator_before_colon($lc) . ":</span> $value</p>";
 
 		if ($field eq 'brands') {
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -4218,3 +4218,14 @@ msgctxt "move_data_and_photos_to_main_language_ignore"
 msgid "Keep existing values and selected photos"
 msgstr "Keep existing values and selected photos"
 
+msgctxt "move_data_and_photos_to_main_language_ignore"
+msgid "Keep existing values and selected photos"
+msgstr "Keep existing values and selected photos"
+
+msgctxt "done_status"
+msgid "Done"
+msgstr "Done"
+
+msgctxt "to_do_status"
+msgid "To do"
+msgstr "To do"
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -4218,13 +4218,6 @@ msgctxt "move_data_and_photos_to_main_language_ignore"
 msgid "Keep existing values and selected photos"
 msgstr "Keep existing values and selected photos"
 
-<<<<<<< HEAD
-msgctxt "move_data_and_photos_to_main_language_ignore"
-msgid "Keep existing values and selected photos"
-msgstr "Keep existing values and selected photos"
-
-=======
->>>>>>> 5f872810091d8f48a6135147011c75707a8f748e
 msgctxt "done_status"
 msgid "Done"
 msgstr "Done"
--- a/po/common/en.po
+++ b/po/common/en.po
@@ -4205,8 +4205,4 @@ msgstr "Done"
 
 msgctxt "to_do_status"
 msgid "To do"
-<<<<<<< HEAD
 msgstr "To do"
-=======
-msgstr "To do"
->>>>>>> 5f872810091d8f48a6135147011c75707a8f748e
--- a/po/common/en.po
+++ b/po/common/en.po
@@ -4198,3 +4198,11 @@ msgstr "Replace existing values and selected photos"
 msgctxt "move_data_and_photos_to_main_language_ignore"
 msgid "Keep existing values and selected photos"
 msgstr "Keep existing values and selected photos"
+
+msgctxt "done_status"
+msgid "Done"
+msgstr "Done"
+
+msgctxt "to_do_status"
+msgid "To do"
+msgstr "To do"


lib/ProductOpener/Display.pm
```diff
@@ -6958,9 +6958,30 @@ sub display_field($$) {
...
+				if (index($val, "to-be") != -1) {
+		 			$to_do_status .=$val . ",";
+  		 		}
+		 		else {
+		 		$done_status .=$val . ",";
+		 		}
+			}
+			$to_do_status =~ s/,$//;
+			$done_status =~ s/,$//;
+			$lang_field = "To Do";
```
stephanegigandet: We use .po files so that we have translated versions (see /po/commons/en.po ). New entries need to be added to the po/commons/common.pot file too.

lib/ProductOpener/Display.pm
```diff
@@ -6958,9 +6958,30 @@ sub display_field($$) {
...
+			$value = $to_do_status;
+			$html .= '<p><span class="field">' . "Done" . separator_before_colon($lc)  . ":</span>" . $done_status . "</p>";
```
stephanegigandet: It would be good not to print the "Done" line or the "To do" line if it does not contain any entry.

teolemon:
Can you provide a small screenshot of the changes in the PR description ?

lib/ProductOpener/Display.pm
areeshatariq: Ok, I'll look into this.

areeshatariq:
> Can you provide a small screenshot of the changes in the PR description?

areeshatariq:
> Can you provide a small screenshot of the changes in the PR description ?

@teolemon of the code or the output? Just added the screenshot of the output.

teolemon:
👍 great

VaiTon:
@areeshatariq news? I'd love to merge this feature!

areeshatariq:
@VaiTon @stephanegigandet , kindly review

lib/ProductOpener/Display.pm
```diff
@@ -6958,9 +6958,47 @@ sub display_field($$) {
...
+			if ($no_entry_to_do_status eq $incomplete){
+				$lang_field = "Done";
+				$value = $done_status;
+			}
+			else {
+				$lang_field = "To Do";
+				$value = $to_do_status;
+			}
+			if ($no_entry_done_status ne $empty and $no_entry_to_do_status ne $incomplete){
+				$html .= '<p><span class="field">' . "Done" . separator_before_colon($lc)  . ":</span>" . $done_status . "</p>";
```
stephanegigandet: So here, you can replace "Done" by lang("done_status"), and that will return the translated value for "Done". In English it will be done, but in French it will return "Fait" (once we have translated it of course).

To do that, you need entries like this in the .po file: 

msgctxt "done_status"
msgid "Done"
msgstr "Done"

msgctxt "to_do_status"
msgid "To do"
msgstr "To do"

lib/ProductOpener/Display.pm
```diff
@@ -6958,9 +6958,47 @@ sub display_field($$) {
...
+			$no_entry_to_do_status =~ s/<a[^>]*>|<\/a>//;
+			if ($no_entry_to_do_status eq $incomplete){
+				$lang_field = "Done";
+				$value = $done_status;
+			}
+			else {
+				$lang_field = "To Do";
+				$value = $to_do_status;
+			}
+			if ($no_entry_done_status ne $empty and $no_entry_to_do_status ne $incomplete){
```
stephanegigandet: Here I think it would be much easier to just test this:

if ($done_status ne "") {
$html .= '<p><span class="field">' . "lang("done_status") . separator_before_colon($lc)  . ":</span>" . $done_status . "</p>";
}

and then just after:

if ($to_do_status ne "") {
$html .= '<p><span class="field">' . "lang("to_do_status") . separator_before_colon($lc)  . ":</span>" . $to_do_status . "</p>";
}

lib/ProductOpener/Display.pm
```diff
@@ -6958,9 +6958,47 @@ sub display_field($$) {
...
+				$html .= '<p><span class="field">' . "Done" . separator_before_colon($lc)  . ":</span>" . $done_status . "</p>";
+			}
+		}
+
```
stephanegigandet: Here you can add an else { 
$html .= '<p><span class="field">' . $lang_field . separator_before_colon($lc) . ":</span> $value</p>";
}

so that you don't print it again (you will already have printed the "Done" and "To do" values above)

lib/ProductOpener/Display.pm
areeshatariq: @stephanegigandet, yes this is an easier approach but I did this because $done_status and $to_do_status are not null even when they have no entries. $done_status contains "Empty" in anchor tag and $to_do_status contains "to be checked" in anchor tag.

lib/ProductOpener/Display.pm
areeshatariq: @stephanegigandet, so for entries that are not to be printed, should they not be added in .po files?

stephanegigandet: the .po files are for strings that we want to display on the web site. The web site can be in any language, so any string we display should come from the .po file, thanks to the lang() function.

lib/ProductOpener/Display.pm
stephanegigandet: "Empty" should go in "To do" in fact. It indicates the product is empty, so it should be completed.

"To be checked" should go in "To do", it's fine to display it always (unless if it is not present in the product has been checked).

areeshatariq:
@stephanegigandet any idea why the tests are failing?

stephanegigandet:
> @stephanegigandet any idea why the tests are failing?

It seems to be unrelated to your changes. I'm investigating.

stephanegigandet:
@areeshatariq : could you try merging the build-taxonomies branch in your branch? that may solve the failing tests

stephanegigandet:
@areeshatariq : so it does work, but only in English, as the checks are on the localized values, not on the normalized tag.... 

See the status for a product on the French site for instance: https://fr.openfoodfacts.org/product/3450970007023/vinaigre-d-alcool-colore-6-d-acidite-1-litre-co-eco

stephanegigandet:
image

stephanegigandet:
Sorry I didn't catch that earlier. I think the logic you created needs to be moved a bit before, in this code:
````
	if (defined $taxonomy_fields{$field}) {
		$value = display_tags_hierarchy_taxonomy($lc, $field, $product_ref->{$field . "_hierarchy"});
	}
````
That's where we create the $value with things like " Complete, Nutrition facts completed, Ingredients completed, Characteristics completed" etc.

$product_ref->{$field . "_hierarchy"} is a reference to an array: 

````
[
"en:to-be-checked",
"en:complete",
"en:nutrition-facts-completed",
"en:ingredients-completed",
"en:expiration-date-to-be-completed",
"en:packaging-code-to-be-completed",
"en:characteristics-completed",
"en:categories-completed",
"en:brands-completed",
"en:packaging-completed",
"en:quantity-completed",
"en:product-name-completed",
"en:photos-validated",
"en:photos-uploaded"
],
````

So what we need to do is to take this array as a source, and to create 2 new arrays: the "to do" array that contains the fields that contain "-to-" or the "en:empty", and a "done" array with the rest.

Then you can call display_tags_hierarchy_taxonomy() on each of those 2 arrays, to store the $todo_value and $done_vallue.

Can you try to do that?

areeshatariq:
@stephanegigandet , yes I am looking into this

Issue #3125, opened 2020-03-27
Github tests fail because too many things are printed to STDERR


teolemon:
```
   'nutriment canonicalized {label => "Omega 3-vetzuren",lc => "nl",nid => "omega-3-fat"}\n' +
    'initializing sublabel {lc => "nl",nid => "omega-3-fat",sublabel => "omega-3-fat"}\n' +
    'nutriment canonicalized {label => "Om'
}
(node:35902) UnhandledPromiseRejectionWarning: RangeError [ERR_CHILD_PROCESS_STDIO_MAXBUFFER]: stderr maxBuffer length exceeded
    at Socket.onChildStderr (child_process.js:386:14)
    at Socket.emit (events.js:311:20)
    at addChunk (_stream_readable.js:294:12)
    at readableAddChunk (_stream_readable.js:271:13)
    at Socket.Readable.push (_stream_readable.js:209:10)
    at Pipe.onStreamRead (internal/stream_base_commons.js:186:23)
(node:35902) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:35902) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

teolemon:
believe it was fixed by @hangy. Closing

PR #3155, opened 2020-03-29
Reduce STDERR output during CI

hangy:
**Description:** The default test output is reduced by only logging `error` or above to the TAP. If required for debugging, the `TAP_LOG_FILTER` environment variable can be set to a different level to override it.
**Related issues and discussion:** #3125


t/allergens_tags.t
```diff
@@ -7,7 +7,6 @@ use utf8;
 
 use Test::More;
 use Log::Any::Adapter 'TAP';
-#use Log::Any::Adapter 'TAP', filter => "none";
```
stephanegigandet: I'd prefer to keep those commented lines: I keep uncommenting / commenting them when I debug the tests.

t/ingredients_nesting.t
```diff
@@ -309,7 +310,7 @@ foreach my $test_ref (@tests) {
 		# inside the test file much easier when tests results need
 		# to be updated. Caveat is that it might interfere with
 		# test output.
-		or print STDERR join("\n", explain $product_ref->{ingredients});
+		or diag join("\n", explain $product_ref->{ingredients});
```
stephanegigandet: Will we still get this in the test output when the tests fail? We do need it.

hangy: Yes, but better. 😄 `diag` just prints the message, but ensures that the order of messages in the test log is correct.

t/allergens_tags.t
hangy: You can just run `prove` with a different `TAP_LOG_FILTER`, though. Keeping them in the source leaves the risk of accidentally committing them in uncommented form. I actually removed few of these instances in this commit.

teolemon:
fixed a little merge conflict

stephanegigandet:
@hangy I'm not sure why, but all the traces for the script convert_auchan_data.pl now go to STDERR and it makes the test fail.

````
RangeError [ERR_CHILD_PROCESS_STDIO_MAXBUFFER]: stderr maxBuffer length exceeded
    at Socket.onChildStderr (child_process.js:386:14)
    at Socket.emit (events.js:311:20)
    at addChunk (_stream_readable.js:294:12)
    at readableAddChunk (_stream_readable.js:271:13)
    at Socket.Readable.push (_stream_readable.js:209:10)
    at Pipe.onStreamRead (internal/stream_base_commons.js:186:23) {
  code: 'ERR_CHILD_PROCESS_STDIO_MAXBUFFER',
  cmd: 'perl -c -CS -Ilib scripts/convert_auchan_data.pl',
  stdout: '',

hangy:
> @hangy I'm not sure why, but all the traces for the script convert_auchan_data.pl now go to STDERR and it makes the test fail.
> 
> ```
> RangeError [ERR_CHILD_PROCESS_STDIO_MAXBUFFER]: stderr maxBuffer length exceeded
>     at Socket.onChildStderr (child_process.js:386:14)
>     at Socket.emit (events.js:311:20)
>     at addChunk (_stream_readable.js:294:12)
>     at readableAddChunk (_stream_readable.js:271:13)
>     at Socket.Readable.push (_stream_readable.js:209:10)
>     at Pipe.onStreamRead (internal/stream_base_commons.js:186:23) {
>   code: 'ERR_CHILD_PROCESS_STDIO_MAXBUFFER',
>   cmd: 'perl -c -CS -Ilib scripts/convert_auchan_data.pl',
>   stdout: '',
> ```

The script probably never worked before. `$lc` isn't defined in that line: https://github.com/openfoodfacts/openfoodfacts-server/blob/7aaeb1749fee97ce0e3f6ff80dcfadbb85f0dfa8/scripts/convert_auchan_data.pl#L59
As to why it never showed an error during CI, I do not know.

stephanegigandet:
@hangy : with those changes, will we be testing many more things than before? e.g. all the scripts in /scripts maybe ?

"Failing after 131m "  --> it will be very problematic if tests take 2 hours.

t/allergens_tags.t
stephanegigandet: "You can just run prove with a different TAP_LOG_FILTER"
--> how do I do that exactly?

hangy:
> @hangy : with those changes, will we be testing many more things than before? e.g. all the scripts in /scripts maybe ?

No, not really. We're compiling the scripts in `master`, too: https://github.com/openfoodfacts/openfoodfacts-server/blob/46b6657/.github/workflows/pull_request.yml#L84
There might be some hidden issue that makes it appear like everything is okay, but I did intent to ensure that scripts compile, as well.

> "Failing after 131m " --> it will be very problematic if tests take 2 hours.

Agreed. That's much too long. Simply checking the Perl syntax/compiling the scripts really shouldn't take that long, though. So there is probably some hidden issue that causes too much data to be processed during compilation. I'll have a look ...

t/allergens_tags.t
hangy: Just set that environment variable when running prove, ie, `TAP_LOG_FILTER=error prove -l -j9`.

hangy:
@stephanegigandet currently, I have no idea what might be causing the build to time out. 🤷‍♂️ I'll look into splitting it up, since some of the changes should make sense on their own.

VaiTon:
@hangy tests failing

stephanegigandet:
t/producers.t                  (Wstat: 512 Tests: 0 Failed: 0)
  Non-zero exit status: 2
  Parse errors: No plan found in TAP output

-> that's weird, I don't understand why it fails.

hangy:
> -> that's weird, I don't understand why it fails.

Me neither. :) I didn't think I modified the files in that way...

VaiTon:
@hangy do you still want to merge this?

PR #3161, opened 2020-03-29
de:mariniert

aleene:
**Description:**
de:mariniert


t/ingredients_processing.t
```diff
@@ -194,24 +194,9 @@ my @tests = (
 				'id' => 'en:shallot',
 				'processing' => 'en:pureed',
 				'text' => 'Schalotte'
-				}
```
stephanegigandet: can you put those tests back? with the new built ingredients taxonomy, they should work

aleene: I added them in

PR #3163, opened 2020-03-30
Update variable_weight_products.txt

TaciteOFF:
Ajout de plusieurs ranges


VaiTon:
@TaciteOFF could we add a link to where did you take this information?

teolemon:
It's a manual listing from the products we have in OFF. Trying to make templates we can apply to the other products

PR #3174, opened 2020-03-31
de:gesüßt and de:geölt

aleene:
**Description:**
de:gesüßt and de:geölt


stephanegigandet:
When it doesn't exist yet, can you add the processing in English as well? That way we won't have to change the German tests when we add it later.

That would mean that we need to check that it works well for English too.

aleene:
Patience

PR #3182, opened 2020-04-02
Fix scripts, move old script to obsolete, comment out traces that flood github CI logs

stephanegigandet:
Github continuous integration build logs are filled with thousands of lines of traces output like:

    'nutriment canonicalized {label => "kcal",lc => "zh_CN",nid => "kcal"}\n' +
    'initializing sublabel {lc => "zh_CN",nid => "energy-kcal",sublabel => "kcal"}\n' +

We don't need those traces (they are only useful for debugging), so while it may not be the best solution to comment them, in the short term, there's no downside and it will make exploiting build logs much easier.


VaiTon:
Can't we just set the logging level above TRACE?

stephanegigandet:
> Can't we just set the logging level above TRACE?

I'm sure it's possible. I don't know where it's done though, for the github CI tests. So as we absolutely don't need those tests, except in rare cases where we would need to debug something related to that, that's an easy (and most importantly, quick) workaround.

stephanegigandet:
Build failing:
[Thu Apr  2 08:42:06 2020] convert_auchan_data.pl: Global symbol "$lc" requires explicit package name (did you forget to declare "my $lc"?) at scripts/convert_auchan_data.pl line 59.
[Thu Apr  2 08:42:06 2020] convert_auchan_data.pl: scripts/convert_auchan_data.pl had compilation errors.

This script hasn't been touched for ages... :-|

stephanegigandet:
For some reason it seems that removing things that were printed to STDERR makes github test more things, like all the scripts. (similar to what seems to happen for PR https://github.com/openfoodfacts/openfoodfacts-server/pull/3155 ).

So I'm fixing the scripts.

stephanegigandet:
Hi @hangy , the tests are now passing. This PR may solve the failing tests in https://github.com/openfoodfacts/openfoodfacts-server/pull/3155

PR #3183, opened 2020-04-02
de:gepoppt, de:zerkleinert

aleene:
**Description:**
de:gepoppt, de:zerkleinert


taxonomies/ingredients_processing.txt
```diff
@@ -47,6 +47,24 @@ de:feingeschnitten, fein geschnittenen, feingeschnittener
 # <en:cut
 de:handgeschnitten, handgeschnittene
 
+de:zerkleinert, zerkleinerte, zerkleinerter, zerkleinertes
```
stephanegigandet, line 50: Could you add the corresponding English entries?  otherwise we will have to change the test results when we add the English entries later.

aleene: No worries, I'll do that. I first finish the german entries and then I'll move to the english one.

By the way do not count on it that everything will be translatable. I already have a counter example.

stephanegigandet: What is that counter example?

aleene: That is `de:geschält`, i.e. `en:peeled` and `en:shelled`. In german there is only one word for removing the outer layer (at least as it used for the ingredients).

I have been looking at the english ones. That is however more work, as there are much more errors in the ingredients.

PR #3188, opened 2020-04-02
Added italian translations to ingredients taxonomy

VaiTon:


taxonomies/ingredients.txt
```diff
@@ -294,7 +294,7 @@ en:agente coagulante, agentes coagulantes
 es:Agentes coagulantes
 fi:hyydyte, koagulantti
 fr:agent de coagulation, coagulant
-it:agente coagulante, coagulante
```
stephanegigandet, line 297: Could you also add all those to the additives-classes.txt taxonomy?
All the entries from the additives-classes taxonomy will be automatically added to the ingredients taxonomy as well.

VaiTon: Done! See b6b96c9fd87101895384bae457ef61dc0007b79c

Issue #59, opened 2015-03-04
Be able to belong to a team


kyzh:
One of the other way to get people to participate where they are strong or interested is to pool effort in a team.
While it could be lightly or strongly coupled with the current system.

Something that allow to drill down (like contributor, states, etc..) and provide another way to  view data.
This would foster more competitions.
This allows for people to complete more product by sharing tasks (eg. people interested in categories and people interested in allergies filling respective fields)

fabricetheytaz:
:+1:

TaciteOFF:
I'm doing it unofficially with Sebleouf who fills the ingredients and I fill the characteristics and nutrition facts. 
An "official" way of doing it could be cool :)

teolemon:
We could even automate joining slack channels "Paris team" "Reviewer teams"…

stephanegigandet:
I plan to work on this feature soon. :) (I'm only 5 years late).

My plan is to do something very simple: in the user profile form, allow everyone to add 1 to 3 teams (essentially just tags) of their choice.

And when they edit products, tag the product with a new facet /team/anything

That way with 3 teams, it allow people to cooperate with close friends or family member + a wider group + a much wider group (like a country).

Also planning to do something similar with a new /mission/ facet, but on a product by product basis, with an easy way to switch missions.

teolemon:
deployed.

PR #3205, opened 2020-04-03
Allow users to be part of teams + new teams facet

stephanegigandet:
issue #59


stephanegigandet:
Currently live on .dev:

image

image

VaiTon:
I think we should make teams, as they are based on team name, case insensitive and without whitespaces.

stephanegigandet:
> I think we should make teams, as they are based on team name, case insensitive and without whitespaces.

You are right, I changed it.

PR #3207, opened 2020-04-04
more salts

stephanegigandet:
Continuing on @raphael0202 's work.

### DIFF
--- a/taxonomies/ingredients.result.txt
+++ b/taxonomies/ingredients.result.txt
@@ -3564,8 +3564,13 @@ en:beef broth
 de:Rinderbrühe, Rindfleischbouillon
 es:Caldo de carne de vaca
 fi:naudanlihaliemi
-fr:bouillon de boeuf
+fr:bouillon de bœuf
 it:brodo di manzo
+sv:nötköttsbuljong
+
+< en:Broth
+en:cooking broth
+fr:bouillon de cuisson
 
 < en:Broth
 < en:Pork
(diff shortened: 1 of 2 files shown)


TaciteOFF:
I found this one too 

https://world.openfoodfacts.org/ingredient/fr:sel-de-mer-fume

And this product has it in many languages : https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=5017764900373

en:smoked sea salt
de:geräuchertes Meersalz
fr:sel de mer fumé
it:sale marino affumicato
nl:gerookt zeezout
es:sal marina ahumada

stephanegigandet:
> en:smoked sea salt
> de:geräuchertes Meersalz
> fr:sel de mer fumé
> it:sale marino affumicato
> nl:gerookt zeezout
> es:sal marina ahumada

Thanks @TaciteOFF . For those we have a new strategy: we will introduce "smoked" as a processing. That way "smoked some-ingredient" or "un-ingrédient fumé" will be automatically recognized, and we can keep a smaller ingredients taxonomy that actually matches more ingredients.

Have a look at https://github.com/openfoodfacts/openfoodfacts-server/blob/master/taxonomies/ingredients_processing.txt

We are doing it very slowly (@aleene is working on the German and Dutch ones) as we first need to analyze the ingredients, to make sure we won't create false positives.

So we should be able to catch a large part of those: https://fr.openfoodfacts.org/ingredients?filter=-fume

stephanegigandet:
New "smoked" processing: https://github.com/openfoodfacts/openfoodfacts-server/pull/3207#issuecomment-609043973

PR #3211, opened 2020-04-04
Display To Do and Done status separately (For all languages)

### LINKED ISSUE #2866: Divide status between: TODO and DONE

areeshatariq:
Signed-off-by: Areesha Tariq <areeshatariq02@gmail.com>

**Description:**

The status is separated as 'To Do' and 'Done', based on whether the status indicates something is done ("Nutrition facts completed" etc.) or not done ("Characteristics to be completed").
This works for all languages
Both the arrays to_do and done contained an extra anchor tag in the beginning
I removed the first element from both the arrays using slicing and then it worked fine

**Related issues and discussion:** fixes #2866


lib/ProductOpener/Display.pm
```diff
@@ -6923,8 +6923,30 @@ sub display_field($$) {
 			$value =~ s/\n/<br>/g;
 		}
 	}
+	
+	my $to_do_status = '';
+	my $done_status = '';
+
+	if ($field eq 'states'){
+		my @to_do_status = '';
```
stephanegigandet, line 6931: That's why your array has an extra element, this assigns ''  (an empty string) to the first element of the array.

To create an empty array:

my @to_do_status = ();

lib/ProductOpener/Display.pm
```diff
@@ -6923,8 +6923,28 @@ sub display_field($$) {
...
+					push(@done_status, $val);
+			}
+			
+		}
+		$to_do_status = display_tags_hierarchy_taxonomy($lc, $field, \@to_do_status);
+		$done_status = display_tags_hierarchy_taxonomy($lc, $field, \@done_status);
+	}
 
-	if (defined $taxonomy_fields{$field}) {
+	if (defined $taxonomy_fields{$field} and $field ne 'states') {
```
VaiTon, line 6947: Can we use an `elif`?

areeshatariq: Yes, would be better, will do that. Thanks

lib/ProductOpener/Display.pm
```diff
@@ -6964,19 +6983,6 @@ sub display_field($$) {
 		
 		# Separate To-Do and Done Status
 		if ($field eq 'states') {
```
stephanegigandet, line 6985: What's in the "if ($field eq 'states') {" condition should now be moved above, (after you set the values $to_do_status and $done_status), as you already test for that condition above.

lib/ProductOpener/Display.pm
```diff
@@ -6964,19 +6983,6 @@ sub display_field($$) {
...
-				if ((index($val, "to-be") != -1) or (index($val, "Empty") != -1)) {
-		 			$to_do_status .=$val . ",";
-  		 		}
-		 		else {
-		 		$done_status .=$val . ",";
-		 		}
-			}
-			$to_do_status =~ s/,$//;
-			$done_status =~ s/,$//;
 			if ($to_do_status ne ""){
```
stephanegigandet, line 6986: What's in the "if ($field eq 'states') {" condition should now be moved above, (after you set the values $to_do_status and $done_status), as you already test for that condition above.

areeshatariq: So should I add "if ($field ne 'states') { $html .= '<p><span class="field">' . $lang_field . separator_before_colon($lc) . ":</span> $value</p>";}"
So that the status does not print again

stephanegigandet: Yes, you can do it like that.

PR #3228, opened 2020-04-07
convert scientific notation to decimal 

### LINKED ISSUE #3227: CSV export contains numbers with exponential notation
e.g.  0       0.004   0.004857        0.0057140.000571        7e-05           1.71e-06

This is probably due to the different ways Perl can store numbers internally. (see https://perldoc.perl.org/perlnumber.html )

stephanegigandet:
fixes #3227

### DIFF
--- a/scripts/export_database.pl
+++ b/scripts/export_database.pl
@@ -398,6 +398,8 @@ XML
 				if ($value =~ /e/) {
 					# 7e-05 1.71e-06
 					$value = sprintf("%.10f", $value);
+					# Remove trailing 0s
+					$value =~ s/\.([0-9]+?)0*$/\.$1/g;
 				}
 				$csv .= $value . "\t";
 			}
--- a/scripts/export_database.pl
+++ b/scripts/export_database.pl
@@ -394,7 +394,12 @@ XML
 			$nid =~ s/-$//g;
 
 			if (defined $product_ref->{nutriments}{$nid . "_100g"}) {
-			$csv .= $product_ref->{nutriments}{$nid . "_100g"} . "\t";
+				my $value = $product_ref->{nutriments}{$nid . "_100g"};
+				if ($value =~ /e/) {
+					# 7e-05 1.71e-06
+					$value = sprintf("%.10f", $value);
+				}
+				$csv .= $value . "\t";
 			}
 			else {
 				$csv .= "\t";


scripts/export_database.pl
```diff
@@ -394,7 +394,12 @@ sub sanitize_field_content {
 			$nid =~ s/-$//g;
 
 			if (defined $product_ref->{nutriments}{$nid . "_100g"}) {
-			$csv .= $product_ref->{nutriments}{$nid . "_100g"} . "\t";
+				my $value = $product_ref->{nutriments}{$nid . "_100g"};
+				if ($value =~ /e/) {
+					# 7e-05 1.71e-06
+					$value = sprintf("%.10f", $value);
```
CharlesNepote, line 400: Are we really need 9 decimal digits? Where are the unit tests ;-)?

stephanegigandet: We do need many decimal digits for some nutrients that are in micrograms. e.g.https://world.openfoodfacts.org/product/0746742290308/sparkling-water-beverage-with-juice-black-bear

We have 7.5e-08. That's 0.000000075

stephanegigandet: There are no unit tests for the CSV export, we can add some, but it will take time and it should be a different bug.

VaiTon: Can't we just add as decimals as we need for a specific number?

Eg: `1.71e-02 -> 0.0171`, `1.71e-06 -> 0,00000171`

From [this page](https://perldoc.perl.org/functions/sprintf.html) in perldocs:
```perl
printf '<%.1f>', 1;  # prints "<1.0>
```

stephanegigandet: @VaiTon : it's possible, I'm just not sure it's worth risking introducing a bug while doing it. :)
e.g. we could use a regex to try to remove trailing 0s, or we could try to parse the 1.71e-02 to know that we need 4 digits. I'm just worried we'll forget about some specific case and break things by trying to be too clever.

VaiTon: > we could use a regex to try to remove trailing 0s

Maybe this is the best option at the moment.

stephanegigandet: @VaiTon @CharlesNepote : done, can you approve the changes? thanks!

PR #3229, opened 2020-04-07
Pumpernickel and churros

moon-rabbitOFF:


taxonomies/categories.txt
```diff
@@ -27513,9 +27521,6 @@ ciqual_food_code:en:19666
 ciqual_food_name:en:Petit-Suisse like fresh cream cheese, plain, around 9% fat
 ciqual_food_name:fr:Fromage frais type petit suisse, nature, 10% MG environ
 
-
-en:Coffee creamers
```
teolemon, line 27517: I added that for the US import. Not sure what parent to add. Dairy products ?

moon-rabbitOFF: From what I see Coffee creamers does not seem to contain milk. We already have a category "Creamers" in milk subtitute, I assume it's the same thing

stephanegigandet: @moon-rabbitOFF @teolemon  It is the same thing indeed, let's just add coffee creamers as a synonym of creamers.

PR #3238, opened 2020-04-09
Sweetened process

aleene:
**Description:**
Sweetened


taxonomies/ingredients_processing.txt
```diff
@@ -303,8 +303,26 @@ de:gepökelt, gepökeltes
...
+nl:gezoet, gezoete
+pt:adoçado, adoçada
+
+en:unsweetened
+nl:ongezoet
+
+#<en:sweetened
+en:sugared
+de:gezuckert, gezuckerte
+fr:sucré, sucrée, sucrés, sucrées, sucree, sucrees #better not add fr:sucre and fr:sucres
```
stephanegigandet, line 318: comments need to be on a separate line, otherwise it will break

aleene: OK, repaired

PR #3239, opened 2020-04-09
Adding a brand owner field for the US import

teolemon:
Adding a brand owner field for the US import

### DIFF
--- a/lib/ProductOpener/Config_off.pm
+++ b/lib/ProductOpener/Config_off.pm
@@ -493,7 +493,7 @@ improvements
 
 # fields in product edit form, above ingredients and nutrition facts
 
-@product_fields = qw(quantity packaging brands categories labels origins manufacturing_places
+@product_fields = qw(quantity packaging brands 	brand_owner categories labels origins manufacturing_places
  emb_codes link expiration_date purchase_places stores countries  );
 
 # fields currently not shown in the default edit form, can be used in imports or advanced edit forms
@@ -536,6 +536,7 @@ improvements
 	quantity
 	packaging
 	brands
+	brand_owner
 	categories
 	labels
 	origin
@@ -565,6 +566,7 @@ improvements
 
 @drilldown_fields = qw(
 	brands
+	brand_owner
 	categories
 	labels
 	packaging
@@ -630,6 +632,7 @@ improvements
 	pnns_groups_1
 	pnns_groups_2
 	states
+	brand_owner
 );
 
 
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Config_off.pm
```diff
@@ -493,7 +493,7 @@ improvements
 
 # fields in product edit form, above ingredients and nutrition facts
 
-@product_fields = qw(quantity packaging brands categories labels origins manufacturing_places
+@product_fields = qw(quantity packaging brands 	brand_owner categories labels origins manufacturing_places
```
stephanegigandet, line 496: This will make the brand owner field viewable and editable in the product edit form. I'm not sure if we really want to do that (as we will possibly get lots of strange values from users for other products).

lib/ProductOpener/Config_off.pm
```diff
@@ -565,6 +566,7 @@ improvements
 
 @drilldown_fields = qw(
 	brands
+	brand_owner
```
stephanegigandet, line 569: drill down fields can only be facets. I don't think it's useful to make this a facet. (we are close to the 64 facets allowed by mongodb indexes)

lib/ProductOpener/Config_off.pm
```diff
@@ -630,6 +632,7 @@ improvements
 	pnns_groups_1
 	pnns_groups_2
 	states
+	brand_owner
```
stephanegigandet, line 635: field is twice in the list, but see comment above

po/common/common.pot
```diff
@@ -248,6 +248,34 @@ msgctxt "brands_without_products"
...
+msgctxt "brand_owner"
+msgid "Brand owner"
+msgstr ""
+
+msgctxt "brand_owner_example"
+msgid "The Coca Cola Company"
+msgstr ""
+
+msgctxt "brands_p"
+msgid "brand owner"
```
stephanegigandet, line 260:
```suggestion
msgid "brand owners"
```

lib/ProductOpener/Config_off.pm
stephanegigandet: if we do want to make brand_owner a facet, then we would need to add values in po/tags/

lib/ProductOpener/Config_off.pm
teolemon: ok, fixing

teolemon, line 496:
```suggestion
@product_fields = qw(quantity packaging brands categories labels origins manufacturing_places
```

lib/ProductOpener/Config_off.pm
teolemon, line 569:
```suggestion
```

teolemon: fixed

PR #3242, opened 2020-04-10
Use estimated_document_count for world root URL

syl10100:
Replace count_documents by estimated_document_count for https://world.openfoodfacts.org/

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -3869,9 +3869,17 @@ sub search_and_display_products($$$$$) {
 			}
 			else {
 				$log->debug("Counting MongoDB documents for query", { query => $query_ref }) if $log->is_debug();
-				$count = execute_query(sub {
-					return get_products_collection()->count_documents($query_ref);
-				});
+				# test if query_ref is empty
+				if (keys %{$query_ref} > 0) {
+					$count = execute_query(sub {
+						return get_products_collection()->count_documents($query_ref);
+					});
+				} else {
+				# if query_ref is empty (root URL world.openfoodfacts.org) use estimated_document_count for better performance
+					$count = execute_query(sub {
+						return get_products_collection()->estimated_document_count();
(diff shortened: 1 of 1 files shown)


hangy:
Have you tested how far off the estimate is? If I remember correctly, it can be quite far off, which could mess with the pagination.

stephanegigandet:
on the dev server (which has an outdated database), it returns the same:

> db.products.count()
950695
> db.products.estimatedDocumentCount()
950695

But I just discovered that for some reason, mongo on production is 4.0.2 while on dev it's 4.0.6.

And estimatedDocumentCount() is new in 4.0.3... 

I guess we'll have to update mongodb and pray it doesn't break anything.

stephanegigandet:
I updated mongodb on the dev server, to check everything still works.

apt-get install  mongodb-org mongodb-org-mongos mongodb-org-server mongodb-org-shell

mongo --version   MongoDB shell version v4.0.17
git version: 0bc918c73390f6e4d6349660e4cd233f5900b69a
OpenSSL version: OpenSSL 1.1.0j  20 Nov 2018
allocator: tcmalloc
modules: none
build environment:
    distmod: debian92
    distarch: x86_64
    target_arch: x86_64

syl10100:
From documentation : https://docs.mongodb.com/manual/reference/method/db.collection.estimatedDocumentCount/ 
The cases where estimated count may be incorrect are : usage of shared cluster or unclean shutdown. 
I think it really worths to use this function.
And sorry @stephanegigandet I did not check the required mongo version

stephanegigandet:
mongodb on production (off2) updated to 4.0.17

> use off
switched to db off
> db.products.count()
1208925
> db.products.estimatedDocumentCount()
1208925

stephanegigandet:
Ah, I forgot about something: we don't want to do this on the platform for producers, as each producers sees only its own products, so the world count is different for each of them.

PR #3244, opened 2020-04-10
Add HTML id to user id

CharlesNepote:
Re-push previous comit.


stephanegigandet:
> Why? Is there any CSS or JavaScript script interacting with this?

Yes, but not in the OFF code per se. It's used by @CharlesNepote power user script: https://en.wiki.openfoodfacts.org/Editing_Tools#Open_Food_Facts_Power_User_Extension

PR #3245, opened 2020-04-10
handle both kinds of apostrophe characters for index and search, 

### LINKED ISSUE #3215: Search and apostrophe
**Describe the bug**
I regularly use my iPhone to search for products on the OpenFoodFacts website.
On my iPhone, the apostrophe is written " ’ ".
On my Macbook, the apostrophe is written " ' ". 
(It probably depends on which keyboard or system locales I'm using)
The problem is that the results I get are very different.
If I want to find the brand "Comme j'aime" for instance.

On my iPhone I get only 13 products when I get 88 products on the website.

Mobile : https://fr.openfoodfacts.org/cgi/search.pl?action=process&search_terms=Comme%20j’aime&sort_by=unique_scans_n&page_size=20

Desktop : https://fr.openfoodfacts.org/cgi/search.pl?search_terms=Comme+j%27aime&search_simple=1&action=process

**Expected behavior**
These ASCII characters should be considered the same and give the same results : 
’
'

Thank you.

stephanegigandet:
should fix #3215

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -1926,14 +1926,16 @@ sub index_product($)
 
 	my %keywords;
 
+	my $product_lc = $product_ref->{lc} ||$lc;
+
 	foreach my $field (@string_fields, @tag_fields) {
 		if (defined $product_ref->{$field}) {
-			foreach my $tag (split(/,|'|\s/, $product_ref->{$field} )) {
+			foreach my $tag (split(/,|'|’|\s/, $product_ref->{$field} )) {
 				if (($field eq 'categories') or ($field eq 'labels') or ($field eq 'origins')) {
 					$tag =~ s/^\w\w://;
 				}
 
-				my $tagid = get_string_id_for_lang($lc, $tag);
+				my $tagid = get_string_id_for_lang($product_lc, $tag);
 				if (length($tagid) >= 2) {
 					$keywords{normalize_search_terms($tagid)} = 1;
 				}


lib/ProductOpener/Products.pm
```diff
@@ -1926,14 +1926,16 @@ sub index_product($)
...
 
+	my $product_lc = $product_ref->{lc} ||$lc;
+
 	foreach my $field (@string_fields, @tag_fields) {
 		if (defined $product_ref->{$field}) {
-			foreach my $tag (split(/,|'|\s/, $product_ref->{$field} )) {
+			foreach my $tag (split(/,|'|’|\s/, $product_ref->{$field} )) {
 				if (($field eq 'categories') or ($field eq 'labels') or ($field eq 'origins')) {
 					$tag =~ s/^\w\w://;
 				}
```
hangy, line 1936: That's actually a right single quotation mark. Would it make sense to only do this if there's no left single quotation mark? I'm not really sure why a quotation mark should be used in place of an apostrophe.

stephanegigandet: Word processors like LibreOffice convert L'été to L’été. The use of the right quotation mark as an apostrophe seems widespread, in fact it seems to be the unicode standard

https://tedclancy.wordpress.com/2015/06/03/which-unicode-character-should-represent-the-english-apostrophe-and-why-the-unicode-committee-is-very-wrong/

stephanegigandet: Anyway, the goal of this is to make sure that in search, we remove articles so that queries like "ail" (garlic) match "crème à l'ail" (garlic cream). So even if it an actual quotation mark, it will be removed anyway.

hangy: Sounds like the consortium's guideline is controversial to say the least, but let's roll with it. 🤷‍♂️

PR #3246, opened 2020-04-10
fix for teams tags

stephanegigandet:
prevent empty team tag


VaiTon:
It could also be useful to check via js that the field is not empty.

PR #3249, opened 2020-04-11
remove not working feature to check if a userid exists when typing it…

stephanegigandet:
… in the account create form

the check_id.pl script was removed at some point (maybe for abuse / security issues, I don't remember) so that feature did not work at all.


teolemon:
From memory,  we disabled it during Envoyé Spécial

PR #3253, opened 2020-04-12
Update docker build script

VaiTon:
Also fixed issue preventing launching docker on windows where paths are expressed relatively.
 See https://github.com/docker/cli/issues/1203.


hangy:
I don't know … do you really want to require devs to install Python just for running a docker container? We could possibly just replace the manual docker commands by a docker-compose file and use `docker-compose … run …`?

VaiTon:
I think we should discuss if we want to use python and nodejs in general. We already have a lot of scripts writte in python while, for example, `refresh_taxonomies.js` is JS. 

What are the benefits of the first and what are the benefits of the second?

hangy:
> We already have a lot of scripts writte in python while, for example, `refresh_taxonomies.js` is JS.

Most of the Python scripts seem to be related to packager codes and are thus run manually in irregular intervals. I created `check_perl.js` and `refresh_taxonomies.js` to be used in the CI scripts, since I know parallel processing in JS better than in Bash, and we have NodeJS installed in CI anyways.

> What are the benefits of the first and what are the benefits of the second?

Node.js+npm are the de-facto standard for building website assets. Since we require Node.js to build (S)CSS+JS anyways, one advantage for JS scripts would be that this will be installed on most developer's machines anyways. On the other hand, Python is still ranked higher than JS on the TIOBE index. Usually, websites that don't use a  Python framework to run should don't require Python to be installed, though. Most of the current Python scripts in this repo are one-off scripts.

Issue #1560, opened 2018-12-07
Use projections to restrict fields returned by MongoDB


stephanegigandet:
From syl10100:

Projections are not used in any mongo query and we could improve performance by using it.
For instance, the home page http://fr.openfoodfacts.com returns 20 documents from mongodb but only few fields are displayed in query results.
One document is about 11kB, so 20 docs are about 220kB. Which means that 220kB are send over network from mongo to apache server each time someone connects to OFF website.
If we add a projection on only fields needed (product_name, code, brand, quantity , what else?), we could divide by 10 the resultset size and, consequently, optimize network throughput between servers.
To do it, projection operator should be add in query as follow :
`{projection => {"_id": 0, "product_name" => 1, "code" => 1, "brands" => 1, "quantity" => 1}}`

stephanegigandet:
A very good point. We'll need to be careful to implement this

1.  we currently use the same function Display::search_and_display_products for both the web site and the API. The API currently returns all fields, and it's possible that some apps depend on those fields being there.  E.g. adding .json or .xml to some urls will return the fields: https://fr.openfoodfacts.org/label/bio.json

2. we use mongodb to cache mongodb queries, the projection will need to be taken into account when generating the cache key

3. fields displayed (product name and images) are language specific, for the same product, we can display field values in different languages depending on the host the user is on (e.g. fr.openfoodfacts.org vs world.openfoodfacts.org)

syl10100:
How to test dataset size in mongo shell :
-without projection :
```
var cursor = db.products.find({"countries_tags" : "en:france"}).sort("sortkey" : -1).limit(20)
var size = 0
cursor.forEach(
    function(doc){
        size += Object.bsonsize(doc)
    }
);
print(size);
```
**-->207015 bytes**

-with projection :
```
var cursor = db.products.find({"countries_tags" : "en:france"},{"_id": 0, "product_name":1, "code":1, "brands": 1, "quantity": 1}).sort("sortkey" : -1).limit(20)
var size = 0
cursor.forEach(
    function(doc){
        size += Object.bsonsize(doc)
    }
);
print(size);
```
**-->2544 bytes**

So it is not divided by 10 but by almost 100

syl10100:
> 
> 
> A very good point. We'll need to be careful to implement this
> 
[...]

PR #3257, opened 2020-04-13
Limit fields returned by MongoDB

syl10100:
add $fields_ref for HTML pages results

 Issue #1560

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -3812,8 +3812,24 @@ sub search_and_display_products($$$$$) {
 	}
 
 	my $count;
+	
+	my $fields_ref;
 
-	my $mongodb_query_ref = [ lc => $lc, query => $query_ref, sort => $sort_ref, limit => $limit, skip => $skip ];
+	#for API (json, xml, rss,...), display all fields
+	if ($request_ref->{json} or $request_ref->{jsonp} or $request_ref->{xml} or $request_ref->{jqm} or $request_ref->{rss}) {
+		$fields_ref = {};
+	} else {
+	#for HTML, limit the fields we retrieve from MongoDB
+		$fields_ref = {
+		"code" => 1,
+		"product_name" => 1,
+		"product_name_$lc" => 1,
+		"brands" => 1,
+		"images" => 1,
+		"quantity" => 1
+		};
+	}
+	my $mongodb_query_ref = [ lc => $lc, query => $query_ref, fields => $fields_ref, sort => $sort_ref, limit => $limit, skip => $skip ];
 
 	my $key = $server_domain . "/" . freeze($mongodb_query_ref);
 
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -3849,6 +3849,17 @@ sub search_and_display_products($$$$$) {
 			products => [],
 		};
 
+		# Fields displayed in HTML results pages
```
stephanegigandet, line 3852: We are going to cache the results of the Mongo query. After we get the results back, we save them in memcached like that:

$memd->set($key, $request_ref->{structured_response}, 3600) or $log->debug("Could not set value for MongoDB query key", { key => $key }); 

So if we limit the fields returned, we need to make the key specific to those fields.

Currently the key is set above:
my $mongodb_query_ref = [ lc => $lc, query => $query_ref, sort => $sort_ref, limit => $limit, skip => $skip ];
my $key = $server_domain . "/" . freeze($mongodb_query_ref);
$key = md5_hex($key);

So we need to move the conditions for limiting the fields  (   if ($request_ref->{json} or $request_ref->{jsonp} or $request_ref->{xml} or $request_ref->{jqm} or $request_ref->{rss}) {   )  above, before we generate the key.

syl10100: Ok I got it. I will correct that.

lib/ProductOpener/Display.pm
```diff
@@ -3812,8 +3812,24 @@ sub search_and_display_products($$$$$) {
...
+		"product_name" => 1,
+		"product_name_$lc" => 1,
+		"brands" => 1,
+		"images" => 1,
+		"quantity" => 1
+		};
+	}
+	my $mongodb_query_ref = [ lc => $lc, query => $query_ref, fields => $fields_ref, sort => $sort_ref, limit => $limit, skip => $skip ];
 
 	my $key = $server_domain . "/" . freeze($mongodb_query_ref);
```
stephanegigandet, line 3834: I tested it on the .dev server, and I realized that freeze() does not sort hash keys, so at each call we get a different string, which makes the key different almost every time.

Can you replace the key generation with this?

	# tied hashes can't be encoded directly by JSON::PP, freeze the sort tied hash
	my $mongodb_query_ref = [ lc => $lc, query => $query_ref, fields => $fields_ref, sort => freeze($sort_ref), limit => $limit, skip => $skip ];

	# canonical makes JSON::PP sort the keys of hashes
	my $json = JSON::PP->new->utf8->canonical->encode($mongodb_query_ref);

	my $key = $server_domain . "/" . $json;

	$log->debug("MongoDB query key", { key => $key }) if $log->is_debug();

	$key = md5_hex($key);

syl10100: Yes sure I can but you mean you got a different key for same url ? Because I tested on my docker and I get same key for same url.

stephanegigandet: yes, in my case, I got many different keys

the fields were in different orders:

openfoodfacts.dev/12345678
 lcfr query     en:francecountries_tags field▒image▒
product_nam▒product_name_f▒brand▒codquantity sort
                                                 Tie::IxHas▒sortkey sort▒ limi▒ ski▒',
												 
 'openfoodfacts.dev/12345678
 lcfr query     en:francecountries_tags fieldquantit▒cod▒product_name_f▒brand▒
product_nam▒images sort
                       Tie::IxHas▒sortkey sort▒ limi▒ ski▒',request => 'LcAS9CtnFk9ZhRJC'}

syl10100: ok

PR #3281, opened 2020-04-17
small translation

teolemon:
small translation

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -474,6 +474,7 @@ ru:EAC
 wikidata:en:Q18407253
 
 fr:Entrepreneurs + Engagés, E+, Entrepreneurs Engagés
+de:Entrepreneurs + Engagés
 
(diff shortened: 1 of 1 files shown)


taxonomies/labels.txt
```diff
@@ -474,6 +474,7 @@ ru:EAC
 wikidata:en:Q18407253
 
 fr:Entrepreneurs + Engagés, E+, Entrepreneurs Engagés
+de:Entrepreneurs + Engagés
```
stephanegigandet, line 477: Why add this?

teolemon: Some values are stored in German, for some reason 🤷

hangy: I think that's the usual disruptive apps, probably some older changes, but we shouldn't update the taxonomy to reflect that IMHO. Not having a taxonomy entry for this actually makes it easier to fix the values.

teolemon: you are perfectly right, I closed the PR

PR #3294, opened 2020-04-19
add brand_owner import and display

stephanegigandet:


hangy:
So the brand owner would be s property of every single product, not the brand itself?

stephanegigandet:
> So the brand owner would be s property of every single product, not the brand itself?

Yes. We may have brand owner properties for the brands at some point too. But this is for 300K US products from the USDA database. They don't have a brand field, but they have a brand owner field.

PR #3298, opened 2020-04-20
More processes and translations

aleene:


aleene:
I can remove `natural`, but does it any harm?. Maybe the processes taxonomy should be extended to `organic` and `fair trade` as well? And we'll call it `ingredient adjectives`.

stephanegigandet:
@aleene : it looks like this branch has no actual changes anymore (as I guess all the changes were merged through another branch)

aleene:
I understand now. You accepted already a newer version. So I'll delete this one.

stephanegigandet:
For organic and fair trade, we handle them already, with the labels taxonomy, which is also used for ingredients parsing. I want to keep labels and processings separate.

aleene:
> For organic and fair trade, we handle them already, with the labels taxonomy, which is also used for ingredients parsing. I want to keep labels and processings separate.

Sounds as you mix labels and ingredients.

Issue #3306, opened 2020-04-23
Images uploaded through the product edit form are automatically selected (before cropping and selection)


stephanegigandet:
Images are selected as the new front / ingredients / nutrition images as soon as they are uploaded.

Images should not be automatically selected, especially if we have a previously selected / cropped image.

stephanegigandet:
It's a recent change in product_image_upload.pl :

````
			if ((($imagefield =~ /^front_/) or ($imagefield =~ /^ingredients_/) or ($imagefield =~ /^nutrition_/))
				# Changed 2020-03-05: overwrite already selected images
				# and ((not defined $product_ref->{images}{$imagefield}) or ($select_image))
				) {
				$log->debug("selecting image", { imgid => $imgid, imagefield => $imagefield}) if $log->is_debug();
				process_image_crop($product_id, $imagefield, $imgid, 0, undef, undef, -1, -1, -1, -1, "full");
			}

hangy:
Wasn't this done intentionally? #2975

stephanegigandet:
@hangy : it was done intentionnaly for images uploaded through apps (we assume they did the cropping if they do select "ingredients", "nutrition" etc. as the image type). This change disables it for images uploaded through the OFF web site.

stephanegigandet:
Deployed.

PR #3309, opened 2020-04-23
do not overwrite selected images when uploading images from the web

stephanegigandet:
bug #3306


stephanegigandet:
Also includes removing a function that is no longer used (done in a different way now that we have language tabs)

Issue #2826, opened 2020-01-03
'List of editors' page is loaded around 5 minutes


sashabrava:
Issue 1. The server has a page https://world.openfoodfacts.org/editors, which allows to get a list of users and amount of their contributions. However, the browser was waiting around 45s to get response from server and around 4.5 min to load the data (browser was consuming over 1GB RAM at that time).
![openfoodfacts](https://user-images.githubusercontent.com/11160215/71725925-d3d02780-2e35-11ea-9f34-ad7b3496cd9b.png)

Seems like this page makes a hard request on the database (I assume it calculates edits of 600.000 users in real time), and it gives a worry that such page might be used for an attack to slow down your server performance (what will happen if there are 100 such requests at the same time?).

Issue 2. I was scrolling the result, and there are users with weird nicknames - 
https://world.openfoodfacts.org/editor/yuka.SEpJYUliczlsL1lycTh3NXhCUEU1TlpFbVppTVpHanNPcnM3SUE9PQ
https://world.openfoodfacts.org/editor/yuka.UjRjeUhLazdvdFlnbk1JdTVSREY2b2w4bk03d2UyMkdkOVlqSVE9PQ
Their nickname in the database is shown as "yuka - user", and repeats multiple times. Seems like they are bots or auto-generated users (many of them have same amount of edits).

**Solutions**
If the page has only few views per day, it makes sense to calculate values once a day and show them 1000 per page with numeration (so that mobile devices won't go crazy).

hangy:
Good catch! This is probably the same underlying issue as #262. In this case, the list is 88 MiB of dynamically generated HTML, which first needs to be downloaded and then gets made filterable/sortable client-side by [DataTables](http://datatables.net/). The reason why `/editors` is now this long is because of the changes in #1975.

Issue #1630, opened 2019-01-31
Performance issue for pages that return a long list of tags


stephanegigandet:
Pages that return a long list of tags (e.g. /ingredients or /codes ) fail to load because they take too long (> 1 minute) to generate. (nginx typically times out before that)

The issue is not with mongo (especially as it as been optimised by bug #1612 ) but with the perl code that processes each of the tags.  (in sub display_list_of_tags($$) )

stephanegigandet:
One option is also to paginate results. But we might not be able to support search easily.
https://www.datatables.net/examples/data_sources/

hangy:
Related to #262. There's also the issue that really large pages (with several MB of HTML) may work in some browser but not others. Especially mobile devices with fewer RAM can struggle to display large sites efficiently (if at all).

stephanegigandet:
For codes, I'm going to reduce the number of entries, and keep only entries with at least 3 x : 

codes_tags: [
"code-13",
"9780201379xxx",
"978020137xxxx",
"97802013xxxxx",
"9780201xxxxxx",
"978020xxxxxxx",
"97802xxxxxxxx",
"9780xxxxxxxxx",
"978xxxxxxxxxx",
"97xxxxxxxxxxx",
"9xxxxxxxxxxxx"
],

In terms of actual uses (to group products by country / brands), it's probably enough. And it will reduce a lot the number of unique entries.

syl10100:
> 
> The issue is not with mongo (especially as it as been optimised by bug #1612 ) but with the perl code that processes each of the tags. (in sub display_list_of_tags($$) )

@stephanegigandet Are you sure about that ? When I analyzed the mongo log file, I saw "exceeded memory limit" for aggregation on ingredients and codes. The solution would be to turn on the option "allowDiskUse".

stephanegigandet:
@syl10100 : that's right, I turned on allowDiskUse, which allows the MongoDB query to return relatively quickly (less than 30 seconds), but then the Perl processing takes much longer.

stephanegigandet:
I created a new function get_taxonomy_tag_and_link_for_lang to replace multiple calls to display_taxonomy_tag / canonicalize_taxonomy_tag_link

It yields a 10X improvement to the speed to generate the HTML for the list of tags.

syl10100:
> 
> 
[...]

Issue #262, opened 2016-04-17
Paginate /ingredients /brands and /codes


teolemon:
The /ingredients /brands and /codes are booming in size, degrading performance on slower laptops.

Paginating them in 2 or 3 pages would solve the issue.

stephanegigandet:
If we paginate, we can't use the datatables filtering and sorting though. Unless there's a newer version with some sort of pagination /lazy loading support.

teolemon:
There's something horribly wrong going on the http://fr.openfoodfacts.org/codes page due to bogus entries

hangy:
DataTables does support [server-side processing](https://datatables.net/examples/server_side/) with several customization options and deferred loading.

Would SEO be an issue? Right now, the tables are completely indexed by search engines, and moving to a purely AJAX based table would most likely change that.

hangy:
I experimented with this a little bit on the `issue/262` branch. While server-side processing is possible, and we could return data in a correct format for DataTables, it would not be possible to use MongoDBs sorting and filtering as well as paging, because the product only has a `labels_tags` property that contains label ids like `en:gluten-free`, so that ordering by the French `sans gluten` would put it on a completely wrong page. Since the translations are only stored in `sto` files and available in a processed form in-memory in the Perl/CGI process, there's no easy way to switch on pagination and support the full taxonomy right now.

stephanegigandet:
Regarding SEO: I think it's fine if we don't have huge lists of tags. Those tags are also linked in product pages. We could also generate sitemap xml files with all the pages we want indexed.

Regarding the sorting / filtering: we could implement it in the Perl code, which would require a lot of processing on all entries, but those kind of queries would probably be relatively rare, as they would be from actual users. I'm guessing that currently, most of the page views for lists of tags are from search engine crawlers.

stephanegigandet:
@hangy : regarding the filtering part, we already have a part of it available.
e.g. https://fr.openfoodfacts.org/labels?filter=gluten --> Perl will process the mongodb results with the taxonomy, and it will return "Sans gluten".

hangy:
> @hangy : regarding the filtering part, we already have a part of it available.
> e.g. https://fr.openfoodfacts.org/labels?filter=gluten --> Perl will process the mongodb results with the taxonomy, and it will return "Sans gluten".

[...]

PR #3310, opened 2020-04-23
add pagination and limit for tags pages

syl10100:
Several changes :

- limit results size to 10000 lines for aggregation results like https://fr.openfoodfacts.org/ingredients
- add pagination if results size is over 10000 lines 
- add sub get_cache_results and set_cache_results for memcached usage
- add mongodb count_documents results in memcached

#262
#1630 
#2826


syl10100:
How it looks like :

OFF_world_ingredients
OFF_world_ingredients_2

stephanegigandet:
Wonderful! Thank you @syl10100 !
I deployed it on the dev server, but for some reason, I don't seem to get the pagination, only the limit for 10000 results.

image

stephanegigandet:
I do get the pagination if I log out.

stephanegigandet:
Overall it seems to be working very well. :)

stephanegigandet:
There's one case for which we want to disable the limit: when there is a ?filter= directive.

e.g. https://fr.openfoodfacts.org/ingredients?filter=oliv

This is used to filter the results based on their value in the local language (so not what's stored in the tags fields). So we have to get all the results, and the filtering is done by the Perl code.

syl10100:
@stephanegigandet  Ok I will check the issue with pagination when you are log in.
For the  ?filter= directive, you want to disable limit for any filter? Because I just tried https://fr.openfoodfacts.org/ingredients?filter=oliv and I get a 504 error (time out)...

stephanegigandet:
> Because I just tried https://fr.openfoodfacts.org/ingredients?filter=oliv and I get a 504 error (time out)...

We had some load issues today.. It's now working again (hopefully). The query takes about 10 second as it fetches all the products.

stephanegigandet:
> For the ?filter= directive, do you want to disable limit for any filter?

@syl10100 : Yes.

lib/ProductOpener/Display.pm
```diff
@@ -1304,7 +1312,42 @@ sub display_mission($)
...
+sub get_cache_results($$){
+
+	my $key = shift;
+	my $request_ref = shift;
+	my $results;
+
+	$log->debug("MongoDB hashed query key", { key => $key }) if $log->is_debug();
+
+	# disable caching if ?nocache=1
+	# or if the user is logged in and nocache is different from 0
```
syl10100, line 1324: @stephanegigandet is it still relevant to disable cache when user is logged in ?

syl10100:
Correction done for ?filter and pagination for user logged in

stephanegigandet:
Looks good to me. I'm going to make a few small changes. I think we can avoid having a separate count query when there are less than 10k results (which is probably the large majority of the case): we have the results, Perl can easily count them.

PR #3311, opened 2020-04-23
add owner field for mongo queries, needed for producers platform

stephanegigandet:
all thumbnails are currently broken on the producers platform..


stephanegigandet:
(deployed on production producer platform)

PR #3316, opened 2020-04-24
New dark scheme 

alifafaruk:
Dark mode support ISSUE #2407

@lucasortizny @jen-lopez and I worked on the dark mode feature for the website. We implemented a toggle button, where the user can turn dark mode on/off. 

Screen Shot 2020-04-24 at 12 35 46 PM
Screen Shot 2020-04-24 at 12 58 42 PM

PR #3317, opened 2020-04-24
feat: New Dark Scheme Part 2

### LINKED ISSUE #2407: Dark mode support
https://developer.mozilla.org/fr/docs/Web/CSS/@media/prefers-color-scheme

alifafaruk:
Dark mode support 
fixes #2407

@lucasortizny @Jen-Lopez and I worked on the dark mode feature for the website. We implemented a toggle button, where the user can turn dark mode on/off. We did a [pull request](https://github.com/openfoodfacts/openfoodfacts-server/pull/3316) earlier, but it had multiple merge conflicts that we weren't able to fix. So we closed that. We fixed the issues on our end and now it works!
 
Screen Shot 2020-04-24 at 5 00 40 PM

Screen Shot 2020-04-24 at 12 58 42 PM


stephanegigandet:
Hi @alifafaruk ! Thank you for the PR!

The dark mode looks very good, but instead of adding a site specific toggle, I think it would make much more sense to just use the CSS prefers-color-scheme, as @teolemon suggested in the bug #2407 

That way we can avoid the 2 javascript scripts that will block the page rendering for all users until they load, and we can have a pure CSS solution. 

Also it would be great to put the CSS styles inside the SCSS ( in the /scss/ directory) so that they get processed along with the rest of the CSS styles.

cc @hangy

teolemon:
Here's the Open Food Facts logo in dark mode: https://github.com/openfoodfacts/openfoodfacts-server/blob/3d1a30516a663d1dedad0733fac66d9d93a249c8/html/images/misc/openfoodfacts-logo-no-tagline-dark-mode.svg

html/js/preDarkMode.js
```diff
@@ -0,0 +1,10 @@
+
+// Before page renders, check if page is in darkmode.
+// Otherwise, it will load regular styles before dark style.
+let dark = localStorage.getItem('darkMode');
+var sheet = document.styleSheets[3];
```
GendelfLugansk, line 5: The same issue as in `darkMode.js`

GendelfLugansk:
It seems like `prefers-color-scheme` requires user to activate dark theme on OS level. Correct me if I'm wrong. In this case I'm not sure if it's better

cc @stephanegigandet @teolemon @hangy

GendelfLugansk:
Red text on blue buttons is hard to read, imo. At least on my screen

screenshot

White text like this is probably a bit too bright. It can bee seen in right sidebar and on some pages with text (like Contribute and Discover)

screenshot

That black text (if user is logged in) is barely visible

screenshots

This block on the right on [category page](http://world.productopener.localhost/category/dairies)

screenshot

Some text on [product page](http://world.productopener.localhost/product/3960051050102/beurre-du-haut-jura-les-moussieres) is black and unreadable

screenshot

Nutrition table on the same product page is light. Not critical but could be darker, imo

screenshot

In footer, black text

screenshot

Many regions of product edit page are too light-colored, imo

screenshot
screenshot

On Graphs and Maps page, graph has white background

screenshot

GendelfLugansk:
Also, it seems like it [might be possible](https://levelup.gitconnected.com/share-localstorage-sessionstorage-between-different-domains-eb07581e9384) to share "dark mode" setting across different domains but this might be complex to implement. I think that would require changes to perl code, docker config and nginx/apache config on production site. I don't have perl expertise to give more detailed advice.

hangy:
> It seems like `prefers-color-scheme` requires user to activate dark theme on OS level. Correct me if I'm wrong. In this case I'm not sure if it's better

It sounds like a good idea to default to the dark mode, if someone has selected it, though. [This guide](https://web.dev/prefers-color-scheme/#loading-strategy) explains that the media query can also be done from JS.

> Also, it seems like it [might be possible](https://levelup.gitconnected.com/share-localstorage-sessionstorage-between-different-domains-eb07581e9384) to share "dark mode" setting across different domains but this might be complex to implement. I think that would require changes to perl code, docker config and nginx/apache config on production site.

Changing some JavaScript should suffice, if I understand the blog post correctly. However, I wonder if we need `localStorage` for this. To no waste too much cookie storage, one could just set a `d=1` cookie to store enabled/changed dark mode setting in a cross-subdomain cookie.

html/js/darkMode.js
```diff
@@ -0,0 +1,30 @@
+var stylesheet = document.styleSheets[3]; // dark-mode stylesheet
+
+let darkMode = localStorage.getItem('darkMode');
+const darkToggle = document.getElementById("switch");
+
+const enableDarkMode = () => {
```
hangy, line 6: Do you have an idea about browser compatibility? We don't run this JS through a transpiler, so using `const` and lambda expressions could limit dark mode to some modern browsers.

lib/ProductOpener/Display.pm
```diff
@@ -6090,7 +6090,13 @@ $options{favicons}
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/select2/4.0.10/css/select2.min.css" integrity="sha256-FdatTf20PQr/rWg+cAKfl6j4/IY3oohFAJ7gVC3M34E=" crossorigin="anonymous">
 <link rel="search" href="$formatted_subdomain/cgi/opensearch.pl" type="application/opensearchdescription+xml" title="$Lang{site_name}{$lang}">
 $header
+<!-- Dark Mode -->
+<link rel="stylesheet" href="$static_subdomain/css/darkMode.css">
```
hangy, line 6094: This basically makes dark mode the default - Might introduce unwanted behaviour if JavaScript is disabled. cc @stephanegigandet

html/js/darkMode.js
GendelfLugansk: [Const](https://caniuse.com/#feat=const) should be fine. [Arrow functions](https://caniuse.com/#feat=arrow-functions) are not compatible with IE and Opera Mini. I guess that's ok

stephanegigandet:
> > It seems like `prefers-color-scheme` requires user to activate dark theme on OS level. Correct me if I'm wrong. In this case I'm not sure if it's better

I don't know either. Is there some research on dark mode users? Do they turn it on or off depending on ambient light? Does it make sense to turn it on and off depending on the web sites users visit?

I was under the impression that it was more of a "global" preference that is valid accross sites and times: you either want dark mode all the time, or you don't. So you set it once in the OS preference, and then hopefully web sites have styles that match your preference.

lib/ProductOpener/Display.pm
stephanegigandet: Dark mode should not be the default.

GendelfLugansk:
> I was under the impression that it was more of a "global" preference that is valid accross sites and times: you either want dark mode all the time, or you don't. So you set it once in the OS preference, and then hopefully web sites have styles that match your preference.

The thing is, when I google "Chrome dark mode", I see [this](https://support.google.com/chrome/answer/9275525?co=GENIE.Platform%3DDesktop&hl=en&oco=0) article which implies that dark mode is supported only in Win 10+ and MacOS 10.14+. As Ubuntu 18.04 user, I don't think I can enable it. So for various Linux users toggle might be useful. Toggle can be combined with `prefers-color-scheme`, as was [suggested above](https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#issuecomment-623132902)

lib/ProductOpener/Display.pm
GendelfLugansk: I didn't test in all browsers but in Chromium adding `disabled="true"` to `link` disables stylesheet. I think this should work in other browsers as well. JS code of this PR is based on `disabled` property and [MDN says](https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet/disabled) it's supported everywhere except IE (unknown)

VaiTon:
> The thing is, when I google "Chrome dark mode", I see [this](https://support.google.com/chrome/answer/9275525?co=GENIE.Platform%3DDesktop&hl=en&oco=0) article which implies that dark mode is supported only in Win 10+ and MacOS 10.14+. As Ubuntu 18.04 user, I don't think I can enable it. So for various Linux users toggle might be useful. Toggle can be combined with `prefers-color-scheme`, as was [suggested above](https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#issuecomment-623132902)

I agree. Other sites (eg. stackoverflow, youtube, etc.) are using a site-specific switch and eventually we could add the "system settings" one.

lib/ProductOpener/Display.pm
VaiTon: > 
> 
> I didn't test in all browsers but in Chromium adding `disabled="true"` to `link` disables stylesheet. I think this should work in other browsers as well. JS code of this PR is based on `disabled` property and [MDN says](https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet/disabled) it's supported everywhere except IE (support is unknown)

Right, but there are some extensions that completely disable Javascript execution. In that case dark mode would be the default. Consider adding `disabled="true"` here and changing it in JS.

alifafaruk:
Hi all! Thank you for taking the time to look into the pull request. My group (@Jen-Lopez and @lucasortizny) and I are new to web development and are unfamiliar with the mechanics of how to implement a prefers-color-scheme. We will be fixing the CSS to fix the bugs.

stephanegigandet:
Hi Everyone, just wondering if there is still some work being done / to do on this PR?

teolemon:
poke @alifafaruk @VaiTon

VaiTon:
> 
> 
> poke @alifafaruk @VaiTon

Cannot push to branch :)

alifafaruk:
Hi, we fixed all the style changes requested for dark mode. We also fixed bugs that were present. However, we don't know how to implement "prefers-color-scheme," so the method used to implement dark mode is still a toggle. If someone could help us change it to a "prefers-color-scheme" method, that would be great!

GendelfLugansk:
This is still an issue:
screenshot
"Android apk" is still unreadable:
screenshot
Not sure about this wide border around each tag (if others are ok with it, it's fine with me):
screenshot
This is still an issue (go to "Charts and maps", choose "Energy" for horizontal axis and click "generate graph"):
screenshot

lib/ProductOpener/Display.pm
GendelfLugansk, line 6094: Add `disabled="true"` to `link` tag

GendelfLugansk:
Please make these requested changes:

https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#pullrequestreview-404621626
https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#pullrequestreview-404634024
https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#pullrequestreview-422133026

Also, please look at these screenshots: https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#issuecomment-637064818

teolemon:
I created an APK logo in line with the others: 
https://github.com/openfoodfacts/openfoodfacts-server/pull/3535/files?short_path=67a224e#diff-67a224e98ae5f4a0802aeac7113905f3

Jen-Lopez:
Hi! Thank you all for the suggestions. I tried making the JS work but I also agree that some browsers may have javascript disabled and by default, dark mode would be enabled (which is something that shouldn't happen). I tried adding the disabled = "true" ([request](https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#pullrequestreview-422133026)) to the dark mode stylesheet but it didn't seem to work on my end (dark mode stopped working completely). I looked into the prefers-color-scheme and it seems that most apps and websites are using something similar to implement a dark mode feature nowadays. Although the prefers-color-scheme is not compatible with some browsers and is based on system preferences, I still think it's a good and efficient way for dark mode implementation without disrupting the original stylesheet (and with no glitches). @alifafaruk and I are still in the process of making [changes](https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#issuecomment-637064818) but the recent commit implements the [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) suggested previously, with JS removed.

teolemon:
Good news, the dark mode compatible APK logo is merged, and available in prod

Jen-Lopez:
I fixed the "Edit a Product" Page and went throughout the site to fix more css bugs. Since the new Android image is available in production and is dark mode aware, this issue should be resolved. However, I was unable to fix the floating WebLinks block in "dairies" since there is no id/class associated with it to change the styling. What page should I look at to add an ID to the element? Also, if there are any changes to be made or any suggestions, please let me know.

teolemon:
This PR has merge conflicts.

CloCkWeRX:
TODO items

- [ ] Android APK - Anyone able to do a revised version of https://static.openfoodfacts.org/images/misc/android-apk.svg ? 
- [ ] Weblinks - is this just https://github.com/openfoodfacts/openfoodfacts-server/blob/ec77a782631701af3db2aa799789214b7e5a2cce/lib/ProductOpener/Display.pm#L3255 with inline styles that should be pulled up to a main stylesheet?
- [ ] Tabs - ```ul.tabs {
    border-top: 1px solid #ccc;
    background-color: #efefef;
}```` ?

CloCkWeRX:
Did a dark friendly (and light friendly) revamp of the logo
android-apk

teolemon:
We just deployed a new design, but without dark mode so far.

stephanegigandet:
We deployed a redesign of the https://world.openfoodfacts.org website, so sadly we won't be able to deploy this PR.
Sorry about that and thank you for your work @alifafaruk @lucasortizny and @Jen-Lopez .

PR #3327, opened 2020-04-26
Update linking to hunger-games

VaiTon:
Removed `/hunger-game` page;
Added redirect to `hunger.openfoodfacts.org` to nginx config.

To not break the userscript we need to wait for it to update to the new URL.

### DIFF
--- a/conf/nginx/sites-available/off
+++ b/conf/nginx/sites-available/off
@@ -244,6 +244,9 @@ server {
 	location = /haz-su-donacion-a-open-food-facts {
 			try_files /donate/es.html =404;
 	}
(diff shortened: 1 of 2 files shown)


stephanegigandet:
I'm not sure if we want to completely replace the https://world.openfoodfacts.org/hunger-game page with https://hunger.openfoodfacts.org    Do we @raphael0202 ?

raphael0202:
I'm in favor of replacing it, hunger game v1 lags compared to v2 feature wise.

teolemon:
let's go @raphael0202 @stephanegigandet

PR #3337, opened 2020-04-28
tags pagination: do not create a 2nd count mongod query if less than 10k results

stephanegigandet:
This is a continuation of @syl10100 's PR : https://github.com/openfoodfacts/openfoodfacts-server/pull/3310

When  we have less than 10000 tags (which for almost all tags type is always the case, except for ingredients, editors, possibly categories), we can avoid making a separate count query: we know the number of results from the query that fetches them.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -1460,7 +1460,7 @@ sub query_list_of_tags($$) {
 
 	if ((not defined $results) or (ref($results) ne "ARRAY") or (not defined $results->[0])) {
 
-		# do not used the smaller cached products_ tags collection if ?nocache=1
+		# do not used the smaller cached products_tags collection if ?nocache=1
 		# or if the user is logged in and nocache is different from 0
 		if ( ((defined $request_ref->{nocache}) and ($request_ref->{nocache}))
 			or ((defined $User_id) and not ((defined $request_ref->{nocache}) and ($request_ref->{nocache} == 0)))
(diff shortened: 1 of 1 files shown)


stephanegigandet:
Currently on dev server if you want to test.

lib/ProductOpener/Display.pm
```diff
@@ -1532,10 +1504,69 @@ sub query_list_of_tags($$) {
...
+
+			# do not used the smaller cached products_ tags collection if ?nocache=1
+			# or if the user is logged in and nocache is different from 0
+			if ( ((defined $request_ref->{nocache}) and ($request_ref->{nocache}))
+				or ((defined $User_id) and not ((defined $request_ref->{nocache}) and ($request_ref->{nocache} == 0)))
+				) {
+				eval {
+					$log->debug("Executing MongoDB aggregate count query on products_tags collection", { query => $aggregate_count_parameters }) if $log->is_debug();
+					$count_results = execute_query(sub {
+						return get_products_tags_collection()->aggregate( $aggregate_count_parameters, { allowDiskUse => 1 } );
```
syl10100, line 1544: I think it is the opposite : if nocache=1 then products collection is used else it should be products_tags.

stephanegigandet: Good catch, I'll fix it.

stephanegigandet: Fixed.

syl10100:
image

also it seems that pagination is not displayed any more

stephanegigandet:
> also it seems that pagination is not displayed any more

I moved it to the bottom of the page. I don't think it's going to be used much, unless you scroll to the bottom and realise you want more results.

syl10100:
> 
> 
> > also it seems that pagination is not displayed any more
> 
> I moved it to the bottom of the page. I don't think it's going to be used much, unless you scroll to the bottom and realise you want more results.

Ok... But in this case, the number "10000 catégories" displayed is wrong.

stephanegigandet:
> Ok... But in this case, the number "10000 catégories" displayed is wrong.

This happens only when the count query fail (and the other did not). The dev server is quite small and MongoDB queries tend to time out there.

Here is how it looks like for me now:

image

Issue #3350, opened 2020-05-01
Do not retry failed MongoDB queries


stephanegigandet:
We currently try failed MongoDB queries (e.g. timeouts) 3 times. If the MongoDB server is overloaded, it probably makes things worse. If we have only one Mongo server to send the query to, I think it's best not to retry queries.

stephanegigandet:
Deployed.

PR #3351, opened 2020-05-01
set max_retries_number to 0 for MongoDB queries, bug #3350

stephanegigandet:
At the start of the project, I was retrying MongoDB queries once, and @hangy made a better system with Action::Retry. But when the MongoDB server is overloaded (as was the case a week ago), retrying queries make things worse, so I think we should turn off retries, unless at some point we put more than one MongoDB server.


syl10100:
Maybe we should increase the timeout before it fails ? It think it is 30s by default.

PR #3357, opened 2020-05-02
ingredient parsing: normalize Finnish vitamins correctly

Ban3:
**Description:**

Fixing a silly mistake I have made earlier of postfixing the vitamin code instead of prefixing it, like it should have been done.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -2068,7 +2068,7 @@ sub normalize_vitamin($$) {
 	if ($a =~ /^[a-z][a-z]?-? ?\d?\d?$/i) {
 		($lc eq 'es') and return "vitamina $a";
 		($lc eq 'fr') and return "vitamine $a";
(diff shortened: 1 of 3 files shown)


Ban3:
> It looks good, thank you. Could you also add a test in ingredients_tags.t so that we can check that we now recognize correctly the individual vitamins?

Of course. Is it expected that the additives id takes preference over the vitamin, or is this an error in the taxonomy definitions? For `Vitamiinit (B2, B12, C, D2)` I get `en:e101` for vitamin B2 and `en:e300` for vitamin C.

stephanegigandet:
@Ban3 : yes, it's expected. We automatically include all additives in the ingredients taxonomy, and they take precedence. The following taxonomies are included in the ingredients taxonomy:

("additives_classes", "additives", "minerals", "vitamins", "nucleotides", "other_nutritional_substances", "ingredients")

Ban3:
I pushed a single test for the vitamin tags.

PR #3368, opened 2020-05-04
Use spawn to check perl scripts

hangy:
This avoids ERR_CHILD_PROCESS_STDIO_MAXBUFFER being called if we have too much output.

### DIFF
--- a/scripts/check_perl.js
+++ b/scripts/check_perl.js
@@ -6,26 +6,33 @@ const util = require("util");
 const glob = util.promisify(require("glob").glob);
 const { spawn } = require("child_process");
 
+function checkFile(path, doneCallback) {
+  console.log(`Checking ${path}`);
+  try {
+    const child = spawn(`perl`, ["-c", "-CS", "-Ilib", path]);
+    child.stdout.on("data", (data) => console.log("[" + path + "] " + data));
+    child.stderr.on("data", (data) => console.error("[" + path + "] " + data));
+    child.on("close", (code) => {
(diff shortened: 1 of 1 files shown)


hangy:
OK, so the main reason for #3155 failing with a timeout is likely to be found in here. 🙂

stephanegigandet:
Mmm.. the test has been running for 2 hours already.

hangy:
> Mmm.. the test has been running for 2 hours already.

I suspect some background process gets killed, which causes the build to not finish in the alloted amount of time. The Travis build runs the same script and shows some interesting log messages. Different issues taken out of context:

```
[scripts/import_csv_file.pl] Out of memory!
[scripts/update_all_products_from_dir_in_mongodb.pl] result: null
[scripts/import_csv_file.pl] result: 1
```

Obviously, the script and build should not exit with a success status if a single file causes an error and return `1`. The `null` return value is also interesting, as I wouldn't know how a spawned process could return without a status.. I suspect I made a mistake when checking the return value. Additionally, the `out of memory´ error is weird, but might be caused by too many subprocesses running at the same time. This is likely the cause why the GitHub Workflow doesn't manage to finish the build. 🙂

hangy:
I think I got it. In the current version of the script, the perl processes actually get started in parallel, which obviously is too much for the build pipelines. The only reason why the Travis CI build works is that it handles this particular problem a bit differently than GitHub Workflows. Let's see what I can do to fix it. 😁

PR #3378, opened 2020-05-05
Add Health Star Rating System from Australia

CharlesNepote:
**Description:**

Health Star Rating System claims to be used by more 10,000 products in Australia.
At least we could add these labels.

http://healthstarrating.gov.au/

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -18239,10 +18239,18 @@ en:Health Star Rating, Health Star Rating System, HSR
 
 en:Health Star Rating 1
 
+en:Health Star Rating 1.5
+
 en:Health Star Rating 2
+
+en:Health Star Rating 2.5
 
 en:Health Star Rating 3
 
+en:Health Star Rating 3.5
+
 en:Health Star Rating 4
+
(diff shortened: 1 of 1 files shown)


teolemon:
It seems that you can have 3,5 @CharlesNepote @stephanegigandet

aleene:
Is there a way we could calculate them?

VaiTon:
Also, why are we adding the general label and then the specific ones? What's the purpose of the general label?

teolemon:
http://www.healthstarrating.gov.au/internet/healthstarrating/publishing.nsf/Content/excel-calculator

CharlesNepote:
> Also, why are we adding the general label and then the specific ones? What's the purpose of the general label?

@VaiTon it's important for some people or projects (e.g. public policies) to know the total number of products that have this label on the front of pack. We do the same thing with Nutri-Score:
* https://world.openfoodfacts.org/label/nutriscore
* https://world.openfoodfacts.org/label/nutriscore-grade-a
* [...]

Maybe we could compute the general label based on the specific ones.

PR #3406, opened 2020-05-10
more ES and IT entries

stephanegigandet:
Added a few common ingredients.


taxonomies/ingredients.result.txt
```diff
@@ -218,6 +218,10 @@ nl:gerstemeel
 pt:farinha de cevada
 ro:făină de orz
 
+< :
+en:flat green beans
```
teolemon, line 222: typo

PR #3416, opened 2020-05-12
put back wikidata fixes

stephanegigandet:
There were some regressions (probably due to a merge, I'm not sure how it happened), I'm putting back fixes that were deleted by error.

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -564,7 +564,7 @@ sk:Emulgujúce (taviace) soli
 sl:Emulgirne soli
 sv:Smältsalter
 wikidata:en:Q50419202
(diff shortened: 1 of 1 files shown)


stephanegigandet:
> Shouldn't the Wikipedia and wikidata links all be `en`?

wikidata yes, wikipedia, we could link to different language specific versions

hangy:
> wikidata yes, wikipedia, we could link to different language specific versions

If I understand wikidata correctly, all Wikipedia page language versions should be linked to the same wikidata entity anyways. @teolemon?

stephanegigandet:
Right, we could decide we don't need the wikipedia properties, or that we could update them automatically from wikidata.

This PR is just to repair wikidata links and put the right language for the few wikipedia links we have. It wasn't my aim to restructure how we store the wikidata / wikipedia properties. :)

PR #3418, opened 2020-05-13
Add a new process: pressed

aleene:
**Description:**
process pressed and variants

### DIFF
--- a/taxonomies/ingredients_processing.txt
+++ b/taxonomies/ingredients_processing.txt
@@ -382,7 +382,6 @@ it:spremuto di
 sv:presset
 
 #<en:pressed
-en:pressed with alkali
 
 #<en:pressed
 en:fresh pressed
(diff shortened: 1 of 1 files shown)


taxonomies/ingredients_processing.txt
```diff
@@ -372,6 +372,45 @@ fr:extrait naturel de, extrait naturel d'
...
+# en:false:compressed
+de:abgepresste
+es:espremido de
+fr:pressé, pressée, pressées
+#fr:false:à pâte pressée, fromage pressé
+it:spremuto di
+sv:presset
+
+#<en:pressed
+en:pressed with alkali
```
stephanegigandet, line 385: that's weird, shouldn't it be "processed with alkali"

aleene: I don't know, thats what I saw in the ingredients.

stephanegigandet: probably a mistake in the ingredients, let's not add it (unless we do find several products that contain it on the ingredients photo)

teolemon, line 385:
```suggestion
```

Issue #3420, opened 2020-05-13
Mobile Layout for Product List (css fix)


Jen-Lopez:
### What
When viewing the site on mobile, there is some extra white space on the right hand-side. The pagination could also start on its own line. A few lines of CSS would fix this and make the mobile UI look great.
<img width="401" alt="product_before" src="https://user-images.githubusercontent.com/21044058/81865201-97c37680-953b-11ea-84dc-bae54a6cb53f.png">
<img width="405" alt="paging_before" src="https://user-images.githubusercontent.com/21044058/81865219-9bef9400-953b-11ea-8b17-af1023e0df14.png">

### Part of
- #5513

teolemon:
The website has been redesigned. 
This might have fixed your issue and opened others.
 Please open new issues in #7295

PR #3421, opened 2020-05-13
Mobile Fix for Product List layout

Jen-Lopez:
I edited the _off.scss file by adding a media query that responds to a mobile device. The styles center the products, dropdown, and pagination. It also gets rid of the white-space as mentioned in the issue. 

**Related issues and discussion:** #3420
product_after
paging_fixed

### DIFF
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -386,3 +386,10 @@ a[aria-controls="idOfLeftMenu"] {
   }
 
 }
+
+@media only screen and (max-width: 28.125em) {
+  .products li, ul.pagination li {float: none; display: inline-block;}
+  ul.products, #pages, ul.button-group  {text-align: center;}
+  ul.pagination li:first-child {display: block;}
+  ul.button-group li {margin-top: 3px;margin-bottom: 3px;}
+}
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -384,10 +384,9 @@ a[aria-controls="idOfLeftMenu"] {
   .icon {
     font-size: 1.8rem;
   }
-
 }
 
-@media only screen and (max-width: 28.125em) {
+@media #{$small-only} {
   .products li, ul.pagination li {float: none; display: inline-block;}
   ul.products, #pages, ul.button-group  {text-align: center;}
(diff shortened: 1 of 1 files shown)


scss/_off.scss
```diff
@@ -386,3 +386,11 @@ a[aria-controls="idOfLeftMenu"] {
   }
 
 }
+
+@media only screen and (max-width: 28.125em) {
```
VaiTon, line 390: Thank you for the PR!

`max-width: 28.125em`

Where did you take this value from? Are we sure that it's good for every mobile device?

Otherwise all good!

GendelfLugansk: Here is what I found on this subject: [https://css-tricks.com/em-based-media-queries-are-based-on/](https://css-tricks.com/em-based-media-queries-are-based-on/). I think using `450px` here would be better: it is more readable and as I understood, browser settings may affect media query with `em`

GendelfLugansk: Examples: 

Font set to "small" in chromium settings: 

screenshot

Font set to "medium" (recommended and default value):

screenshot

hangy: Maybe the breakpoint variables from Foundation could be used?

VaiTon: > 
> 
> Maybe the breakpoint variables from Foundation could be used?

We're still on Foundation 5, so we should do as [the documentation](https://get.foundation/sites/docs/media-queries.html#upgrading-from-foundation-5) says.

Eg.
```sass
@media #{$medium-only} {
}
```

Jen-Lopez: Hi! I just assumed that em was used based on the other media queries in the file. I converted 450px to em, which was 28.125 em. And I'm not sure if all devices are under 450px width (majority of them are). But I noticed that at that width, the content starts to shift to one side.

above_450
below_450

stephanegigandet: Let's use $small-only

VaiTon:
Cannot push changes to branch. @Jen-Lopez can you update?

Jen-Lopez:
Hi! Just updated the media query

Issue #3441, opened 2020-05-18
Cannot start vscode docker remote development


VaiTon:
**Describe the bug**
When trying to start remote development in vscode errors are thrown and vscode errors out with "An error occurred setting up the container".
```
ERROR: for backend  Cannot start service backend: OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \
"rootfs_linux.go:58: mounting \\\"/var/lib/docker/volumes/docker_html_data/_data\\\" to rootfs \\\"/var/lib/docker/overlay2/ef1209c4bbaf66cb6fe639cce89b54d84967e145da305c8a215c0c239
3785f17/merged\\\" at \\\"/var/lib/docker/overlay2/ef1209c4bbaf66cb6fe639cce89b54d84967e145da305c8a215c0c2393785f17/merged/opt/product-opener/html/data\\\" caused \\\"mkdir /var/lib
/docker/overlay2/ef1209c4bbaf66cb6fe639cce89b54d84967e145da305c8a215c0c2393785f17/merged/opt/product-opener/html: file exists\\\"\"": unknown
ERROR: Encountered errors while bringing up the project.
[890713 ms] Command failed: docker-compose --project-name docker -f c:\Users\*\openfoodfacts-server\docker\docker-compose.yml -f c:\Users\*\openfoodfacts-server\docker\docker-compose.dev.yml -f c:\Users\*\web\openfoodfacts-server\docker\docker-compose.admin-uis.yml -f c:\Users\*\openfoodfacts-server\docker\docker-compose.vscode.yml up -d --build
```

**To Reproduce**
1. Remove every docker container and image
2. Open vscode
3. Try to start remote container

**Desktop (please complete the following information):**
 - OS: Win 10

VaiTon:
cc @hangy

hangy:
Does this work if you use the docker-compose command manually? I haven't seen that exact message before.

VaiTon:
> 
> 
> Does this work if you use the docker-compose command manually? I haven't seen that exact message before.

Yup, same message.

hangy:
I was unable to reproduce this with a clean clone of the repo and a freshly pruned Docker Desktop environment. 😕

VaiTon:
I deleted and recloned the repo and apparently the error is no more. Closing :)

PR #3443, opened 2020-05-19
use $options{product_type} instead of "food" in $server_domain

stephanegigandet:
This should solve the issue of the docker environment not computing the Nutri-Score.


VaiTon:
Testing to see if this fixes #3441

PR #3459, opened 2020-05-22
translation for some nutriments

teolemon:
Add some entries from Food.pm to speed up their translations in a normalized way


stephanegigandet:
I don't think it's a good idea to add translations that we won't be using (at least with the current code), it will make things confusing.

The right solution will be - at some point - to replace the strings in Food.pm by a taxonomy for the nutrients (but it is complex, has side effects etc. That's why the nutrients taxonomy that was created some time ago is not used in Product Opener.)

teolemon:
I'm trying a pragmatic and limited approach to something that won't be fixed anytime soon otherwise.

stephanegigandet:
I know. But sometimes it makes more harm than good to add partial solutions to problems. It creates debt. Just like the nutrients taxonomy: I don't know what to make of all the open PRs on it.

PR #3464, opened 2020-05-22
Web version of the Advanced User Guide

teolemon:
Web version of the Advanced User Guide by @bcatelin
You can see a preview at https://world.openfoodfacts.net/advanced-user-guide


VaiTon:
Why not on the wiki?

teolemon:
It was a Word document that I converted to HTML and cleaned. We can convert it to Wiki syntax, and upload manually each image.
Pro: editable
Con: Will it have as much visibility as on the main website ? More conversion and manual upload work

lang/en/texts/advanced-user-guide.html
```diff
@@ -0,0 +1,1015 @@
...
+  .image {
+    margin-left: 0px;
+    margin-top: 0px;
+    transform: rotate(0rad) translateZ(0px);
+    -webkit-transform: rotate(0rad) translateZ(0px);
+  }
+</style>
+
+<img
+  alt="https://static.openfoodfacts.org/images/misc/openfoodfacts-logo-en-356x300.png"
```
hangy, line 14: This will be available on OBF and OPFF, too. Do we want to hard-code the domain?

teolemon: Just deploying it manually on OFF. I think we need specific guides for OBF and OPFF.

PR #3480, opened 2020-05-26
Ingredient and process additions

aleene:


taxonomies/ingredients.txt
```diff
@@ -9874,11 +9874,13 @@ sv:Färsk grädde
...
 ca:Crema agra, crema àcida
+da:syrnet fløde
 de:saure Sahne, Sauerrahm
+el:κρέμα γαλακτοσ
 eo:Acidkremo
-es:Crema agria, nata agria
+es:Crema agria, nata agria, nata ácida
 fa:خامه ترش
-fi:Hapankerma
+fi:Hapankerma, ranskankerma
```
Ban3, line 9883: Ranskankerma = Crème fraîche. I think it's already a synonym for that.

aleene: Great, I just copied it from ingredient lists. So you see how well producers translate.

stephanegigandet: So ", ranskankerma" should be removed, right?

Ban3: @stephanegigandet Actually, I noticed that ranskankerma was under a wrong ingredient (`en:fresh-cream`) even before aleene's commit. I'll fix it by pushing to this branch.

stephanegigandet: Perfect, thank you @Ban3 !

Issue #3483, opened 2020-05-27
Remove non-significant digits from nutrition values passed to the API


stephanegigandet:
Yuka seems now to pass floating point values with non significant digits.

e.g. https://fr.openfoodfacts.org/api/v0/produit/2227153013512/terrine-de-campagne

carbohydrates: 2.9000000953674,
sugars: 1.6000000238419,

https://fr.openfoodfacts.org/api/v0/produit/8001420001624/riso-arborio-gallo-1-kg

saturated-fat: 0.89999997615814,
sodium_100g: 0.359999990463256,
sugars_value: 2.5999999046326,

Those values are bogus, due to how they store numbers internally. We should remove non significant digits. (and make sure 0.359999990463256 becomes 0.36)

We also need to fix existing products with that kind of values.

hangy:
> Those values are bogus, due to how they store numbers internally. We should remove non significant digits. (and make sure 0.359999990463256 becomes 0.36)

But that's just one app that's sending decimals as floats with low precision (or something like that). Other apps might use slower but more precise data types for these values. Also, the website allows entering arbitrary strings, and that kind of truncation might be unexpected.

stephanegigandet:
> Other apps might use slower but more precise data types for these values. Also, the website allows entering arbitrary strings, and that kind of truncation might be unexpected.

Any value with 0000 or 9999 in the decimals in it is not an actual nutrient value that comes from a product label, there is never such a precision in values.

VaiTon:
Has this been fixed by #3489?
@stephanegigandet

stephanegigandet:
@VaiTon : not yet, I made a function to remove the non-significant digits, but it has not been plugged to the API, and we have not run it against existing products.

Issue #3487, opened 2020-05-27
Significant trailing zeroes for nutrients are not kept


stephanegigandet:
While investigating bug #3483 , I realized that we are not keeping trailing 0s in nutrients sent by users or the API. e.g. is a nutrient value is 2.50, we convert the value to a number and it becomes 2.5.

We could try to preserve trailing 0s, but it may add a lot of complexity

aleene:
I would prefer to keep the trailing zero's, as it is an indication of accuracy.

stephanegigandet:
Yes, I would prefer too. It makes things quite complex though, as the values can be sent with one unit, but then converted / displayed in another. Or maybe we just store a new field carbohydrates_zeroes=2 if we are sent 2.00 g or 2000.00 mg.

aleene:
In that way it would be dependent on the unit that has been used. Maybe it is better to define a field with the number of (significant?) digits, independent on the location of the comma. So 2.00 g would be 3, 2000.00 mg would be 6.

VaiTon:
What about just having two decimal positions? 
3.956 -> 3.96
1.444 -> 1.45
2 -> 2.00
1.2 -> 1.20
130 -> 130.00

teolemon:
@stephanegigandet has it been deployed ?

teolemon:
Hasn't the nutrition refactor fixed this ?

teolemon:
+2020 PR

Freso:
FWIW, if significant trailing zeroes are stored/kept now, they’re not being shown to the user when editing… but that’s probably a separate issue to this one.

PR #3489, opened 2020-05-27
Keep significant digits, remove insignificant digits for nutrition facts

stephanegigandet:
This is a function to solve bug #3483

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -106,6 +106,8 @@ BEGIN
 					&create_nutrients_level_taxonomy
 
 					&assign_category_properties_to_product
+					
+					&remove_insignificant_digits
 
 					);	# symbols to export on request
 	%EXPORT_TAGS = (all => [@EXPORT_OK]);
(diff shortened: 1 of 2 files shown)


stephanegigandet:
> * Is the decimal point expected to be normalised to `.` before the method is called?

Currently we have this in the code when we assign a nutrient value (that comes from the web site, a producer, an app):

````
sub assign_nid_modifier_value_and_unit($$$$$) {

[..]
	$value =~ s/(\d) (\d)/$1$2/g;
	$value =~ s/,/./;
````

I'm planning to call the remove_insignificant_digits function after.

> * We store nutrients and other numbers as int/float right now, so that changing this to string (in mongodb) would be a massive breaking change for API consumers, and might break the search functionality of the off website.

I don't plan to change the existing values to strings, we'll convert them back to numbers when we store them to the existing fields. We might keep a new field to store the number of significant digits or trailing 0s, so that we don't lose information when we have a value like "2.40".

> * I don't understand why you want to remove precision from 2.5999999046326 but not 2.00.

That's because 2.5999999046326 is not precision, but always an artefact resulting of storing float numbers somewhere along the chain (an app, a database of a producer etc.). At the maximum there is 4 or 5 significant digits on product labels (and in most countries there are rules to round the values even if somehow a greater precision could be measured).

stephanegigandet:
Also related: #3487

Issue #2416, opened 2019-10-10
Use a templating system for the HTML code


teolemon:
### What
Extract all the HTML code from CGI files and put them in a separate template, using a standard templating system

### Why
As a Frontend developper, I don't want to think of the underlying Perl code when I improve the OFF interface

### Tasks

- [x] #3502
- [x] #3846
- [x] #3503
- [x] #3631
- [x] #3728
- [x] #3802
- [x] #3845
- [x] #3858
- [x] #3897
- [x] #3898
- [x] #3902
- [x] #3976
- [x] #3985
- [x] #3995
- [x] #4003
- [x] #4021
- [x] #5530
- [x] #5468

### Part of 
- #5527 
- #5513

stephanegigandet:
Note: doing this is quite complex. There is a lot of logic in the CGI that generate the HTML code. Most of the logic probably needs to remain there (otherwise if we move all the if / for loops etc. to the templating system, we won't gain much).

We need a way to start small. e.g. before going through the CGI scripts, it may be easier and more useful to first use a template for the code that generates the structure and navigation of all HTML pages, which is in Display.pm display_new() function.

stephanegigandet:
We also need to discuss which template system to use. See also bug #2361 from @zigouras .
https://metacpan.org/pod/HTML::Template is a good option.

hangy:
> https://metacpan.org/pod/HTML::Template is a good option.

Other maintained candidates that I found during research include https://metacpan.org/pod/Text::Xslate and https://metacpan.org/pod/Template::Toolkit

areeshatariq:
@stephanegigandet @hangy I would like to start working on this issue. Need your guidance on which template to use and which files to start with

stephanegigandet:
@areeshatariq : @teolemon is suggesting to start with Made Near Me, which is separate from the main product opener code, and is already a bit templated: https://github.com/openfoodfacts/openfoodfacts-server/tree/master/madenearme
It would be a good thing to try to use HTML::Template or Template::Toolkit to generate it.

See https://madenear.me 

The corresponding code is in /cgi/madenearme.pl

areeshatariq:
@stephanegigandet I'll start working on it. Thanks

areeshatariq:
> The corresponding code is in /cgi/madenearme.pl
Can't find this file

CharlesNepote:
> > https://metacpan.org/pod/HTML::Template is a good option.
> 
[...]

PR #3503, opened 2020-05-30
NutritionScore Calculation Details HTML Template

areeshatariq:
**Description:**
HTML Template has been created for Nutrition Score Calculation Details using Template::Toolkit
The template file is used for the presentation purpose only, while all the logic has been kept in /lib/ProductOpener/Display.pm

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -8636,6 +8636,15 @@ sub display_nutriscore_calculation_details($) {
 
 	my $nutriscore_data_ref = shift;
 	
+	my $is_beverage;
+	
+	if ($nutriscore_data_ref->{is_beverage}) {
+		$is_beverage = lang("nutriscore_is_beverage");
+	}
+	else {
+		$is_beverage = lang("nutriscore_is_not_beverage");
+	}
+
 	# Select message that explains the reason why the proteins points have been counted or not
 	
 	my $nutriscore_protein_info;
@@ -8662,7 +8671,7 @@ sub display_nutriscore_calculation_details($) {
 		
 		lang => \&lang,
 				
-		is_beverage => $nutriscore_data_ref->{is_beverage},
+		is_beverage => $is_beverage,
 		is_fat => $nutriscore_data_ref->{is_fat},
 		
 		nutriscore_protein_info => $nutriscore_protein_info,
@@ -8716,8 +8725,9 @@ sub display_nutriscore_calculation_details($) {
 		 
 		 push @{$template_data_ref->{points_groups}}, $points_group_ref;
 	 }
+
 	# Nutrition Score Calculation Template
-		my $config = {
+	my $config = {
 		INCLUDE_PATH => $data_root . '/templates',
 		INTERPOLATE => 1,
 		EVAL_PERL => 1,
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -159,6 +159,7 @@ use Storable qw(freeze);
 use Digest::MD5 qw(md5_hex);
 use boolean;
 use Excel::Writer::XLSX;
+use Template;
 
 use Log::Any '$log', default_adapter => 'Stderr';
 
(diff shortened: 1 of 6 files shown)


stephanegigandet:
Note for reviewers: @areeshatariq is working on using a template engine for the HTML code generated by Product Opener. Copying here Areeesha's message in #productopener on Slack:

"I've opened a Pull request on HTML Template of Nutrition Score Calculation Details. Started with Nutrition Score as it was a more independent module. We used Template::Toolkit for this. What do you think of the approach and how the template looks?"

lib/ProductOpener/Display.pm
```diff
@@ -8634,79 +8635,101 @@ the rounded value according to the Nutri-Score rules, and the corresponding poin
...
+				maximum => scalar(@{$points_thresholds{$nutrient_threshold_id}}),
+				value => $nutriscore_data_ref->{$nutrient},
+				rounded => $nutriscore_data_ref->{$nutrient . "_value"},
+			};
+		 }
+		 
+		 push @{$template_data_ref->{points_groups}}, $points_group_ref;
+	 }
+	# Nutrition Score Calculation Template
+		my $config = {
```
VaiTon, line 8720:
```suggestion
	my $config = {
```
Also, aren't these config we can use for every template? Can't we put them outside of the sub?

templates/nutrition_score.tt
```diff
@@ -0,0 +1,41 @@
+
```
VaiTon, line 1:
```suggestion
```

templates/nutrition_score.tt
```diff
@@ -0,0 +1,41 @@
...
+  [% END %]
+
+    <p>[% nutriscore_protein_info %] </p>
+    <p>
+      <strong> [% lang('nutriscore_score') %][% lang('sep') %]: [% score %] </strong> ( 
+      [% negative_points %] - [% positive_points %])
+    </p>
+    <p><strong> [% lang('nutriscore_grade') %][% lang('sep') %]: [% grade %] </strong></p>
+
+</div>
```
VaiTon, line 41:
```suggestion
</div>

```

templates/nutrition_score.tt
```diff
@@ -0,0 +1,41 @@
...
+  [% FOREACH group IN points_groups %]
+
+    <p> 
+      <strong> [% lang("nutriscore_" _ group.type _ "_points") %][% lang('sep') %]: [% group.points %]  </strong> 
+    </p>
+
+    <ul>
+      [% FOREACH nutrient IN group.nutrients %]
+        <li>
+          <strong> [% lang("nutriscore_points_for_" _ nutrient.id) %][% lang('sep') %]: [% nutrient.points %]&nbsp</strong>
```
VaiTon, line 25: `[% lang("nutriscore_points_for_" _ nutrient.id) %]`

What's the `_` before `nutrient.id` for?

areeshatariq: @VaiTon it's for concatenation

Ban3: @VaiTon The underscore operator is explained at least on this part of the manual: http://template-toolkit.org/docs/manual/Directives.html#section_SET

Maybe the double quoted string interpolation would be clearer.

VaiTon: > In Perl 5, the . is used for string concatenation, but in Perl 6, as in the Template Toolkit, the . will be used as the method calling operator and ' _ ' will be used for string concatenation.

I was used to the `.` for concatenation in Perl.

> Maybe the double quoted string interpolation would be clearer.
I agree. The others? @openfoodfacts/open-food-facts-contributors

CharlesNepote:
That's a VERY good start, thank you @areeshatariq!

The Template::Toolkit is not as usable as I would like, but it's a good enhancement.

PROs:
* I'm able to read the template and work on it easily (for example, I would like to add CSS classes or ids to some paragraphs)
* the template is/can be prettified for better readability and ease of use (indentation etc.)

CONs:
* Github doesn't provide syntax highlighting for the template

Questions:
* is it possible to insert comments in the template that won't be rendered in the HTML: I think it's very important to let back-end devs document some things
* is it possible to validate HTML code during the pull request workflow
* shouldn't we avoid conditionals IF-THEN-ELSE as much as we can: e.g.
```
  [% IF is_beverage %]
    <p> [% lang('nutriscore_is_beverage') %] </p>
  [% ELSE %]
    <p> [% lang('nutriscore_is_not_beverage') %] </p>
  [% END %]
```
shouldn't we compute a `nutriscore_beverageness` in the back-end?

templates/nutrition_score.tt
```diff
@@ -0,0 +1,41 @@
...
+          /&nbsp [% nutrient.maximum %][% lang("points") %] ([% lang("nutriscore_source_value") %][% lang("sep") %]: [% nutrient.value %],
```
stephanegigandet, line 26: the &nbsp in the code should be &nbsp; with a ;

templates/nutrition_score.tt
stephanegigandet: I just checked, we can do this instead:

[% lang("nutriscore_${group.type}_points") %]

[% lang("nutriscore_points_for_${nutrient.id}") %]

I think it's more clear than the [% lang("nutriscore_" _ group.type _ "_points") %] syntax.

areeshatariq:
>     * is it possible to insert comments in the template that won't be rendered in the HTML: I think it's very important to let back-end devs document some things
Yes we can insert comments that won't be rendered in the HTML

>     * shouldn't we avoid conditionals IF-THEN-ELSE as much as we can: e.g.
> 
> 
> ```
>   [% IF is_beverage %]
>     <p> [% lang('nutriscore_is_beverage') %] </p>
>   [% ELSE %]
>     <p> [% lang('nutriscore_is_not_beverage') %] </p>
>   [% END %]
> ```
> 
> shouldn't we compute a `nutriscore_beverageness` in the back-end?

Yes, we did try to avoid as many conditions as we can to make the template easier to understand. I'll add this If Else to the back-end. Thank you

templates/nutrition_score.tt
VaiTon, line 26:
```suggestion
          /&nbsp; [% nutrient.maximum %][% lang("points") %] ([% lang("nutriscore_source_value") %][% lang("sep") %]: [% nutrient.value %], 
```

lib/ProductOpener/Display.pm
```diff
@@ -8636,6 +8636,15 @@ sub display_nutriscore_calculation_details($) {
 
 	my $nutriscore_data_ref = shift;
 	
+	my $is_beverage;
```
VaiTon, line 8639: Would rename to "beverage_view" or something like this. The is_ prefix makes me think of a bool.

stephanegigandet:
> * shouldn't we avoid conditionals IF-THEN-ELSE as much as we can: e.g.
> 
> ```
>   [% IF is_beverage %]
>     <p> [% lang('nutriscore_is_beverage') %] </p>
>   [% ELSE %]
>     <p> [% lang('nutriscore_is_not_beverage') %] </p>
>   [% END %]
> ```
> 
> shouldn't we compute a `nutriscore_beverageness` in the back-end?

Hi @CharlesNepote,

That's a very good question.




> * shouldn't we avoid conditionals IF-THEN-ELSE as much as we can: e.g.
> 
> ```
>   [% IF is_beverage %]
>     <p> [% lang('nutriscore_is_beverage') %] </p>
>   [% ELSE %]
>     <p> [% lang('nutriscore_is_not_beverage') %] </p>
>   [% END %]
> ```
> 
> shouldn't we compute a `nutriscore_beverageness` in the back-end?

That's a very good question @CharlesNepote .

@areeshatariq changed it to:

  <p> [% beverage_view %] </p>

It does make the template simpler, but we lost the connection to the string id in the .po file. That's another interesting question: should we try to put the calls to lang () to internationalize the strings in the template, or should we try to do all the calls to lang() in the Perl code.

The next statement is interesting too:
  
  [% IF is_fat %]
    <p> [% lang('nutriscore_proteins_is_added_fat') %] </p>
  [% END %]

We need the if statement because we don't want the <p> to be generated in the Perl code.

We could to the call to lang() in Perl though, and pass the translated message directly, in the "is_fat" variable (instead of just passing 0 or 1).

  [% IF is_fat %]
    <p> [% is_fat %] </p>
  [% END %]

stephanegigandet:
> CONs:
> 
> * Github doesn't provide syntax highlighting for the template

@CharlesNepote , @areeshatariq : there is a simple solution to that: rename the .tt file in .html :) (it's in the /templates/ directory so it's obvious it's a template).

I just tried it, we have HTML highlighting in github: https://github.com/openfoodfacts/openfoodfacts-server/blob/areesha-stephane-wip/templates/nutrition_score.html

and also in the text / code editors:

image

Ban3:
I added github syntax highlighting as HTML for the templates (per suggestion from @stephanegigandet). If there's a better option than HTML, see [here](https://github.com/github/linguist#overrides) how to configure it.

stephanegigandet:
> I added github syntax highlighting as HTML for the templates (per suggestion from @stephanegigandet). If there's a better option than HTML, see [here](https://github.com/github/linguist#overrides) how to configure it.

Thanks @Ban3 . We probably will have templates for dynamically generated JS code as well (and possibly CSS). So I think the simplest thing is just to rename the template so that they have extension of what we are generating.  /templates/nutriscore.html (instead of .tt)

hangy:
> We probably will have templates for dynamically generated JS code as well (and possibly CSS).

AFAIK, it is usually recommended to not create JS/CSS on-the-fly, but to use techniques like unobtrusive JS and to use selectors efficiently. 🙂 

> So I think the simplest thing is just to rename the template so that they have extension of what we are generating. /templates/nutriscore.html (instead of .tt)

The templates are not valid HTML/CSS/JS, so that working syntax highlighting basically depends on how forgiving the parser is. 🤷‍♂️

PR #3510, opened 2020-05-30
chore: Removed powders from ingredients taxonomy

aleene:
**Description:**
Removing powders from ingredients taxonomy. Added missing translations to processes taxonomy. Doublechecked removals in ingredient parents (conjugations).


stephanegigandet:
It will be great to make the ingredients taxonomy smaller. :)
It is still a work in progress right?  Or you want to merge it now?

aleene:
It is still a work in process

aleene:
I repaired this old pull request. time to merge

taxonomies/ingredients.txt
```diff
@@ -9806,21 +9823,22 @@ es:queso de leche pasteurizada
...
-fr:fromage en poudre, poudre de fromage
-hu:sajtpor, porított sajt
-it:formaggio in polvere
-nl:kaaspoeder
-pl:ser w proszku
-pt:queijo em pó
-ru:Сырный порошок
-sk:sirotka v prahu
-sv:ostpulver
+#en:processing:en:powder
```
stephanegigandet, line 9826: I'm not sure I understand. Why do we keep en:cheese powder in the taxonomy? Is that because some languages don't have support for "powder" parsing yet? In that case we should leave those languages until we do have support I think.

aleene: There are more examples like this. It is mostly due to children, which I can not delete. For instance when there is an associated CIQUAL.

benbenben2:
@aleene @stephanegigandet 

Please review last changes.

ingredients.pm has been updated to cover cases where processing is in a separated word (before the ingredient), such as "jauhettu vaniljatanko"

"hedelmätäysmehutiivisteet" was a bit particular (somewhat between ingredient and ingredients_processing). Instead it has been replaced by dried banana.

aleene:
Great you resolved this one.

stephanegigandet (approved):
Thank you!!!

PR #3511, opened 2020-05-30
Refresh the Open Beauty Facts data page + (OPF/OPFF pages)

teolemon:
Refresh the Open Beauty Facts data page


stephanegigandet:
The opf and opff texts seem inverted

teolemon:
fixed @stephanegigandet thx

PR #3522, opened 2020-06-01
Logo for Haute Valeur Environnementale

teolemon:
Logo for Haute Valeur Environnementale


html/images/lang/fr/labels/haute-valeur-environnementale-500x500.svg
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="162px" height="162px" viewBox="0 0 162 162" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 63.1 (92452) - https://sketch.com -->
```
teolemon, line 3:
```suggestion
```

html/images/lang/fr/labels/haute-valeur-environnementale-500x500.svg
```diff
@@ -0,0 +1,75 @@
...
+    <title>hve_prod_coul</title>
+    <desc>Created with Sketch.</desc>
```
teolemon, line 5:
```suggestion
    <desc>Created from official PDF file.</desc>
```

stephanegigandet:
Please rename the suffix to .90x90.svg  (height needs to be 90 in the file name)

teolemon:
Does it have to be 90 for a SVG ?

stephanegigandet:
the SVG itself can stay as is, you just need to change the file name. This is so that we know how wide the image is when displayed at 90px height.

teolemon:
How about the raster ? Does it need a specific name as well or can it be just a confirmation of the ratio ? (90x90 with an actual 500x500)

stephanegigandet:
> How about the raster ? Does it need a specific name as well or can it be just a confirmation of the ratio ? (90x90 with an actual 500x500)

If there is a png file, it should be resized to 90px height, and the name of the file should be .[actual width]x90.png

stephanegigandet:
Only one file is needed for each label. If there is a a SVG, it will be used and the PNG won't be used.

Issue #338, opened 2016-06-26
Load all remaining EU packaging codes


teolemon:
```[tasklist]
### Tasks
- [ ] Austria #333
- [x] Belgium #1492
- [ ] #6512
- [x] Load packager codes for Croatia
- [ ] Load packager codes for Republic of Cyprus
- [ ] Load packager codes for Czech Republic
- [ ] Load packager codes for Denmark
- [x] Load packager codes for Estonia
- [x] Load packager codes for Finland
- [x] Load packager codes for France
- [x] Load packager codes for Germany
- [ ] Load packager codes for Greece
-  [x] Hungary #1494
-  [ ] Ireland #1572
-  [x] Load packager codes for Italy
-  [ ] Load packager codes for Latvia
-  [x] Lithuania #1493
-  [ ] Luxembourg #331
-  [ ] Load packager codes for Malta
-  [ ] Netherlands #335
-  [ ] Poland #334
-  [ ] Portugal #1468
-  [ ] Load packager codes for Romania
-  [ ] Slovakia #1555
-  [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/8958
-  [x] Load packager codes for Spain
-  [ ] Sweden #1477
- [x] Load packager codes for UK
```
### Part of
- https://github.com/openfoodfacts/openfoodfacts-server/issues/5508

MaximeLaurenty:
I claim Portugal !

MaximeLaurenty:
please update

teolemon:
done

teolemon:
Luxembourg claimed and done by @k3lyn

MaximeLaurenty:
Portugal claimed & under progress

MaximeLaurenty:
A small code that I wrote to help me aggregate all csv file into one (useful when there are a lot of them ...)
Feel free to use.

```
import csv
import os
### Lines with 3 ### might need adaptation to your need
MAIN_CSV = '/Users/Maxime/Downloads/Portugal/Portugal.csv' ### PATH of the csv where you will agregate everything
main_csv_file = open(MAIN_CSV, 'a', newline='',encoding = "ISO-8859-1")  # 'a' to append and not overwrite ; encoding might need adaptation to handle special characters
main_csv_writer = csv.writer(main_csv_file, delimiter=';',quotechar='|') ### DELIMITER to choose wisely

[...]

PR #3524, opened 2020-06-01
Add packager codes for Estonia

Ban3:
**Description:**

- Add approval numbers available from https://www.agri.ee/et/avaandmed
- Add a script that does the conversion to TSV and geocoding for future reference.
- Update the built packager codes.

**Related issues and discussion:** #338


scripts/packager-codes/toidukaitlejad-tsv.xsl
```diff
@@ -0,0 +1,56 @@
...
+
+<xsl:template match="row">
+  <xsl:for-each select="child::*">
+    <xsl:if test="position() != last()">
+      <xsl:value-of select="normalize-space(.)"/>
+      <xsl:value-of select="$separator" />
+    </xsl:if>
+    <xsl:if test="position() = last()">
+      <xsl:value-of select="normalize-space(.)"/>
+      <xsl:text>&#xa;</xsl:text>
```
Ban3, line 51: Factor out the repeated line from the ifs.
```suggestion
    <xsl:value-of select="normalize-space(.)"/>
    <xsl:if test="position() != last()">
      <xsl:value-of select="$separator" />
    </xsl:if>
    <xsl:if test="position() = last()">
      <xsl:text>&#xa;</xsl:text>
```

PR #3536, opened 2020-06-02
Get output (errors) from the taxonomy build in the travis test logs

stephanegigandet:
Let's see if this works.


VaiTon:
Appparently it does work!

stephanegigandet:
> Appparently it does work!

I does now. :)

PR #3559, opened 2020-06-05
Small changes to Nutri-Score template

stephanegigandet:
- moved template initialization at the start of the Display.pm module so that it's done only once when Apache+mod_perl loads (on Apache process can initialize once and then serve many requests)
- changed template extension from .tt to .html in order to have syntax highlighting in github, code editors etc.
- removed extra spaces between tags
- removed the deprecated lang("sep") (including in a lot of places I had added it to, I forgot it was replaced a while back)
- changed &raquo to &raquo; (I prefer to be on the safe side :) )


stephanegigandet:
If you want to see @areeshatariq's work in action, it's currently on the dev server: https://world.openfoodfacts.dev/product/3760091726667/chocolat-noir-80-equateur-nicaragua-ethiquable (login : off  , password : off ), then click on "Details of the Nutri-Score calculation".

VaiTon:
I think it would be better if we named the templates with a custom "suffix". Like template.tmpl.html

stephanegigandet:
> I think it would be better if we named the templates with a custom "suffix". Like template.tmpl.html

We could name them like nutriscore.tt.html (tt for template toolkit).

But they are located in the /templates/ directory, so it should be already clear that they are templates.

stephanegigandet:
I just added Template options so that the templates are compiled and cached.

That means changes to the template on disk may not be reflected for 60 seconds (the module won't stat the file again to see if it has changed).

The cache is for each mod_perl process.

When developing and changing the template, it's best to stop and start Apache again (or wait 60 seconds after changing the template).

PR #3583, opened 2020-06-10
copy all js, including minified script

stephanegigandet:
used by image upload form in producers platform

### DIFF
--- a/cgi/import_photos_upload.pl
+++ b/cgi/import_photos_upload.pl
@@ -307,11 +307,11 @@ HTML
     </script>
 
     <!-- The Templates plugin is included to render the upload/download listings -->
-    <script src="/js/dist/tmpl.min.js"></script>
+    <script src="/js/dist/tmpl.js"></script>
     <!-- The Load Image plugin is included for the preview images and image resizing functionality -->
     <script src="/js/dist/load-image.all.min.js"></script>
     <!-- The Canvas to Blob plugin is included for image resizing functionality -->
-    <script src="/js/dist/canvas-to-blob.min.js"></script>
+    <script src="/js/dist/canvas-to-blob.js"></script>
 
     <!-- The Iframe Transport is required for browsers without support for XHR file uploads -->
     <script src="/js/dist/jquery.iframe-transport.js"></script>
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -848,7 +848,6 @@ HTML
 	$scripts .= <<HTML
 <script type="text/javascript" src="/js/dist/webcomponentsjs/webcomponents-loader.js"></script>
 <script type="text/javascript" src="/js/dist/cropper.js"></script>
-<script type="text/javascript" src="/js/jquery.tagsinput.20160520/jquery.tagsinput.min.js"></script>
 <script type="text/javascript" src="/js/jquery.form.js"></script>
 <script type="text/javascript" src="/js/dist/tagify.min.js"></script>
 <script type="text/javascript" src="/js/dist/jquery.iframe-transport.js"></script>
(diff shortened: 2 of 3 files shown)


stephanegigandet:
> > I don't think that pre-minified versions should be copied. They're minified through the `terser()` call and everything else might cause issues.
> 
> Note that https://static.openfoodfacts.org/js/dist/jquery.fileupload.js _is_ minified.

@hangy : I'm a bit confused.

We have those files:

root@off1:/srv/off/html/js/dist# ls -lrt canvas-to-blob.*
-rw-r--r-- 1 root root 1282 Oct 26  1985 canvas-to-blob.min.js
-rw-r--r-- 1 root root 1279 Oct 26  1985 canvas-to-blob.js
-rw-r--r-- 1 root root 5914 Jun 10 15:09 canvas-to-blob.js.map
-rw-r--r-- 1 root root 3187 Jun 10 15:09 canvas-to-blob.min.js.map

Which one should actually be loaded with the <script> tag?

Do we need to copy the corresponding .map file too?

What about the other scripts? Should we copy only [script name].js and replace all the loading of [script name].min.js by [script name].js ?

      "./node_modules/@webcomponents/**/webcomponentsjs/**/*.js",
      "./node_modules/foundation-sites/js/vendor/*.js",
      "./node_modules/foundation-sites/js/foundation.js",
      "./node_modules/papaparse/papaparse.js",
      "./node_modules/osmtogeojson/osmtogeojson.js",
      "./node_modules/leaflet/dist/leaflet.js",
      "./node_modules/leaflet.markercluster/dist/leaflet.markercluster.js",
      "./node_modules/blueimp-tmpl/js/*.js",
      "./node_modules/blueimp-load-image/js/load-image.all.min.js",
      "./node_modules/blueimp-canvas-to-blob/js/*.js",
      "./node_modules/blueimp-file-upload/js/*.js",
      "./node_modules/@yaireo/tagify/dist/tagify.min.js",
      "./node_modules/cropper/dist/*.js"

hangy:
> root@off1:/srv/off/html/js/dist# ls -lrt canvas-to-blob.*
> -rw-r--r-- 1 root root 1282 Oct 26 1985 canvas-to-blob.min.js
> -rw-r--r-- 1 root root 1279 Oct 26 1985 canvas-to-blob.js
> -rw-r--r-- 1 root root 5914 Jun 10 15:09 canvas-to-blob.js.map
> -rw-r--r-- 1 root root 3187 Jun 10 15:09 canvas-to-blob.min.js.map

Based on a clean dist directory, Gulp should only create these
- canvas-to-blob.js 
- canvas-to-blob.js.map
 
> Which one should actually be loaded with the <script> tag?

canvas-to-blob.js is the only one that should exist after cleaning and re-creating the dist directory using Gulp, so that should be used.
 
> Do we need to copy the corresponding .map file too?

Gulp creates those. They're not actually required, but can help debugging JS issues in the browser.

> What about the other scripts? Should we copy only [script name].js and replace all the loading of [script name].min.js by [script name].js ?
> 
> ```
>   "./node_modules/@webcomponents/**/webcomponentsjs/**/*.js",
>   "./node_modules/foundation-sites/js/vendor/*.js",
>   "./node_modules/foundation-sites/js/foundation.js",
>   "./node_modules/papaparse/papaparse.js",
>   "./node_modules/osmtogeojson/osmtogeojson.js",
>   "./node_modules/leaflet/dist/leaflet.js",
>   "./node_modules/leaflet.markercluster/dist/leaflet.markercluster.js",
>   "./node_modules/blueimp-tmpl/js/*.js",
>   "./node_modules/blueimp-load-image/js/load-image.all.min.js",
>   "./node_modules/blueimp-canvas-to-blob/js/*.js",
>   "./node_modules/blueimp-file-upload/js/*.js",
>   "./node_modules/@yaireo/tagify/dist/tagify.min.js",
>   "./node_modules/cropper/dist/*.js"
> ```

Where non-minified versions were available, I usually chose those, so that Gulp would need to re-minimize them during the build process. Also, already minified scripts aren't that useful with map files. 🙂 However, some of the node.js modules don't provide non-minified compiled versions in their dist, so that I had to use the minified versions.

stephanegigandet:
@hangy : thanks for the explanation, I updated gulpfile.js and the CGI scripts to load the right files.

PR #3585, opened 2020-06-10
feat: Official videos for Nutri-Score

teolemon:
### What
- [ ] Here's I've added everything in bulk
- [ ] I'd very much like to put the videos in an array, and pass only the relevant ones in this page for a given language, and possibly as a JSON for use in the mobile apps/knowledge panels/…
### Part of
- #5517


teolemon:
- [ ] Closing and moving to https://github.com/openfoodfacts/openfoodfacts-web/pull/345

PR #3597, opened 2020-06-11
Remove the petition from the iPhone tagline

teolemon:
Remove the petition from the iPhone tagline


VaiTon:
@teolemon do crowdin takes json files?

teolemon:
It wouldn't work because all languages are in the same file (and moreover the format is totally custom). We would need a script to generate the file from per-language files.

PR #3619, opened 2020-06-13
Process(-result) Juice added

aleene:
**Description:**
Process(-result) Juice added


taxonomies/ingredients.txt
```diff
@@ -32355,10 +32361,11 @@ nl:ammonium- en natriumcarbonaten, ammonium carbonaten en natrium
 #
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
-
+en:fruit and vegetable
```
stephanegigandet, line 32364: I would prefer not to add entries for "something and something else".

aleene: I added it to capture `fruit and vegetable juice`. I see it as a combined ingredient, not as two separate ones. Thus I made it a parent of `en:fruit` and `en:vegetable`. For other and-constructions I do the inverse, i.e. `en:fruit` and `en:vegetable` would be parents of `en:fruit and vegetable`.

stephanegigandet: OK, in that case we should deal with "fruit and vegetable juice" in a different way, without making "fruit and vegetable" an ingredient. We will need to handle it when we do the parsing. Because otherwise we would need add countless combinations so that we can handle "mango and apple juice" etc.

We also can't have "fruit and vegetable" as a parent of "fruit" and "vegetable", otherwise all products that have either a fruit, or a vegetable, will have "fruit and vegetable" as an ingredient, which would be very confusing.

taxonomies/ingredients_processing.txt
```diff
@@ -894,3 +934,5 @@ es:sin hueso
 
 en:inflated
 es:hinchado,hinchada,hinchados,hinchadas
+
+#en:for color, added for color, colored with
```
stephanegigandet, line 938: Those are in fact additives classes, we will deal with it in a differenty way, but it needs some changes to the parsing code.

aleene: That's why I put those in comments. I am just gathering all adjectives.

PR #3624, opened 2020-06-13
normalize strings to Unicode form NFC, fixes #3615

### LINKED ISSUE #3615: Accent handling in ingredients
I'm coming across cases where accented characters aren't being correctly handled.
For example (Spanish):
The ingredients list includes arándanos
In the ingredients taxonomy arándanos (with accent exists) but arándanos doesn't get processed
arandanos (without accent) does get processed.

It appears ingredients in the ingredients.txt shouldn't have accents but there are hundreds of entries with accent (in all languages).

stephanegigandet:
fixes #3615, thanks @Ban3 for the suggestion of Unicode::Normalize

### DIFF
--- a/lib/ProductOpener/Store.pm
+++ b/lib/ProductOpener/Store.pm
@@ -121,8 +121,6 @@ sub get_string_id_for_lang {
 		$string =~ s/œ|Œ/oe/g;
 		$string =~ s/æ|Æ/ae/g;
 		$string =~ s/ß|\N{U+1E9E}/ss/g;
-		# Remove diacritics marks -> use Unicode::Normalize instead
-		#$string =~ s/\N{COMBINING ACUTE ACCENT}|\N{COMBINING GRAVE ACCENT}|\N{COMBINING CIRCUMFLEX ACCENT}|\N{COMBINING TILDE}|\N{COMBINING DIAERESIS}//g;
 	}
 
 	# turn special chars and zero width space to -
--- a/lib/ProductOpener/Store.pm
+++ b/lib/ProductOpener/Store.pm
@@ -49,6 +49,7 @@ use Storable qw(lock_store lock_nstore lock_retrieve);
 use Encode;
 use Encode::Punycode;
 use URI::Escape::XS;
+use Unicode::Normalize;
 use Log::Any qw($log);
 
 # Text::Unaccent unac_string causes Apache core dumps with Apache 2.4 and mod_perl 2.0.9 on jessie
@@ -86,6 +87,10 @@ sub get_string_id_for_lang {
 	if (not defined $string) {
 		return "";
 	}
+	
+	# Normalize Unicode characters
+	# Form NFC
+	$string = NFC($string); 
 
 	my $unaccent = $string_normalization_for_lang{default}{unaccent};
 	my $lowercase = $string_normalization_for_lang{default}{lowercase};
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Store.pm
```diff
@@ -49,6 +49,7 @@ use Storable qw(lock_store lock_nstore lock_retrieve);
 use Encode;
 use Encode::Punycode;
 use URI::Escape::XS;
+use Unicode::Normalize;
```
hangy, line 52: Should this be added to `cpanfile` to pin it to a specific version? I'm not sure if this is distributed with core perl.

lib/ProductOpener/Store.pm
```diff
@@ -116,6 +121,8 @@ sub get_string_id_for_lang {
 		$string =~ s/œ|Œ/oe/g;
 		$string =~ s/æ|Æ/ae/g;
 		$string =~ s/ß|\N{U+1E9E}/ss/g;
+		# Remove diacritics marks -> use Unicode::Normalize instead
+		#$string =~ s/\N{COMBINING ACUTE ACCENT}|\N{COMBINING GRAVE ACCENT}|\N{COMBINING CIRCUMFLEX ACCENT}|\N{COMBINING TILDE}|\N{COMBINING DIAERESIS}//g;
```
hangy, line 125: Do we need the comment?

lib/ProductOpener/Store.pm
```diff
@@ -86,6 +87,10 @@ sub get_string_id_for_lang {
 	if (not defined $string) {
 		return "";
 	}
+	
+	# Normalize Unicode characters
+	# Form NFC
+	$string = NFC($string);
```
hangy, line 93: Am I correct in the assumption that the `NFC` sub uses native code instead of pure perl, so that the performance should be quite good? I'm just wary of using a 3rd party method here, since I seem to remember the performance impact of changes in this method was quite important for overall ProductOpener performance. 🙂

lib/ProductOpener/Store.pm
Ban3: @hangy it's a core module http://perlpunks.de/corelist/mversion?module=Unicode%3A%3ANormalize

lib/ProductOpener/Store.pm
Ban3: > Am I correct in the assumption that the NFC sub uses native code instead of pure perl

Correct. [It's implemented in XS](https://metacpan.org/source/KHW/Unicode-Normalize-1.26/Normalize.xs#L576)

stephanegigandet: I did some testing, running get_string_id_for_lang 1M times with a small string:

Original function:
root@ns3135968:/srv/off/scripts# time ./test_get_string_id.pl 

real	0m12.530s
user	0m12.524s
sys	0m0.004s

---

Regexp approach:
root@ns3135968:/srv/off/scripts# time ./test_get_string_id_regexp.pl 

real	0m15.718s
user	0m15.704s
sys	0m0.012s

---
With Unicode::Normalize:
root@ns3135968:/srv/off/scripts# time ./test_get_string_id_nfc.pl 

real	0m15.754s
user	0m15.744s
sys	0m0.008s

I think it's worth using Unicode::Normalize, as we actually need to normalize to match entries to the taxonomies (in all languages with potentially composed characters like accents, not only when we want to remove accents).

lib/ProductOpener/Store.pm
stephanegigandet: I'm happy to remove it.

stephanegigandet: Removed.

lib/ProductOpener/Store.pm
hangy: > Correct. [It's implemented in XS](https://metacpan.org/source/KHW/Unicode-Normalize-1.26/Normalize.xs#L576)

Perfect, thanks. 🙂 

> I think it's worth using Unicode::Normalize, as we actually need to normalize to match entries to the taxonomies (in all languages with potentially composed characters like accents, not only when we want to remove accents).

Agreed, and the benchmarks don't look like the impact is too bad. Thanks for adding the benchmark numbers. 🙂

PR #3628, opened 2020-06-15
Nordic unknown ingredients

Ban3:
**Description:**

First batch of unknown ingredients from the dk, fi, is, no and sv domains.

I'll add the Perl parts (abbreviations, traces, processing...) in a later pull request.

### DIFF
--- a/taxonomies/ingredients_processing.txt
+++ b/taxonomies/ingredients_processing.txt
@@ -22,8 +22,13 @@
 #######################################################################
 
 en:raw
+da:rå
+fi:raaka
 fr:cru, crue, crus, crues
 es:crudo, cruda, crudos, crudas
+# da:parse:rå$
+# fi:parse:raaka$
+# sv:parse:rå$
 
 en:fresh
 de:frische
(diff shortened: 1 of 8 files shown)


taxonomies/ingredients_processing.txt
```diff
@@ -22,8 +22,13 @@
 #######################################################################
 
 en:raw
+da:rå
+fi:raaka
 fr:cru, crue, crus, crues
 es:crudo, cruda, crudos, crudas
+# da:parse:rå$
```
stephanegigandet, line 29: What is the meaning of this comment? Do you have some examples of ingredients with rå or raaka?
Are there risks of false positives?

Ban3: > What is the meaning of this comment?

Just to indicate it has apperead as a prefix, without space.

> Do you have some examples of ingredients with rå or raaka?

For danish, there's already a taxonomified [rå rørsukker](https://dk.openfoodfacts.org/ingredient/R%C3%A5%20r%C3%B8rsukker). As part of a compound word there's at least Finnish [raakalakritsi](https://fi.openfoodfacts.org/ingredient/raakalakritsi), same product has Danish `rålakrids`.

> Are there risks of false positives?

For Danish and Swedish I would say yes. Same goes for the other short ones like `hel`. Råg in Swedish comes to mind as a ingredient starting with rå but unrelated to raw. There's much less risk for the longer Finnish word.

PR #3638, opened 2020-06-15
restrict cross language tag matching, fixes #3626

### LINKED ISSUE #3626: "cal." or "cal" (for "calibre") adds E529
"cal." or "cal" (for "calibre") adds E529
```Orange Valencia Late (cal. 6/7), traité avec Imazalil. E904, E914.```
https://fr.openfoodfacts.org/produit/8437016428276/oranges-a-jus-la-grappe

stephanegigandet:
We have been historically quite aggressive when matching tags: if it did not match in the target language, we tried in some other languages. The risk of false positives was low, but as we have now very extended taxonomies, the risk of false positives is higher (and it is less useful because our data is cleaner).

This issue is to introduce stricter rules, so that we specify which site tries which language for only a very limited set of taxonomies:

"la" is Latin, it's quite useful for species in ingredients, and the mysterious "aqua" ingredient in beauty products :)

```perl
				if ($options{product_type} eq "food") {
				
					if ($tagtype eq "ingredients") {
						@test_languages = ("la");
					}
					elsif ($tagtype eq "additives") {
						@test_languages = ("en");
					}				
				}
				elsif ($options{product_type} eq "beauty") {
				
					# Beauty products ingredients are often in English
					if ($tagtype eq "ingredients") {
						@test_languages = ("en", "la");
					}
					elsif ($tagtype eq "additives") {
						@test_languages = ("en");
					}				
				}

### DIFF
--- a/lib/ProductOpener/Config_obf.pm
+++ b/lib/ProductOpener/Config_obf.pm
@@ -175,6 +175,8 @@ use ProductOpener::Config2;
 
 );
 
+$options{product_type} = "beauty";
+
 @edit_rules = ();
 
 
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -2133,7 +2133,7 @@ sub display_taxonomy_tag_link($$$) {
 	my $css_class = get_tag_css_class($target_lc, $tagtype, $tag);
 
 	my $html;
-	if ((defined $tag_lc) and ($tag_lc ne $lc)) {
+	if ((defined $tag_lc) and ($tag_lc ne $target_lc)) {
 		$html = "<a href=\"/$path/$tagurl\" class=\"$css_class\" lang=\"$tag_lc\">$tag_lc:$tag</a>";
 	}
 	else {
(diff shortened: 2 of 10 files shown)


teolemon:
The description of the benefit is quite nebulous. Can you explain how it is going to impact up Open Beauty Facts?

stephanegigandet:
> The description of the benefit is quite nebulous. Can you explain how it is going to impact up Open Beauty Facts?

It's not going to affect OBF much, if at all. For OBF it's important that we match ingredients not only to the target language, but also to Latin and English. We did that before, and we will keep doing that.

What changes is that we won't try to match tags (e.g. categories or additives) that are supposed to be in say French, to tags in Spanish, Italian or Russian. (which is what happened in the bug #3626 that you reported)

stephanegigandet:
For ingredients on OFF, the change is that we won't try to match tags to English as we currently do. It does not make a lot of sense anyway, that "feature" was not intended for ingredients. It was mostly useful for categories at one point, when we had issues of categories being entered in the wrong language a few years ago.

PR #3664, opened 2020-06-17
Brands taxonomy for OBF

teolemon:
Brands taxonomy for OBF


stephanegigandet:
> I think _all_ entries need an `en:` prefix for the taxonomy to be usable.

That's right, it won't build otherwise.

teolemon:
@stephanegigandet @hangy fixed

PR #3666, opened 2020-06-17
UNSPSC from slack file

teolemon:
UNSPSC from slack file (I couldn't find the production one on the server)


teolemon:
It's the categories for Open Products Facts, drawn from an international taxonomy by GS1.

PR #3668, opened 2020-06-17
better percent parsing, fixes #3667

### LINKED ISSUE #3667: Ingredient analysis: handle "minimum 10%"
Our current ingredients analysis parsing can parse things like "cocoa 10% min" or "Cocoa (10% minimum)", but not "Cocoa min. 10%"

stephanegigandet:
- Support for "some ingredient: min 10%" (previously we matched only "some ingredient: 10% min"
- Made a $percent_regexp to avoid repeating the same complex regexp 10 times
- Added some support for Spanish (e.g. "Cacao: 57% mínimo.")

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -690,12 +690,19 @@ sub extract_ingredients_from_image($$$$) {
 }
 
 
+my %min_regexp = (
+	en => "min|min\.minimum",
+	es => "min|min\.|mín|mín\.|mínimo|minimo|minimum",
+	fr => "min|min\.|mini|minimum",
+);
+
 # Words that can be ignored after a percent
 # e.g. 50% du poids total, 30% of the total weight
 
 my %ignore_strings_after_percent = (
 	en => "of (the )?total weight",
-	fr => "(min|min\.|mini|minimum)|(dans le chocolat( (blanc|noir|au lait))?)|(du poids total|du poids)",
+	es => "(en el chocolate( con leche)?)",
+	fr => "(dans le chocolat( (blanc|noir|au lait))?)|(du poids total|du poids)",
 );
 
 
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -690,12 +690,19 @@ sub extract_ingredients_from_image($$$$) {
 }
 
 
+my %min_regexp = (
```
hangy, line 693: As these are used to build a combined regex, and the sub is probably used quite often, it might be useful to [precompile](https://stackoverflow.com/a/953076/11963) the regular expressions in this hash, and other regexes this will be used in combination.

stephanegigandet: @hangy : I just tried it, but I didn't see any time benefit. Maybe my test is flawed, or Perl just compiles the same strings only once anyway.

```#!/usr/bin/perl -w

use strict;

my %min_regexp = (
	en => "min|min\.minimum",
	es => "min|min\.|mín|mín\.|mínimo|minimo|minimum",
	fr => "min|min\.|mini|minimum",
);

# Words that can be ignored after a percent
# e.g. 50% du poids total, 30% of the total weight

my %ignore_strings_after_percent = (
	en => "of (the )?total weight",
	es => "(en el chocolate( con leche)?)",
	fr => "(dans le chocolat( (blanc|noir|au lait))?)|(du poids total|du poids)",
);

my $product_lc = "fr";

my $min_regexp = "";
if (defined $min_regexp{$product_lc}) {
	$min_regexp = $min_regexp{$product_lc};
}
my $ignore_strings_after_percent = "";
if (defined $ignore_strings_after_percent{$product_lc}) {
	$ignore_strings_after_percent = $ignore_strings_after_percent{$product_lc};
}

my $percent_regexp = '(<|' . $min_regexp . '|\s)*(\d+((\,|\.)\d+)?)\s*(\%|g)\s*(' . $min_regexp . '|' . $ignore_strings_after_percent . '|\s|\)|\]|\}|\*)*';

my $re_ref = { fr => qr/^$percent_regexp$/i } ;  #precompile phase
my $re = qr/^$percent_regexp$/i;  #precompile phase

my $ingredients = "50% Min";

my $i = 0;
my $j = 0;

for ($i = 0; $i < 10 * 1000 * 1000; $i++) {

	if ($ingredients =~ /^$percent_regexp$/i) {
	#if ($ingredients =~ $re_ref->{fr}) {
		$j++;
	}
}

print STDERR "$i tests - $j matches\n";

hangy: According to other [benchmarks](http://www.pal-blog.de/entwicklung/perl/precompile-regexps-or-not.html) there could be some advantage of using precompiled regexes. Even if there's no performance penalty, using precompiled ones might reduce memory.

PR #3672, opened 2020-06-18
Self-service platform for producers: self-creation of professional accounts and orgs

stephanegigandet:
This PR is to allow producers (manufacturers or brands) to create professional accounts. Professional accounts have access to the platform for producers, where they can import Excel files with data for their products, and photos in bulk.

Changes:

1. Added a checkbox in the user creation / edit form to declare that it is a "professional account"
2. Checking that box asks for an organization
3. The organization is created if it doesn't exist, if there is already an organization with the same name, the requested org is pending (there's no system yet for org admins to receive and validate those requests, will be done later).
4. Pro accounts can be created either on the public platform, or the producer platform directly.
5. Welcome messages are different based on whether it is a pro account or not, and whether we are on the producers platform or not.

This is just a minimum viable product so that producers can start to use the producers platform without us having first to set up accounts for them (as we currently do). There are lots of things that are not implemented yet (e.g. editing organization profiles, notifications) or that can be improved (welcome email, suggesting first steps etc.). This will be done later.

If you want to test the different scenarios of creating users (pro or not) and using them, those changes are currently deployed on https://world.openfoodfacts.dev and https://world.pro.openfoodfacts.dev  (user and pass: off )

image


teolemon:
I have translated my usability document into English, and have added a few more issues @stephanegigandet 
https://docs.google.com/document/d/14e_CU8XUZdeOVNtGc_tcwynSB2kjoVSLzMlWjpClF2M/edit#

stephanegigandet:
> Would have been nice templatified

It's in @areeshatariq 's pipeline.

PR #3674, opened 2020-06-18
Fix add_tags_to_field()

stephanegigandet:
Fixed add_tags_to_field to prevent mix up of languages when all the fields added are already there.
Call compute_field_tags() directly from add_tags_to_field, only when fields have changed.

See #3675 for details.

Adding more tests to verify that the add_tags_to_field() function works correctly.


teolemon:
Is that app also behaving wrongly ?

stephanegigandet:
> Is that app also behaving wrongly ?

the off app? no it's good

PR #3684, opened 2020-06-19
OPF index to avoid OFF index

teolemon:
OPF index to avoid OFF index


hangy:
> OPF index to avoid OFF index

What does that mean?

teolemon:
Sorry for the short desc: we currently load the OFF index on OPF. This gives OPF its very own index.

PR #3686, opened 2020-06-19
Try to reduce memory used by Docker: start only 1 Apache server

stephanegigandet:
@areeshatariq is running into issues with docker, the Apache container sometimes fails because there isn't enough memory. (on a 12 Gb machine).

This is to change the Apache config to start only one server process (instead of 2), in the hope that it will reduce memory (although as some memory is shared, I'm not sure how much is actually saved).

This is what I get in top with this change:

``Tasks:   5 total,   1 running,   4 sleeping,   0 stopped,   0 zombie
%Cpu(s):  7.5 us,  2.1 sy,  0.0 ni, 89.8 id,  0.1 wa,  0.0 hi,  0.6 si,  0.0 st
MiB Mem :  15899.2 total,    187.0 free,  13855.0 used,   1857.2 buff/cache
MiB Swap:   2048.0 total,      0.1 free,   2047.9 used.   1016.4 avail Mem 

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                      
      1 root      20   0    2384    692    624 S   0.0   0.0   0:00.26 apache2ctl                   
      8 root      20   0    3864   2884   2456 S   0.0   0.0   0:00.24 bash                         
     15 root      20   0    8044   2976   2540 R   0.0   0.0   0:00.02 top                          
     20 root      20   0 2690500   2.5g  15936 S   0.0  16.2   0:29.29 /usr/sbin/apach              
     25 www-data  20   0 4616416   2.5g   8296 S   0.0  16.2   0:00.28 /usr/sbin/apach   

Without it, there is a 3rd 2.5g Apache process.

Something else we could try is to make smaller versions of taxonomy files (especially ingredients and categories).

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -55,6 +55,10 @@ FROM modperl AS runnable
 # Prepare Apache to include our custom config
 RUN rm /etc/apache2/sites-enabled/000-default.conf
 
+# Reduce the number of Apache processes
+RUN sed -i 's/StartServers\s*2/StartServers 1/' /etc/apache2/mods-enabled/mpm_event.conf
+RUN sed -i 's/MaxRequestWorkers\s*150/MaxRequestWorkers 25/' /etc/apache2/mods-enabled/mpm_event.conf
+
 # Copy Perl libraries from the builder image
 COPY --from=builder /tmp/local/ /opt/perl/local/
 ENV PERL5LIB /opt/perl/local/lib/perl5/


hangy:
Is there a significant difference when you look at `docker stats`? From the top of my head, I don't think I have seen the backend container taking up more than 3 GiB.

VaiTon:
Can confirm that allocating less than 4GB makes the container crash on startup on my machine (8 GB).

VaiTon:
Memory leak?

areeshatariq:
> Memory leak?

This was the error in my logs: Cannot allocate memory: AH00481: fork: Unable to fork new process

hangy:
For example, https://world.openfoodfacts.dev:8082/ running on the dev server, currently has the following usage:
```
CONTAINER ID        NAME                     CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
cbb43266e6d8        off-docker_mongodb_1     0.50%               74.29MiB / 31.38GiB   0.23%               155kB / 124kB       5.19MB / 208MB      34
7c80686118d8        off-docker_memcached_1   0.03%               5.145MiB / 31.38GiB   0.02%               79.3kB / 49.4kB     3.37MB / 0B         10
04031e6afe74        off-docker_backend_1     0.00%               2.778GiB / 31.38GiB   8.85%               289kB / 3.22MB      75.9MB / 52.3MB     56
82ce3a92afa6        off-docker_frontend_1    0.00%               7.051MiB / 31.38GiB   0.02%               3.16MB / 3.05MB     459kB / 0B          9
```

This is why I'm a bit surprised that you reported a few gigs not being enough. However, `top` running inside the container clearly shows more 2.5g RES for three processes:
```
top - 12:55:10 up 65 days,  4:58,  0 users,  load average: 0.93, 0.50, 0.31
Tasks:   6 total,   1 running,   5 sleeping,   0 stopped,   0 zombie
%Cpu(s):  4.6 us,  2.7 sy,  0.0 ni, 92.4 id,  0.2 wa,  0.0 hi,  0.1 si,  0.0 st
MiB Mem :  32128.5 total,   1799.3 free,  21211.0 used,   9118.2 buff/cache
MiB Swap:   2044.0 total,      0.0 free,   2043.9 used.  10134.6 avail Mem

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                                                                                                     1 root      20   0    0.0g   0.0g   0.0g S   0.0   0.0   0:00.06 apache2ctl                                                                                                                                                                 14 root      20   0    2.6g   2.5g   0.0g S   0.0   8.0   0:34.38 /usr/sbin/apach
   19 www-data  20   0    4.5g   2.5g   0.0g S   0.0   8.0   0:01.16 /usr/sbin/apach
   20 www-data  20   0    4.5g   2.5g   0.0g S   0.0   8.1   0:02.41 /usr/sbin/apach
   75 root      20   0    0.0g   0.0g   0.0g S   0.0   0.0   0:00.06 bash
   85 root      20   0    0.0g   0.0g   0.0g R   0.0   0.0   0:00.02 top                                                                                                                                                                     
```

> Can confirm that allocating less than 4GB makes the container crash on startup on my machine (8 GB).
> Memory leak?

I think there might be a difference in memory reported in `top` and `docker stats`. Also, it's possible that the process uses more memory on startup or during `scripts/build_lang.pl`. Have you observed something like this, @stephanegigandet?

> This was the error in my logs: Cannot allocate memory: AH00481: fork: Unable to fork new process

Have you tried it with the patch in this branch? It would be interesting to know if that helps on you machine. 🙂

stephanegigandet:
With docker stats, I don't see any different when there is one less process. So I'm guessing the memory is shared between the Apache processes. (a good thing :) )

stephanegigandet:
I'm testing a script that does nothing but loading Tags.pm (and thus the taxonomies).
It takes almost 1 Gb of memory.

16341 root      20   0  919876 735436  11464 S   0.0  2.2   0:05.47 test_memory.pl

Maybe something to investigate. The taxonomies on disk are only 60 Mb. The lang file with the translations is 10 Mb. Where does all the memory go?

VaiTon:
> 
> 
> I'm testing a script that does nothing but loading Tags.pm (and thus the taxonomies).
> It takes almost 1 Gb of memory.
> 
> 16341 root 20 0 919876 735436 11464 S 0.0 2.2 0:05.47 test_memory.pl
> 
> Maybe something to investigate. The taxonomies on disk are only 60 Mb. The lang file with the translations is 10 Mb. Where does all the memory go?

In building the tree IMHO

stephanegigandet:
https://metacpan.org/pod/Devel::Size

Doing more tests, adding this at the end of Tags.pm:

print STDERR "Tags.pm memory usage\n";

use Devel::Size qw(total_size);

print total_size(\%loaded_taxonomies) . " - loaded_taxonomies\n";
print total_size(\%stopwords) . " - stopwords\n";
print total_size(\%synonyms) . " - synonyms\n";
print total_size(\%synonyms_for) . " - synonyms_for\n";
print total_size(\%synonyms_for_extended) . " - synonyms_for_extended\n";
print total_size(\%just_synonyms) . " - just_synonyms\n";
print total_size(\%translations_from) . " - translations_from\n";
print total_size(\%translations_to) . " - translations_to\n";
print total_size(\%level) . " - level\n";
print total_size(\%direct_parents) . " - direct_parents\n";
print total_size(\%direct_children) . "  - direct_children\n";
print total_size(\%all_parents) . " - all_parents\n";
print total_size(\%properties) . " - properties\n";

Results in:

Tags.pm memory usage
2996 - loaded_taxonomies
96067 - stopwords
58505405 - synonyms
45784491 - synonyms_for
65333005 - synonyms_for_extended
12391 - just_synonyms
21373452 - translations_from
18129531 - translations_to
2118691 - level
4621210 - direct_parents
2415538  - direct_children
6599519 - all_parents
10576705 - properties
started

So the taxonomies do take a lot of memory.

stephanegigandet:
Someday we should revisit the taxonomy structures, when I created them, I didn't pay too much attention to memory usage...

In fact I think the biggest one is used only for taxonomy building, not after:

65 333 005 - synonyms_for_extended

And it's possible some others could be replaced, or structured in a better way.

VaiTon:
The total should be about 235 MB. It is a big chunk of the memory but we're far from 2.5GB mentioned earlier.

stephanegigandet:
Food.pm uses 250 Mb, 100 Mb for the categories stats for all countries, and 150 Mb for the geo coded addresses of the packager codes.

areeshatariq:
> Have you tried it with the patch in this branch? It would be interesting to know if that helps on you machine. slightly_smiling_face

Yea, I tried it. Didn't work for me

stephanegigandet:
Closing this PR as it does not seem to help much to reduce memory usage.

PR #3693, opened 2020-06-20
Spanish Catalan and more taxonomy

AcuarioCat:
Additions and corrections

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -8397,7 +8397,7 @@ it:latte intero sterilizzato, latte sterilizzato intero
 en:UHT pasteurised whole milk
 ca:llet sencera esterilitzada UHT, llet esterilitzada UHT sencera
 de:Haltbare Vollmilch, H-Vollmilch
-es:leche entera esterilizada UHT, leche esterilizada UHT entera
+es:leche entera esterilizada UHT, leche esterilizada UHT entera,leche UHT entera
 fi:iskukuumennettu täysmaito
 fr:lait entier stérilisé uht, lait entier stérilisé u h t, lait stérilisé uht entier
 hu:pasztőrözött UHT teljes tej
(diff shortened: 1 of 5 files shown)


taxonomies/ingredients.txt
```diff
@@ -19271,7 +19287,7 @@ fr:tournesol oléique
 en:high oleic sunflower oil
 ca:oli refinat de gira-sol alt oleic
 de:High Oleic-Sonnenblumenöl
-es:aceite de girasol alto oleico,aceite refinado de girasol alto oleico,huile de tournesol haut oléique
+es:aceite de girasol alto oleico,aceite refinado de girasol alto oleico,huile de tournesol haut oléique,omega 9
```
stephanegigandet, line 19290: why "omega 9" ?
also "huile de tournesol haut oléique" should be in the fr: after (but it's from another change)

taxonomies/ingredients.txt
```diff
@@ -71687,6 +71752,9 @@ wikidata:en:Q658413
 wikipedia:en:https://en.wikipedia.org/wiki/Bread_crumbs
 # ingredient/fr:panure has 462 products in 6 languages @2019-05-17
 
+<en:breadcrumbs
+en:wholmeal breadcrumbs
```
stephanegigandet, line 71756: wholemeal ?

AcuarioCat: Typo.. fixed

taxonomies/ingredients.txt
AcuarioCat: Omega 9 from these 3:
https://es.openfoodfacts.org/ingrediente/omega-9

stephanegigandet: "Aceite de glrasol con alto contenido de acido oleico (omega 9),"
--> we should not make "Omega 9" a synonym of "aceite de girasol alto oleico". There are lots of other possible sources of Omega 9.

Instead we can add Omega 9 as a single ingredient. (without linking it to the different oils).

AcuarioCat: ok

PR #3698, opened 2020-06-21
Open Beauty Facts dark mode logo

teolemon:
Open Beauty Facts dark mode logo


hangy:
Could we use a `style` tag in the `svg` instead of having to include a separate icon? Might be worth a test. 🙂 

- https://catalin.red/svg-favicon-light-dark-theme/#how-to-switch-the-svg-favicon-when-in-dark-mode
- https://blog.tomayac.com/2019/09/21/prefers-color-scheme-in-svg-favicons-for-dark-mode-icons/
- https://hiddedevries.nl/en/blog/2018-12-24-making-single-color-svg-icons-work-in-dark-mode (`currentColor` is used in the inline SVG icons, I think)

html/images/misc/openbeautyfacts-logo-no-tagline-dark-mode.svg
```diff
@@ -0,0 +1,356 @@
...
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   inkscape:export-ydpi="90"
+   inkscape:export-xdpi="90"
+   inkscape:export-filename="C:\Users\Stéphane\Desktop\Blogs\OpenBeautyFacts\openbeautyfacts-logo-en-178x150.png"
```
hangy, line 12: Private information like this should be removed before the file is added to git.

VaiTon:
> `style`

I don't think this would save much space and in the long run it's more difficult editing these kind of svg in inkscape/illustrator.

VaiTon:
> 
> 
> Can we minify this? (There's lossless minification that still allows full editing but doesn't include not-so-useful metadata.)

Don't we already do this in gulp? Would you prefer a pure SVG?

hangy:
>> `style
>
> I don't think this would save much space and in the long run it's more difficult editing these kind of svg in inkscape/illustrator.

I wasn't actually thinking about saving space, but rather lowering the translation and deployment complexity. If there's only one logo for a language, then that should be easier to manage and include in the website.

>> Can we minify this? (There's lossless minification that still allows full editing but doesn't include not-so-useful metadata.)
>
> Don't we already do this in gulp? Would you prefer a pure SVG?

Right now, `svgmin` is only used for the icons. Files like the OFF logo are used as-is. 😔

VaiTon:
> Right now, `svgmin` is only used for the icons. Files like the OFF logo are used as-is. 😔

What if we move original svg and make gulp pass them through `svgmin`?

hangy:
> What if we move original svg and make gulp pass them through `svgmin`?

Yes, that should be useful. But to be honest, I'd like to avoid mixing «dist» directories and files that should be in the git repository. We've had a few missed files (in the case of a matching .gitignore entry) in commits, and erroneously committed generated files.

Another thought I had about this, is if we could automatically generate the ‘dark mode’ styles using gulp, to avoid keeping the translated logos in sync manually. What do you think about it?

VaiTon:
> Another thought I had about this, is if we could automatically generate the ‘dark mode’ styles using gulp, to avoid keeping the translated logos in sync manually. What do you think about it?

This particular logo has nothing we can translate. Which one are you talking about, can we move the discussion to a new issue?

> I'd like to avoid mixing «dist» directories and files that should be in the git repository.

What if we move all files to a folder, like "src" and then build to the current folder.

hangy:
> > Another thought I had about this, is if we could automatically generate the ‘dark mode’ styles using gulp, to avoid keeping the translated logos in sync manually. What do you think about it?
> 
> This particular logo has nothing we can translate. Which one are you talking about, can we move the discussion to a new issue?

Sure. 👍 I think, I was mainly thinking about the logo variants that do have a translatable tagline and might need a dark mode version, too.

> > I'd like to avoid mixing «dist» directories and files that should be in the git repository.
> 
> What if we move all files to a folder, like "src" and then build to the current folder.

Yup, or we could update the logo URLs/translations to use minified versions from the `dist` folder. I think the latter might be more in-line of how the deployment is done right now. cc @stephanegigandet?

html/images/misc/openbeautyfacts-logo-no-tagline-dark-mode.svg
teolemon, line 12:
```suggestion
   inkscape:export-filename="openbeautyfacts-logo-en-178x150.png"
```

teolemon: @hangy I removed it.

PR #3701, opened 2020-06-22
OBF body parts taxonomy fix

teolemon:
Shameful I forgot that part (OBF body parts)


teolemon:
I don't know how much we want to get into details. On the mobile app, I would adopt a discreet approach.
It would actually even make sense to have a distinction between sexes (shampoos that respect vaginal flora and things like that).

Issue #3702, opened 2020-06-22
Add search sort key to order results based on popularity, completeness etc. for personal search project


stephanegigandet:
See https://wiki.openfoodfacts.org/Server-side_product_indexing_and_search_-_Functional_Specs
and https://wiki.openfoodfacts.org/Server-side_product_indexing_and_search_-_Technical_Specs

PR #3717, opened 2020-06-23
add search_key, bug #3702

stephanegigandet:
This is a new search order key, for the Personal Search project.
Please see #3702 for details.

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -96,7 +96,7 @@ BEGIN
 		&compute_languages
 		&compute_changes_diff_text
 		&compute_data_sources
-		&compute_sort_key
+		&compute_sort_keys
 
 		&add_back_field_values_removed_by_user
 
@@ -624,7 +624,7 @@ sub change_product_server_or_code($$$) {
 }
 
 
-=head2 compute_sort_key ( $product_ref )
+=head2 compute_sort_keys ( $product_ref )
 
 Compute sort keys that are stored in the MongoDB database and used to order results of queries.
 
@@ -642,7 +642,7 @@ Used for the Personal Search project to provide generic search results that apps
 
 =cut
 
-sub compute_sort_key($) {
+sub compute_sort_keys($) {
 
 	my $product_ref = shift;
 
@@ -876,7 +876,7 @@ sub store_product($$) {
 
 	compute_data_sources($product_ref);
 
-	compute_sort_key($product_ref);
+	compute_sort_keys($product_ref);
 
 	if (not defined $product_ref->{_id}) {
 		$product_ref->{_id} = $product_ref->{code} . ''; # treat id as string
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -646,7 +646,7 @@ sub compute_sort_keys($) {
 
 	my $product_ref = shift;
 
-	# put obsolete products last  		(add 200000000000 when products are not obsolete)
+	# put obsolete products last  		(substract 200000000000 when products are obsolete)
 	# then put complete products first	(add 100000000000 when products are complete)
 	# otherwise sort by last_modified_t (e.g.  1571384133)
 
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Products.pm
```diff
@@ -623,6 +623,25 @@ sub change_product_server_or_code($$$) {
...
+
+=head3 last_modified_t - date of last modification of the product page
+
+Used on the web site for facets pages, except the index page.
+
+=head3 sortkey - date of last modification, with obsolete products last and complete products first
+
+Used on the web site for index page.
+
+=head3 search_key - Popular and recent products
```
hangy, line 639: I'm not sure the variable/field name is good, considering it will be exposed in the database and possibly CSV. Right now, I think that something like `popularity_score` might represent the field content more precisely.

lib/ProductOpener/Products.pm
```diff
@@ -632,6 +651,47 @@ sub compute_sort_key($) {
...
+			}
+		}
+		# Keep only the latest year, and make the latest year count more than previous years
+		if (defined $latest_year) {
+			$search_key += $latest_year * 1000000 * 1000 - $years{$latest_year} * 1000;
+		}
+	}
+	
+	# unique_scans_n : number of unique scans for the last year processed by scanbot.pl
+	(defined $product_ref->{unique_scans_n}) and $search_key += $product_ref->{unique_scans_n};
```
hangy, line 685: Is this the same as this?
```
if (defined $product_ref->{unique_scans_n}) {
    $search_key += $product_ref->{unique_scans_n};
}
```

lib/ProductOpener/Products.pm
stephanegigandet: I was thinking a more generic name would allow us to change the formula later. Today it's mostly popularity (as in scans), but later we could include other things (like for instance availability in stores if we get data from stores someday). The current formula in the code is in fact quite experimental: it's just one possible way, with the data we have today. Then I want to populate it in the full database, see what kind of results we get for different searches etc. and possibly adapt the formula.

lib/ProductOpener/Products.pm
stephanegigandet: Yes, good catch, thanks.

lib/ProductOpener/Products.pm
stephanegigandet: ok I renamed it to popularity_key . if we change the formula too much, we can always give it a new name / compute a different key.  I prefer "key" instead of "score" because the formula I made tries to achieve a certain order, but there's absolutely no scale for the key.

lib/ProductOpener/Products.pm
```diff
@@ -623,6 +623,25 @@ sub change_product_server_or_code($$$) {
...
+
+Used on the web site for index page.
+
+=head3 popularity_key - Popular and recent products
+
+Used for the Personal Search project to provide generic search results that apps can personnalize later.
+
+=cut
+
 sub compute_sort_key($) {
```
VaiTon, line 645: Can we rename to `compute_sort_keys`?

lib/ProductOpener/Products.pm
```diff
@@ -632,6 +651,49 @@ sub compute_sort_key($) {
...
+	if (defined $product_ref->{last_image_t}) {
+
+		my $age = int((time() - $product_ref->{last_image_t}) / (86400 * 30));	# in months
+		if ($age < 12) {
+			$popularity_key += 12 - $age;
+		}
+	}
 
 	if ((not defined $product_ref->{obsolete}) or (not $product_ref->{obsolete})) {
 		$sortkey += 200000000000;
```
VaiTon, line 699: I don't really like these really big numbers, expecially as mongo index 😕. Could subtracting be an idea? Does mongo deals with negative indexes?

lib/ProductOpener/Products.pm
```diff
@@ -641,6 +703,7 @@ sub compute_sort_key($) {
 	}
```
VaiTon, line 704:
```suggestion

# Add 0 so we are sure the key is saved as int
```

lib/ProductOpener/Products.pm
stephanegigandet: sure

lib/ProductOpener/Products.pm
stephanegigandet: I think negative numbers should work. I do not know if there's an issue with big numbers though.

lib/ProductOpener/Products.pm
stephanegigandet: done

lib/ProductOpener/Products.pm
stephanegigandet: I changed the keys to make obsolete products negative. We still have big numbers, but I don't really think that's an issue for mongo, it seems to work quite well. We can revisit the keys later, we will probably have to change them anyway.

PR #3723, opened 2020-06-24
Adding an OBF version of the Crowdsourcing illustration

teolemon:
Crowdsourcing icon for Open Beauty Facts


html/images/svg/crowdsourcing-obf-round-smartphones.svg
```diff
@@ -0,0 +1,432 @@
...
+             xlink:href="#path-1"
+             id="use18"
+             x="0"
+             y="0"
+             width="100%"
+             height="100%" />
+        </mask>
+        <g
+           id="g21" />
+        <image
```
stephanegigandet, line 143: That's not good, we shouldn't have a bitmap image in a SVG...
That's the bottle cap, you can find a vector version of it in the OFF logo.

PR #3740, opened 2020-06-25
Update states.txt (Fr-->À?)

alexouille123:


taxonomies/states.txt
```diff
@@ -94,14 +94,14 @@ zh:完成
...
 en:To be completed
-ca:Pendent de completar
+ca:Pendent de completar
 cs:Před dokončením
 de:Auszufüllen
 el:Να συμπληρωθεί
 es:A completar
-fi:Tarvitsee täydentää
+fi:Tarvitsee täydentää
 fr:A compléter
```
alexouille123, line 105: If so, I have to do it here, too.

alexouille123, line 105: Also here, no ?

PR #3747, opened 2020-06-26
Add countries to the Store taxonomy

teolemon:
Added country property to stores. Note that more countries need to be added for Carrefour and other multi-country store, but the property does not bear any consequences yet

### DIFF
--- a/taxonomies/stores.txt
+++ b/taxonomies/stores.txt
@@ -1,244 +1,244 @@
 # French stores
 fr:Auchan
-country:en:France
+country:en: en:France
 
 fr:A 2 Pas
-country:en:France
+country:en: en:France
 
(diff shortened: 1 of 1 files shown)


taxonomies/stores.txt
```diff
@@ -1,180 +1,244 @@
 # French stores
 fr:Auchan
+country:en:France
```
stephanegigandet, line 3: Should be country:en: en:france

teolemon: fixed

stephanegigandet:
We do not know yet how we will need to structure the stores taxonomy (just like brands) to make it work, so we can commit as-is, but it would probably be good to wait before doing more work on this.

PR #3755, opened 2020-06-26
Preliminary attempt to categorize labels into clusters

teolemon:
Preliminary attempt to categorize labels into clusters so that we can display them into wide categories in the website and the app.

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -759,6 +759,7 @@ he:תווית איכות תאילנדית
 
 <en:Thai quality label
 en:Thailand Diversity & Refinement
+origins:en: en:Thailand
 
 <en:Thai quality label
 en:Thai fish sauce quality label
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -175,6 +175,7 @@ hu:Szabad tartású
 nl:Vrije uitloop
 pt:Criação em liberdade
 ru:свободный выгул, открытый воздух
+label_categories:en: en:Animal welfare
 
 en:French eggs, Eggs from France
 ca:Ous francesos, Ous de França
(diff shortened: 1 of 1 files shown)


teolemon:
I would put `en:Environment` first. Perhaps `en:Origins` second. But to be honest, I was thinking geographic origins.

taxonomies/labels.txt
```diff
@@ -175,6 +175,7 @@ hu:Szabad tartású
 nl:Vrije uitloop
 pt:Criação em liberdade
 ru:свободный выгул, открытый воздух
+label_categories:en: en:Animal welfare
```
hangy, line 178: `en:en:` seems redundant

taxonomies/labels.txt
```diff
@@ -573,207 +579,6 @@ de:FSC Recycling
 fi:FSC-kierrätys
 fr:FSC Recyclé, FSC Recyclage
 
-<en:FSC
```
hangy, line 576: Why do you want to remove the FSC labels and how does that change relate to clustering?

taxonomies/labels.txt
teolemon: it's the way it is for multilingual taxonomies embedded into other multilingual taxonomies.

taxonomies/labels.txt
teolemon: unrelated. They were moved to packager codes, and this was a really tiny subset of all the FSC codes.

taxonomies/labels.txt
stephanegigandet: That's the way it currently works when a property value is an entry in another taxonomy.
e.g. for additives, we have:
additives_classes:en: en:colour

That means that the English value for the additives_classes property is the entry en:colour.

The properties in taxonomy must have a language (and the value can be different based on that language). 

We should do this though:

label_categories:en: en:animal-welfare

teolemon: I'm confused @stephanegigandet in other occurrences I could swear you've asked me to put the real value and not the tag. Should I create a taxonomy of label clusters ?

taxonomies/labels.txt
teolemon: @hangy should I remove them in a separate PR or are you ok with getting rid of them in this one ?

PR #3759, opened 2020-06-26
Apply perl critic automatically using GitHub actions

teolemon:
Applied at level 5, gentle. For other levels, we'd need to update the source action.


VaiTon:
I don't think changes are then commited to the repo.

hangy:
We already run perlcritic through `prove` when running all other tests: https://github.com/openfoodfacts/openfoodfacts-server/blob/master/t/perlcritic.t The settings are managed by https://github.com/openfoodfacts/openfoodfacts-server/blob/master/.perlcriticrc

teolemon:
Ok. closing for now. As a pro: The GitHub action promises to make the Critic results much more visible. As a con: by default, it's only set to Level 5

hangy:
Can it be modified to use the rc- file? If it offers better visibility of errors, we could remove the test and move it to a GitHub action.

Ban3:
> Can it be modified to use the rc- file?

I glanced at the [source](https://github.com/Difegue/action-perlcritic), and it doesn't support the rc-file argument. [Looks like it just feeds the directories as positional arguments to perlcritic](https://github.com/Difegue/action-perlcritic/blob/master/entrypoint.sh). Should be easy enough to fork & modify, if that's what's desired.

hangy:
> 
> 
> > Can it be modified to use the rc- file?
> 
> I glanced at the [source](https://github.com/Difegue/action-perlcritic), and it doesn't support the rc-file argument. [Looks like it just feeds the directories as positional arguments to perlcritic](https://github.com/Difegue/action-perlcritic/blob/master/entrypoint.sh). Should be easy enough to fork & modify, if that's what's desired.

I think the args are just passed through from the YAML declaration to [perlcritic](https://metacpan.org/pod/perlcritic). Basically, all commandline options could be passed manually. However, since the program seems to be run in the repo directory, and `perlcritic` searches for the `.perlcriticrc` in the current directory by default, it probably just uses the RC file already. (That would also make Difegue/action-perlcritic#1 redundant, since the level can be set in the RC files.) We should just consider removing the the perlcritic unit test from the normal CI build, so that it's not executed twice.

teolemon:
@hangy is the mode harsh ?

hangy:
> @hangy is the mode harsh ?

No, we wouldn't pass anything with a lower severity than 5: https://github.com/hangy/openfoodfacts-server/runs/887003513?check_suite_focus=true#step:4:8

PR #3761, opened 2020-06-27
Start adding properties to brands

teolemon:
Start adding properties to brands for a future constraints system

### DIFF
--- a/taxonomies/brands.txt
+++ b/taxonomies/brands.txt
@@ -637,319 +637,319 @@ fr:Marque Repère
 
 <fr:Marque Repère
 fr:Autour du Dessert
-brand_categories: fr:Gâteaux et pâtisseries surgelés
+categories: fr:Gâteaux et pâtisseries surgelés
 
 <fr:Marque Repère
 fr:Bio Village
-brand_label:Bio
+labels:en: en:organic
 
 <fr:Marque Repère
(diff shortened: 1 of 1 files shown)


taxonomies/brands.txt
```diff
@@ -2571,7 +2574,7 @@ es:Campos de Jumilla
 
 Coop. Bio Südtirol
 <es:VOG Coop. Soc. Agr.
-es:Bio Südtirol, Bio Sudtirol
+es:Südtirol, Bio Sudtirol
```
teolemon, line 2577:
```suggestion
es:Bio Südtirol, Bio Sudtirol
```

taxonomies/brands.txt
```diff
@@ -2162,6 +2162,7 @@ es:Carrefour Baby
 
 <es:Carrefour S.A.
 es:Carrefour Bio
+label:en: en:Organic
```
stephanegigandet, line 2165: It will be easier with the exact facets ids we use (in plurals)
```suggestion
labels:en: en:organic
```

stephanegigandet:
Looks good but at this point we don't know how the brands taxonomy will work, so it would probably be good to wait before doing too much work on it.

teolemon:
Fixed, and also normalized other constraints that were already there

Issue #3779, opened 2020-06-26
Ensure all images have a JSON


teolemon:
I have products where I have no JSON on all the images (this can be linked to an upload during a high peak of contribution traffic where Cloud Vision can no longer absorb the peak)

https://static.openfoodfacts.org/images/products/326/385/186/3614/3.json
https://static.openfoodfacts.org/images/products/326/385/186/3614/3.jpg
https://world.openfoodfacts.org/product/3263851863614/quinoa-rouge-franprix

We should have a CRON script ensuring we have a JSON for each image on the disk, and regenerating if needed.

teolemon:
The script is here, we should ensure it's run periodically.

PR #3780, opened 2020-07-02
Add script to generate missing OCR JSON (#3779)

raphael0202:
Add a new Python script to regenerate corrupted or missing OCR JSON.
This script has already run successfully in production.
Solve issue #3779.


raphael0202:
Indeed it only generates JSON files for images with digits (1.jpg, 2.jpg,...)

raphael0202:
I've updated the instruction at the top of the script to specify it should be run as off user.

Issue #3793, opened 2020-07-04
Prevent products with invalid codes (less than 8 digits) to be created


stephanegigandet:
Very bad things can happen when products are created with short codes (especially 3 digits), due to the directory structure we use to store products.

This change is to prevent those products to be created.

stephanegigandet:
e.g. when a product "123" is created, it is stored in products/123/ which is also the root of products like /123/456/ etc. 
When the "123" products gets moved to another barcode by a moderator, all the paths of the other products are broken.

stephanegigandet:
I moved / merged back the 123 products files and images to /123/

The original + moved files before merging are also stored in /srv2/off/bug-123 just in case.

We seem to have also moved the /951/ products somewhere else.

PR #3795, opened 2020-07-04
prevent short barcodes, bug #3793

stephanegigandet:
This is to prevent entry of barcodes with less than 8 or more than 24 digits.

### DIFF
--- a/cgi/product_jqm_multilingual.pl
+++ b/cgi/product_jqm_multilingual.pl
@@ -72,14 +72,15 @@ if ($code eq "new") {
 	$response{code} = $code . "";	# Make sure the code is returned as a string
 }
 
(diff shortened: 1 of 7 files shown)


stephanegigandet:
It will make sure that #3793 does not happen again, but we still need to look for the misplaced products (and hopefully somehow merge them back).

stephanegigandet:
> Some Spanish retailers have very long barcodes

Long equals more than 24 digits?

hangy:
> > Some Spanish retailers have very long barcodes
> 
> Long equals more than 24 digits?

Also, how interested are we in non-GTIN barcodes in general?

IMHO, fix this bug first, then see about unusual barcode formats.

PR #3797, opened 2020-07-05
Some Belgian ingredient parsing additions

aleene:
**Description:**
Some dutch ingredient parsing additions


taxonomies/additives_classes.txt
```diff
@@ -364,7 +364,7 @@ da: Emulgator, Emulgatorer, Emulgeringsmiddel
 el: Γαλακτωματοποιητής
 es: Emulgente, Emulsificante, emulsionante, agente dispersante, agente enturbiador, agentes enturbiadores, agente tensoactivo, corrector de la densidad, estabilizadores de una suspensión, estabilizador de una suspensión, inhibidor de la cristalización, plastificante
 fi: Emulgointiaine, Emulgointiainetta, Emulgointiaineet, Emulgointiaineita
-fr: Émulsifiant, agent de dispersion, agent de surface, agent de suspension, agent d'ajustement de la densité, inhibiteur de cristallisation, nébulisant, plastifiant
+fr: Émulsifiant, émulsiﬁant, agent de dispersion, agent de surface, agent de suspension, agent d'ajustement de la densité, inhibiteur de cristallisation, nébulisant, plastifiant
```
stephanegigandet, line 367: émulsiﬁant with a single fi character is incorrect, we will fix it through spellchecking.

lib/ProductOpener/Ingredients.pm
```diff
@@ -2535,6 +2536,7 @@ fr => [
 '(produit )?(a|à) protéger de ', # humidité, chaleur, lumière etc.
 '(produit )?conditionn(e|é) sous atmosph(e|è)re protectrice',
 'N(o|ò)us vous conseillons',
+'Sans conservateur',
```
Ban3, line 2540: Wouldn't this be better in labels?
https://github.com/openfoodfacts/openfoodfacts-server/blob/52d473d71612f1ad27441b8eeefcd39d7c2b7b5d/taxonomies/labels.txt#L2532

aleene: No, as it is used to parse the ingredient lists. And it is only used to find the end of the ingredients. The labels is what can be found elsewhere on the packaging.

However if it is mentioned in the ingredient list, it could be added to the labels.

Ban3: Yeah, but labels at the end of the list will be picked up as labels by the parser. That's why I'm asking.

aleene: OK, next question, where can I specify that in the ingredients.pm

Ban3: If it's a standalone label like what I'm assuming `Sans conservateur` to be (e.g `ingredient1, ingredient2. Sans conservateur`), nothing needs to be done in Ingredients.pm. It only needs to be defined in labels.txt. The detection of those labels happens here: https://github.com/openfoodfacts/openfoodfacts-server/blob/d06d989211dd232e16ba54cd1d468cfe4f5de8e1/lib/ProductOpener/Ingredients.pm#L1245-L1257

The ones that are part of the ingredient or `*`s needs to be included in this list to be detected: https://github.com/openfoodfacts/openfoodfacts-server/blob/d06d989211dd232e16ba54cd1d468cfe4f5de8e1/lib/ProductOpener/Ingredients.pm#L381-L385

That's at least my limited understanding of the analyzer. @stephanegigandet  might want to correct me here if I'm wrong.

PR #3798, opened 2020-07-05
Commit label logos

teolemon:
Commit label logos that are in production but not on the repository.


html/images/lang/en/labels/real-california-milk-90x90.png
```diff
@@ -0,0 +1,278 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr" class="client-nojs">
+<head>
+<title>File:Real-california-milk-90x90.png - Open Food Facts EN</title>
```
hangy, line 4: This seems to be HTML from Wikipedia.

teolemon: fixed. the proper logo was already commited.

teolemon:
If I move them from fr to en, it will break display on fr.off.org
see: https://github.com/openfoodfacts/openfoodfacts-server/issues/3821

hangy:
At least for these two logos, the English version should be used for other languages.

teolemon:
@hangy
- I added FSC to labels/en/
- Max Havelaar, there's already a SVG and PNG copy in labels/en/

PR #3800, opened 2020-07-05
Add Australian Health Star rating label images (PNG)

teolemon:


teolemon:
Should I loose quality by shrinking or lie about the size ?

hangy:
> Should I loose quality by shrinking or lie about the size ?

Ideally, we'd have an SVG version. 😉 However, since that doesn't seem to be available, resizing the file manually should be better. The website will display it at the 90px height anyways. Using a specific scaling algorithm instead of letting the browser choose one sounds more reasonable, IMHO.

PR #3817, opened 2020-07-08
test if merges can be done when tests fail

stephanegigandet:
It looks like it's possible to merge branches when some tests fail. This PR is to test it.


stephanegigandet:
Cannot be merged as test failed. Good. :)

Issue #3840, opened 2020-07-13
Separate ciqual categories and agribalyse categories in categories taxonomy


stephanegigandet:
Currently:

<en:Snacks and desserts for babies
en:Cocoa powder for baby beverage
fr:Poudre cacaotée pour bébé
ciqual_food_code:en:42501
ciqual_food_name:en:Cocoa powder for baby beverage
ciqual_food_name:fr:Poudre cacaotée pour bébé

If there is a corresponding Agribalyse category, add:

agribalyse_ciqual_food_code:en:42501

Otherwise, add support for a proxy category:

agribalyse_proxy_food_code:en:43244
agribalyse_proxy_food_name:en:Cocoa powder

PR #3841, opened 2020-07-13
add new agribalyse_food_code property

stephanegigandet:
Start of #3840

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -108,7 +108,7 @@ BEGIN
 
 					&create_nutrients_level_taxonomy
 
-					&assign_category_properties_to_product
+					&assign_categories_properties_to_product
 					
 					&remove_insignificant_digits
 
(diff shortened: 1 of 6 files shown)


stephanegigandet:
> What is agribalyse_food_codes.csv ?

Just a list of the ciqual codes that are in Agribalyse, I added it.

lib/ProductOpener/Food.pm
```diff
@@ -5793,40 +5793,51 @@ This function was first designed to assign a CIQUAL category to products, based
...
+		my $property_name = $property;
+		$property_name =~ s/:en$//;
 
 		# Find the first category with a defined value for the property
 
 		if (defined $product_ref->{categories_tags}) {
 			foreach my $categoryid (reverse @{$product_ref->{categories_tags}}) {
 				if ((defined $properties{categories}{$categoryid}) and (defined $properties{categories}{$categoryid}{$property})) {
-					$product_ref->{category_properties}{$property} = $properties{categories}{$categoryid}{$property};
+					$product_ref->{categories_properties}{$property} = $properties{categories}{$categoryid}{$property};
```
hangy, line 5821: Is this a breaking change? I've found this property set in existing products in the public API.

stephanegigandet: It's something I added a couple weeks ago, it is a work in progress and subject to change, it's not something documented and it is not used by anyone.
I'll update all products to remove the field with the old name.

PR #3845, opened 2020-07-14
display_new() templatization

areeshatariq:
**Description:**

HTML Templatization of display_new() using Template::Toolkit
A template has been created for pages structure: header, footer, column


lib/ProductOpener/Display.pm
```diff
@@ -6363,145 +6320,21 @@ HTML
...
-		$image_banner = "";
-	}
-
-	if ($image_banner ne "") {
+	my $link = lang("donate_link");
+	my $image;
+	my $utm;
+	my @banners = qw(independent personal research);
+	my $banner = $banners[time() % @banners];
+	$image = "/images/banners/donate/donate-banner.$banner.$lc.800x150.svg";
```
stephanegigandet, line 6328: For some languages, this file will not exist. We need to check if the file exists (with -e "$www_root//images/banners/donate/donate-banner.$banner.$lc.800x150.svg")  --> if it does not exist, we need to use the English banner.

templates/display_new.tt.html
```diff
@@ -0,0 +1,272 @@
+<!doctype html>
+<html class="no-js" lang="[% language %]" data-serverdomain="[% server_domain %]">
+<head>
+    <meta charset="utf-8">
+    <title>[% title %]</title>
+    [% meta_description %]
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta property="fb:app_id" content="219331381518041">
```
hangy, line 8: Could you move this to a config setting, like the for example the `twitter_account` value?

lib/ProductOpener/Display.pm
```diff
@@ -6851,6 +6446,7 @@ HTML
 	# init javascript code
 
 	$html =~ s/<initjs>/$initjs/;
```
hangy, line 6448: `$html` is empty at this point, so that all of the following fix-ups don't do anything. IMHO: Move https://github.com/openfoodfacts/openfoodfacts-server/blob/8ad091812d7e88d2e52c8e8e0def72e4f2609a6c/lib/ProductOpener/Display.pm#L6208 to https://github.com/openfoodfacts/openfoodfacts-server/blob/8ad091812d7e88d2e52c8e8e0def72e4f2609a6c/lib/ProductOpener/Display.pm#L6479 and just use `$html` instead of `$temp_html`, and then run https://github.com/openfoodfacts/openfoodfacts-server/blob/8ad091812d7e88d2e52c8e8e0def72e4f2609a6c/lib/ProductOpener/Display.pm#L6419-L6448 after `$tt->process`. This should fix issues like not all images being requested from the `static` subdomain.

templates/nutrition_facts_table.tt.html
```diff
@@ -47,13 +47,13 @@
 [% FOREACH table IN tables %]
 
 	<table id="[% table.id %]_data_table" class="data_table">
-
+    <caption>Nutrition Facta Table</caption>
```
stephanegigandet, line 50: small typo: Facta instead of Facts. But this should be internationalized, using lang("nutrition_data")

templates/display_new.tt.html
```diff
@@ -0,0 +1,272 @@
...
+			<li class="show-for-xlarge-up"><a href="/cgi/search.pl">[% display_icon('add') %] [% lang('advanced_search') %]</span></a></li>
+			<li class="show-for-large-only"><a href="/cgi/search.pl?graph=1" title="[% lang('graphs_and_maps') %]">[% display_icon('bar_chart') %]</a></li>
+			<li class="show-for-xlarge-up"><a href="/cgi/search.pl?graph=1">[% display_icon('bar_chart') %] [% lang('graphs_and_maps') %]</span></a></li>
+			[% IF !(server_options_producers_platform) %]
+                <li class="show-for-large-up divider"></li>
+                <li><a href="[% lang('menu_discover_link') %]">[% lang('menu_discover') %]</a></li>
+                <li><a href="[% lang('menu_contribute_link') %]">[% lang('menu_contribute') %]</a></li>
+                <li class="show-for-large"><a href="/[% lang('get_the_app_link') %]" title="[% lang('get_the_app') %]" class="button success">[% display_icon('phone_android') %]</a></li>
+                <li class="show-for-xlarge-up"><a href="/[% lang('get_the_app_link') %]" class="button success">[% display_icon('phone_android') %] [% lang('get_the_app') %]</a></li>
+            [% END %]
```
hangy, line 80: Could you fix the indentation for this block? I think the first list elements are indented too much, and the `% IF` statement was actually a bit difficult to discover.

PR #3846, opened 2020-07-14
HTML Templatization of Product page

areeshatariq:
**Description:**
HTML Templatization of Product page using Template::Toolkit
display_product() has been templatized


templates/display_product.tt.html
```diff
@@ -0,0 +1,243 @@
...
+
+    [% IF has_tag == 'categories-en:alcoholic-beverages' %]
+        <p class="panel callout">
+            [% lang('alcohol_warning') %]
+        </p>
+    [% END %]
+[% END %]
+
+<!-- other fields -->
+[% IF other_fields != "" %]
```
hangy, line 191: https://github.com/openfoodfacts/openfoodfacts-server/blob/65b6c9cc6d0e6c080c0e05b909487d278c73f8f5/lib/ProductOpener/Display.pm#L7920-L7922

I think this is only set if it's not empty. How does TT handle the case of the property probably being `undef`?

stephanegigandet:
I'm testing on the dev server, one issue is that the same image is displayed for the front of the product, the ingredients, and the nutrition facts.
e.g. https://world.openfoodfacts.dev/product/2000000116683 (I may deploy other branches on the dev server)

image

areeshatariq:
> I'm testing on the dev server, one issue is that the same image is displayed for the front of the product, the ingredients, and the nutrition facts.

I'll look into this

templates/display_product.tt.html
areeshatariq: > I think this is only set if it's not empty. How does TT handle the case of the property probably being `undef`?

'undef' fields won't be displayed

hangy: OK, cool. I was wondering if that might cause any kind of error, but in that case, TT is apparently clever enough. 🙂 Thanks!

lib/ProductOpener/Display.pm
```diff
@@ -7480,124 +7455,42 @@ HTML
...
-<div property=\"gr:hasEAN_UCC-13\" content=\"$code\" datatype=\"xsd:string\"></div>\n";
+		$template_data_ref->{html_upc} = $html_upc;
 	}
 
-
 	# obsolete product
 
 	if ((defined $product_ref->{obsolete}) and ($product_ref->{obsolete})) {
-
+		$template_data_ref->{product_ref_obsolete} = $product_ref->{obsolete};
```
hangy, line 7464: `product_is_obsolete` might be a better key name. `$product_ref` is just named like this since it's a reference variable.

templates/display_product.tt.html
```diff
@@ -0,0 +1,243 @@
...
+            [% IF no_nutrition_data == 'on' %]
+                <div class='panel callout'>
+                    [% lang("no_nutrition_data") %]
+                </div>
+            [% END %]
+            [% display_nutrition_table %]
+        </div>
+    
+        <div class="show-for-large-up large-4 xlarge-4 xxlarge-4 columns" style="padding-left:0">
+            $html_image
```
hangy, line 179: Should this be a TT variable instead?

areeshatariq: Oops, yes!

stephanegigandet:
(closing this pull request and making a new one, for some reason github shows changes that are in fact already merged, and the test fails)

PR #3837, opened 2020-07-13
Some more french ingredients

aleene:
**Description:**
Still going through the unknown Belgian ingredients

PR #3862, opened 2020-07-16
special case for 'fer élémentaire', fixes #3837

stephanegigandet:
not pretty but I don't have a better solution right now


lib/ProductOpener/Ingredients.pm
```diff
@@ -3497,7 +3497,10 @@ sub preparse_ingredients_text($$) {
 			$text =~ s/($prefixregexp)\s?(:|\(|\[)\s?($suffixregexp)\b(\s?(\)|\]))/normalize_enumeration($product_lc,$1,$5)/ieg;
 
 			# Huiles végétales de palme, de colza et de tournesol
+			# Carbonate de magnésium, fer élémentaire -> should not trigger carbonate de fer élémentaire.
```
VaiTon, line 3500:
```suggestion
			# Carbonate de magnésium, fer élémentaire -> should not trigger carbonate de fer élémentaire.
			# TODO 18/07/2020 remove when we have a better solution
```

PR #3869, opened 2020-07-16
Rename haute-valeur-environnementale-90x90.svg + Halal Labels

teolemon:
Rename haute-valeur-environnementale-90x90.svg to haute-valeur-environnementale.90x90.svg


teolemon:
Re-requesting review, since I added halal labels

stephanegigandet:
The way the images work:

html/images/lang/en/labels/[English name of the label, if it exists in the taxonomy].90x90.png
html/images/lang/fr/labels/[French name of the label, if it exists in the taxonomy].90x90.png

So we should not have html/images/lang/en/labels/controle-de-la-mosquee-d-evry-courcouronnes.90x90.png for instance

teolemon:
Well, if we do the way you suggest, we don't get the labels displayed on the English version @stephanegigandet

teolemon:
EG: AB-Bio that I had to copy paste in prod

PR #3870, opened 2020-07-16
label categories taxonomy

teolemon:
Here's a stub taxonomy for the new Label categories property of the label taxonomy


stephanegigandet:
can you rename it to labels_categories.txt ?

teolemon:
done @stephanegigandet

Issue #3852, opened 2020-07-15
Refresh and automate French packager codes


teolemon:
- [ ] Fetch files
```
    # TODO concatenate while adding an extra column to indicate establishment type (variable name)
    # "Numero de département","Numéro agrément/Approval number","SIRET","Raison SOCIALE - Enseigne commerciale/Name","Adresse/Adress","Code postal/Postal 
    base_url = 'https://fichiers-publics.agriculture.gouv.fr/dgal/ListesOfficielles/'
    0 = 'SSA1_ACTIV_GEN.txt'
    1 = 'SSA1_VIAN_ONG_DOM.txt'
    2 = 'SSA1_VIAN_COL_LAGO.txt'
    3 = 'SSA1_VIAN_GIB_ELEV.txt'
    4 = 'SSA1_VIAN_GIB_SAUV.txt'
    5 = 'SSA1_VIAND_HACHE_VSM.txt'
    6 = 'SSA4_AGSANPROBASEVDE_PRV.txt'
    7 = 'SSA4B_AS_CE_PRODCOQUI_COV.txt'
    8 = 'SSA4B_AS_CE_PRODPECHE_COV.txt'
    9 = 'SSA1_LAIT.txt'
    10 = 'SSA1_OEUF.txt'
    11 = 'SSA1_GREN_ESCARG.txt'
    12 = 'SSA4_AGSANGREXPR_PRV.txt'
    13 = 'SSA4_AGR_ESVEBO_PRV.txt'
    14 = 'SSA4_AGSANGELAT_PRV.txt'
    15 = 'SSA4_AGSANCOLL_PRV.txt'
    16 = 'SSA_PROD_RAFF.txt'
    # RC = 'SSA4_ASCCC_PRV.txt'
```
- [ ] Add section name to each line
- [ ] Concatenate ( cf Spain example at: https://github.com/openfoodfacts/openfoodfacts-server/blob/master/scripts/packager-codes/portugal-concatenate-csv-sections.py)
- [ ] geocode

teolemon:
Thanks @Ban3 , closing this one thanks to you 👍

PR #3884, opened 2020-07-18
Refresh FR packager codes

Ban3:
**Description:**

Script to update FR packager codes and also the resulting CSV.

I used some modules that are not dependencies of ProductOpener for convenience:

- `Data::Table`: for tabular data processing.
- `Geo::Coder::Google`: for geocoding.
- `Future`, `Future::Utils`, `Net::Async::HTTP`, `IO::Async`: for async processing of the sections. Geocoding is awfully slow and this makes it a bit more bearable.

**Related issues and discussion:** #3852

### DIFF
--- a/scripts/packager-codes/fr-packagers-refresh.pl
+++ b/scripts/packager-codes/fr-packagers-refresh.pl
@@ -1,0 +1,209 @@
+#!/usr/bin/env -S perl -w
+
+# This file is part of Product Opener.
+#
+# Product Opener
+# Copyright (C) 2011-2020 Association Open Food Facts
+# Contact: contact@openfoodfacts.org
+# Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
+#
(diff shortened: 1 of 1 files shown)


Ban3:
[Here's also the 75 rows](https://gist.github.com/Ban3/900bea96adf5b22314d823deb398ba2d) that didn't get coordinates from the geocoder. They probably could be added to some of them manually.

Ban3:
There's at least one faulty line where the approval number has been split into multiple fields:
```
34,34,301,656,75197227400011,MICHELOT JULIEN,MAS CONCHYLICOLE DU BARROU RUE DES CORMORANS,34200,SETE,DC;PC,https://fichiers-publics.agriculture.gouv.fr/dgal/ListesOfficielles/SSA4B_AS_CE_PRODCOQUI_COV.txt,,
```
The issue is in the source data as well. Naturally it's split to just the right amount of columns so it will just slip through. I will fix this by hand.

PR #3889, opened 2020-07-19
Added Unsaturated fat

aleene:
**Description:**
Added `unsaturated fat`, which I encountered on a product.


aleene:
Shall I change that one as well?

PR #3897, opened 2020-07-19
User Form Templatization

areeshatariq:
**Description:**

HTML Templatization of User form using Template::Toolkit

### DIFF
--- a/cgi/user.pl
+++ b/cgi/user.pl
@@ -38,8 +38,27 @@ use URI::Escape::XS;
 use Storable qw/dclone/;
 use Log::Any qw($log);
 
+use Template;
+use Data::Dumper;
+
 my $type = param('type') || 'add';
 my $action = param('action') || 'display';
+
+
+# Initialize the Template module
+my $tt = Template->new({
+	INCLUDE_PATH => $data_root . '/templates',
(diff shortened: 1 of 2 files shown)


cgi/user.pl
```diff
@@ -38,9 +38,28 @@
 use Storable qw/dclone/;
 use Log::Any qw($log);
 
+use Template;
+use Data::Dumper;
```
hangy, line 42: Do we need it in here?

cgi/user.pl
```diff
@@ -38,9 +38,28 @@
...
+
 my $type = param('type') || 'add';
 my $action = param('action') || 'display';
 
+
+# Initialize the Template module
+my $tt = Template->new({
```
hangy, line 49: I noticed that the initialisation of `$tt` and `$template_data_ref` is now done several times in Display.pm and some .pl files in the same way. Would it make sense to introduce a new `ProductOpener::Template` module with methods that know how to do just that? It could reduce duplicated code and avoid errors.

stephanegigandet: @hangy : maybe we can just make the $tt in Display.pm an exported variable. Then it's initialize once at module startup, and that's it.

PR #3911, opened 2020-07-21
Make some image aware of dark mode

hangy:
Can be used in correlation with #3317 to support #2407.
image

### DIFF
--- a/html/images/svg/product-silhouette.svg
+++ b/html/images/svg/product-silhouette.svg
@@ -10,6 +10,14 @@
    viewBox="0 0 100 100"
    height="100"
    width="100">
+ <style>
+   #rect6554{fill:#d7e3f4;}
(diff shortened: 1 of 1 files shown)


hangy:
The product silhouette and English OFF icon already work quite well. I'll look into updating the other icons, and fixing the translations to include SVGs where possible.

teolemon:
How about keeping both versions with different names ? Although the color can be changed programmatically, having light and dark versions would also be useful for marketing purposes.

CloCkWeRX:
@hangy did you feel like this could be mergable in its current form? Any chance you'd adopt or improve the android SVG logo on https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#issuecomment-1073799200 into this PR?

PR #3912, opened 2020-07-21
override csv separator for FR

stephanegigandet:
The French file is generated with , instead of ; as separator.

This is an ugly fix, at some point we will have to improve the script so that it uses Text::CSV to load csv files.


stephanegigandet:
@Ban3 : ok, then maybe it would be better to regenerate with ;
Thank you!

PR #3915, opened 2020-07-22
fix: updated path for ean8 and short barcodes (padding with zeroes)

stephanegigandet:
Please see #3818 for details


lib/ProductOpener/Products.pm
```diff
@@ -75,6 +75,7 @@ BEGIN
 		&server_for_product_id
 		&data_root_for_product_id
 		&www_root_for_product_id
+		&split_code
```
hangy, line 78: It's already exported here: https://github.com/openfoodfacts/openfoodfacts-server/blob/89119c8f92ee67120b95f1037c8ad453c1929a55/lib/ProductOpener/Products.pm#L73

stephanegigandet: Thanks

stephanegigandet:
> Do the images need to be moved, too?

Yes, I forgot to mark the PR as a draft, I have not finished.

teolemon:
@stephanegigandet merge conflict on the test before a rebase is possible

stephanegigandet:
I will make a new PR.

PR #3952, opened 2020-07-27
Add a CSS class for weblinks

hangy:
**Related issues and discussion:** https://github.com/openfoodfacts/openfoodfacts-server/pull/3317#issuecomment-664549326


hangy:
@Jen-Lopez I think this should actually be enough to fix the issue of not being able to style this floating box. You can also just do the change in your branch instead. 🙂

Issue #938, opened 2017-11-10
Prevent crawlers from going to deep in facets


stephanegigandet:
I'm thinking to make those links no follows:
- 2nd level drilldown by facets links in the product list resulting of a 1st drilldown by facet

The list of facet values in the 2nd drilldown are already no follow, e.g. in page /packaging/surgele/additives:

<tr><td><a href="/packaging/surgele/additive/e471-mono-and-diglycerides-of-fatty-acids" class="tag well_known" rel="nofollow">E471 - Mono- and diglycerides of fatty acids</a></td>

teolemon:
so brand XXX additive YYY would not be discovered ?

stephanegigandet:
no, those are already nofollow, brand XXX additives should not be crawled either.

we just get too many pages crawled and recrawled for nothing. and those are expensive pages.

teolemon:
fair enough.

PR #3956, opened 2020-07-28
Fix for broken blue "Explore by" button link

### LINKED ISSUE #3919: The blue "Explore by" button is broken
The links have `https://us.openfoodfacts.org/misc/nutriscore-computed/nutrition-grades%20rel=`
As a result they are broken.

areeshatariq:
**Description:**

Fixed broken link of 'Explore by' button. 

**Related issues and discussion:** fixes #3919


VaiTon:
Why was that there in the first place?

hangy:
> Why was that there in the first place?

See #938

stephanegigandet:
Right, we do want to keep the rel="nofollow". @hangy's suggestion seems to be the right way.

PR #3973, opened 2020-08-02
Areesha product page

stephanegigandet:
Small fixes for @areeshatariq 's product page templatization.


stephanegigandet:
Thank you @areeshatariq !

PR #3979, opened 2020-08-03
chore: Reorganisation of label images

teolemon:
- Moving Spanish Protected origins images from `English labels folder` to `Spanish Categories folder`
- Doing the same with French Concours General Agricole


teolemon:
We have PDO as categories, actually.

hangy:
> We have PDO as categories, actually.

Can you show me an example? I cannot find `PDO` or `protected designation of origin` in https://github.com/openfoodfacts/openfoodfacts-server/blob/a22adb4aa19fe58aec26670b70bf6e1fb6b38a43/taxonomies/categories.txt.

Honestly, this confuses me a bit, because I understand those protection indicators as lables, but to me, they don't make a product different than another product, just because they're from some specific region. For example, https://de.openfoodfacts.org/produkt/22122568/natives-oliven%C3%B6l-extra-cucina is an olive oil with some GU label, but it's just an olive oil. 🤷

teolemon:
We've treated AOC/PGI/IGP as subclass of the category they are in:
it makes sense that https://world.openfoodfacts.org/category/de:Mainfranken%20Biere is a subclass of bier, rather than putting it in labels.
Most if all of those products, in addition to the geographic zoning have some strictly defined characteristics (a % of alcohol, specific raising methods…). So they are a subclass of the category they are in.
As a result, Robotoff can also categorize them easily using the category model.

stephanegigandet:
@teolemon for categories, we don't have PDO / DO etc. in the canonical entry name, so to be matched it should be removed from the file names.

stephanegigandet (approved):
Merging, but will probably require changes for some of the logos to be displayed, in order to have matching names.

PR #3985, opened 2020-08-04
Top Translators using Template::Toolkit

areeshatariq:
**Description:**

HTML Templatization of top_translators.pl using Template::Toolkit.
Made $tt an export variable so that we don't need to initialize it in every file and to avoid code redundancy. 


teolemon:
Merging. Thanks @areeshatariq :-)

PR #3998, opened 2020-08-06
Taxonomy error repairs

aleene:
**Related issues and discussion:** #3990

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -66339,18 +66339,13 @@ fr:Uroteuthis edulis, Calmar épée, Calmar, Encornet
 la:Uroteuthis edulis, Loligo edulis
 wikidata:en:Q2202434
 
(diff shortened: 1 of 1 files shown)


taxonomies/ingredients.txt
```diff
@@ -62370,67 +62353,63 @@ en:lota lota
...
-# kk-latn:Moýva
-# kk-tr:Moýva
-# vep:Hohol'
-# zh-cn:毛鳞鱼
-# zh-hans:毛鳞鱼
-# zh-hant:毛鱗魚
-# zh-hk:毛鱗魚
-# zh-sg:毛鳞鱼
-# zh-tw:毛鱗魚
+en:capelin, caplin, Mallotus villosus
```
stephanegigandet, line 62359: For ingredients, we match latin automatically, there is no need to add it to all entries.

PR #4006, opened 2020-08-08
Show Get the app on 13inch laptops

teolemon:
Show Get the app on 13inch laptops, so that people get a chance to see it.


stephanegigandet:
I think I mistakenly reverted this one when I fixed merge conflicts. I hadn't seen that PR and I thought it was a mistake.

We intentionnaly do not show the "Get the app" on smaller screens (just the icons) because in some languages there's not enough space in the top navigation bar, and it becomes ugly as it goes on 2 lines.

PR #4007, opened 2020-08-08
Refresh ES packager codes

Ban3:
**Description:**

Updated ES codes and a script to convert them.

Small change to the dataset from @teolemon: The RGSEAA Excel downloads are actually HTML files with XLS extension. Corrected the extension to avoid confusion.

[Here's the 641 rows that didn't receive coordinates](https://gist.github.com/Ban3/8f7d53361cc0b1d43687eb7381bd8877). Special prize to this "address":
``
23.004363/GC;Santana Cárdenes, Salvador; #$%;SANTA BRIGIDA;Palmas, Las;Canarias;;
``

### DIFF
--- a/scripts/packager-codes/es-packagers-html2csv.pl
+++ b/scripts/packager-codes/es-packagers-html2csv.pl
@@ -1,0 +1,281 @@
+#!/usr/bin/env -S perl -w
+
+# This file is part of Product Opener.
+#
+# Product Opener
+# Copyright (C) 2011-2020 Association Open Food Facts
+# Contact: contact@openfoodfacts.org
+# Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
+#
+# Product Opener is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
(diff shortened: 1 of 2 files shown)


hangy:
Nice! How large are the HTML files on disk? GitHub mobile says that this PR adds more than 4 million lines of text. Not sure how badly this might impact cloning the repo. (Which is already bad on low bandwidth.) Removing the HTML files and squashing the changes could end up saving some space/time.

Ban3:
> Nice! How large are the HTML files on disk? GitHub mobile says that this PR adds more than 4 million lines of text. Not sure how badly this might impact cloning the repo. (Which is already bad on low bandwidth.) Removing the HTML files and squashing the changes could end up saving some space/time.

Yeah, they were large enough to give magit issues when committing. I removed the files from the commit.

hangy:
> > Nice! How large are the HTML files on disk? GitHub mobile says that this PR adds more than 4 million lines of text. Not sure how badly this might impact cloning the repo. (Which is already bad on low bandwidth.) Removing the HTML files and squashing the changes could end up saving some space/time.
> 
> Yeah, they were large enough to give magit issues when committing. I removed the files from the commit.

Awesome, thanks!

PR #4020, opened 2020-08-13
More taxonomy errors resolved

aleene:


stephanegigandet:
@aleene : we tried to fix it at the same time :) I think there's another e322i that needed to be added.

stephanegigandet (approved):
Thank you!

PR #4023, opened 2020-08-14
Improve Perl critique

Ban3:
See #3895, or run `perlcritic` locally. Feel free to push to this branch if you work on these.

### DIFF
--- a/lib/ProductOpener/Cache.pm
+++ b/lib/ProductOpener/Cache.pm
@@ -64,9 +64,9 @@ use vars qw();
 sub get_multi_objects($)
 {
 	my $keys_ref = shift;
-	my $values_ref = $memd->get_multi(keys %$keys_ref);
+	my $values_ref = $memd->get_multi(keys %{$keys_ref});
 
-	foreach my $key (keys %$keys_ref) {
+	foreach my $key (keys %{$keys_ref}) {
 		if (not defined $values_ref->{$key}) {
 			if ($key =~ /\/blogs\/(.*)$/) {
 				my $blog_ref = retrieve("$data_root/index/blogs/$1/blog.sto");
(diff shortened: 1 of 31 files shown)


Ban3:
There's a lot of uses of match variables and violations of this policy: https://metacpan.org/pod/Perl::Critic::Policy::Variables::ProhibitMatchVars

But `perlvar` states:
>In Perl 5.20.0 a new copy-on-write system was enabled by default, which finally fixes all performance issues with these three variables, and makes them safe to use anywhere. 

So should we just add this policy to ignores?

hangy:
> There's a lot of uses of match variables and violations of this policy: https://metacpan.org/pod/Perl::Critic::Policy::Variables::ProhibitMatchVars
> 
> But `perlvar` states:
> 
> > In Perl 5.20.0 a new copy-on-write system was enabled by default, which finally fixes all performance issues with these three variables, and makes them safe to use anywhere.
> 
> So should we just add this policy to ignores?

That's probably a good idea. I don't really like the syntax and would prefer explicit match variables, but there's thousands of occurrences in the source code and "fixing" all of that doesn't seem to be feasible or necessary, because there are no perf implications, any more.

Ban3:
Policy https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::RequireArgUnpacking doesn't seem to understand `shift` as a legimate way to unpack args. Innocent subroutines like this will be flagged by it: https://github.com/openfoodfacts/openfoodfacts-server/blob/dcb7ba19690bf532f99f1882576b962faab52410/lib/ProductOpener/Store.pm#L235-L240

hangy:
> Policy https://metacpan.org/pod/Perl::Critic::Policy::Subroutines::RequireArgUnpacking doesn't seem to understand `shift` as a legimate way to unpack args. Innocent subroutines like this will be flagged by it:
> 
> https://github.com/openfoodfacts/openfoodfacts-server/blob/dcb7ba19690bf532f99f1882576b962faab52410/lib/ProductOpener/Store.pm#L235-L240

I've opened an issue with [Perl::Critic](https://github.com/Perl-Critic/Perl-Critic/issues/919) to investigate this.

Issue #4024, opened 2020-08-15
HTML elements not changing from left to right when l Right to left languages like "Arabic,Hebrew" selected


kmlhsn:
**Describe the bug**
When I select country "Saudi Arabia", languages changes to Arabic but HTML elements are not changing its direction from "Left to Right" to "Right to Left".    

**To Reproduce**
1. Go to 'https://world.openfoodfacts.org/' or run from your local machine
2. Click on 'Country' Drop Down Box
3. Scroll down to 'Saudi Arabia or Israel' or other Right to Left language 
4. Select country

**Expected behavior**
When "Right to Left" language is selected, HTML elements should change its direction from "Left to Right" to "Right to Left"

**Screenshots**
![image](https://user-images.githubusercontent.com/32615215/90307282-6b324880-def6-11ea-8854-559c5b95955d.png)

**Desktop (please complete the following information):**
Ans: Windows 10
Ans: Mozilla Firefox
79
**Smartphone (please complete the following information):**
Ans: Nexus 4
Ans: Android 5
Ans: Mozilla Firefox
Ans" 68.11

**Additional context**

**Number of products impacted**
If relevant, number (or estimate) of products impacted by the bug or feature at the time of creation of the issue

**Timeloss per product**
Estimate of timeloss provoked or saved by the issue, per product

hangy:
@kmlhsn Could you please have a look at https://sa.openfoodfacts.dev/ (user/password: off/off) and check how the RTL adaptation works for you, and if there are any major issues? Thank you! 🙂

(Images not showing or other features not working perfectly are expected on this dev site.)

PR #4025, opened 2020-08-15
Improve RTL support

hangy:
**Description:**
Based off #3845, this modifies a few UI elements to improve usability with languages that use right to left text orientation.

**Related issues and discussion:** #4024

 Interesting reads:
- https://hacks.mozilla.org/2015/09/building-rtl-aware-web-apps-and-websites-part-1/
- https://hacks.mozilla.org/2015/10/building-rtl-aware-web-apps-websites-part-2/
- https://meiert.com/en/blog/css-how-to-host-right-to-left-styling/


teolemon:
Is the spreadsheet applied to only languages that are actually RTL ?

hangy:
> Is the spreadsheet applied to only languages that are actually RTL ?

`app-ltr.css` is applied to LTR languages, `app-rtl.css` to RTL languages. We need separate stylesheets because of the way that Foundation handles RTL support internally. I decided to add a `text_direction` translation for the switch, so that translators can decide if/when a translation is good enough to enable RTL mode.

hangy:
> Thank you! Looks good to me, I'll merge it, and install it on the dev server.

Thanks!

> One thing is that all the donate html files are generated through crowdin, so we may lose the change from app.css to app-ltr.css at some point. But we can worry about it later.

I know, but I was hoping that crowdin might pick it up, because the original was changed, too. It's not a big issue, if it doesn't. 😄

teolemon:
I'll make sure we won't get crowdin *** back.

PR #4028, opened 2020-08-16
Taxonomy edits

svensven:
add "garbanzo beans" as synonym of "chickpea"


svensven:
Sorry, I'm making a mess of this.
I thought Stearic Acid/E570 was a missing ingredient, but then I saw it was in the additives file instead, so I'm not sure what to do about that.

PR #4031, opened 2020-08-17
Taxonomy edits

svensven:
Adding ingredient spellings seen in photos. Changing Vitamines to Vitamins in English.

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -70911,7 +70911,7 @@ ciqual_food_code:en:11086
 ciqual_food_name:en:Spirulina, (spirulina sp.), dried
 
 <en:Dietary supplements
-en:Vitamins
+en:Vitamins, vitamines
 de:Vitamine
 fi:vitamiinit
 fr:Vitamines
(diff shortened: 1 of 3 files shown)


svensven:
Now I'm wondering if "Vitamins" should be added as an alias to "en:Vitamines" instead of replacing it. Will replacing the word break existing tagging?
https://world.openfoodfacts.org/category/vitamines

stephanegigandet:
> Now I'm wondering if "Vitamins" should be added as an alias to "en:Vitamines" instead of replacing it. Will replacing the word break existing tagging?

In that case, "vitamins" is definitely better than "vitamines", so what we can do is to rename it, but add "Vitamines" as a synonym, so that existing links don't break.

taxonomies/categories.txt
```diff
@@ -70911,7 +70911,7 @@ ciqual_food_code:en:11086
 ciqual_food_name:en:Spirulina, (spirulina sp.), dried
 
 <en:Dietary supplements
-en:Vitamines
+en:Vitamins
```
stephanegigandet, line 70914:
```suggestion
en:Vitamins, vitamines
```

(adding vitamines as a synonym so that old links etc. don't break)

PR #4038, opened 2020-08-18
Display barcode bug fix

### LINKED ISSUE #4015: The barcode is not displayed on the product page anymore
The barcode is not displayed on the product page anymore. This massively hurts SEO and discoverability. P1.

areeshatariq:
**Related issues and discussion:** fixes #4015

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -7461,7 +7461,7 @@ CSS
 	if ($code =~ /^2000/) { # internal code
 	}
 	else {
-		$template_data_ref->{upc} = 'defined';
+		$template_data_ref->{upc_code} = 'defined';
 		# Also display UPC code if the EAN starts with 0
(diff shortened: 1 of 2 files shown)


areeshatariq:
> I'm not sure I understand how this change solves the issue. You rename the variable in both places, but the value seems to be the same (from a first look).

I was passing 'upc' to the template twice with different values, e.g.
$template_data_ref->{upc} = 'defined';
$template_data_ref->{upc} =$upc;
So, I changed one

hangy:
> > I'm not sure I understand how this change solves the issue. You rename the variable in both places, but the value seems to be the same (from a first look).
> 
> I was passing 'upc' to the template twice with different values, e.g.
> $template_data_ref->{upc} = 'defined';
> $template_data_ref->{upc} =$upc;
> So, I changed one

Thanks for the clarification. 🙂

Issue #4011, opened 2020-08-11
New search API - tags fields


stephanegigandet:
The existing Open Food Facts search API is a hack of the web site search form, and thus the parameters are the ones that make sense for how the search form is designed (being able to add criterias etc.), but they make less sense for a search API.

e.g. this is needlessly complex:

	my $tagtype = remove_tags_and_quote(decode utf8=>param("tagtype_$i"));
	my $tag_contains = remove_tags_and_quote(decode utf8=>param("tag_contains_$i"));
	my $tag = remove_tags_and_quote(decode utf8=>param("tag_$i"));

Proposal for a new search API:

Use the tags fields directly:

labels_tags=en:organic
labels_tags_fr=bio -> bio is assumed to be in French and thus converted to en:organic

For AND queries, use comma separated values:

labels_tags=en:organic,en:fair-trade

For negative queries, prefix tag with -
labels_tags=-en:organic,-en:fair-trade

For OR queries, use | to separate values
labels_tags=en:organic|en:fair-trade|en:some-other-label

stephanegigandet:
Those parameters would be added to search.pl and to all OFF urls.
e.g. https://de.openfoodfacts.org/?labels_tags=en:organic&categories_tags=-en:cakes would work.

stephanegigandet:
The existing search.pl parameters and endpoint would continue to work, but we could introduce a cleaner endpoint, like /api/v2/search for instalce.

stephanegigandet:
Documenting on https://wiki.openfoodfacts.org/Open_Food_Facts_Search_API_Version_2

PR #4039, opened 2020-08-18
New search API + simplified parameter handling

stephanegigandet:
1. New search API, see #4011 for details.

sample queries:

https://world.openfoodfacts.dev/api/v0/search?fields=product_name&categories_tags_fr=gateaux,-madeleines,-truc,-cake%20aux%20fruits

https://world.openfoodfacts.dev/api/v0/search?fields=product_name&categories_tags_fr=gateaux|sodas|chocolats&additives_tags_fr=stevia

bonus:

the parameters can also be added to any URLs, for instance to get an unlimited level of drilldown with facets:

https://world.openfoodfacts.dev/?labels_tags=en:organic&categories_tags_fr=gateaux&brands_tags=carrefour

2. Simplified parameter handling: we now use CGI param() to parse the query string, so all parameters can be sent either on the query or by POSTing


teolemon:
- [ ] We're currently using facets on the Android apps
- [ ] We could do a native advanced search quite easily
- [ ] We can probably make more advanced queries on Hunger Games

VaiTon:
What about /v2/

stephanegigandet:
> What about /v2/

@VaiTon : You mean forcing to use /v2/ to use the new /api/v?/search API? We could, but there are no breaking changes, only new things.

stephanegigandet:
Currently this change is independent of the API version, the new fields are applied to all versions.

hangy:
I think the overall design is much better than the previous approach. Can AND and OR queries be combined when searching? It might be interesting to search for (pseudocode) `(categories_tags=A && additives_tags=B) || (categories_tags=C && additives_tags=D)`.

Would it be useful to have a REST endpoint that lists the tags fields that are configured for the site? Do we already have something like that? That could help with programmatic API discovery (hateoas?) of the API.

> the parameters can also be added to any URLs, for instance to get an unlimited level of drilldown with facets:
> 
> https://world.openfoodfacts.dev/?labels_tags=en:organic&categories_tags_fr=gateaux&brands_tags=carrefour

I'm wary of publishing this in some kind of API docs. I think it's better to limit this kind of functionality to some kind of versioned endpoint, so that it can be switched off at some point in time, when an old API version gets retired. Also, it's probably not really 100% intuitive at which URL that's kind of query is possible. We could just have a simple HTML site endpoint that provides this functionality.

hangy:
Some interesting cases (of which I don't know if/how they work in the current search, to be honest):

https://world.openfoodfacts.dev/api/v0/search?fields=product_name_de&categories_tags=de:wasser currently returns some empty products. I have no idea how other APIs that return JSON lists do this, but it does look weird at a first glance.
```JSON
{"count":2798,"page_size":20,"skip":0,"products":[{},{},{},{},{},{},{},{"product_name_de":""},{},{},{},{},{"product_name_de":"Mineralwasser Mit Kohlensäure Vers..."},{},{},{},{},{},{},{}],"page":1}
```

Have the pagination parameters changed from the old API? https://world.openfoodfacts.dev/api/v0/search?fields=product_name,code&categories_tags=de:proteinpulver&page_size=5&page=5 doesn't seem to return a different page.

According to https://wiki.openfoodfacts.org/index.php?title=API/Read/Search&oldid=8819, the current search API supports searching for something like `energy-kj ≥ 1000`. Is this supported in the new API, too?

stephanegigandet:
> Can AND and OR queries be combined when searching? It might be interesting to search for (pseudocode) `(categories_tags=A && additives_tags=B) || (categories_tags=C && additives_tags=D)`.

Currently no. MongoDB does support queries like that, but I don't think it's worth adding complexity to support queries that are likely to be quite rare (and that may be expensive to run).

> Would it be useful to have a REST endpoint that lists the tags fields that are configured for the site? Do we already have something like that? That could help with programmatic API discovery (hateoas?) of the API.

We don't have that, I'm not familiar with it. I guess it can be added later if you think it's useful.

> I'm wary of publishing this in some kind of API docs. I think it's better to limit this kind of functionality to some kind of versioned endpoint, so that it can be switched off at some point in time, when an old API version gets retired. Also, it's probably not really 100% intuitive at which URL that's kind of query is possible. We could just have a simple HTML site endpoint that provides this functionality.

We could just have a /search endpoint for that. That would mimic what we have for product pages:  /product/ for the web site, and /api/v?/product/ for the API.

stephanegigandet:
> https://world.openfoodfacts.dev/api/v0/search?fields=product_name_de&categories_tags=de:wasser currently returns some empty products. I have no idea how other APIs that return JSON lists do this, but it does look weird at a first glance.

It returns products that don't have a product_name_de field, but the products are not empty: https://world.openfoodfacts.dev/api/v0/search?fields=code,product_name_de&categories_tags=de:wasser

> Have the pagination parameters changed from the old API? https://world.openfoodfacts.dev/api/v0/search?fields=product_name,code&categories_tags=de:proteinpulver&page_size=5&page=5 doesn't seem to return a different page.

I have not implemented them, good point.
 
> According to https://wiki.openfoodfacts.org/index.php?title=API/Read/Search&oldid=8819, the current search API supports searching for something like `energy-kj ≥ 1000`. Is this supported in the new API, too?

Not yet. I'm not sure what's the right way to do it.  &energy_kj=>=1000 looks a bit strange. I guess it would work.

hangy:
> > Can AND and OR queries be combined when searching? It might be interesting to search for (pseudocode) `(categories_tags=A && additives_tags=B) || (categories_tags=C && additives_tags=D)`.
> 
> Currently no. MongoDB does support queries like that, but I don't think it's worth adding complexity to support queries that are likely to be quite rare (and that may be expensive to run).

Agree, I'm not sure about how often that would be required. I just thought it could be a good option to think about in a new version of the search API, since grouping search terms like that would likely require a little bit of a different API design.

I just remembered that some people were [talking](https://openfoodfacts.slack.com/archives/C032PK9D8/p1422017716000009) about a possible SPARQL endpoint for the data/search, but I guess you didn't want to go that far?

> > Would it be useful to have a REST endpoint that lists the tags fields that are configured for the site? Do we already have something like that? That could help with programmatic API discovery (hateoas?) of the API.
> 
> We don't have that, I'm not familiar with it. I guess it can be added later if you think it's useful.

Yup, can be added later. No worries. 🙂 

> > I'm wary of publishing this in some kind of API docs. I think it's better to limit this kind of functionality to some kind of versioned endpoint, so that it can be switched off at some point in time, when an old API version gets retired. Also, it's probably not really 100% intuitive at which URL that's kind of query is possible. We could just have a simple HTML site endpoint that provides this functionality.
> 
> We could just have a /search endpoint for that. That would mimic what we have for product pages: /product/ for the web site, and /api/v?/product/ for the API.

Sounds good. I hope that this would make it more obvious for end users, and also is a single endpoint that could be switched off some day (if required).

> > https://world.openfoodfacts.dev/api/v0/search?fields=product_name_de&categories_tags=de:wasser currently returns some empty products. I have no idea how other APIs that return JSON lists do this, but it does look weird at a first glance.
> 
> It returns products that don't have a product_name_de field, but the products are not empty: https://world.openfoodfacts.dev/api/v0/search?fields=code,product_name_de&categories_tags=de:wasser

I understand that, but it might be unexpected from an API user's perspective. Maybe it would be useful to always include the `_id` or `code` (whichever is supposed to be unique long-term (I'm thinking about product versions etc)) as a mandatory field, so that no entry is just `{}`? I don't know. 🙂 

> > According to https://wiki.openfoodfacts.org/index.php?title=API/Read/Search&oldid=8819, the current search API supports searching for something like `energy-kj ≥ 1000`. Is this supported in the new API, too?
> 
> Not yet. I'm not sure what's the right way to do it. &energy_kj=>=1000 looks a bit strange. I guess it would work.

It looks a bit odd, but it should work, if URL encoded correctly, ie. `energy_kj=%3E%3D+1000&alcohol=%3C%3D+42` or `energy_kj=%E2%89%A5+1000&alcohol=%E2%89%A4+42`. I've also seen other [APIs](https://docs.camunda.org/manual/7.13/reference/rest/variable-instance/post-query/#request) using a separate field for that, but this would require processing JSON documents for search queries. Looks of the parameter shouldn't matter too much, as long as people escape their parameter values correctly. 😄

hangy:
Isn't this less functional than the previous search API then? I don't know about this "personal search" feature, I but I suppose we'd need a tracking bug or milestone for New Search API todos? Is "personal search" a bigger epic?

stephanegigandet:
> Isn't this less functional than the previous search API then? I don't know about this "personal search" feature, I but I suppose we'd need a tracking bug or milestone for New Search API todos? Is "personal search" a bigger epic?

Yes right now it's less functional. Overtime we can add the missing features (and filing bugs for them is a good idea). The existing Search API is not going away and will still be available. 

Here is the wiki page for the Personal Search project: https://wiki.openfoodfacts.org/Project:Personalized_Search

stephanegigandet:
Adding some docs :)
https://wiki.openfoodfacts.org/Open_Food_Facts_Search_API_Version_2

teolemon (approved):

stephanegigandet:
Thanks for all the feedback, I'm merging this one, and I will make a separate PR for the conditions on nutrients.

PR #4042, opened 2020-08-19
Taxonomy edits - ingredients and categories

svensven:
More anomalies found while editing records.

Not sure if the merge being in there is a bad thing. I had to resolve a conflict.


aleene (approved):
Looks good. By the way it is enough to enter synonyms that are actually used on products.

stephanegigandet (approved):

stephanegigandet:
Thanks!

PR #4043, opened 2020-08-20
Use PERL5LIB for lib location instead of a symlink

hangy:
While working on #4023, I noticed that we still use symlinks and to get the lib directory when using one of the Perl scripts. Using the `PERL5LIB` environment variable instead is a bit more straight-forward for users, and much easier when searching for perlcritic or similar issues, since you don't have to constantly ignore the symlinked directories.


stephanegigandet:
I'm a bit worried about the production and dev environments where we have multiple instances of product opener in different directories (sometimes with different code, and each with a different Config2.pm file) with the same off user. The symlink ensured that all the modules, scripts, cgi etc. in one of the instances use that instance's modules.
e.g. in production, we have 4 instances for off, obf, opf, opff + the off-pro instance

hangy:
> I'm a bit worried about the production and dev environments where we have multiple instances of product opener in different directories (sometimes with different code, and each with a different Config2.pm file) with the same off user. The symlink ensured that all the modules, scripts, cgi etc. in one of the instances use that instance's modules.
> e.g. in production, we have 4 instances for off, obf, opf, opff + the off-pro instance

The same approach has previously been used by https://github.com/openfoodfacts/openfoodfacts-server/blob/93a9d714918788b15be9a72c5382426482e97952/scripts/equadis-import/run_equadis_import.sh#L24 and https://github.com/openfoodfacts/openfoodfacts-server/blob/93a9d714918788b15be9a72c5382426482e97952/scripts/gen_opff_leaderboard.sh#L4, so I dont't not sure that matters. The scope of the variable change is also limited to child processes, ie.
```
root@7d9a58b62ecf:/opt/product-opener# echo $PERL5LIB
/opt/product-opener/lib/:/opt/perl/local/lib/perl5/
root@7d9a58b62ecf:/opt/product-opener# bash
root@7d9a58b62ecf:/opt/product-opener# export PERL5LIB=a
root@7d9a58b62ecf:/opt/product-opener# echo $PERL5LIB
a
root@7d9a58b62ecf:/opt/product-opener# bash
root@7d9a58b62ecf:/opt/product-opener# echo $PERL5LIB
a
root@7d9a58b62ecf:/opt/product-opener# export PERL5LIB=b
root@7d9a58b62ecf:/opt/product-opener# echo $PERL5LIB
b
root@7d9a58b62ecf:/opt/product-opener# exit
exit
root@7d9a58b62ecf:/opt/product-opener# echo $PERL5LIB
a
root@7d9a58b62ecf:/opt/product-opener# exit
exit
root@7d9a58b62ecf:/opt/product-opener# echo $PERL5LIB
/opt/product-opener/lib/:/opt/perl/local/lib/perl5/
root@7d9a58b62ecf:/opt/product-opener# 
```

stephanegigandet:
OK, let's try it. :)

stephanegigandet (approved):

Issue #4040, opened 2020-08-19
Automatically generate sample Excel file on the producers platform


stephanegigandet:
Offer producers to generate a sample Excel file with the right column names, comments to help to fill them etc.

There will be a very simple form for producers to indicate if their products have nutrition facts per 100g/serving, as sold, prepared etc. so that we can generate the right columns (and not all possible columns).

We can use comments to provide suggestions / help / examples.

In practice it will be very similar (and will reuse the code) to the dropdown we have in the fields selection, with a grouping by type.

![image](https://user-images.githubusercontent.com/8158668/90632944-7dc5be00-e225-11ea-9481-9a27d1a563e4.png)

teolemon:
- With common_name_XX depending on the country ? with potentially several XX for multilingual countries ?
- What is the desired effect/outcome of letting them customize the fields to the extreme ? Reducing anxiety ?

stephanegigandet:
The fields names will be in plain language (the language they are using on the platform), not the OFF ids in English and underscores.

Simple things first: a simple file with only one language, and the most common fields.

There's no extreme customization planned, just a few simple boxes for relatively common things (like nutrition facts values for the prepared product). The goal is to not have 4532 columns in the sample file.

teolemon:
Additional thought: would letting them export their current products be useful ?

PR #4044, opened 2020-08-20
generate sample import file + small fixes, bug #4040

stephanegigandet:
Start of a script to let producers generate a simple sample import file in their language. See #4040

### DIFF
--- a/cgi/generate_sample_import_file.pl
+++ b/cgi/generate_sample_import_file.pl
@@ -1,0 +1,134 @@
+#!/usr/bin/perl -w
+
+# This file is part of Product Opener.
+#
+# Product Opener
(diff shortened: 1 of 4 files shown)


stephanegigandet:
Currently generate sample file: 
[
[openfoodfacts_import (1).xlsx](https://github.com/openfoodfacts/openfoodfacts-server/files/5102807/openfoodfacts_import.1.xlsx)
](url)

More helpful comments will be provided through new .po entries.

stephanegigandet:
In progress: removing all the optional nutrients, and keeping only the default one. We may add a checkbox at some point to add vitamins and minerals.

teolemon (approved):

PR #4050, opened 2020-08-21
ci: Improve Perl critique lvl 4

hangy:
### What
- See #4023 and build errors, or run `perlcritic` locally. Feel free to push to this branch if you work on these.
### Part of
- #5512


teolemon:
@hangy is it in a mergeable state ? Why not merge now instead of later ?

Issue #4052, opened 2020-08-22
Packaging code auto completion


TaciteOFF:
Hello,

The packaging code field has an auto completion feature but it appears that it has been populated with very odd values. 
I let you see on this image : 
http://data.nimages.fr/creenshot20200822at20325889p.png

Also, it can be quite hard to click on the result we want, as seen on this video : 
https://imgur.com/a/pWJUPJW

I'm using Safari 13.1.2.

Thanks

PR #4066, opened 2020-08-24
Replace separators embedded in FR and ES packager code fields

Ban3:
**Description:**
Fixes the first issue described in #4052. The custom parser doesn't support quoted separator characters in fields causing incorrect field splitting. This replaces all separators appearing in fields with commas.

As a bonus, looks like ES got a couple new coordinates.

**Related issues and discussion:** #4052


stephanegigandet (approved):
Thank you!

PR #4067, opened 2020-08-24
revert bad commit aeeca45 - add back agribalyse_food_code to categories taxonomy

stephanegigandet:
Not sure what happened, but a commit seems to have removed all the agribalyse_food_codes 
https://github.com/openfoodfacts/openfoodfacts-server/commit/aeeca450e68267986fe6927e6e3442bac2c9f22b#diff-f10308cb67581847fc8c5b78473832bf

reverting commit aeeca45

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -1314,7 +1314,6 @@ sub parse_ingredients_text($) {
 							'nl' => [
 								'in wisselende verhoudingen',
 								'harde fractie',
-								'o.a.',
 							],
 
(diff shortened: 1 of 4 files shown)


stephanegigandet:
@aleene : we can commit this revert that adds back the agribalyse properties (and some ingredients), then you can make a new PR to apply only your Dutch changes

Ban3:
I'm currently reverting the merge so that it only affects categories. If you give me a little time to check that I'm reverting only what's wanted I can make PR that doesn't require reapplying the other changes.

PR #4081, opened 2020-08-27
Taxonomy edits 6 - ingredients & additives

svensven:
More of the same, although 66f6e90 assumes that the ingredients list is _always_ used case insensitively, and doesn't need both upper and lower case versions of names.


stephanegigandet (approved):
Looks good, thank you!

PR #4086, opened 2020-08-29
Port update_packager_codes to using Text::CSV & SE code refresh

Ban3:
**Description:**

Ported `update_packager_codes.pl` to using `Text::CSV` from the custom parser. Hopefully this will reduce the chances of bugs like #4052 from happening.

Additionally fix some issues that surfaced during development:
- Re-created SE codes because it was easier than trying to figure out the encoding issues by hand.
- Corrected SE country code from SV so the codes actually get loaded
- Added missing code for loading SK codes
- Tweaked the RS code normalization regex to correctly normalize the codes in the form `RS-XXX-XXX`
- Minor corrections to other CSV files to parse them correctly


stephanegigandet (approved):
Thanks a lot for all the cleaning and improvements!

PR #4092, opened 2020-08-30
Templatize revision info and navigator

hangy:
Updates `display_rev_info` to use `Template::Toolkit`. Also updates the styling to be compatible with RTL text direction.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -9433,7 +9433,7 @@ sub display_rev_info {
 
 	$comment =~ s/^Modification :\s+//;
 	if ($comment eq 'Modification :') {
-		$comment = qw{};
+		$comment = q{};
 	}
 
 	$comment =~ s/\new image \d+( -)?//;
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -9405,59 +9406,51 @@ sub display_rev_info {
...
+		$next_link = $product_url . '?rev=' . ($rev + 1);
 	}
 
 	my $comment = $change_ref->{comment};
 	$comment = lang($comment) if $comment eq 'product_created';
 
 	$comment =~ s/^Modification :\s+//;
 	if ($comment eq 'Modification :') {
-		$comment = '';
+		$comment = qw{};
```
Ban3, line 9436: Did you mean
```suggestion
		$comment = q{};
```
Perl sees `qw{}` as a empty list:
```sh
perl -MO=Deparse -e '$comment = qw{}'

$comment = ();
-e syntax OK
```

stephanegigandet: Why replace '' by qw{} ?

hangy: > Why replace '' by qw{} ?

The policy [`ValuesAndExpressions::ProhibitEmptyQuotes`](https://metacpan.org/pod/Perl::Critic::Policy::ValuesAndExpressions::ProhibitEmptyQuotes) doesn't like `''`. We don't use this level right now, but I'm trying to apply it to my current changes to avoid future policy violations. 🙂

stephanegigandet (approved):
Looks good

PR #4094, opened 2020-08-30
Templatize product history

hangy:
Updates `display_product_history` to use `Template::Toolkit`.


stephanegigandet (approved):
Looks good, thank you

PR #4096, opened 2020-08-30
Taxonomy errors 5

aleene:


aleene:
@stephanegigandet what is the issue? There is no fail.

taxonomies/additives.txt
```diff
@@ -16422,63 +16509,77 @@ additives_classes:en: en:sequestrant
 vegan:en:yes
 vegetarian:en:yes
 
-en:E578, Calcium gluconate
+comment:en:CALCIUM GLUCONATE is for the moment listed in both the additves and ingredients taxonomy.
+#<en:Calcium
+<en:E578
+en:E578, calcium gluconate
```
svensven, line 16515: Guessing it doesn't like this.
2020-09-07T15:43:53.8871819Z Errors in the additives taxonomy definition:
2020-09-07T15:43:53.8909631Z ERROR - en:e578 is a parent of itself
2020-09-07T15:43:53.8910404Z Errors in the additives taxonomy definition at lib/ProductOpener/Tags.pm line 1618.
2020-09-07T15:43:54.1858128Z [additives] child process exited with code 25

stephanegigandet (changes requested):
Putting descriptions / comments above entries is not supported.

description:en:CALCIUM CARBONATE is a chemical compound with the formula CaCO3.
comment:en:This entry must be doubled in the minerals and additives taxonomy (for the time being).
#<en:calcium
#en:E170(i)
<en:E170	<en:E170

If those are comments : just start the line with # .

If those are properties (like description:en: ) they need to be after the entry and all the synonyms.

aleene:
Note that I did it like that everywhere (all taxonomies).

stephanegigandet (approved):
Thank you!

PR #4097, opened 2020-08-30
Templatize leaflet maps

hangy:
Updates some leaflet based map views to use `Template::Toolkit`. At the same time, it removes manual JSON building and moves some previously inline JavaScript to TT templates or separate JavaScript files.


stephanegigandet (approved):
Thank you!

PR #4109, opened 2020-08-31
fix: use HTML::Entities instead of require

### LINKED ISSUE #4108: Issue with the Fleury-Michon import
Some ingredient lists were replaced by HTML::Entities

Ban3:
**Description:**

Fixes #4108

calling `HTML::Entities->decode_entities` will return a list where
the first value is the class name and second is the actual return value.

**Related issues and discussion:** #4108

### DIFF
--- a/scripts/import_fleurymichon.pl
+++ b/scripts/import_fleurymichon.pl
@@ -51,6 +51,7 @@ use Encode;
 use JSON::PP;
 use Time::Local;
 use XML::Rules;
+use HTML::Entities qw/decode_entities/;
 
 $lc = "fr";
 $country = "en:france";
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Interesting bug.

Ban3:
> Interesting bug.

Can't say I understand perl objects well, but I think it's because the thing on the left side will be passed as the first argument when using `->`

```perl
(trepanpl): eval@ HTML::Entities->decode_entities('&copy')
[
    [0] "HTML::Entities",
    [1] "\251"
]

(trepanpl): eval@ HTML::Entities::decode_entities('HTML::Entities','&copy')
[
    [0] "HTML::Entities",
    [1] "\251"
]

```

VaiTon (approved):

PR #4135, opened 2020-09-03
support rev for product read API, #4134

### LINKED ISSUE #4134: Retrieve a specific revision of a product through the API
Something like this would be very useful for debugging:

https://world.openfoodfacts.org/api/v0/product/8410422407028/salsa-picante-louisiana-chovi&rev=28

stephanegigandet:
fixes #4134 

add support for queries like https://world.openfoodfacts.org/api/v0/product/8410422407028/salsa-picante-louisiana-chovi?rev=28

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -9206,7 +9206,6 @@ sub display_product_api($)
 	if (defined $rev) {
 		$log->info("displaying product revision") if $log->is_info();
 		$product_ref = retrieve_product_rev($product_id, $rev);
-		$header .= '<meta name="robots" content="noindex,follow">';
 	}
 	else {
 		$product_ref = retrieve_product($product_id);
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -9198,7 +9198,19 @@ sub display_product_api($)
...
-	my $product_ref = retrieve_product($product_id);
+	
+	my $product_ref;
+
+	my $rev = param("rev");
+	local $log->context->{rev} = $rev;
+	if (defined $rev) {
+		$log->info("displaying product revision") if $log->is_info();
+		$product_ref = retrieve_product_rev($product_id, $rev);
+		$header .= '<meta name="robots" content="noindex,follow">';
```
hangy, line 9209: As this is for the API, which does not have any HTML output, setting `$header` should not do anything. I'd just remove it. 🙂

teolemon (approved):
Many great uses of this

lib/ProductOpener/Display.pm
stephanegigandet: Good catch, I'll remove it. Thanks!

PR #4138, opened 2020-09-04
Taxonomy edits 10 - category dedupes and translations

svensven:


taxonomies/categories.txt
```diff
@@ -33565,54 +33479,63 @@ es:Quesucos de Liébana
 protected_name_file_number:en: PDO-ES-0090
 protected_name_type:en: pdo
 origins:en: en:Spain
+country:en:Spain
```
stephanegigandet, line 33482: In fact we need to remove all the country:en: entries, that was an experiment, but we use the countries field to indicate where the product is sold, not where it is from (we use the origins field for that). cc @teolemon

stephanegigandet (approved):
Looks good to me, thank you!

taxonomies/categories.txt
svensven: hmm, apparently there are 2217 "^country" entries. not one to edit by hand. :D

svensven: I've removed the "country:"s from categories that had an origins as well, so in theory, one could just search+replace the remaining country lines into origins ones?

svensven: also, is the space in the origins lines intentional?

Ban3: > also, is the space in the origins lines intentional?

I think it's there just to make the double `en:` clearer. Parser shouldn't care about that.

Issue #4117, opened 2020-09-01
Add "Non-organic", "Non-fair trade", "Non-vegetarian", "Non-vegan" labels


stephanegigandet:
To compute the new eco-score environmental score, and for the personal search project, we need to know if a product has some specific labels like "Organic" or "Fair trade". If we don't have a label in our labels field, we can't be 100% sure that the label is not present on the product, it might not have been added by the user yet, and it may not have been recognized in product photos by robotoff.

One solution could be to add a very limited set of "negative" labels: "Not organic", "Not fair trade", and gradually add them to products (starting with the most scanned).

stephanegigandet:
Some producers send us data files with columns for "Organic", "Vegetarian" etc. with values like YES and NO. When they send NO, we can add the non-organic etc. labels.

stephanegigandet:
Adding a new property to indicate the opposite of an entry:

<en:Vegetarian
en:Vegan, Suitable for vegans
opposite:en: en:non-vegan

en:Non-vegan, not vegan, not suitable for vegans
opposite:en: en:vegan

PR #4146, opened 2020-09-05
Add opposite labels through imports + small improvements

stephanegigandet:
Some producers send us data file with columns like "Organic" with Yes/No values. We previously only used the Yes values to assign the en:organic label. We can now add the en:non-organic label for No values, thanks to the new opposite labels. (see #4117)


lib/ProductOpener/Config_off.pm
```diff
@@ -1,4 +1,5 @@
 # This file is part of Product Opener.
+# This file is part of Product Opener.
```
svensven, line 2:
```suggestion
```

teolemon (approved):

PR #4149, opened 2020-09-06
Support /.well-known/change-password URL

### LINKED ISSUE #4140: Support for /.well-known/change-password
New versions of web browsers have password managers that let you change your password easily if it has been compromised in a breach (and that you have reused it many times on many websites).

To make it easy for user to change password, a new .well-known path has been proposed
https://web.dev/change-password-url/

/.well-known/change-password

hangy:
**Description:** Adds a new redirect, login form, and change password form to support the `/.well-known/change-password`.

**Related issues and discussion:** Closes #4140

### DIFF
--- a/cgi/change_password.pl
+++ b/cgi/change_password.pl
@@ -1,0 +1,101 @@
+#!/usr/bin/perl -w
+
+# This file is part of Product Opener.
+#
+# Product Opener
+# Copyright (C) 2011-2020 Association Open Food Facts
+# Contact: contact@openfoodfacts.org
+# Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
+#
+# Product Opener is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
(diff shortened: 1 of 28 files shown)


stephanegigandet (changes requested):
The /.well-known directory is used for other things (in particular to store the acme-challenges created by letsencrypt to renew SSL certificates), so I think it's best to keep it served by nginx instead of passing it to Apache. We can just add an nginx rewrite rule to redirect /.well-known/change-password to /cgi/change_password.pl

hangy:
> The /.well-known directory is used for other things (in particular to store the acme-challenges created by letsencrypt to renew SSL certificates)

Is it still used for that? AFAIK, Let's Encrypt doesn't create wildcard certs with the HTTP challenge.

> so I think it's best to keep it served by nginx instead of passing it to Apache. We can just add an nginx rewrite rule to redirect /.well-known/change-password to /cgi/change_password.pl

I can do that, but do we need to? A lot of the URIs listed at http://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml seem to be related to application specific data. (For example OIDC, which I'd still prefer to the current API access.)

stephanegigandet:
> Is it still used for that? AFAIK, Let's Encrypt doesn't create wildcard certs with the HTTP challenge.

For OFF, we use DNS validation with acme.sh, but for madenearme etc. we still use letsencrypt HTTP validation, so we should not change the nginx config for those.

hangy:
> > Is it still used for that? AFAIK, Let's Encrypt doesn't create wildcard certs with the HTTP challenge.
> 
> For OFF, we use DNS validation with acme.sh, but for madenearme etc. we still use letsencrypt HTTP validation, so we should not change the nginx config for those.

Fixed 🙂

stephanegigandet (approved):
Looks good, thank you!

PR #4152, opened 2020-09-06
add terms to ignore on the end of ingredients list for english

svensven:
Editing records, a often ingredients lists still have a bit of "nutrition..." tacked on ([example](https://uk.openfoodfacts.org/product/4771189250086/chef-tomato-ketchup)), both when OCRing and in the ingredients analysis.
This is my attempt to reduce that a bit, if I haven't completely misunderstood the code/tests.


stephanegigandet (approved):
Looks good, thank you!

PR #4154, opened 2020-09-06
Allergens markup in ingredients taxonomy

aleene:
**Description:**
Adding the allergens to the ingredients taxonomy. 

Purpose is to have a easier maintainable allergens list (no doubling, completer).


taxonomies/ingredients.txt
```diff
@@ -21827,6 +21831,7 @@ en:crustaceans flavouring
 de:Krustentieraroma, Schalentieraroma
 fi:äyriäisaromi
 fr:arôme de crustacés
+allergen:en:en:en:crustaceans
```
svensven, line 21834:
```suggestion
allergen:en:en:crustaceans
```

taxonomies/ingredients.txt
```diff
@@ -23426,7 +23435,7 @@ zh:大麦
 # yue:大麥
 wikidata:en:Q61665121
 wikipedia:en:https://en.wikipedia.org/wiki/Barley
-
+allergen:en:en:gluten
```
svensven, line 23438:
```suggestion
allergen:en:en:gluten

```

svensven:
what's the significance of removing the wheat parents on "glucose de blé" etc?
also, pizza bases can be gluten free, even if it's very unlikely.

aleene:
The wheat parents have been removed as it is a wrong usage of parents. They idea is that you should be able to replace an ingredient on a product with its parent. Thus `wheat glucose` can be replaced with `glucose`, but not with `wheat`. Another way to look at it is that wheat is an adjective and more specific and glucose is a noun and more generic.

aleene:
You are right, you can buy gluten free pizza bases. For allergens I prefer the conservative approach (it is an allergen, unless).

stephanegigandet:
For compound ingredients, we should not add allergens info unless we know for sure it contains the allergen. In actual ingredients lists, the compound ingredients are followed by the individual ingredients.

e.g. Pizza Base (Wheat Flour, Water, Rapeseed Oil, Yeast, Dextrose, Salt)

See https://world.openfoodfacts.org/ingredient/pizza-base for more examples.

stephanegigandet:
Can you rename allergen:en: to allergens:en: ? We may not have many ingredients with multiple allergens, but it makes things easier if we use the name of the taxonomy (allergens in plural).

stephanegigandet:
Also allergen:en: en:gluten (with a space) is a bit easier to read.

aleene:
For compound ingredients, we might set the allergen to maybe, otherwise to yes and a child to no. Similar to the approach used for vegan.

The syrups should be moved the ingredients processing and can then be deleted from the ingredients.

stephanegigandet:
> The syrups should be moved the ingredients processing and can then be deleted from the ingredients.

I don't think we should do that, corn syrup is so different than corn that it really makes sense to keep 2 ingredients.

stephanegigandet:
> For compound ingredients, we might set the allergen to maybe, otherwise to yes and a child to no. Similar to the approach used for vegan.

That would mean separate properties for each allergen. I think we need to consider what we want to use this information for. The goal is to be able to identify individual allergens in ingredients list, so that we can for instance bold them automatically (like we do today thanks to the allergens taxonomy). We do not need to identify compounds that may or may not contain an allergen because if they do contain an allergen, it will be listed as an individual ingredient.

It's very different from vegan / vegetarian where we need to assess each and every ingredient listed in order to know if the product is vegan / vegetarian. For allergens, we just need to identify the individual allergens, not compounds that contain them.

aleene:
I disagree. For allergens it is even more important. If you are allergic you might die from it. So if we have any doubt we should show it.

stephanegigandet:
One thing I forgot: the value of the property must match the canonical name in the allergens taxonomy. e.g. en:nuts (in plural)

aleene:
> The syrups should be moved the ingredients processing and can then be deleted from the ingredients.
I don't think we should do that, corn syrup is so different than corn that it really makes sense to keep 2 ingredients.

We can leave exceptions in the ingredients if I remember well.

stephanegigandet:
The allergens:en: property needs to be precise, if there are false positives on compound ingredients (e.g. speculoos made with corn flour), then we can't use it to replace the current allergens taxonomy to identify allergens.

One solution is to introduce another property that indicates high likelyhood (e.g. > 90%) that a compound ingredient contains an allergen. We can use that other allergens likelyhood property to create quality facets / warnings.

e.g.

en:speculoos
likely_allergens:en: en:gluten

Then we can tag all products with likely allergens that don't have the allergen, and review them manually to either add "contains gluten" or "gluten-free" in the labels.

aleene:
> One solution is to introduce another property that indicates high likelyhood (e.g. > 90%) that a compound ingredient contains an allergen. We can use that other allergens likelyhood property to create quality facets / warnings.

e.g.

en:speculoos
likely_allergens:en: en:gluten

Ehm `allergens:en: en:gluten:maybe`

Then we can also do `en:gluten:no`

svensven:
> So if we have any doubt we should show it.

I like the idea of telling a user "you should double check, this thing often contains $allergen, but we're not sure that it does here". But saying it does when it may not feels like lying, and defeating the point of trying to have as accurate data as possible.

svensven:
Although given how incomplete a lot of the data is, I don't think we could ever declare "does not contain $allergen" for most products anyway, unless explicitly labelled by the producer. Often a lack of information is a null rather than a negative.

stephanegigandet:
> Ehm `allergens:en: en:gluten:maybe`

OK, let's try that.

svensven:
re my last, i actually found a product that declares itself allergen-free! https://uk.openfoodfacts.org/product/5012869542015/falafel

taxonomies/ingredients.txt
```diff
@@ -444,7 +444,7 @@ be:каратыноід
 bg:Каротиноид
 ca:Carotenoide, carotenoides
 cs:karotenoidy
-de:Carotinoide, Karotinoide
+de:Carotinoide, Krarotinoide
```
svensven, line 447: is this intentional?

aleene: Looks like a typo

aleene:
> re my last, i actually found a product that declares itself allergen-free
I call that a **producer claim**, which need to be checked. Do we think that ALL *all* the ingredients are allergen free as well.

I agree that we should not claim that a product is allergen free. We can say that we detect nothing. Only if there is a label of a third party certifier has been found, we can do that.

taxonomies/ingredients.txt
```diff
@@ -70984,6 +71022,7 @@ fr:spéculoos
 it:speculoos
 nl:speculoos
 ru:Спекулос
+allergens:en: en:gluten
```
stephanegigandet, line 71025:
```suggestion
allergens:en: en:gluten:maybe
```

There are gluten-free speculoos (I found one made with corn flour)

stephanegigandet (approved):
Thank you!

PR #4161, opened 2020-09-07
ingredients parsing - ignore "allergy advice" in ingredients lists

svensven:
Several of the top unknown ingredients for the UK are variations of these. These patterns match 1423 lines in the CSV dump.

(also a random comment from me working out how the code worked, and how "suitable for vegetarians" could be in an ingredient text but not parsed...)

Looking at the CSV, French has a similar issue ("Pour les allergènes, voir les ingrédients indiqués en gras" etc), but I'm not as familiar with the ingredient conventions. But if this PR is the right way to do this, I could have a go.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -1186,6 +1186,7 @@ sub parse_ingredients_text($) {
 					}
 				}
 
+				# Check ingredient against labels taxonomy. If it exists there, don't parse it as an ingredient, and add it to labels. See also further down.
 				if (defined $labels_regexps{$product_lc}) {
 					# start with uncomposed labels first, so that we decompose "fair-trade organic" into "fair-trade, organic"
 					foreach my $labelid (reverse @labels) {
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -1186,6 +1186,7 @@ sub parse_ingredients_text($) {
 					}
 				}
 
+				# Check ingredient against labels taxonomy. If it exists there, don't parse it as an ingredient, and add it to labels. See also further down.
```
stephanegigandet, line 1189: It's a bit more complex: it checks if it is an ingredient + a label, for a limited subset of labels:
my @labels = ("en:fair-trade-organic", "en:organic", "en:fair-trade", "en:pgi");

e.g. "organic tomatoes" -> we add "en:organic" to labels, and we keep "tomatoes" as the ingredient

```suggestion
				# Check if we have an ingredient + some specific labels like organic and fair-trade.
				# If we do, remove the label from the ingredient and add the label to labels
```

t/ingredients_tags.t
```diff
@@ -121,6 +121,12 @@ my @tests = (
 	[ { lc => "es", ingredients_text => "caramelo E-150c"} , ["en:e150c"]],
 	# mismatch between name and number
 	[ { lc => "fr", ingredients_text => "acide citrique E120"} , ["fr:acide citrique e120"]],
+	
+	# removal of "allergy advice..." in %ignore_regexps
+	[ { lc => "en", ingredients_text => "salt, spice. allergy advice! for allergens, see ingredients in bold, water."}, ['en:salt','en:spice','en:water']],
+	[ { lc => "en", ingredients_text => "salt, spice. allergy advice: for allergens, see ingredients in bold. May contain traces of nuts."}, ['en:salt','en:spice']],
+	[ { lc => "en", ingredients_text => "salt, spice. allergen advice: for allergens including cereals containing gluten, see ingredients in bold. May contain traces of nuts."}, ['en:salt','en:spice']],
```
stephanegigandet, line 128: It would be good to also add the last test to allergens_tags.t to check that we do keep the nuts as traces.

svensven: something like 25d81949a62684c2922144ed62df4ee4a2fd418f ?

stephanegigandet: Perfect, thank you!

PR #4164, opened 2020-09-07
fix .json, fixes #4163

### LINKED ISSUE #4163: .json urls give incorrect results
https://world.openfoodfacts.dev/label/organic.json gives back 0 results, due to a bug in the parsing of json (which is not removed, so this looks for the label "organic-json".

stephanegigandet:
I made another mistake when I changed the way query parameters are parsed. This fixes it.

fixes #4163


teolemon (approved):

teolemon:
What kind of path did the mistake have an impact onto ?

stephanegigandet:
all urls with .json (e.g. /additives.json) except the ones with page numbers like 1.json

PR #4175, opened 2020-09-10
Taxonomy edits 13 - mostly added synonyms/translations

svensven:
removed translations of "grated cheese", in the hope they'll be picked up by ingredients_processing. If that works, there are a few specific variants too, like grated emmental.

"pâte à tarte" was a duplicate.


taxonomies/ingredients.txt
```diff
@@ -9672,15 +9676,6 @@ pt:queijo de pasta dura, queijo duro
...
 
-<en:cheese
-fr:fromage râpé
-ca:formatge ratllat
-de:geriebener Käse
-es:queso rallado
-it:formaggio grattugiato
-nl:geraspte kaas
-# fr:fromage-râpé has 185 products in 4 languages @2108-10-28
-
```
aleene, line 9683: This kind of pruning still needs to be done for everything that occurs in the processing taxonomy.

taxonomies/ingredients.txt
```diff
@@ -9689,6 +9684,34 @@ es:queso suizo
...
+<en:sauce
+en:cheese sauce
+es:salsa de queso
+fi:juustokastike
+fr:sauce au fromage
+nl:kaassaus
+sv:ostsås
+wikidata:en:Q27972032
+# ingredient/en:cheese-sauce has 425 products in 2 languages @2020-06-07
+
```
aleene, line 9696: You could delete this. Should be covered by processing?

taxonomies/ingredients.txt
```diff
@@ -23355,7 +23395,7 @@ fr:blé amidonnier, amidonnier
 ru:двузернянка
 
 <en:cereal
-en:teff
+en:teff, teff seeds
 es:teff, semilla de teff
```
aleene, line 23399: I prefer to keep the seeds separate. Just `teff` does not say anything about form, shape or process.

(but maybe I am exaggerating)

aleene (approved):

taxonomies/ingredients.txt
svensven: Yeah, this was just my first attempt at it, because I noticed the english version was already being handled that way.

taxonomies/ingredients.txt
svensven: oh, the addition was just moving it out of the named-cheeses section. there's a deletion of the same just below.

taxonomies/ingredients.txt
svensven: I actually looked it up on wikipedia at the time, and that says "It is cultivated for its edible seeds, also known as teff", so I just put them together. Presumably any use of the term in ingredients lists would be grain rather than the plant anyway? i dunno. :)

aleene: I would make `teff seeds` a child of `teff`, and `teff flour` a child of `teff`, and `teff seed flour`, a child of `teff seed` and `teff flour`.

svensven: I went through the CSV to see what other terms there were, and just confused myself trying to decide a hierarchy, so I just made them all children of teff.

taxonomies/ingredients.txt
svensven: I was wondering how I could check whether moving ingredients to processing will work other than just pushing to live and hoping, but falling foul of the unit tests just now made me realise I could write ones to check the changes I've made do what I expect, and run them locally on my machine.

Issue #3753, opened 2020-06-26
Problems with accents in Spanish


AcuarioCat:
The following is in ingredients:
es:Fruta, frutas

The following is in ingredients_processing:
es:pure,puré,pure de

'puré de frutas' does not parse 
'pure de frutas' parses ok

'puré frutas' does not parse
'pure frutas' parses ok

svensven:
For some reason, in ingredients_processing.txt if the unaccented version is first, or the only version, it won't match accented terms.
But if the accented version is first, or the only version, that will match both accented and unaccented terms.

So, ```es:puré, pure, puré de, pure de, ...``` or ```es:puré, puré de, ...``` both match all 4 ingredients, 
but ```es:pure, pure de, ...``` doesn't match the accented ingredients.

stephanegigandet:
> For some reason, in ingredients_processing.txt if the unaccented version is first, or the only version, it won't match accented terms.

That sounded very strange, but I check, and indeed, that is the case. What happens is that unaccented synonyms are added automatically, and when an accented synonym is after the un-accented version, it's not added at all.

VaiTon:
Superseeded by #4810.

PR #4184, opened 2020-09-11
ingredients processing - Quick fix for example in #3753

svensven:
Rearranging terms to be accented-first. 

Don't know if there's a wider issue about the parsing of accented terms in ingredients_processing.txt.


stephanegigandet (approved):
Thank you!

PR #4187, opened 2020-09-12
performance - stop Lang.pm sleep(1)-ing every time it's loaded

svensven:
I'm assuming the sleep is just there to make the debug message easier to read?


stephanegigandet (approved):
Looks good to me.

PR #4188, opened 2020-09-12
performance - Make Tags.pm perform its 9600 file-exists checks on demand instead of on load

svensven:
As part of playing with NYTProf, I discovered why unit tests took so long to run for me.
The population of %tags_texts and %tags_levels happens every time Tags.pm is loaded, and number-of-languages * number-of-tag-types = 9600 file exists checks, for variables that are only used in 3 places in Display.pm. 
This probably isn't noticeable most of the time, but I was running over a network share, where this added 39 seconds to running each unit test file...
I took inspiration from https://github.com/openfoodfacts/openfoodfacts-server/pull/2833, but may have got the perl wrong...

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -1731,6 +1731,8 @@ sub display_list_of_tags($$) {
 			$log->debug("missing_property defined", {missing_property => $missing_property});
 		}
 
+		init_tags_texts_levels() unless %tags_levels;
+
 		foreach my $tagcount_ref (@tags) {
 
 			$i++;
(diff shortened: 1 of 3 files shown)


stephanegigandet:
That's a very good idea. In fact in practice, I think we have tags texts only for additives (and 1 category), and only for French. That's a very old system I put in place in 2012 in order to have some info pages for additives, with a color level, that we have been meaning to remove for quite a while...

stephanegigandet (approved):
Looks good to me, thank you!

hangy (approved):

PR #4189, opened 2020-09-12
performance - make build_tags_taxonomy() use string comparisons instead of regexps

svensven:
Building the categories taxonomy, according to NYTProf the modified code is run approx **141 million times**. (3 regexps, so approx 422 million calls.) 
Borrowing a trick from PHP optimisation, I tried replacing the regexp comparisons with builtin string functions. Sadly, perl's builtin functions are more limited, hence the clumsy "does the string end with this" test - "is the start point of the last match + the length of the needle string equal to the length of the haystack string".
For me, this change reduced the taxonomy build times to 1/5 of the original time for all the taxonomies (apart from the tiny ones, where startup time is larger).

It seems to produce identical .results.txt files (and the same complaints/errors to the console), but different but identical-size .sto files. I don't know if that's significant, maybe they're different every time?
And the other major functions seem to be called exactly the same number of times, suggesting the same code flow.

This does of course assume those comparisons never _need_ to be regexps.

### With regexp comparisons:
_Profile of scripts/build_tags_taxonomy.pl for 3770s (of 4412s), executing 1293706561 statements and 854791389 subroutine calls in 911 source files and 285 string evals._
image 
image
image

### With string comparisons:
_Profile of scripts/build_tags_taxonomy.pl for 584s (of 1387s), executing 1716511043 statements and 9193823 subroutine calls in 911 source files and 285 string evals._
image
image
image

Output files:
image

### DIFF
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -1080,16 +1080,21 @@ sub build_tags_taxonomy($$$) {
 
 					# replace whole words/phrases only
 
-					if ($tagid =~ /-${tagid2}-/) {
+					# String comparisons are many times faster than the regexps, as long as tags only ever need simple string matching.
+					#if ($tagid =~ /-${tagid2}-/) {
+					if (index($tagid, "-${tagid2}-") >= 0) {
 						$replace = "-${tagid2}-";
 						$before = '-';
 						$after = '-';
 					}
-					elsif ($tagid =~ /-${tagid2}$/) {
+					#elsif ($tagid =~ /-${tagid2}$/) {
+					# despite how convoluted it is, this is still faster than the regexp.
+					elsif (rindex($tagid, "-${tagid2}") + length("-${tagid2}") == length($tagid)) {
 						$replace = "-${tagid2}\$";
 						$before = '-';
 					}
-					elsif ($tagid =~ /^${tagid2}-/) {
+					#elsif ($tagid =~ /^${tagid2}-/) {
+					elsif (index($tagid, "${tagid2}-") == 0) {
 						$replace = "^${tagid2}-";
 						$after = '-';
 					}


svensven:
this is what i was hoping to see :D
my other pull requests:
image
image
this one:
image

Ban3:
I guess there's no way avoiding the regex compilation because of the variables used? That's probably the cause for the slowness because those regexes never actually get to the regex engine but are optimized to substr searches:

```
Matching REx "-lovech-" against "sandanski"
UTF-8 pattern and string...
Intuit: trying to determine minimum start position...
  doing 'check' fbm scan, [0..9] gave -1
  Did not find anchored substr "-lovech-"...
Match rejected by optimizer
```

stephanegigandet:
WOW, very cool results! That's going to make a lot of people very happy! :)

> It seems to produce identical .results.txt files (and the same complaints/errors to the console), but different but identical-size .sto files. I don't know if that's significant, maybe they're different every time?

From the Storable doc, it seems you can set $Storable::canonical = 1 to force the .sto file to have sorted keys, that should allow to verify that the resulting structures are exactly the same:

"Normally Storable stores elements of hashes in the order they are stored internally by Perl, i.e. pseudo-randomly. If you set $Storable::canonical to some TRUE value, Storable will store hashes with the elements sorted by their key. This allows you to compare data structures by comparing their frozen representations (or even the compressed frozen representations), which can be useful for creating lookup tables for complicated queries."

https://metacpan.org/pod/Storable#CANONICAL-REPRESENTATION

stephanegigandet (approved):
Amazing results, thank you very much!

stephanegigandet:
@hangy also used that rindex trick to speed up the Store::get_string_id_for_lang function that we use billions of times (during the taxonomy building but also very extensively when we analyze ingredients, serve queries etc.).

hangy (approved):

svensven:
I put ```$Storable::canonical = 1;``` in build_tags_taxonomy.pl and the .sto files came out identical in both versions.

PR #4195, opened 2020-09-13
declare &init_tags_texts_levels

stephanegigandet:
so that it can be called from Display.pm without prefixing with ProductOpener::Tags::
Otherwise we get an error when accessing a product page:

Software error:
Undefined subroutine &ProductOpener::Display::init_tags_texts_levels called at /srv/off/lib/ProductOpener/Display.pm line 7549.

### DIFF
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -100,6 +100,8 @@ BEGIN
 		&compute_field_tags
 		&add_tags_to_field
 
+		&init_tags_texts_levels
+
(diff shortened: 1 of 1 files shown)


svensven (approved):
I did say I may have got the perl wrong :D

svensven:
ahh, that'll also be why i had to refer to it by the package name in the unit test file.

stephanegigandet:
> ahh, that'll also be why i had to refer to it by the package name in the unit test file.

right, the other option is to use the package name.

PR #4196, opened 2020-09-13
Ingredients.pm - try to silence a repeated warning in unit tests

svensven:
"Use of uninitialized value in concatenation (.) or string at /home/runner/work/openfoodfacts-server/openfoodfacts-server/lib/ProductOpener/Ingredients.pm line 4764."
Caused by $product_ref->{$field} being undefined.


Ban3:
Perl actually has specific defined-or operator `//` you might want to use. Probably doesn't matter much in this case, but any false value im `$product_ref->{$field}` will cause the right hand side to be evaluated. [perlop](https://perldoc.perl.org/perlop.html#Logical-Defined-Or)

stephanegigandet (approved):

Issue #4198, opened 2020-09-13
Cause of "uninitialized value $canon_tagid" warnings on building taxonomy


svensven:
The "Use of uninitialized value $canon_tagid in hash element at lib/ProductOpener/Tags.pm line 1321, <$IN> line 14442." lines on building the ingredients taxonomy (and possibly others) come from uncommented ```description:en:...``` and ```comment:en:...``` lines in the text file. ("line 14442" refers to the input txt file.)

Presumably because they're usually above the entry they refer to, sometimes with extra newlines, the script doesn't know what entry ($canon_tagid) to attach them to. All but one of each type just get discarded, and the result file just has the last instance of each line type near the top:
```
comment:en: Maybe it is possible to distinguish between ice and ice cream by look if it is a compound ingredient.
description:en: SOY SAUCE is a liquid condiment of Chinese origin, made from a fermented paste of soybeans, roasted grain, brine, and Aspergillus oryzae or Aspergillus sojae molds.
```

As they're being discarded anyway, silencing the error could just be as simple as wrapping the lines in ```if (defined $canon_tagid)``` or whatever. 
Or maybe warning about those description/comment lines so they can be commented out.
Or something else?

svensven:
per https://github.com/openfoodfacts/openfoodfacts-server/pull/4096#pullrequestreview-488775044 those lines aren't supposed to be where they are, at least uncommented, so warning about them is the correct solution.

PR #4205, opened 2020-09-14
Building Taxonomies - log the fact we're discarding lines, rather than just causing an unhelpful perl warning.

svensven:
Improvement for #4198, until someone decides the best way to handle those lines.


stephanegigandet (approved):
Thank you!

PR #4211, opened 2020-09-15
performance - Move two Storable loads from BEGIN to INIT phase, to speed up syntax check

svensven:
The variables packager_codes and geocode_addresses are only used from within functions in Display.pm, not top level code that runs during BEGIN, and in cgi/ & scripts/ code that will only use them on executing anyway.

For one thing, this shaves approx 1 second off syntax checking each script in the pull request workflow, and there are 150 scripts...

(Usual perl-n00b caveat applies, I'm getting my understanding from https://perldoc.perl.org/perlmod.html#BEGIN%2c-UNITCHECK%2c-CHECK%2c-INIT-and-END )

Neither of the variables are referenced directly in any unit tests, but they're loaded conditionally anyway.
```
if (-e "$data_root/packager-codes/packager_codes.sto") {
```

The other big chunk of Storable::* is the loop of ``` retrieve_tags_taxonomy($taxonomyid); ``` in Tags.pm, but it populates variables that other BEGIN-time code relies on, so moving that to INIT breaks things, at least without tackling a lot more of the code.

Doing ``` time perl -d:NYTProf -c -CS -Ilib cgi/change_password.pl ```:

Before:
image

After:
image

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -5597,16 +5597,21 @@ sub get_canon_local_authority($) {
 	return $canon_local_authority;
 }
 
-if (-e "$data_root/packager-codes/packager_codes.sto") {
-	my $packager_codes_ref = retrieve("$data_root/packager-codes/packager_codes.sto");
-	%packager_codes = %{$packager_codes_ref};
-}
+# Slow, and the variables are only used from within functions or non-resident scripts,
+# so only run this when actually executing (INIT), not just checking syntax (BEGIN).
+INIT {
 
-if (-e "$data_root/packager-codes/geocode_addresses.sto") {
-	my $geocode_addresses_ref = retrieve("$data_root/packager-codes/geocode_addresses.sto");
-	%geocode_addresses = %{$geocode_addresses_ref};
-}
+	if (-e "$data_root/packager-codes/packager_codes.sto") {
+		my $packager_codes_ref = retrieve("$data_root/packager-codes/packager_codes.sto");
+		%packager_codes = %{$packager_codes_ref};
+	}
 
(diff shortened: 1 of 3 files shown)


svensven:
Never mind, mod_perl doesn't load it if it's in INIT. 
I suppose it could get the same treatment as init_emb_codes(), and be a function called in startup_apache.pl.

hangy:
> Never mind, mod_perl doesn't load it if it's in INIT. 
> I suppose it could get the same treatment as init_emb_codes(), and be a function called in startup_apache.pl.

Probably, but there could be some scripts that need emb codes (madenaerme comes to my mind). Maybe it could be interesting to lazily execute init_emb_codes() where they get accessed, and are not loaded, yet. startup_apache could do that greedily, because it knows that the website needs that data …

svensven:
I mean, it'd be nice not to have to load packager_codes.sto in one big lump at all. It seems to be responsible for 1.2GB of the 3.2GB RAM apache uses in the dev docker. And loading it on the fly would be slooow.
See https://openfoodfacts.slack.com/archives/CELR31Q83/p1600172810013000

svensven:
This version seems to do the trick.

hangy:
> I mean, it'd be nice not to have to load packager_codes.sto in one big lump at all. It seems to be responsible for 1.2GB of the 3.2GB RAM apache uses in the dev docker. And loading it on the fly would be slooow.
> See https://openfoodfacts.slack.com/archives/CELR31Q83/p1600172810013000

Yeah if it uses that much RAM, it should be interesting to store that elsewhere, and possibly use a smaller version for local development.

stephanegigandet (approved):
I'm not familiar at all with module execution and the INIT, BEGIN blocks etc. But from what I understand, this looks good to me.

PR #4213, opened 2020-09-15
Add HTTP caching header to suggest.pl

svensven:
There's no need for the browser to fetch the same data over and over for tags that only change occasionally.

image
vs
image


svensven:
(the choice of 1 day is arbitrary, btw. an hour or a minute would still be better than nothing.)

stephanegigandet (approved):
Looks good!

PR #4214, opened 2020-09-16
Sort processing regexes by length

Ban3:
**Description:**

Currently only the synonyms for a single process are in sorted order, but the whole array for a language is not. I bumped into this while enabling processing for the Nordic languages and couldn't match `fi:jauhettu` because `fi:jauhe` appeared before it.

I removed other `sort`s from the subroutine because I couldn't see the point in sorting the canonical tags or the languages.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -468,34 +468,32 @@ my %ingredients_processing_regexps = ();
 
 sub init_ingredients_processing_regexps() {
 
-	foreach my $ingredients_processing (sort { (length($b) <=> length($a)) || ($a cmp $b) } keys %{$translations_to{ingredients_processing}}) {
+	foreach my $ingredients_processing ( keys %{ $translations_to{ingredients_processing} } ) {
 
-		foreach my $l (sort keys %{$translations_to{ingredients_processing}{$ingredients_processing}}) {
+		foreach my $l ( keys %{ $translations_to{ingredients_processing}{$ingredients_processing} } ) {
 
 			defined $ingredients_processing_regexps{$l}  or $ingredients_processing_regexps{$l}  = [];
-
-			my %synonyms = ();
 
 			# the synonyms below also contain the main translation as the first entry
 
 			my $l_ingredients_processing = get_string_id_for_lang($l, $translations_to{ingredients_processing}{$ingredients_processing}{$l});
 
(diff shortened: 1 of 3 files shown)


Ban3:
I don't know why this test ever worked:
https://github.com/openfoodfacts/openfoodfacts-server/blob/fd47e6b35f163a4d4b986f2c9b3bafbbcd665ffb/t/ingredients_tags.t#L72

because the ingredients taxonomy has:
```txt
<en:chicken fillet
fr:filet de poulet traité en salaison cuit
nl:verhitte gepekelde kipfilet
# ingredient/fr:filet-de-poulet-traité-en-salaison-cuit has 4 products @2019-05-19
```
and it's supposed to stop stripping processes when it finds a match from the ingredients.

I updated the test instead of removing the ingredient because I don't think the dutch translation is in the processing taxonomy.

stephanegigandet:
> I removed other `sort`s from the subroutine because I couldn't see the point in sorting the canonical tags or the languages.

It was to remove some randomness. The languages ordering didn't change the result, but it ensured we always do the processing in the same order (which can be useful if we add debug statements).

The canonical tags ordering did change the results, as we didn't sort the whole array of regexps afterwards. I think it's why this test worked before:

 [ { lc => "fr", ingredients_text => "émincé de filet de poulet traité en salaison cuit rôti, Pourcentages exprimés sur les pâtes alimentaires aux oeufs "}, ["en:chicken-fillet" ], ], 

We remove the processings one by one, and stop when we find a known entry. But we do it in the order of the regexes, so it can go different ways, for instance:

"émincé de filet de poulet traité en salaison cuit rôti"
"émincé de filet de poulet traité en salaison cuit"
"émincé de filet de poulet traité en salaison"
"émincé de filet de poulet"
"filet de poulet"

or

"émincé de filet de poulet traité en salaison cuit rôti"
"filet de poulet traité en salaison cuit rôti"
"filet de poulet traité en salaison cuit"

stephanegigandet (approved):
Looks good to me.

Could you add a test for this? So that we don't break it at another point.

"I bumped into this while enabling processing for the Nordic languages and couldn't match fi:jauhettu because fi:jauhe appeared before it."

Ban3:
> Could you add a test for this? So that we don't break it at another point.

I couldn't see good examples of it in DE or NL, so I just added the FI test. It's commented out right now because I'd like to do a bit more testing on the language before I enable processing for it.

Issue #4220, opened 2020-09-17
Unable to scan the barcode for some uploaded images - ImageMagick CorruptImageProfile


stephanegigandet:
For some images, we get an exception from ImageMagick:

[4561128] /srv/off-pro/lib/ProductOpener/Images.pm 223 ProductOpener.Images {error => 'Exception 365: CorruptImageProfile `xmp\' @ warning/profile.c/SetImageProfileInternal/1704'} cannot read file to scan barcode

This seems to be in fact just a warning, as the image seems to be read correctly anyway.

stephanegigandet:
![IMG_20200810_114851](https://user-images.githubusercontent.com/8158668/93456912-98e82400-f8de-11ea-9f73-9bab4a47e1ec.jpg)

Issue #4218, opened 2020-09-17
Date in image file name mistaken for barcode for bulk photos upload


stephanegigandet:
On the producers platform, when producers import image files in bulk, we try to find the barcode from the file name.

But the same bulk image upload script is also used for scan parties where contributors take lots of photos of food products in the same order: barcode, product front, ingredients, nutrition facts. In that mode, we analyze the image to identify the barcode so that we can put the images on the right products.

Some phones / cameras name photos with the date like this:  IMG_20200810_111707.jpg

In that case, we think 20200810 is the barcode, and all images are assigned to it.

PR #4221, opened 2020-09-17
Fix 2 small bugs for bulk image upload

stephanegigandet:
- dates that look like barcode numbers #4218 
- failing to scan barcode for some images #4220

### DIFF
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -220,11 +220,10 @@ sub scan_code($) {
 	my $x = $magick->Read($file);
 	local $log->context->{file} = $file;
 	
-	# Disabling the check for $x, as ImageMagick triggers an exception for some images
-	# that it can read anyway
-	# Exception 365: CorruptImageProfile `xmp' @ warning/profile.c/SetImageProfileInternal/1704
-	# if ("$x") {
-	if (0) {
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Images.pm
```diff
@@ -219,7 +219,12 @@ sub scan_code($) {
 	my $magick = Image::Magick->new();
 	my $x = $magick->Read($file);
 	local $log->context->{file} = $file;
-	if ("$x") {
+	
+	# Disabling the check for $x, as ImageMagick triggers an exception for some images
+	# that it can read anyway
+	# Exception 365: CorruptImageProfile `xmp' @ warning/profile.c/SetImageProfileInternal/1704
+	# if ("$x") {
+	if (0) {
```
svensven, line 227: You could check for a number above 400, instead of any response.
https://www.imagemagick.org/script/perl-magick.php#exceptions says " A status code less than 400 is a warning. This means that the operation did not complete but was recoverable to some degree. A numeric code greater or equal to 400 is an error and indicates the operation failed completely."

svensven (approved):

lib/ProductOpener/Images.pm
stephanegigandet: I didn't know that, thank you. I'll change the code accordingly.

Issue #4223, opened 2020-09-17
Simplify bulk image upload form


stephanegigandet:
The pro platform has a bulk image upload form, used by producers, and also by contributors who have hundreds of photos to upload.

1. Start upload automatically. Today there is a "Start upload" button that need to be pressed after selecting the files to upload, and it works only for images that the browser has already analyzed, so it's very easy to forget to upload some images if hundreds of images have been selected.

2. Remove the Start upload / Cancel / Close buttons.

![image](https://user-images.githubusercontent.com/8158668/93469884-6a277900-f8f1-11ea-95a8-d284dab4d226.png)

3. Better status messages when there are errors, images already uploaded etc.
- always show the file name
- add link to product if we have a code

![image](https://user-images.githubusercontent.com/8158668/93470005-8fb48280-f8f1-11ea-9ba3-bb3ea91e3d09.png)

Before: first row
After: second row

stephanegigandet:
Deployed on pro platform.

PR #4225, opened 2020-09-17
Issue 4223/simplify bulk image upload

stephanegigandet:
#4223 : auto start uploads, remove buttons to start / cancel upload for each file

### DIFF
--- a/cgi/import_photos_upload.pl
+++ b/cgi/import_photos_upload.pl
@@ -180,10 +180,11 @@ JS
               <span>$Lang{add_photos}{$lang}</span>
               <input type="file" name="files[]" multiple accept="image/*" data-url="/cgi/product_image_import.pl" />
             </span>
+            <!-- automatically start uploads, hide start upload button
             <button type="submit" class="button small btn-primary start">
               @{[ display_icon('arrow_upward') ]}
               <span>$Lang{start_upload}{$lang}</span>
-            </button>
+            </button> -->
             <!-- The global file processing state -->
             <span class="fileupload-process"></span>
           </div>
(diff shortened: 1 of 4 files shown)


cgi/import_photos_upload.pl
```diff
@@ -180,10 +180,11 @@
               <span>$Lang{add_photos}{$lang}</span>
               <input type="file" name="files[]" multiple accept="image/*" data-url="/cgi/product_image_import.pl" />
             </span>
+            <!-- automatically start uploads, hide start upload button
```
hangy, line 183: You can just remove the button. If required, it can be restored from git.

cgi/product_image_upload.pl
```diff
@@ -129,20 +129,29 @@
...
+			filename => $filename . "",	# Make filename a scalar
+	}],
+};
+
 if ((not defined $code) or ($code eq '')) {
 
 	$log->warn("no code");
-	my %response = ( status => 'status not ok');
-	$response{error} = "error - missing product code";
+	$response_ref->{status} = 'status not ok';
```
hangy, line 142: Are we missing translations?

cgi/product_image_upload.pl
```diff
@@ -157,16 +166,19 @@
...
 
 	$log->debug("path determined", { imagefield => $imagefield, path => $path, delete => $delete });
 
 	if ($path eq 'invalid') {
 		# non numeric code was given
 		$log->warn("no code", { code => $code });
-		my %response = ( status => 'status not ok');
-		$response{error} = "error - invalid product code: $code";
-		my $data =  encode_json(\%response);
+		$response_ref->{status} = 'status not ok';
```
hangy, line 178: Translations? (See above)

cgi/product_image_upload.pl
```diff
@@ -304,9 +320,9 @@
 	else {
 
 			$log->warn("no image field defined");
-			my %response = ( status => 'status not ok');
-			$response{error} = "error - imagefield not defined";
-			my $data =  encode_json(\%response);
+			$response_ref->{status} = 'status not ok';
```
hangy, line 323: Translations? (Same as above)

cgi/product_image_upload.pl
stephanegigandet: Those were intended for the API initially so I think we never added translations for them.

cgi/import_photos_upload.pl
stephanegigandet: If we remember that we had it once. :)

cgi/product_image_upload.pl
teolemon: not sure we want to translate error messages

cgi/product_image_upload.pl
teolemon: not sure it's wise, especially people rely on parsing.

teolemon (approved):

cgi/import_photos_upload.pl
stephanegigandet: Removed the commented-out code.

cgi/product_image_upload.pl
stephanegigandet: We now do display the error message to users in the bulk image upload form, so I guess we could translate that. It's much less likely that an app would parse the error messages.

stephanegigandet: I kept the "status not ok" as-is (it's never shown to users), but I added translations for error messages that can be shown to users (we already had some for some other image upload errors).

Issue #1728, opened 2019-03-28
Show barcode image on product page


CharlesNepote:
Barcode image should be displayed on the product page, not only barcode numbers.

**Rationale:**
* it becomes easier to test our mobile apps with products we don't have physically
* it becomes easier to compare Open Food Facts data with other mobile apps
* it makes data more trustable: I can verify the barcode of my product is the same as displayed on the web page
* some people, for whatever reason, could love to print the product page and scan it later
* it makes the web page more graphical
* last, but not least, it becomes easier to provide services for producers which do not have barcodes: we could build special pages for them

Here is **an interesting JS library** that can produce SVG based on the numbers: https://lindell.me/JsBarcode/

sandy9999:
Where can I find the source/script that renders the individual product page?

flyer1996:
it's so difficult!

PR #4227, opened 2020-09-18
Fixed Issue #1728

ArishSultan:
`Show barcode image on product page`

* Added JsBarcode CDN link to the frontend.
* Now the `barcode number` is converted to a vector image.

**Checklist:**

* [ ] Code is well documented
* [ ] Include unit tests for new functionality
* [ ] Code passes Travis builds in your branch
* [ ] If you have multiple commits please combine them into one commit by squashing them.
* [ ] Read and understood the [contribution guidelines](https://github.com/openfoodfacts/openfoodfacts-server/blob/master/CONTRIBUTING.md)

**Description:**

Describe the changes made and why they were made instead of how they were made.

**Related issues and discussion:** #[ISSUE NUMBER]

### DIFF
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -311,6 +311,13 @@ HTML
 
 
 	$scripts .= <<JS
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/jsbarcode@3.11.0/dist/JsBarcode.all.min.js"></script>
+<script type="text/javascript">
+const barcode = document.getElementById('_barcode');
+JsBarcode("#barcode", barcode.innerHTML);
+
+barcode.parentNode.removeChild(barcode);
+</script>
 <script type="text/javascript" src="/js/dist/jquery.iframe-transport.js"></script>
 <script type="text/javascript" src="/js/dist/jquery.fileupload.js"></script>
 <script type="text/javascript" src="/js/dist/load-image.all.min.js"></script>
@@ -445,9 +452,9 @@ sub get_code_and_imagefield_from_file_name($$) {
 	}
 
 	# Check for a specified imagefield
-	
+
 	$filename =~ s/(table|nutrition(_|-)table)/nutrition/i;
-	
+
 	if ($filename =~ /((front|ingredients|nutrition)((_|-)\w\w\b)?)/i) {
 		$imagefield = $1;
 		$imagefield =~ s/-/_/;
@@ -479,7 +486,7 @@ sub process_image_upload($$$$$$$) {
 	my $debug_string_ref = shift;    # to return debug information to clients
 
 	$log->debug("process_image_upload", { product_id => $product_id, imagefield => $imagefield }) if $log->is_debug();
-	
+
 	# The product_id can be prefixed by a server (e.g. off:[code]) with a different $www_root
 	my $product_www_root = www_root_for_product_id($product_id);
 	my $product_data_root = data_root_for_product_id($product_id);
@@ -527,7 +534,7 @@ sub process_image_upload($$$$$$$) {
 			}
 		}
 	}
-	
+
 	local $log->context->{imagefield} = $imagefield;
 	local $log->context->{uploader}   = $userid;
 	local $log->context->{file}       = $file;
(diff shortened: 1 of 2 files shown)


templates/display_product.tt.html
```diff
@@ -52,7 +52,8 @@ <h1 property="food:name" itemprop="name">[% title %]</h1>
 <!-- Display UPC code if the EAN starts with 0 -->
 [% IF upc_code == 'defined' %]
     <p id="barcode_paragraph">
-        [% lang("barcode") %]:  <span id="barcode" property="food:code" itemprop="gtin13" style="speak-as:digits;">[% code %]</span>[% upc %]
+        [% lang("barcode") %]: <p id="_barcode">[% code %]</p>
```
hangy, line 55: Doesn't this nest two `p` elements? I don't think that's a recommendation HTML structure. What's bad about the previous `span`?

templates/display_product.tt.html
```diff
@@ -52,7 +52,8 @@ <h1 property="food:name" itemprop="name">[% title %]</h1>
...
+      <svg id="barcode" property="food:code" itemprop="gtin13" style="speak-as:digits;"></svg>
```
hangy, line 56: I'm not sure if the structured data, and `speak-as:digits` work on `svg` elements. Wouldn't the attributes be more useful on the `id=_barcode` element? Then, the SVG could be added dynamically in JavaScript, if it's required, so that there wouldn't be an empty dangling SVG.

templates/display_product.tt.html
ArishSultan: That p tag is removed by the above javascript code, i actually used it to get the barcode text inside the javascript code i.e. innerHTML

templates/display_product.tt.html
ArishSultan: The JsBarcode library requires the element to exist inside dom and we just provide the id of that element to JsBarcode so, i don't think that creating svg dynamically would be a great idea, since the script tag is placed at the end of the body tag, so i dont think there is a chance of dangling svg.

hangy: I thought there could be a dangling/empty SVG, if the user has JavaScript disabled.

ArishSultan: Ah, i didn't thought about that u are right 😅, i'll fix it, any other problem ?

ArishSultan:
Can someone review it ?

templates/display_product.tt.html
ArishSultan: I have resolved the issues, can u review the updated commit ??

ArishSultan: Please can u approve the latest commit for merge, if there is no issue ?

lib/ProductOpener/Images.pm
```diff
@@ -311,6 +311,25 @@ HTML
...
 	$scripts .= <<JS
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/jsbarcode@3.11.0/dist/JsBarcode.all.min.js"></script>
+<script type="text/javascript">
+// This code will execute only when javascript is enabled.
+
+// Get the barcode `span` element from DOM.
+const barcode = document.getElementById('barcode');
+
+// Get the barcode_paragraph `p` element from DOM.
+const barcodeParagraph = document.getElementById('barcode_paragraph');
```
hangy, line 322: The `barcode_paragraph` element does not exist, if the barcode doesn't start with a `0`, so that should likely be checked before doing all of this. Maybe even the `JsBarcode.all.min.js` script only needs to be included, if we know the barcode will be replaced by an SVG.

hangy: Also, it may need to be limited to some page types, because right now, the barcode is also replaced if I still have to log in: image

ArishSultan: Can you mention the pages, on which the `barcode` is needed

hangy: @stephanegigandet @teolemon Where did you initially want the barcode to be displayed?

ArishSultan: No one is replying 🙁

stephanegigandet: Hi @ArishSultan, sorry for my late reply. We should show the barcode only the product info page: https://world.openfoodfacts.org/product/5022019160310/organic-ginger-beer-belvoir

Ideally it would appear when we click on a barcode button.

ArishSultan: Ok, i am on it

teolemon:
Did @ArishSultan 's latest commit fix your concerns about the display only on the product page ? 
@hangy

PR #4228, opened 2020-09-19
Remove unchecked donation banner

hangy:
I think the checkbox with "Hey, a box to uncheck?!" is not really useful and disturbs the flow. Therefore, I'd like to completely hide the banner and checkbox once it's unchecked. Since we don't need the HTML then, we can also just not add it to the DOM in the first place, when the site is loaded the next time.


stephanegigandet (approved):
Looks good to me. We will probably create some new banners very soon (and we'll probably want to show them even if the box has been checked: the checkbox should be banner specific)

PR #4230, opened 2020-09-19
Remove title and metadata from SVGs

hangy:
When the SVG gets inlined, search engines actually index this, so that search keys like "IcoFont Icons" are uselessly used for all OFF pages.


stephanegigandet (approved):
Thanks!

PR #4231, opened 2020-09-19
performance improvements for scripts/update_all_products.pl

svensven:
Some low-hanging-fruit in avoiding looping over CORE:regcomp. It feels like there's more to squeeze out, but I don't know the data structures well enough. It's also hard to compare like with like, if you let the code actually update the stored data.

### 1)
A section that still has slow regexes looped over a lot is the
```
($new_ingredient =~ /(^($regexp)\b|\b($regexp)$)/i)
($new_ingredient =~ /($regexp)$/i) )
($new_ingredient =~ /(^($regexp)|($regexp)$)/i)
```
expressions in: 
https://github.com/svensven/openfoodfacts-server/blob/01235cb7d8e77b7d62e44efba8528898e828ba19/lib/ProductOpener/Ingredients.pm#L1245-L1263

### 2)
Also, preparse_ingredients_text() is called twice, presumably with the same inputs:
```
# spent 8.49s (914ms+7.57) within ProductOpener::Ingredients::preparse_ingredients_text which was called 490 times, avg 17.3ms/call: 
# 245 times (519ms+4.15s) by ProductOpener::Ingredients::parse_ingredients_text at line 823, avg 19.1ms/call 
# 245 times (395ms+3.42s) by ProductOpener::Ingredients::extract_ingredients_classes_from_text at line 3984, avg 15.6ms/call
```
from update_all_products.pl:
```
			extract_ingredients_from_text($product_ref);
			extract_ingredients_classes_from_text($product_ref);
```
I thought about trying a static/state variable, but it would have to be done in a way that didn't just keep consuming more RAM.

### Pictures
example before & after:

before
after

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -1252,7 +1252,7 @@ sub parse_ingredients_text($) {
 									
 									#  match after, do not require a space
 									# currently no language
-									or ( ($product_lc eq 'xx') and ($new_ingredient =~ /($regexp)$/i) )
+									#or ( ($product_lc eq 'xx') and ($new_ingredient =~ /($regexp)$/i) )
 									
 									#  Dutch: match before or after, do not require a space
 									or (
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -4412,16 +4412,27 @@ sub extract_ingredients_classes_from_text($) {
 					#$product_ref->{$tagtype . "_debug_ingredients_ids" } .= " -> no exact match ";
 
 					foreach my $id (@{$ingredients_classes_sorted{$class}}) {
-						if (($ingredient_id =~ /^$id\b/) and (not defined $seen{$ingredients_classes{$class}{$id}{id}})) {
+						
+						if (index($ingredient_id, $id) == 0) {
+							# only compile the regex if we can't avoid it
+							if (
+								($ingredient_id =~ /^$id\b/)
+								and (not defined $seen{$ingredients_classes{$class}{$id}{id}})
+							) {
 
-							next if (($ingredients_classes{$class}{$id}{id} eq 'huile-vegetale') and (defined $all_seen{"huile-de-palme"}));
+								next if (
+									($ingredients_classes{$class}{$id}{id} eq 'huile-vegetale')
+									and (defined $all_seen{"huile-de-palme"})
+								);
 
-							#$product_ref->{$tagtype . "_debug_ingredients_ids" } .= " -> match $id - $ingredients_classes{$class}{$id}{id} ";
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -1244,13 +1244,22 @@ sub parse_ingredients_text($) {
...
+									(
+										#($product_lc =~ /^(en|es|it|fr)$/)
+										( ($product_lc eq 'en') or ($product_lc eq 'es') or ($product_lc eq 'fr') or ($product_lc eq 'it') )
+										and ($new_ingredient =~ /(^($regexp)\b|\b($regexp)$)/i)
+									)
+									
 									#  match after, do not require a space
 									# currently no language
-									or  (($product_lc =~ /^(xx)$/) and ($new_ingredient =~ /($regexp)$/i))
+									or ( ($product_lc eq 'xx') and ($new_ingredient =~ /($regexp)$/i) )
```
stephanegigandet, line 1255: In fact that line can be commented out (there's no language "xx", it's just there in case for some languages we need to match after, but currently we don't do it).

scripts/update_all_products.pl
```diff
@@ -203,7 +207,9 @@
 	and (not $remove_team) and (not $remove_label) and (not $remove_nutrient)
 	and (not $mark_as_obsolete_since_date)
 	and (not $assign_categories_properties) and (not $restore_values_deleted_by_user) and not ($delete_debug_tags)
-	and (not $compute_codes) and (not $compute_carbon) and (not $check_quality) and (scalar @fields_to_update == 0) and (not $count) and (not $just_print_codes)) {
+	and (not $compute_codes) and (not $compute_carbon) and (not $check_quality) and (scalar @fields_to_update == 0) and (not $count) and (not $just_print_codes)
+	and (not $mongodb_to_mongodb)
```
stephanegigandet, line 211: The purpose of this check is to make sure that the user specified something to update. mongodb to mongodb is more of an option, so specifying it should not prevent the check from verifying we'll actually update something.

stephanegigandet:
Looks very good to me.

Regarding preparse_ingredients_text() being called twice, it definitely does not make sense, as the 2 functions that call it are always called one after the other anyway.

One way could be to change the function to pass $product_ref to it, so that we can store the results in $product_ref->{ingredients_text_preparsed}.
Then we would need to remove it after calling the 2 functions, so that we don't store it in mongodb and the .sto files.

lib/ProductOpener/Ingredients.pm
svensven:
```suggestion
									#or ( ($product_lc eq 'xx') and ($new_ingredient =~ /($regexp)$/i) )
```

scripts/update_all_products.pl
svensven:
```suggestion
```

stephanegigandet (approved):

Issue #4004, opened 2020-08-08
The Google Translate link is not working anymore in the Taxonomy translation system


teolemon:
The Google Translate link is not working anymore in the Taxonomy translation system. We're not passing the original text in the Google translate URL anymore.

PR #4235, opened 2020-09-20
fix google translate link in taxonomy translation page - for #4004

svensven:
Finding synonyms of each term was using the wrong key to look up entries in the hash, thus not finding any, and the Translate link is populated with the list of synonyms.

Although now it's ugly, because the first shown synonym is the term itself. I'm not sure what the correct behaviour is re link and synonyms:

image


stephanegigandet (approved):
Thank you!

PR #4236, opened 2020-09-20
Improve error handling in update_packager_codes_pl, refresh DE codes.

Ban3:
**Description:**

The script was ignoring errors silently and dropping the remaining CSV file after
a problematic line. Enable the `auto_diag` parameter to automatically report any
errors, and only discard the line with the error instead of the remaining file.

Also fixes the errors that were triggering this behavior:
- DE: It hasn't been updated in a while, so just recreate the file and fix the errors at the same time.
- EE: Correctly quote fields containing quote character.


stephanegigandet (approved):
Thank you!

Issue #4121, opened 2020-09-01
Personalized Search - Javascript UI to rank and filter search results according to local user preferences


stephanegigandet:
Allow users to see search results ranked and filtered according to locally saved user preferences.

- User preferences are set by #4119 
- Corresponding product attributes #4113 are returned by the new search API #4039 
- Each returned product is scored according to user preferences + product attributes using a JS library #4120 
- Search results are displayed according to the computed score

Ideally: display the UI to select preferences on top of the search results, so that the user can change the preferences and see the result list being reordered accordingly.

stephanegigandet:
Deployed.

PR #4240, opened 2020-09-21
New /search endpoint + start of local display of results for personal search

stephanegigandet:
- Displays warnings and descriptions for attributes in the user preferences selection UI
- Creates a new /search endpoint for the new search API (with a dedicated display_search_results() function instead of abusing the display_tag() function
- Solves small issues for pagination
- Start of the /search?user_preferences=1 interface to have the client request search results from the API, and ranking, filtering and displaying on the client. #4121

### DIFF
--- a/cgi/display.pl
+++ b/cgi/display.pl
@@ -94,7 +94,7 @@ if (defined $request{api}) {
 	}
 }
 elsif (defined $request{search}) {
-	display_tag(\%request);
+	display_search_results(\%request);
 }
 elsif (defined $request{text}) {
 	display_text(\%request);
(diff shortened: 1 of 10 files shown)


stephanegigandet:
I have something that is starting to work: it's possible to set some preferences, and to get search results that are ranked and filtered according to those search preferences.

image

You can test it on the dev server:
https://world.openfoodfacts.dev/search?categories_tags_en=sweet%20spreads&user_preferences=1

There are still lots of things missing:
- many strings are currently hardcoded in English
- many attributes titles, descriptions
- many attributes computations (e.g. allergens)
- there's no design of the selected preferences, preferences selection form, and search results

svensven (approved):
TIL `</li>` [is optional in html5](https://www.w3.org/TR/2014/REC-html5-20141028/syntax.html#optional-tags). i used to write xhtml back in the day.

PR #4246, opened 2020-09-22
Renamed German category to 'Müsli'

KommX:
Renamed German category to 'Müsli', this is the word everyone in Germany uses, see https://de.wikipedia.org/wiki/M%C3%BCsli

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -40566,7 +40566,7 @@ bg:Мюсли
 ca:Musli
 cs:müsli
 da:Müsli
-de:Müsli, Müesli
+de:Müslis, Müeslis
 eo:Muslio
 es:Muesli, Mueslis
 et:Müsli
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -40566,7 +40566,7 @@ bg:Мюсли
 ca:Musli
 cs:müsli
 da:Müsli
-de:Müesli, Müslis
+de:Müsli, Müesli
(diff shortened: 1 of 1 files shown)


taxonomies/categories.txt
```diff
@@ -40566,7 +40566,7 @@ bg:Мюсли
 ca:Musli
 cs:müsli
 da:Müsli
-de:Müesli, Müslis
+de:Müsli, Müesli
```
hangy, line 40569: We kinda want plurals as the primary name, where they exist. Maybe use Müsli as an alias?

KommX: I don't know too much about the inside of openfoodfacts, but I hoped to achieve that I can select "Müsli" (or "Müslis", I don't care) as a category. Right now, I can't do that, and I have to enter "Müesli" which is a word I've never heard before.

hangy: Yep, me neither. Never heard of that spelling. 🤷‍♂️

svensven: blame the swiss, that wiki link above says "...in der Schweiz und im südlichen Baden-Württemberg Müesli..." :)

KommX: So, I changed the names to their plural forms in an additional commit. I hope, I did it right.

svensven: looks good. thanks :)

svensven (approved):

PR #4249, opened 2020-09-23
Experimentally add some TODO tests

svensven:
The testing process should allow TODO tests to fail without complaining, but they're a way to track known issues, and see if they get fixed in the process of doing something else.
https://perldoc.perl.org/Test/More.html#*TODO%3a-BLOCK*

Also normalised the leading whitespace in the test files to be the tabs .editorconfig says they should be. 
(We could probably lose the first level or two of indentation, as in some other test files, but I thought I'd at least start out trying to be Correct.)

### DIFF
--- a/t/additives.t
+++ b/t/additives.t
@@ -25,15 +25,15 @@ diag explain $product_ref->{additives};
 # vitamine C is not used as an additive (no fuction)
 
 is_deeply($product_ref->{additives_original_tags}, [
-                              ],
+	],
 );
 
 is_deeply($product_ref->{vitamins_tags}, [
-        "en:thiamin",
-        "en:vitamin-c",
-        "en:vitamin-e",
-        "en:riboflavin",
-                              ],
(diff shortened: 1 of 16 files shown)


stephanegigandet (approved):
Very good idea, thank you.

teolemon:
I'm trying to see where we can see test "fail" 
I tried in https://github.com/openfoodfacts/openfoodfacts-server/pull/4249/checks?check_run_id=1152691439

teolemon:
I'm also going to create a label for bugs that can be turned into TODO tests

Ban3:
> I'm trying to see where we can see test "fail"
> I tried in https://github.com/openfoodfacts/openfoodfacts-server/pull/4249/checks?check_run_id=1152691439

Is this what you're looking for? https://github.com/openfoodfacts/openfoodfacts-server/pull/4249/checks?check_run_id=1152691439#step:5:2613

PR #4253, opened 2020-09-24
add w/ & w/o abbreviations, fixes #3270

### LINKED ISSUE #3270: Ingredient analysis parsing: cocoa powder processed w/alkali ->with alkali
https://world.openfoodfacts.org/product/0858687003001/brownie-mix-moon-rabbit-foods

svensven:
seems to work?

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -284,6 +284,8 @@ da => [
 ],
 
 en => [
+	["w/o", "without"],
+	["w/", "with "], # note trailing space
 	["vit.", "vitamin"],
 
 ],
--- a/t/ingredients_parsing.t
+++ b/t/ingredients_parsing.t
@@ -263,24 +263,24 @@ my @lists =(
 	["es", "E102(i)-E101i", "e102i - e101i"],
 	["es", "E102(i)", "e102i"],
 	["es", "S.I.N.:160 b", "e160b"],
-	["pt", "estabilizadores (E 422, E 412)","estabilizadores (e422, e412)"],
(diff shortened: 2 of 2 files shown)


Ban3 (approved):
Looks like it works nicely. Now I'm thinking how are these going to analyzed. Could the processing taxonomy handle things like `(cooked without salt)` or `(processed with alkali)`?

svensven:
> Looks like it works nicely. Now I'm thinking how are these going to analyzed. Could the processing taxonomy handle things like `(cooked without salt)` or `(processed with alkali)`?

i was actually wondering how far the processing taxonomy is supposed to go, before i started looking at removing any ingredients that could be handled by processing instead.

i can understand physical transformations like "chopped" being stripped out of ingredients, but some of them seem more "chemical" transformations, or ones where maybe there's a significance to the transformation that someone might want to look for that version specifically rather than the base ingredient? (can't think of the example that occurred to me a while back)

stephanegigandet (approved):
Looks good, thank you!

stephanegigandet:
> i can understand physical transformations like "chopped" being stripped out of ingredients, but some of them seem more "chemical" transformations, or ones where maybe there's a significance to the transformation that someone might want to look for that version specifically rather than the base ingredient? (can't think of the example that occurred to me a while back)

Right, we need to be careful not too put things that completely change the nature of the ingredient. e.g. "XYZ natural flavouring" should not be considered the same as "XYZ" with some processing.

Also currently in MongoDB we index the base ingredients in ingredients_tags, but we could try to support ways to query products to find all the ones that contain chopped tomatoes. e.g. something like /ingredient/tomatoes:chopped maybe.

PR #4258, opened 2020-09-26
performance - gen_top_tags_per_country.pl

svensven:
I tried running this script with both a 10k and 100k limit on what it queries from mongodb, and either way, it seems to spend a large chunk of its time on generating "Use of uninitialized value" warnings. Maybe those are down to not having the full dataset from mongodb, I don't know. The 10k run generated 262MB of stderr output.

```
1000086  Use of uninitialized value in numeric comparison (<=>) at scripts/gen_top_tags_per_country.pl line 710.
 555391  Use of uninitialized value in addition (+) at scripts/gen_top_tags_per_country.pl line 490.
 193331  Use of uninitialized value in numeric comparison (<=>) at scripts/gen_top_tags_per_country.pl line 854.
   1320  Use of uninitialized value $value in string ne at /opt/product-opener/lib/ProductOpener/Display.pm line 8581.
   1320  Use of uninitialized value $value in pattern match (m//) at /opt/product-opener/lib/ProductOpener/Display.pm line 8577.
   1320  Use of uninitialized value in pattern match (m//) at scripts/gen_top_tags_per_country.pl line 446.
...
```
image

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -8574,11 +8574,7 @@ sub add_product_nutriment_to_stats($$$) {
 	my $nid = shift;
 	my $value = shift // '';
 
-	if (lc($value) =~ /nan/) {
-
-		return -1;
-	}
-	elsif ($value ne '') {
+	if ($value ne '') {
 
 		if (not defined $nutriments_ref->{"${nid}_n"}) {
 			$nutriments_ref->{"${nid}_n"} = 0;
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -8572,9 +8572,9 @@ sub add_product_nutriment_to_stats($$$) {
 
 	my $nutriments_ref = shift;
 	my $nid = shift;
-	my $value = shift;
+	my $value = shift // '';
 
-	if ($value =~ /nan/i) {
+	if (lc($value) =~ /nan/) {
 
 		return -1;
 	}
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -218,7 +218,7 @@ sub normalize_nutriment_value_and_modifier($$) {
 
 	return if not defined ${$value_ref};
 
-	if (${$value_ref} =~ /nan/i) {
+	if (lc(${$value_ref}) =~ /nan/) {
 		${$value_ref} = '';
 	}
 
--- a/scripts/gen_top_tags_per_country.pl
+++ b/scripts/gen_top_tags_per_country.pl
@@ -166,11 +166,11 @@ my $end_t = 1424476800 - 12 * 3600 + 10 * 86400;
 
 
 my %nutrition_grades_to_n = (
-a => 1,
-b => 2,
-c => 3,
-d => 4,
-e => 5,
+	a => 1,
+	b => 2,
+	c => 3,
+	d => 4,
+	e => 5,
 );
 
 while (my $product_ref = $cursor->next) {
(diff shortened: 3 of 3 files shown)


svensven:
### First fixes
1. One of the warnings was easy to pass a default value to.
2. Even though perl optimised out the regexp replace, it still took a long time to replace a space with a new line, possibly just from having to manipulate an ever-growing string. This is from generating the `2132, 2136, 2141, 2180, 2190, 2199, 2202, 2218, 2228, 2232,` style lines in e.g. lang/fr/texts/products_stats_fr.html
image
whereas unless I'm being dense, this does the same, much faster:
image

which brings it down to

image

Ban3:
I can see countries getting sorted indentically at multiple places.

https://github.com/openfoodfacts/openfoodfacts-server/blob/c099820ced4d51802db3baa89f7957469dd7ca98/scripts/gen_top_tags_per_country.pl#L501

https://github.com/openfoodfacts/openfoodfacts-server/blob/c099820ced4d51802db3baa89f7957469dd7ca98/scripts/gen_top_tags_per_country.pl#L523

https://github.com/openfoodfacts/openfoodfacts-server/blob/c099820ced4d51802db3baa89f7957469dd7ca98/scripts/gen_top_tags_per_country.pl#L676

https://github.com/openfoodfacts/openfoodfacts-server/blob/c099820ced4d51802db3baa89f7957469dd7ca98/scripts/gen_top_tags_per_country.pl#L842

1. Can it be reduced to one sort saving the results somewhere?
2. Does it need to be sorted at all?

svensven:
> I can see countries getting sorted indentically at multiple places.
Those sorts aren't the problem...
image

The sorts that are the problem are the ones generating the warnings:

image

image

I don't really understand what these are doing, as 
a) I don't know the data structure being manipulated without getting more in depth, 
b) `$countries_dates{$country}{$date}{$a} <=> $countries_dates{$country}{$b}` looks wrong to me, as one side has {$date} and the other doesn't,
c) `@sorted_dates` isn't used anywhere after them, unless the sort is working on `$countries_dates` in-place?

svensven:
### Question: 
can this be optimised to a literal string eq comparison, with lc() if necessary? Again, I don't know the source data.
image

Ban3:
> c) `@sorted_dates` isn't used anywhere after them, unless the sort is working on `$countries_dates` in-place?

Must be a typo. Sort doesn't work in-place.

svensven:
> 
> 
> > c) `@sorted_dates` isn't used anywhere after them, unless the sort is working on `$countries_dates` in-place?
> 
> Must be a typo. Sort doesn't work in-place.

in which case, those two sort lines don't seem to serve any purpose?

Ban3:
> in which case, those two sort lines (and all the warnings they generate) don't seem to serve any purpose?

Looks like it. Only place where I can see that sort used is here:
https://github.com/openfoodfacts/openfoodfacts-server/blob/a9a9b26f87dd274967ac74f8d0f44872518c007f/scripts/gen_top_tags_per_country.pl#L473-L475

Without delving too much into the code, I'm guessing it has been copy pasted to other places. Or the unsorted array has been accidentally used.

stephanegigandet:
Indeed, only the first @sorted_dates is used, the other 2 buggy ones can be removed. (I'm the one to blame for this code in case you wonder ;) )

stephanegigandet:
> can this be optimised to a literal string eq comparison, with lc() if necessary? Again, I don't know the source data.

The /nan/i was to remove NaN values that were sent by some 3rd party apps at some point. We should just remove those in the database if we still have some, and forbid them. Then we can just get rid of this check.

stephanegigandet:
> 2\. Even though perl optimised out the regexp replace, it still took a long time to replace a space with a new line, possibly just from having to manipulate an ever-growing string.

I guess we could try to just remove the line feeds completely. I probably did it for readability or something, but I'm guessing javascript doesn't mind huge lines.

Ban3:
> can this be optimised to a literal string eq comparison, with lc() if necessary? Again, I don't know the source data.

In the meantime, _if_ it needs to be regex match, would this be any faster?

```diff
-				if ($products_nutriments{$code}{$nid} =~ /nan/i) {
+				if ( lc $products_nutriments{$code}{$nid} =~ /nan/ ) {
```

At least the optimizer can't reject the match because of the `/i` modifier. Simple test

```sh
$ perl -Mre=debugcolor -e '1.01 =~ /nan/i'  
Compiling REx "nan"
Final program:
   1: EXACTFU <nan> (3)
   3: END (0)
stclass EXACTFU <nan> minlen 3 
Matching REx "nan" against "1.01"
Matching stclass EXACTFU <nan> against "1.01" (4 bytes)
Contradicts stclass... [regexec_flags]
Match failed
Freeing REx: "nan"
```

Same thing without the `/i`:

```sh
Compiling REx "nan"
Final program:
   1: EXACT <nan> (3)
   3: END (0)
anchored "nan" at 0..0 (checking anchored isall) minlen 3 
Matching REx "nan" against "1.01"
Intuit: trying to determine minimum start position...
  doing 'check' fbm scan, [0..4] gave -1
  Did not find anchored substr "nan"...
Match rejected by optimizer
Freeing REx: "nan"
```

svensven:
without the sort warnings:
image

svensven:
> In the meantime, _if_ it needs to be regex match, would this be any faster?
> 
> ```diff
> -				if ($products_nutriments{$code}{$nid} =~ /nan/i) {
> +				if ( lc $products_nutriments{$code}{$nid} =~ /nan/ ) {
> ```
> 
> At least the optimizer can't reject the match because of the `/i` modifier. Simple test

ooh, cunning, I'll give it a try. 
I might go back to trying 100k records as well, now it's not going to take as long. (even generating the html report from the .out file took 25 mins)

svensven:
yep, case-sensitive regexp is 3-4 times faster than /i :

image

vs

image

svensven:
I'm not sure I can find anything more performance wise. The slow bits are mongodb and the two stats functions, which don't have any really obvious bottlenecks, they're just being called many millions of times.

Two possible outstanding issues:
1) The links generated in countries.html are nested:
https://github.com/openfoodfacts/openfoodfacts-server/blob/69aa123a3b8a0a78b4800626b2f03c705ee225c9/scripts/gen_top_tags_per_country.pl#L516-L518

2) The yAxis title in products_countries.js is an undefined variable `$Lang{products_p}{$lang}`. Presumably it's supposed to be the translation of the word "Products", but I'm not seeing it in the .po files.


Last run of 100k query:

image

And so to bed.

stephanegigandet:
I verified that we don't have NaN values in the production database, and that apps can't write it anymore. So we can safely remove the /nan/ check.

stephanegigandet (approved):
Thank you!

PR #4259, opened 2020-09-26
Fix use of undefined translation

Ban3:
**Description:**

As pointed out by @svensven in https://github.com/openfoodfacts/openfoodfacts-server/pull/4258#issuecomment-699492604, `products_p` is not defined in the gettext catalog. Switch references to it to the actually defined `products`.

Also, noticed that `$series_n{$seriesid}` can be undefined when using nutrition grade colors for a graph and the grade has 0 products. Initialize it to 0 before interpolating it.


svensven (approved):

hangy (approved):

stephanegigandet (approved):
Thank you!

PR #4261, opened 2020-09-27
clear most of the warnings on the unit tests

svensven:
There are still 2 uses of a "test" taxonomy in tags.t that cause
> Use of uninitialized value $path in concatenation (.) or string at /home/runner/work/openfoodfacts-server/openfoodfacts-server/lib/ProductOpener/Tags.pm line 2171.

Dunno if that's an intentional test of a bogus taxonomy, as the test expects the incorrect URL it gets.

### DIFF
--- a/lib/ProductOpener/ImportConvert.pm
+++ b/lib/ProductOpener/ImportConvert.pm
@@ -695,12 +695,12 @@ drained_weight => '(peso )?(neto )?(escurrido)',
 		}
 
 		# 1/2 , 3/4
-		if ($product_ref->{quantity} =~ /^'?\s*\d+((\/)\d+)\s*$/i) {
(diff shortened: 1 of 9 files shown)


stephanegigandet:
> Dunno if that's an intentional test of a bogus taxonomy, as the test expects the incorrect URL it gets.

The test is intended to test the part that is after /$path/ , using a test taxonomy that has special cases built in. But the issue is that we have not declared the taxonomy in the po/tags/ files, so the $path is not defined. If we do declare it, it would make urls like /test/something valid on the web site, so I'm not sure that's a good idea. Maybe we could just find a way to silence the warnings for that specific test.

stephanegigandet (approved):
Thank you!

PR #4274, opened 2020-09-28
 Taxonomy for the various parts of the environmental impact.

teolemon:
Taxonomy for the various parts of the environmental impact.
From the Excel and http://app.agribalyse.fr/aliments/7615

### DIFF
--- a/taxonomies/agribalyse-step.txt
+++ b/taxonomies/agribalyse-step.txt
@@ -1,0 +1,20 @@
+en:Agriculture
+fr:Agriculture
+
+en:Processing
+fr:Transformation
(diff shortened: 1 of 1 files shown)


stephanegigandet (commented):
At this point, I don't know how we will use those strings exactly.

In general, I believe that we should not create taxonomies until we actually need them and know how they will be used. Otherwise, it makes it difficult to see what taxonomy is actually used, is a WIP, or has been superseded by something else.

In this particular example, there isn't really a hierarchy, so it may make more sense to have the strings in .po files instead of a taxonomy.

Issue #3857, opened 2020-07-16
CSV/JSONL/RDF export controlled


CharlesNepote:
Some people are yelling that the CSV/JSONL export is sometimes not complete. Indeed, it was 200 MB two days ago, 3 GB yesterday, and 140 MB today (2020-07-16). Even if we find a bug, there WILL BE other bugs, so we shall discuss how to produce a good export at least once a week. Here is my proposal (I have tried to make it simple):

1. export to `en.openfoodfacts.org.products.YYYY-MM-DD.csv`
2. control number of CSV lines is not different by 1% than the number of products
3. if there's no difference: move the file to `en.openfoodfacts.org.products.csv`
4. if there're some differences: delete the produced file and keep the last one, and send an alert to #infrastructure slack channel ; if we have alerts two or three successive days, we will be able to see if it's important to fix it
5. the script should return a "0" value when finishing, and it should be launched by another script/system (Minion?) to catch that it ends correctly

CharlesNepote:
Related script is: https://github.com/openfoodfacts/openfoodfacts-server/blob/master/scripts/export_database.pl

stephanegigandet:
Still an issue on most days... :-| 

/srv/off/html/data# wc -l *csv
     25940 be-fr.openfoodfacts.org.products.20190520.csv
    179064 en.openfoodfacts.org.products.csv
    638284 en.openfoodfacts.org.products.nutriscore.csv
         2 en.openfoodfacts.org.products.test.csv
   1454379 fr.openfoodfacts.org.products.csv
         2 fr.openfoodfacts.org.products.test.csv

svensven:
does the process die with any error message?

svensven:
Looking at the timestamps of the various things gen_feeds_daily_off.sh generates, assuming that is the script being used.
Tonight the CSV export seems to have died about 40 mins in. 
And because the french exports haven't been modified for 6 days, it's died during the english phase every night, but at different times, because the size has varied.

If it's a MongoDB problem, I was wondering if either of these might help:
- merging both languages into one combined crawl of the database, generating all 4 en/fr csv/rdf files simultaneously, instead of once for english and once for french, given the actual query is identical.
- querying the database in chunks, say 10k records at a time, looping until there are no more results.

However, I was wondering if it was a RAM problem instead, because perl's keeping many GB of output text in memory somehow while writing the files, and getting OOM-killed.

### Tonight's gen_feeds_daily_off.sh outputs:

[...]

PR #4283, opened 2020-09-29
export_database.pl - only overwrite previous export if the new one is bigger

svensven:
A quick workaround, re #3857.

Although if the script just dies part way through, this won't really help.

(The assumption being that the data set will only ever grow, even with occasional deletions. Manually deleting the old dump should cause the new dump to be used regardless.)

### DIFF
--- a/scripts/export_database.pl
+++ b/scripts/export_database.pl
@@ -129,12 +129,16 @@ foreach my $l ("en", "fr") {
 
 	$langs{$l} = 0;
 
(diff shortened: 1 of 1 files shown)


svensven:
"[categories] child process exited with code null" this time, instead of additives.

svensven:
"[additives] child process exited with code null" this time.

svensven:
dunno what's going on with travis...
"[vitamins] child process exited with code null"

stephanegigandet (approved):
Looks good to me, thank you!

PR #4284, opened 2020-09-30
Added the Swedish word for "cheese" to allergens.

Kezxo:
Not sure how it was missing for so long.
https://world.openfoodfacts.org/allergen/sv:ost

I hope I'm doing this correctly.

### DIFF
--- a/taxonomies/allergens.txt
+++ b/taxonomies/allergens.txt
@@ -227,7 +227,7 @@ ro:lapte, lactoză
 ru:молоко, лактоза, сливки, молочная сыворотка, йогурт
 sk:mlieko, laktózy
 sl:mleko, laktozo
-sv:mjölk, mjölkpulver, skummjölk, skummjölkspulver, mellanmjölk, mellanmjölkspulver, kärnmjölk, standardmjölk, standardmjölkspulver, laktos, grädde, smör, smörfett, fårmjölk, getmjölk, komjölk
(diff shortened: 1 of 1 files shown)


Ban3 (approved):
> I hope I'm doing this correctly.

You are :slightly_smiling_face:

aleene:
Allergens detection is a bit different at the moment. It is now hidden in ingredients.pm (?), and will be moved into the ingredients taxonomy in order to cover all synonyms. Allergens.txt is now only used to provide translations for the allergen categories.

Ban3:
> Allergens detection is a bit different at the moment. It is now hidden in ingredients.pm (?), and will be moved into the ingredients taxonomy in order to cover all synonyms. Allergens.txt is now only used to provide translations for the allergen categories

AFAIK the allergens taxonomy is still used even if there's work to switch to ingredients taxonomy.

stephanegigandet:
Right now, we still use the allergens taxonomy for detection of allergens in the ingredients list. At some point we will probably switch to the new properties of the ingredients taxonomy to do it instead, but it has not been coded yet.

Ban3:
> At some point we will probably switch to the new properties of the ingredients taxonomy to do it instead, but it has not been coded yet.

We also need to remember to move synonyms from allergens taxonomy that are not yet in ingredients.

Issue #4287, opened 2020-09-30
Compute an EcoScore with data from the French Agribalyse food products environmental impacts database


stephanegigandet:
The Agribalyse V3 database has been released as open data. It contains detailed environmental impacts for 2500 food products categories.

In the last few months, we have been adding properties in the categories taxonomy so that we can match our categories with the Agribalyse categories.

This will allow us to compute an EcoScore using the AgriBalyse data, adjusted with other parameters (e.g. the packagings, some labels like the organic label etc.).

The EcoScore will be computed and made available in a similar way than the NutriScore: we will keep all the computation details of every step so that we can explain the score (like we do for the NutriScore), and the score will also be made available to apps as a new product attribute.

VaiTon:
@stephanegigandet @teolemon can we close?

PR #4288, opened 2020-09-30
Initial and partial Eco-Score computation, #4287

stephanegigandet:
See #4287 for an explanation of the Eco-Score.

This is an initial very partial implementation, based only on the product category and the corresponding Agribalyse data.

Next step will be to create an ecoscore facet, similar to the nutriscore facet, so that we can observe and improve the coverage of products with Agribalyse category matches.

Further down the line, we will include other Eco-Score components such as labels, origins of ingredients, packagings etc.


lib/ProductOpener/Ecoscore.pm
```diff
@@ -0,0 +1,240 @@
...
+	if (open (my $io, "<:encoding($encoding)", $agribalyse_details_by_step_csv_file)) {
+
+		# @$headers_ref = $csv->header ($io, { detect_bom => 1 });
+		# the header function crashes with some csv files... use getline instead
+		my $row_ref;
+
+		# Skip 3 first lines
+		$csv->getline ($io);
+		$csv->getline ($io);
+		$csv->getline ($io);
```
Ban3, line 103: I think `getline` parses the line even when called in void context. Maybe just using the `<>` to avoid extra work would be better.

```suggestion
		$. = 0
		do { <$io> } until $. == 3 || eof;
```

lib/ProductOpener/Ecoscore.pm
```diff
@@ -0,0 +1,240 @@
...
+
+	my $csv_options_ref = { binary => 1, sep_char => "," };    # should set binary attribute.
+
+	my $csv = Text::CSV->new ( $csv_options_ref )
+		or die("Cannot use CSV: " . Text::CSV->error_diag ());
+
+	if (open (my $io, "<:encoding($encoding)", $agribalyse_details_by_step_csv_file)) {
+
+		# @$headers_ref = $csv->header ($io, { detect_bom => 1 });
+		# the header function crashes with some csv files... use getline instead
```
Ban3, line 97: Is this related to the `Strings with code points over 0xFF may not be mapped into in-memory file handles` error? I worked around that one in `update_packager_codes.pl` by setting the encoding 
only after the headers were read.

https://github.com/openfoodfacts/openfoodfacts-server/blob/12d17130cb5cca3db85a70cbcfc3a5b6fd5ab5b1/scripts/update_packager_codes.pl#L209

...

https://github.com/openfoodfacts/openfoodfacts-server/blob/12d17130cb5cca3db85a70cbcfc3a5b6fd5ab5b1/scripts/update_packager_codes.pl#L219-L234

lib/ProductOpener/Ecoscore.pm
```diff
@@ -0,0 +1,240 @@
...
+				name_en => $row_ref->[5], # LCI Name
+				dqr => $row_ref->[6], # DQR (data quality rating)
+				# warning: the AGB file has a hidden H column
+				ef_agriculture => $row_ref->[8], # Agriculture
+				ef_processing => $row_ref->[9], # Transformation
+				ef_packaging => $row_ref->[10], # Emballage
+				ef_transportation => $row_ref->[11], # Transport
+				ef_distribution => $row_ref->[12], # Supermarché et distribution
+				ef_consumption => $row_ref->[13], # Consommation
+				ef_total => $row_ref->[14], # Total
```
Ban3, line 117: Alternative to the header is [bind_columns](https://metacpan.org/pod/Text::CSV#bind_columns), which according to the [documentation is also faster than using hash refs](https://metacpan.org/pod/Text::CSV#getline_hr). So if we get the header row like this:

```perl
$. = 0;
do { <$io> } until $. == 7 || eof;

my @hdrs = @{$csv->getline($io)};
my $row_ref = {};
$csv->bind_columns(\@{$row_ref}{@hdrs});
```

This can be changed to this:
```suggestion

		while ($csv->getline ($io)) {
			$agribalyse{$row_ref->{'Code AGB'}} = {
				name_fr => $row_ref->{'Nom du Produit en Français'},
				name_en => $row_ref->{'LCI Name' },
				dqr => $row_ref->{'DQR'}, # data quality rating
				# warning: the AGB file has a hidden H column
				ef_agriculture => $row_ref->{'Agriculture'},
				ef_processing => $row_ref->{'Transformation'},
				ef_packaging => $row_ref->{'Emballage'},
				ef_transportation => $row_ref->{'Transport'}, 
				ef_distribution => $row_ref->{'Supermarché et distribution'},
				ef_consumption => $row_ref->{'Consommation'},
				ef_total => $row_ref->{'Total'},
```
Which is more readable if nothing else.

lib/ProductOpener/Ecoscore.pm
stephanegigandet: I don't remember if there was an error message, I think it was a segmentation fault.

In EcoScore.pm we don't use the header line at all in fact, so I'm going to just remove the comment.
I copied it from Producers.pm (whe we do need the headers line):

````
			# @$headers_ref = $csv->header ($io, { detect_bom => 1 });
			# the header function crashes with some csv files... use getline instead
			my $row_ref;

			while ((not defined $row_ref) and ($row_ref = $csv->getline ($io))) {
			}

			if (defined $row_ref) {

				@{$headers_ref} = @{$row_ref};

				while ($row_ref = $csv->getline ($io)) {
					push @{$rows_ref}, $row_ref;
				}
			}
			else {
				$results_ref->{error} = "Could not read header line in CSV $file: $!";
			}

File: lib/ProductOpener/Ecoscore.pm (hunk 1 above)
Comment (stephanegigandet) [reply lib/ProductOpener/Ecoscore.pm:103]: In this specific case, I think we are losing readability / understandability if we don't use getline. (it took me a while to understand why we read 8 lines instead of 3).

This file is going to be read only once at startup, so there is no performance issue.

Comment (Ban3) [reply lib/ProductOpener/Ecoscore.pm:103]: Fair enough.

File: lib/ProductOpener/Ecoscore.pm (hunk 3 above)
Comment (stephanegigandet) [reply lib/ProductOpener/Ecoscore.pm:117]: I tried it, but unfortunately it does not work because the column names are not unique. There are multiple "Total" columns... :-(

Comment (Ban3) [reply lib/ProductOpener/Ecoscore.pm:115]: Ah. You're right. Looks like it picked the last one total column when I tried it.

**Comment by Ban3** [comment]
> The job exceeded the maximum time limit for jobs, and has been terminated.

Travis has been weird with the taxonomies lately

**Comment by teolemon** [comment]
```
Errors in the ingredients taxonomy definition:
ERROR - de:natriumcitrat already is a synonym of de:e331 for entry en:e331 - de:natriumcitrat cannot be mapped to entry en:sodium-citrate
ERROR - et:naatriumtsitraat already is a synonym of et:e331 for entry en:e331 - et:naatriumtsitraat cannot be mapped to entry en:sodium-citrate
ERROR - fi:natriumsitraatti already is a synonym of fi:e331 for entry en:e331 - fi:natriumsitraatti cannot be mapped to entry en:sodium-citrate
ERROR - hu:nátrium-citrát already is a synonym of hu:e331 for entry en:e331 - hu:nátrium-citrát cannot be mapped to entry en:sodium-citrate
ERROR - sk:citran-sodný already is a synonym of sk:e331 for entry en:e331 - sk:citran-sodný cannot be mapped to entry en:sodium-citrate
ERROR - fr:tocopherol already is a synonym of fr:e307a for entry en:e307a - fr:vitamine-e cannot be mapped to entry en:vitamin-e
```

teolemon (approved):

stephanegigandet:
> Travis has been weird with the taxonomies lately

Yes. :(

But isn't the Travis test doing exactly the same thing as the Github tests? I'm a bit confused. Do we actually need travis?
 cc @hangy @svensven

svensven:
> But isn't the Travis test doing exactly the same thing as the Github tests? I'm a bit confused. Do we actually need travis?
> cc @hangy @svensven

Looking at the [history](https://travis-ci.org/github/openfoodfacts/openfoodfacts-server/builds)/[branches](https://travis-ci.org/github/openfoodfacts/openfoodfacts-server/branches) pages, it's been active for ages. It just didn't show up as an entry in the checks list on PRs until recently.

PR #4294, opened 2020-10-02
Perl modules from apt

svensven:
This is an attempt to speed up the pull request workflow (and building the backend docker image) by installing as many of the perl cpan dependencies from apt instead of through cpanm, because it's much faster.

The risk is that apt often has slightly older packages, especially on the Ubuntu 18.04 LTS image the pull request workflow uses, as opposed to the Debian Buster the docker image uses. (Some that were _really_ out of date, i didn't bother to add.)

However, if cpanfile specifies a version number, or the perl module itself specifies a version number in its dependencies, cpanm will download a newer version of the package from cpan, even if it's also installed through apt. I'm _assuming_ perl uses the newest installed version of a module if there's more than one.

AFAICT, this shaves 1-2 mins off the build time.

before: 3m18s
>2020-10-01T10:50:19.6980225Z local::lib is up to date. (2.000024)
>2020-10-01T10:53:37.8565832Z 159 distributions installed

after: 1m40s
>2020-10-02T14:28:11.3062539Z local::lib is up to date. (2.000024)
>2020-10-02T14:29:51.1941831Z 63 distributions installed

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -46,6 +46,7 @@ jobs:
         sudo apt-get install -y --no-install-recommends cpanminus tesseract-ocr
         # apt packages from cpanfile
         sudo apt-get install -y --no-install-recommends graphviz \
+        libtie-ixhash-perl \
         libwww-perl \
         libimage-magick-perl \
         libxml-encoding-perl \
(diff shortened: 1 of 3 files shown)


stephanegigandet (approved):
Looks good to me.

hangy:
> I'm assuming perl uses the newest installed version of a module if there's more than one.

I'm not sure it does version checking. It's more likely, that the first match in `@INC` is used.

PR #4295, opened 2020-10-02
Capture signal when refreshing taxonomies

Ban3:
**Description:**

Lately the taxonomy build script has been sporadically exiting with a null exitcode. This change logs the received signal when the exit code is null for more information about the failure.

I suspect it's getting killed with SIGKILL from the oomkiller, and it's probably something that should be taken into account when waiting for the spawned processes to exit.


svensven (approved):

stephanegigandet (approved):
Thank you!

Issue #4286, opened 2020-09-30
New image type for packaging info / recycling instructions


stephanegigandet:
We currently have 3 types of images that can be cropped and selected : front of the product, ingredients list, and nutrition facts table.

As more and more products have packaging information and recycling instructions printed on their label in more and more countries, it would be very useful to enable users to upload and select / crop images for recycling instructions.

In addition to "front", "ingredients" and "nutrition", we can add a new "packaging" image type, to hold either packaging info / logos, or if available recycling instructions. (if both are available, they are often close)

teolemon:
fixed

PR #4304, opened 2020-10-03
Add new packaging image type + packaging_text field - #4286

stephanegigandet:
This adds a new image type for packaging info / recycling instructions, in addition to the front/ingredients/nutrition images.
See #4286

There is also a new packaging_text field that can be populated by OCR of the recycling instructions, and edited by hand to specify more details.

In turn, we can later use the packaging_text field to populate a new packagings data structure: https://wiki.openfoodfacts.org/Packagings_data_structure

And after that, we can use the packagings data structure to take it into account in the upcoming EcoScore. :)


cgi/product_multilingual.pl
```diff
@@ -1479,7 +1481,7 @@ ($$$$$$)
 
 
 
-	$html .= display_tabs($product_ref, $select_add_language, "nutrition_image", $product_ref->{sorted_langs}, \%Langs, ["nutrition_image"]);
+	$html .= display_tabs($product_ref, $select_add_language, "nutrition_image", $product_ref->{sorted_langs}, \%Langs, ["nutrition_image"]);
```
svensven, line 1484:
```suggestion
	$html .= display_tabs($product_ref, $select_add_language, "nutrition_image", $product_ref->{sorted_langs}, \%Langs, ["nutrition_image"]);
```

aleene:
What is the difference between this new text field and the already existing one?

stephanegigandet:
> What is the difference between this new text field and the already existing one?

@aleene: The existing "packaging" field is a tag field: we split it using the commas, and each separate item goes into packaging_tags.

The "packaging_text" field is more like the "ingredients_text" field: we are going to extract structured data out of it (like the ingredients data structure), and also tags (like the ingredients_tags). But that will be done with a packaging analysis (like we do for ingredients), not just by splitting commas.

The goal is to eventually deprecate the current "packaging" field, once we have migrated its data to the new structure (and to a new preservation_methods field for the frozen/fresh/canned data that is also currently in the existing "packaging" field).

https://wiki.openfoodfacts.org/Packagings_data_structure details the overall plan.

teolemon (approved):

PR #4315, opened 2020-10-05
taxonomy edits 18 - ingredient additions/corrections

svensven:


taxonomies/ingredients.txt
```diff
@@ -6511,6 +6511,9 @@ vegetarian:en:yes
 carbon_footprint_fr_foodges_ingredient:fr:Lait semi-écrémé, pasteurisé
 carbon_footprint_fr_foodges_value:fr:1.2
 
+<en:milk
+en:unpasteurised milk
```
stephanegigandet, line 6515: maybe we can make unpasteurised a processing instead? (even though it's an absence of processing)

stephanegigandet (approved):
Looks good to me, thank you!

taxonomies/ingredients.txt
svensven: oh, it's already in there, just under -zed instead of -sed. :D

PR #4316, opened 2020-10-05
new convert_string_to_number function, fixes #4312

### LINKED ISSUE #4312: Strange number format with commas : 3,697,00 (sent by Spanish producer through the producers platform)
We get 3,697,00 for the kJ field, which corresponds to 3697 kJ, but that we convert to 3,697 kJ = 3.697 kJ.

stephanegigandet:
New convert_string_to_number function so that we avoid replacing comma by a dot in many places while forgetting that some numbers may be formatted differently.

This function is not perfect, we may very well find examples where it doesn't work (e.g. numbers in India). It is designed to work for numbers that we actually see on product labels, not any possible number in any possible language and numbering convention.

So it's more an improvement on what we have today, that we may need to improve again in the future (but at least we have a defined place to improve it).

### DIFF
--- a/lib/ProductOpener/Numbers.pm
+++ b/lib/ProductOpener/Numbers.pm
@@ -1,0 +1,148 @@
+# This file is part of Product Opener.
+#
+# Product Opener
+# Copyright (C) 2011-2020 Association Open Food Facts
+# Contact: contact@openfoodfacts.org
+# Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
+#
+# Product Opener is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+=head1 NAME
+
+ProductOpener::Numbers - normalize numbers sent as strings in various formats
+with different sets of separators for digit grouping and to indicate decimals.
+
+=head1 DESCRIPTION
+
+=cut
+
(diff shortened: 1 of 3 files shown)


stephanegigandet:
Also fixes #4312

t/numbers.t
```diff
@@ -38,5 +38,24 @@ foreach my $test_ref (@tests) {
...
 
+@tests = (
+["1","1"],
+["1.0","1.0"],
+["2,2","2.2"],
+["2,20","2.20"],
+["100 000","100000"],
+["33,000,000","33000000"],
+["3,760,00","3760"],
+);
```
svensven, line 49:
```suggestion
["1,000.22","1000.22"],
);
```
both a thousand separator and a decimal at the same time, for completeness?

svensven (approved):

lib/ProductOpener/Numbers.pm
```diff
@@ -0,0 +1,148 @@
...
+by guessing which characters (spaces, commas, dots) are used as
+digit grouping separators or used to indicate decimals.
+
+=cut
+
+sub convert_string_to_number($) {
+	
+	my $value = shift;
+	
+	$value =~ s/(\d) (\d)/$1$2/g;
```
hangy, line 127: Do you know how often this gets called? This, and other, regex replaces could be replaced by [`tr`](https://stackoverflow.com/a/3847062/11963).

stephanegigandet: Thanks @hangy , I merged the PR too soon. It's called whenever a nutrient value is updated. So it could be called a dozen of times for one product. In an import of 1000 products, it could be called in the order of 10 000 times. It's only when writing data, so I'm guessing it may not be worth spending too much time to improve its performance. We can replace the regexes with tr (especially if it helps readability).

t/numbers.t
stephanegigandet: Thanks @svensven , I'll add it.

lib/ProductOpener/Numbers.pm
svensven: ooh, interesting. for some reason i saw the slashes, and assumed tr involved regex overhead too. i only ever use s/// with sed.

t/numbers.t
stephanegigandet: added in https://github.com/openfoodfacts/openfoodfacts-server/pull/4317

Issue #4105, opened 2020-08-31
Implement Product Attributes


stephanegigandet:
This is a tracking issue for the new Product Attributes feature: https://wiki.openfoodfacts.org/Product_Attributes

stephanegigandet:
Deployed.

PR #4327, opened 2020-10-06
added attributes for allergens, + taxonomy changes #4105

stephanegigandet:
This adds an attribute for each allergen, so that we can use it for personalized search.

I also made some changes to the allergens / labels taxonomy to make it more predictive (in particular, always making "no something" the canonical entry instead of sometimes "something-free" or "without something).


po/common/en.po
```diff
@@ -4834,4 +4834,12 @@ msgstr "Recycling instructions and/or packaging information text has been extrac
 
 msgctxt "product_js_extracting_packaging"
 msgid "Extracting recycling instructions and/or packaging information"
-msgstr "Extracting recycling instructions and/or packaging information"
\ No newline at end of file
+msgstr "Extracting recycling instructions and/or packaging information"
+
+msgctxt "contains"
+msgid "Contains"
```
teolemon, line 4840: This is a French centric approach. We should have "Contains %" so that translators can potentially do "% Bla Bla"

stephanegigandet: It's to be used as "Contains: XYZ". The issue with "Contains %" is that it will create bad sentences in some languages. e.g. in French, we won't have the right article. "Contient gluten" and "Ne contient pas gluten" sounds bad.

teolemon: "Contains: %" then.

teolemon: Arabic is RTL, so it should be one sentence, otherwise we will have issues.

stephanegigandet: ok

stephanegigandet: @teolemon : fixed, thank you!

svensven:
poking travis...

taxonomies/labels.txt
```diff
@@ -2735,7 +2738,7 @@ fr:Contient du soja
 hu:Szójatartalmú, szóját tartalmaz
 it:Contiene soia
 pt:Contem soja
-
+opposite:en: en:contains-soy
```
svensven, line 2741:
```suggestion
opposite:en: en:no-soy
```

svensven (approved):

PR #4328, opened 2020-10-07
perl syntax checking in parallel, but limit the number of processes.

svensven:
Adapted from the refresh_taxonomies.js script.

The syntax checking is slow (5-6 mins on github), but checking over 100 scripts simultaneously is also unworkable:
https://github.com/openfoodfacts/openfoodfacts-server/pull/3368

The limit is configurable, in case the environment running the checks ever has more than the 2 cores it currently does. Increasing the limit on 2 cores works (at least with 3 & 4), but is no faster.

Should shave another 2-3 mins off the pull request workflow.

Update: With the changed files workflow plugin, it removes the whole 5-6 mins.

### DIFF
--- a/scripts/check_perl.js
+++ b/scripts/check_perl.js
@@ -42,7 +42,7 @@ async function main() {
 
     // count how many spawned processes don't have an exit code yet
     running = spawns.reduce((pv, cv) => {
-      if (cv.exitCode === null) {
+      if (cv.exitCode === null && cv.signalCode === null) {
         return pv + 1;
       }
       else {
@@ -99,7 +99,7 @@ async function main() {
   running = Number.MAX_SAFE_INTEGER;
   while (running > 0) {
     running = spawns.reduce((pv, cv) => {
-      if (cv.exitCode === null) {
+      if (cv.exitCode === null && cv.signalCode === null) {
         return pv + 1;
       }
       else {
--- a/scripts/check_perl.js
+++ b/scripts/check_perl.js
@@ -2,37 +2,113 @@
 /*eslint no-await-in-loop: "off"*/
 
 const process = require("process");
-const util = require("util");
+const util = require('util');
 const glob = util.promisify(require("glob").glob);
-const { spawn } = require("child_process");
+const { spawn } = require('child_process');
 
-function checkFile(path, doneCallback) {
-  console.log(`Checking ${path}`);
-  try {
-    const child = spawn(`perl`, ["-c", "-CS", "-Ilib", path]);
-    child.stdout.on("data", (data) => console.log("[" + path + "] " + data));
-    child.stderr.on("data", (data) => console.error("[" + path + "] " + data));
-    child.on("close", (code) => {
-      console.log("[" + path + "] result: " + code);
-      process.exitCode = code;
-      doneCallback();
-    });
-  } catch (e) {
-    console.error("[" + path + "] " + e);
-    process.exitCode = e.code;
-    throw e;
-  }
+// Github actions have 2 CPUs: 
+// https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources
+// Travis also seems to have 2 CPUs:
+// https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system
+
+const maxRunning = 2;
+
+function sleep(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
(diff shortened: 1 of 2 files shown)


scripts/check_perl.js
```diff
@@ -2,37 +2,113 @@
...
+
+  }
+
+  console.log('Finished queueing files.');
+
+  // don't return until all child processes have finished
+  running = Number.MAX_SAFE_INTEGER;
+  while (running > 0) {
+    running = spawns.reduce((pv, cv) => {
+      if (cv.exitCode === null) {
```
Ban3, line 102: You might want to check `cv.signalCode` here as well. If the process gets killed for out of memory reason, the exitCode will stay as null and CI wont fail until timeout is reached.

This is a issue with refresh-taxonomies.js too, as we saw recently with Travis failures.

From the exit event documentation. This refers to the function arguments, but it also applies to the properties.:

> The 'exit' event is emitted after the child process ends. If the process exited, code is the final exit code of the process, otherwise null. If the process terminated due to receipt of a signal, signal is the string name of the signal, otherwise null. One of the two will always be non-null.

svensven: so, `if (cv.exitCode === null && cv.signalCode === null) {` ?

Ban3: @svensven Yeah, I think that's good enough for our purposes. I think technically signalCode can be some non-null signal that didn't lead to exit (like `SIGUSR2` or whatever) but seems pretty unlikely to happen.

svensven:
well this is a new one.


> Wed, 07 Oct 2020 19:37:26 GMT 61 distributions installed
> Wed, 07 Oct 2020 19:37:27 GMT Can't locate DateTime/TimeZone/Catalog.pm in @INC (you may need to install the DateTime::TimeZone::Catalog module) (@INC contains: lib /home/runner/perl5/lib/perl5/5.26.1/x86_64-linux-gnu-thread-multi /home/runner/perl5/lib/perl5/5.26.1 /home/runner/perl5/lib/perl5/x86_64-linux-gnu-thread-multi /home/runner/perl5/lib/perl5 /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.26.1 /usr/local/share/perl/5.26.1 /usr/lib/x86_64-linux-gnu/perl5/5.26 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.26 /usr/share/perl/5.26 /home/runner/perl5/lib/perl5/5.26.0 /home/runner/perl5/lib/perl5/5.26.0/x86_64-linux-gnu-thread-multi /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base) at /home/runner/perl5/lib/perl5/DateTime/TimeZone.pm line 14.
> Wed, 07 Oct 2020 19:37:27 GMT BEGIN failed--compilation aborted at /home/runner/perl5/lib/perl5/DateTime/TimeZone.pm line 14.
> Wed, 07 Oct 2020 19:37:27 GMT Compilation failed in require at /home/runner/perl5/lib/perl5/x86_64-linux-gnu-thread-multi/DateTime.pm line 17.
> Wed, 07 Oct 2020 19:37:27 GMT BEGIN failed--compilation aborted at /home/runner/perl5/lib/perl5/x86_64-linux-gnu-thread-multi/DateTime.pm line 17.
>

svensven:
ah, it's a bug in the cpan module that just got updated. https://github.com/houseabsolute/DateTime-TimeZone/issues/43

svensven:
> ah, it's a bug in the cpan module that just got updated. [houseabsolute/DateTime-TimeZone#43](https://github.com/houseabsolute/DateTime-TimeZone/issues/43)

and fixed.

stephanegigandet:
If it's just syntax checking, maybe we could try a way to run the tests only when the scripts have changed, like we do for the taxonomies. It doesn't make much sense to test 100 scripts every time.

I don't know if we can get the info of whether a file has been changed in a PR when we run the tests though.
One option could be to periodically touch and commit a file, but it's not very elegant.

svensven:
> I don't know if we can get the info of whether a file has been changed in a PR when we run the tests though.

That would be nice. Given that someone's written this https://github.com/jitterbit/get-changed-files but, looking at its issues, github doesn't expose enough information to handle force-pushes etc, it may be difficult to achieve.

svensven:
this one seems to use a different technique, which maybe doesn't have the same issue: 
https://github.com/marketplace/actions/changed-files-exporter

stephanegigandet (approved):

PR #4216, opened 2020-09-16
Flakes

moon-rabbitOFF:

PR #4329, opened 2020-10-07
refresh_taxonomies.js - handle signals & limit parallelism (for Travis)

svensven:
Applying the same quick fix as in #4328. See https://github.com/openfoodfacts/openfoodfacts-server/pull/4328#discussion_r501245467

After the previous change, this Travis run now logs that one of the processes had been SIGKILLed, but the script still counted it as running. https://travis-ci.org/github/openfoodfacts/openfoodfacts-server/builds/733734997
> [additives] child process exited with signal SIGKILL

### DIFF
--- a/scripts/refresh_taxonomies.js
+++ b/scripts/refresh_taxonomies.js
@@ -69,6 +69,7 @@ async function main() {
           }
           else if ( signal !== null ) {
             console.log(`[${file}] child process exited with signal ${signal}`);
+            process.exitCode = 128; // consider a spawn getting killed a failure
           }
         });
       } catch (e) {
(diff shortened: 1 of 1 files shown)


Ban3:
Now that I think about it, we should also change `process.exitCode` so that SIGKILL won't count as a success. Maybe something like this would work (128 being just a random non-zero value):

```diff
          else if ( signal !== null ) {
            console.log(`[${file}] child process exited with signal ${signal}`);
+           process.exitCode = 128;
          }
```

svensven:
how do we rerun a failed travis build? "rerun all jobs" doesn't seem to do it: https://github.com/openfoodfacts/openfoodfacts-server/pull/4216

Ban3:
> how do we rerun a failed travis build? "rerun all jobs" doesn't seem to do it: #4216

Closing and reopening PR is the classic trick.

Ban3 (approved):

PR #4139, opened 2020-09-04
Corrections ciqual

moon-rabbitOFF:

PR #4332, opened 2020-10-08
"do we need the Travis check?"

svensven:
ERROR - de:geflügelschinken already is a synonym of de:hähnchen-brustfilet-in-scheiben for entry en:cooked-chicken-breast-slices - de:geflügelschinken cannot be mapped to entry en:poultry-hams

Broken in #4139. 
Clearly the duplicate detection isn't deterministic. I'm guessing that if one term is a primary translation (which aren't checked for duplicates), and the other a synonym of a translation, it won't flag an error if it happens to process the synonym before the primary translation.

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -57895,7 +57895,7 @@ nl:Kippendijen, Kipdij
 
 <en:Poultry hams
 en:Cooked chicken breast slices, chicken ham
-de:Hähnchen-brustfilet in Scheiben, Geflügelschinken
+de:Hähnchen-brustfilet in Scheiben
 fi:kanaleikkele
 fr:Blancs de poulet en tranches, Blancs de poulet, Blanc de poulet, Jambons de poulet, Jambon de poulet
 nl:Kippendijen, Kippendij


svensven:
so fed up with travis.

Ban3:
> so fed up with travis.

Should we just limit the concurrency on Travis? Maybe the process limit code from syntax checking could be adapted here.

svensven:
> > so fed up with travis.
> 
> Should we just limit the concurrency on Travis? Maybe the process limit code from syntax checking could be adapted here.

Yeah, could do. I'll add it to my other PR in a mo.

svensven:
> 
> 
> > > so fed up with travis.
> > 
> > 
> > Should we just limit the concurrency on Travis? Maybe the process limit code from syntax checking could be adapted here.
> 
> Yeah, could do. I'll add it to my other PR in a mo.

Done.

stephanegigandet:
This may be a stupid question, but isn't Travis testing exactly the same things that are tested in the "Pull Request / test (pull request)" tests? https://github.com/openfoodfacts/openfoodfacts-server/pull/4332/checks?check_run_id=1225733052

If so, is there any reason to keep the Travis test? cc @hangy

hangy:
> This may be a stupid question, but isn't Travis testing exactly the same things that are tested in the "Pull Request / test (pull request)" tests? https://github.com/openfoodfacts/openfoodfacts-server/pull/4332/checks?check_run_id=1225733052
> 
> If so, is there any reason to keep the Travis test? cc @hangy

Yes, it's probably redundant with GitHub actions being used.

PR #4334, opened 2020-10-08
more performance tweaks for update_all_products.pl

svensven:
### unac_string_perl()
After [learning about tr/// the other day](https://github.com/openfoodfacts/openfoodfacts-server/pull/4316#discussion_r499659223), this seemed like a good candidate, _assuming I'm understanding how it works correctly_.
For a test of 1500ish records, it went from 13.4s to approx 3s for 644k calls.
> image

vs
> image

### preparse_ingredients_text()
This is called twice with the same inputs for each product. 
> image

Adding a simple cache using state variables means we can return immediately on the 2nd invocation, as shown by the calls from extract...() being 1000x faster:
> image

(This also means we're calling unac_string_perl() half as much.)

to
> image

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -3335,6 +3335,20 @@ sub preparse_ingredients_text($$) {
 
 	$log->debug("preparse_ingredients_text", { text => $text }) if $log->is_debug();
 
+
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Perfect, thank you!

PR #4335, opened 2020-10-09
fixed categories taxonomy, built taxonomies

stephanegigandet:
- fix from https://github.com/openfoodfacts/openfoodfacts-server/pull/4332
- built all taxonomies in the hope that all tests will pass and won't fail due to a timeout


teolemon (approved):

stephanegigandet:
removing branded waters from categories (introduced from CIQUAL/Agribalyse, but not relevant for categories) and rebuilding categories

PR #4336, opened 2020-10-09
removed mineral water brands

stephanegigandet:
CIQUAL has categories for individual brands (e.g. Perrier), they don't really make sense as categories for us, and their Agribalyse impact is the same anyway as they were not evaluated individually.


taxonomies/categories.txt
```diff
@@ -6799,777 +6799,6 @@ nl:Natuurlijke mineraalwaters, Natuurlijk mineraalwater
...
-nl:Met koolzuur versterkte mineraalwaters, Met koolzuur versterkt mineraalwater
-
-<en:Carbonated natural mineral waters
-en:Mildly carbonated natural mineral water
-de:natürliches Mineralwasser mit wenig Kohlensäure versetzt, natürliche Mineralwässer mit wenig Kohlensäure versetzt
-fi:kevyesti hiilihapotettu luonnollinen kivennäisvesi
-fr:Eau minérale naturelle légèrement gazéifiée
-it:Acqua minerale naturale mediamente carbonata
-nl:Met weinig koolzuur versterkte mineraalwaters, Met weinig koolzuur versterkt mineraalwater
-
```
svensven, line 7572: are the last 2 deletions intentional? they're not brands.

stephanegigandet: good  catch, I'll put them back

teolemon:
I kind of told @TaciteOFF and @moon-rabbitOFF that we'd think about a wide-ranging solution, but it's a good start 👍

teolemon (approved):

aleene:
You might need to add the brand minerals as synonyms, or you will lose the ingredients recognition.

stephanegigandet:
> You might need to add the brand minerals as synonyms, or you will lose the ingredients recognition.

Hi @aleene Sorry I don't understand, can you elaborate?

PR #4338, opened 2020-10-09
default userid for logged in users

stephanegigandet:
This is to allow to create links to change the logged in user profile without putting the userid explicitly in the link.

In particular, we can create links for moderators of the producers platform of the form
https://fr.pro.openfoodfacts.dev/cgi/user.pl?action=process&type=edit_owner&pro_moderator_owner=org-carrefour-espana
to change the org to org-carrefour-espana


teolemon (approved):

teolemon (approved):
Thank you very much.

PR #4339, opened 2020-10-09
product editing - add back current-image html for non-ocr images

### LINKED ISSUE #4324: Current "front" image no longer visible on product edit page.
**Describe the bug**
The record has a front image saved,
> image

but on editing it's not showing up as the "current image" the way the ingredients/etc pictures do:
> image

(unless I'm completely imagining that it did that in the past?)

vs

> image

**To Reproduce**
https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=0250980001102

svensven:
fixes #4324, hopefully.

### DIFF
--- a/html/js/product-multilingual.js
+++ b/html/js/product-multilingual.js
@@ -455,6 +455,10 @@ function update_display(imagefield, first_display) {
 				$(document).foundation('equalizer', 'reflow');
 			});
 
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
I broke that last week, thanks for fixing it!

PR #4340, opened 2020-10-09
Corrections

moon-rabbitOFF:


taxonomies/categories.txt
```diff
@@ -24322,7 +24297,7 @@ fr:Pain ukrainien nouveau
 nl:Nieuwe Oekraïense broden
 
 <en:Breads
-en:Sliced breads, Sandwich loaf
+en:Soft breads, Sliced breads, Sandwich loaf
 ca:Pa de motlle
 de:geschnittenes Brot
 es:Panes de molde
```
svensven, line 24303: i'm not sure "soft bread" is a common term in english, whereas the other two are immediately recognisable.
to me, "soft bread" makes me think of brioche.

moon-rabbitOFF: No it doesn't seem to be common, but I think (maybe I'm wrong), that it is more accurate than "Sliced bread". To me, sliced bread refers to any bread which is in slices, while this category describe a particular type of bread, this one : https://fr.wikipedia.org/wiki/Pain_de_mie
Apparently there is no strict synonym of "pain de mie" in english, but I saw that those breads seems to be refered as "Soft bread" in the UK : https://world.openfoodfacts.org/product/20675615/soft-medium-white-rowan-hill

moon-rabbitOFF: In the US https://world.openfoodfacts.org/product/0025911000512/vermont-bread-company-organic-soft-white-bread and in Australia https://world.openfoodfacts.org/product/9310645254128/soft-white-toast-loaf-coles too I think
Maybe this term is no common, because this type of bread is not common in English-speaking countries

moon-rabbitOFF: According to Wikipedia, it is called pan loaf in English https://en.wikipedia.org/wiki/Pan_loaf
but we don't have only one corresponding product in the database https://world.openfoodfacts.org/cgi/search.pl?search_terms=pan+loaf&search_simple=1&action=process
so I think it is not really used

svensven: but those examples are all _white_ bread, and the "soft" is because of that.
we have sliced brown/wholemeal/etc bread in the same shape too. E.g.
https://world.openfoodfacts.org/product/00642118/super-seeded-marks-spencer
https://world.openfoodfacts.org/product/3250392265021/pain-complet-aux-flocons-d-avoine-chabrior

svensven: having just been to the supermarket, yes, the white bread was often labelled as soft. but "sliced bread" should then be a separate term, broader than that.
Also found https://en.wikipedia.org/wiki/Sandwich_bread, and _of course_ the english term "pain de mie" means something slightly different from the french: https://en.wikipedia.org/wiki/Pain_de_mie :smile:

svensven:
BTW, if you merge from, or rebase onto, upstream/master, the tests should run much faster and the travis test not keep failing.
I think you created the branch just before we fixed that, so it's still using the old code.

taxonomies/categories.txt
moon-rabbitOFF: Hmm indeed, so what term would you keep?

svensven: TBH i think i'd keep it as it was, "sliced breads". or at most rename it to "sliced sandwich breads".
looking at the ciqual site they translate "pain de mie" as "sandwich loaf", which is correct, but probably not a term a user would type, especially as it doesn't contain the word "bread".
and in english, "sliced bread" is such a strong term there's the idiom "it's the best thing since sliced bread".

but speaking of white bread, the one thing https://world.openfoodfacts.org/category/sliced-breads doesn't have is "sliced white breads", which could be a child of both "sliced breads" and "white breads".

similarly, "Gluten-free sliced breads" could have the extra parent "Gluten-free breads", and maybe "Wholemeal English bread" -> "brown breads", and "Bran grain sandwich loaf" -> "Bran bread"?

moon-rabbitOFF: Ok I'll change that

moon-rabbitOFF: BTW shouldn't we rename "Brown breads" into "Wholemeal breads"? I think there products that contain almost no wholemeal flour that are called brown breads

svensven: AIUI not all "brown" bread is wholemeal. sometimes it's just dyed brown, or from a mixture of white & whole flour. https://en.wikipedia.org/wiki/Brown_bread.
So maybe "brown and wholemeal breads"? or just "wholemeal breads" could be it's own section under "breads".

moon-rabbitOFF: Yes sometimes they are just dyed brown and I think a categorisation of bread by color is not meaningful. The world brown is confusing for the consumer, which is convenient for some brands but not for us. So I think it is better not to have a category for brown breads but solely for wholemeal ones. Furthermore, I think this category was initially created to  contain wholemeal breads as the translations (except the dutch and the romanian ones)  all means Wholemeal bread

svensven: Yeah, fair enough. Looking through [tesco](https://www.tesco.com/groceries/en-GB/shop/bakery/bread-and-rolls/all?aisle=Brown%20%26%20Wholemeal%20Bread&viewAll=aisle) and [sainsbury's](https://www.sainsburys.co.uk/shop/gb/groceries/bakery/wholemeal-brown-bread) it seems like the only breads labelled brown are a few 50/50 flour cases, and everything else is labelled wholemeal.

moon-rabbitOFF: I did all the changes!

stephanegigandet (approved):
Looks good to me, thank you!

PR #4342, opened 2020-10-09
Don't count '... photo selected' as a TODO state

Ban3:
**Description:**

Currently the new '... photo selected' states are shown in the TODO section because they contain the substring 'to '.

- Change 'not' to 'to be' in the new states to match previous incomplete states
- Make the TODO/Done check more specific to 'to be' instead just 'to '


stephanegigandet (approved):
Looks good, thank you!

If at some point it gets to convoluted to make decisions based on the state name, we could alternatively add a property to the states in the taxonomy, to indicate if they are "to do" or "to be done" states.

PR #4347, opened 2020-10-11
Read API: new images_to_update_[lc] field, fixes #4292

### LINKED ISSUE #4292: Create an API to indicate which images would need to be taken (again)
Create an API to indicate which images would need to be taken (again), with language and field, based on some conditions

```
ingredients_fr_to_refresh
ingredients_pt_to_take
recycling_instructions_to_take
labels_to_take
packager_code_to_take
```

if front_fr=selected and ingredients_fr=empty
if category=meat and packager_codes=to_be_completed
if recycling_instructions=to_be_completed (or a more sophisticated condition required for Eco-Score)

stephanegigandet:
See #4292. This is so that the OFF app (or other apps) can request a field like images_to_update_en to know which images should be taken or re-taken (because they are too old, or possibly too small / blurry) by users.

https://fr.openfoodfacts.dev/api/v0/produit/3483130043180/beurre-cru-a-la-baratte-les-petites-laiteries?fields=images_to_update_fr

It returns a hash of image types + language code (only for the requested language code -> the language of the app user). The value is 0 for images we don't have, or the age of the image (in seconds) for apps that want to add some context like "Our photo of the ingredients is 14 months old, could you take a new one?".

```````
product: {
images_to_update_fr: {
packaging_fr: 0,
front_fr: 83734290,
ingredients_fr: 83734290
}
},

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -4364,7 +4364,35 @@ sub customize_response_for_product($) {
 		elsif ($field eq "attribute_groups") {
 			compute_attributes($product_ref, $lc);
 			$customized_product_ref->{$field} = $product_ref->{"attribute_groups_" . $lc};
-		}		
+		}
+		
+		# Images to update in a specific language
+		elsif ($field =~ /^images_to_update_([a-z]{2})$/) {
+			my $target_lc = $1;
+			$customized_product_ref->{$field} = {};
+
+			foreach my $imagetype ("front", "ingredients", "nutrition", "packaging") {
+				
+				my $imagetype_lc = $imagetype . "_" . $target_lc;
+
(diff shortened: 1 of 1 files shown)


svensven (approved):

teolemon:
Related Android issue: https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/3543

teolemon:
iOS issue: https://github.com/openfoodfacts/openfoodfacts-ios/issues/777

aleene:
I am not a fan of these field names. I do not like field names that depend on the data that is in the json. I have now to create all possible variants of language codes and images in order to check whether the fields is present.

kartikaysharma01:
I agree with @aleene .
@teolemon @stephanegigandet Can we have this change done, otherwise it will be a messed up on client side.

PR #4353, opened 2020-10-12
Add icons for allergens, and add them to the product attributes

stephanegigandet:
SVG icons for allergens:

image


svensven (approved):
cool.

PR #4356, opened 2020-10-14
Handle edit page ajax failures

svensven:
- set data-language when adding a language, to allow power user script ingredient analysis to work on new languages without saving and reloading.
- handle failures in ajax post requests. currently, when the server's timing out, the page will hide buttons and just sit in an "in progress" state, needing a refresh to work again.

### DIFF
--- a/html/js/product-multilingual.js
+++ b/html/js/product-multilingual.js
@@ -18,6 +18,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+/*eslint dot-location: ["error", "property"]*/
 /*eslint no-console: "off"*/
 /*global lang admin otherNutriments Tagify*/
 /*global toggle_manage_images_buttons */ // These are weird.
@@ -340,7 +341,7 @@ function change_image(imagefield, imgid) {
     $('.cropbutton_' + imagefield).hide();
     $('.cropbuttonmsg_' + imagefield).html('<img src="/images/misc/loading2.gif" /> ' + lang().product_js_image_saving);
     $('.cropbuttonmsg_' + imagefield).show();
-    var crop_post = $.post(
+    $.post(
       '/cgi/product_image_crop.pl',
       {
         code: code, id: imagefield, imgid: imagefield_imgid[imagefield],
@@ -449,7 +450,7 @@ function update_display(imagefield, first_display) {
 				// alert(event.data.imagefield);
 				$('div[id="ocrbutton_loading_' + imagefield +'"]').html('<img src="/images/misc/loading2.gif" /> ' + lang()["product_js_extracting_" + imagetype]).show();
 				$('div[id="ocrbuttondiv_' + imagefield +'"]').hide();
-				var ocr_post = $.post(
+				$.post(
 					'/cgi/' + imagetype + '.pl',
 					{code: code, id: imagefield, process_image:1, ocr_engine:"google_cloud_vision" },
 					null,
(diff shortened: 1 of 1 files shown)


html/js/product-multilingual.js
```diff
@@ -437,22 +450,32 @@ function update_display(imagefield, first_display) {
...
+					})
+					.fail(function() {
+						$('div[id="ocrbuttondiv_' + imagefield +'"]').show();
+						$('div[id="ocrbutton_loading_' + imagefield +'"]').html(lang().job_status_failed);
+					})
+					.always(function() {
+						$(document).foundation('equalizer', 'reflow');
+					});
+
+				//$(document).foundation('equalizer', 'reflow');
```
hangy, line 478: If this is actually not required any more, just remove the code. (Same in the function below.)

html/js/product-multilingual.js
```diff
@@ -18,6 +18,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+/*eslint dot-location: "off"*/
```
hangy, line 21: IMHO disabling a purely stylisting linter rule for a single file defeats the purpose of using style rules in general. We can definitely discuss the [`dot-location`](https://eslint.org/docs/rules/dot-location) rule for all files (and change it in `.eslintrc.js`), as it simply extends from the `eslint:recommended` rule set. I wouldn't want to disable it for one file, though, because then the style wouldn't be consitent among all JS files any more.

html/js/product-multilingual.js
svensven: I didn't want to remove it without being sure it wasn't specified twice for a reason.

html/js/product-multilingual.js
svensven: i mean, it's not like anything else in product-multilingual.js is consistent, even within the file...

svensven: but with dot-location in particular, a convention of having the dot on the end of the previous line is one i've not seen before, and doesn't make sense to me.

html/js/product-multilingual.js
stephanegigandet: It doesn't happen at the same time. In practice the reflow is needed when we change the content in a way that may change the overall height of the content. Otherwise the bottom of the content might be hidden by the footer. In this specific case, we are just showing an icon and hiding a button, so I guess we could remove it (the commented out one).

stephanegigandet (approved):
Thank you!

html/js/product-multilingual.js
svensven: also before, the inner one would only have run on success, whereas now it should always run.

html/js/product-multilingual.js
hangy: As I said, it's in `eslint:recommended`. Changing it for just one file is super unexpected and shouldn't be done IMHO.

svensven: well as i mentioned in my commit, it's going directly against the examples in the jquery docs https://api.jquery.com/jQuery.post/#jqxhr-object, so the rule may be going too far. 

imho disabling one rule in one file is the least worst option (well, i suppose there's the disabling it for specific lines). lint is just a tool that's there to be helpful, not a religion, and even it recognises that you may not want a rule. https://eslint.org/docs/rules/dot-location#when-not-to-use-it
i'd argue that changing the setting globally just to get one file to pass is more disruptive. and i did fix the other errors lint threw up.

Issue #4370, opened 2020-10-15
Add profiles for organizations, e.g. to add customer service information for producers


stephanegigandet:
Food manufacturers can create professional accounts on the platform for producers to send us data and photos directly.
Those pro user accounts are associated with an organization account (so that multiple people of a company can access the producers platform).

This feature is to allow members of organizations to edit an organization profile (similar to the user profiles we have) so that they can provide useful information, such as customer service or commercial service contact information. This hopefully will reduce the number of emails we get from people who complain about products. :) 

Edit profile form:

![image](https://user-images.githubusercontent.com/8158668/96162693-212b0a80-0f19-11eb-880b-f413ac035b0a.png)

Profile page:

![image](https://user-images.githubusercontent.com/8158668/96162898-694a2d00-0f19-11eb-9212-00e5bc2d3df1.png)

We could also surface some of the info directly on the product page on the web site or app (e.g. the phone number of the customer service)

stephanegigandet:
Deployed.

Issue #4351, opened 2020-10-12
Allow users to add a location (city and country) to their profile, and create a map of contributors


stephanegigandet:
It would be nice to let users put more information on their profile pages.

Maybe a short "about" text, possibly links to social media accounts.

And a location of the form city + country. That would allow us to create a map of contributors. :)

e.g. this is how OpenStreetMap shows contributors near you:

![image](https://user-images.githubusercontent.com/8158668/95729537-53b8d700-0c7d-11eb-93da-3a8b7b81145e.png)

In practice, we could add an optional ?map=1 to the user facet.

e.g. from this url:
https://world.openfoodfacts.org/brand/coca-cola/editors

we could show a map at this url: https://world.openfoodfacts.org/brand/coca-cola/editors?map=1

PR #4371, opened 2020-10-15
Add org profiles #4370, start to refactor user profiles

stephanegigandet:
This adds profiles for organizations (see #4370 for details and screenshot).

I also started some refactoring to improve user profiles, to make it easier to redesign them and add fields (such as the map: 
#4351 )


teolemon:
image

stephanegigandet:
I replaced the table and removed all the responsive columns in the org profile edit form.
That way the behavior is the same at all sizes.
I think we should do the same for the user edit form too.

image

teolemon (approved):

Issue #4296, opened 2020-10-03
JS cropper is unmaintained, deprecated in favour of jquery-cropper


svensven:
https://fengyuanchen.github.io/cropper/ says 
>As of v4.0.0, the core code of Cropper is replaced with Cropper.js. I recommend you to use the jquery-cropper instead of Cropper. 

https://github.com/fengyuanchen/jquery-cropper

https://github.com/fengyuanchen/cropper has been archived, so won't get any more updates.

hangy:
I was sind If we could just ignore the jQuery wrapper and use the native JavaScript version Cropper.js instead. 😀

svensven:
yeah, https://github.com/fengyuanchen/cropperjs

i'm afraid i haven't looked into the JS side much, i don't know how the npm/etc stuff works.

PR #4376, opened 2020-10-18
migrate from defunct cropper to cropperjs+jquery-cropper

svensven:
Using jquery-cropper means it's a drop-in replacement, and gets us off a dead-end without having to change any of our JS for now.
See:
https://github.com/openfoodfacts/openfoodfacts-server/issues/4296
https://github.com/fengyuanchen/cropper
https://github.com/fengyuanchen/cropperjs
https://github.com/fengyuanchen/jquery-cropper

(tested on my docker instance)


hangy (approved):
Good idea! 👍

PR #4384, opened 2020-10-19
Icons, accordion and styling for user food preferences

stephanegigandet:
The icons help to make the radio buttons less frightening. The next step will be to use input range with steps to have a slider instead of the buttons.

I was trying to add an accordion but couldn't get it to work with our old version of Foundation (it's supposed to work though). It has to do with the fact that the accordion is dynamically created. But in fact I kind of like to have all options shown without having to click multiple times on the accordion so I guess I'll leave it like that. :)

To test:

https://world.openfoodfacts.dev/search?user_preferences=1&categories_tags_en=sausages&labels_tags=en:organic&fields=product_name,ecoscore_data,images#attribute_group_processing_a

(login and pass : off )

image

### DIFF
--- a/html/images/icons/0-additives.svg
+++ b/html/images/icons/0-additives.svg
@@ -1,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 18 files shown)


teolemon (approved):

teolemon:
Approved, but will need l10n and i18n

stephanegigandet:
> Approved, but will need l10n and i18n for the JavaScript part

Yes, definitely.

PR #4385, opened 2020-10-20
Default language-less xx: values for tags, fixes #3872, enables brands taxonomy

### LINKED ISSUE #3872: Default language entries + language-less entries for taxonomies, using the xx: prefix
The current taxonomy work very well for entries that are in one language and can be translated in other languages (e.g. categories, ingredients), but it is much less adapted for entries that do not have translations in other languages (e.g. labels like "AB - Agriculture Biologique" or "Red Tractor" should not really be translated and displayed in another language), or that are more or less language-less (like many brands, and labels that are in fact brands like Demeter or Max Havelaar).

This is causing issues for labels where we have to add the same entries like "Demeter" in many different languages, so that we can match it for products entered in different languages, and also so that we can display it without a fr: or de: etc. prefix.

To create a brands taxonomy, we also need language-less entries that match in all languages, and are displayed in all languages. (there are a few rare exceptions where brands are translated)

The proposed solution is a xx: language prefix that can be used as a default language:

fr:AB Agriculture Biologique -> canonical name is fr:ab-agriculture-biologique
xx:AB Agriculture Biologique -> makes AB Agriculture Biologique match in all languages, and also makes AB Agriculture Biologique a default translation for all languages (unless they have their own entry)

And that can also be used for language less entries:

xx:Nestlé -> xx:nestle is the canonical name, it matches in all languages, and is the default translation for all languages

the xx: code is currently unassigned, and using a 2 letter code will require much less changes (and create much less risks of breaking something) than using something different.

stephanegigandet:
This is an implementation of the proposal made in July: #3872 

It creates a new xx: prefix for taxonomy that can be used as a default wildcard language: it matches all languages.

This will remove the need to define entries in all languages for things like the "Demeter" label.

For labels that are specific to a country and are defined in one language (e.g. the French "AB Agriculture Biologique"), we should use fr:AB Agriculture Biologique as the main language, and add a xx:AB Agriculture Biologique entry below  so that it can be recognized and displayed in all languages.

xx: entries can also be used as the main language. This will be very useful to make the taxonomy system work for brands: by default all brands will be defined with the xx: wildcard language. And in the rare cases where brands are spelled differently or translated ("The laughing cow" / "La vache qui rit" / "La vaca que rie"), we can add local translations.


teolemon:
How do you handle the unknown languages cases we already have:
https://world.openfoodfacts.org/language/unknown-language.json?fields=lc

svensven:
i like the idea. personally, i'm curious how presenting the xx tags on the product view, and especially editing view, will work. i'm guessing suggest.pl will need modifying too?

svensven:
> How do you handle the unknown languages cases we already have and that are prefixed xx:
> https://world.openfoodfacts.org/language/unknown-language.json?fields=lc

are these created on an ongoing basis, or are those just 46 records that someone who's bored could go through and correct, and that would be that?

stephanegigandet:
> How do you handle the unknown languages cases we already have and that are prefixed xx:

Those are old products, we should not have products with unknown languages. We can just review the 46 products and assign the right language instead.

https://world.openfoodfacts.org/language/unknown-language

stephanegigandet:
> Do we know if there are pre-existing applications of the `xx` code? I read that ISO 639-2 specifies `zxx` for "non-linguistic content", but ISO 639-1 does not have any value for it.

I don't know, it seems it's at least not used right now. It makes everything much easier if we stick with a 2 letter code, as the required changes are minimal. Using more letters will require more code changes and increase a lot the risk of bugs.

Similarly, I'm hoping to use the "lc" language code in the API, so that apps can request &lc=fr&fields=ingredients_text_lc to get ingredients in French (that way the fields names are not dependent on the language)

stephanegigandet:
> i like the idea. personally, i'm curious how presenting the xx tags on the product view, and especially editing view, will work. i'm guessing suggest.pl will need modifying too?

In all languages, they will be displayed without the xx: and edited without the xx:

e.g. on the dev platform (login and pass: off)
https://de.openfoodfacts.dev/produkt/20975012/m%C3%A9lange-de-quinoa-et-ses-l%C3%A9gumes-golden-sun

This product has the fr:ab-agriculture-biologique label, which is displayed as AB Agriculture Biologique because there is a xx:AB Agriculture Biologique synonym.

It's the same if the xx: entry is the canonical name of the tag.

I created a xx:XX Label in the labels taxonomy on the dev platform to test:

https://uk.openfoodfacts.dev/product/8410076481597/tortilla-nachips-old-el-paso
the same product in another language: https://fr.openfoodfacts.dev/product/8410076481597/tortilla-nachips-old-el-paso

stephanegigandet:
> are these created on an ongoing basis, or are those just 46 records that someone who's bored could go through and correct, and that would be that?

That _should_ be that, but we might still have some bugs that create those xx entries.

e.g. there might still be some bugs in some apps that send us things that the API interprets in a bad way:

August 29, 2020 at 12:13:22 PM CEST - date-limite-app Selected images (Added: front_xx) - new image front_xx : 1.13 - view

stephanegigandet:
> . i'm guessing suggest.pl will need modifying too?

Good point. I just checked in changes to suggest.pl for that:

https://world.openfoodfacts.dev/cgi/suggest.pl?lc=fr&tagtype=labels&term=xx

aleene:
I removed the unknown languages except one: must be set to mongolian, which is available in the dropdown.

aleene:
Note that these xx are caused by apps, so a check must be added to the write api.

hangy:
> > Do we know if there are pre-existing applications of the `xx` code? I read that ISO 639-2 specifies `zxx` for "non-linguistic content", but ISO 639-1 does not have any value for it.
> 
> I don't know, it seems it's at least not used right now. It makes everything much easier if we stick with a 2 letter code, as the required changes are minimal. Using more letters will require more code changes and increase a lot the risk of bugs.
> 
> Similarly, I'm hoping to use the "lc" language code in the API, so that apps can request &lc=fr&fields=ingredients_text_lc to get ingredients in French (that way the fields names are not dependent on the language)

hmm okay. I somewhat fear the situation that one of those unused codes might be used in the future, which would mess with the internal server code and external API/data consumers.

IIRC we mostly use a regex like `^(?:(\w{2}):)?(.+)$` to get the language and tag? The two-letter ISO language codes don't seem to use any other word characters, so maybe we could use one of those? For example `0x` or even `--` instead of `xx` for no language, and `0c` or `__` instead of `lc`? It's not as intuitive but *might*  be more future proof. (It's very possible that I'm overly cautious, though. 😄)

aleene:
Agree, maybe safer to use `**` or some other non-escaped characters.

svensven:
AFAICT iso 639-1 hasn't changed since 2002-3, and the barrier to adding a new code seems quite high, so I doubt we have to worry too much. "xx" is probably safe, but "lc" might be slightly more risky. However, i don't think they would affect anything until we actually tried to add support for the new language code, especially if the system is set to reject updates with invalid language codes to prevent more xx records.

We probably also wouldn't want to use any codes that had special meanings in other contexts though, for ease of programming. E.g. underscores are single character wildcards in SQL, and asterisks in globs/regexps/etc.

aleene:
木木 it is then.

stephanegigandet:
> I removed the unknown languages except one: must be set to mongolian, which is available in the dropdown.

Thank you!

stephanegigandet:
Let's try to decide on what code to use for the default / language-less entries (we can discuss the "lc" code later, it's just an initial idea and it would be used only as a field name in the API, it wouldn't be returned or stored in the database).

The criterias that have been discussed above:

1. Futureproofness: avoid or minimize the risk of picking something that could become a code for an actual language in the future
2. Avoid or minimize the risk of breaking changes/bugs on OFF and for reusers: a code that is likely not to break something somewhere (e.g. favor  a 2 letter code instead of a 3 letter code)
3. Ease of implementation/coding (OFF + reusers): (e.g. avoid characters with special meaning, that may need escaping etc.)
4. Ease of maintaining taxonomies (e.g. no strange characters that would be difficult to type on some keyboards)
5. Intuitiveness: favor something obvious/easy to remember (e.g. **: would be a very good choice for that criteria)

Anything missing?

What are your recommendations?

My personal recommendation would still be xx: as it matches well criteria 2 to 5 . The risk for 1 seems minimal given the guidelines for adding new 2 letter codes (e.g. no 2 letter codes will be added for languages that already have a 3 letter code).

svensven:
i think xx is fine.
absolute worst case, "xx:" is also nice and distinct for search+replace.

stephanegigandet:
ok, can we merge as-is then? :)

svensven (approved):

svensven:
ooh, we got a new unknown language record since the other day: https://world.openfoodfacts.org/product/8711200475230/yula-amazonian-energy-drink

aleene:
I do not think you should pick the `xx`, as it is in use by several apps. Take some other ascii thing `££` ?

svensven:
> I do not think you should pick the `xx`, as it is in use by several apps. Take some other ascii thing `££` ?

Interesting. What does it mean in those apps?

svensven:
£ isn't ascii.
I was looking at https://en.wikipedia.org/wiki/ASCII#Character_set to see what characters don't have escaping issues.
I thought maybe @@, but that would need escaping in URLs. and 00 apparently means "toilet"...

stephanegigandet:
> I do not think you should pick the `xx`, as it is in use by several apps.

@aleene: What apps? Apps that reuse OFF you mean, or other apps? What do they use it for?

stephanegigandet:
> I thought maybe @@, but that would need escaping in URLs.

Good point, if we can avoid something that would need to be escaped in URLs, it would be a good thing.

stephanegigandet:
I was thinking of --: as well, but in facet URLs we can currently do things like /category/-en:cakes  (not in the en:cakes category). 
/category/---:something would be very confusing.

aleene:
I mean apps that reuse OFF. That is where the current xx come from.

aleene:
I guess ```
xx
``` means now **not set**

svensven:
I suppose there is a distinction between "language unknown" and "language-free" that it might be useful to encode.

But AIUI products with xx are an anomaly that shouldn't be created by apps or accepted by the server anyway, whereas this is about the taxonomies rather than products.

stephanegigandet:
In fact the apps that use OFF do not use or send us "xx", it's something that we do ourselves on our side, when they send bogus language codes that cannot be mapped to a language we know. 

`````if (not defined $lang_lc{$product_ref->{lc}}) {
                $product_ref->{lc} = 'xx';
        }
````

Probably something we should change to just reject the language code.

stephanegigandet:
I just commited a change that changes how we deal with apps that want to change the main language of a product to an incorrect language code: previously we would set the main language to "xx". Now we will just ignore that change and keep the existing main language.

So we should not have new "xx".

We do have 19 products with their main language set to "xx":

> db.products.find({ lang : "xx"}, {code : 1})
{ "_id" : "0032100000420", "code" : "0032100000420" }
{ "_id" : "4009903011119", "code" : "4009903011119" }
{ "_id" : "4019339303047", "code" : "4019339303047" }
{ "_id" : "4032523004612", "code" : "4032523004612" }
{ "_id" : "4049128181426", "code" : "4049128181426" }
{ "_id" : "42286110", "code" : "42286110" }
{ "_id" : "5410038102150", "code" : "5410038102150" }
{ "_id" : "5703161804160", "code" : "5703161804160" }
{ "_id" : "7610849682034", "code" : "7610849682034" }
{ "_id" : "8410022110960", "code" : "8410022110960" }
{ "_id" : "9020200014028", "code" : "9020200014028" }
{ "_id" : "5400112452165", "code" : "5400112452165" }
{ "_id" : "8711200475230", "code" : "8711200475230" }
{ "_id" : "7896022206451", "code" : "7896022206451" }
{ "_id" : "14423765", "code" : "14423765" }
{ "_id" : "8606100935420", "code" : "8606100935420" }
{ "_id" : "8886350068080", "code" : "8886350068080" }
{ "_id" : "8656021011109", "code" : "8656021011109" }
{ "_id" : "4388860399080", "code" : "4388860399080" }

svensven:
> We do have 19 products with their main language set to "xx":

Fixed. We're back down to the mongolian record. Any other way to find them on the site besides https://world.openfoodfacts.org/data-quality/main-language-unknown ?

stephanegigandet:
OK, once we merge the last commit and with Sven's and Aleene's cleaning, we should not have xx: entries for bogus language codes sent by apps anymore.

I think that clears the potential short-term confusion there was with xx: 

The long term risk of xx being assigned to a new language seems very small, and even if it happens, we would have years to switch to something else (i.e. there would need to be a new language, not currently with a 3 letter code, that would be assigned xx and that would be used on printed food labels).

So I'll merge this PR with xx:

We will use xx: first only as a catch-all (i.e. the canonical name is en:something or es:something, and there is a xx:entry after). So it will not be stored in the database or returned in the JSON, it will be used only for matching and displaying.

Language-less entries (where xx:something is the canonical name) should be limited to very specific taxonomies (e.g. brands).

PR #4390, opened 2020-10-21
add ecoscore bonuses for labels 

stephanegigandet:
The Eco-Score is computed from life cycle analysis (LCA) data from the Agribalyse database (based on its category), to which bonuses and maluses are added depending on labels, ingredients, origins of ingredients and packaging.

We have already implemented the LCA / Agribalyse part, this adds the labels bonuses for the production system (e.g. organic)

Related issue: #4287

### DIFF
--- a/lib/ProductOpener/Ecoscore.pm
+++ b/lib/ProductOpener/Ecoscore.pm
@@ -147,8 +147,98 @@ sub compute_ecoscore($) {
 
 	my $product_ref = shift;
 	
-	$product_ref->{ecoscore_data} = {};
+	$product_ref->{ecoscore_data} = {
+		adjustments => {},
+	};
 	
+	# Compute the LCA Eco-Score based on AgriBalyse
+	
+	compute_ecoscore_agribalyse($product_ref);
+	
+	# Compute the bonuses and maluses
+	
+	compute_ecoscore_production_system_adjustment($product_ref);
+	
+	# Compute the final Eco-Score and assign the A to E grade
+	
+	# We need an AgriBalyse category match to compute the Eco-Score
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Ecoscore.pm
```diff
@@ -188,51 +278,85 @@ sub compute_ecoscore($) {
...
+	foreach my $label_ref (@production_system_labels) {
+		
+		my ($label, $value) = @$label_ref;
+		
+		if (has_tag($product_ref, "labels", $label)
+			# Label Rouge labels is counted only for beef, veal and lamb
+			and (($label ne "fr:label-rouge")
+				or (has_tag($product_ref, "categories", "en:beef"))
+				or (has_tag($product_ref, "categories", "en:veal-meat"))
+				or (has_tag($product_ref, "categories", "en:lamb-meat")))) {
```
svensven, line 350: i don't know if other labels/categories are referred to in code, but is it worth putting a "do not rename from x" comment on the taxonomy entries, and/or unit tests to make sure they're not renamed?

svensven (approved):

lib/ProductOpener/Ecoscore.pm
stephanegigandet: It's a good idea. Another way could be to canonicalize the tags once during the module initialization, so that it still works if we switch the canonical name with a synonym.

stephanegigandet: ok I'm going to add some tests, but they are going to fail until https://github.com/openfoodfacts/openfoodfacts-server/pull/4385 is merged (because I renamed a couple of labels from en: to fr: )

so I will first merge, and then make a new PR for the tests

PR #4397, opened 2020-10-22
[WIP] Fix broken redirects in the search form

### LINKED ISSUE #4396: Search - Searching with world-en redirects to the wrong page
### What
- The following links work (https://world-de.openfoodfacts.org/cgi/search.pl?action=process&search_terms=Mehl&sort_by=unique_scans_n&json=true, https://world-fr.openfoodfacts.org/cgi/search.pl?action=process&search_terms=Mehl&sort_by=unique_scans_n&json=true)
- but not this one (https://world-en.openfoodfacts.org/cgi/search.pl?action=process&search_terms=Mehl&sort_by=unique_scans_n&json=true). 
- The only thing changed was "-en" instead of "-fr" or "-de".

### Part of
- #5492

svensven:
### What

- QUERY_STRING doesn't include the path, so redirecting to domain+query_string will lose the path. 
- REQUEST_URI is path+query_string.
- It also means the checks for `($ENV{QUERY_STRING} !~ /(cgi|api)\//)` aren't currently working.

E.g. 
https://world-en.openfoodfacts.org/cgi/search.pl?action=process&search_terms=Mehl&sort_by=unique_scans_n&json=true
currently redirects to 
https://world.openfoodfacts.org/action=process&search_terms=Mehl&sort_by=unique_scans_n&json=true

### Part of
- #5492

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -394,9 +394,9 @@ sub init()
 
 		$log->debug("subdomain matches known country name", { subdomain => $subdomain, lc => $lc, cc => $cc, country => $country }) if $log->is_debug();
 	}
-	elsif ($ENV{QUERY_STRING} !~ /(cgi|api)\//) {
+	elsif ($ENV{REQUEST_URI} !~ /(cgi|api)\//) {
 		# redirect
-		my $redirect = "$world_subdomain/" . $ENV{QUERY_STRING};
+		my $redirect = "$world_subdomain/" . $ENV{REQUEST_URI};
 		$log->info("request could not be matched to a known format, redirecting", { subdomain => $subdomain, lc => $lc, cc => $cc, country => $country, redirect => $redirect }) if $log->is_info();
 		$r->headers_out->set(Location => $redirect);
 		$r->status(301);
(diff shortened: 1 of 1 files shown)


svensven:
I'm not sure what the `($ENV{QUERY_STRING} !~ /(cgi|api)\//)` tests were actually meant to do, because most things seem to go through /cgi/display.pl anyway, even the front page, and explore link, etc. 
Changing the test to REQUEST_URI means it will probably never match anything.

stephanegigandet:
> I'm not sure what the `($ENV{QUERY_STRING} !~ /(cgi|api)\//)` tests were actually meant to do,

The redirect is meant for real users, so that they see the right subdomain in their browser. e.g. if they ask for ch-de, or fr-fr, they should get redirected to ch. and to fr. We also don't want search engines to index subdomains like fr-fr

But some API users might have issues with following redirects, and there is no value to it: we want to give back the answer right away.

It does work with the /api/ part of the path:

https://fr-fr.openfoodfacts.org/api/v0/product/3302740030918

That's because we pass the path as the query to display.pl so that display.pl can interpret it:

````````
location / {

                proxy_set_header Host $host;
                proxy_set_header X-Real-IP $remote_addr;
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;

                proxy_pass http://127.0.0.1:8001/cgi/display.pl?;
        }
````````
You are right that it does not work with /cgi as we don't pass the path as a query string for the /cgi location.

svensven:
so, would the correct behaviour be to test and redirect based on QUERY_STRING for /api, and on REQUEST_URI for /cgi ?

stephanegigandet:
> so, would the correct behaviour be to test and redirect based on QUERY_STRING for /api, and on REQUEST_URI for /cgi ?

That sounds like a good plan.

stephanegigandet:
The original bug was fixed, so I'm closing the PR. Thanks @svensven !

https://github.com/openfoodfacts/openfoodfacts-server/issues/4396

Issue #4423, opened 2020-10-26
Compute a possible estimate percent for each ingredient and sub-ingredient


stephanegigandet:
We currently compute the possible percent min and max values for each ingredient and sub-ingredient.

In order to weight some computations like the Eco-Score that depends on the origin of each ingredient, we also need to compute an estimate percent for each ingredient, and the sum of all estimates needs to be 100%.

PR #4424, opened 2020-10-26
refactor ingredient.t - move expected test results to JSON files

stephanegigandet:
A lot of the test files like ingredients.t have a simple input product, on which we run functions like processing ingredients.
The expected test result is inside the ingredients.t file which makes it very big, and also very difficult to update when the expected results change.

For instance, I'm implementing #4423 which is going to add a new "percent_estimate" to all ingredients and sub ingredients, and that will make all tests fail. Updating them is a very lengthy and error prone process with a lot of copy and paste.

So I'm refactoring the tests so that each test result is saved in an individual JSON file. When the tests run, the actual results are also saved as JSON files in a separate directory. So to see and update the diffs, we can just copy the output files over the expected files, verify the diffs are expected, and commit the new files.

### DIFF
--- a/t/ingredients.t
+++ b/t/ingredients.t
@@ -8,2349 +8,277 @@ use utf8;
 use Test::More;
 use Log::Any::Adapter 'TAP';
 
+use JSON;
+use File::Path qw/make_path/;
+use Cwd;
+
+my $dir = cwd();
+
+use ProductOpener::Config qw/:all/;
 use ProductOpener::Tags qw/:all/;
 use ProductOpener::TagsEntries qw/:all/;
 use ProductOpener::Ingredients qw/:all/;
 
-init_emb_codes();
 
-# dummy product for testing
+my @tests = (
 
-my $product_ref = {
-	lc => "fr",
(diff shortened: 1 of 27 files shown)


stephanegigandet:
We could then do the same thing in some of the other test files.

stephanegigandet:
Here is a PR that shows updated tests after adding a new percent_estimate field:
https://github.com/openfoodfacts/openfoodfacts-server/pull/4425

t/ingredients.t
```diff
@@ -8,2349 +8,277 @@ use utf8;
...
-delete $product_ref->{ingredients_percent_analysis};
-
-
-is ($product_ref->{labels}, undef) or diag explain $product_ref->{labels};
-is_deeply ($product_ref->{labels_tags}, undef) or diag explain $product_ref->{labels_tags};
+);
+
+my $testdir = "ingredients";
+
+if (! -e "test_results") {
```
hangy, line 246: Maybe place this in a well-known (`/tmp`?) or configured directory (`$data-root`). Otherwise, this depends on the exact path where the test is run from. Also, this dir should probably be `.gitignore`.

svensven:
could the writing the output files be optional? i imagine 99% of the time they won't be used, and it just adds overhead.

stephanegigandet:
@hangy @svensven : thanks for the feedback, I added an optional --results [path of results dir] option, that way we'll save the results files only when we need them to inspect the diffs / update the tests, and users can specify the path they want.

svensven (approved):

Issue #4427, opened 2020-10-27
Create an origins taxonomy, populate it and use it


stephanegigandet:
We currently use the countries taxonomy to parse origins of ingredients. It works well for countries, but origins are sometimes defined with another granularity, such as a bigger region like "South America", or a smaller region like "Normandy".

This bug is to track efforts to define this taxonomy (e.g. what to include (regions, states, fishing zones..) and what not to include in it (e.g. smaller regions, cities)), to implement a minimal version of it with a few extra entries so that we can test its integration in features such as ingredients analysis, and to see how we can populate it automatically from sources such as Wikidata.

svensven:
it would be useful for providing tagify suggestions to the not-just-country location fields on the edit form too.

PR #4428, opened 2020-10-27
Initial creation of an "origins" taxonomy + use for ingredients parsing

stephanegigandet:
This is the first phase for #4427 to have a very small origins taxonomy to start to implement it inside features such as ingredients analysis:

- Created a new origins.txt taxonomy to add a few country groups (e.g. South America) and regions (e.g. California, Brittanny)
- Automatically includes countries.txt in the origins taxonomy (same system that we use to automatically add all additives, vitamins etc. to ingredients)
- Loading of the taxonomy. That has implications for the existing origins field which now becomes taxonomized, instead of language-less entries like "France", we will have "fr:france".
- Changed ingredients parsing of origins to use the origins taxonomy instead of the countries taxonomy, so that we can match the new entries.
- Improved ingredients parsing of origins to add new types of matches like "California almonds", "Peaches from Florida".

### DIFF
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -774,12 +774,14 @@ sub display_field($$) {
 		$class = "tagify-me";
 		if ((defined $taxonomy_fields{$fieldtype}) or ($fieldtype eq 'emb_codes')) {
 			$autocomplete = "$formatted_subdomain/cgi/suggest.pl?tagtype=$fieldtype&";
-					}
+		}
 	}
 
 	my $value = $product_ref->{$field};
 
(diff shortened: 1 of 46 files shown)


svensven (approved):

aleene:
What will be added to the json?

stephanegigandet:
> What will be added to the json?

2 changes:
- The origins_tags field will contain taxonomized entries like "en:belgium" instead of "belgium" or "belgique".
- In the ingredients field, each ingredient may have an "origins" field (which was named "origin" previously, but it sometimes contains multiple origins that are comma separated, that was already the case).

PR #4429, opened 2020-10-27
Compute a possible estimate of the percent of each ingredients (with the sum equal to 100%)

stephanegigandet:
See #4423 

This is needed in particular for the Eco-Score, so that we can weight each ingredient.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -89,6 +89,7 @@ BEGIN
 		&set_percent_min_values
 		&set_percent_max_values
 		&delete_ingredients_percent_values
+		&compute_ingredients_percent_estimates
 
 		&add_fruits
 		&estimate_nutriscore_fruits_vegetables_nuts_value_from_ingredients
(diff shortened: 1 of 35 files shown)


svensven:
Doesn't this happen already? Or is this a replacement for that code?

On this topic, it might be nice if the system could understand all the american lists with "contains 2% or less of the following" type text in them, when calculating percentages. (or maybe it already does, but just doesn't take it out of the unknown ingredients.)

stephanegigandet:
> Doesn't this happen already? Or is this a replacement for that code?

It is an extension of the existing code:

- we have been computing the possible percent min and max for all ingredients and sub-ingredients (based on actual % provided and the position of each ingredient in the list). but this only gives us possible ranges for each ingredient.

- for some uses, we need an estimate (not a range), and we need the sum of all those estimates to be 100%. This is what this new code does: computing those "estimates" (that at this point are more a set of possible values that match all the min and max constraints).

stephanegigandet:
> On this topic, it might be nice if the system could understand all the american lists with "contains 2% or less of the following" type text in them, when calculating percentages. (or maybe it already does, but just doesn't take it out of the unknown ingredients.)

It definitely would be nice, but we don't have code for that yet.

lib/ProductOpener/Ingredients.pm
```diff
@@ -1976,6 +1979,98 @@ sub set_percent_sub_ingredients($) {
 }
 
 
+=head2 compute_ingredients_percent_values ( total_min, total_max, ingredients_ref )
```
svensven, line 1982: this looks like the blurb for the wrong function?

stephanegigandet: indeed, I just changed it.

svensven (approved):

PR #4449, opened 2020-10-30
Taxonomy edits 22 - ingredients & labels additions

svensven:
I'm assuming the organic label is how asterisked organic ingredients are detected in ingredients lists?

There's some duplication between the "function section" of ingredients.txt and additives_classes.txt, dunno if one should be removed, so I edited both in parallel.


stephanegigandet (approved):
Thank you!

stephanegigandet:
> I'm assuming the organic label is how asterisked organic ingredients are detected in ingredients lists?

That's right, it currently works for en:organic, en:fair-trade, it's done at preparsing time (replacing the * by organic etc.): in preparse_ingredients_text()

stephanegigandet:
> There's some duplication between the "function section" of ingredients.txt and additives_classes.txt, dunno if one should be removed, so I edited both in parallel.

Actual additives classes (as defined by the US, EU etc. food additives directives/laws) should be only in additives_classes.txt (which is automatically included in ingredients.txt). I noticed that we now have a few other functions in additives_classes.txt that are not necessarily additives classes. I guess we can keep them there for now.

svensven:
> Actual additives classes (as defined by the US, EU etc. food additives directives/laws) should be only in additives_classes.txt (which is automatically included in ingredients.txt). I noticed that we now have a few other functions in additives_classes.txt that are not necessarily additives classes. I guess we can keep them there for now.

ah, from reading the file, i'd assumed it was just a collection of generic terms, rather than based on anything formal.

PR #4465, opened 2020-11-03
Many small fixes and improvements for the Eco-score

stephanegigandet:
- new /ecoscore facet (similar to /nutrition-grades for Nutri-Score
- fixed issue with aggregation of origins percentages for nested ingredients
- added en:unknown entry to origins taxonomy
- better display of Eco-score computation details (in particular, list of the aggregated origins and percentages)
- compute the Eco-score when updating products using the edit form or the API


lib/ProductOpener/Ecoscore.pm
```diff
@@ -102,7 +102,7 @@ sub load_agribalyse_data() {
 		$csv->getline ($io);
 
 		while ($row_ref = $csv->getline ($io)) {
-			$agribalyse{$row_ref->[0]} = {
+			$agribalyse{$row_ref->[0] . ''} = {
```
VaiTon, line 105: Is this to make sure that we get a string? Couldn't we create a new "utils" method for this so it is more explicit? This way we could use it when we need without adding lots of comment explaining why do we do this.

po/common/common.pot
```diff
@@ -5067,3 +5067,12 @@ msgstr "Missing ingredients list"
 msgctxt "missing_nutrition_facts"
 msgid "Missing nutrition facts"
 msgstr "Missing nutrition facts"
+
+msgctxt "ecoscore_p"
```
VaiTon, line 5071: Is this "p" for plural?
```suggestion
# Plural noun for ecoscore
msgctxt "ecoscore_p"
```

stephanegigandet: It is for plural, but for the Eco-score there's no plural, we don't want Eco-scores.

lib/ProductOpener/Ecoscore.pm
stephanegigandet: In fact Perl does it automatically for hash keys, I'll remove it.

stephanegigandet: Removed

teolemon (approved):

PR #4469, opened 2020-11-04
Fix translations breaking html & translating slack channel name

svensven:
Noticed mismatched A tags on the turkish site, turning everything after them into a link.
Also the slack channel name was translated, despite the comment saying not to.
Added tests to prevent recurrence, which showed up 2 more errors. Fixed on crowdin too.

Update: Added test to check _all_ A tags are matched, and consequently found another error.

image


stephanegigandet (approved):
Good idea, thank you.

PR #4471, opened 2020-11-04
test: GitHub action to validate HTML in the static files

teolemon:
### What
- Validate HTML static files
- Note: I've made the same PR on openfoodfacts-web
### Part of 
- #5512


VaiTon:
@teolemon it would be better if we run this only if HTML files are changed.

PR #4472, opened 2020-11-04
Oceans and seas

teolemon:
Most popular oceans and seas, up until `ocean-indien-ouest | 16 | *`


stephanegigandet (approved):
Let's merge it but please don't add more entries as we will deal with FAO zones differently (not repeated for every language) and we'll need tests for it.

For origins, I think it's best that the canonical name is always an en: entry

teolemon:
I added en: entries

stephanegigandet (approved):

PR #4473, opened 2020-11-04
Packaging adjustment for the ecoscore, #4287

stephanegigandet:
Taking into account packaging is the last remaining adjustment for the Eco-score computations.

This is a start, there are subtleties that will need to be implemented (e.g. the adjustment for PET depends on whether it is a bottle or something else).

The overall quality of the packaging adjustment will also directly depend on the quality of the packaging data structure, the underlying data and the parsing/analysis of it.

Included:
- some additions to the countries and origins taxonomies so that the Eco-score countries can be easily matched.
- improvements to the packaging_materials taxonomy to account for all the materials taken into account by the Eco-score
- ecoscore packaging malus computation
- updated tests

Will be done soon:
- displaying the Eco-score packaging computation details in the Eco-score details


lib/ProductOpener/Ecoscore.pm
```diff
@@ -496,6 +685,8 @@ The percentages are stored in $aggregated_origins_ref
 
 =cut
 
+sub aggregate_origins_of_ingredients($$$);
+
```
svensven, line 689: is this intended? i don't know if it's duplication, or some perl feature i don't know about. :)

svensven (approved):

lib/ProductOpener/Ecoscore.pm
stephanegigandet: It's intended, the function is recursive, and otherwise Perl emits a warning that the function is "called too early to check prototype". The solution to avoid it is to predeclare the function.

See https://www.perlmonks.org/bare/?node_id=130071

svensven: fascinating. i have seen that message before in the logs, didn't know that was the/a cause.

PR #4474, opened 2020-11-06
Display search results with cards using user preferences

stephanegigandet:
This is currently only for the new /search endpoint, to test personal search.

Each product is shown on a card with the photo + name, as well as small icons for the attributes that correspond to selected user preferences.

The background of the card is green, red or grey, depending on whether the product matches the user preferences.

Once we finalize this display, the goal is to make it the display for facets as well, not just search results.

With a lot of attributes selected:

image

"normal" usage would have much less attributes selected. By default it would be Nutri-Score + NOVA + Eco-score.

### DIFF
--- a/html/js/product-preferences.js
+++ b/html/js/product-preferences.js
@@ -88,7 +88,8 @@ function display_selected_preferences (target_selected, target_selection_form, p
 		$( target_selection_form ).show();
 		$(document).foundation('equalizer', 'reflow');
 	});
-
+	
+	$(document).foundation('reflow');
 }
 
 
@@ -169,7 +170,7 @@ function display_user_product_preferences (target_selected, target_selection_for
 				
 				if (attribute.description_short) {
 					attribute_group_html += "<p class='attribute_description_short'>" + attribute.description_short + "</p>";
-				}
+				}			
 				
 				attribute_group_html += "<hr style='clear:left;border:none;margin:0;margin-bottom:0.5rem;padding:0;'>";
 				
--- a/html/js/product-search.js
+++ b/html/js/product-search.js
@@ -146,16 +146,8 @@ function display_products(target, product_groups ) {
 		$.each( product_group, function(key, product) {
 		
 			var product_html = "";
-			
-			var color = '#eee';
-			if (product.match_status == "yes") {
-				color = '#cfc';
-			}
-			else if (product.match_status == "no") {
-				color = '#fcc';
-			}
 						
-			product_html += '<li><a href="' + product.url + '" class="list_product_a" style="background-color:' + color + ';">';
+			product_html += '<li><a href="' + product.url + '" class="list_product_a list_product_a_match_' + product.match_status + '">';
 			product_html += '<div class="list_product_img_div">';
 			
 			if (product.image_front_thumb_url) {
--- a/html/js/product-search.js
+++ b/html/js/product-search.js
@@ -118,6 +118,7 @@ function rank_products(products, product_preferences) {
 	});
 	
 	var product_groups = {
+		"all" : [],
 		"yes" : [],
 		"unknown" : [],
 		"no" : [],
@@ -126,6 +127,7 @@ function rank_products(products, product_preferences) {
 	$.each( products, function(key, product) {
 
 		product_groups[product.match_status].push(product);
+		product_groups["all"].push(product);
 	});
 	
 	return product_groups;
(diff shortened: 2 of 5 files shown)


stephanegigandet:
If you want to test, it's currently live on the dev server:
https://fr.openfoodfacts.dev/search?user_preferences=1&codes_tags=330xxxxxxxxxx&ecoscore_tags=a

html/js/product-preferences.js
```diff
@@ -169,7 +170,7 @@ function display_user_product_preferences (target_selected, target_selection_for
 				
 				if (attribute.description_short) {
 					attribute_group_html += "<p class='attribute_description_short'>" + attribute.description_short + "</p>";
-				}
+				}
```
svensven, line 173:
```suggestion
				}
```

html/js/product-search.js
```diff
@@ -126,6 +127,7 @@ function rank_products(products, product_preferences) {
 	$.each( products, function(key, product) {
 
 		product_groups[product.match_status].push(product);
+		product_groups["all"].push(product);
```
svensven, line 130:
```suggestion
		product_groups.all.push(product);
```

svensven:
I'm not a huge fan of how much more space each result takes up. (and the "Your food preferences" box too, tbh.) on my laptop, 6 per row instead of 9 per row. but that's a trend in web stuff generally that annoys me - cf twitter & facebook redesigns.
also, if you have no preferences, everything shows up as green and "yes", which doesn't really mean anything.

html/js/product-search.js
```diff
@@ -143,45 +145,50 @@ function display_products(target, product_groups ) {
...
-			product_html += "<span>" + product.product_name + "</span>";
-			
-			product_html += '</a>';
-			
+			product_html += '<div style="height:80px;overflow:hidden;">' + product.product_name + "</div>";
+									
 			$.each(product.match_icons.mandatory.concat(product.match_icons.very_important, product.match_icons.important), function (key, icon_url) {
 				
-				product_html += '<img src="' + icon_url + '" class="match_icons">';
+				product_html += '<img src="' + icon_url + '" style="height:24px;margin-right:0.2rem;margin-top:0.2rem;">';
```
svensven, line 171: will the style= stuff go into classes at some point?

svensven (approved):

html/js/product-search.js
stephanegigandet: Yes, once the styles are finalized we will definitely move them. To tinker with the styles I find it much easier to have them close to the HTML and somewhere where the effects are immediate without having to rebuild the scss stuff etc.

stephanegigandet:
> I'm not a huge fan of how much more space each result takes up. (and the "Your food preferences" box too, tbh.) on my laptop, 6 per row instead of 9 per row. but that's a trend in web stuff generally that annoys me - cf twitter & facebook redesigns.

The "Your food preferences" box is too big indeed, maybe we can remove the box completely and just leave a button or something.

I tweaked things a bit to display more results per row (1, 4, 6, 8 or 10 depending on screen size, compared to 1, 2, 4, 6, or 8 before). On your laptop you should get 8 per row now.

image

I also improved the 1 result per row on small screens, to display the product name and icons (bigger icons in fact) to the right of the product image, to save vertical space.

> also, if you have no preferences, everything shows up as green and "yes", which doesn't really mean anything.

Right, if there are no preferences, then I guess we can just remove the tabs completely. And we can keep the green as a default color, or put some light grey I guess.

html/js/product-search.js
stephanegigandet: Fixed.

stephanegigandet:
> I'm not a huge fan of how much more space each result takes up. [..]
> also, if you have no preferences

@svensven : That made me wonder how to make you a huge fan of it. The current preferences are geared towards people who are looking to eat or buy products. But they are not very useful to OFF contributors. But maybe we could use the same preferences mechanism to display other things that contributors would be more interested in: like products that need to be completed in a certain way, products that have Robotoff suggestions or things like that. Any ideas of things that could be useful to you?

svensven:
> @svensven : That made me wonder how to make you a huge fan of it. The current preferences are geared towards people who are looking to eat or buy products. But they are not very useful to OFF contributors. But maybe we could use the same preferences mechanism to display other things that contributors would be more interested in: like products that need to be completed in a certain way, products that have Robotoff suggestions or things like that. Any ideas of things that could be useful to you?

Hmm, that's an interesting idea. some ideas off the top of my head:
- has photos uploaded but no front photo selected (i.e. whether it's worth clicking through to see if a front photo can be selected)
- has unrecognised ingredients 
- has any of the more interesting/fixable flags from https://world.openfoodfacts.org/data-quality , i.e. not nutrition-value-very-high/low-for-category*, ingredients-*-ending-comma, nutrition-value-under*salt...
- has ingredients photo selected but no ingredients entered
- has nutrition photo selected but no nutrition data entered
- you are the most recent editor of this record (handy when you're working through search results and lose track)

stephanegigandet:
> Hmm, that's an interesting idea. some ideas off the top of my head:

Thanks for the ideas! I put them on a separate wiki page, so that we can broaden a bit the discussion. I think it's worth making it a project on its own.

https://wiki.openfoodfacts.org/Product_Attributes_for_Contributors

PR #4480, opened 2020-11-08
PNG version of the attributes icons + conversion script

stephanegigandet:
This PR adds a PNG version of the icons used for allergens, ingredients analysis, labels etc. as it's difficult to use the SVG icons on iOS and Android.

There's also a convert script which needs Inkscape version 1 or greater to be installed.

@VaiTon : I drew the icons based on a 24x24 grid, so if you can, it's best to display them with a height that is a multiple of that, like 24px or 48px. I rendered them at 192px which should be enough even for very high density displays.


hangy:
Couldn't this be done via gulp, and not with static files in the repo? That way, it would be easier to keep the SVG and generated PNG in sync, and we wouldn't be redundant files in git.

stephanegigandet:
> Could we do this via gulp instead, and not with static files in the repo? That way, it would be easier to keep the SVG and generated PNG in sync, and we wouldn't be redundant files in git.

Probably, but I'm really not used to gulp, so the easy thing for me is just to commit the png files...

I'm not sure if it is really worth it to spend time on this, and whether we'll have issues (now or in the future) if we have gulp install inkscape and its dependencies to regenerate the icons each time we run it.

Alternatively, we could just commit the convert_svg_icons_to_png.sh and remove the PNG icons. I will just run it once in production. The PNG icons are used only by apps.

stephanegigandet:
BTW, we have to use Inkscape instead of ImageMagick has ImageMagick SVG support isn't very good (e.g. with convert / mogrify, we can lose transparency and it's hard to get a specific height and not losing quality).

mogrify removes transparency and creates a weird interpolation.

image

Inkscape converts the file as intended:

image

hangy:
> BTW, we have to use Inkscape instead of ImageMagick has ImageMagick SVG support isn't very good

There's alternatives like https://github.com/ben-eb/gulp-rsvg, which can probably be adapted to use
https://github.com/f3lang/node-rsvg-prebuilt.

stephanegigandet:
> There's alternatives like https://github.com/ben-eb/gulp-rsvg, which can probably be adapted to use
> https://github.com/f3lang/node-rsvg-prebuilt.

Certainly, but what's the point? I mean yes it's certainly possible, but there may be issues (e.g. from the example, I don't know if / how the SVG icons that are defined at something x 24px can be correctly converted to 192 px height) that will need time to investigate. Time that I would prefer to spend on things that will make a bigger difference.

My view of it is that we have to balance cost / benefit / risk, and sometimes automating something may not be the most beneficial investment of our time.

How about either just leaving the PNG files in git (they are not very big), or removing them and just leaving the one line convert script so that we know how to generate them when we need them?

svensven (approved):

PR #4481, opened 2020-11-08
Adding the Israeli mandatory labels

yarons:
The green label is described here (in Hebrew):
https://www.health.gov.il/Subjects/FoodAndNutrition/Nutrition/Adequate_nutrition/FoodLabeling/Pages/green-labeling.aspx

The red labels are described here (in Hebrew):
https://www.health.gov.il/Subjects/FoodAndNutrition/Nutrition/Adequate_nutrition/FoodLabeling/Pages/labeling.aspx

The SVGs were extracted from the official PDFs accordingly.


stephanegigandet:
The files need to be renamed from high-level-of-fat.113x113.svg  to high-level-of-fat.90x90.svg : it is the way the server knows what is the width of a logo when it is displayed at 90px height.

stephanegigandet (changes requested):
The files need to be renamed from high-level-of-fat.113x113.svg to high-level-of-fat.90x90.svg : it is the way the server knows what is the width of a logo when it is displayed at 90px height.

yarons:
Hmmmm I'll do that of course but I really don't understand why, it's SVG, the whole purpose of SVG is to just fit everywhere and adjust to whatever contraint you impose on it while keeping maximum resolution.

yarons:
I wish I could squash commits but this repo is huge and I don't want to wait for it to download for such a trivial change, thank you and sorry.

stephanegigandet:
Yes, you don't have to change the SVG, it's fine if it is 113 x 113, just change the name. It's just a naming convention we use so that we can get the aspect-ratio of the file. We could do it differently, but that's how it's implemented today. Thank you!

stephanegigandet:
Another thing is that we need to add entries for those labels in the labels taxonomy: labels.txt

For some reason I can't load the https://www.health.gov.il/Subjects/FoodAndNutrition/Nutrition/Adequate_nutrition/FoodLabeling/Pages/labeling.aspx page.

We need to add entries with names that indicate that they are specific to Israel (at least for the canonical name).

The entries should look like this:

he:[name in Hebrew]
xx:Israel Green Label  --> the xx: entry will be used to describe the label for other languages

Could you give me the URL of some products on OFF that have those labels?

yarons:
The website doesn't work for me either, it's probably down.

The Israeli Cottage contains the high level of fat red label:
https://il.openfoodfacts.org/product/7290004127336/%D7%A7%D7%95%D7%98%D7%92-9-%D7%AA%D7%A0%D7%95%D7%91%D7%94

These cereals contains the high level of sugar red label:
https://il.openfoodfacts.org/product/5053827194891/cereal-kellogg-s

I still have this around the house and it contains both high fat and high sodium:
https://il.openfoodfacts.org/product/7290003075553/%D7%A8%D7%92%D7%B3%D7%91

I can't find anything with the green label.

BTW if I'm adding this to the labels.txt, where is the connection to the SVG?

stephanegigandet:
> BTW if I'm adding this to the labels.txt, where is the connection to the SVG?

We will have to rename the SVGs so that they match the label names in the taxonomy.

yarons:
> > BTW if I'm adding this to the labels.txt, where is the connection to the SVG?
> 
> We will have to rename the SVGs so that they match the label names in the taxonomy.

OK, which names do you think we should use?

BTW I have many more labels for Israeli Kashrut, this upload is basically a test for the next round.

stephanegigandet:
@yarons sorry I missed your comment.

For the labels taxonomy, it should be like this:

he:[name of the label in Hebrew, as it appears on packaging or as it is commonly called]
xx:Israel Green Label 

he xx: entry will be used to describe the label for other languages

Then for the file name, it should match the name in Hebrew (hopefully it will work in file names, we'll need to test)

yarons:
Is #4611  correct?
We should really start transitioning this data to Wikidata, it'll be a lot easier to manage (but not to translate, unless we'll create some master SPARQL queries for translators).

yarons:
@stephanegigandet I've added all the required changes, can you please verify I did this correctly?

stephanegigandet (approved):

PR #4484, opened 2020-11-08
Add support for the Eco-Score for Offline-Scan

teolemon:
Add support for the Eco-Score for Offline-Scan


stephanegigandet (approved):

teolemon:
Note that this PR creates potential early leaks. It won't be visible directly in the apps, unless there's an explicit update to load and show the column though.

PR #4495, opened 2020-11-09
bigger button icons, fixes #4494

### LINKED ISSUE #4494: Icons are too small on the OFF web site
It's very hard to see the buttons icons on the web site. The material icons are designed at 24x24 px, but on my screen, they are displayed at 13x13 in most of the buttons.

image

stephanegigandet:
On my screen the button icons are so small (13x13) that it's very hard to see what they are. 

Before:
image

After:
image

### DIFF
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -344,14 +344,33 @@ details.recent {
   cursor: pointer;
 }
 
+.top-bar .button, .top-bar button {
+  padding-top:0.2rem;
+  padding-bottom:0.2rem;
+}
+
+.button.small, button.small {
+  padding:0.5rem;
+}
+
+.postfix.button.small, button.postfix.small {
+  padding:0;
+}
+
 .icon {
   display: inline-block;
   fill: currentColor;
-  width: 1em;
-  height: 1em;
+  width: 24px;
+  height: 24px;
   margin-left: 0.25em;
   margin-right: 0.25em;
   overflow: hidden;
+  vertical-align:middle;
+}
+
+.tiny svg.icon {
+  width:16px;
+  height:16px;
 }
 
 .ingredients_analysis {
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -345,12 +345,14 @@ details.recent {
 }
 
 .top-bar .button, .top-bar button {
-  padding-top:0.2rem;
-  padding-bottom:0.2rem;
+  padding-top:0.4rem;
+  padding-bottom:0.4rem;
 }
 
-.button.small, button.small {
+.button.small, button.small, button.tiny, .button.tiny {
   padding:0.5rem;
+  padding-left:0.75rem;
+  padding-right:0.75rem;
 }
 
 .postfix.button.small, button.postfix.small {
@@ -360,17 +362,13 @@ details.recent {
 .icon {
   display: inline-block;
   fill: currentColor;
-  width: 24px;
-  height: 24px;
   margin-left: 0.25em;
   margin-right: 0.25em;
   overflow: hidden;
   vertical-align:middle;
-}
-
-.tiny svg.icon {
-  width:16px;
-  height:16px;
+  font-size:18px;
+  width:1em;
+  height:1em;
 }
 
 .ingredients_analysis {


svensven (approved):

hangy (commented):
Have you tested this in other screen sizes and dpis? `1em` should be pretty good since it scales with the font size automatically. Switching to fixed `px` sounds like a disadvantage IMHO.

hangy:
> After:
> image
> 

The smartphone on the top right and the gear next to the sign-out button are way too big.

VaiTon:
Maybe increase to 1.5em?

teolemon:
Extra points if you manage to get "Get the app" at a 13-inch resolution next to the phone, while taking lenght variation by country, and not making the other parts disappear.

stephanegigandet:
> Have you tested this in other screen sizes and dpis? `1em` should be pretty good since it scales with the font size automatically. Switching to fixed `px` sounds like a disadvantage IMHO.

1em is what we have today. On a desktop screen it's very small.

Fixed px have an advantage for small icons that have been drawn on a specific grid (24x24 for the material icons): it reduces anti-aliasing if we select sizes that more or less match multiples of the original grid.

e.g. this what I see on my screen:

1em --> resulting size = 13x13 px

image

1.25em --> resulting size = 20x20 px

image
image

1.5em --> resulting size = 24x24 px

image

hangy:
> > Have you tested this in other screen sizes and dpis? `1em` should be pretty good since it scales with the font size automatically. Switching to fixed `px` sounds like a disadvantage IMHO.
> 
> 1em is what we have today. On a desktop screen it's very small.

I never saw it as an issue, to be honest. Also, desktop screens vary massively, from 1920x1080 up to 4k res on the same screen sizes. We need to look good on mobile, too.

> Fixed px have an advantage for small icons that have been drawn on a specific grid (24x24 for the material icons): it reduces anti-aliasing if we select sizes that more or less match multiples of the original grid.

IIRCs, every guide about SVG icons that I've read recommended `em` sizes. After all, they're often used (as in our case) to replace icon fonts, which were naturally linked to the size of the surrounding text. Maybe there are better relative sizes that could be used, but generally, sizes relative to the text size have the advantage of scaling with user's overrides (accessibility), and that they can be used everywhere (ie. `h1` and `footer`) without having to define new icon sizes.

stephanegigandet:
> Also, desktop screens vary massively, from 1920x1080 up to 4k res on the same screen sizes. We need to look good on mobile, too.

Yes, definitely.

> sizes relative to the text size have the advantage of scaling with user's overrides

That's actually true for the icons with a fixed height in pixels too. If you use Ctrl-+ you will see the icons getting bigger.
Same thing for displays with high density: on my phone, icons with a height set at 24px are actually displayed at 96 pixels.

Here's the actual size that is rendered on my phone:

image

> IIRCs, every guide about SVG icons that I've read recommended em sizes. 

I don't know, but that's not what Google recommends for the material design icons we use:
https://google.github.io/material-design-icons/#sizing

> Although the icons in the font can be scaled to any size, in accordance with material design icons guidelines, we recommend them to be shown in either 18, 24, 36 or 48px. The default being 24px.
> 
> CSS rules for the standard material design sizing guidelines:
> 
> .material-icons.md-18 { font-size: 18px; }
> .material-icons.md-24 { font-size: 24px; }
> .material-icons.md-36 { font-size: 36px; }
> .material-icons.md-48 { font-size: 48px; }
> 
> Material icons look best at 24px, but if an icon needs to be displayed in an alternative size, using the above CSS rules can help:

Maybe the best way to go forward is to test in as many different kinds of displays / conditions as possible (both the current design with the 1em icons, and the new design with 24px icons), so that we can see what needs to be changed.

I'll make the gears and smartphone icons smaller.

stephanegigandet:
OK, I tested with the smallest 18px size recommended by Google for the material icons, and I think it's good compromise. The icons are much more legible than what we have today, but they aren't too big either.

I also changed the style to use font-size:18px;width:1em;height:1em; 

Here is what it looks like:
image

Everyone: it's on the dev server if you want to try it on your own devices. Could you try and compare with what we have in production? If there is something that needs to be fixed, please add a screenshot. Thank you!

VaiTon (approved):

PR #4496, opened 2020-11-10
Add lang= attribute to moderator ingredients edit on product view

svensven:
So the browser can decide what to spellcheck it as. Also some tidying.


stephanegigandet (approved):
Looks good to me.

hangy (approved):

PR #4508, opened 2020-11-12
improvements for packaging parsing

stephanegigandet:
- adds singular for "metals" (to recognize "metal lid" / "couvercle en métal" etc.)
- new en:unrecognized shape so that we know when we have an entry that we could not recognize (e.g. for a product packed in something that is not in the taxonomy), and also so that we can avoid merging entries that should not be merged


lib/ProductOpener/Display.pm
```diff
@@ -4372,12 +4372,10 @@ sub customize_response_for_product($) {
 				}
 			}
 		}
-		
 		# Eco-Score
 		elsif (($field =~ /^ecoscore/) and (not defined $product_ref->{ecoscore_data})) {
 			compute_ecoscore($product_ref);
 			$customized_product_ref->{$field} = $product_ref->{$field};
```
svensven, line 4378:
```suggestion
			$customized_product_ref->{$field} = $product_ref->{$field};
		}
```

svensven (approved):

lib/ProductOpener/Display.pm
stephanegigandet: Thanks!

PR #4510, opened 2020-11-13
Personal search - Enable user preferences for facets pages and search results

stephanegigandet:
This makes the new products cards view the default (only for moderators at this point) for facets pages and search results.

image
Currently live on the dev server, product images not shown because their are not on the dev server)

In practice the search results are still generated as HTML, but also added as JSON data so that they can be displayed with cards. (that way we still display results when javascript is OFF, and search engines can continue to easily index them).

This is only for moderators so that we can use it, see how well it works in practice (for regular users but also for contributors).

Some caveats:

1. The ranking according to user preferences only applies to the results loaded on the page. (I changed the number of results per page from 20 to 24 so that we have full rows for all breakpoints (1, 2, 4, 6 and 8). (before we did not have breakpoints, it was just a list of floats)

2. I'm not sure what's the best thing to do with the tabs, in terms of what goes in them, and whether to show them or not, and with what wording.

Currently products can go into "Not OK" if the user has set a "mandatory" preference. And they can go in "maybe" if there is an "important" preference for which we don't have enough data.

On small screens the 4 tabs can be on 2 lines which is not very good. One way could be to remove the numbers of products in each tab, or we could replace the text by icons (check mark, question mark, cross mark)

3. This completely changes the HTML structure of the results page, so user scripts like the power user script will not work (unless they are changed to use the new format).

### DIFF
--- a/html/images/icons/match-no.svg
+++ b/html/images/icons/match-no.svg
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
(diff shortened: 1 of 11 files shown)


stephanegigandet:
I think I found a better solution for tabs:

image

teolemon (approved):

PR #4513, opened 2020-11-14
Honey

moon-rabbitOFF:

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -26620,7 +26620,7 @@ it:Caramelle
 nl:Snoep
 nl_be:Snoep
 pl:Cukierki, słodycze, cukierek, słodkości
-pt:Bombons, doces
+pt:Rebuçados e gomas
 ru:Сладости
 sv:Godis, Sötsaker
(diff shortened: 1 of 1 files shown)


taxonomies/categories.txt
```diff
@@ -69431,7 +69489,7 @@ hu:Dzsemek, Lekvárok
 it:Confetture, Confettura
 nl:Jams, Confituren
 pl:Dżemy, konfitury, powidła
-pt:Geléias
+pt:Doces
```
stephanegigandet, line 69492: Hi @moon-rabbitOFF ,
There's an issue with this change:

ERROR - pt:doces already is a synonym of pt:bombons for entry en:bonbons - pt:doces cannot be mapped to entry en:jams

stephanegigandet (changes requested):

taxonomies/categories.txt
moon-rabbitOFF: I fixed it

stephanegigandet (approved):
Thank you!

PR #4533, opened 2020-11-18
add class for page type, fixes #4528

### LINKED ISSUE #4528: Add page type information into HTML code to help third-party scripts
Third-party scripts, at least Power User Script, need to know the **type of the page** to add features or not, based on it. For example, Power User Script has different behaviours based on these different types:
* product view
* edit view
* saved-product view (seen after validation of an edit view)
* product list (facet, search result, home page, etc.)
* recent changes
* list of ingredients
* search form

Maybe there are other types which I have forgotten.

This information can also help for debugging purposes and even styling.

**Implementation**
It can be done with `<meta>` but the best solution could be a class attached to the `<body>` or the `<div id="main_column">`. Example: `<body class="product-view">` or `<div id="main_column" class="product-view">`.

stephanegigandet:
fixes #4528

this adds a class to the body tag, e.g. <body about="/produit/26018423/farine-de-ble-bio-type-65-bon-et-bio" typeof="food:foodProduct" class="product_page">

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -1090,7 +1090,8 @@ sub display_error($$)
 	display_new( {
 		title => lang('error'),
 		content_ref => \$html,
-		status => $status
+		status => $status,
+		page_type => "error",
 	});
 	exit();
 }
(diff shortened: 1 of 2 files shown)


CharlesNepote (approved):
I didn't test but if I understand well the code, it seems ok for me.

PR #4534, opened 2020-11-18
Improvements for product cards

stephanegigandet:
1. Display product name + brand + quantity and not just product name
2. Display the barcode number if we don't have a product name
3. Display a jar silhouette icon if we don't have a product picture
4. Add a title tooltip when hovering on top of the attributes icons to show what they are and the associated details / recommendation.


teolemon (approved):
Approved
- [ ] Small conflict to solve
- [ ] A screenshot of the card (before/after) would be nice

stephanegigandet:
@teolemon : it looks like this:

image

+ tooltips when you hover the icons, but I can't capture them

teolemon:
image

PR #4554, opened 2020-11-26
Mostly German ingredients

jolo0815:
**Description:**

I added one Danish and some German translations for ingredients.

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -11514,7 +11514,7 @@ wikipedia:en:https://en.wikipedia.org/wiki/Retinyl_acetate
 
 <en:vitamin A
 en:Retinyl Palmitate
-de:Retinylpalmitat
+de:Retinylpalmitat, Vitamin A
 es:palmitato de retinilo
 fi:retinyylipalmitaatti
 fr:palmitate de rétinol
@@ -11540,7 +11540,7 @@ bg:Бета каротин
 ca:beta-carotè, beta-carotens,betacarotè
 cs:beta-karoten
 da:Betakaroten
-de:Beta-Carotin, Vitamin A, Vit. A
+de:Beta-Carotin
 el:β-καροτένιο
 es:beta-caroteno
 et:Beetakaroteen
@@ -11630,7 +11630,7 @@ ca:tiamina, Vitamina B1
 cs:thiamin, vitamin B1
 cy:Thiamin, Fitamin B1
 da:Tiamin, Vitamin B1
-de:Thiamin, Vitamin B1, Vit. B1
+de:Thiamin, Vitamin B1
 dv:ވިޓަމިން ބީ1
 el:θιαμίνη, βιταμίνη Β1
 eo:Tiamino, Vitamino B1
@@ -11726,7 +11726,7 @@ ca:riboflavina, Vitamina B2
 cs:riboflavin, vitamín B2
 cy:Ribofflafin, Fitamin B2
 da:Riboflavin, Vitamin B2
-de:Riboflavin, Vitamin B2, Vit. B2
+de:Riboflavin, Vitamin B2
 dv:ވިޓަމިން ބީ2
 el:ριβοφλαβίνη, βιταμίνη Β2
 eo:riboflavino, vitamino B2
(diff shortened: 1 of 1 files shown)


taxonomies/ingredients.txt
```diff
@@ -43234,7 +43235,7 @@ carbon_footprint_fr_foodges_value:fr:0.8
 <en:cabbage
 en:white cabbage
 da:hvidkål
-de:Weißkohl, Weisskohl
+de:Weißkohl, Weisskohl, Weißkohl
```
svensven, line 43238: this appears to be a duplicate

taxonomies/ingredients.txt
```diff
@@ -11540,7 +11540,7 @@ bg:Бета каротин
 ca:beta-carotè, beta-carotens,betacarotè
 cs:beta-karoten
 da:Betakaroten
-de:Beta-Carotin
+de:Beta-Carotin, Vitamin A, Vit. A
```
svensven, line 11543: this should probably go in the vitamin A entry just above beta carotene.

taxonomies/ingredients.txt
```diff
@@ -29307,6 +29307,7 @@ es:proteína de guisante hidrolizada,proteínas de guisante hidrolizadas
 <en:pea protein
 en:pea protein isolate
 es:aislado de proteina de guisante
+de:Erbsenproteinisolat
```
svensven, line 29310: ideally, this would go above the es: entry so the languages are in alphabetical order (after en:)

jolo0815: Oh, yeah. Is there a simple way to change it in this pull request?

taxonomies/ingredients.txt
jolo0815: Ups, yes it is.

taxonomies/ingredients.txt
jolo0815: Okay, seems like I overlooked the Vitamin entry. I was wondering why there "were" none.

Ban3: `Vit.` also can't be handled with the taxonomies because of the `.`. It needs to be expanded in code.

taxonomies/ingredients.txt
svensven: any commits you push to the branch the pull request is from will be automatically added on to the pull request.

taxonomies/ingredients.txt
jolo0815: Okay, so I should delete all `Vit. X` entries, right? We should probably explain that to people adding Ingredients because I saw it this way on the packeting.

jolo0815:
How can I make the corrections to the ingredients file?

svensven:
> How can I make the corrections to the ingredients file?

edit the one you've already changed, commit it again, push the branch up to github again, and the pull request will update.

taxonomies/ingredients.txt
svensven: hmm, it looks like there's no german section of %abbreviations in lib/ProductOpener/Ingredients.pm yet.

jolo0815:
Like this?

taxonomies/ingredients.txt
```diff
@@ -11514,7 +11514,7 @@ wikipedia:en:https://en.wikipedia.org/wiki/Retinyl_acetate
 
 <en:vitamin A
 en:Retinyl Palmitate
-de:Retinylpalmitat
+de:Retinylpalmitat, Vitamin A
```
svensven, line 11517:
```suggestion
de:Retinylpalmitat
```
Ah, the "Vitamin A" isn't needed, as the translation already exists in the parent entry. (The <en:... line indicates the entry's parent entry.) A bit further up in ingredients.txt is "de:Vitamin A". 

The structure here is that "vitamin A" is the generic term, and then there are child terms for specific forms of vitamin A, like beta carotene etc.

Another way to view the hierarchy (although it's not a live view of what's in github) is on the OFF site. E.g.
https://world.openfoodfacts.org/ingredient/vitamin-a
https://de.openfoodfacts.org/zutat/vitamin-a

taxonomies/ingredients.txt
stephanegigandet: Right, we will handle "Vit. X" entries during parsing. We have some code in place to expand abbrevations: https://github.com/openfoodfacts/openfoodfacts-server/blob/master/lib/ProductOpener/Ingredients.pm#L261

svensven (approved):
thanks for the contributions. you should be able to click the "merge" button now. :)

jolo0815:
Feels good my first contribution on Github (except opening some issues)

jolo0815:
Do I have to redownload the file before making a new PR?

PR #4563, opened 2020-11-27
mostly added German words that end the ingredient list

jolo0815:
**Description:**

Added German "ingredient list end words" and one stopword.
I am not sure about the syntax. I try to keep the edits small so if there is a fatal mistake it is not to much work which goes to waste.

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -65,192 +65,22 @@ stopwords:zh::配料
 # comment:en:These are words that end ingredient lists.
 
 #de:
-#Allergenhinweis
-#aus biologischem
-#aus biologischer
-#aus fairem
-#aus fairer
-#aus kontrolliert
-#aus kontrollierter
-#aus ökologischem
-#aus ökologischer
-#das mischungsverhältnis
-#diese packung
-#	'davon ges(â|a|ä)ttigte'
-#	'Durchschnittliche N(â|a|ä)hrwerte'
-#	'Ern(â|a|ä)hrungswerte'
-#Glutenfrei
-#Kann andere
-#Kann andere Schalenfrüchte enthalten.
-#Kann auch
-#Kann bei
-#Kann ei
-#Kann enthalten
-#Kann Erdn(ü|u)sse
-#Kann gluten
-#Kann glutenhaltige
-#Kann glutenhaltiges
-#Kann Hasseln(ü|u)sse
-#Kann Kern
-#Kann Kerne
-#Kann Mandeln
-#Kann milch
-#Kann N(ü|u)sse
-#Kann schalen 
-#kann schalenfrüchte
-#Kann sellerie
-#Kann sesamsamen
-#Kann Spuren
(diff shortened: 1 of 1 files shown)


taxonomies/ingredients.txt
```diff
@@ -65,17 +65,54 @@ stopwords:zh::配料
 # comment:en:These are words that end ingredient lists.
 
 #de:
+#Allergenhinweis
```
stephanegigandet, line 68: The lines starting with # are comments, they don't actually change how ingredients are parsed. I'm not sure why we have those comments here. We do have some code to remove everything after some phrases in Ingredients.pm (but it must be used with carefully, many of the entries below can be before or inside ingredients lists).

jolo0815: So I would have to added them somewhere else, where is that file excatly?
Okay I understand that I have to be careful because some of them could  also be before the Ingredient list like "diese packung" but which of those do you think can be inside the ingredient list, could you give an example because I haven't found any. I haven't just added "kann" because it also is in "PeKANNuss"...

Ban3: Regular expressions for the start of the list in German: https://github.com/openfoodfacts/openfoodfacts-server/blob/f61c9ca957672f758a6cf41c6c42c430c9d166f8/lib/ProductOpener/Ingredients.pm#L2562-L2565

End of the list: https://github.com/openfoodfacts/openfoodfacts-server/blob/f61c9ca957672f758a6cf41c6c42c430c9d166f8/lib/ProductOpener/Ingredients.pm#L2829-L2856

jolo0815: Can I undo the ingredients commit or do I have to create a new PR?

stephanegigandet: @jolo0815 : I don't speak German, but phrases like Glutenfrei for instance could be listed inside ingredients list, for a specific ingredient.  e.g. in English we can have things like "Ingredients: chocolate (may contain nuts), sugar, milk (organic)" etc.

Those phrases are used to cut ingredients lists that we get from the OCR (some people don't crop ingredients, so with the OCR we get a bunch of phrases that appear before or after the ingredients list). We want to remove them, but only when we are 100% sure that we are not going to remove parts of the ingredients as well.

You can close this PR and make a new one, or you could create a new commit to remove the comments.

I think we should remove the comments of this section in the other languages as well, they are confusing. They might have been useful at some point, but it's not useful if people add some other entries expecting that it will do something. cc @aleene

Ban3: > but it's not useful if people add some other entries expecting that it will do something.
I agree with this. I almost did the same thing first time editing thinking the comments are parsed somehow. At least a comment where the regexes are actually defined would be good.

jolo0815: @Ban3 @stephanegigandet Okay good. Should I also remove the language codes?

jolo0815: Like this?

stephanegigandet (approved):
Looks good to me, thank you!

PR #4583, opened 2020-12-04
script to import all products revisions in mongodb

stephanegigandet:
This script is to import all revisions of all products (and not only the last revision) in a MongoDB collection.
The _id of each revision is "[code].[rev]".
This is for a student project where students will look at the history of changes (by users, by products etc.).
It does not use the ProductOpener modules so that it can be used more easily, just with a dump of all the .sto files for the products revisions.

### DIFF
--- a/scripts/update_all_products_revisions_from_dir_in_mongodb.pl
+++ b/scripts/update_all_products_revisions_from_dir_in_mongodb.pl
@@ -1,0 +1,146 @@
+#!/usr/bin/perl -w
+
+# This file is part of Product Opener.
+#
+# Product Opener
+# Copyright (C) 2011-2020 Association Open Food Facts
+# Contact: contact@openfoodfacts.org
+# Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
+#
+# Product Opener is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as
+# published by the Free Software Foundation, either version 3 of the
+# License, or (at your option) any later version.
+#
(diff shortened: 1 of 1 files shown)


hangy:
Do you intend to run this in production? It could easily cause the products collection to contain several hundred million documents, which could cause the software to behave badly. Every version would be indexed after all.

teolemon:
@hangy It's for a school project. That won't be run in prod and we'll get feedback from the student project before we attempt at deploying that ourselves.

teolemon:
Also, it will likely be an adhoc collection they generate. That wouldn't be the same collection.

stephanegigandet:
@hangy : It will be done on another server. It will be interesting to see how well MongoDB can handle it, and if it could be a solution to store revisions instead of using the file system. e.g. we could keep a product collection with only the latest version as we have today, and then have a products x revisions collection, that would be queried differently (e.g. in many cases with a specific product barcode), and that could have different indexes more tailored to those uses.

hangy:
Thanks for your clarifications. 🙂

hangy (approved):

cyrilbs:
I faced a problem with the script when importing sto files with 4 digits folders so I did this quick fix that let me inserted most of the data in mongoDb.
OLD:
        if ($code =~ /^(...)(...)(....?)(.*)$/) {
                $path = "$1/$2/$3/$4";
NEW:
        if ($code =~ /^(...)(...)(....?)(.*)$/) {
                $path = "$1/$2/$3/$4";

Not perfect as I have still a few data still rejected. I have not investigated further yet and not commited the change.
Please confirm if I should

PR #4584, opened 2020-12-05
Correct invalid pot file

teolemon:
Correct invalid pot file. Such mistakes would not happen every time if the files were compiled with poedit or checked with translatetoolkit.

msgctxt "list_of_sources"
msgid "Some of the data for this product has been collected in collaboration with %s"	msgid "Some of the data and/or photos for this product come from those sources:"
msgstr "Some of the data for this product has been collected in collaboration with %s"	msgstr "Some of the data and/or photos for this product come from those sources:"


stephanegigandet (approved):
Thank you.

PR #4601, opened 2020-12-08
better ecoscore details display

stephanegigandet:
Added a little bit of structure to the details of the computation.

There will probably be more to come, e.g. some icons, probably some details about the LCA impacts of ingredients, processing, transportation etc.

image

### DIFF
--- a/templates/ecoscore_details.tt.html
+++ b/templates/ecoscore_details.tt.html
@@ -5,8 +5,11 @@
 
 	<h3>Score de référence : Analyse de cycle de vie (ACV)</h3>
 
+	<div class="row">
+	<div class="small-12 columns">
+		<div class="panel" style="background-color:#dfd">
 	<p>
-		<strong>Catégorie Agribalyse : [% agribalyse.agribalyse_food_name_fr %]</strong><br>
+		Catégorie Agribalyse : [% agribalyse.agribalyse_food_name_fr %] 
 		[% IF agribalyse.agribalyse_proxy_food_code %]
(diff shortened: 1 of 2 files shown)


templates/ecoscore_details.tt.html
```diff
@@ -43,19 +56,34 @@ <h4>Origine des ingrédients</h4>
 		<strong>Bonus / malus - Politique environnementale : [% round(adjustments.origins_of_ingredients.epi_value) %]</strong><br>
```
CharlesNepote, line 56: "Bonus / malus - Politique environnementale : -5" is hard to read. I had to understand that the first "-" is not a "minus" but just a "connection" sign. I would write a ">" instead or event nothing:
=> "Bonus / malus > Politique environnementale : -5"

templates/ecoscore_details.tt.html
```diff
@@ -43,19 +56,34 @@ <h4>Origine des ingrédients</h4>
...
 		<strong>Bonus / malus - Transport : [% round(adjustments.origins_of_ingredients.transportation_value) %]</strong><br>
```
CharlesNepote, line 57: Same comments than "Bonus / malus - Politique environnementale".

teolemon (approved):

stephanegigandet:
@CharlesNepote : good point, I'm removing the words Bonus / Malus, it's shorter and clearer that way:

image

teolemon (approved):
I think the footprint icon is going to create confusion among users. I would put something related to deforestation instead.

PR #4602, opened 2020-12-08
Forest footprint improved display + attribute for personal search

stephanegigandet:
Forest footprint icons, description strings, and attribute.


teolemon (approved):
I think the footprint icon is going to create confusion among users about the scope about forest footprint (it's just deforestation and not carbon footprint). I would put something related to just deforestation instead (like a tree).

PR #4607, opened 2020-12-09
Add 1 Spanish synonym

mrmd8:
Adding one widely used Spanish synonym for High-fructose corn syrup


stephanegigandet (approved):
Thank you!

PR #4611, opened 2020-12-10
Added the Israeli Mandatory labels

yarons:
As per #4481

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -16145,6 +16145,22 @@ country:en:United States
 en:Real California Cheese
 country:en:United States
 
+en:Green Label
+he:הסימון הירוק
+country:en:Israel
+
+en:High level of fat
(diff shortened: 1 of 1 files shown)


stephanegigandet (changes requested):
For labels that are specific to a country, if the name is not distinctive (e.g. if the name could be used in other countries as well), then we should probably add the name of the country, so that we don't mix different labels together.

e.g. maybe something like:

en:Green Label - Israel
he:הסימון הירוק
countries:en: en:israel

en:High level of fat - Israel
he:שומן בכמות גבוהה
countries:en: en:israel

yarons:
Done, anything else I should change?

PS We should really switch to Wikidata :)

teolemon:
Ok for me. @stephanegigandet needs to approve so that we can merge.

stephanegigandet (approved):

PR #4624, opened 2020-12-12
mount logs, update Config.pm, fixes #4591

### LINKED ISSUE #4591: Docker environment not starting, related to ecoscore
Reported by @leventgunay : 

Log::Any::Proxy=HASH(0x563476144e58)[Sun Dec 06 10:59:36.205749 2020] [perl:error] [pid 24:tid 140342709499008] 30 unrecognized materials in CSV /mnt/podata/ecoscore/data/Eco_score_Calculateur.csv.11 at /opt/product-opener/lib/ProductOpener/Ecoscore.pm line 285, <$io> line 31.\nCompilation failed in require at (eval 2) line 1.\n
[Sun Dec 06 10:59:36.205917 2020] [perl:error] [pid 24:tid 140342709499008] Can't load Perl file: /opt/product-opener/lib/startup_apache2.pl for server productopener.localhost:0, exiting...

12 h 02
could anyone help me addressing this issue? i try to run start_dev but failing while processing eco score data.
12 h 02
[3688] /opt/product-opener/lib/ProductOpener/Ecoscore.pm 264 ProductOpener.Ecoscore {material => 'PS',material_id => 'fr:PS'} ecoscore material does not exist in taxonomy

stephanegigandet:
- mount the logs for easier debugging as suggested by @leventgunay 
- update Config.pm so that it loads the new taxonomies (and also add support for many other things like product attributes)

 fixes #4591

### DIFF
--- a/docker/backend-dev/conf/Config.pm
+++ b/docker/backend-dev/conf/Config.pm
@@ -252,6 +252,7 @@ $page_size = 20;
 
 $google_analytics = '';
 
+
 my @icons = (
 	{ "platform" => "ios", "sizes" => "57x57", "src" => "https://static.$server_domain/images/favicon/apple-touch-icon-57x57.png" },
 	{ "platform" => "ios", "sizes" => "60x60", "src" => "https://static.$server_domain/images/favicon/apple-touch-icon-60x60.png" },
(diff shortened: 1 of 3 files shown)


docker/backend-dev/conf/Config.pm
```diff
@@ -553,17 +690,61 @@ $options{import_export_fields_groups} = [
...
+# allow moving products to other instances of Product Opener on the same server
+# e.g. OFF -> OBF
+
+$options{current_server} = "off";
+
+$options{other_servers} = {
+	obf =>
+	{
+		name => "Open Beauty Facts",
+		data_root => "/srv/obf",
```
hangy, line 716: These don't exist in the docker environment, though.

stephanegigandet: Well it doesn't really hurt. I'd like to keep the Config.pm file as close as possible to the Config_off.pm file. Ideally they should be the same. Maybe we can just move anything that needs to be different to Config2.pm (like the google analytics stuff)

Regarding the "admin" user, is there such a user?

hangy: > Well it doesn't really hurt.

Doesn't it cause an unusable functionality for admin users?

> I'd like to keep the Config.pm file as close as possible to the Config_off.pm file. Ideally they should be the same. Maybe we can just move anything that needs to be different to Config2.pm (like the google analytics stuff)

How much of that is specific to _off, _obf_, etc.? 

> Regarding the "admin" user, is there such a user?

There are no users given by default, so you can register an admin with whatever password you want.

stephanegigandet: > How much of that is specific to _off, _obf_, etc.?

I think the docker environment should by default use Config_off.pm, which turns all the specific OFF feature on.

stephanegigandet: > Doesn't it cause an unusable functionality for admin users?

It's going to enable the feature that allows admin users to move a product to OBF etc. if they enter "obf" as the new barcode. And it will not work, but that's not going to cause an issue. On the contrary, it will force us to handle the case gracefully when the configuration specifies paths etc. that don't actually exist.

leventgunay (approved):
i have applied these changes on my local and it is finally up and running :) thx

stephanegigandet:
I'm merging this as it solves the issue #4591 . And we can improve how to use/divide Config.pm in another issue.

PR #4629, opened 2020-12-12
Plain fermented skimmed milk drink

aleene:
Is an accurate description for buttermilk for ecoscore


stephanegigandet (approved):

stephanegigandet (changes requested):
In fact here instead of putting "buttermilks" under the Agribalyse category, we should add the Agribalyse category as a match (or proxy match). That's because all "buttermilks" are not "plain": they can have fruit juice etc.

https://world.openfoodfacts.org/category/buttermilks/ingredients

PR #4632, opened 2020-12-14
Mostly German Ingredients

jolesh:
Started Translating and adding German Ingredients


stephanegigandet (approved):

taxonomies/ingredients.txt
```diff
@@ -31810,6 +31835,10 @@ fi:Luontainen hiiva
 fr:levure naturelle
 it:naturale-fermento, lievito naturale
 
+<en:yeast
+en:organic yeast
+de:Biologische Hefe
+
```
Ban3, line 31841: This should be instead picked up as a label by the ingredients parser. Add `Biologische` as a synonym here: https://github.com/openfoodfacts/openfoodfacts-server/blob/49dea456c786a8adac8b3a8eb371a0dc1413a132/taxonomies/labels.txt#L2902

taxonomies/ingredients.txt
```diff
@@ -44770,6 +44811,10 @@ fr:oignons frais
 it:cipolla fresca
 # ingredient/fresh-onions has 485 products in 3 languages @2018-12-06
 
+<en:onion
+en:Onion pieces
+de:Zwiebelstücke
+
```
Ban3, line 44817:
```suggestion
```
This is unneeded as `stücke` is properly handled by the processing taxonomy. See [test](https://world-de.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=Zwiebelst%C3%BCcke&type=add&action=process&.submit=1)

taxonomies/ingredients.txt
```diff
@@ -39610,14 +39648,17 @@ nl:geconcentreerd aardbeiensap
...
+fr:fraises lyophilisées
 fi:pakkaskuivattu mansikka, pakastekuivattu mansikka
 it:fragole liofilizzate
 # openfoodfacts:https://world.openfoodfacts.org/ingredient/fr:fraises-lyophilisées
 # 38 products in 3 languages @2108-10-01
 
-<fr:fraises lyophilisées
+<en:freeze-dried strawberries
+en:freeze-dried strawberry pieces
+de:gefriergetrocknete Erdbeerstücke, gefriergetrocknete Erdbeerscheiben
```
Ban3, line 39661: These are also handled by the processing taxonomy if you add `Erdbeer` as a synonym for strawberry in the ingredients taxonomy and `scheiben` to the processing taxonomy.

taxonomies/ingredients.txt
```diff
@@ -65842,6 +65897,10 @@ vegan:en:no
 vegetarian:en:yes
 # ingredient/honey has 9679 products in 17 languages @2018-12-17
 
+<en:honey
+en:bee  honey
```
Ban3, line 65901:
```suggestion
en:bee honey
```

taxonomies/ingredients.txt
jolesh: Okay thanks, I saw it that way at other points but it would be tedious  to make that for every ingredient.

taxonomies/ingredients.txt
jolesh: Okay, I think I saw it as not recognized on a product page but I am not sure.

taxonomies/ingredients.txt
jolesh: to "en:sliced"? "Scheiben" means "slices"

jolesh:
Any idea how managed to fail the test?

stephanegigandet (approved):
Looks good!

stephanegigandet:
@jolesh : the tests fail because the ouput of the analysis has changed, but it's probably good changes:

https://github.com/openfoodfacts/openfoodfacts-server/pull/4632/checks?check_run_id=1570168988

can you review the changes to see if they are correct?

jolesh:
> Looks good!

I only have to solve that about the "Scheiben" and find out why the PR test failed.

jolesh:
> @jolesh : the tests fail because the ouput of the analysis has changed, but it's probably good changes:
> 
> https://github.com/openfoodfacts/openfoodfacts-server/pull/4632/checks?check_run_id=1570168988
> 
> can you review the changes to see if they are correct?

Yeah, what is it exactly I have to look at?

taxonomies/ingredients.txt
Ban3: en:sliced sounds good

stephanegigandet:
e.g.

````````
#   Failed test at t/ingredients_processing.t line 1298.
#          $got->[0]{processing} = 'en:peeled'
#     $expected->[0]{processing} = 'de:gesch�lt'
[

**Comment by stephanegigandet** [comment]
You can search for "fail" in the test log, it will show the results of tests that have different output than expected.

**Comment by Ban3** [comment]
Searching for `failed` in that log: [First failure](https://github.com/openfoodfacts/openfoodfacts-server/pull/4632/checks?check_run_id=1570168988#step:6:2710)

> Failed test at t/ingredients_processing.t line 1298.

The actual test based on the lines above: https://github.com/openfoodfacts/openfoodfacts-server/blob/e744eaa74e23215355bb9302c1986b18e2a6ec90/t/ingredients_processing.t#L495-L525

Start of the difference to the expected results:
>      Structures begin differing at:
>           $got->[0]{processing} = 'en:peeled'
>      $expected->[0]{processing} = 'de:gesch�lt'

You added geschält under `en:peeled` but it already had its own German definition: https://github.com/openfoodfacts/openfoodfacts-server/blob/dd2929829b392b5befed9919333fc6c2357115c2/taxonomies/ingredients_processing.txt#L303-L306

I think you should remove it under the `en:peeled` and `en:unpeeled`

**Comment by stephanegigandet** [comment]
@jolesh : I made 2 small changes to make the tests pass and to fix the conflict with the Spanish changes.

We'll have to find a better solution for this entry:

#de:comment:geschält is a generic term and can mean en:peeled or en:shelled
de:geschält, geschälte, geschälter

Ideally, the best thing would be that all entries have translations in all languages (in particular in English). Those translations have to be unique.

The issue is that if geschält means both shelled and peeled, then we can't put it as a translation of each of them.

One solution could be:

en:peeled or shelled
de:geschält, geschälte, geschälter

en:peeled
de:(something that means peeled in German, but that does not exactly match de:geschält)

en:shelled
de:(something that means shelled in German, but that does not exactly match de:geschält)

cc @aleene @hangy

**Comment by jolesh** [comment]
Oh, sorry I forgot that the PR wasn't done yet and my Github didn't send me any mails.

File: taxonomies/ingredients_processing.txt [hunk 5]
```diff
@@ -361,7 +361,7 @@ en:zest
 
 # en:description:The soft center of a fruit
 en:pulp
-de:fruchtfleisch
+de:mark, pulpe, fruchtfleisch
```
jolesh, line 364: "Passierte Fruchtmassen werden ebenfalls als Mark bezeichnet, zum Beispiel [Tomatenmark](https://de.wikipedia.org/wiki/Tomatenmark) oder Himbeermark. Fruchtmark kann Säften oder Nektaren und vielen weiteren Nahrungsmitteln zugesetzt werden." [Wikipedia](https://de.wikipedia.org/wiki/Mark_(Lebensmittel))
So "mark" is also used for other things and I even think it is used more often in these other ways.
According to that evidence do we have to change sth?

PR #4638, opened 2020-12-15
Choice of sort order for products + interaction with Personal Search with user preferences

stephanegigandet:
This adds ways to sort products on the home page / facets pages by popularity, Nutri-Score, Eco-Score, last modification date and creation date.

I think I found a reasonable way to articulate the overall sort and the personal search sort:

1. We display 100 results per page instead of 20 or 24.
2. The default sort order needs to be decided (it could be the popularity / most scanned products, it is a reasonable choice for first time users (not contributors) and it makes sense when used in combination with user preferences)
3. Users will be able to change that sort order to see products with the best Nutri-Score, Eco-score, or recently added / edited products
4. Products will be color coded according to the user preferences, and the corresponding icons will be shown, but by default, they won't be sorted / filtered according to those preferences.
5. There is a switch that allows to rank the results on the page (so only those 100 results, not the whole database) using user preferences. That ranking / filtering is done locally in the browser. When the switch is on, we show tabs for products that match / may match / do not match the user preferences.
6. The pagination at the end allows to load more results (the next 100 products etc.)

It's currently live on .dev if you want to take a look:

https://uk.openfoodfacts.dev/

Here is how it looks like:

image


CharlesNepote:
Thank you very much Stéphane, it's very hard indeed. Here are my first thoughts about it.

I think it's very complicated for the beginner:
* there are so many buttons/choices in a very small space; there is even more buttons/choices in the whole page (eg. the "donate" and "sign in/out" buttons are not very far)
* there is at least 3 very different levels of usages:
  * all 572 products that can be refined based on some categories (brands, categories... but also popularity, Eco-Score)
  * all 572 (or 100?) products that can be sorted based on popularity, Eco-Score, etc;
  * 100 products that can be refined based on "food preferences"
  * at first, I could think there is another level with the colors which I don't understand very well: are green products good ones?
* why these 100 products? what is the sort key for the XXX products?
* how can I sort the 572 products by Eco-Score?
* I don't understand why my Food preferences are not used for the whole 572 products
* sort by nutri-score isn't working: https://uk.openfoodfacts.dev/?sort_by=nutriscore_score
* etc.

The page is very rich, so I'm afraid there is no magical solution, but I think we can try to make it clearer:
* group buttons/features dealing with the same things
  * the website level: donation banner seems to refer to contents displayed below it; at least "[ ] I have already donated ..." should be inside the donation box; the banner could be in the same grey area used for website level features on the left panel
  * the database level: 572 products, Drilldown..., Order..., Pages...
  * the page level: classify products..., Edit your food preferences
* be explicit where it's unclear:
  * "Order these 572 products"
  * "Classify these 100 displayed products according to your preferences"
  * "Edit your food preferences for (this page|the displayed products)"
  * Something like: "Food preferences are a private choice. **Open Food Facts respect your privacy by design**. Your food preferences are kept in your browser and never sent to Open Food Facts or anyone else. Your food preferences are applied by your browser, only for the displayed products."
* may be the "Pages: [1] 2 3 ..." could be repeated at the beginning of the list to explicitly show the context of these 100 products (a page of 100 products).

stephanegigandet:
Thanks for all the feedback and suggestions Charles!

> * the website level: donation banner seems to refer to contents displayed below it; at least "[ ] I have already donated ..." should be inside the donation box;

That was the easy one. :)

image

stephanegigandet:
Many small changes based on @CharlesNepote's feedback and suggestions:

image

1. The dropdown to select the ordering by "Most scanned products" etc. Is now right next to the number of total products.

2. The Explore by is moved below, and I added a small phrase to explain what it is. (a lot of people have no idea of what this button does). Also renamed "Drilldown" to "Explore".

3. Added pagination at the top, to make it more clear that the user preferences apply only to a single page of results. Also added a "(100 products per page)" at the end of the pagination to make more clear why we are talking about 100 products.

4. Add the number of products "Classify the 100 products below".

5. When the switch to usse user preferences to classify product is not selected, we do display the icons for the user preferences, but the color of all products is grey. When the switch is on, the tabs and the colors appear.

image

CharlesNepote:
Well, that's far better in my opinion.

I think we can make the database level (13,776 products...) even more straight and simpler. Here is a test:
-----------------------------------
image

* I deleted the text and added icons often used when dealing with sorting and filtering (for filter icon, none exists in UTF-8 but we probably [find many](https://www.google.com/search?q=filter+icon) in SVG).
* The two buttons should use the same style. All the line "13,776 products ......." should use the same style. We could also provide more feature at this level, such as a search field.

At the "page" level:
* "Edit your food preferences" might be styled differently from "database" level or "website" levels of information. It could be a small icon after "according to your preferences ⚙️" (with a bubble open by default). BTW, we should use different styles in each diferent levels of information.
* we might be able to select a different number of products per page (not sure).

stephanegigandet:
Thanks @CharlesNepote , I'll add the icons and display the first line with the number of results + sort + explore by as in your mockup.

Regarding the styles : sort and explore by are a bit different. The sort is a select box, it indicates the current sort and offers to change it. It changes the order of the result set, but not the actual results. The "explore by" is more of an action button that completely changes the view.

Regarding having different styles for the "edit your preferences" button, it is a possibility, but at this point we have the same style for all buttons across all features of the site, so it is probably something that we should review for the whole site before introducing new styles.

stephanegigandet:
Here is a new version, with icons as suggested by @CharlesNepote .

I changed the styles of the use preferences switch + the edit preferences button to round + green. That way it's more clear that they work together.

image

lang/en/texts/index.html
```diff
@@ -25,8 +25,3 @@ <h2>Contribute</h2>
 
 	</div>
 </div>
-
-<h2>Recently updated products:</h2>
```
teolemon, line 29: Do we still have a way to see recently completed products ?

teolemon (approved):

PR #4647, opened 2020-12-19
Synonyms for Aloysia citrodora

mrmd8:
New synonyms added:

en:Lemon verbena
es:Cedrón, verbena de Indias


stephanegigandet (approved):
Thank you!

PR #4649, opened 2020-12-20
fix export for Nutri-Score rename

teolemon:
fix export for Nutri-Score rename


stephanegigandet (approved):

teolemon:
it needs an additional fix to keep the same name in the export

teolemon:
fixed, merging.

PR #4657, opened 2020-12-22
Nordic unknown ingredients W52

Ban3:
**Description:**

Some unknown ingredients from the dk,is,fi and no domains.


taxonomies/allergens.txt
```diff
@@ -366,7 +372,7 @@ en:sulphur dioxide and sulphites, sulphur dioxide, sulphites, sulfites
 ar:ثاني أكسيد الكبريت والكبريتات, ثاني أكسيد الكبريت, الكبريتات
 bg:серен диоксид и сулфити, серен диоксид, сулфити
 cs:oxid siřičitý a siřičitany, oxid siřičitý, siřičitany
-da:svovldioxid og sulfitter, svovldioxid, sulfitter
+da:svovldioxid og sulfitter, svovldioxid, sulfitter, natriumsulfit
```
svensven, line 375: the other languages don't mention sodium sulphite (which is apparently E221). 
don't know if that's intentional/significant, but it stood out to me.
(also in sv:)

svensven (approved):

taxonomies/allergens.txt
Ban3: I don't know if it belongs either, but I think I added it because it appears highlighted on ingredient lists. [example](https://se.openfoodfacts.org/product/7310751194518/ica-potatismos)

Ban3: Now that I think about it, you don't need to list it for languages like French because it's in separate words, i.e **sulfite** de sodium. OFF currently breaks the ingredient if you try to mark part of a compound word as an allergen.

taxonomies/allergens.txt
```diff
@@ -13,13 +14,13 @@ en:gluten, cereals containing gluten, barley, barley malt flour, malted barley,
...
...rkkelystä, kauratärkkelyksestä, kauratärkkelykset, kauratärkkelyksiä, kauratärkkelyksistä, spelttitärkkelys, spelttitärkkelystä, spelttitärkkelyksestä, spelttitärkkelykset, spelttitärkkelyksiä, spelttitärkkelyksistä, vehnämallasjauho, vehnämallasjauhoa, vehnämallasjauhosta, vehnämallasjauhot, vehnämallasjauhoja, vehnämallasjauhoista, durumvehnämallasjauho, durumvehnämallasjauhoa, durumvehnämallasjauhosta, durumvehnämallasjauhot, durumvehnämallasjauhoja, durumvehnämallasjauhoista, ruismallasjauho, ruismallasjauhoa, ruismallasjauhosta, ruismallasjauhot, ruismallasjauhoja, ruismallasjauhoista, ohramallasjauho, ohramallasjauhoa, ohramallasjauhosta, ohramallasjauhot, ohramallasjauhoja, ohramallasjauhoista, kauramallasjauho, kauramallasjauhoa, kauramallasjauhosta, kauramallasjauhot, kauramallasjauhoja, kauramallasjauhoista, spelttimallasjauho, spelttimallasjauhoa, spelttimallasjauhosta, spelttimallasjauhot, spelttimallasjauhoja, spelttimallasjauhoista, vehnäkuitu, vehnäkuituja, vehnäkuiduista, vehnänjyvä, vehnänjyvää, vehnänjyvästä, vehnänjyvät, vehnänjyviä, vehnänjyvistä, kauranjyvä, kauranjyvää, kauranjyvästä, kauranjyvät, kauranjyviä, kauranjyvistä, muita viljoja, muita viljatuotteita
```
stephanegigandet, line 23: this list is really scary. at some point we'll need to find a better way to deal with all the possible forms on words in Finnish.

taxonomies/ingredients.txt
```diff
@@ -8,6 +8,8 @@ synonyms:en:hydrolysed, hydrolyzed
 
 synonyms:en:pasteurised, pasteurized
 
+synonyms:en:liquorice, licorice
```
stephanegigandet, line 11: there's already an ingredient en:liquorice, licorice, so we do not need to add it as a synonym here.

taxonomies/ingredients.txt
```diff
@@ -21184,8 +21335,17 @@ es:Saborizante de manzana natural
 fi:luontainen omena-aromi
 fr:arôme naturel de pomme
 hu:természetes alma aroma
+is:náttúruleg eplabragðefni
 nl:natuurlijk appelaroma
 
+<en:natural apple flavouring
+<en:with other natural flavouring
+en:natural apple flavouring with other natural flavourings
```
stephanegigandet, line 21343: we should remove this entry, and file a bug to improve the ingredients parser so that it can recognize it automatically. We do not want to create entries for all possible "natural [something]  flavour with other natural flavourings"

taxonomies/ingredients.txt
```diff
@@ -39633,6 +39912,76 @@ ro:pudră de căpșuni
...
+da:skovjordbær
+de:wald-erdbeere
+el:χαμοκέρασο
+eo:arbara fragujo
+es:fragaria vesca
+et:metsmaasikas
+eu:marrubi arrunt
+fa:توت‌فرنگی وحشی
+fi:ahomansikka, metsämansikka
+fr:fraisier des bois
```
stephanegigandet, line 39936: "fraisier des bois" is the name of the plant, and "fraise des bois" is the name of the fruit. If those were taken from wikidata / wikipedia, it's possible that some of the other languages list the plant and not the fruit.

stephanegigandet:
Thanks for all the changes! I just have a few comments above regarding a couple of entries.

taxonomies/allergens.txt
Ban3: It is. Couple of things that would improve it somewhat:

1. Using an allergen property in the ingredient taxonomy. That way the long list would be split into individual entries.
2. Detecting allergens inside compound words like `täysjyvä_vehnä_jauho`. Tricky to do automatically, but would allow only listing the allergens instead of every compound word where it appears.

taxonomies/ingredients.txt
Ban3: It's a Wikidata import indeed. I'll see if I can confirm some of them.

taxonomies/ingredients.txt
Ban3: Removed.

taxonomies/ingredients.txt
Ban3: I removed all the translations I couldn't find on OFF or otherwise wasn't quite sure about. There was actually [Wikidata item for the fruit](https://www.wikidata.org/wiki/Q104219130), but that wasn't much of a help.

taxonomies/ingredients.txt
Ban3: Removed and filed: #4666

stephanegigandet (approved):
Thank you!

PR #4660, opened 2020-12-23
Ingredient analyzer patterns sourced from Nordic ingredient lists

Ban3:
**Description:**

Mostly `ignore_regexps` so as not to remove anything useful following them.


Ban3:
Not included in the PR, but I noticed that I can't add `sisältää pieniä määriä` to trace regexps, or this test fails https://github.com/openfoodfacts/openfoodfacts-server/blob/8161351eb1cb194cf0fc53a104582397e0c984cc/t/allergens.t#L359-L362

`Pähkinöitä` won't be detected as an allergen with that addition. I have to take a look what's going on with this expression at some point https://github.com/openfoodfacts/openfoodfacts-server/blob/8161351eb1cb194cf0fc53a104582397e0c984cc/lib/ProductOpener/Ingredients.pm#L4991

stephanegigandet (approved):
Thank you!

PR #4683, opened 2020-12-29
New colors and icons for Eco-Score details + impacts for each LCA step

stephanegigandet:
Added Agribalyse detailed PEF + CO2 impacts for each LCA steps.

image


teolemon (approved):
Let it be noted in the record that I am not a fan of the color scheme.

teolemon:
image

stephanegigandet:
> Let it be noted in the record that I am not a fan of the color scheme.

Well we can keep everything green like today:

image

But I think it's useful to have different colors for each of the adjustments. I tried with pastel colors but it doesn't look very good either.

image

Ban3:
I actually prefer the pastel colors over the ones in the first screenshot.

stephanegigandet:
Let's go with pastel colors then. :)

image

Issue #4686, opened 2020-12-29
Ecoscore data: no way to know if "packaging" or "origins_of_ingredients" scores are provided or not


FredJul:
**Describe the bug**
While playing a bit with the ongoing ecoscore API during my PR (https://github.com/openfoodfacts/openfoodfacts-dart/pull/63), I noticed that there is no way for an app to know how to correctly display the "packaging" and "origins_of_ingredients" score if this one is not known.

**To Reproduce**
For instance, we can notice that values are not exact (it's hard to believe that 0 is a correct score for a plastic bottle) if we look at this product:
https://world.openfoodfacts.org/api/v0/product/5000112548167?fields=product_name,ecoscore_data,ecoscore_alpha

**Expected behavior**
I suggest to simply not return these fields when there is no information. Another solution would be to add a known/unknown status but I believe this is a bit more complex.

teolemon:
Related: 
https://github.com/openfoodfacts/openfoodfacts-server/issues/4663 (Origin display bug for the Eco-Score explanation)
https://github.com/openfoodfacts/openfoodfacts-androidapp/pull/3644 (adds prompt for missing info related to Eco-Score (solves issue 3624) )

stephanegigandet:
> (it's hard to believe that 0 is a correct score for a plastic bottle)

In fact the Eco-Score formula is quite complex, and for packaging material, the score indicates the "circularity" of the material. 0 is the worse score. (not a recycled material, not recycled). Other types of plastic have better scores.

stephanegigandet:
@FredJul : there are now more warning fields in ecoscore_data, that we use to display the details of the Eco-Score on the OFF site.
Can we close this bug?

FredJul:
I guess it clearly does the job, thanks!

PR #4690, opened 2020-12-31
Ecoscore - add Eco-Score warning when origins are 100% unknown

stephanegigandet:
- Add an "Unspecified" entry in the origins taxonomy. That way we can add "Unspecified" in the origins of ingredients field if the origins are not specified in the label. It will be different than the "Unknown" entry which means we don't have data.

- Add a warning in the ecoscore_data structure when 100% of the origins of ingredients are "Unknown" (the warning won't be triggered if the origin is "Unspecified" or if we have at least one origin for an ingredient).

- Add a corresponding data quality warning so that we can easily identify the products with 100% unknown origins.

This will partially solve #4686 .

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -1070,6 +1070,32 @@ sub check_ingredients_percent_analysis($) {
 	return;
 }
 
+
+=head2 check_ecoscore_data( PRODUCT_REF )
+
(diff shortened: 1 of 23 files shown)


teolemon (approved):
- [ ] A "en:origins-to-be-completed" would also be very crucial for the mobile app. 
- [ ] Passing those warnings and the score computation as HTML in environment_infocard_fr

PR #4697, opened 2021-01-02
Ecoscore fix for products without ingredients, use origin from origins field

stephanegigandet:
If the ingredients array does not exist or is empty, consider that 100% of the ingredients are coming from the origins in the origins field.


teolemon (approved):
Thank you

PR #4713, opened 2021-01-05
Prompts to add missing Ecoscore data

stephanegigandet:
with link to producers platform

image


teolemon (approved):
Approved. There's the subcase if which we don't compute an Eco-Score at all

PR #4719, opened 2021-01-06
Added some wikidata links

yarons:
Hoping we can pull some useful data and translations from there

### DIFF
--- a/taxonomies/packaging_materials.txt
+++ b/taxonomies/packaging_materials.txt
@@ -24,6 +24,7 @@ fr:Matériau recyclable
 hu:Újrahasznosítható anyag
 it:Materiale riciclabile
 pt:Material reciclável
+wikidata:en:Q57305800
 
 <en:Recyclable material
 <en:Plastic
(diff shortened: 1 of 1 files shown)


taxonomies/packaging_materials.txt
```diff
@@ -357,11 +372,13 @@ hu:Pamut, 60 COT
 it:Cotone
 nl:Katoen
 ru:Хлопок
+wikidata:en:Q8231603
```
CharlesNepote, line 375: Would https://www.wikidata.org/wiki/Q11457 better fit?

CharlesNepote: I reviewed all of them, good job. Just a little issue with "cotton", see my previous comment.

yarons: I'm not sure about Q11457, it's the raw material, the other one is a fabric made out of cotton which is more suitable for wrapping products (I guess).

stephanegigandet (approved):

PR #4723, opened 2021-01-09
Microbial culture changes

mrmd8:
* probiotic cultures entry added
* synonyms for bifidus culture and lactic cultures added

Thanks!


stephanegigandet (approved):
Thank you!

PR #4727, opened 2021-01-09
ecoscore: distinguish heavy and light aluminium

stephanegigandet:
distinction is based on shape

includes new tests


stephanegigandet:
should fix #4718

teolemon (approved):

PR #4736, opened 2021-01-11
Added German translations and some new ingredients

jolesh:
Added random German translation and ingredients I found to be missing

### DIFF
--- a/taxonomies/ingredients_processing.txt
+++ b/taxonomies/ingredients_processing.txt
@@ -875,12 +875,13 @@ pt:adoçado, adoçada
 sv:sötad
 
 en:unsweetened
+de:ungesüßt, ungesüßte, ungesüßter, ungesüßtes
 fr:non sucré, non sucrée
 nl:ongezoet
 
 #<en:sweetened
 en:sugared
-de:gezuckert, gezuckerte
+de:gezuckert, gezuckerte, gezuckerter, gezuckertes
 es:azucarado,azucarada,azucarados,azucaradas
 fi:sokeroitu, sokeroidut, sokeroitua, sokeroituja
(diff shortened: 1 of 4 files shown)


taxonomies/ingredients_processing.txt
```diff
@@ -908,6 +911,7 @@ en:semi candied, semi-candied
 # <en:unsweetened
 de:ungezuckert
 
+en:oiled
```
jolesh, line 914: The test probably failed because I added "en:oiled" but I don't know why that is a problem.

Ban3: The test expects German entry to be the main entry: https://github.com/openfoodfacts/openfoodfacts-server/blob/dcb9640ea1490429a38281a57c5d14dd202111fe/t/ingredients_processing.t#L425-L439

I'd ping @aleene in case there's a reason why it doesn't have English entry.

stephanegigandet: In English, I don't see any ingredients that are "oiled": https://us.openfoodfacts.org/ingredients?filter=oiled
So I guess it's fine to add it in English, there wouldn't be false positives.

What about the German ones with geölt, geölte, geölter, geöltes, what kind of ingredients are those?

stephanegigandet (approved):

taxonomies/ingredients_processing.txt
stephanegigandet: in another PR, we should add "boiled", "parboiled", "charbroiled", "flamebroiled" though :)

jolesh: I know that raisins are sometimes oiled but haven't found any ingredients containing "geölt".

PR #4740, opened 2021-01-12
group and optimize SVG icons for attributes, fixes #4739

### LINKED ISSUE #4739: Group and optimize SVG icons for attributes
SVG rendering seems to take a lot of time, especially on mobile (Flutter). See discussion in https://github.com/openfoodfacts/smooth-app/issues/52

- move all sources SVG files for attributes to html/images/attributes/src  (from images/misc or images/icons )
- add gulpfile config to optimize SVG files and put them to /html/images/attributes/
(I prefer to keep the source images close to the destination images, and not to lengthen the url with an extra /dist/)
- fix nuts and nutriscore-a icon

stephanegigandet:
SVG rendering seems to take a lot of time, especially on mobile (Flutter). See discussion in openfoodfacts/smooth-app#52

- move all sources SVG files for attributes to html/images/attributes/src (from images/misc or images/icons )
- add gulpfile config to optimize SVG files and put them to /html/images/attributes/
(I prefer to keep the source images close to the destination images, and not to lengthen the url with an extra /dist/)
fix nuts and nutriscore-a icon


teolemon (approved):
Are some symlinks planned for the old paths ?

gulpfile.js
```diff
@@ -42,6 +42,19 @@ function icons() {
     pipe(dest("./html/images/icons/dist"));
 }
 
+function attributes_icons() {
```
hangy, line 45: `_` somewhat breaks with the naming convention of all other methods in this file. Maybe use `attributeIcons` instead?

gulpfile.js
```diff
@@ -42,6 +42,19 @@ function icons() {
...
+  return src("*.svg", { cwd: "./html/images/attributes/src" }).
+    pipe(
+      svgmin({
+      plugins: [
+        { removeMetadata: true },
+        { removeTitle: true },
+      ]
+      })
+    ).
+    pipe(dest("./html/images/attributes"));
```
hangy, line 55: Are the generated files ignored by `.gitignore`?

gulpfile.js
stephanegigandet: sure, I'll change it.

stephanegigandet:
> Are some symlinks planned for the old paths ?

in production I will just leave the existing files where they are, at least for the time being

gulpfile.js
stephanegigandet: good point, I'll add that.

gulpfile.js
stephanegigandet: changed.

gulpfile.js
stephanegigandet: generated files added to .gitignore

PR #4742, opened 2021-01-12
ci: Label stale issues automatically

Ban3:
**Description:**

Label issues with no activity for 90 days as `stale`. Does not close or comment on issues.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,16 @@
+name: "Stale issue handler"
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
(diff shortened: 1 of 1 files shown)


M123-dev:
The suggestion was to automatically close issues that have been inactive for a long time,since they are then mostly irrelevant, so it is done in many big repos. In this repo the oldest open issue is from Feb 2014. After a small discussion on [slack ](https://openfoodfacts.slack.com/archives/C02KVRT2C/p1610475803177600)we came to the decision that it is better to label the issues only to make the closure / check of the relevants manually.

stephanegigandet (approved):

Issue #3961, opened 2020-07-30
[categories] Oatmeal cookies is not in Cookies


aleksejrs:

jolesh:
Thanks for telling us about the issue @aleksejrs . In the meantime it was already fixed by somebody else.

PR #4748, opened 2021-01-13
added Oatmeal cookies to Drop cookies

jolesh:
Fix for issue #3961 "[categories] Oatmeal cookies is not in Cookies"

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -23295,8 +23295,9 @@ ciqual_food_name:fr:Biscuit sec ou tuile, aux amandes
 fr:tuile aux amandes
 
 <en:Biscuits
+<en:Drop cookies
 en:Oatmeal cookies
-de:Hafer-Cookies
+de:Hafer-Cookies, Haferkekse
 es:Galletas de avena
 fi:Kaurakeksit
 fr:Biscuits à l'avoine


taxonomies/categories.txt
```diff
@@ -23295,8 +23295,9 @@ ciqual_food_name:fr:Biscuit sec ou tuile, aux amandes
 fr:tuile aux amandes
 
 <en:Biscuits
+<en:Drop cookies
```
stephanegigandet, line 23298: in fact we should not add <en:Drop cookies as it's a special type of cookies.

the "en:biscuits" entry already has cookies as a synonym:

en:Biscuits, biscuit, cookies, cookie

jolesh: So we can't fix issue #3961 /it is already fixed?

stephanegigandet: I think it's fixed in fact:

https://world.openfoodfacts.org/category/oatmeal-cookies

-> it is in "biscuits", which is the main name for cookies in English.

We made a distinction between "cookies" and "drop cookies".  In France, "cookies" are only "drop cookies". But in the US / UK, they can be any type of biscuits.

stephanegigandet: But we should merge your other changes.

jolesh: Okay, good

Issue #3892, opened 2020-07-19
"en:Fruit sodas" category is incorrectly translated as "ru:Напитки на основе фруктов"


aleksejrs:
**Describe the bug**
"en:Fruit sodas" is incorrectly translated as "ru:Напитки на основе фруктов", which puts Russian fruit-based beverages into "Sodas".

**To Reproduce**
1. Go to https://ru.openfoodfacts.org/%D0%BA%D0%B0%D1%82%D0%B5%D0%B3%D0%BE%D1%80%D0%B8%D1%8F/%D0%BD%D0%B0%D0%BF%D0%B8%D1%82%D0%BA%D0%B8-%D0%BD%D0%B0-%D0%BE%D1%81%D0%BD%D0%BE%D0%B2%D0%B5-%D1%84%D1%80%D1%83%D0%BA%D1%82%D0%BE%D0%B2
2. Look at the parent categories.
3. See "en:Sodas".

**Additional context**

```
929c7530f9 (root                     2019-03-26 07:45:55 +0000  3061) <en:Sodas
929c7530f9 (root                     2019-03-26 07:45:55 +0000  3062) <en:Fruit-based beverages
929c7530f9 (root                     2019-03-26 07:45:55 +0000  3063) en:Fruit sodas
e90157b744 (root                     2019-03-26 08:04:27 +0000  3064) de:Fruchtlimonaden
929c7530f9 (root                     2019-03-26 07:45:55 +0000  3065) es:Sodas a base de frutas
2774a4e4dc (Stéphane Gigandet        2019-09-02 12:14:03 +0200  3066) fi:Hedelmävirvoitusjuomat
929c7530f9 (root                     2019-03-26 07:45:55 +0000  3067) fr:Sodas aux fruits
2774a4e4dc (Stéphane Gigandet        2019-09-02 12:14:03 +0200  3068) hu:Szénsavas gyümölcs ital, gyümölcs fröccs
929c7530f9 (root                     2019-03-26 07:45:55 +0000  3069) nl:Vruchtenfrisdranken
929c7530f9 (root                     2019-03-26 07:45:55 +0000  3070) ro:Băuturi carbogazoase cu fructe, Bauturi carbog
azoase cu fructe, Băuturi carbogazoase din fructe, Bauturi carbogazoase din fructe
929c7530f9 (root                     2019-03-26 07:45:55 +0000  3071) ru:Напитки на основе фруктов
[...]

PR #4749, opened 2021-01-13
removed incorrect Russian translation in categories.txt

jolesh:
see issue "en:Fruit sodas" category is incorrectly translated as "ru:Напитки на основе фруктов" #3892


stephanegigandet (approved):
Thank you!

Issue #2220, opened 2019-08-24
Add "Confirmed by manufacturer" labels for info not on the product label but sent by producers by email (e.g. vegan, vegetarian, palm oil free etc.)


stephanegigandet:
Sometimes ingredients are ambiguous (e.g. could be vegan or not), and some contributors ask producers for more information. They often get back emails with more precise information.

When we get such information, we should:
1. Upload an image of the correspondence with the producer to the corresponding products
2. Add the extra info to the product.

Proposal:

- Create labels like "Confirmed palm oil free by manufacturer" (do not make them children "of "Palm oil free", so that we can differentiate what is actually on the product label.
- When doing ingredient analysis, use both "Palm oil free" and "Confirmed palm oil free by manufacturer" to override the result of ingredient analysis.

jolesh:
Vegan, Vegetarian, Palm oil free, gluten free, what else should I add?

stephanegigandet:
> Vegan, Vegetarian, Palm oil free, gluten free, what else should I add?

@jolesh : A good question. I'm having second thoughts about this, as we would essentially need to double almost all labels regarding allergens, ingredients etc. then translate them etc. Maybe the simplest thing is just to use the existing labels.

stephanegigandet:
In practice, since this issue was opened, I don't recall any label info sent by producers by email, so I'm thinking we could just close this, and if it does happen, we just use the normal "No gluten" etc. labels.

jolesh:
Okay seems good to me. The person adding the label could/should then describe the source in the description of the edit.

PR #4756, opened 2021-01-14
add "Confirmed by manufacturer" labels

jolesh:
as stated in issue #2220


jolesh:
We decided against it because we would have to do that for almost every label

PR #4764, opened 2021-01-18
add org-panzani-sa, org-cristalco to nightly equadis imports

stephanegigandet:
nightly updates from equadis


teolemon (approved):
We'll need to ask them for the reminder of their product. Tracking spreadsheet updated accordingly.

PR #4765, opened 2021-01-18
update_all_products.pl script - allow OR queries for tags

stephanegigandet:
small improvement to allow OR queries


teolemon (approved):
Approving without a good understanding of the implications or the rationale, praying it doesn't start the apocalypse :-)

PR #4777, opened 2021-01-21
buttons to remove nutrients, fixes #4776 #1141

stephanegigandet:
fixes #4776 / fixes #1090 (for moderators)
fixes #1141 (for everyone)

image

PR #4781, opened 2021-01-22
make save button green in product edit form

stephanegigandet:
Suggestion from @CharlesNepote in https://github.com/openfoodfacts/openfoodfacts-server/pull/4777#issuecomment-764697968

"Small suggestion: all buttons are blue but this can lead to confusion. For example, I often click on "Save" instead of "Validate and/or resize image" because there are not far from each other (see below)... Maybe the "non-return" buttons should be different ("Save", "Erase something", "Delete product", etc.)."

image


VaiTon (approved):

hangy (approved):

CharlesNepote:
Great! Thanks!

PR #4784, opened 2021-01-22
Added Danish taxonomy

jolesh:
All taken from ingredient list of Danish products I added

### DIFF
--- a/taxonomies/additives.txt
+++ b/taxonomies/additives.txt
@@ -7276,7 +7276,7 @@ en:E322(i), Lecithin
 bg:E322(i), Лецитин
 ca:E322(i), lecitina
 cs:E322(i), lecitin
-da:E322(i), Lecitin
+da:E322(i), Lecitin, lecithin
 de:E322(i), Lecithin
 el:E322(i), λεκιθίνη
 eo:E322(i), Lecitino
@@ -12080,7 +12080,7 @@ en:E450, Diphosphates, Pyrophosphate, diphosphate
 bg:E450, Дифосфат
 ca:E450, Difosfat
 cs:E450, E450 food additive
-da:E450, E450 food additive
+da:E450, E450 food additive, diphosphater
 de:E450, Diphosphate
 el:E450, E450 food additive
 es:E450, Difosfatos, Difosfato, Pirofosfato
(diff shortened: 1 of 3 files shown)


taxonomies/additives.txt
```diff
@@ -12080,7 +12080,7 @@ en:E450, Diphosphates, Pyrophosphate, diphosphate
 bg:E450, Дифосфат
 ca:E450, Difosfat
 cs:E450, E450 food additive
-da:E450, E450 food additive
+da:E450, E450 food additive, diphosphater
```
Ban3, line 12083:
```suggestion
da:E450, diphosphater
```

jolesh: Tak

jolesh:
Does anybody understand why the test fails? I know that it is caused by:

"<en:egg
en:pasteurized egg, pasteurized eggs
da:pasteuriseret æg
de:pasteurisiertes Ei, pasteurisierte Eier"

but I don't know why

taxonomies/ingredients.txt
```diff
@@ -69076,6 +69087,11 @@ allergens:en: en:eggs
 carbon_footprint_fr_foodges_ingredient:fr:Oeuf, moyenne nationale
 carbon_footprint_fr_foodges_value:fr:2.6
 
+<en:egg
+en:pasteurized egg, pasteurized eggs
+da:pasteuriseret æg
+de:pasteurisiertes Ei, pasteurisierte Eier
+
```
Ban3, line 69094:
```suggestion
```
This is why the test fails. Pasteurized is in processing. It's not yet enabled for Danish (I'm working on finishing the tests for it)

jolesh: Okay great. I am often unsure if a specific combination of ingredient and processing should get their own entry in the ingredients.txt. Because there are a lot of these combinations but they are often not really necessary as far as I know.

stephanegigandet: Historically, we were not able to parse the processing, so we added entries like "pasteurized eggs". But now that we parse the processing, we should remove those combined entries. (the tricky part is to make sure we have working processing parsing for as many languages as possible)

jolesh: Strange couldn't commit through the browser. So I should definitely not add new one and I could delete old ones but it has no high priority since the don't cause any harm, right? @stephanegigandet

stephanegigandet: That's right. We need to be careful when we delete entries though, as they may be covered by the processing algorithm/taxonomies for some languages, but not all the ones for the combined entry. If possible we should then add processing (but then we have also to be very careful about false positives) for more languages.

stephanegigandet (approved):
Thank you!

PR #4828, opened 2021-02-04
Eco-Score updates : include milks in beverages, updated packaging and shapes

stephanegigandet:
Many small updates to the Eco-Score.


lib/ProductOpener/Packaging.pm
```diff
@@ -316,6 +316,12 @@ sub analyze_and_combine_packaging_data($) {
 		
 		my $packaging_ref = parse_packaging_from_text_phrase($phrase, $product_ref->{lc});
 		
+		# If the shape is "capsule" and the product is in category "en:coffees", mark the shape as a "coffee capsule"
+		if ((defined $packaging_ref->{"shape"}) and ($packaging_ref->{"shape"} eq "en:capsule")
+			and (has_tag($product_ref, "categories", "en:coffees"))) {
+			$packaging_ref->{"shape"} = "en:coffee-capsule";
+		}
+
```
teolemon, line 324: We have specific categories for Tassimo/Nespresso systems

teolemon (approved):

lib/ProductOpener/Packaging.pm
stephanegigandet: I made the choice to use the broader "coffees" category to catch all the products that do not have the finer categories.
e.g. see https://fr.openfoodfacts.org/search?brands_tags=nespresso&categories_tags_fr=-cafes-en-dosettes-compatible-nespresso,cafes

It's also more robust as it's less likely that the canonical name of "en:coffees" could change.

PR #4838, opened 2021-02-08
small improvements to handle more variations of producer data

stephanegigandet:
e.g. "Nutriscore_A"
also a fix to convert "NESTLE, RICORE" to "Nestle, Ricore" and not "Nestle, ricore"


CharlesNepote (approved):
Sounds good.

PR #4841, opened 2021-02-09
small fixes for producers platform

stephanegigandet:
- does not fail if the producers give an image url that returns something else than an image
- ignore nutrients if they contain something that does not look like nutrients


CharlesNepote (approved):
Sounds good (not tested).

PR #4855, opened 2021-02-12
New GS1 conversion pipeline for Equadis and CodeOnline imports

stephanegigandet:
This is a reimplementation of the GS1 pipeline created by @odtvince for the Equadis import, so that it can also handle CodeOnline, and possibly other sources of GDSN data.

Main changes:
- as CodeOnline is a JSON export of the same data structures used in the Equadis XML, instead of converting the xml to CSV, we now convert the xml to JSON, using the node module xml2json. It removes the dependency on the modified xml2csv.
The JSON from CodeOnline and the one from xml2json is not exactly the same, but it's close. The differences are only about how to represent the source XML. The source XML is the same.
- the resultings JSON are parsed in Perl, to convert to the OFF format
- added some fields that are specific to cosmetics (e.g. cosmetics ingredients list)
- added tests so that we can more easily fix things / add support for new fields etc. without breaking other things


teolemon:
```
t/i18n.t ......................... ok
en 12345678.jpg
en 12345678_photo.jpg
en 12345678_photo-3510.jpg
en 12345678_2.jpg
en 20200201131743_2.jpg
en 4 LR GROS LOUE_3 251 320 080 419_3D avant.png
t/images.t ....................... ok
# {
#   'drained_weight' => '336 g',
#   'lc' => 'es',
#   'net_weight' => '480 g',
#   'quantity' => '480 g',
#   'total_weight' => 'Peso neto: 480 g (6 x 80 g) Peso neto escurrido: 336 g (6x56 g)'
# }
t/import.t ....................... ok
# error: gs1_maps - entry not in taxonomy {tag => "Oats",tagtype => "allergens"}
Died at /home/runner/work/openfoodfacts-server/openfoodfacts-server/lib/ProductOpener/GS1.pm line 279, <DATA> line 2104.
Compilation failed in require at t/import_gs1.t line 17, <DATA> line 2104.
BEGIN failed--compilation aborted at t/import_gs1.t line 17, <DATA> line 2104.
t/import_gs1.t ................... 
Dubious, test returned 255 (wstat 65280, 0xff00)
```

CharlesNepote (approved):

Issue #4662, opened 2020-12-23
Add the Eco-Score computation explanation to the HTML field we created for the v1 of Carbon impact


teolemon:
Add the Eco-Score computation explanation to the HTML field for the v1 of Carbon impact so that we can have the detailed computations in the Environment tab at launch for Android and iOS

See original PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/1656/files

```
<h2>♺ Analyse de cycle de vie (ACV)</h2>
🌎Catégorie Agribalyse : Tomate, séchée<br>
Score environnemental PEF : 1.96 (plus le score est bas, plus l'impact est faible)<br>
➡️dont impact sur le changement climatique : 6.47 kg CO2 eq/kg de produit<br>
<h3>Détail des impacts par étapes du cycle de vie</h3>
🚜Agriculture: 34.1 % <br>
🏭Transformation : 59.8 % <br>
🥫Emballage : 3.6 % <br>
🚚Transport : 1,3 % <br>
🛒Distribution : 1,0 % <br>
🍳Consommation : 0,2 % <br><br>
Score ACV sur 100 : 22

<h2>Bonus et malus complémentaires</h2>
<h3>Mode de production</h3>
Pas de labels pris en compte pour le système de production.

<h3>Origine des ingrédients</h3>
Italie: 100%<br>
Politique environnementale : +2 points<br>
Transport : +8 points<br>

<h3>🌴- Espèces menacées</h3>
Aucun ingrédient dont la culture menace des espèces n'a été détecté.
<h3>♺ Recyclabilité de l'emballage</h3>
Emballage<br>
Boîte de conserve (ratio : 1) - Verre (score : 80.5)<br>
Score de tous les composants : 80.5<br><br>

Emballage : -1.95<br>
<h2>Score final</h2>
Score Éco-score : 30 - Note Éco-score : D<br><br>

➽ Si vous êtes le fabricant de ce produit, vous pouvez nous transmettre les informations avec notre plateforme gratuite pour les producteurs (https://fr.pro.openfoodfacts.org ou producteurs@openfoodfacts.org)<br><br>

➽ L'Eco-Score ne peut être calculé que si le produit possède une catégorie suffisamment précise. Vous pouvez modifier la fiche produit pour ajouter une catégorie plus précise.<br><br>

```

![image](https://user-images.githubusercontent.com/1689815/103008893-1a7ecc00-4536-11eb-923f-cafcdd3b311c.png)

[...]

PR #4882, opened 2021-02-18
remove old ecological table

### LINKED ISSUE #4875: Remove old "ecological impact" table below the nutrition table
This table was created with ADEME's FoodGES database, before we computed the Eco-Score.

I will remove it as it is confusing when shown with the Eco-Score.

image

stephanegigandet:
as we now have the Eco-Score. Otherwise it is quite confusing.

fixes #4875

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -9339,8 +9339,6 @@ sub display_nutrition_table($$) {
 				rows => [],
 			},
 		],
-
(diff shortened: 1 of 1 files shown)


teolemon (approved):
Approving, but 

- it feels like a downgrade for the rest of the world. 
- Plus we should implement https://github.com/openfoodfacts/openfoodfacts-server/issues/4662

PR #4916, opened 2021-03-01
Add some Russian additive classes

blazern:
Adding some Russian additive classes (in singular and plural forms).


stephanegigandet (approved):
Thanks!

PR #4919, opened 2021-03-02
Adding some Russian ingredients and additives classes

blazern:
Adding some Russian ingredients and additives classes


stephanegigandet (approved):
Thanks!

PR #4922, opened 2021-03-03
Add some Russian ingredients, additives and additives classes

blazern:
Adding some Russian ingredients, additives and additives classes


stephanegigandet (approved):
Looks good, thanks!

PR #4923, opened 2021-03-03
Add some Russian ingredients, additives and additives classes

blazern:
**Please verify that I don't corrupt things by this PR** - I'm not really sure about half of the changes in this PR (I'm sure about translations, but not sure if I've put the phrases into proper places).

### DIFF
--- a/taxonomies/additives.txt
+++ b/taxonomies/additives.txt
@@ -3588,7 +3588,7 @@ nl:E200, Sorbinezuur, Sorbinezuur
 nn:E200, sorbinsyre
 pl:E200, Kwas sorbowy, Kwas sorbowy, Kwas sorbinowy
 pt:E200, Ácido sórbico, Ácido sórbico, Sorbato
-ru:E200, Acid sorbic, сорбиновая кислота, консервант сорбиновая кислота
+ru:E200, Acid sorbic, sorbic acid, сорбиновая кислота, консервант сорбиновая кислота
 ro:E200, Acid sorbic, Acid sorbic
 sh:E200, Sorbinska kiselina
 sk:E200, Kyselina sorbová, Kyselina sorbová
@@ -3863,7 +3863,7 @@ nl:E211, Natriumbenzoaat
 pl:E211, Benzoesan sodu
 pt:E211, Benzoato de sódio
 ro:E211, Benzoat de sodiu
-ru:E211, консервант бензоат натрия, бензоат натрия
+ru:E211, Бензоат натрия, консервант бензоат натрия
 sh:E211, Natrijum-benzoat
 sk:E211, Benzoan sodný
 sl:E211, Natrijev benzoat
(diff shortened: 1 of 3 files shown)


taxonomies/additives.txt
```diff
@@ -3587,6 +3587,7 @@ nl:E200, Sorbinezuur, Sorbinezuur
 nn:E200, sorbinsyre
 pl:E200, Kwas sorbowy, Kwas sorbowy, Kwas sorbinowy
 pt:E200, Ácido sórbico, Ácido sórbico, Sorbato
+ru:E200, Acid sorbic, sorbic acid, сорбиновая кислота, консервант сорбиновая кислота, Е200, Е-200
```
stephanegigandet, line 3590: we don't need the E200 at the end, as it's already at the beginning (there are other existing entries that have the E-number repeated, but that's a mistake).

we don't need E-200 either, as for additives, variations like E-200 are automatically generated

taxonomies/ingredients.txt
```diff
@@ -4837,6 +4838,13 @@ vegetarian:en:maybe
 # comment:en:All vegan sites state that industrial produced lactic ferments is vegan. Unclear if exceptions do occur. Still need an authoritative source for this.
 # comment:en:A wide variety of synonyms are used, but they all seem to point to the same bacteria.
 
+<en:fermentation starter
+en:fermentation starter
```
stephanegigandet, line 4842: <en:fermentation starter
en:fermentation starter

Those 2 lines mean that there is an "en:fermentation started" which is a child of "en:fermentation starter", so it does not work.

what does "закваска" mean ? does it mean "fermentation starter" ? or a more specific kind of fermentation starter?

taxonomies/ingredients.txt
```diff
@@ -16907,6 +16914,11 @@ hu:pálmamagolaj és -zsír
 nl:palmpitolie en vet
 from_palm_oil:en:yes
 
+<milk fat substitute
+en:milk fat substitute
+ru:заменитель молочного жира
```
stephanegigandet, line 16919: same here, if заменитель молочного жира means "milk fat substitute", then it should be added as a translation of the existing "en:milk fat substitute" entry, not as a child

taxonomies/ingredients.txt
blazern: > what does "закваска" mean ? does it mean "fermentation starter" ? or a more specific kind of fermentation starter?

It just means "fermentation starter", its origin is not specified. It could be anything from yeast to abomasum.
https://ru.wikipedia.org/wiki/%D0%97%D0%B0%D0%BA%D0%B2%D0%B0%D1%81%D0%BA%D0%B0

> Those 2 lines mean that there is an "en:fermentation started" which is a child of "en:fermentation starter", so it does not work.

How should I fix it?
Should I just remove the first line?

taxonomies/ingredients.txt
blazern: I don't understand.
There's no other `en:milk fat substitute`, only this one.
Do you mean that I should remove the `<milk fat substitute` line?

stephanegigandet: ok, so if en:milk fat substitute does not exist yet, you can create it:

en:milk fat substitute
ru:заменитель молочного жира

without <milk fat substitute  before

taxonomies/ingredients.txt
stephanegigandet: There's an "en:ferment" entry:

#
#                              F E R M E N T S     S E C T I O N
#

# description:en:FERMENT is an agent, such as an enzyme, bacterium, or fungus, that brings about fermentation.
# comment:en:Ferment is thus more generic than enzyme.

en:ferment


I think "fermentation starter" should just be a synonym of "ferment".

So you can add "fermentation starter" as a synonym in the existing entry

en:ferment, fermentation starter

and add a Russian translation

blazern:
Fixed issues discussed above

stephanegigandet (approved):
Thanks!

PR #4929, opened 2021-03-04
Add more of Russian ingredients, additives and additives classes

blazern:
Adding more of Russian ingredients, additives and additives classes.

In the `additives.txt` file the `E-` additives are sometimes duplicated by this PR, but that's intentional.
For example, in next line:
```
ru:E322, Лецитины, е322
```
...the first `E322` starts with a latin `E`, and the second `е322` starts with a cyrillic `e`.
They have different unit codes in all encodings I worked with.


stephanegigandet (approved):

blazern:
The PR is has an approve but is not merged for some reason
@Steinhagen, is that intentional?

stephanegigandet:
@blazern : I wanted to keep it open while I opened the https://github.com/openfoodfacts/openfoodfacts-server/issues/4931 bug and then forgot to merge it.

PR #4933, opened 2021-03-05
Make exports more flexible, fixes #4932

### LINKED ISSUE #4932: Make CSV exports more flexible
We have an export_csv_file.pl script and export functions (also used to export the producers platform data to the public database), but it can be made more flexible:

- support advanced query filters (similar to the search V2 API : products without a tag, with multiple tags, with one of many tags etc.)
- exporting data for a specific set of barcodes
- exporting fields that are not at the root of the product structure

e.g.

./export_csv_file.pl --query 'additives_tags=en:e101|en:e102' --query ecoscore_grade=exists --fields=code,ecoscore_grade,ecoscore_data.score,ecoscore_data.agribalyse.agribalyse_food_code --query-codes-from-file codes

stephanegigandet:
fixes #4932 

- make all fields available for filtering in the Search API V2
- use the same filter parsing code than the Search API V2 for the export_csv_file.pl script
- support advanced query filters (similar to the search V2 API : products without a tag, with multiple tags, with one of many tags etc.)
- exporting data for a specific set of barcodes (from a file)
- exporting fields that are not at the root of the product structure

e.g. it's now possible to do things like this:

./export_csv_file.pl --query 'additives_tags=en:e101|en:e102' --query ecoscore_grade=exists --fields=code,ecoscore_grade,ecoscore_data.score,ecoscore_data.agribalyse.agribalyse_food_code --query-codes-from-file codes

The corresponding changes will also make it easier to do things like this through a web interface at a future point.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -4229,6 +4229,32 @@ Reference to the MongoDB query object.
 
 =cut
 
+# Parameters that are not query filters
+
+my %ignore_params = (
+	fields => 1,
+	format => 1,
+	json => 1,
+	jsonp => 1,
+	xml => 1,
+	keywords => 1,	# added by CGI.pm
+	api_version => 1,
+	api_method => 1,
+	search_simple => 1,
+	search_terms => 1,
+	userid => 1,
+	password => 1,
+	action => 1,
+	type => 1,
+);
+
+# Parameters that can be query filters
+# It is safer to use a positive list, instead of just the %ignore_params list
+
+my %valid_params = (
+	code => 1,
+);
+
 sub add_params_to_query($$) {
 	
 	my $request_ref = shift;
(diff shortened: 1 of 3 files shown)


teolemon (commented):
It could pave the way to custom exports for the offline scan feature for the apps.
My understanding is that v2 enables:
- [ ] country filtering
- [ ] popularity filtering
- [ ] field by field filtering

CharlesNepote (approved):
Great PR. I documented export_csv.pl.
Doesn't the PR need more tests? And doc? I'm not able to understand modifications in Display.pm and Export.pm.

stephanegigandet:
> Great PR. I documented export_csv.pl.

Thanks!

> Doesn't the PR need more tests? And doc? I'm not able to understand modifications in Display.pm and Export.pm.

I added a comment in Export.pm :)

This part of the code does not have tests. Yes, it would be great to add some, at some point.

PR #4939, opened 2021-03-07
Cioccolato di Modico

aleene:
**Description:**
Maybe we should go through the EU database to add al PGI's?


teolemon:
@aleene There's categories.wip.txt. PGI integration is still work in progress. `Cioccolato di Modico` might be in there ?

stephanegigandet (approved):
Thanks

PR #4952, opened 2021-03-10
Add Russian ingredients, additives and additives classes

blazern:
Adding Russian ingredients, additives and additives classes.


stephanegigandet (approved):
Thank you!

PR #4955, opened 2021-03-10
better missing data warnings for Eco-Score (easy reviews)

stephanegigandet:
fixes a bug in the template and provides slightly more specific messages about what's missing (e.g. no packaging info or information that is not precise enough)


teolemon (approved):

teolemon:
approved @stephanegigandet 
thanks @VaiTon for applying my patch

PR #4958, opened 2021-03-11
Add some Russian additives, ingredients and minerals

blazern:
Adding some Russian additives, ingredients and minerals.


taxonomies/additives.txt
```diff
@@ -1487,7 +1487,7 @@ oc:E150, caramèl
 pl:E150, karmel
 pt:E150, caramelo
 ro:E150, Carmel
-ru:E150, карамель
+ru:E150, карамель, burnt sucrose
```
stephanegigandet, line 1490: the translations should only be in the target language, why add "burnt sucrose"?
and it looks like we have no product with it: https://world.openfoodfacts.org/ingredient/burnt-sucrose

have you seen it in an actual product?

blazern: Oops, looks like a mistake

blazern:
Removed `burnt sucrose`.

stephanegigandet (approved):

blazern:
Resolved a conflict

PR #4967, opened 2021-03-12
Added Russian ingredients, additives and additives classes which have Russian 'and' in their name, and which appear more than 1 time in products of OFF

blazern:
Added Russian ingredients, additives and additives classes which have Russian 'and' in their name, and which appear more than 1 time in products of OFF.


stephanegigandet (approved):
Thanks!

Issue #4890, opened 2021-02-22
The producer signup form is not responsive


teolemon:
The producer signup form is not responsive
![image](https://user-images.githubusercontent.com/1689815/108745177-1291d700-753b-11eb-8269-ca2ccf0787d5.png)

andersonarorjdev:
Is this bug on App or in web?

stephanegigandet:
This has been done.

PR #4973, opened 2021-03-12
display register or save in user form submit button, fixes #4889

### LINKED ISSUE #4889: Submit button of the create user form has no associated label, causing "localized string not found" in some contexts
Missing translations on the Producer Platform
image

stephanegigandet:
fixes #4889

### DIFF
--- a/templates/user_form.tt.html
+++ b/templates/user_form.tt.html
@@ -26,8 +26,10 @@
                     <input type="hidden" name="type" value="[% type %]" />
                     [% IF userid %]
 						<input type="hidden" name="userid" value="[% userid %]" />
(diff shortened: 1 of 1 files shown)


teolemon:
@stephanegigandet Another point is that the form is not responsive, unlike the rest of the page, which reduces conversion rate.

hangy (approved):

stephanegigandet:
> Another point is that the form is not responsive, unlike the rest of the page, which reduces conversion rate.

There's more work needed for https://github.com/openfoodfacts/openfoodfacts-server/issues/4890 , it's not as simple. Ideally we shoud refactor the user form's code to be more like the org form.

PR #4999, opened 2021-03-16
Import: fix for "not specified" values, fix for obsolete and unobsolete

stephanegigandet:
This solves issues with data we got from producers that was unexpected, and a bug regarding the obsolete products.

### DIFF
--- a/lib/ProductOpener/Import.pm
+++ b/lib/ProductOpener/Import.pm
@@ -1023,69 +1023,98 @@ EMAIL
 							$new_field_value =~ s/litre|litres|liter|liters/l/i;
 							$new_field_value =~ s/kilogramme|kilogrammes|kgs/kg/i;
 					}
-
+					
 					$new_field_value =~ s/\s+$//g;
-					$new_field_value =~ s/^\s+//g;
-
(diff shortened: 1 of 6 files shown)


lib/ProductOpener/Import.pm
```diff
@@ -1023,69 +1024,98 @@ EMAIL
...
 					}
-
+					
 					$new_field_value =~ s/\s+$//g;
-					$new_field_value =~ s/^\s+//g;
+					$new_field_value =~ s/^\s+//g;					
+					
+					# Some fields like "obsolete" can have yes/no values
+					if ($field eq "obsolete") {
+						if ($new_field_value =~ /^\s*(1|y|yes|o|oui|s|si)\s*$/i) {
```
teolemon, line 1033: Can this be factored ?

stephanegigandet: We can move the strings in a hash table above the function as we do for other regexps for ingredients analysis etc.

stephanegigandet: It's done.

teolemon (approved):

PR #5001, opened 2021-03-17
Ignore some Russian not ingredients when they're within ingredients list

blazern:
Ignoring some Russian not ingredients when they're within ingredients list.


stephanegigandet (approved):

blazern:
Wow that was fast! :)

Issue #5017, opened 2021-03-19
Oils parentheses enumerations are not expanded for Russian language


blazern:
On Russian products manufacturers often specify different oils in the next way: "vegetable oil (sunflower, soy)" ("масло растительное (подсолнечное, соевое)" in Russian).
The "sunflower" and "soy" parts are being treated by OFF as separate ingredients: https://ru.openfoodfacts.org/%D0%B8%D0%BD%D0%B3%D1%80%D0%B5%D0%B4%D0%B8%D0%B5%D0%BD%D1%82/%D0%BF%D0%BE%D0%B4%D1%81%D0%BE%D0%BB%D0%BD%D0%B5%D1%87%D0%BD%D0%BE%D0%B5
But those are not separate ingredients, those are specifications of the vegetable oil kinds.

There's already a mechanism which does the `vegetable oil (sunflower, soy)` transformation into `sunflower vegetable oil, soy vegetable oil` for English and French.
But it's in Perl and I don't know Perl :(

<details>
  <summary>
But I know Python and I wrote a little snippet which seems to do what is needed:</summary>

```python
import re

def prettify_oils_ru(input_str):
    # (oil|oil from vegetables|vegetable oil)
    vegetable_oil_ru = '(масло|масло растительное|растительное масло)'

    # will match: 'vegetable oil (olive, sunflower)'
    # won't match (nested parentheses): 'vegetable oil (olive, tasty (soybean, palm), sunflower)'
    regex = '{}\s?\([^()]*\)$'.format(vegetable_oil_ru)
    if not re.search(regex, input_str):
        return None
        print('Not vegetable oil or not acceptable format')
        return

    # 'vegetable oil (olive, sunflower and soy)' -> 'olive, sunflower and soy'
    types_str = input_str[input_str.find('(')+1 : input_str.rfind(')')]
    # 'olive, sunflower and soy' -> ['olive', 'sunflower', 'soy']
    types = re.compile('\s*,\s*|\s+и\s+|\s+или\s+').split(types_str)
    # ['olive', 'sunflower', 'soy'] -> ['olive oil', 'sunflower oil', 'soy oil']
    different_oils = map(lambda oil_type: 'масло {}'.format(oil_type), types)

    return different_oils

input_strs = [
    'масло растительное (подсолнечное, соевое, оливковое)',
    'растительное масло (подсолнечное, соевое, оливковое)',
    'масло (подсолнечное, соевое, оливковое)',
[...]

PR #5018, opened 2021-03-19
Make ingredients categories+types (e.g. oils (rapeseed, oil) -> rapeseed oil, olive oil) more generic

stephanegigandet:
We have some ingredients parsing code for French to transform ingredients that are specified with a category + some types (e.g. vegetable oils (olive, palm and sunflower) to individual ingredients: olive vegetable oils, palm vegetable oils, sunflower vegetable oils.

This PR is to make this code more generic, so that it can be applied or adapted more easily to other languages.

I added a few examples for English and Russian (from @blazern : #5017 ).

We can add more categories and more types gradually (and more corresponding tests too).

We can also probably extend the system to more languages easily (e.g. Italian, Portuguese, Spanish). I'm not sure about the languages with compound words like Dutch, Finnish or German, but I'm guessing we can find a way.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -2502,7 +2502,6 @@ sub normalize_a_of_b($$$) {
 	$b =~ s/^\s+//;
 
 	if ($lc eq "en") {
-
 		return $b . " " . $a;
 	}
 	elsif ($lc eq "es") {
@@ -2518,6 +2517,9 @@ sub normalize_a_of_b($$$) {
 			return $a . " de " . $b;
 		}
 	}
+	elsif ($lc eq "ru") {
+		return $b . " " . $a;		
+	}
 }
 
 
@@ -2531,13 +2533,19 @@ sub normalize_enumeration($$$) {
 	my $enumeration = shift;
 
 	$log->debug("normalize_enumeration", { type => $type, enumeration => $enumeration }) if $log->is_debug();
+	
+	# If there is a trailing space, save it and output it
+	my $trailing_space = "";
+	if ($enumeration =~ /\s+$/) {
+		$trailing_space = " ";
+	}
 
 	my $and = $Lang{_and_}{$lc};
 	#my $enumeration_separators = $obrackets . '|' . $cbrackets . '|\/| \/ | ' . $dashes . ' |' . $commas . ' |' . $commas. '|'  . $Lang{_and_}{$lc};
 
 	my @list = split(/$obrackets|$cbrackets|\/| \/ | $dashes |$commas |$commas|$and/i, $enumeration);
 
-	return join(", ", map { normalize_a_of_b($lc, $type, $_)} @list);
+	return join(", ", map { normalize_a_of_b($lc, $type, $_)} @list) . $trailing_space;
 }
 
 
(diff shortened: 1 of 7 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -3661,6 +3669,413 @@ sub replace_additive($$$) {
...
+			my $and_of = ' - ';
+			if (defined $and_of{$product_lc}) {
+				$and_of = $and_of{$product_lc};
+			}
+			my $and_or = ' - ';
+			if (defined $and_or{$product_lc}) {
+				$and_or = $and_or{$product_lc};
+			}	
+			
+			if (($product_lc eq "en") or ($product_lc eq "ru")) {
```
blazern, line 3999: Do you mind if logic for English and Russian would be a bit different?
In Russian we never say "подсолнечное масло растительное" ("sunflower oil made of vegetables"), instead we say "масло растительное подсолнечное" ("vegetable oil made of sunflowers").
So instead of converting `масло растительное (подсолнечное)` into `подсолнечное масло растительное`, I think it should be converted into `масло растительное подсолнечное`.

On the other hand, although current expanded phrasing sounds a bit odd for a Russian ear, it definitely would be properly understood and current patch solves the problem of #5017 
So it'd be ok if the patch was merged like that :)

stephanegigandet: We definitely should try to convert to what sounds best in Russian.

I was looking at https://ru.openfoodfacts.org/ingredients?filter=%D0%BC%D0%B0%D1%81%D0%BB%D0%BE and my impression was that in most cases масло was the last word. e.g. Пальмовое масло

In English we convert "Oil (Palm, Olive)" in "Palm oil, olive oil".
In French we do the reverse: "Huile (palme, olive)" becomes "Huile de palme, huile d'olive" (litterrally "oil of palm, oil of olive").
In Russian, what should we convert "Масло (пальмовое, оливковое)" to?
And what about "масло растительное (пальмовое, оливковое)" ?

I obviously don't speak Russian at all, so please tell me what's the best way to transform it. We can certainly improve the logic.

blazern: > I was looking at https://ru.openfoodfacts.org/ingredients?filter=%D0%BC%D0%B0%D1%81%D0%BB%D0%BE and my impression was that in most cases масло was the last word. e.g. Пальмовое масло

You're absolutely correct about most common cases.

In Russian if the phrase has only word `масло` (`oil`), then the order does not matter (`подсолнечное масло` has absolutely same meaning as `масло подсолнечное`). That's the case you've described.

But if the phrase has 2 words instead of 1: `масло растительное` (`oil made from vegetables`) or `растительное масло` (`vegetable oil`), then the location of `подсолнечное` (`sunflower`) starts to matter a bit:
1. `oil` first: `масло растительное подсолнечное` and `растительное масло подсолнечное` both sound natural to a Russian ear (literal meaning: `oil made of vegetable - sunflower` and `vegetable oil made of sunflowers`).
2. `sunflower` first: `подсолнечное растительное масло` (`sunflower vegetable oil`) is also okay, but `подсолнечное масло растительное` sounds weird (literal meaning: `sunflower oil made of vegetables`).

So when `oil`/`vegetable oil` is put first, all phrases sound okay. But when `sunflower` (or any other type) is put first, then there's a case which sounds weird.

> In Russian, what should we convert "Масло (пальмовое, оливковое)" to?
> And what about "масло растительное (пальмовое, оливковое)" ?

The most simple solution would be to always put `oil`/`vegetable oil` before oil type:
- `Масло (пальмовое, оливковое)` -> `Масло пальмовое`, `масло оливковое`
- `масло растительное (пальмовое, оливковое)` -> `масло растительное пальмовое`, `масло растительное оливковое`

stephanegigandet: @blazern Thanks a lot for the explanation. I made changes so that we always put the category first (oil) and then the type (palm).

How does it look now?

It implies that we need to make sure we have synonyms in the ingredients taxonomy so that entries like масло растительное пальмовое, масло растительное оливковое are recognized.

blazern:
Is the `ingredients.result...` files update needed for the tests?
Initially the files weren't updated by the PR.

lib/ProductOpener/Ingredients.pm
blazern: > How does it look now?

Looks good! Thank you!

> It implies that we need to make sure we have synonyms in the ingredients taxonomy so that entries like масло растительное пальмовое, масло растительное оливковое are recognized.

I'm not sure I get it.
Do you mean that Russian oils should move from the `unknown` section of ingredients list into `ingredients.txt` file? :)

stephanegigandet:
> Is the `ingredients.result...` files update needed for the tests?
> Initially the files weren't updated by the PR.

The ingredients.result.* files are the built version of the ingredients.txt taxonomy. I updated the ingredients taxonomy to add a couple of synonyms for Russian oils (but we will have to add similar ones for all the oils).

blazern (approved):
Logic for Russian LGTM.
But I don't know Perl so I couldn't review most of the code changes.

CharlesNepote (approved):

PR #5022, opened 2021-03-21
Fix warning during execution of CodeQL workflow

hangy:
`HEAD^2` is now unnecessary according to https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/671276345

> git checkout HEAD^2 is no longer necessary. Please remove this step as Code Scanning recommends analyzing the merge commit for best results.

The CodeQL config is now updated to match https://github.com/github/codeql-action/blob/97a70e6013beb841a16b2660ce2c5a30982128c1/README.md#usage.


teolemon (approved):
CodeQL seems to be working

PR #5033, opened 2021-03-22
New Portugal recycling containers images

laralem:
New images for recycling containers used in Portugal. PNG and SVG versions. The SVG version was saved in Inkscape in "SVG optimized" format, without changes in Inkscape options (default options). I hope it's the right format.


stephanegigandet (approved):
Looks good to me!

The only change I would make would be to use "ecoponto-amarelo.90x90.svg" as the file name, instead of "ecoponto-amarelo-color.90x90.svg", as we will just use "Ecoponto Amarelo" for the label name.

PR #5035, opened 2021-03-22
Categories taxonomy: put liquorice under candies, fix German entry for pickles

stephanegigandet:
From Rochus Wessels on Slack in #taxonomies :

The German translation of "Pickles" should be "Eingelegte Lebensmittel" (pickled food) or something like that, but not "Gurken", because that is "Cucumbers" and a lot of pickled foods are no cucumbers.

The other one is the categorization of "Liquorice" as sub-category of "Hebal teas". Liquorice teas exist, but the category is mostly used for the sweets/candies/sugary snacks and those are definitely no teas, so it would probably be better a subcategory of "Sugary snacks" and have a separate category for "Liquorice teas".

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -6124,8 +6124,9 @@ wikidata:en:Q158746
 
 # description:en:LIQUORICE is the root of Glycyrrhiza glabra from which a sweet flavour can be extracted.
 
-<en:Herbal teas
-# <en:Plant
+# Almost all entries under the en:liquorice category are candies
+
+<en:candies
 en:liquorice, licorice
 af:Soethout
 ar:عرقسوس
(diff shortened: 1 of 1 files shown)


stephanegigandet:
In fact for all herbal tea entries, we probably should put "XYZ herbal tea" at the canonical name, and "XYZ" as a synonym, only if there is no doubt that XYZ refers to the herbal tea.  e.g. for the entry "Juniper berries", it should not be an herbal tea either.

taxonomies/categories.txt
```diff
@@ -68394,7 +68399,7 @@ wikidata:Q1886522
 en:Pickles
 ca:Conserves
 da:Syltede agurker
```
svensven, line 68401:
```suggestion
```
This also means "pickled cucumbers", so could move to the en:Pickled cucumbers section around line 75405.

stephanegigandet: @svensven : thanks, I just moved it.

moon-rabbitOFF (approved):

PR #5036, opened 2021-03-22
Many small improvements for the producers platform

stephanegigandet:
- Org profiles have an extra checkbox for admins to disable CodeOnline imports
- Org profiles allow admins to input a list of GLNs. The GS1 org ids (GLN) are now stored in a hash (orgs_glns.sto), and they are used to force the match of a GLN to an org during Equadis / CodeOnline imports.
- Imports to the public platform now skip products that would result in an owner change
- The export form has an extra checkbox to allow overwriting owners
- When editing an org profile, admins can now create the org profile if it does not exist yet (and they are warned that they are editing an unknown org profile)
- The side column form for admins to switch to a specific org now defaults to organizations if the entered it does not start by user- or org-. i.e. it's possible to just input "nestle-france".

### DIFF
--- a/cgi/org.pl
+++ b/cgi/org.pl
@@ -66,7 +66,13 @@ my $org_ref = retrieve_org($orgid);
 
 if (not defined $org_ref) {
 	$log->debug("org does not exist", { orgid => $orgid }) if $log->is_debug();
-	display_error($Lang{error_org_does_not_exist}{$lang}, 404);
+	
+	if ($admin) {
+		$template_data_ref->{org_does_not_exist} = 1;
+	}
+	else {
+		display_error($Lang{error_org_does_not_exist}{$lang}, 404);
+	}
 }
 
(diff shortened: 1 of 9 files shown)


cgi/org.pl
```diff
@@ -94,9 +100,39 @@
...
+				if (not defined $org_ref) {
+					$org_ref = create_org($User_id, $orgid);
+				}
+				
+				foreach my $field ("enable_manual_export_to_public_platform", "activate_automated_daily_export_to_public_platform", "do_not_import_codeonline") {
 					$org_ref->{$field} = remove_tags_and_quote(decode utf8=>param($field));
 				}
+				
+				# List of GS1 GLNs
+				# Remove existing GLNs
```
teolemon, line 114: Can we perhaps send a mail to the org owner or to platform admins ?

teolemon (approved):

PR #5039, opened 2021-03-22
Portuguese translation in packaging_shapes.txt

laralem:
New Portuguese translations and a few fixes in Portuguese existing translations.

**Removed**:

> en:Mold
> fr:Moule

since they were duplicates (line 113 and line 124). Please check. Thank you.

By the way, I could also add a few translations in Brazilian Portuguese, but how should I do that?
-Add to "pt:"? Since most of the words are the same with a few exceptions.
-Add to "pt-br:" or "pt-BR:"? In this case it will have also some words used in "pt:" translation.

Example of one of those exceptions:

> pt:Carica, rolha
> pt-BR:Chapinha, rolha

But I don't see a problem using just one translation:

> pt:Carica, chapinha, rolha

I'm asking this because in Crowdin we have both Portuguese and Brazilian Portuguese translations.

### DIFF
--- a/taxonomies/packaging_shapes.txt
+++ b/taxonomies/packaging_shapes.txt
@@ -2,12 +2,14 @@
 
 en:Unknown
 fr:Inconnu, inconnue, inconnus, inconnues
+pt:Desconhecido
 
 # Use singular as the main name, and add plural as synonym
 en:Tray, trays
 de:Tablettverpackung, Tablett
 fr:Barquette, barquettes
 hu:Tálca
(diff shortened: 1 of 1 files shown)


stephanegigandet:
> But I don't see a problem using just one translation:
> 
> > pt:Carica, chapinha, rolha

Currently we don't support language variants such as pt_PT and pt_BR, so please try to add all synonyms under pt:

laralem:
> Currently we don't support language variants such as pt_PT and pt_BR, so please try to add all synonyms under pt:

OK, I will merge also some translations already in some taxonomies but changed to comments since they are not supported as you said. Thank you.

PR #5051, opened 2021-03-24
Indicate source of data for CodeOnline and Equadis + dedicated pages

stephanegigandet:
This will add a link at the bottom of product pages to indicate that some data comes from CodeOnline, or from Equadis. The link will lead to a page with a small text about each, plus the list of corresponding products. Essentially it's just 2 additional values to the /data-sources facet.


stephanegigandet:
Sample products: 

https://fr.openfoodfacts.dev/produit/3083681023398/xyz-laitue-mache-bonduelle

https://fr.openfoodfacts.dev/produit/3080210004446/1664-4x50cl-boite-1664-5-50-degre-alcool

teolemon (approved):

PR #5060, opened 2021-03-26
Translations for packaging_recycling.result.txt

laralem:
Adding Portuguese translations for packaging_recycling.result.txt


laralem:
Should I update packaging_recycling.txt also? It has German translations not present in this PR.

stephanegigandet (approved):
Thanks!

Issue #4640, opened 2020-12-16
Improve accessibility by fixing issues reported by the Accessibility Fundation


stephanegigandet:
The Accessibility Foundation has made an accessibility of the Open Food Facts web site. There is a number of issues that they pointed out and that we should fix.

[Report OpenFoodFact 15-12-2020.pdf](https://github.com/openfoodfacts/openfoodfacts-server/files/5702582/Report.OpenFoodFact.15-12-2020.pdf)

Who would be interested in fixing some of them?

jolesh:
Is there a good way of keeping track of what we already have implemented and what is still to do? A project for example?

Ban3:
> Is there a good way of keeping track of what we already have implemented and what is still to do? A project for example?
Not yet, but doing something like that was my first thought when reading the report. A project sounds like a good idea.

stephanegigandet:
@ban3 @jolesh : a project is a good idea, does one of you want to create it, and create the corresponding bugs with the items mentioned in the report?

@leventgunay : do you still want to address some of the reported issues to improve accessibility?

roshnaeem:
@stephanegigandet can I start working on it? I want to address few issues to improve accessibility.
.

stephanegigandet:
Hi @roshnaeem , sure, thank you!
Could you create a separate bug for each of the issues you intend to work on, and then link them from this bug?
That way it will be easier to discuss each issue.
I just created one for the links underlining: https://github.com/openfoodfacts/openfoodfacts-server/issues/5074

roshnaeem:
Yes, sure @stephanegigandet. It will be easier for everyone.

yuktea:
> @Ban3 @jolesh : a project is a good idea, does one of you want to create it, and create the corresponding bugs with the items mentioned in the report?
> 
> @leventgunay : do you still want to address some of the reported issues to improve accessibility?

@stephanegigandet should I work on creating this project? (I'll point out and document what we have fixed and create issues for what we have not from the accessibility report)

alexgarel:
@yuktea yes, go ahead, it's a good idea. You should verify for some points are already fixed.

PR #5071, opened 2021-03-29
Fixed accessibility issue 1.1.1

roshnaeem:
Description:
Issue 1.1.1: Changed the alt text for the logo.

Related issue  https://github.com/openfoodfacts/openfoodfacts-server/issues/4640

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -2446,6 +2446,10 @@ msgctxt "site_name"
 msgid "Open Food Facts"
 msgstr ""
 
+msgctxt "logo_site_name"
+msgid "Open Food Facts logo"
+msgstr ""
+
 msgctxt "sort_by"
 msgid "Sort by"
 msgstr ""
--- a/po/common/en.po
+++ b/po/common/en.po
@@ -2507,6 +2507,10 @@ msgctxt "site_name"
 msgid "Open Food Facts"
 msgstr "Open Food Facts"
 
+msgctxt "logo_site_name"
+msgid "Open Food Facts logo"
+msgstr "Open Food Facts logo"
+
 msgctxt "sort_by"
 msgid "Sort by"
 msgstr "Sort by"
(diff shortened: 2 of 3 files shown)


po/common/common.pot
```diff
@@ -285,9 +285,9 @@ msgid ""
 "svg/donate-icon.svg\" alt=\"Donate to Open Food Facts\" /></"
 "a><p><<site_name>> is made by a non-profit association, independent from the "
 "industry. It is made for all, by all, and it is funded by all. You can "
-"support our work by <a href=\"https://world.openfoodfacts.org/donate-to-open-food-facts\">donating to Open Food Facts</a> and "
+"support our work by <a href=\"https://world.openfoodfacts.org/donate-to-open-food-facts\" style="text-decoration:underline;">donating to Open Food Facts</a> and "
```
stephanegigandet, line 288: We have many other links like this one that are in the middle of other text. I'm thinking a better idea would be to change the CSS to underline links inside paragraph tags.

To discuss: https://github.com/openfoodfacts/openfoodfacts-server/issues/5074

po/common/en.po
```diff
@@ -2507,6 +2507,10 @@ msgctxt "site_name"
 msgid "Open Food Facts"
 msgstr "Open Food Facts"
 
+msgctxt "logo_site_name"
+msgid "Logo Open Food Facts"
```
stephanegigandet, line 2511: Could you change it to "Open Food Facts logo"?

po/common/common.pot
roshnaeem: Yes, I'll start working on it.

po/common/en.po
roshnaeem: yes, sure. You want "L" to be small?

stephanegigandet (changes requested):
I think it would be better to split this into 2 PRs. The alt text for the logo is easy, but the underline links would probably be better done in another way: https://github.com/openfoodfacts/openfoodfacts-server/issues/5074

If we split it, then we can merge each PR independently.

po/common/en.po
stephanegigandet: yes, small L.

You can keep this PR for the logo, but I think the other bug (the underlined links) should be done in another way, and moved to another PR.

roshnaeem: > I think it would be better to split this into 2 PRs. The alt text for the logo is easy, but the underline links would probably be better done in another way: #5074
> 
> If we split it, then we can merge each PR independently.

Yes, it'll be better. I will open PR for each issue separately now.

stephanegigandet (approved):
Thanks!

PR #5077, opened 2021-03-30
Some improvements

CharlesNepote:


taxonomies/ingredients.txt
```diff
@@ -19947,6 +19949,31 @@ wikipedia:en:https://en.wikipedia.org/wiki/Amaretto
 #ingredient/amaretto has 15 products in 4 languages @2018-12-21
 
 
+<en:liqueur
+ca:Limoncello
+cs:Limoncello
+de:Limoncello
+en:Limoncello
```
svensven, line 19956:
```suggestion
en:Limoncello
ca:Limoncello
cs:Limoncello
de:Limoncello
```

taxonomies/ingredients.txt
```diff
@@ -8628,6 +8628,7 @@ sv:pastöriserad vispgrädde
 <en:cream
 en:whipped cream
 ca:crema de llet montada, nata montada
+da:piskefløde
```
svensven, line 8631: this is already under "en:whipping cream" too (along with german and dutch duplicates)

taxonomies/ingredients.txt
aleene: Looks like `xx:Limoncello` is more appropriate.

CharlesNepote: So it would be:
```
en:Limoncello
ja:リモンチェッロ
ko:리몬첼로
lt:Limončelas
ru:Лимончелло
zh:檸檬酒
xx:Limoncello
wikipedia:en:https://en.wikipedia.org/wiki/Limoncello
wikidata:en:Q736208
# https://world.openfoodfacts.org/cgi/search.pl?search_terms=limoncello&search_simple=1&action=process
# 226 "Limoncello" via OFF search on 2021-01-11
```
right?

Also I don't find it as a category, should I add it the same as in ingredients?

taxonomies/ingredients.txt
```diff
@@ -13233,7 +13234,7 @@ es:Sal marina, sal marina de mesa,sal del mar
 eu:Marsalo
 fa:نمک دریایی
 fi:Merisuola
-fr:sel marin, sel de mer, sels minéraux de mer, sel marin de table
+fr:sel marin, sel de mer, sels minéraux de mer, sel marin de table, sel gros de mer
```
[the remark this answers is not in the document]
CharlesNepote: Right, deleted.

taxonomies/ingredients.txt
stephanegigandet: Looks good. We should add it as a category as well indeed

taxonomies/ingredients.txt
stephanegigandet: Right, we should keep the distinction between whipped cream and whipping cream (which is not yet whipped :) )

stephanegigandet (approved):

Issue #1501, opened 2018-11-12
Store the number of scans for each product / day / month / year / total


stephanegigandet:
It would be very interesting to store the number of scans we get for each product (at least for the official OFF apps) so that we can use it to sort by popularity, to see the "life and death" of products, seasonality etc.

hangy:
Considering that the number of scans is not part of the product, I imagine the scans should probably not be stored in the `product.sto` file or the `products` collection, but maybe rather a different collection. The reason is that this would produce a lot of writes writes to the `.sto` file, and storing it in a different collection would enable us to implement #164 by storing non-successful scans, too. Additionally, #291 could be implemented by storing some (GDPR compliant) location of the scan with the originating software, without exposing that data through the MongoDB dump.

stephanegigandet:
Agreed, I think it makes sense to use a different collection for scans. We could update the product collection with aggregate scan data nightly (e.g. a sort key for product popularity). To store the scans themselves, we could also do it nightly from the nginx logs, with possibly some preprocessing to filter out some scans (e.g. if one individual makes too many scans) and aggregation. (e.g. by country and product).

stephanegigandet:
I was thinking more about not only scans, but all kind of user events:
- product scans (both successful and non successfull (i.e. no product in OFF)
- product edits
- photos uploads

I think it would be useful to store all those events, for each user (unless they opt-out), including not logged in users, so that we can not only derive scan statistics, but also create useful user features (e.g. scan history available both on the phone and on the web site), gamification (points, monthly lists of contributors etc.).

I'm thinking that it might make more sense to use a SQL database for those events, with columns like:
timestamp, userid (or anonymized user id if not logged in), barcode, type of event (scan, edit, image upload), event status code (e.g. scan succesfull or not). 

Then from that table of events, we could compute aggregates etc. Having it in a SQL database will also reduce the risk of making mongodb grow too much, and would allow us to move the database to other servers etc. later if needed.

We would need to do some rough sizing estimates (e.g. can we store 1M events a day in a table? can a single table have 1 billion lines?)

What do you think?

hangy:
Instead of a typical SQL database, we could also test some time series databases like OpenTSDB, ClickHouse, or InfluxDB. Their main advantage is that they're purpose built for ingesting large amounts of time based data points. If we want to store the events in order to being able to create time based stats, storing this outside of the normal mongo collection or the .sto files sounds like a sound idea.

[...]

PR #5079, opened 2021-03-30
Improve scanbot, store each product scans in scans.json files

stephanegigandet:
We need scans data to be available for each product. While it would be cool to have data by year, months, days etc. (see #1501), it would be already be very useful to have it by year and by country.

This update to scanbot creates a new scans.sto file with aggregated scan data by year and country.

Example:
```

./sto_to_xml.pl /srv/off/products/800/227/001/4901/scans.sto 
/srv/off/products/800/227/001/4901/scans.sto$VAR1 = {
          '2020' => {
                      'unique_scans_n_by_country' => {
                                                       'world' => 4,
                                                       'fr' => 2,
                                                       'be' => 2
                                                     },
                      'unique_scans_n' => 4,
                      'scans_n' => 4,
                      'unique_scans_rank_by_country' => {
                                                          'fr' => 19,
                                                          'world' => 1,
                                                          'be' => 1
                                                        }
                    }
        };

```

It also changes the way we compute the top products per country, to make it more accurate. Currently the top lists of products per country are derived from the top list of product for the world, and if the product is sold in a country, it is added to the top list of each country. But that means that the most globally popular products always come first (especially for countries where we have few scans compared to others).

The new way really computes the top products for a country based on only the scans for that country.

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -58,6 +58,8 @@ requires 'Text::Fuzzy';
 requires 'Spreadsheet::CSV'; # deps: libspreadsheet-parseexcel-perl
 requires 'File::chmod::Recursive'; # deps: libfile-chmod-perl
 requires 'Devel::Size'; # deps: libdevel-size-perl
+requires 'JSON::Create';
+requires 'JSON::Parse';
 
 # Mojolicious/Minion
 requires 'Mojolicious::Lite';
--- a/lib/ProductOpener/Store.pm
+++ b/lib/ProductOpener/Store.pm
@@ -34,6 +34,8 @@ BEGIN
 		&get_ascii_fileid
 		&store
 		&retrieve
+		&store_json
+		&retrieve_json
 		&unac_string_perl
 		&get_string_id_for_lang
 		&get_url_id_for_lang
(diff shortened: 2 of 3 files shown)


stephanegigandet:
The scans data per country are needed for a research project, so see if we can determine in a better way which product is sold in which country.

VaiTon:
Could we use [JSON::Create](https://metacpan.org/pod/JSON::Create) to create the file? JSON is more common than STO format.

stephanegigandet:
> Could we use [JSON::Create](https://metacpan.org/pod/JSON::Create) to create the file? JSON is more common than STO format.



> Could we use [JSON::Create](https://metacpan.org/pod/JSON::Create) to create the file? JSON is more common than STO format.

We could, if we think that something not written in Perl will read those files. There will be a performance hit, but it may not matter too much.

There are many options, we could also just store this data in a new mongodb collection, or as it is very tabular, a SQL database would probably work very well.

We need scan data quickly for a research project, so unfortunately I won't have time to experiment too much right now. I'm thinking to just implement one way, and then after that we can revisit it.

svensven (approved):

stephanegigandet:
@VaiTon I changed the scans.sto file to a JSON scans.json file.

I used JSON::Create as you suggested. From reading the docs, I'm not sure if it is better than the older JSON::XS module. But in any case, it works.

svensven:
Re speed, I was reading https://metacpan.org/pod/JSON#DESCRIPTION and the cpanfile specifies JSON and JSON::PP (perl version), but not JSONL::XS (C version), so unless it's pulled in by default it may be worth adding.

stephanegigandet:
Discussion regarding the JSON modules: https://github.com/openfoodfacts/openfoodfacts-server/issues/5081

stephanegigandet:
I tested on 2019 data, and now the top products for each country are much more local.

before:
image

after:
image

Issue #5086, opened 2021-04-01
Use product data and scan data to compute the main countries where a product is sold


stephanegigandet:
I plan to create a new MainCountries module:

Products on Open Food Facts have a countries_tags field that list the countries
where a product is sold. The information is entered by users, food producers,
and additional countries are added by the scanbot.pl script if a product
is scanned by 5 or more different IP addresses in a specific country.

In some cases, products may be marked as sold in a country even it is not widely
sold in that country. For instance, a lot of discount shops import products that
are intended for sale to other countries. Scan data can be tricky as well:
people who live close to a border often shop in another country, and they
may scan products when they are back in their home country.

This module tries to determine which countries a product is "mostly" sold in.

The determination can be made on different factors:
- scan data
- languages
- brands
- etc.

PR #5088, opened 2021-04-01
New feature to detect main countries of products based on scans and languages 

stephanegigandet:
Issue #5086

This is just a start, right now it just puts some tags in the /misc facets so that we can experiment and see what works.


teolemon:
How about we start a misc taxonomy ? The tags are now numerous, and really complicated to understand

stephanegigandet:
> How about we start a misc taxonomy ? The tags are now numerous, and really complicated to understand

The misc facet is for things that are not intended to last / be used externally. It's for experiments, debugging, one-offs etc. They can be renamed / deleted etc. at any time.

teolemon (approved):

PR #5091, opened 2021-04-01
Document the ProductOpener::GeoIP.pm module

roshnaeem:
Description: Updated the `GeoIP.pm` file to add its documentation in perlpod format.

Fixes: https://github.com/openfoodfacts/openfoodfacts-server/issues/5089

Related Issue: #2203

### DIFF
--- a/lib/ProductOpener/GeoIP.pm
+++ b/lib/ProductOpener/GeoIP.pm
@@ -18,6 +18,28 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+=head1 NAME
+
+ProductOpener::GeoIP - finds the country name and iso code from the given IP address.
+
+=head1 SYNOPSIS
+
+C<ProductOpener::GeoIP> is used to find the country name and iso code from the IP address using GeoIP2 API.
+
+    use ProductOpener::GeoIP qw/:all/;
+
+    $geolite2_path = '/usr/local/share/GeoLite2-Country/GeoLite2-Country.mmdb';
+    $geolite2_path is a path to geo IP location max mind database file.
+
+    Max Mind DataBase file(.mmdb) is a database format that maps the given IP addresses to their geolocation.
+
+=head1 DESCRIPTION
+
+The module implements the functionality to find the country name, and iso code by using the GeoIP2 API.
+The functions used in this module take the IP address and return the geolocation.
+
+=cut
+
 package ProductOpener::GeoIP;
 
 use utf8;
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Looks good, thank you!

lib/ProductOpener/GeoIP.pm
```diff
@@ -68,6 +106,23 @@ sub get_country_for_ip {
 	return $country;
 }
 
+=head1 FUNCTIONS
+
+=head2 get_country_code_for_ip()
+
+C<get_country_code_for_ip()> takes the IP address as input paramter and returns the country iso code where IP address is located.
```
CharlesNepote, line 113:
```suggestion
C<get_country_code_for_ip()> takes the IP address as input parameter and returns the country iso code where IP address is located.
```

roshnaeem: Thank you so much for pointing this out!!
I have incorporated changes, Please check!!

lib/ProductOpener/GeoIP.pm
```diff
@@ -68,6 +106,23 @@ sub get_country_for_ip {
...
+
+C<get_country_code_for_ip()> takes the IP address as input parameter and returns the country iso code where IP address is located.
+
+=head3 Arguments
+
+One scalar variable ip is passed as an argument.
+
+=head3 Return values
+
+If the function executes successfully it returns the iso code for the country where the IP address is located.
```
CharlesNepote, line 121:
```suggestion
If the function executes successfully it returns the two-character ISO 3166-1 (http://en.wikipedia.org/wiki/ISO_3166-1) alpha code for the country where the IP address is located (eg. "AF" for Afghanistan). ```

File: lib/ProductOpener/GeoIP.pm [hunk 3]
```diff
@@ -49,6 +71,22 @@ if (-e $geolite2_path) {
...
+
+C<get_country_for_ip()> takes the IP address as an input parameter and returns the country name where the IP address is located.
+
+=head3 Arguments
+
+One scalar variable ip is passed as an argument.
+
+=head3 Return values
+
+If the function executes successfully it returns the country name. On the other hand, if it throws an exception, it simply returns undefined.
```
CharlesNepote, line 86: It would be good to specify what is the reference for the "country name". I think it's the english version of Geonames.org names but not the ISO 3166-1 -- https://en.wikipedia.org/wiki/ISO_3166-1. The "United Kingdom" for example, seems to be named "United Kingdom of Great Britain and Northern Ireland" in ISO 3166-1. @stephanegigandet could you tell us?

CharlesNepote (approved):

PR #5094, opened 2021-04-02
Add language agnostic image for the Eco-Score scale

teolemon:
- [ ] Add language agnostic image for the Eco-Score scale.
- [ ] The French image needs to be committed, by the way


stephanegigandet:
The image has some gray extra pixels left from the text. I couldn't find the SVG source so I created a new one.

stephanegigandet (approved):

Issue #5095, opened 2021-04-02
Use a Template::Toolkit template for the HTML code generation of the test_ingredients_analysis.pl cgi


stephanegigandet:
The script currently generates HTML code directly. We should instead make the script compute the necessary data for HTML generation, and the HTML generation should be in a template in /templates/

PR #5104, opened 2021-04-05
Test Ingredients Analysis Templatization

roshnaeem:
Description: Templatization of the file `cgi/test_ingredients_analysis.pl`

Fixes: https://github.com/openfoodfacts/openfoodfacts-server/issues/5095

### DIFF
--- a/cgi/test_ingredients_analysis.pl
+++ b/cgi/test_ingredients_analysis.pl
@@ -44,30 +44,21 @@ use Log::Any qw($log);
 
 ProductOpener::Display::init();
 
+my $template_data_ref = {
+	lang => \&lang,
+};
+
 my $type = param('type') || 'add';
 my $action = param('action') || 'display';
 
 my $ingredients_text = remove_tags_and_quote(decode utf8=>param('ingredients_text'));
 
 
-my $html = '<p>You can use this form to see how an ingredient list is analyzed.</p>';
+my $html= '';
+$template_data_ref->{action} = $action;
+$template_data_ref->{type} = $type;
 
-$html .= start_form(-method => "GET");
-
-$html .= <<HTML
-Ingredients text (language code: $lc): <br/>
-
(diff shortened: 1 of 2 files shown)


cgi/test_ingredients_analysis.pl
```diff
@@ -44,30 +44,21 @@
 
 ProductOpener::Display::init();
 
+my $template_data_ref = {
+	lang => \&lang,
```
stephanegigandet, line 48: There is a new process_template() function that adds lang => \&lang by default, so it's not needed here.

cgi/test_ingredients_analysis.pl
```diff
@@ -104,6 +96,9 @@
 	$full_width = 0;
 }
 
+$tt->process('test_ingredients_analysis.tt.html', $template_data_ref, \$html);
```
stephanegigandet, line 99: It's best to use this new function:

process_template('test_ingredients_analysis.tt.html', $template_data_ref, \$html) or $html = "<p>" . $tt->error() . "</p>";

stephanegigandet (approved):
Looks good, thank you!

cgi/test_ingredients_analysis.pl
stephanegigandet: That function probably was not in the script you used as an example, we'll need to change the old scripts as well so that they use it.

cgi/test_ingredients_analysis.pl
roshnaeem: Alright, let me fix it.

cgi/test_ingredients_analysis.pl
roshnaeem: yes, it was in `spellcheck_test.pl`, should I update that file too?

stephanegigandet: sure you can do it, in a new PR maybe

stephanegigandet (approved):

PR #5131, opened 2021-04-11
build: added cache to github actions

VaiTon:


svensven:
if you pull in at least some of the perl modules using apt rather than cpan, like in the pull request workflow, it'll run much faster. also IIRC some of the modules have non-perl dependencies, like tesseract.

VaiTon:
@svensven sure thanks! If you could do it I would be very grateful. Also the build is failing now because of these modules like tesseract

hangy:
I started something similar in #2431, but IIRC the question was, if we actually want that amount of changes to the binary sto files in this repo, see https://github.com/openfoodfacts/openfoodfacts-server/pull/2431#issuecomment-769869463.

VaiTon:
> 
> 
> I started something similar in #2431, but IIRC the question was, if we actually want that amount of changes to the binary sto files in this repo, see [#2431 (comment)](https://github.com/openfoodfacts/openfoodfacts-server/pull/2431#issuecomment-769869463).

You're right, let's keep your PR. I will only ref the workflow files. Thanks for pointing out!

VaiTon:
@teolemon @hangy @svensven could you review?

teolemon (approved):
Cleaner. No breaking changes from what I can see.

teolemon:
done @VaiTon 
@hangy I believe the changes to be innocuous. NPM is cached, and the yml should be more readable now

VaiTon:
@teolemon @hangy @stephanegigandet bump

.github/workflows/pull_request.yml
```diff
@@ -7,164 +7,212 @@ on:
...
+        with:
+          node-version: "12.x"
+
+      - name: Cache node modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/cpanfile') }}
```
hangy, line 28: `package-lock.json` instead?

.github/workflows/pull_request.yml
```diff
@@ -7,164 +7,212 @@ on:
...
+          # libmodule-build-pluggable-perl
+          #
+
+      - name: Cache perl modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-perl-modules
+        with:
+          path: ~/.cpanm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
```
hangy, line 190: Key should probably be `cpanfile`

hangy (approved):

teolemon:
Thanks @hangy 
@VaiTon : https://github.community/t/github-actions-pull-requests-always-include-some-outdated-checks/16157/4

teolemon:
Can I disable the requirement for gulp, lint and test ?

hangy:
> Can I disable the requirement for gulp, lint and test ?

Temporarily? I don't think they've been replaced …?

PR #5138, opened 2021-04-12
Add app stores logo in app banner on mobile web

### LINKED ISSUE #5092: Redesign the app banner to make it more effective
Redesign the app banner to make it more effective
image

stephanegigandet:
fixes #5092

I added a way to have different messages based on page type at first (e.g. product page or not), but I don't have a very good idea of a page specific message, so I guess let's keep it simple.

image


CharlesNepote (approved):

VaiTon:
Why don't we implement this in a native way using Web app manifest?

https://developer.mozilla.org/en-US/docs/Web/Manifest/related_applications

PR #5139, opened 2021-04-13
create orgs directory if it does not exist

stephanegigandet:
this is to make it easier to work on organization in dev environments.


VaiTon (approved):
Seems good to me

PR #5146, opened 2021-04-14
Translations for allergens.txt

laralem:
Adding and fixing Portuguese translations for allergens.txt


stephanegigandet (approved):
Thank you!

Issue #5142, opened 2021-04-14
categories.txt portuguese translation


laralem:
Sorry, this request maybe  a bit silly. Since the file I changed it's big, I can't edit in browser and don't know a way to upload the file (maybe in GitHub Desktop, but my local installation it's a mess). So I'm asking someone to open a pull request with this file
[categories.txt](https://github.com/openfoodfacts/openfoodfacts-server/files/6307325/categories.txt)

After that and while the pull request is open I will compare and check if I didn't deleted something in other languages by mistake. Since this file its to big to translate it, it's a work in progress.

Thanks in advance.

laralem:
I found how to pull a request with a big file. It seems the file I've upload needs fixing. I'm closing this request.

PR #5147, opened 2021-04-14
Update categories.txt

M123-dev:
Added portuguese translation to categories.txt 

FROM @laralem in #5142 
> So I'm asking someone to open a pull request with this file categories.txt


laralem:
@M123-dev I've found how to make a pull request with a big file. Unfortunately it seems the file I've uploaded needs fixing. I've made a mess somehow and will be painful fixing it, since it takes some time, but I will do it bit by bit.

Please cancel this pull request. I'm really sorry.  Thank you for your help.

M123-dev:
@laralem , no problem

Issue #5106, opened 2021-04-06
Refactor the user creation / user profile form: use Template::Toolkit and make it responsive


stephanegigandet:
The user creation and user profile edit form code is in user.pl + Users.pm.
user.pl has a Template::Toolkit template, but most of the HTML code is in fact generated in Users.pm functions and passed as-is to the template.

The form uses tables to display the field names + field values in separate columns, and it makes it unresponsive.

It would be a good thing to use the approach used for the organization edit form: org.pl:
- org.pl creates a $template_data_ref->{sections} data structure that contains different sections with different fields.

e.g.
	foreach my $contact ("customer_service", "commercial_service") {
		
		push @{$template_data_ref->{sections}}, {
			id => $contact,
			fields => [
				{ field => $contact . "_name" },
				{ field => $contact . "_address", type => "textarea" },
				{ field => $contact . "_email" },
				{ field => $contact . "_link" },
				{ field => $contact . "_phone" },				
				{ field => $contact . "_info", type => "textarea"  },
			],
		};
	}

Then the labels / descriptions / notes etc. are fetched from the .po files, and the existing values are added.

And the template is responsible to display the different sections and fields in any way.

Suggested approach:

- copy the org.pl and its template to a user2.pl script + user2 template.
- convert it to a very simple user edit form (e.g. just editing the user name)
- gradually add other fields and sections that are editable in the current user.pl form
- when all of the user.pl features are available in user2.pl, replace user.pl by user2.pl

stephanegigandet:
Will solve #4890

stephanegigandet:
This has been done.

PR #5150, opened 2021-04-14
Refactor the user creation / user profile form

roshnaeem:
**Description:**
1. Created three new files to provide the same functionality `user.pl` and `user-form.tt.html` were providing.
2. `user.pl` uses three functions defined in `Users.pm` for user creation and user profile form. These three functions consist highly of HTML.
3. The newly created three files which can be viewed in this PR are using Template::Toolkit to provide the same functionality and remove the usage of HTML including functions. 
4. Moreover, removed the usage of `<table>` to make the form responsive.

**Related issues and discussion:** https://github.com/openfoodfacts/openfoodfacts-server/issues/5106

### DIFF
--- a/cgi/user2.pl
+++ b/cgi/user2.pl
@@ -167,7 +167,7 @@ if ($action eq 'display') {
 
 	$template_data_ref->{sections} = [];
 
-	if ($user_ref) {
+		if ($user_ref) {
 		push @{$template_data_ref->{sections}}, {
 			id => "user",
 			fields => [
@@ -198,10 +198,16 @@ if ($action eq 'display') {
 		# Professional account
 		push @{$template_data_ref->{sections}}, {
 			id => "professional",
+			name => lang("pro_account"),
+			description => "if_you_work_for_a_producer",
+			note => "producers_platform_description_long",
 			fields => [
 				{
 					field => "pro",
 					type => "checkbox",
+					label => lang("this_is_a_pro_account"),
+					warning => sprintf(lang("this_is_a_pro_account_for_org"),"<b>" . $user_ref->{org} . "</b>"),
+					value => "off",
 				},
 				{
 					field => "pro_checkbox",
@@ -210,11 +216,19 @@ if ($action eq 'display') {
 				},
 				{
 					field => "requested_org",
+					label => lang("producer_or_brand") . ":",
 				}
 			]
 		};
 
-		my $team_section_ref = { id => "teams", fields => [] };
+		# Teams section
+		my $team_section_ref = {
+			id => "teams",
+			name => lang("teams") . " (" . lang("optional") . ")",
+			description => "teams_description",
+			note => "teams_names_warning",
+			fields => []
+		};
 		for (my $i = 1; $i <= 3; $i++) {
 			push @{$team_section_ref->{fields}}, {
 				field => "team_". $i,
@@ -224,9 +238,15 @@ if ($action eq 'display') {
 
 		push @{$template_data_ref->{sections}}, {%$team_section_ref};
 
-		my $administrator_section_ref = { id => "administrator", fields => [] };
+		# Admin section
+		my $administrator_section_ref = {
+			id => "administrator",
+			name => "Administrator fields",
+			fields => []
+		};
 		push  @{$administrator_section_ref->{fields}}, {
 			field => "org",
+			label => lang("organization"),
 		};
 		foreach my $group (@user_groups) {
 			push @{$administrator_section_ref->{fields}}, {
@@ -240,8 +260,30 @@ if ($action eq 'display') {
 		push @{$template_data_ref->{sections}}, {%$administrator_section_ref};
 	}
 
-	# Add labels, types, descriptions, notes and existing values for all fields
+	if ( ( defined $user_ref->{org} ) and ( $user_ref->{org} ne "" ) ) {
+
+		$template_data_ref->{accepted_organization} = $user_ref->{org};
+		$template_data_ref->{pro_account_org} = sprintf(lang("this_is_a_pro_account_for_org"),"<b>" . $user_ref->{org} . "</b>");
+	}
+	elsif ((defined $options{product_type}) and ($options{product_type} eq "food")) {
+		my $requested_org_ref = retrieve_org($user_ref->{requested_org});
+		$template_data_ref->{requested_org_ref} = $requested_org_ref;
+		$template_data_ref-> {org_name} = sprintf(lang("add_user_existing_org"), org_name($requested_org_ref));
+		$template_data_ref->{teams_flag} = not ((defined $server_options{private_products}) and ($server_options{private_products}));
+	}
+
+		# Add labels, types, descriptions, notes and existing values for all fields
 	foreach my $section_ref (@{$template_data_ref->{sections}}) {
+
+		# Descriptions and notes for sections
+		if (defined $section_ref->{id}) {
+			if ($section_ref->{description}) {
+				$section_ref->{description} = lang($section_ref->{description});
+			}
+			if ($section_ref->{note}) {
+				$section_ref->{note} = lang($section_ref->{note});
+			}
+		}
 
 		foreach my $field_ref (@{$section_ref->{fields}}) {
 
@@ -250,7 +292,7 @@ if ($action eq 'display') {
 			# Default to text field
 			if (not defined $field_ref->{type}) {
 				$field_ref->{type} = "text";
-			}
+			};
 
 			# id to use for lang() strings
 			my $field_lang_id = $field;
@@ -262,28 +304,16 @@ if ($action eq 'display') {
 			# Label
 			if (not defined $field_ref->{label}) {
 				$field_ref->{label} = lang($field_lang_id);
-			}
+			};
+
+			if (((defined $user_ref->{pro}) and ($user_ref->{pro}))
+			or ((defined $server_options{producers_platform}) and ($type eq "add"))) {
+				if (($section_ref->{id} eq "professional") and $field_ref->{type} eq "checkbox") {
+					$field_ref->{value} = "on";
+				}
+			};
 		};
 	};
-
-	$template_data_ref->{accepted_organization} = $user_ref->{org};
-	$template_data_ref->{pro_account_org} = sprintf(lang("this_is_a_pro_account_for_org"),"<b>" . $user_ref->{org} . "</b>");
-
-	my $pro_checked = '';
-
-	# Check the "pro account" checkbox for register screen on the producers platform
-
-	if (((defined $user_ref->{pro}) and ($user_ref->{pro}))
-		or ((defined $server_options{producers_platform}) and ($type eq "add"))) {
-		$pro_checked = "on";
-	}
-
-	$template_data_ref->{product_type} = $options{product_type};
-	$template_data_ref->{pro_checked} = $pro_checked;
-	my $requested_org_ref = retrieve_org($user_ref->{requested_org});
-	$template_data_ref->{requested_org_ref} = $requested_org_ref;
-	$template_data_ref-> {org_name} = sprintf(lang("add_user_existing_org"), org_name($requested_org_ref));
-	$template_data_ref->{teams_flag} = not ((defined $server_options{private_products}) and ($server_options{private_products}));
 	$template_data_ref->{admin_flag} = $admin;
 
 }
--- a/cgi/user2.pl
+++ b/cgi/user2.pl
@@ -1,0 +1,365 @@
+#!/usr/bin/perl -w
+
+# This file is part of Product Opener.
+#
+# Product Opener
+# Copyright (C) 2011-2020 Association Open Food Facts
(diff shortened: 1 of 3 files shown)


templates/user_form2.tt.html
```diff
@@ -0,0 +1,241 @@
...
+[% END %]
+
+[% IF action == 'display' %]
+
+    <!-- Start form -->
+
+    <form method="post" action="/cgi/user2.pl" enctype="multipart/form-data">
+		[% FOREACH section IN sections %]
+			[% IF user OR section.id == "user" %]
+				[% FOREACH field IN section.fields %]
```
stephanegigandet, line 37: Instead of having 3 different [% FOREACH field IN section.fields %] loops, for each of the sections, it would be better to keep only one that works for all 3 sections, like it is done for the org.pl form:

		[% FOREACH section IN sections %]
		
			[% IF admin OR section.id != "admin" %]
		
				[% IF section.name %]
					<fieldset[% IF section.id == "admin" %] class="panel callout"[% END %]>
						<legend>[% section.name %]</legend>
				[% END %]
				
				[% IF section.description %]
					<p>[% section.description %]</p>
				[% END %]
				[% IF section.note %]
					<p>[% section.note %]</p>
				[% END %]

				[% FOREACH field IN section.fields %]

					[% IF field.description %]
						<p>[% field.description %]</p>
					[% END %]								

					[% IF field.type == 'text' %]
						<label for="[% field.field %]">[% field.label %]</label>
						<input type="text" id="[% field.field %]" name="[% field.field %]" value="[% field.value %]" />
					[% ELSIF field.type == 'textarea' %]
						<label for="[% field.field %]">[% field.label %]</label>
						<textarea id="[% field.field %]" name="[% field.field %]" style="height:100px;">[% field.value %]</textarea>
					[% ELSIF field.type == 'checkbox' %]
						<label for="[% field.field %]">
							<input type="checkbox" id="[% field.field %]" name="[% field.field %]" [% IF field.value == 'on' %]checked="checked"[% END %] />
							[% field.label %]
						</label>
					[% END %]
					
					[% IF field.note %]
						<p style="font-size:.8em">&rarr; [% field.note %]</p>
					[% END %]

				[% END %]
				
				[% IF section.name %]
					</fieldset>
					[% IF section.name == "admin" %]
						</div>
					[% END %]
				[% END %]
			
			[% END %]

		[% END %]

templates/user_form2.tt.html
```diff
@@ -0,0 +1,241 @@
...
+[% INCLUDE 'error_list.tt.html' %]
+[% END %]
+
+[% IF action == 'display' %]
+
+    <!-- Start form -->
+
+    <form method="post" action="/cgi/user2.pl" enctype="multipart/form-data">
+		[% FOREACH section IN sections %]
+			[% IF user OR section.id == "user" %]
```
stephanegigandet, line 36: [% IF user OR section.id == "user" %]

--> there's no "user" variable in the template data $template_ref

in the org template, there is a similar line, but only for administrators:

[% IF admin OR section.id != "admin" %]

--> this is just to hide the "admin" section for administrators.  The "admin" variable is set in the template_data.

cgi/user2.pl
```diff
@@ -0,0 +1,361 @@
...
+					field => "requested_org",
+				}
+			]
+		};
+
+		my $team_section_ref = { id => "teams", fields => [] };
+		for (my $i = 1; $i <= 3; $i++) {
+			push @{$team_section_ref->{fields}}, {
+				field => "team_". $i,
+				label2 => sprintf(lang("team_s"), $i),
```
stephanegigandet, line 221: Why not use the "label" instead of a new "label2" ?

cgi/user2.pl
```diff
@@ -0,0 +1,361 @@
...
+		push @{$template_data_ref->{sections}}, {%$team_section_ref};
+
+		my $administrator_section_ref = { id => "administrator", fields => [] };
+		push  @{$administrator_section_ref->{fields}}, {
+			field => "org",
+		};
+		foreach my $group (@user_groups) {
+			push @{$administrator_section_ref->{fields}}, {
+				field =>  "user_group_". $group,
+				label2 =>  "user_group_" . ${group} . "_description",
```
stephanegigandet, line 234: in the org template, there is a "description" that could be used for what you put in "label2"

e.g. in org.pl it's populated automatically if we have a "[fieldid]_description" entry in the .po files:

			# Label
			$field_ref->{label} = lang($field_lang_id);
			
			# Descriptions and notes for fields
			if (lang($field_lang_id . "_description")) {
				$field_ref->{description} = lang($field_lang_id . "_description");
			}
			if (lang($field_lang_id . "_note")) {
				$field_ref->{note} = lang($field_lang_id . "_note");
			}

templates/user_form2.tt.html
```diff
@@ -0,0 +1,241 @@
...
+
+                <p>[% lang("if_you_work_for_a_producer") %]</p>
+                <p>[% lang("producers_platform_description_long") %]</p>
+
+                [% IF accepted_organization.defined %]
+                    <p>[% pro_account_org %]</p>
+                [% ELSIF product_type.defined && product_type == "food" %]
+                    [% FOREACH field IN section.fields %]
+                        [% IF field.type == 'checkbox' %]
+                            <input type="checkbox" id="[% field.field %]" name="[% field.field %]" [% IF pro_checked == 'on' %]checked="checked"[% END %] />
```
stephanegigandet, line 74: Here you could do it like in the org template:

						<label for="[% field.field %]">
							<input type="checkbox" id="[% field.field %]" name="[% field.field %]" [% IF field.value == 'on' %]checked="checked"[% END %] />
							[% field.label %]
						</label>

cgi/user2.pl
```diff
@@ -0,0 +1,361 @@
...
+		$pro_checked = "on";
+	}
+
+	$template_data_ref->{product_type} = $options{product_type};
+	$template_data_ref->{pro_checked} = $pro_checked;
+	my $requested_org_ref = retrieve_org($user_ref->{requested_org});
+	$template_data_ref->{requested_org_ref} = $requested_org_ref;
+	$template_data_ref-> {org_name} = sprintf(lang("add_user_existing_org"), org_name($requested_org_ref));
+	$template_data_ref->{teams_flag} = not ((defined $server_options{private_products}) and ($server_options{private_products}));
+	$template_data_ref->{admin_flag} = $admin;
```
stephanegigandet, line 290: There is already a "admin" variable that is automatically set by process_template():

sub process_template($$$) {
	
	my $template_filename = shift;
	my $template_data_ref = shift;
	my $result_content_ref = shift;
	
	# Add functions and values that are passed to all templates

	$template_data_ref->{product_type} = $options{product_type};
	$template_data_ref->{admin} = $admin;

cgi/user2.pl
```diff
@@ -0,0 +1,361 @@
...
+	my $pro_checked = '';
+
+	# Check the "pro account" checkbox for register screen on the producers platform
+
+	if (((defined $user_ref->{pro}) and ($user_ref->{pro}))
+		or ((defined $server_options{producers_platform}) and ($type eq "add"))) {
+		$pro_checked = "on";
+	}
+
+	$template_data_ref->{product_type} = $options{product_type};
```
stephanegigandet, line 284: this is already set by process_template()

stephanegigandet:
Hello @roshnaeem !

Thanks a lot for the PR. I tested it and it is working as expected in all the cases that I could think of.

I think the template code can be made much more generic though, with less hardcoded values and special cases for each section.

Here there is a lot of logic that is now in the template. We should keep the logic in the Perl code, and put the result of the logic in the template_data (the different sections and the different fields for each section). Then the template can decide how to display them (e.g. whether to use a table, or fieldsets etc.).

That way, if we had a new section, we should not have to modify the template.

Also we should try to make the template as close as possible to the template used for the org.pl form. At some point we could even see if we could have one "form" template that we could use for all similar forms.

templates/user_form2.tt.html
roshnaeem: Actually in `user.pl` the msgctxt used from `.po` files have different names, they are not following a rule like in `org.pl`. `org.pl` have msgctxt ending with `_note` or `_description`. But they are different for each section in `user.pl`, so we can not add them in the start here for every section. That's why I have used `foreach` separately for each section so things don't get mixed up.

templates/user_form2.tt.html
roshnaeem: `user` is the section id for the first section. every section has a different unique ID. These IDs are helping to distinguish the sections from each other.

cgi/user2.pl
roshnaeem: We are setting the `label` equal to `lang($field_lang_id)` if it is defined. I checked setting it to `label` instead of `label2`, but it applies `lang` function twice on it resulting in bugs.

cgi/user2.pl
roshnaeem: yes you are right, we can use the description here instead of `label2` in the administrator section.

templates/user_form2.tt.html
```diff
@@ -0,0 +1,241 @@
...
+                        [% ELSE %]
+                            <label for="[% field.field %]">  [% lang(field.label2) %] [% lang('username_info')  %]</label>
+                            <input type="text" id="[% field.field %]" name="[% field.field %]" value="[% field.value %]" autocomplete="[% field.field %]" onkeyup="normalize_string_value(this)" />
+                        [% END %]
+					[% ELSIF field.type == 'text' %]
+                        <label for="[% field.field %]">[% field.label %]</label>
+                        <input type="text" id="[% field.field %]" name="[% field.field %]" value="[% field.value %]" autocomplete="[% field.field %]"/>
+					[% ELSIF field.type == 'password' %]
+						<label for="[% field.field %]">[% lang(field.label2) %]</label>
+						<input type="password" id="[% field.field %]" name="[% field.field %]" value="[% user_password_field %]" autocomplete="new-password"/>
```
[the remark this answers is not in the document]
roshnaeem: ` "\n<tr><td>$Lang{password}{$lang}</td><td>"
	. password_field(-name=>'password', -value=>$user_ref->{password}, -autocomplete=>'new-password', -override=>1) . "</td></tr>"
	. "\n<tr><td>$Lang{password_confirm}{$lang}</td><td>"
	. password_field(-name=>'confirm_password', -value=>$user_ref->{password}, -autocomplete=>'new-password', -override=>1) . "</td></tr>"`

This is the code snippet of `Users.pm`, here for both `password` and `confirm_password` value is same. That's why I stored it in one variable and used it for both fields. 

But I guess it can be replaced.

templates/user_form2.tt.html
roshnaeem: The issue here was that in `users.pm` it is used like this. The label is associated with a hidden field. I first tried it like this, but it was causing issues of different names I guess.  So I tried to stick to the `Users.pm` to avoid errors.

Here is the code snippet of `Users.pm`:

`"<input type=\"checkbox\" id=\"pro\" name=\"pro\" $pro_checked>"
		. "<label for=\"pro\">"
		. "<input type=\"hidden\" name=\"pro_checkbox\" value=\"1\"> "
		. lang("this_is_a_pro_account") . "</label>"`

cgi/user2.pl
roshnaeem: Alright, `$admin` was causing me a lot of issues, and I did not where it was being set. That is why I changed it to `$admin_flag`. Let me check it again how can I fix it.

cgi/user2.pl
roshnaeem: Let me check this as well. Sorry, I did not explore `process_template()` which I should have.

templates/user_form2.tt.html
stephanegigandet: For the fields where the "_note" or "_description" does not work, in user2.pl, you could still assign the value in the user2.pl file:

				{
					field => "requested_org",
					label => lang("producer_or_brand"),
				}

roshnaeem:
> Hello @roshnaeem !
> 
> Thanks a lot for the PR. I tested it and it is working as expected in all the cases that I could think of.
> 
> I think the template code can be made much more generic though, with less hardcoded values and special cases for each section.
> 
> Here there is a lot of logic that is now in the template. We should keep the logic in the Perl code, and put the result of the logic in the template_data (the different sections and the different fields for each section). Then the template can decide how to display them (e.g. whether to use a table, or fieldsets etc.).
> 
> That way, if we had a new section, we should not have to modify the template.
> 
> Also we should try to make the template as close as possible to the template used for the org.pl form. At some point we could even see if we could have one "form" template that we could use for all similar forms.

Hello Stephane, I have updated the files and incorporated all changes, Please have a look.

stephanegigandet (approved):
Thanks Rosheen!

It is a huge improvement to what we had before, it's great that the form is now responsive and that a lot of HTML code has been moved to the template.

There are a few things that I think could be improved, but it's something that we can do later, or that maybe I can do on top of your PR.

So let's merge this, deploy it and try it out (as you made a separate user2.pl script it will be easy to test / compare with what we had before).

Thank you!

PR #5169, opened 2021-04-23
Fixes #4968: Chrome autofills username field with e-mail address

### LINKED ISSUE #4968: Chrome autofills username field with e-mail address
A lot of users are created with their email address as the username, it's probably because Chrome can autofill the username field with the email address:

image

The signup form has this:

<tr><td>Nom</td><td><input type="text" name="name"  size="80" autocomplete="name" id="name" /></td></tr>
--
  | <tr><td>Adresse e-mail</td><td><input type="text" name="email"  size="80" autocomplete="email" type="email" /></td></tr>
  | <tr><td>Nom d'utilisateur<br/><span class="info">(lettres non accentuées, chiffres et/ou tirets)</span></td><td><input type="text" name="userid"  size="40" autocomplete="username" id="userid" onkeyup="normalize_string_value(this)" /></td></tr>

I'm not sure how to handle that. If we remove autocomplete="username", then the browser won't remember it. If we do put it, we have the risk that the browser autofills with an email address.

hstauff:
**Checklist:**

* [ ] Code is well documented
* [ ] Include unit tests for new functionality
* [ ] Code passes Travis builds in your branch
* [ ] If you have multiple commits please combine them into one commit by squashing them.
* [ ] Read and understood the [contribution guidelines](https://github.com/openfoodfacts/openfoodfacts-server/blob/master/CONTRIBUTING.md)

**Description:**

Describe the changes made and why they were made instead of how they were made.

**Related issues and discussion:** #4968

### DIFF
--- a/lib/ProductOpener/Users.pm
+++ b/lib/ProductOpener/Users.pm
@@ -210,12 +210,18 @@ sub display_user_form($$$) {
 	. textfield(-name=>'email', -value=>$user_ref->{email}, -size=>80, -autocomplete=>'email', -type=>'email', -override=>1) . "</td></tr>"
 	. "\n<tr><td>$Lang{username}{$lang}<br/><span class=\"info\">" . (($type eq 'edit') ? '': $Lang{username_info}{$lang}) . "</span></td><td>"
 	. (($type eq 'edit') ? $user_ref->{userid} :
-		( textfield(-id=>'userid', -name=>'userid', -value=>$user_ref->{userid}, -size=>40, -onkeyup=>"normalize_string_value(this)", -autocomplete=>'username' -oninput='this.value = this.value.replace(/[^a-zA-Z0-9 -]/, '')'))) . "</td></tr>"
+		( textfield(-id=>'userid', -name=>'userid', -value=>$user_ref->{userid}, -size=>40, -onkeyup=>"normalize_string_value(this)", -autocomplete=>'username'))) . "</td></tr>"
 	. "\n<tr><td>$Lang{password}{$lang}</td><td>"
 	. password_field(-name=>'password', -value=>$user_ref->{password}, -autocomplete=>'new-password', -override=>1) . "</td></tr>"
 	. "\n<tr><td>$Lang{password_confirm}{$lang}</td><td>"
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Users.pm
```diff
@@ -210,7 +210,7 @@ sub display_user_form($$$) {
 	. textfield(-name=>'email', -value=>$user_ref->{email}, -size=>80, -autocomplete=>'email', -type=>'email', -override=>1) . "</td></tr>"
 	. "\n<tr><td>$Lang{username}{$lang}<br/><span class=\"info\">" . (($type eq 'edit') ? '': $Lang{username_info}{$lang}) . "</span></td><td>"
 	. (($type eq 'edit') ? $user_ref->{userid} :
-		( textfield(-id=>'userid', -name=>'userid', -value=>$user_ref->{userid}, -size=>40, -onkeyup=>"normalize_string_value(this)", -autocomplete=>'username'))) . "</td></tr>"
+		( textfield(-id=>'userid', -name=>'userid', -value=>$user_ref->{userid}, -size=>40, -onkeyup=>"normalize_string_value(this)", -autocomplete=>'username' -oninput='this.value = this.value.replace(/[^a-zA-Z0-9 -]/, '')'))) . "</td></tr>"
```
stephanegigandet, line 213: The issue is that Chrome fills this field with an email address. (see https://github.com/openfoodfacts/openfoodfacts-server/issues/4968 )
so we get values like someusername@gmail.com for instance. We want to reject those values entirely, here we would just remove the @ character.
Maybe a solution is to check if the value contains @, and if it does, we completely delete the value.

VaiTon (commented):
What happens then? Does the user get warned about it? Should we add an info in the signup form?

stephanegigandet:
> What happens then? Does the user get warned about it? Should we add an info in the signup form?

I think it's best to fix it in the form itself, before it is submitted. We populate the field with js, the best place to fix it is with js.

stephanegigandet:
I'll make a browser side fix.

stephanegigandet:
fixed on the browser side, closing this PR. Thanks for your work @hstauff !

PR #5189, opened 2021-05-01
Soybeans and reconstituted crisps

moon-rabbitOFF:


svensven:
Aren't mung and soy beans different things? Especially if one is a major allergen and the other isn't.

moon-rabbitOFF:
> Aren't mung and soy beans different things? Especially if one is a major allergen and the other isn't.

Yes they are. Mung sprouts are erroneously called "soy sprouts" by brands and consumers. Real soy sprouts are not edible raw, and are not commercialized. In France, the mention soy sprout is not authorized on the packaging anymore, but in other countries it is still the case. http://leparticulier.lefigaro.fr/jcms/p1_1600765/haricots-mungo-et-soja-ne-doivent-pas-etre-confondus

stephanegigandet (approved):
Thank you!

PR #5196, opened 2021-05-03
Capitalize 1st letter of all lowercase product names, ingredients etc.

stephanegigandet:
Some CodeOnline products have all lowercase product names


hangy:
Lowercase product or brand names can be a specific decision by the manufacturer/brand. Also, at least in German, uppercasing all ingredients might look weird/be incorrect in cases like "Soja, glutenfreies Weizen" vs. "Soja, Glutenfreies Weizen".

stephanegigandet:
@hangy : for ingredients, it would uppercase the 1st letter of the whole ingredient lists, not each ingredient. Would that be incorrect in some cases in German? e.g. if there's only one ingredient, "Glutenfreies Weizen" would be incorrect? If so I can make the rule only for French. 

For product names, it applies only to the complete product name, with no capital anywhere.

Brands is a bit more tricky, as a few brands are indeed all lowercase. Those we probably can fix once we implement a brands taxonomy.

hangy:
> @hangy : for ingredients, it would uppercase the 1st letter of the whole ingredient lists, not each ingredient. Would that be incorrect in some cases in German? e.g. if there's only one ingredient, "Glutenfreies Weizen" would be incorrect? If so I can make the rule only for French.

That's good. :)

stephanegigandet:
@hangy : so can I merge this as is, or should I apply it only for French?

hangy (approved):

PR #5197, opened 2021-05-03
feat: Improve dev onboarding

teolemon:
Improve dev onboarding

image

### DIFF
--- a/docker/automatic-dev-environment-setup.sh
+++ b/docker/automatic-dev-environment-setup.sh
@@ -11,11 +11,11 @@ brew cask install docker
 # opening docker for the first time in GUI mode to grant it priviledges
 open docker
 cd ./openfoodfacts-server/docker/
-echo "🥫 Building visual assets…"
-./build_npm.sh
 echo "🥫 Building Product Opener…"
 echo "🥫 Your ventilator is probably going to start"
 ./build_dev.sh
+echo "🥫 Building visual assets…"
+./build_npm.sh
 echo "🥫 TODO: describe me…"
 echo "🥫 This phase is going to be very long, up to an hour. Look for the increasing number snapshot min"
 ./start_dev.sh
--- a/docker/automatic-dev-environment-setup.sh
+++ b/docker/automatic-dev-environment-setup.sh
@@ -1,24 +1,24 @@
 #if mac:
 # check if git is installed on mac
-echo "Checking git is installed…"
+echo "🥫 Checking git is installed…"
 git --version
-echo "Cloning Open Food Facts web server - product-opener…"
+echo "🥫 Cloning Open Food Facts web server - product-opener…"
 git clone https://github.com/openfoodfacts/openfoodfacts-server.git
-echo "Installing Docker… you may have to open it manually to grant permissions"
+echo "🥫 Installing Docker… you may have to open it manually to grant permissions"
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 brew cask install docker
 # opening docker for the first time in GUI mode to grant it priviledges
 open docker
 cd ./openfoodfacts-server/docker/
-echo "Building visual assets…"
+echo "🥫 Building visual assets…"
 ./build_npm.sh
-echo "Building Product Opener…"
-echo "Your ventilator is probably going to start"
+echo "🥫 Building Product Opener…"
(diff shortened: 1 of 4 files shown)


hangy:
Why do you want to do this in `build_dev.sh`? That script is something you may run multiple times a day, if you change something that modifies the local image. It's not a one-time setup helper.

IMHO, the instructions in https://github.com/openfoodfacts/openfoodfacts-server/blob/9af7d8767d3e2a84b055f96b9fa18dc3a45988e6/installation/dev-environment-quick-start-guide.md contain all of the information that's required for the start. Also, as far as I'm confirmed, it's not a good idea to have this script run something like `curl -fsSL https://get.docker.com -o get-docker.sh` transparently, as one might want to use the distribution-provided Docker package, ie. from `pacman`, instead.

svensven:
or at least have it as a separate install_docker script, with a big warning it may not cover all use cases.

teolemon:
noted @hangy @svensven 
I'd like eventually a set-up-dev-machine.sh script for lazy ppl :-)

docker/install-full-database.sh
```diff
@@ -0,0 +1,13 @@
+echo "🥫 Downloading the full MongoDB dump…"
+wget https://static.openfoodfacts.org/data/openfoodfacts-mongodbdump.tar.gz
+echo "🥫 Copying the dump to Docker…"
+docker cp openfoodfacts-mongodbdump.tar.gz docker_mongodb_1:/data/db
+echo "🥫 Opening a shell within the Mongo docker…"
+docker exec -it docker_mongodb_1 bash
```
hangy, line 6: How does it work? I don't quite understand how the following script is run inside the container.

teolemon: install-full-database.sh is meant to be run outside
this instruction will be run to open a shell inside
https://www.oreilly.com/library/view/hands-on-devops-with/9781789138054/a9a81868-1652-4e57-a064-0c64de55eefc.xhtml

docker/automatic-dev-environment-setup.sh
```diff
@@ -0,0 +1,22 @@
+#if mac:
+# check if git is installed on mac
+echo "🥫 Welcome to the Open Food Facts dev environment setup"
+echo "🥫 Checking git is installed…"
+git --version
+echo "🥫 Cloning Open Food Facts web server - product-opener…"
+git clone https://github.com/openfoodfacts/openfoodfacts-server.git
+echo "🥫 Installing Docker… you may have to open it manually to grant permissions"
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+brew cask install docker
```
ocervell, line 10: not everyone uses `brew cask`, I prefer the Desktop install of Docker on Mac for instance.

docker/automatic-dev-environment-setup.sh
```diff
@@ -0,0 +1,22 @@
...
+# opening docker for the first time in GUI mode to grant it priviledges
+open docker
```
ocervell, line 12: not a valid CMD :)

docker/automatic-dev-environment-setup.sh
```diff
@@ -0,0 +1,22 @@
...
+cd ./openfoodfacts-server/docker/
+echo "🥫 Building Product Opener…"
+echo "🥫 Your ventilator is probably going to start"
+./build_dev.sh
+echo "🥫 Building visual assets…"
+./build_npm.sh
+echo "🥫 Populating database…"
+docker-compose -f docker-compose.yml -f docker-compose.dev.yml exec backend bash /opt/scripts/import_sample_data.sh
+echo "🥫 You should be able to access your local install of Open Food Facts at http://0.0.0.0…"
+echo "🥫 You have around 200 test products. Please run ./install-full-database.sh if you want a full dump."
```
ocervell, line 22: All those steps will be replaced by a few commands in the Makefile, documented in the existing guide, basically `make dev` should build a complete dev env.

docker/install-full-database.sh
ocervell: This will be automated as part of the `restructure-docker` branch in the Makefile. Note to myself: `wget` doesn't work on Powershell.

ocervell (commented):
Not sure if this PR is necessary, since we're updating a bit how the dev setup will be made.

stephanegigandet:
@teolemon I believe this is not needed anymore, can we close it?

PR #5201, opened 2021-05-04
[WIP] Prototype - deploy-branches-with-rsync.yml

teolemon:
### Proposed approach 1
Taking some inspiration from https://zellwk.com/blog/github-actions-deploy/

- [ ] Would such an approach work safety wise within a ProxMox container ? 
- [ ] Do we restrict automated deployments to a `test-deploy` branch, before eventually deploying every commit ?

Another approach since docker images seem to be published by another action would be to install them from the server. 
GitHub would only ping the install script to trigger an install. 

Any thoughts ?

### Proposed approach 2
Having GitHub notify the server, which then wget an install script on github
image

### DIFF
--- a/.github/workflows/deploy-branches-with-rsync.yml
+++ b/.github/workflows/deploy-branches-with-rsync.yml
@@ -1,17 +1,174 @@
 name: deploy
 on:
   push:
-  schedule:
-    - cron: '0 0 * * *' # Everyday at 12am
+    branches:
+      - deploy-with-rsync
+
 jobs:
-  build-and-deploy:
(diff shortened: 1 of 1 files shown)


stephanegigandet:
I think it would be best to define the use case / problem before selecting a technical solution.

e.g. if the use case is to have an up-to-date instance running on .net, then it's much simpler to have a cron on the server to fetch the code and deploy it. It could be done every 30 mins for instance.

If the use case is to provide a test environment for people who want to test a specific branch, then it would also be much simpler to do it server side as well.

PR #5204, opened 2021-05-05
Main countries: small fixes for the best remap export

stephanegigandet:
We now compute main countries when saving products.


svensven (approved):

lib/ProductOpener/MainCountries.pm
```diff
@@ -212,7 +216,12 @@ sub compute_main_countries($) {
 					if ($product_name_in_country_language < 1) {
 						defined $product_ref->{misc_tags} or $product_ref->{misc_tags} = [];
 						push @{$product_ref->{misc_tags}}, "en:main-countries-$cc-product-name-not-in-country-language";
-					}					
+					}
+					
+					if ($ingredients_text_in_country_language < 1) {
+						defined $product_ref->{misc_tags} or $product_ref->{misc_tags} = [];
```
VaiTon, line 222: Why don't we put this line before the if and remove it from every if?

PR #5222, opened 2021-05-11
Added alcoholic beverage categories

aleene:
Purpose is to describe alcoholic beverages that have been mixed. These can be based on other liquors, like wine, beer, etc.

The idea is to have a place where we can pit the Bacardi Breezers, Smirnoff ice and other premixed (designer) drinks.

This category is not very well defined on Wikipedia, but is related to [Ready to Drink](https://en.wikipedia.org/wiki/Ready_to_drink), [Alcopop](https://en.wikipedia.org/wiki/Alcopop), [Wine cooler](https://en.wikipedia.org/wiki/Wine_cooler), etc.

There might be an overlap with flavoured drinks, like flavoured beers, vodka based beers, etc.

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -1076,9 +1076,44 @@ pnns_group_2:en:Alcoholic beverages
 nova:en:3
 
 <en:Alcoholic beverages
+en:Premixed alcoholic beverages
+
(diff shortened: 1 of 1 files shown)


teolemon (approved):
Looks good to me

PR #5234, opened 2021-05-17
small fix for data quality food facet

stephanegigandet:
fruits vegetables estimate is a computed value, it should not count for empty / non-empty values

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -594,12 +594,14 @@ sub check_nutrition_data($) {
 
 					push @{$product_ref->{data_quality_errors_tags}}, "en:nutrition-value-over-1000-$nid2";
 				}
-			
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -594,12 +594,14 @@ sub check_nutrition_data($) {
 
 					push @{$product_ref->{data_quality_errors_tags}}, "en:nutrition-value-over-1000-$nid2";
 				}
-			
-				if ($product_ref->{nutriments}{$nid} == 0) {
-					$nid_zero++;
-				}
-				else {
-					$nid_non_zero++;
+
```
teolemon, line 597:
```suggestion
		# fruits vegetables estimate is a computed value, it should not count for empty / non-empty values
```

teolemon (approved):

PR #5236, opened 2021-05-18
Imports: do not overwrite product name with generic name

stephanegigandet:
For imports, when we had a generic name but not a product name, we used to copy the generic name to the product name (because in most cases the generic name was also the product name).

But with the current GS1 imports, it's very frequent that the product name is in fact an abbreviated product name, so I made an organization checkbox to assign the GS1 product name to the abbreviated product name. That means we don't have a product name for many products.

So this:
- remove the automatic generic name copy to product name when it does not exist
- changes the display of products to display the generic name when we don't have a product name

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -4843,6 +4843,10 @@ sub search_and_display_products($$$$$) {
 		"code" => 1,
 		"product_name" => 1,
 		"product_name_$lc" => 1,
+		"generic_name" => 1,
+		"generic_name_$lc" => 1,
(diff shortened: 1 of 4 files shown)


teolemon (approved):

teolemon:
#   Failed test at t/import.t line 268.
#          $got->{product_name_en} = Does not exist
#     $expected->{product_name_en} = '-'
# {
#   'generic_name_en' => '-',
#   'labels' => '',
#   'lc' => 'en',
#   'origins' => '',
#   'warning_en' => '-'
# }
Use of uninitialized value in substitut

@stephanegigandet

PR #5243, opened 2021-05-20
Deploy new templated user form

stephanegigandet:
This is to deploy the new user form that has been templated by @roshnaeem .  The new form was available at /cgi/user2.pl for testing, this now moves it in place of the real /cgi/user.pl form.


teolemon:
CC-ing Vaiton and Aleene. We can't release Android right now, so please don't break the "Thank you page" after login

teolemon:
https://github.com/openfoodfacts/openfoodfacts-androidapp/blob/1b3956398d8f81838d7a9f2fd67a0a2b53f5c906/app/src/main/java/openfoodfacts/github/scrachx/openfood/features/HomeFragment.kt#L122

stephanegigandet:
Well it doesnt' seem that the Android app is checking anything except that there is some valid HTML returned. (which is a weird test). Anyway, I'm noting that we should verify logging in with the Android app still works when we deploy this.

teolemon:
I can't find the proper file right now, but we're checking for a string if I remember right.

templates/user_form.tt.html
```diff
@@ -7,32 +42,101 @@
...
+
+            [% IF section.note %]
+                <p >[% section.note %]</p>
+            [% END %]
+
+            [% FOREACH field IN section.fields %]
+                [% IF field.field == 'userid' %]
+                    [% IF type == 'edit' %]
+                        <label for="[% field.field %]"> [% lang(field.label) %]</label>
+                        <p> [% field.value %] </p>
```
CharlesNepote, line 64:
```suggestion
                        <p id="userid">[% field.value %]</p>
```

CharlesNepote, line 64:
```suggestion
                        <p id="userid"><a href="/editor/[% field.value %]">[% field.value %]</a></p>
```

CharlesNepote (approved):

templates/user_form.tt.html
stephanegigandet: in fact there's already an hidden field with the userid, so maybe I can add id="userid" to it, to make it consistent with the other fields.

        [% IF userid %]
            <input type="hidden" name="userid" value="[% userid %]" />
            <input type="submit" name=".submit" class="button" value="[% lang("save") %]" />
        [% ELSE %]

stephanegigandet:
tested in prod :) both the ios and android app can login.

roshnaeem:
Thank you so much @stephanegigandet for working on it and simplifying it even more. Great work!! :100:

Issue #5258, opened 2021-05-26
Allow installing an instance of the producers platform through Docker


stephanegigandet:
In production, we have 2 instances of Product Opener running:

- 1 for the public database https://world.openfoodfacts.org
- 1 for the producers platform https://world.pro.openfoodfacts.org

The code is the same, but the producers platform has a few differences:

- there is a different root path (in production: /srv/off-pro/ instead of /srv/off)
- in Config2.pm:

# server constants
$server_domain = "pro.openfoodfacts.org";

# server paths
$www_root = "/srv/off-pro/html";
$data_root = "/srv/off-pro";
$mongodb = "off-pro";

[..]

%server_options = (

        cookie_domain => "openfoodfacts.org",   # if not set, default to $server_domain
        private_products => 1,  # Make products visible only to the owner
        producers_platform => 1,        # Enable special features for producers
        minion_backend => {'Pg' => 'postgresql://off:Zo4pByAt2tt@localhost:5432/minion'},
        export_servers => { public => "off" },
        minion_local_queue => "pro.openfoodfacts.org",
        minion_export_queue => "openfoodfacts.org",
        export_data_root => "/srv/off",
        export_www_path => "/srv/off/html",

There are a few differences in directory structure:

the products and html/products directories have one extra level for the owner (e.g. org-nestle-france) of the products

In addition, there are also 2 minion server queues that are running, one for the producers platform (e.g. to import Excel file from producers, and to export data to the public platform), and one for the public platform (to import data from the producers platform).

The minion servers are run with systemctl:

root@off1:/srv/off-pro/products# cat /etc/systemd/system/minion-off-pro.service 
[Unit]
Description=off-pro minion workers
After=postgresql.service

[Service]
Type=simple
#ExecStart=/srv/off-pro/scripts/minion.off-pro.sh
User=off
WorkingDirectory=/srv/off-pro/scripts
Environment="PERL5LIB=."
ExecStart=/srv/off-pro/scripts/minion_producers.pl minion worker -m production -q pro.openfoodfacts.org
KillMode=process

[Install]
WantedBy=multi-user.target

root@off1:/srv/off-pro/products# 
root@off1:/srv/off-pro/products# 
root@off1:/srv/off-pro/products# cat /etc/systemd/system/minion-off.service 
[Unit]
Description=off minion workers
After=postgresql.service

[Service]
Type=simple
#ExecStart=/srv/off-pro/scripts/minion.off-pro.sh
User=off
[...]

PR #5262, opened 2021-05-27
Changes to make the producers platform work in Docker

stephanegigandet:
Everything is not working yet, but this makes a few things work.
Tracking bug: #5258

### DIFF
--- a/docker/docker-compose.pro.dev.yml
+++ b/docker/docker-compose.pro.dev.yml
@@ -1,0 +1,42 @@
+﻿version: "3.7"
+services:
+  backend:
+    image: productopener-backend-pro-dev
+    build:
+      context: ../
+      dockerfile: ./docker/backend/Dockerfile
+      target: runnable
+    volumes:
+      - type: bind
+        source: ./backend-pro-dev/scripts/
+        target: /opt/scripts/
+      - type: bind
+        source: ./backend-pro-dev/conf/Config.pm
+        target: /opt/product-opener/lib/ProductOpener/Config.pm
+      - type: bind
+        source: ./backend-pro-dev/conf/Config2.pm
+        target: /opt/product-opener/lib/ProductOpener/Config2.pm
+      - type: bind
+        source: ./backend-pro-dev/conf/log.conf
+        target: /mnt/podata/log.conf
+      - type: bind
+        source: ./backend-pro-dev/conf/minion_log.conf
+        target: /mnt/podata/minion_log.conf
+      - type: bind
+        source: ./backend-pro-dev/conf/apache.conf
+        target: /etc/apache2/sites-enabled/product-opener.conf
+      - type: bind
+        source: ./backend-pro-dev/conf/po-foreground.sh
+        target: /usr/local/bin/po-foreground.sh
+      - type: bind
+        source: ../
+        target: /opt/product-opener/
+      - ./logs/apache2-pro:/var/log/apache2
+      - ./logs/apache2-pro:/mnt/podata/logs
+      - ./logs/apache2-pro:/var/log/httpd
+  frontend:
+    image: nginx:stable-alpine
+    volumes:
+      - type: bind
+        source: ../
+        target: /opt/product-opener/
--- a/docker/docker-compose.pro.dev.yml
+++ b/docker/docker-compose.pro.dev.yml
@@ -1,10 +1,10 @@
 ﻿version: "3.7"
 services:
-  backend:
+  backend-pro:
     image: productopener-backend-pro-dev
     build:
       context: ../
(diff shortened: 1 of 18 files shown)


docker/docker-compose.yml
```diff
@@ -41,6 +41,9 @@ services:
       - type: bind
```
roshnaeem, line 41: I had to replace `Config2.pm` with `Config2_pro.pm` at `line 39, 40` to avoid errors.

hangy: I think it should probably be run in a 2nd backend container, so that both versions can be tested.

stephanegigandet: Right, Config2.pm is to run an instance that looks like the public Open Food Facts platform, and Config2_pro.pm is to run an instance that looks like the producers platform on https://world.pro.openfoodfacts.org

There's probably a way to change the startup scripts to make it possible to specify which config file should be used.

stephanegigandet:
I added a separate config file for the producers platform, and scripts to build and run the backend container with the configuration of the producers platform:
./build_pro_dev.sh
./start_pro_dev.sh

@hangy : ideally we would have a separate container for the public and pro instances, but it's already quite long to install 1 backend container, so that's why I'm thinking we can use the backend container for both the backend-dev and backend-pro-dev config.

docker/docker-compose.pro.dev.yml
```diff
@@ -0,0 +1,42 @@
+﻿version: "3.7"
+services:
+  backend:
```
hangy, line 3:
```suggestion
  backend-pro:
```
Let's use a different service name, so that both can run in parallel, if the user wants them to.

hangy:
I think you may need to update the frontend to contain a 2nd server entry for the "pro" subdomain, so that it can be accessed.

docker/docker-compose.pro.dev.yml
stephanegigandet: If I replace backend by backend-pro, I'm not able to build it:

version: "3.7"
services:
  backend-pro:
    image: productopener-backend-pro-dev
    build:
      context: ../
      dockerfile: ./docker/backend/Dockerfile
      target: runnable
    volumes:
      - type: bind
        source: ./backend-pro-dev/scripts/
        target: /opt/scripts/

stephane@stephane-hp:~/openfoodfacts-server/docker$ ./build_pro_dev.sh 
backend uses an image, skipping

Maybe I also need to duplicate the "backend" dir in "backend-pro"? Which means all the Apache modules etc. would need to be installed for this new backend-pro as well?

stephanegigandet:
> I think you may need to update the frontend to contain a 2nd server entry for the "pro" subdomain, so that it can be accessed.

in frontend-git/conf/nginx.conf ?

In : 

proxy_pass http://backend/cgi/display.pl?;

I'm guessing "backend" is like a host for the backend container?

So we would create a backend-pro service, and proxy pass to http://backend-pro ?

hangy:
> > I think you may need to update the frontend to contain a 2nd server entry for the "pro" subdomain, so that it can be accessed.
> 
> in frontend-git/conf/nginx.conf ?
> 
> In :
> 
> proxy_pass http://backend/cgi/display.pl?;
> 
> I'm guessing "backend" is like a host for the backend container?
> 
> So we would create a backend-pro service, and proxy pass to http://backend-pro ?

Yup, inside Docker Swarm, the service name "backend" acts as a (load balanced) DNS alias for the containers.

stephanegigandet:
> > I think you may need to update the frontend to contain a 2nd server entry for the "pro" subdomain, so that it can be accessed.

The issue I'm running into is that if both backend and backend-pro are in the nginx config, then it fails to start if one of the backends is not running:

frontend_1     | 2021/06/01 07:27:09 [emerg] 1#1: host not found in upstream "backend" in /etc/nginx/conf.d/default.conf:59
frontend_1     | nginx: [emerg] host not found in upstream "backend" in /etc/nginx/conf.d/default.conf:59

So I'm guessing that means I need a separate frontend-pro container too.

stephanegigandet:
> So I'm guessing that means I need a separate frontend-pro container too.

Except of course only one of them can listen on port 80.

stephanegigandet:
Trying using a variable: https://sandro-keil.de/blog/let-nginx-start-if-upstream-host-is-unavailable-or-down/

stephanegigandet:
@hangy : the last commit is working. It creates a new backend-pro container, and the frontend container nginx config dispatches the *.pro.productopener.localhost urls to it. I can run ./start_dev.sh and ./start_pro_dev.sh in parallel.

teolemon (approved):

Issue #5250, opened 2021-05-23
This product should trigger a gluten alert


teolemon:
This product should trigger a gluten alert. 
https://world.openfoodfacts.org/product/3367651008869/pause-pasta-poulet-roti-mimolette-daunat

![image](https://user-images.githubusercontent.com/1689815/119271631-a2aaef80-bc02-11eb-8224-1d64e858a74d.png)

teolemon:
```
<en:semolina
<en:durum wheat
# <en:Couscous
en:durum wheat semolina, durum semolina
da:durumhvede semulje, durumhvedesemulje
de:Hartweizengrieß, Hartweizengriess, Hartweizengries
el:σιμιγδάλισκληρού σιταριού
es:sémola de trigo duro, sémola de trigo candeal
fi:durumvehnäsuurimo
fr:semoule de blé dur, semoule de blé dure
hr:krupica durum pšenice
hu:durumbúzadara, durum búzadara
it:semola di grano duro
nb:durumhvete semule, durum semulegryn
nl:harde tarwe griesmeel, durum tarwegries, harde tarwegries, griesmeel van harde tarwe
pl:Kasza z pszenicy durum
pt:sêmola de trigo duro
ro:griș din grâu, griș de grău
ru:манная крупа, крупа манная
sl:pšenični zdrob durum
sv:durumvetegryn
tr:durum buğdayı ırmığı
zh:硬质粗粒小麦粉
```

teolemon:
```
en:gluten, cereals containing gluten, barley, barley malt flour, malted barley, malted barley extract, malted barley flour, kamut, rye, rye flour, spelt, speltflour, wheat, wheat flour, wheatflour, wheat semolina, oats, oat fiber
ar:غلوتين
bg:глутен, ечемик, ечемичен, ечемично-малцов, пшеница, пшенично, ръж,  овес, шпелта, камут, вид едра твърда пшеница
cs:lepek, pšenice, žito, ječmen, oves, špalda, kamut
da:gluten, hvede, durumhvede, rug, byg, havre, spelt, kamut, bygmalt, bygmaltmel, bygmaltekstrakt, havremel, hvedemel, durumhvedemel, fuldkornshvede, fuldkornshvedemel, havregryn, ruggryn, havreflager, speltflager, hvedeklid, hvedestivelse, hvedegluten, bygmalteddike, kornprodukter som indeholder gluten
[...]

Issue #3297, opened 2020-04-20
Use ingredients taxonomy to add allergens from ingredients list


stephanegigandet:
### What

- We currently use synonyms in the allergens taxonomy to detect allergens in the ingredients list.
- We could add allergens properties in the ingredients taxonomy so that we can find more allergens, without having to list tons of synonyms in the allergens taxonomy.
- e.g. "blancs d'oeufs en poudre" should trigger the "eggs" allergen.
- https://fr.openfoodfacts.org/produit/3256226086497/sables-nappes-chocolat-au-lait-sans-gluten-u

### Part of
- #5706

aleene:
@stephanegigandet There are already a lot of properties `allergens:en:` in the ingredients taxonomy. Are these currently used?
- if not, can we check the ingredients first before reverting to allergens.txt
- if yes, can we deprecate allergens.txt and moving entries to ingredients.txt
- and do both, so we have a safe upgrade path?

PR #5272, opened 2021-06-01
Detect allergens using the ingredients taxonomy and ingredients analysis

stephanegigandet:
We currently detect allergens using the allergens taxonomy. This adds detection through the ingredients taxonomy, so that we can catch more ingredients containing allergens. e.g. in the allergens taxonomy, we have some common ingredients containing gluten listed as synonyms of gluten, but we don't have all possible ingredients containing gluten (e.g. "durum wheat semolina").

On the other hand, thanks to the ingredients taxonomy, we know that "durum wheat semolina" is a child of "wheat", and that "wheat" has the "allergens:en: en:gluten" property.

Main issue: #3297

Will also fix: #5250

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -5216,10 +5216,11 @@ sub replace_allergen_between_separators($$$$$$) {
 }
 
 
-
 =head2 detect_allergens_from_ingredients ( $product_ref )
 
-Detects allergens from the ingredients extracted from the ingredients text.
+Detects allergens from the ingredients extracted from the ingredients text,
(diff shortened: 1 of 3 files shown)


aleene:
When this is committed, we neded to double check to see if all entries have been copied into the ingredients.

aleene (approved):

stephanegigandet:
> When this is committed, we neded to double check to see if all entries have been copied into the ingredients.

Yes. We'll also need to change a few properties from allergens:en: to something like likely_allergens:en: for compound ingredients that are very likely to contain an allergen (e.g. "pizza crust" -> gluten) but may also not contain it.

Issue #5261, opened 2021-05-27
Improvements to better onboard designers and front-end devs


CharlesNepote:
I tried to list here what could help designers and front-end devs to suggest and work on small site improvements.

- [ ] for each template, add the name of the template in a comment at the beginning (with path?)
- [ ] add ids on each different meaningful part of the page, to let people understand which part of the template they have to work on and to ease manipulation and design of each meaningful part; examples of meaningful part which don't have an id:
  - [ ] left column
  - [ ] main logo
  - [ ] tagline
  - [ ] sign-in block
  - [ ] "not registered yet?"
  - [ ] add a product
  - [ ] Donate to support our work block
- [ ] use a different name for templates related to the whole page and templates related to parts of pages (the later could have an "inc" prefix) 
- [ ] add perl comments to understand where to find content; example: what is [% content_ref %] and where is it produced?
- [ ] use a tool for the indentation of the templates' code (at least the templates related to whole pages) 
- [ ] try to avoid HTML code inside perl template variables; example: `[% tagline %]` which contains `<p>blabla</p>`; other examples:
  - [ ] [% blocks %]
  - [ ] [% h1_title %]
  - [ ] [% join_us_on_slack %]
  - [ ] [% lang('footer_and_the_facebook_group') %]
  - [ ] [% lang('footer_follow_us') %]

roshnaeem:
Thank you @Charles Nepote. These are really good suggestions. We are already working on a few of them. Hopefully, we will soon be able to incorporate other suggestions as well.

roshnaeem:
Thank you @CharlesNepote. These are really good suggestions. We are already working on a few of them. Hopefully, we will soon be able to incorporate other suggestions as well.

stephanegigandet:
> * [ ]  for each template, add its name in an HTML comment at the beginning of the template (with path?)
> *

This is a great idea. I think it's good to add the full path, so that designers can quickly find the files in the github directory.

As you suggested, we can put this:

```
<!-- start templates/[% template.name %] -->
[..]
<!-- end templates/[% template.name %] -->
```

at the start and end of each template, and then we get this generated:

`
`

stephanegigandet:
New templates directory structure:

- templates > web > common > site_layout.tt.html : main template for all the navigation etc. of all pages
- templates > web > common > includes : parts of multiple pages
- templates > web > pages > product > product_page.tt.html : main page content
- templates > web > pages > product >includes > .. : parts of the page

PR #5274, opened 2021-06-01
Structure the templates directory

stephanegigandet:
This starts to put a little bit of structure in the templates directory, as suggested by @CharlesNepote in #5261

- As we now have emails templates in templates/emails/ , I moved all web pages related templates to templates/web/
- The not intuitively display_new template is now renamed to web/site_layout.html (that's the template that has all the menus, side column and footer)
- in templates/web there are a pages/ and a common/ directory:
- templates/web/common/ is for things that are common to several pages (e.g. the donate_banner that is included by site_layout)
- templates/pages contains the different types of pages, in sub directories:
- I created the dir for only one page, as an example: templates/pages/product contains all the product page related templates
- templates/pages/product/product_page.tt.html : template ending in _page is the main page template
- templates/pages/product/edit_history.tt.html etc. (not suffixed by _page) are components of the page.

If everything looks good, we can continue to move all the other templates to their respective pages, and then continue with Charles' other suggestions like adding a comment at the top of each template to indicate which part of the page is generated by which template.


templates/web/pages/product/forest_footprint_details.tt.html
```diff
@@ -0,0 +1,35 @@
+<p>
```
CharlesNepote, line 1:
```suggestion
<!-- [% template.name %] -->
<p id="forest_footprint_calculation_details">
```

[% template.name %] should print the template name.

templates/web/pages/product/forest_footprint_details.tt.html
```diff
@@ -0,0 +1,35 @@
...
+	[% ELSE %]
+		<p>Les informations sur l'emballage sont manquantes, son impact n'a pas pu être pris en compte.</p>
+	[% END %]
+	
+	<h3>Score final</h3>
+
+	<p><strong>[% lang('ecoscore_score') %][% sep %]: [% round(score) %]</strong></p>
+	<p><strong>[% lang('ecoscore_grade') %][% sep %]: [% grade FILTER ucfirst %]</strong></p>
+
+</div>
```
CharlesNepote, line 35:
```suggestion
</div>
<!-- / [% template.name %] -->
```

CharlesNepote (approved):
I would have used a different name for templates related to the whole page and templates related to parts of pages (the later could have an "inc" prefix). "inc" let you know it's a piece/component and not the whole page. But maybe _page is sufficient.

But it's already a good improvement.

I made 2 suggestions, feel free to integrate it or not.

templates/web/pages/product/forest_footprint_details.tt.html
stephanegigandet: Perfect, I didn't know [% template.name %]

stephanegigandet:
> I would have used a different name for templates related to the whole page and templates related to parts of pages (the later could have an "inc" prefix). "inc" let you know it's a piece/component and not the whole page.

I would prefer to create a specific includes directory instead of using file name prefixes. Something like templates/web/pages/products/includes/ I'll try to see how it looks.

It would also work well for the templates/web/common directory, to have the site_layout.tt.html in common/ and things like the donate_banner.tt.html in common/includes/

roshnaeem:
Thank you @stephanegigandet for working on this. I just have a small question. How are we gonna deal with `.tt.js` files. Are we gonna make new separate `js` directory/directories for them or these files will be in the same folder as their corresponding `tt.html` files?

stephanegigandet:
> How are we gonna deal with `.tt.js` files. Are we gonna make new separate `js` directory/directories for them or these files will be in the same folder as their corresponding `tt.html` files?

@roshnaeem: Maybe we can just leave them in the same folder as the corresponding tt.html file for now.

stephanegigandet:
Merging as I have other bug fixes that depend on this. We can always adjust the directory structure later if needed.

PR #5280, opened 2021-06-02
User form template update and fix of email address autocompleted in username field

### LINKED ISSUE #4968: Chrome autofills username field with e-mail address

stephanegigandet:
Adopt the new template structure discussed in https://github.com/openfoodfacts/openfoodfacts-server/pull/5274

Fixes #4968 in Javascript (alternate option to the server side fix proposed by @hstauf in https://github.com/openfoodfacts/openfoodfacts-server/pull/5169


teolemon (approved):
It would be great to mass rename usernames with gmail or hotmail into them, and notify users.

stephanegigandet:
> It would be great to mass rename usernames with gmail or hotmail into them, and notify users.

Good idea. Filed #5285

templates/web/pages/user_form/user_form.tt.js
```diff
@@ -1,15 +1,15 @@
...
-		\$('.tr_teams').show();
+		\$('#teams_section').show();
 	}
 	\$(document).foundation('equalizer', 'reflow');
 });
 
 if (\$('#pro').prop('checked')) {
 	\$('.pro_org_display').show();
-	\$('.tr_teams').hide();
+	\$('#teams_section').hide();
```
CharlesNepote, line 14: Here we have an id: $('#teams_section') and in user_form_page.tt.html we have a class:
```
    .teams_section {
        display: block;
    }
```
I believe one of those is probably wrong, isn't it.

stephanegigandet: Good catch. The style is not needed anymore, I'll remove it.

stephanegigandet: fixed by https://github.com/openfoodfacts/openfoodfacts-server/pull/5287

PR #5281, opened 2021-06-02
improve milk recognition, fix nutriscore issue

stephanegigandet:
This is to fix a nutriscore issue signaled on twitter for https://world.openfoodfacts.org/product/7610900251759/caffe-latte-double-zero-macchiato-emmi

Beverages that contain 80% of milk are considered food for the nutriscore. We rely on the ingredients analysis to determine if there is milk and how much there is in it.

### DIFF
--- a/lib/ProductOpener/Config_off.pm
+++ b/lib/ProductOpener/Config_off.pm
@@ -424,22 +424,20 @@ XML
 ;
 
 
-# Nutriscore: milk and drinkable yogurts are not considered beverages
-# list only categories that are under en:beverages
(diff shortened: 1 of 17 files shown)


teolemon (approved):
I'm not quite sure why the dairy drink (in addition to the %) was removed, but validated

stephanegigandet:
> I'm not quite sure why the dairy drink (in addition to the %) was removed, but validated

It's to make sure we don't consider a beverage with 80% milk as a beverage, even if the dairy drinks category is not there.

Issue #5294, opened 2021-06-07
Organization form throwing error "error_list.tt.html not found"


roshnaeem:
**Describe the bug**
The structure of the templates directory has changed. The path of `error_list.tt.html` needs to be updated.

PR #5295, opened 2021-06-07
Updated path of error_list.tt.html in org_from.tt.html

roshnaeem:
Fixes: https://github.com/openfoodfacts/openfoodfacts-server/issues/5294


stephanegigandet (approved):
Thanks!

PR #5296, opened 2021-06-07
Updated error_list.tt.html path

roshnaeem:
Fixes: https://github.com/openfoodfacts/openfoodfacts-server/issues/5294


stephanegigandet (approved):
Thanks!

PR #5316, opened 2021-06-15
Templatization of import_photo_upload.pl

roshnaeem:
**Description:**
Added the template for the HTML code of the form in `import_photo_upload.pl`.

### DIFF
--- a/cgi/import_photos_upload.pl
+++ b/cgi/import_photos_upload.pl
@@ -53,6 +53,8 @@ my $action = param('action') || 'display';
 
 my $title = lang("import_photos_title");
 my $html = '';
+my $js = '';
+my $template_data_ref = {};
 
 local $log->context->{type} = $type;
 local $log->context->{action} = $action;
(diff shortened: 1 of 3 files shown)


cgi/import_photos_upload.pl
```diff
@@ -64,153 +66,24 @@
...
-          <tbody class="files"></tbody>
-        </table>
-      </form>
-
-	<div id="empty_space_for_equalizer" style="height:200px;width:100px;">&nbsp;</div>
-
-HTML
-;
+  my $i = 0;
+  $template_data_ref->{i} = $i;
```
stephanegigandet, line 85: $i is always 0 in fact, so you can remove it, and replace all instances of [% i %] by 0.

roshnaeem: yeah, I checked that, `$i` is 0 everywhere, but it thought to keep it generic, in case of modifications. If someone modifies the code, he/she will only have to change the value of `$i` instead of replacing all the variables.

stephanegigandet: ok

stephanegigandet (approved):
Thanks!

PR #5318, opened 2021-06-15
Templatization of import_product_categories_from_public_database.pl 

roshnaeem:
**Description:**

Added a template file of `import_product_categories_from_public_database.pl` for `action==display`.

### DIFF
--- a/cgi/import_products_categories_from_public_database.pl
+++ b/cgi/import_products_categories_from_public_database.pl
@@ -55,23 +55,15 @@ my $action = param('action') || 'display';
 
 my $title = lang("import_products_categories_from_public_database");
 my $html = '';
+my $js = '';
+my $template_data_ref = {};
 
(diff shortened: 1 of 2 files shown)


cgi/import_products_categories_from_public_database.pl
```diff
@@ -55,23 +55,15 @@
 
 my $title = lang("import_products_categories_from_public_database");
 my $html = '';
+my $js = '';
```
stephanegigandet, line 58: I don't think the $js variable is used?

stephanegigandet (changes requested):

cgi/import_products_categories_from_public_database.pl
roshnaeem: Oh yeah sorry, added it for the next process file. Removed now.

stephanegigandet (approved):

PR #5321, opened 2021-06-16
Templatization of import_product_categories_from_public_database

roshnaeem:
Description:

* Added a new template file for process part of `import_product_categories_from_public_database`. 
* Changed the `<div>` tags to `<span>` for better UI.

### DIFF
--- a/templates/web/pages/import_product_categories_process/import_product_categories_from_public_database_process.tt.html
+++ b/templates/web/pages/import_product_categories_process/import_product_categories_from_public_database_process.tt.html
@@ -1,0 +1,10 @@
+<!-- start templates/[% template.name %] -->
+
+<p> job_id: [% job_id %]</p>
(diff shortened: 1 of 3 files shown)


templates/web/pages/import_product_categories_process/import_product_categories_from_public_database_process.tt.html
```diff
@@ -0,0 +1,10 @@
+<!-- start templates/[% template.name %] -->
+
+<p> job_id: [% job_id %]</p>
+<a href="[% link %]">status</a>
+
+</br>Poll: <span id="poll"></span>
```
stephanegigandet, line 6: I don't think </br> is valid, it can be simply <br>, or <br/> if we want valid XHTML (but that's not needed).

stephanegigandet (approved):
Thanks!

PR #5323, opened 2021-06-16
Templatization of import_file_select_format.pl

roshnaeem:
Description:

Templatization of `import_file_select_format.pl`.

Fixes: https://github.com/openfoodfacts/openfoodfacts-server/issues/5322


roshnaeem:
It is working fine now, the only issue I am facing now is the value of `placeholder` on line 40 of `import_file_select_format.tt.js`. Upon checking the source code it can be checked in the input field with the `id= select_field_oprion_tag_`.

@stephanegigandet

stephanegigandet:
There's an error in some languages like French, probably because the Javascript uses the single quote as a delimiter for some strings, instead of double quotes as it was before. So if the string contains a single quote, it will fail.

e.g. http://fr.pro.productopener.localhost/cgi/import_file_select_format.pl?file_id=1624029285-elle-vire-fichier-openfoodfacts&action=display  (just replace world. by fr. in the URL)

stephanegigandet:
> It is working fine now, the only issue I am facing now is the value of `placeholder` on line 40 of `import_file_select_format.tt.js`. Upon checking the source code it can be checked in the input field with the `id= select_field_oprion_tag_`.

Right there's something weird, e.g. in French I get this:

```
			if (field ==  "categories_specific") {

				var input = '<input id="select_field_option_tag_' + col + '" name="select_field_option_tag_' + col +  '" placeholder= catégorie style="width:150px;margin-bottom:0;height:28px;">';

```

and in English:
`
				var input = '<input id="select_field_option_tag_' + col + '" name="select_field_option_tag_' + col +  '" placeholder= category style="width:150px;margin-bottom:0;height:28px;">';
`

It's missing quotes.

roshnaeem:
> There's an error in some languages like French, probably because the Javascript uses the single quote as a delimiter for some strings, instead of double quotes as it was before. So if the string contains a single quote, it will fail.
> 
> e.g. http://fr.pro.productopener.localhost/cgi/import_file_select_format.pl?file_id=1624029285-elle-vire-fichier-openfoodfacts&action=display (just replace the world. by fr. in the URL)

Is it only in $instructions or with other tags as well? I changed the quotes of  
`instructions += "<p>[% lang('${tagtype}_specific_tag') %]</p>" + "<p>[% lang('${tagtype}_specific_tag_value') %]</p>";`
and it is working fine now. Should we change the quotes of other html tags too?

roshnaeem:
> > It is working fine now, the only issue I am facing now is the value of `placeholder` on line 40 of `import_file_select_format.tt.js`. Upon checking the source code it can be checked in the input field with the `id= select_field_oprion_tag_`.
> 
> Right there's something weird, e.g. in French I get this:
> 
> ```
> 			if (field ==  "categories_specific") {
> 
> 				var input = '<input id="select_field_option_tag_' + col + '" name="select_field_option_tag_' + col +  '" placeholder= catégorie style="width:150px;margin-bottom:0;height:28px;">';
> ```
> 
> and in English:
> `var input = '<input id="select_field_option_tag_' + col + '" name="select_field_option_tag_' + col + '" placeholder= category style="width:150px;margin-bottom:0;height:28px;">';`
> 
> It's missing quotes.

Sorry, I could not understand the difference.

stephanegigandet:
> Should we change the quotes of other html tags too?

Yes. Ideally, you should generate exactly the same code as the one that is currently generated in production. That way we can be sure there won't be bad surprises.

stephanegigandet:
> Sorry, I could not understand the difference.

It's missing quotes in both French and English:

 placeholder= category style="[..]

should be placeholder="category"  (the placeholder can have spaces in it)

stephanegigandet (approved):
Everything looks good, thank you!

PR #5334, opened 2021-06-21
Templatization of remove_products.pl

roshnaeem:
**Description:**
* Added the template file for `html` code. 

**Fixes**: https://github.com/openfoodfacts/openfoodfacts-server/issues/5333

### DIFF
--- a/templates/web/pages/remove_products/remove_products.tt.html
+++ b/templates/web/pages/remove_products/remove_products.tt.html
@@ -1,0 +1,20 @@
+<!-- start templates/[% template.name %] -->
+
+[% IF action == 'display' %]
+
+    <p>[% lang("remove_products_from_producers_platform_description") %]</p>
+    <p>[% lang("this_action_cannot_be_undone") %]</p>
+
+    <!--Start the form-->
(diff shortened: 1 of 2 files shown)


templates/web/pages/remove_products/remove_products.tt.html
```diff
@@ -0,0 +1,20 @@
+<!-- start templates/[% template.name %] -->
+
+[% IF action == 'display' %]
+
+    <p>[% lang("remove_products_from_producers_platform_description") %]</p>
+    <p>[% lang("this_action_cannot_be_undone") %]</p>
+
+    <!--Start the form-->
+    <form id="remove_products_form" method="post" onsubmit="return confirm('[% confirm_alert %]')" enctype="multipart/form-data">
```
stephanegigandet, line 9: there is a possibility that in one of the languages contains a single quote, which would brea the confirm() javascript.

I think that you can use the squote filter from Template Toolkit to escape any single quotes: http://template-toolkit.org/docs/manual/VMethods.html#section_squote

stephanegigandet (approved):
Thanks!

PR #5340, opened 2021-06-23
[WIP] product.pl templatization

roshnaeem:
**Description:**
Moving the `html` code to template files.


stephanegigandet (commented):
/*global toggle_manage_images_buttons */ // These are weird.

we needed that line because toggle_manage_images_buttons() was previously defined in product.pl, now that we have moved it in product-multilingual.js, we can remove it this line I think

cgi/product_multilingual.pl
```diff
@@ -69,6 +70,8 @@
 
 my @errors = ();
 
+$admin = 1;
```
stephanegigandet, line 73: We'll need to remove that :)

cgi/product_multilingual.pl
```diff
@@ -964,302 +923,47 @@ ($$)
...
-    }
-  });
-  alerts.on('close.fndtn.alert', function(event) {
-    \$.cookie('state_' + \$(this)[0].id, 'none', { path: '/', expires: 365, domain: '$server_domain' });
-  });
-});
-</script>
-HTML
-;
+	$template_data_ref_display->{obsolete} = $Lang{obsolete}{$lang};
```
stephanegigandet, line 931: We can remove those lines and use directly the lang() function in the template.

cgi/product_multilingual.pl
```diff
@@ -964,302 +923,47 @@ ($$)
...
-	\$([]).selectcrop('init_images',data.images);
-	\$(".select_crop").selectcrop('show');
-
-  },
-  error : function(jqXHR, textStatus, errorThrown) {
-	\$('div[id="moveimagesmsg"]').html(lang().product_js_images_delete_error + ' - ' + textStatus);
-  },
-  complete: function(XMLHttpRequest, textStatus) {
-
+		$template_data_ref_display->{copy_data} = $Lang{copy_data}{$lc};
```
stephanegigandet, line 960: we can remove those lines and use directly the lang() function in the template

roshnaeem: using the lang function after removing these lines results in hash number instead of text. 
Screenshot from 2021-07-19 12-44-41

Issue #5359, opened 2021-07-06
New images.openfoodfacts.org subdomain to serve product images from another server (ovh3)


stephanegigandet:
We currently have a static.openfoodfacts.org domain that we use for JS, CSS and all images (product images + logos for labels, attributes etc.).

The static domain is currently served by the nginx reverse proxy on off1.

We now automatically copy all product images to the ovh3 server. With a new images.openfoodfacts.org pointing to ovh3, we can have ovh3 serve the product images, and remove some load from off1.

@cquest has configured ovh3 so that it can serve the images locally, or proxy to off1 when we don't have the image yet (e.g. new product images)

PR #5360, opened 2021-07-06
serve product images from images.openfoodfacts.org

stephanegigandet:
This is to serve product images from ovh3 instead of off1.

@cquest has setup an automated copy of images from off1 to ovh3, and ovh3 can serve images locally if it has them, or get them from off1 and cache them otherwise (e.g. for new product images that have not been copied to ovh3 yet).

details in #5359

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -109,6 +109,7 @@ BEGIN
 		$original_subdomain
 		$subdomain
 		$formatted_subdomain
+		$images_subdomain
 		$static_subdomain
 		$world_subdomain
 		$producers_platform_url
@@ -294,6 +295,7 @@ $default_request_ref = {
 use vars qw();
 
 $static_subdomain = format_subdomain('static');
+$images_subdomain = format_subdomain('images');
 $world_subdomain = format_subdomain('world');
 
 my $user_preferences;	# enables using user preferences to show a product summary and to rank and filter results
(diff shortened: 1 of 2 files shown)


teolemon (commented):
This is going to affect robotoff, tools and apps, OCR generation…
How will the change affect them?

stephanegigandet:
> This is going to affect robotoff, tools and apps, OCR generation…
> How will the change affect them?

It should not affect them at all.

either they use the image URLs from the API, and they will get images.openfoodfacts.org instead of static.openfoodfacts.org, and they can load the images with the new URLs

or if they have hardcoded the static domain to reconstruct images urls, they will continue to load images from static.openfoodfacts.org

hangy (approved):

cquest:
Maybe the very last change for the SVG should not be changed, to only cache product images and nothing else.
Except this, it looks fine to me.

cquest (approved):

CharlesNepote:
Shouldn't Nginx conf be documented and published in a repo also?

stephanegigandet:
> Maybe the very last change for the SVG should not be changed, to only cache product images and nothing else.

Yes, good idea, I changed that back.

stephanegigandet:
> Shouldn't Nginx conf be documented and published in a repo also?

That would be good. We could add it somewhere in https://github.com/openfoodfacts/openfoodfacts-server/tree/main/conf

or maybe in the openfoodfacts-infrastructure repo, so that we can put all others nginx proxy configs from OVH3.
What do you think @cquest ?

Something we discussed with @ocervell yesterday is to reverse our "working" flow: instead of saving config changes after they are deployed, we could push them to github first and then have them deployed automatically from github.

I'm guilty of tinkering with nginx config in production until it works, which is why I did it in the reverse way (and more often did not do it: what is in /config/ is probably outdated and not what we have in prod today). But I guess it would be better if all the tinkering is done on a dev or preprod environment instead. :)

PR #5363, opened 2021-07-09
docs: Use `compose` command from Docker CLI

hangy:
Use [Compose V2](https://docs.docker.com/compose/cli-command/) commands.


hangy:
This works well in Docker Desktop for Windows 3.5.1. Can anyone verify other platforms? 🙂

VaiTon:
On Ubuntu 20.04 LTS still not supported. Could we add it as a suggestion instead of replacing the old command?

hangy:
I think we can leave it as-is for the moment. Apparently, the new docker CLI command also works as a drop-in replacement for docker-compose.

ocervell (approved):
Looks good to me

VaiTon:
> 
> 
> I think we can leave it as-is for the moment. Apparently, the new docker CLI command also works as a drop-in replacement for docker-compose.

@hangy do you want to close this? Put it as a WIP?

ocervell:
@hangy We run Debian 10 on Proxmox VMs, but the `docker compose` command is not supported there either

alexgarel:
@hangy hum sorry, this would have been cool if it was supported…

hangy:
> @hangy hum sorry, this would have been cool if it was supported…

Yup, but this PR was badly outdated given the recent changes to Docker infrastructure and build scripts.

PR #5371, opened 2021-07-13
small fixes for Spanish kefirs ingredient parsing

stephanegigandet:
1. added some kefir ferments ingredients
2. added a Spanish translation for the organic ingredients
3. made the Spanish translation work (special case because it contains " e ", which we turn to " y " to ease parsing)


t/expected_test_results/ingredients/fi-percents.json
```diff
@@ -2,23 +2,23 @@
...
+         "percent_estimate" : "6.5",
          "text" : "Appelsiini",
          "vegan" : "yes",
          "vegetarian" : "yes"
       },
       {
          "id" : "en:apple",
-         "percent" : 3.5,
+         "percent" : "3.5",
          "percent_estimate" : 81.2,
```
CharlesNepote, line 22: Why percent is string while percent_estimate is a number?

CharlesNepote (approved):

t/expected_test_results/ingredients/fi-percents.json
stephanegigandet: It's a long standing issue, not related to this change: scalars in Perl have no type, and internally Perl changes the internal type depending on the last operation made on something. 
Definitely not an ideal situation for languages that expect a specific type in the JSON. In practice we try to force the type on some fields, but I probably didn't do it for the ingredients.

PR #5378, opened 2021-07-17
Translations for the NOVA groups taxonomy

teolemon:
Translations for the NOVA groups taxonomy


teolemon (commented):
There's no difference between processed and ultra processed in the Turkish translation

alexgarel (approved):

PR #5393, opened 2021-07-19
Created template file for function display_blocks() in Display.pm

roshnaeem:


templates/web/common/includes/display_blocks.tt.html
```diff
@@ -0,0 +1,19 @@
+<!-- start templates/[% template.name %] -->
+
+[% FOREACH block IN blocks %]
+
+    <div class="block">
+        <h3 class="block_title">[% block.block_ref_title %]</h3>
```
stephanegigandet, line 6: Here you can check if [% block.block_ref_title %] is empty, if it is, don't display the <h3> block.

lib/ProductOpener/Display.pm
```diff
@@ -7066,25 +7066,23 @@ sub display_blocks($)
...
-$block_ref->{content}
-</div>
-</div>
-";
-		if ((defined $block_ref->{id}) and ($block_ref->{id} eq 'my_block')) {
-			$html .= "<!-- end off canvas blocks for small screens -->\n";
-		}
-
+		push(@blocks, {
+			block_ref_content => $block_ref->{content},
```
stephanegigandet, line 7074:
```suggestion
			content => $block_ref->{content},
```

We can remove the block_ref_ part from the name.

stephanegigandet: In fact you don't even need to recreate a @blocks array: we already have one in $request_ref->{blocks_ref}

so you can do :

my $template_data_ref->{blocks} = $request_ref->{blocks_ref}

lib/ProductOpener/Display.pm
```diff
@@ -7063,28 +7063,14 @@ sub display_bottom_block($)
...
-";
-		if ((defined $block_ref->{id}) and ($block_ref->{id} eq 'my_block')) {
-			$html .= "<!-- end off canvas blocks for small screens -->\n";
-		}
-
-	}
+	my $template_data_ref_blocks->{blocks} = $request_ref->{blocks_ref};
 
 	# Remove empty titles
 	$html =~ s/<div class=\"block_title\"><\/div>//g;
```
stephanegigandet, line 7071: # Remove empty titles
	$html =~ s/<div class=\"block_title\"><\/div>//g;

--> this can now be removed as you now check for empty titles in the template

stephanegigandet (approved):
Looks good, thanks!

PR #5394, opened 2021-07-19
Removed joghurtkulturen

aleene:
**Description:**
Removed joghurtkulturen from the allergens, so that no longer triggers the milk allergen.


stephanegigandet (approved):
Thanks!

PR #5396, opened 2021-07-20
Fixed mixed vegetables categories

stephanegigandet:
in particular, frozen mixes should not trigger fresh vegetables


TaciteOFF (approved):
👍🏼

PR #5397, opened 2021-07-20
Added template file for function 'display_product_search_or_add' in D…

roshnaeem:


templates/web/common/includes/display_product_search_or_add.tt.html
```diff
@@ -0,0 +1,22 @@
...
+
+[% IF !server_options_producers_platform %]
+  [% display_search_image_form %]
+[% END %]
+
+<form method="post" action="/cgi/product.pl" enctype="multipart/form-data">
+
+    <div class="row collapse">
+        <div class="small-9 columns">
+          <input type="text" name="code" placeholder="[% lang('or') %] [% lang('barcode') %]">
```
stephanegigandet, line 11: lang('or') is not the same thing as what is in $template_data_ref_content->{or}

roshnaeem: Using $template_data_ref_content->{or}as [% or %] in template file was throwing an error. 
`"Unexpected token or [% or %]" `
So i used it directly, still thinking of way for 

```
$or =~ s/( |\&nbsp;)?://;
```

lib/ProductOpener/Display.pm
```diff
@@ -7530,59 +7530,33 @@ sub display_product_search_or_add($)
...
 
-		$html = <<HTML
-&rarr; <a href="/cgi/import_file_upload.pl">$Lang{import_product_data}{$lc}</a><br>
-&rarr; <a href="/cgi/import_photos_upload.pl">$Lang{import_product_photos}{$lc}</a><br>
-&rarr; <a href="/cgi/export_products.pl">$Lang{export_product_data_photos}{$lc}</a><br>
-&rarr; <a href="/cgi/remove_products.pl">$Lang{remove_products_from_producers_platform}{$lc}</a><br>
-</p>
-HTML
-;
+		process_template('web/common/includes/display_product_search_or_add_producer.tt.html', $template_data_ref_content_producer, \$html) || ($html = "template error: " . $tt->error());
```
stephanegigandet, line 7542: here we use $html for 2 different content, we should use a new variable $html_producer instead, and declare it inside this block.
The declaration of my $template_data_ref_content_producer = {}; should be moved inside this block as well.

stephanegigandet (approved):

PR #5411, opened 2021-07-26
New ingredients

aleene:
For product https://fr.openfoodfacts.org/produit/3273220085301/dessert-vegetal-noisette-amande-sojasun

I do now know how to add `sels de calcium`, as there are many of those.


stephanegigandet (approved):
Looks good

PR #5423, opened 2021-07-29
Templatization of function display_field() in Display.pm

roshnaeem:


templates/web/common/includes/display_field.tt.html
```diff
@@ -0,0 +1,21 @@
...
+    [% IF to_do_status != "" %]
+        <p><span class="field">[% lang("to_do_status") %]: </span>[% to_do_status %]</p>
+    [% END %]
+    [% IF done_status != "" %]
+    <p><span class="field">[% lang("done_status") %]: </span>[% done_status %]</p>
+    [% END %]
+[% END %]
+
+[% IF value_check.defined && value_check!= '' %]
+    [% IF field != 'states' %]
```
stephanegigandet, line 13: we could remove that check here, and use an ELSE clause after the [% IF field == 'states' %] block

roshnaeem: What about the if clause at line 12? I am moving it inside the else block.

stephanegigandet: I think it would be better to do the reverse: in the Perl code, move the 	process_template('web/common/includes/display_field.tt.html', $template_data_ref_field, \$html) || return "template error: " . $tt->error();  inside the loop, so that it is not called at all if the field does not have a defined value. That would remove a bunch of code.

roshnaeem: Which loop? the value of $field is not being defined inside the loop. Sorry,I could not understand this, can you please explain a little bit more.

stephanegigandet: I meant in the condition, sorry. In the Perl code, you have this condition:

if ((defined $value) and ($value ne '')) {
}
and then after:
process_template('web/common/includes/display_field.tt.html', $template_data_ref_field, $html) 

but if there's no defined value, then there's no point to call the template processing, so you can move process_template('web/common/includes/display_field.tt.html', $template_data_ref_field, $html)  inside the condition, at the end of it

roshnaeem: The template file also includes the **To Do:** and **Done:** before this condition depending upon the value of `$field`. 

```
if ($field eq 'states'){
}
```

Moving `process_template()` inside the 
```
if ((defined $value) and ($value ne '')) {
}
```
will not print the value of **To Do** and **Done** if `$value` will not be defined.

stephanegigandet (approved):

PR #5434, opened 2021-08-02
Templatization of display_possible_improvement_description in Display.pm

roshnaeem:


lib/ProductOpener/Display.pm
```diff
@@ -7739,31 +7739,25 @@ sub display_possible_improvement_description($$) {
...
-				$product_ref->{improvements_data}{$tagid}{current_value},
-				$product_ref->{improvements_data}{$tagid}{new_value},
-				sprintf("%d", $product_ref->{improvements_data}{$tagid}{difference_percent})) . "</p>";
-		}
-	}
+	my $template_data_ref_improvement = {};
+
+	$template_data_ref_improvement->{tagid} = $tagid;
+	$template_data_ref_improvement->{product_ref_improvements_data} = $product_ref->{improvements_data};
+	$template_data_ref_improvement->{product_ref_improvements_data_tagid} = $product_ref->{improvements_data}{$tagid};
```
stephanegigandet, line 7746: If $product_ref->{improvements_data} is not defined, the line above will trigger an error.

That's why we had "	if ((defined $product_ref->{improvements_data}) and (defined $product_ref->{improvements_data}{$tagid})) {" in the Perl code before.

templates/web/common/includes/display_possible_improvement_description.tt.html
```diff
@@ -0,0 +1,21 @@
+<!-- start templates/[% template.name %] -->
+
+[% IF product_ref_improvements_data.defined && product_ref_improvements_data_tagid.defined %]
```
stephanegigandet, line 3: this check should be in the Perl code, and we should not call the template if we don't have improvements data defined

lib/ProductOpener/Display.pm
```diff
@@ -7742,27 +7742,26 @@ sub display_possible_improvement_description($$) {
 
 	if ((defined $product_ref->{improvements_data}) and (defined $product_ref->{improvements_data}{$tagid})) {
 
-		# Comparison of product nutrition facts to other products of the same category
+		my $template_data_ref_improvement = {};
+
+		$template_data_ref_improvement->{tagid} = $tagid;
+		$template_data_ref_improvement->{product_ref_improvements_data} = $product_ref->{improvements_data};
```
stephanegigandet, line 7748: we should keep the checks if ($tagid =~ /^en:nutrition-(very-)?high/) { and if ($tagid =~ /^en:better-nutri-score/) { here, otherwise we will get undefined values for the sprintf function calls like sprintf(lang("value_for_the_category"), display_taxonomy_tag($lc, "categories", $product_ref->{improvements_data}{$tagid}{category}));

stephanegigandet (approved):
Thanks!

Issue #5205, opened 2021-05-05
Split Display.pm into API.pm and Frontend.pm, make frontend use the API


stephanegigandet:
Related to the discussion about rearchitecturing Product Opener: #5170

I've been trying to find ways to better separate things in Product Opener: what's relative to the web frontend, to the API, to MongoDB queries, to processing products when we write them etc.

I'm thinking we could try to split Display.pm which is probably the worse module in terms of how everything is mixed together.

We could do it gradually:
- create API.pm and Frontend.pm
- gradually move functions from Display.pm to API.pm and Frontend.pm if there is a clear split
- if there isn't a clear split, refactor the function so that there is one
- gradually remove dependencies to Display.pm in other modules and scripts
- gradually remove MongoDB / product .sto files write and read access from the frontend, and have the frontend use the API instead
- create missing APIs (e.g. user management, products aggregates (list of tags for a facet) etc.)

What do you think?

M123-dev:
@stephanegigandet, as I mentioned in #5170 I think this is a good idea. But how do you imagine the transition time, this process will take a relatively long and possibly lead to many errors and broken things. 
Should this repo be continued, should API.pm and Frontend.pm be new repositories or what were you thinking of?

stephanegigandet:
I think we can do it gradually without breaking anything, we can create new empty modules API.pm and Frontend.pm, and move functions one by one.

M123-dev:
sounds good

CharlesNepote:
A great idea IMHO. I think there are many benefits:
* code architecture will be more clear for newcomers
* it will help front-end devs
* it will help to fine-tune the API
* it's a concrete and first step to architecture changes we talked about (languages, etc.)
* as Stéphane already said, it allows moving gradually without breaking things

M123-dev:
Is there already any progress here

teolemon:
@stephanegigandet @M123-dev I believe there will be progress tomorrow when this is deployed to .net

teolemon:
For the record, @VaiTon has started an alternative web frontend: https://openfoodfacts-explorer.vercel.app/

PR #5435, opened 2021-08-03
Moved the function display_my_block() to Web.pm

roshnaeem:
**Description:** 

"Re-architecturing the Open Food Facts Perl modules and functions, possibly making some of them more generic and publishing them on CPAN."
Specifically, `Display.pm` is way too big, and it's very difficult to understand what all the functions do. Trying to move the functions from `Display.pm` to new files like `Web.pm`, `Api.pm`, etc

Moved the function display_my_block() from `Dispaly.pm` to `Web.pm`.

**Related to :** https://github.com/openfoodfacts/openfoodfacts-server/issues/5205

### DIFF
--- a/lib/ProductOpener/Web.pm
+++ b/lib/ProductOpener/Web.pm
@@ -41,14 +41,37 @@ use utf8;
 use Exporter qw(import);
 
 use ProductOpener::Display qw/:all/;
+use ProductOpener::Store qw(:all);
+use ProductOpener::Config qw(:all);
+use ProductOpener::Tags qw(:all);
+use ProductOpener::TagsEntries qw(:all);
+use ProductOpener::Users qw(:all);
+use ProductOpener::Index qw(:all);
+use ProductOpener::Lang qw(:all);
+use ProductOpener::Images qw(:all);
+use ProductOpener::Food qw(:all);
+use ProductOpener::Ingredients qw(:all);
+use ProductOpener::Products qw(:all);
+use ProductOpener::Missions qw(:all);
+use ProductOpener::MissionsConfig qw(:all);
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Web.pm
```diff
@@ -74,4 +97,55 @@ sub display_blocks($)
 	return $html;
 }
 
+=head1 FUNCTIONS
+
+=head2 display_my_block ( $request_ref )
+
+The sidebar of home page consists of blocks. This function is used to to display one block.
```
stephanegigandet, line 104:
```suggestion
The sidebar of home page consists of blocks. This function is used to to display one block with information and links related to the logged in user.
```

lib/ProductOpener/Web.pm
```diff
@@ -41,14 +41,37 @@ use utf8;
 use Exporter qw(import);
 
 use ProductOpener::Display qw/:all/;
+use ProductOpener::Store qw(:all);
```
stephanegigandet, line 44: Do we really need all those modules?

One thing that would be very nice to do is to remove the circular dependencies that we currently have (module A that use module B, and module B that uses module A). So we should only use the minimum set of modules that are actually used.

roshnaeem: I added them because we are gonna need them for the next functions we are gonna move to Web.pm. Also, I was not sure where `$server_options`  is present.

stephanegigandet: $server_options is imported from Config.pm

stephanegigandet (approved):

stephanegigandet (approved):
Thanks!

Issue #5275, opened 2021-06-01
Use docker to install and maintain the openfoodfacts.net test server?


stephanegigandet:
We currently run the test server (used by apps etc. to test the API) https://world.openfoodfacts.net on a very cheap (but that we have to pay) OVH server. We are going to replace that server with proxmox containers on much more powerful servers provided for free by OVH, so we need to move https://world.openfoodfacts.net to a proxmox container. (and also openfoodfacts.dev that we use for development)

I'm wondering if we should also try to use docker to install that instance of product opener and to keep it up-to-date. It would have the advantage of forcing us to keep the docker environment current and working, and so would facilitate things for developers.

It would probably require some changes though, as the product and images directories would be synchronized from production regularly. I'm guessing we might want to just bind all the data directories to the host filesystem instead of using docker volumes.

Any thoughts or suggestions?

stephanegigandet:
I think one reason to use docker is that product opener has tons of dependencies (libraries, external programs, Perl modules) that are difficult to track.

VaiTon:
I fully support the idea. It would also encourage us to build in "microservices" instead of a big monorepo.

PR #5436, opened 2021-08-03
[WIP] Deploy image to pre-prod Proxmox VM

ocervell:
**Description:**

Deploy container image to pre-prod OVH server on push to `deploy-*` branches. This PR will be used for testing the changes, and we'll modify `deploy-*` to `main` once everything works okay.

**Related issues and discussion:** #5275

### DIFF
--- a/.github/workflows/deploy_preprod.yml
+++ b/.github/workflows/deploy_preprod.yml
@@ -1,0 +1,27 @@
+name: Deploy to pre-prod Proxmox VM (Docker)
+
+on:
+  push:
+    branches:
+      - deploy-*
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Deploy image to pre-prod Proxmox VM
+      uses: appleboy/ssh-action@master
+      with:
+        host: ${{ secrets.HOST }}
+        username: ${{ secrets.USERNAME }}
+        key: ${{ secrets.SSH_PRIVATE_KEY }}
+        proxy_host: ${{ secrets.PROXY_HOST }}
+        proxy_username: ${{ secrets.USERNAME }}
+        script: |
+          git clone https://github.com/openfoodfacts/openfoodfacts-server/
+          git checkout ${{ github.sha }}
+          cd docker/
+          ./build_dev.sh
+          nohup sh start_dev.sh 2>&1 > /var/log/docker_compose.log &
+          build_npm.sh
+          docker-compose -f docker-compose.yml -f docker-compose.dev.yml exec backend bash /opt/scripts/import_sample_data.sh


.github/workflows/container-deploy.yml
```diff
@@ -0,0 +1,56 @@
...
+        proxy_host: ${{ secrets.PROXY_HOST }}
+        proxy_username: ${{ secrets.USERNAME }}
+        proxy_key: ${{ secrets.SSH_PRIVATE_KEY }}
+        script: |
+          git clone https://github.com/openfoodfacts/openfoodfacts-server/
+          git checkout ${{ github.sha }}
+          cd docker/
+          ./build_dev.sh
+          nohup sh start_dev.sh 2>&1 > /var/log/docker_compose.log &
+          build_npm.sh
```
hangy, line 55: I think it could be better to run `build_npm.sh` before `start_dev.sh`, because the server uses the dist files' timestamps for cache busting of some static resources with longer cache times.

hangy:
Nice to see progress on this!

You'll probably want to have an environment-specific `Config2.pm` referenced in an additional compose override file, becuase of the configured `$server_domain`: https://github.com/openfoodfacts/openfoodfacts-server/blob/1486b043e164424f7a549a3428f57a627fc6b3b0/docker/backend-dev/conf/Config2.pm#L50

.github/workflows/container-deploy.yml
```diff
@@ -0,0 +1,56 @@
...
+      uses: appleboy/ssh-action@master
+      with:
+        host: ${{ secrets.HOST }}
+        username: ${{ secrets.USERNAME }}
+        key: ${{ secrets.SSH_PRIVATE_KEY }}
+        proxy_host: ${{ secrets.PROXY_HOST }}
+        proxy_username: ${{ secrets.USERNAME }}
+        proxy_key: ${{ secrets.SSH_PRIVATE_KEY }}
+        script: |
+          git clone https://github.com/openfoodfacts/openfoodfacts-server/
```
hangy, line 50: May reduce fetch depth a little bit? The repo is pretty massive and you don't need the whole history of some large binary files for this deployment.

.github/workflows/container-deploy.yml
```diff
@@ -0,0 +1,56 @@
...
+          git checkout ${{ github.sha }}
+          cd docker/
```
hangy, line 52:
```suggestion
          cd openfoodfacts-server/docker/
```
Just missing the repo path, I think.

ocervell: Thanks, fixed !

.github/workflows/container-deploy.yml
ocervell: +1

ocervell:
Some more context for reference is available in #5440

docker/docker-compose.yml
```diff
@@ -6,6 +6,8 @@ services:
       - dbdata:/var/lib/mongodb
     networks:
       - webnet
+    ports:
+      - 27017:27017
```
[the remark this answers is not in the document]
ocervell: It exposes it from the Docker container to the Proxmox VM ; not to the outside world (we would need to open port 27017 on the machine for this). The reason we did this is because the Robotoff container (same machine, separate Proxmox container) needs to access this db as well.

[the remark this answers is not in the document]
ocervell: + I agree that eventually we'll still have to put some credentials through GitHub secrets on the MongoDB deployment

ocervell:
Getting this error at server start after adding the new variables:
```
backend_1    | <pre>Global symbol &quot;$google_cloud_vision_api_key&quot; requires explicit package name (did you forget to declare &quot;my $google_cloud_vision_api_key&quot;?) at /opt/product-opener/lib/ProductOpener/Config2.pm line 69.
backend_1    | Global symbol &quot;$crowdin_project_identifier&quot; requires explicit package name (did you forget to declare &quot;my $crowdin_project_identifier&quot;?) at /opt/product-opener/lib/ProductOpener/Config2.pm line 70.
backend_1    | Global symbol &quot;$crowdin_project_key&quot; requires explicit package name (did you forget to declare &quot;my $crowdin_project_key&quot;?) at /opt/product-opener/lib/ProductOpener/Config2.pm line 71.
backend_1    | Global symbol &quot;$csrf_secret&quot; requires explicit package name (did you forget to declare &quot;my $csrf_secret&quot;?) at /opt/product-opener/lib/ProductOpener/Config2.pm line 72.
```

I think we need to export them at the top of the package, will try.

docker/backend-dev/conf/Config2.pm
```diff
@@ -39,23 +39,30 @@ BEGIN
...
 	);
 	%EXPORT_TAGS = (all => [@EXPORT_OK]);
 }
 use vars @EXPORT_OK ; # no 'my' keyword for these
+use strict;
+use utf8;
 
 # server constants
-$server_domain = "productopener.localhost";
+$server_domain = $ENV{PRODUCT_OPENER_DOMAIN};
```
stephanegigandet, line 56: Shouldn't we keep the || "productopener.localhost"; part ?

e.g. for development, I run start_dev.sh without any environment variables defined. It would be nice to keep a default URL.

ocervell: It will be the default already in `docker/.env` so there is no need to duplicate.

ocervell: `.env` is loaded when `docker-compose` runs :) so in the local environment, the repo's `.env` will be loaded

docker/backend-dev/conf/Config2.pm
```diff
@@ -39,23 +42,30 @@ BEGIN
...
 
-@ssl_subdomains = qw();
+@ssl_subdomains = qw(
+*
```
hangy, line 62: This doesn't work with the simple `start_dev.bat` script, or when you start it from VS Code as a dev container, because we currently don't have TLS set up for local development. We could either tell nginx to serve HTTPS for local dev, or make this configurable via `$ENV`, too.

.github/workflows/container-deploy.yml
```diff
@@ -0,0 +1,141 @@
...
+      env:
+        DOCKER_CLIENT_TIMEOUT: 120
+        COMPOSE_HTTP_TIMEOUT: 120
+        PRODUCT_OPENER_DOMAIN: ${{ secrets.PRODUCT_OPENER_DOMAIN }}
+        ROBOTOFF_URL: ${{ secrets.ROBOTOFF_URL }}
+        GOOGLE_CLOUD_VISION_API_KEY: ${{ secrets.GOOGLE_CLOUD_VISION_API_KEY }}
+        CROWDIN_PROJECT_IDENTIFIER: ${{ secrets.CROWDIN_PROJECT_IDENTIFIER }}
+        CROWDIN_PROJECT_KEY: ${{ secrets.CROWDIN_PROJECT_KEY }}
+        CSRF_SECRET: ${{ secrets.CSRF_SECRET }}
+        GEOLITE2_PATH: ${{ secrets.GEOLITE2_PATH }}
```
hangy, line 46: Is this GitHub secret set in the current test setup? I think there'd probably be a volume mount missing for it, in that case. It might be easier to employ https://github.com/openfoodfacts/openfoodfacts-server/blob/9af7d8767d3e2a84b055f96b9fa18dc3a45988e6/docker/docker-compose.geolite2.yml inside this YAML pipeline: https://github.com/openfoodfacts/openfoodfacts-server/blob/b0d5043c25dca8b99b197e2a54c95df96938f66b/docker/README.md#L15

docker/backend-dev/conf/Config2.pm
```diff
@@ -66,13 +76,20 @@ $memd_servers = [ "memcached:11211" ];
 $facebook_app_id = "";
 $facebook_app_secret = "";
 
+$google_cloud_vision_api_key = $ENV{GOOGLE_CLOUD_VISION_API_KEY};
+$crowdin_project_identifier = $ENV{CROWDIN_PROJECT_IDENTIFIER};
+$crowdin_project_key = $ENV{CROWDIN_PROJECT_KEY};
+$csrf_secret = $ENV{CSRF_SECRET};
```
hangy, line 82: We actually removed CSRF in #1362, so maybe this config setting, and the corresponding environment variable should be deleted.

docker/docker-compose.yml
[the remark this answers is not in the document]
hangy: Would this work with Robotoff if we used 
```suggestion
      - "127.0.0.1:27017:27017"
```
instead? Or possibly even with an environment variable for the interface? The default binding `27017:27017` publishes the port on all interfaces, which makes it somewhat easy to accidentally open the port.

docker/backend-dev/conf/Config2.pm
```diff
@@ -26,53 +26,71 @@ use Exporter    qw< import >;
...
+use strict;
+use utf8;
 
 # server constants
-$server_domain = "productopener.localhost";
+$server_domain = $ENV{PRODUCT_OPENER_DOMAIN};
 
-@ssl_subdomains = qw();
+@ssl_subdomains = split / /, $ENV{SSL_SUBDOMAINS};
+$producers_platform = $ENV{PRODUCERS_PLATFORM};
```
hangy, line 62: Can we use this to replace the https://github.com/openfoodfacts/openfoodfacts-server/tree/main/docker/backend-pro-dev and update https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docker/docker-compose.pro.dev.yml accordingly? @stephanegigandet

hangy (approved):

docker/backend-dev/conf/Config2.pm
ocervell: Will make a separate PR for this, since adding env variables gives us a lot of freedom regarding having a unified Config2.pm / Dockerfile / docker-compose file, parameterized through env vars.

PR #5439, opened 2021-08-04
Fix order of deploy commands

hangy:
`git clone` doesn't switch to the cloned dir on it's own, so that `git checkout ${{ github.sha }}` didn't work. I only switched around the order of those command, and added an addition `git pull`, so that the newest commits are also pulled, even if the directory `openfoodfacts-server` directory existed previously.

### DIFF
--- a/.github/workflows/container-deploy.yml
+++ b/.github/workflows/container-deploy.yml
@@ -12,8 +12,8 @@ jobs:
     concurrency: off-net
     runs-on: ubuntu-latest
     steps:
-    # TODO: Uncomment when changing from deploy-* to main branch - this will 
-    # wait for container-build.yml build action to run before deploying the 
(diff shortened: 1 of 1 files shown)


.github/workflows/container-deploy.yml
```diff
@@ -53,12 +53,15 @@ jobs:
...
           [ ! -d 'openfoodfacts-server' ] && git clone --depth 1 https://github.com/openfoodfacts/openfoodfacts-server/ --no-single-branch
 
-          # Checkout current commit SHA
-          git checkout ${{ github.sha }}
-
           # Go to docker/ directory
           cd openfoodfacts-server/docker/
 
+          # Fetch newest commits (in case it wasn't freshly cloned)
+          git fetch --depth 1
```
ocervell, line 60: good call here :) will avoid some trouble down the road.

ocervell (approved):

PR #5440, opened 2021-08-04
Suggestion on how container deloyment might work

hangy:


ocervell:
I think your second suggestion is what we should be doing: using pre-built containers with an exact tag. That will ensure what we're deploying to prod will be working similarly.

IMO for .net and .info deployments we should be as close as possible to a production setup and treat it as a production setup.

That would mean:
- **Build Docker images with more precise tags** --> everytime we commit / merge to `main`, we should publish the `latest` tag, *and* the `<COMMIT_NUMBER>` tag
- **Wait for the container build workflow to run before running the deploy workflow** (see commented part in `container-deploy.yml` that should do just that)

Referring to [using `docker-compose` in production](https://docs.docker.com/compose/production/), here is an additional list of things to prepare for production setups:

- **Removing any volume bindings for application code, so that code stays inside the container and can’t be changed from outside** --> this means removing the `docker-compose.dev.yml` override as it binds host directories to the container. Probably also removing the existing binds in `docker-compose.yml` and use `COPY` in the Dockerfile to put those files in the correct directories on the container:

  - `./backend-dev/conf/Config.pm` to `/opt/product-opener/lib/ProductOpener/Config.pm`
  - `./backend-dev/conf/Config2.pm` to `/opt/product-opener/lib/ProductOpener/Config2.pm`
  - `./backend-dev/conf/log.conf` to `/mnt/podata/log.conf`
  - `./backend-dev/conf/minion_log.conf` to `/mnt/podata/minion_log.conf`
  - `./backend-dev/conf/apache.conf` to `/etc/apache2/sites-enabled/product-opener.conf`
  - `./backend-dev/conf/po-foreground.sh` to `/usr/local/bin/po-foreground.sh`
  - `./frontend-git/conf/nginx.conf` to `/etc/nginx/conf.d/default.conf`

- **Binding to different ports on the host**
- **Setting environment variables differently, such as reducing the verbosity of logging, or to specify settings for external services such as an email server**
- **Specifying a restart policy like restart: always to avoid downtime**
- **Adding extra services such as a log aggregator**

.github/workflows/container-deploy.yml
```diff
@@ -74,4 +83,4 @@ jobs:
           # sudo docker exec -it docker_mongodb_1 -c "cd /data/db && tar -xzvf openfoodfacts-mongodbdump.tar.gz && mongorestore"
 
           # Import sample data into docker-compose db
-          sudo docker exec -it docker_backend_1 -c "bash /opt/scripts/import_sample_data.sh"
+          sudo docker-compose -f ./docker-compose.yml -f ./docker-compose.dev.yml -f ./docker-compose.deploy-test.yml exec backend bash /opt/scripts/import_sample_data.sh
```
ocervell, line 86: what's the difference between this and the previous command ?

docker/backend-dev/conf/Config2-deploy-test.pm
```diff
@@ -0,0 +1,78 @@
...
+	    $facebook_app_secret
+		$robotoff_url
+		%server_options
+	);
+	%EXPORT_TAGS = (all => [@EXPORT_OK]);
+}
+use vars @EXPORT_OK ; # no 'my' keyword for these
+
+# server constants
+$server_domain = "openfoodfacts.net";
```
ocervell, line 50: If we used an env variable here `ENV{PRODUCT_OPENER_DOMAIN} || 'openfoodfacts.localhost'`, we would not need this additional config file

.github/workflows/container-deploy.yml
```diff
@@ -60,10 +60,19 @@ jobs:
           cd openfoodfacts-server/docker/
 
           # Pull latest container images
-          sudo docker-compose pull
+          sudo docker pull node:lts
+          sudo docker pull ghcr.io/openfoodfacts/openfoodfacts-server/backend:latest
+          sudo docker-compose -f ./docker-compose.yml -f ./docker-compose.dev.yml -f ./docker-compose.deploy-test.yml pull
+
+          # Rebuild npm
+          sudo bash ./build_npm.sh
```
ocervell, line 68: see discussions on using pre-built containers instead of re-building the frontend assets

.github/workflows/container-deploy.yml
hangy: With `docker exec` you need to assume the actual container name (ie. `docker_` prefix and `_1` suffix etc.) while `docker-compose` knows the service name through the provided compose files, and can simply execute the command in a service it knows to be running.

hangy:
> Referring to [using `docker-compose` in production](https://docs.docker.com/compose/production/), here is an additional list of things to prepare for production setups:

I think the first question we need to ask is if we actually want `docker-compose`, or if it should be at least `docker swarm` with the option to distribute load to different nodes.

>     * **Removing any volume bindings for application code, so that code stays inside the container and can’t be changed from outside** --> this means removing the `docker-compose.dev.yml` override as it binds host directories to the container. Probably also removing the existing binds in `docker-compose.yml` and use `COPY` in the Dockerfile to put those files in the correct directories on the container:
>       
>       * `./backend-dev/conf/Config.pm` to `/opt/product-opener/lib/ProductOpener/Config.pm`
>       * `./backend-dev/conf/Config2.pm` to `/opt/product-opener/lib/ProductOpener/Config2.pm`
>       * `./backend-dev/conf/log.conf` to `/mnt/podata/log.conf`
>       * `./backend-dev/conf/minion_log.conf` to `/mnt/podata/minion_log.conf`
>       * `./backend-dev/conf/apache.conf` to `/etc/apache2/sites-enabled/product-opener.conf`
>       * `./backend-dev/conf/po-foreground.sh` to `/usr/local/bin/po-foreground.sh`
>       * `./frontend-git/conf/nginx.conf` to `/etc/nginx/conf.d/default.conf`

They don't really need to be volume bindings, but could be used as [`configs`](https://docs.docker.com/compose/compose-file/compose-file-v3/#configs) instead. The different config files (https://github.com/openfoodfacts/openfoodfacts-server/blob/07af23cb0c63283944ec0d7c7c616c38a1f149b0/lib/ProductOpener/Config_obf.pm, https://github.com/openfoodfacts/openfoodfacts-server/blob/07af23cb0c63283944ec0d7c7c616c38a1f149b0/lib/ProductOpener/Config_off.pm, https://github.com/openfoodfacts/openfoodfacts-server/blob/07af23cb0c63283944ec0d7c7c616c38a1f149b0/lib/ProductOpener/Config_opf.pm, https://github.com/openfoodfacts/openfoodfacts-server/blob/07af23cb0c63283944ec0d7c7c616c38a1f149b0/lib/ProductOpener/Config_opff.pm) can be quite different, so that simply replacing/using some environment variables might not suffice.

https://github.com/openfoodfacts/openfoodfacts-server/blob/0dd16aacb2f91d9470764bf21f06081d03b4fa3c/docker/backend-dev/conf/po-foreground.sh is a workaround to reflect the symlinks used in production right now. It might not need be part of the final image.

>     * **Binding to different ports on the host**
> 
>     * **Setting environment variables differently, such as reducing the verbosity of logging, or to specify settings for external services such as an email server**
> 
>     * **Specifying a restart policy like restart: always to avoid downtime**
> 
>     * **Adding extra services such as a log aggregator**

Isn't there already some kind of log aggregator deployed externally? I remember reading about Grafana recently. @teolemon 

> **EDIT:**
> Regarding the config file(s):
> 
>     1. Is it really necessary to create a new `Config.pm` for this ? It seems the only difference between this one and the existing one is the `PRODUCT_OPENER_DOMAIN`, which we could configure through environment variables (with a default `productopener.localhost` for local setups), so essentially modify the variable like so: `$server_domain = $ENV{PRODUCT_OPENER_DOMAIN} || "productopener.localhost"`.

Have a look at the other `Config_*.pm` I linked to above. Especially the `%server_options` need to be configurable, and doing all of that with env vars only *might* be too much.

>     2. There would be an additional change in the `nginx.conf`: `server_name productopener.localhost *.productopener.localhost;`. We could use  something like `envsubst` before running the container to replace this with the correct `PRODUCT_OPENER_DOMAIN` since NGINX configs don't support env variables.

The current `nginx.conf` is pretty much only good for local development IMHO. We need nginx to host the static files (more efficiently than with apache2), but the config either needs to be updated to host HTTPS, or a load balancer could be used to offload SSL termination.

PR #5442, opened 2021-08-05
Templatization of display_data_quality_issues_and_improvement_opportunities() in Display.pm

roshnaeem:


templates/web/common/includes/display_data_quality_issues_and_improvement_opportunities.tt.html
```diff
@@ -0,0 +1,25 @@
+<!-- start templates/[% template.name %] -->
+
+[% FOREACH tagtype IN tagtypes %]
+    [% IF tagtype.product_ref_tagtype.defined && tagtype.scalar_product_ref_tagtype > 0 %]
```
stephanegigandet, line 4: this check should not be needed in the template: in the Perl code, we should only populate the tagtypes array with non empty values

templates/web/common/includes/display_data_quality_issues_and_improvement_opportunities.tt.html
```diff
@@ -0,0 +1,25 @@
...
+        <h2>[% tagtype.tagtype_heading %]</h2>
+
+        [% FOREACH tagid IN tagtype.tagids %]
+            <strong>[% tagid.display_taxonomy_tag %]</strong><br>
+
+            [% IF tagid.properties.defined %]
+                <p>[% tagid.properties %]</p>
+            [% END %]
+
+            [% IF tagtype.match('^data_quality') %]
```
stephanegigandet, line 14: we should do this if in the template, we can do it only once in the Perl code, and simpy have a tagid.description property

lib/ProductOpener/Display.pm
```diff
@@ -7782,29 +7782,39 @@ sub display_data_quality_issues_and_improvement_opportunities($) {
...
-				if ($tagtype =~ /^data_quality/) {
-					$html .= display_data_quality_description($product_ref, $tagid);
-				}
-				elsif ($tagtype eq "improvements") {
-					$html .= display_possible_improvement_description($product_ref, $tagid);
-				}
+				push(@tagids, {
+					display_taxonomy_tag => display_taxonomy_tag($lc, $tagtype, $tagid),
+					properties => $properties{$tagtype}{$tagid}{"description:$lc"},
+					display_data_quality_description => display_data_quality_description($product_ref, $tagid),
```
stephanegigandet, line 7804: we should not call both display_data_quality_description() and display_possible_improvement_description(), one of them is not relevant, based on the tagtype, so we should have the tagtype check here in the Perl code.

stephanegigandet: and based on the check, do description = display_possible_improvement_description()  or $description = the other function

stephanegigandet (approved):
Thanks!

Issue #5416, opened 2021-07-28
Duplicated pages might harm ranking of OFF's product page in search engines


teolemon:
Per @calsmith:
```
Because the same product pages are accessible at multiple URLs, e.g. the same Nutella product page is accessible at uk., .world., etc this may be harming the website’s SEO ranking on Google

Here’s some documentation we share externally: https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls

the link rel=“canonical” header tag should probably point to each page’s “world.” page to prevent ranking for the content being divided among each country subdomain thus reducing each individual page’s ranking
```

hangy:
Is the "world" page always the correct canonical URL? It can only display the English version or the main language version of a product.

Consider a product that has French and Italian (main) content: The world subdomain would show the Italian content, and if we tell Google that world is canonical, then the French version would probably never be indexed.

stephanegigandet:
Right, we can't assign "world" as the canonical URL for all product pages. We could do it for product pages in English I guess.

PR #5455, opened 2021-08-10
Fixing the SEO of product page

roshnaeem:
**Description:**  Changed the domain to the `http://world.productopener.localhost` to keep it permanent. 
**Related to:** https://github.com/openfoodfacts/openfoodfacts-server/issues/5416

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -7025,7 +7025,7 @@ sub display_page($) {
 		$canon_description = lang("site_description");
 	}
 	my $canon_image_url = "";
-	my $canon_url = $formatted_subdomain;
+	my $canon_url = 'http://world.productopener.localhost';
 
 	if (defined $request_ref->{canon_url}) {
 		if ($request_ref->{canon_url} =~ /^http:/) {
@@ -7107,7 +7107,7 @@ sub display_page($) {
 	$template_data_ref->{google_analytics} = $google_analytics;
 	$template_data_ref->{bodyabout} = $bodyabout;
 	$template_data_ref->{site_name} = $site_name;
-	
+
 	if (defined $request_ref->{page_type}) {
 		$template_data_ref->{page_type} = $request_ref->{page_type};
 	}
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -7025,7 +7025,7 @@ sub display_page($) {
 		$canon_description = lang("site_description");
 	}
 	my $canon_image_url = "";
-	my $canon_url = $formatted_subdomain;
+	my $canon_url = 'http://world.productopener.localhost';
```
stephanegigandet, line 7028: There are 2 issues:

1. http://world.productopener.localhost is a url that works only on your developer machine. In practice, we want to have the domain https://world.openfoodfacts.org/ in production, https://world.openfoodfacts.dev on the dev server etc.

2. we only want to change the subdomain from uk/ca/us.openfoodfacts.org etc. to world.openfoodfacts.org for the canon_url of pages that are both product pages, and that are in English.

In Display.pm, in the display_product() function, there is this line:

$request_ref->{canon_url} = product_url($product_ref);

That's where you want to change the subdomain (if the language is in English :  $lc == 'en').

Note that Display.pm sets a $formatted_subdomain (e.g. uk.openfoodfacts.org or uk.productopener.localhost on docker) and a $world_subdomain (world.openfoodfacts.org or world.productopener.localhost)

lib/ProductOpener/Display.pm
```diff
@@ -7870,6 +7870,12 @@ CSS
 
 	$request_ref->{canon_url} = product_url($product_ref);
 
+	if ($lc eq 'en') {
+		my $subdomain = 'https://world.openfoodfacts.org';
+		$request_ref->{canon_url} = $subdomain;
+		$request_ref->{canon_url} .= product_url($product_ref);
```
stephanegigandet, line 7876: That's not going to work, product_url() gives back the full URL.
You can try it with docker, it will give you a broken url in the <link rel="canonical"> tag.

product_url() gives you an url like https://uk.(openfoodfacts.org|productopener.localhost)/... 

and instead we want the same url, with uk (or us etc.) replaced by world, and the rest of the url as it was.

e.g. it can be done with a regex to replace https://(something without a dot). with https://world.

$request_ref->{canon_url} =~ s/https:\/\/([^\.])\./https:\/\/world./ ;

stephanegigandet (approved):
Thank you!

PR #5459, opened 2021-08-11
Added herb sumac to ingredients tax

aleene:
**Description:**
Added herb sumac


stephanegigandet (approved):
One of my favorite spices. :)

Issue #5463, opened 2021-08-11
Canonicalize taxonomy properties that refer to another taxonomy


stephanegigandet:
e.g. in the additives_classes taxonomy, en:stabiliser is the canonical name, and en:stabilizer is a synonym
in the additives taxonomy, property entries like additives_classes:en: en:stabilizer are kept as-is, without canonicalization.

This results in display issues:

https://fr.openfoodfacts.org/additif/e170i-carbonate-de-calcium

Fonctions : Colorant, en:stabilizer

PR #5471, opened 2021-08-13
Canonicalize properties that refer to other taxonomies

stephanegigandet:
fixes part of #5463

When a taxonomy like additives refer to another taxonomy for some properties (e.g. additives_classes), we now canonicalize the value (e.g. if we have additives_classes:en: en:stabilizer for an additive, we change it to en:stabiliser which is the canonical name in the additives_classes taxonomy).

This currently work only for properties that have exactly the same name as an already loaded taxonomy.

We'll also need to do it for mandatory_additive_class (and we will probably rename it to additives_classes.mandatory or something so that we can make the match).

### DIFF
--- a/lib/ProductOpener/Config_off.pm
+++ b/lib/ProductOpener/Config_off.pm
@@ -485,8 +485,11 @@ $options{categories_exempted_from_nutrient_levels} = [qw(
 )];
 
 # fields for which we will load taxonomies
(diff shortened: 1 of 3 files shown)


teolemon (approved):

hangy (approved):

teolemon:
@stephanegigandet good to merge

PR #5472, opened 2021-08-13
Fix and improve Estonian ingredients

jaens:
* Move some mis-translations.
* Add some plural and declensed forms.
* Remove "harilik" ("common") prefix from some plants which is rarely used in actual writing/speaking.
* Add some colloquial forms of words.

(I am a native Estonian speaker)

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -734,7 +734,7 @@ de:Carrageen
 el:Καραγενάνη
 eo:Karageno
 es:Carragenano, carragenanos, carragenato, carragenatos, carragenina, carrageninas,carragen
-et:Karraginaanid
+et:Karraginaanid, karrageenid
 fa:کاراگینان
 fi:karrageeni, karrageenit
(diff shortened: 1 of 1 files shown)


aleene:
Great that we have now a native estonian who can work on things.

Be aware that we put only ingredients in the taxonomy that are actually found on products. So add only as required, which might save you some work.

Thanks.

jaens:
Yeah, most of the translations added were to add wordings that are actually used on products vs. what looked like Wikipedia/Wikidata imports...

aleene:
I am probably responsible for the wikidata imports. Just hoped to kickstart translations without too many errors.

moon-rabbitOFF:
Great! Thank you for those translations. If you want to help improving ingredients in estonian, you can also take pictures of estonian products and add them to the database. You would then have more template to work with. I could help entering the data for each products (list of ingredients, nutrition table, categories...).

stephanegigandet (approved):
Thank you!

PR #5473, opened 2021-08-13
Config for olivier

teolemon:


docker/backend-dev/conf/Config2.pm
```diff
@@ -61,18 +76,29 @@ $mongodb = "off";
 $mongodb_host = "mongodb://mongodb:27017";
 $mongodb_timeout_ms = 50000; # config option max_time_ms/maxTimeMS
 
-$memd_servers = [ "memcached:11211" ];
+$memd_servers = [ "127.0.0.1:11211" ];
```
hangy, line 79: Not correct for all dev containers (local development)

hangy:
Have you tested the local development still works?

teolemon:
@hangy it's just to help @ocervell with debugging. Marking as WIP to reflect that

docker/backend-dev/conf/Config2.pm
ocervell: It's fine to keep the existing `memcached:11211` here, since the containers are running on the same VM.

ocervell:
Thanks, closing now :)

PR #5530, opened 2021-08-23
Templatization of display_recent_changes() in Display.pm

roshnaeem:
### Part of
- #2416


stephanegigandet (approved):
Looks good, thanks!

PR #5578, opened 2021-08-30
test file for knowledge panels

stephanegigandet:
This is a test file to demonstrate the response format of the new Knowledge Panels API (see https://docs.google.com/document/d/1vJ9gatmv8pCXxyOERmYD16jOKRWJpz1RaQQ5MEcTxms/edit#heading=h.f5zx2tcu9c5t )

Test file on https://fr.openfoodfacts.dev/files/knowledge_panel_test.json (login and pass: off)

### DIFF
--- a/html/files/knowledge_panel_test.json
+++ b/html/files/knowledge_panel_test.json
@@ -1,0 +1,91 @@
+{
+    "doyouknow_brands_nutella_423": {
+        "type": "doyouknow",
+        "level": "trivia",
+        "topics": [
+            "ingredients"
+        ],
+        "title": "Do you know why Nutella contains hazelnuts?",
+        "subtitle": "It all started after the second world war...",
+        "elements": [
+            {
+                "element_type": "text",
+                "element": {
+                    "html": "Cocoa beans were expensive and hard to come by after the second world war, so in Piedmont (Italy) where Pietro Ferrero created Nutella, they were replaced with hazelnuts to make <em>gianduja</em>, a mix of hazelnut paste and chocolate."
+                }
+            }
+        ]
+    },
+    "ecoscore": {
+        "type": "ecoscore",
+        "level": "info",
+        "topics": [
+            "environment"
+        ],
+        "title": "Eco-Score D",
+        "subtitle": "High environmental impact",
+        "grade": "d",
+        "icon_url": "https://static.openfoodfacts.org/images/attributes/ecoscore-d.svg",
+        "elements": [
+            {
+                "element_type": "text",
+                "element": {
+                    "text_type": "summary",
+                    "html": "The <a href=\"https://world.openfoodfacts.org/eco-score-the-environmental-impact-of-food-products\">Eco-Score</a> is an experimental score that summarizes the environmental impacts of food products."    
+                }
+            },
+            {
+                "element_type": "text",
+                "element": {
+                    "text_type": "note",
+                    "html": "The Eco-Score was initially developped for France and it is being extended to other European countries. The Eco-Score formula is subject to change as it is regularly improved to make it more precise and better suited to each country."    
+                }
+            },
+            {
+                "element_type": "panel",
+                "element": {
+                    "panel_id": "ecoscore_lca"
+                }  
+            },
+            {
+                "element_type": "text",
+                "element": {
+                    "text_type": "warning",
+                    "html": "<strong>Warning: some information necessary to calculate the Eco-Score with precision is not provided (see the details of the calculation below). You can edit the product to add the missing information.<br><br>If you are the manufacturer of this product, you can send us the information with our <a href=\"https://world.pro.openfoodfacts.org\">free platform for producers</a>."
+                }
+            }
+        ]
+    },
+    "ecoscore_lca" : {
+        "type": "ecoscore_lca",
+        "level": "info",
+        "title": "Lifecyle Analysis (LCA) for the product category",
+        "subtitle": "Score: 40/100",
+        "grade": "b",
+        "elements": [
+            {
+                "element_type": "text",
+                "element": {
+                    "text_type": "summary",
+                    "html": "<p>Agribalyse category: <a href=\"https://www.agribalyse.fr/app/aliments/31032\" title=\"Exact match with the product category\">Chocolate spread with hazelnuts</a></p><ul><li>PEF environmental score: 0.74 (the lower the score, the lower the impact)</li><li>- including impact on climate change: 9.87 kg CO2 eq/kg of product</li></ul>"
+                }
+            },
+            {
+                "element_type": "table",
+                "element": {
+                    "table_type": "percents",
+                    "title": "Details of the impacts by stages of the life cycle",
+                    "headers": ["Steps", "Impact"],
+                    "rows": [
+                        {"title": "Agriculture", "value": "82.7"},
+                        {"title": "Processing", "value": "11.5"},	
+                        {"title": "Packaging", "value": "2.8"},
+                        {"title": "Transportation", "value": "2.4"},
+                        {"title": "Distribution", "value": "0.6"},
+                        {"title": "Consumption", "value": "0.0"}
+                    ]
+                }
+            }
+        ]
+    }
+}
--- a/html/files/knowledge_panel_test.json
+++ b/html/files/knowledge_panel_test.json
@@ -13,7 +13,15 @@
                 "element": {
                     "html": "Cocoa beans were expensive and hard to come by after the second world war, so in Piedmont (Italy) where Pietro Ferrero created Nutella, they were replaced with hazelnuts to make <em>gianduja</em>, a mix of hazelnut paste and chocolate."
                 }
-            }
+            },
+            {
+                "element_type": "image",
+                "element": {
+                    "url": "https://static.openfoodfacts.org/images/attributes/contains-nuts.png",
+                    "width": 192,
+                    "height": 192
+                }
+            }            
         ]
     },
     "ecoscore": {
--- a/html/files/knowledge_panel_test.json
+++ b/html/files/knowledge_panel_test.json
@@ -1,5 +1,6 @@
 {
     "doyouknow_brands_nutella_423": {
+        "parent_panel_id": "root",
         "type": "doyouknow",
         "level": "trivia",
         "topics": [
@@ -26,6 +27,7 @@
         ]
     },
     "ecoscore": {
+        "parent_panel_id": "root",
         "type": "ecoscore",
         "level": "info",
         "topics": [
@@ -66,8 +68,12 @@
         ]
     },
     "ecoscore_lca" : {
+        "parent_panel_id": "ecoscore",
         "type": "ecoscore_lca",
         "level": "info",
+        "topics": [
+            "environment"
+        ],        
         "title": "Lifecyle Analysis (LCA) for the product category",
         "subtitle": "Score: 40/100",
         "grade": "b",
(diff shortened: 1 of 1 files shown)


html/files/knowledge_panel_test.json
```diff
@@ -0,0 +1,91 @@
...
+    "ecoscore": {
+        "type": "ecoscore",
+        "level": "info",
+        "topics": [
+            "environment"
+        ],
+        "title": "Eco-Score D",
+        "subtitle": "High environmental impact",
+        "grade": "d",
+        "icon_url": "https://static.openfoodfacts.org/images/attributes/ecoscore-d.svg",
```
jasmeet0817, line 28: is this correct? inside a panel the following are required fields:

type,
level,
topics,
title,
elements

following are optional:
subtitle,
icon_url,
grade

html/files/knowledge_panel_test.json
```diff
@@ -0,0 +1,91 @@
...
+                "element_type": "text",
+                "element": {
+                    "text_type": "summary",
+                    "html": "<p>Agribalyse category: <a href=\"https://www.agribalyse.fr/app/aliments/31032\" title=\"Exact match with the product category\">Chocolate spread with hazelnuts</a></p><ul><li>PEF environmental score: 0.74 (the lower the score, the lower the impact)</li><li>- including impact on climate change: 9.87 kg CO2 eq/kg of product</li></ul>"
+                }
+            },
+            {
+                "element_type": "table",
+                "element": {
+                    "table_type": "percents",
```
jasmeet0817, line 76: can we call this something more descriptive, like ecoscore_impact_breakdown ?

html/files/knowledge_panel_test.json
stephanegigandet: Sounds good.

jasmeet0817 (commented):
looks good to me, just a couple comments.

html/files/knowledge_panel_test.json
stephanegigandet: We can add a "table_id" field to tables so that the client can identify a specific table.

html/files/knowledge_panel_test.json
```diff
@@ -0,0 +1,91 @@
...
+                    "title": "Details of the impacts by stages of the life cycle",
+                    "headers": ["Steps", "Impact"],
+                    "rows": [
+                        {"title": "Agriculture", "value": "82.7"},
```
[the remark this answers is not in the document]
stephanegigandet: > will each row in a table have title, value fields ?

I think most tables will be simple tables with 2 columns. Those are the ones that we can display in creative ways (with bars, spider graphs etc.)

We could have another element type for tables with more columns that would be more generic.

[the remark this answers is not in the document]
stephanegigandet: I changed the panel type to "table_2_columns" to make it very explicit. And I added ids for tables and rows, so that clients can recognize them and do something specific with them if they want to.

html/files/knowledge_panel_test.json
jasmeet0817: got it, so my understanding is that table_type: "percents" should have 2 headers and each row should contain a title and value.

Can we document this, clients would be happy to follow the documentation to understand the structure.

html/files/knowledge_panel_test.json
```diff
@@ -0,0 +1,92 @@
...
+            },
+            {
+                "element_type": "table_2_columns",
+                "element": {
+                    "table_id": "ecoscore_lca_impacts_by_stages",
+                    "table_type": "percents",
+                    "title": "Details of the impacts by stages of the life cycle",
+                    "headers": ["Steps", "Impact"],
+                    "rows": [
+                        {"id": "agriculture", "title": "Agriculture", "value": "82.7"},
```
jasmeet0817, line 81: what's the use case of this id ?

html/files/knowledge_panel_test.json
jasmeet0817: cool, lets document this stuff somewhere in the server (can be in another PR).

jasmeet0817 (approved):
looks good, just a question aboute table-row.id

html/files/knowledge_panel_test.json
```diff
@@ -0,0 +1,92 @@
...
+            "environment"
+        ],
+        "title": "Eco-Score D",
+        "subtitle": "High environmental impact",
+        "grade": "d",
+        "icon_url": "https://static.openfoodfacts.org/images/attributes/ecoscore-d.svg",
+        "elements": [
+            {
+                "element_type": "text",
+                "element": {
```
jasmeet0817, line 32: can you add an image element as well ?

jasmeet0817 (approved):

html/files/knowledge_panel_test.json
jasmeet0817: is "topics" required? ecoscore_lca panel is missing this field

html/files/knowledge_panel_test.json
jasmeet0817: mmm actually, it would be better if we can just create generic tables, right now I have to create a specific object for table_2_columns and potentially another object for other tables.

It would be better to create a generic_table type, with N headers and N values (value1, value2 ... valueN). If we want to distinguish some special tables (like percent) that can also be rendered as pie chart/graphs etc, we can use "table_type" for it, like "percent" here. This should accomplish all use cases you have in mind (AFAICT) and is easier to implement on the client side.

if you want, we can chat more on this, I can show you why the code is less complicated with generic tables

jasmeet0817: so my suggestion is to turn the rows into an array, something like this:

'table_type': 'percents',
'headers': ['Steps', 'Impact'],
'rows': [
  ['Agriculture', '82.7'],
  ['Processing', '11.5'],
  ['Packaging', '2.8'],
  ['Transportation', '2.4'],
  ['Distribution', '0.6'],
  ['Consumption', '0.0']
]

Since the table_type is percents, the client should be able to decipher that it can also render this table as a bar/spider-graph/pie-chart etc.

html/files/knowledge_panel_test.json
stephanegigandet: It's to allow clients to decide to display specific rows in a specific way if they choose to.

html/files/knowledge_panel_test.json
```diff
@@ -0,0 +1,92 @@
...
+        "level": "trivia",
+        "topics": [
+            "ingredients"
+        ],
+        "title": "Do you know why Nutella contains hazelnuts?",
+        "subtitle": "It all started after the second world war...",
+        "elements": [
+            {
+                "element_type": "text",
+                "element": {
```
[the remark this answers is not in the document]
stephanegigandet: So my idea is that some texts would not have a type, and that we would just mark the text elements that have special types. Kind of a default type.

html/files/knowledge_panel_test.json
stephanegigandet: Sure, I added an image to the first "doyouknow" panel.

html/files/knowledge_panel_test.json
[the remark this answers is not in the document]
jasmeet0817: hmm, don't you think it's better to always have a type? We can default to "description" type for such? I think not having a type introduces a potential for interpretation.

html/files/knowledge_panel_test.json
jasmeet0817: do you have an example? I'm still not able to fully imagine the use case in my head, how will the client choose to display these differently, if they would use different icons etc, should we also pass the icons?

Anyways, I was kind of hoping we would start the first version with the basic cases, and slowly move to complicated cases like these. My principle is to avoid coding specialized use cases unless absolutely needed.

html/files/knowledge_panel_test.json
jasmeet0817: Thank you

html/files/knowledge_panel_test.json
stephanegigandet: Then you can ignore the "id" field for now, and we will see if we need it later. It's a good test case anyway: we will extend the knowledge panels, and clients should not break if an extra field they don't know about is added.

html/files/knowledge_panel_test.json
```diff
@@ -0,0 +1,100 @@
...
+                    "text_type": "warning",
+                    "html": "<strong>Warning: some information necessary to calculate the Eco-Score with precision is not provided (see the details of the calculation below). You can edit the product to add the missing information.<br><br>If you are the manufacturer of this product, you can send us the information with our <a href=\"https://world.pro.openfoodfacts.org\">free platform for producers</a>."
+                }
+            }
+        ]
+    },
+    "ecoscore_lca" : {
```
jasmeet0817, line 67: in order to distinguish that this panel doesn't belong to root, should we add a parent_id (as you suggested on the doc), the parent_id for this would be "ecoscore" and the parent_id for others would be "root".

html/files/knowledge_panel_test.json
stephanegigandet: We can use an array for the column values indeed, but I still think we need properties for rows.
Here is what I propose:

```
                    "rows": [
                        {"id": "agriculture", "icon_url": "https://static.openfoodfacts.org/images/icons/dist/agriculture.svg", "values": ["Agriculture", "82.7%"], "percent": 82.7},
                        {"id": "processing", "values": ["Processing", "11.5%"], "percent": 11.5},	
                        {"id": "packaging", "values": ["Packaging", "2.8%"], "percent": 2.8},
                        {"id": "transportation", "values": ["Transportation", "2.4%"], "percent": 2.4},
                        {"id": "distribution", "values": ["Distribution", "0.6%"], "percent": 0.6},
                        {"id": "consumption", "values": ["Consumption", "0.0%"], "percent": 0}
                    ]

```
each row has optional properties, and at least a "values" array.

The properties can be used to pass icons etc.

We can also use them to pass the percent info as a number. We need both a string and a number, because the string value we display will be translated (e.g. in French we will have "75,2%" with a comma, and in English it will be "72.5%" with a dot.

html/files/knowledge_panel_test.json
jasmeet0817: Cool, sounds good.

html/files/knowledge_panel_test.json
jasmeet0817: Sounds good, this looks better, thanks Stephane.

html/files/knowledge_panel_test.json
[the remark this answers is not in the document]
jasmeet0817: Thanks for the change

html/files/knowledge_panel_test.json
stephanegigandet: Sounds good. I added a parent_panel_id field.

html/files/knowledge_panel_test.json
stephanegigandet: Let's make topics optional.

jasmeet0817 (approved):

html/files/knowledge_panel_test.json
```diff
@@ -0,0 +1,107 @@
+{
+    "doyouknow_brands_nutella_423": {
```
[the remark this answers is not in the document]
stephanegigandet: @CharlesNepote : so in fact it's not a name at all, it won't be shown to users, it is a unique id.

You get all the panels in a hash, and the key to that hash is their id. Then some panels can show other subpanels, and to do so, they refer to them with their ids.

Here is the current structure returned when you asked for panels:

https://world.openfoodfacts.org/api/v2/product/3017620422003/nutella-ferrero?fields=knowledge_panels&cc=uk&lc=en

Issue #5557, opened 2021-08-26
Release the Eco-Score in the United Kingdom


teolemon:
### Who for
- UK users
- UK apps based on Open Food Facts
### What
- [x] Ensure transportation specific to the UK is properly loaded (it is not)
  - https://github.com/openfoodfacts/openfoodfacts-server/blob/main/ecoscore/data/fr_countries.csv
  - #5586 
- [x] Add the United Kingdom to supported countries
- [ ] Ensure we grasp data quick wins
  - [gb-org-05](https://hunger.openfoodfacts.org/questions?type=label&value_tag=en%3Agb-org-05)
  - [gb-org-02](https://hunger.openfoodfacts.org/questions?type=label&value_tag=en%3Agb-org-02)
  - [gb-org-04](https://hunger.openfoodfacts.org/questions?type=label&value_tag=en%3Agb-org-04)
  - [assured-food-standards](https://hunger.openfoodfacts.org/questions?type=label&value_tag=en%3Aassured-food-standards)
  - [Asda](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=asda)
  - [Lidl](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=lidl)
  - [Aldi](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=aldi)
  - [Morrisons](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=morrisons)
  - [Woolworths](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=woolworths)
  - [Iceland](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=iceland)
  - [Tesco](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=tesco)
  - [Waitrose](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=waitrose)
  - [Sainsbury's](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=sainsburys)
   - [Sainsbury's](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=sainsbury-s)
  - [The-co-operative](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=the-co-operative)
  - [Co-op](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=co-op)
  - [Coop](https://hunger.openfoodfacts.org/questions?type=brand&value_tag=coop)
[...]

PR #5586, opened 2021-09-01
Load Eco-Score distances for all available countries, enable Eco-Score for UK, fix regions origins

stephanegigandet:
Changes:

1. Distances are now loaded from a copy of @aleene 's https://docs.google.com/spreadsheets/d/1wPVzY8d1otgguHZ0Qd2jIStRpVpnsGPbFv91fMasQkc/edit#gid=1666241930 table.
--> Except France where we use the values from the original Eco-Score, so that we don't have discrepancies with the official doc.

2. Country specific Eco-Score are now computed for UK (+ the original 8 European countries we launched).
+ the Eco-Score is now shown for UK users

3. When the origin was specified directly in the ingredients list, it worked only with countries (e.g. "Milk (France)"). It now also works with the regions that are present in the origins taxonomy. (e.g. "Fraises (Bretagne)"). 

### Part of
- #5557


teolemon (approved):
Great 👍

Issue #5541, opened 2021-08-24
Serve Eco-Score data in the new KP API


jasmeet0817:
### Description
Serve eco score data in the new KP API

New KP API: https://docs.google.com/document/d/1vJ9gatmv8pCXxyOERmYD16jOKRWJpz1RaQQ5MEcTxms/edit#
Example JSON response: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/html/files/knowledge_panel_test.json

### Acceptance criteria
Feature parity with the current API

### What would a demo look like
The API can be queried to see the results.

### Tasks
TBD

### Part of 
- #5493

stephanegigandet:
This has been done and deployed.

PR #5597, opened 2021-09-02
Initial plumbing for knowledge panels API + start of Eco-Score panel

stephanegigandet:
- Start of #5541:

- added a way to request knowledge panels, similar to the product attributes API
- started to implement the Eco-Score panels

to test: https://world.openfoodfacts.dev/api/v2/product/3017620422003/nutella-ferrero?fields=knowledge_panels

### Part of
- #5493

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -125,6 +125,7 @@ BEGIN
 
 		$show_ecoscore
 		$attributes_options_ref
+		$knowledge_panels_options_ref
 
 		);    # symbols to export on request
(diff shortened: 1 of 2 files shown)


jasmeet0817 (approved):

teolemon (approved):
Do you plan on having the logic for all panels in the same file ? Wouldn't self-contained files be better on the long term ?

stephanegigandet:
> Do you plan on having the logic for all panels in the same file ? Wouldn't self-contained files be better on the long term ?

A good question, let's do a few panels first, a lot of that code is likely to change, some of it might also be moved to HTML or JSON templates, we can have a better idea of how to organize things when we have more actual examples.

Issue #2938, opened 2020-02-22
[WIP] Format for knowledge cards


teolemon:
title
icon
short_text
long_text
learn_more_url
bullet_point_1_icon
bullet_point_1_title
bullet_point_1_subtitle
bullet_point_1_click_url

teolemon:
#5539

PR #5603, opened 2021-09-03
Use JSON template to generate Eco-Score Agribalyse LCA knowledge panel 

stephanegigandet:
### What
This is an implementation of the Eco-Score Agribalyse sub panel with LCA impacts for the different steps. (see #5541 )

I tried a different approach for this one: instead of using Perl code to generate the expected structure for the panel, the Perl code just passes the necessary data to a template that generates JSON for the structure.

It's a bit more tricky than templates to generate HTML code, but it works.

The goal is to have a better separation between model/data and display logic and form, and to allow non programmers to change, reorder etc. the panels just by changing the templates.

### Part of
- #2938

### DIFF
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -49,8 +49,8 @@ BEGIN
 	use vars       qw(@ISA @EXPORT_OK %EXPORT_TAGS);
 	@EXPORT_OK = qw(
 
-		&compute_knowledge_panels
-		&compute_ecoscore_panels
+		&create_knowledge_panels
+		&create_ecoscore_panels
 
 		);    # symbols to export on request
 	%EXPORT_TAGS = (all => [@EXPORT_OK]);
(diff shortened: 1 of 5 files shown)


stephanegigandet:
Test URL: https://world.openfoodfacts.dev/api/v2/product/3017620422003/nutella-ferrero?fields=knowledge_panels (you can test other products that have an Eco-Score)

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -152,20 +154,126 @@ sub compute_knowledge_panels($$$$) {
...
+
+=cut
+
+sub create_panel_from_json_template ($$$$$$) {
+
+    my $panel_id = shift;
+    my $panel_template = shift;
+    my $panel_data_ref = shift;
+    my $product_ref = shift;
+	my $target_lc = shift;
```
jasmeet0817, line 210: nit, line spacing is inconsistent

jasmeet0817 (approved):

lib/ProductOpener/KnowledgePanels.pm
stephanegigandet: thanks, I fixed the spacing

PR #5608, opened 2021-09-03
More packaging shapes

teolemon:
### What
More packaging shapes

- https://bo.citeo.com/sites/default/files/2019-07/20190617_Guide_Info-tri_Citeo_FR.pdf
- https://bo.citeo.com/sites/default/files/2019-07/20190617_Guide_Info-tri_Citeo_EN.pdf

### Part of 
- #5520


stephanegigandet (changes requested):
To avoid issues, all entries in this taxonomy should have a unique English canonical name. It's really not simple as many of those French words are often translated interchangeably by different English names.

I would suggest to start with a few entries first, the ones that are the most common and associated with food.

We also need to make sure that any additional entry will be correctly mapped to an Eco-Score entry.

teolemon:
All those are in the CITEO system and guidelines, and as a result are likely to be used in French recycling instructions. Barring any packaging shape facet, that's the best I can do.

taxonomies/packaging_shapes.txt
```diff
@@ -277,9 +285,111 @@ fr:Batonnet, bâtonnets
...
+fr:Flacon à pompe, Flacon à pistolet, Spray
+packaging_materials:en: en:plastic, en:metal, en:glass
+
+en:Aerosol can
+fr:Aérosol
+packaging_materials:en: en:plastic, en:metal
+
+en:Ball flask, Roll-on bottle
+fr:Flacon à bille, Roll
+packaging_materials:en: en:plastic, en:glass
```
stephanegigandet, line 374: Adding properties like that is not a good idea if we are not sure how we are going to use them. Let's wait until we do know and until we actually use them somewhere, as otherwise it will be very hard later to figure out what to do with them, or we will have unwanted side issues.

e.g. "packaging_materials:en: en:plastic, en:glass" --> if later we implement tag rules, then that would add both plastic and glass. That's probably not wanted.

teolemon: I've added them as comments

taxonomies/packaging_shapes.txt
```diff
@@ -277,9 +285,111 @@ fr:Batonnet, bâtonnets
...
+#goes_with:en:glass bottle
+
+en:hanger
+fr:Cintre
+
+en:Wedge, corner piece
+fr:Cale, Coin, Cornière
+#suggest:packaging_materials:en: en:plastic, en:cardboard, en:metal, en:other
+
+# close to bottle, but also ampoule is translated to vial by citeo
```
stephanegigandet, line 342: So we have English translations for all Citeo elements? Then can we just add them to all the new fr: only entries?

stephanegigandet (approved):
Perfect, thank you!

teolemon:
@stephanegigandet some tests are failing

stephanegigandet:
Some of the new entries were already there as synonyms.

Errors in the packaging_shapes taxonomy definition:
ERROR - en:cup already is a synonym of en:pot for entry en:pot - en:cup cannot be mapped to entry en:cup
ERROR - en:tie already is a synonym of en:fastener for entry en:fastener - en:tie cannot be mapped to entry en:tie
ERROR - en:band already is a synonym of en:strip for entry en:strip - en:string cannot be mapped to entry en:string
ERROR - en:cover already is a synonym of en:lid for entry en:lid - en:cover cannot be mapped to entry en:cover
Errors in the packaging_shapes taxonomy definition at lib/ProductOpener/Tags.pm line 995.

I'll try to fix them.

stephanegigandet:
@teolemon : I think I fixed the conflicting entries, and I added some plurals + some tests. Remaining work needed: assigning a shape ratio for the new entries so that they can be taken into account correctly for the Eco-Score.

teolemon:
@stephanegigandet good to merge, and shape ratio in the next PR ?

PR #5609, opened 2021-09-04
Disable translation upload

teolemon:
I believe we don't need to reupload translations everytime. Not only does it slow down the process, but it recontaminates translations from the old source code translations.

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -21,7 +21,7 @@ jobs:
     - name: crowdin action
       uses: crowdin/github-action@1.3.3
       with:
-        upload_translations: true # default is false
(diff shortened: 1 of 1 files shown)


stephanegigandet:
What does "upload_translations" do exactly?

> I believe we don't need to reupload translations everytime. 

Meaning wil will reupload them sometime? When is that needed?

Just trying to figure out how this works as it's really unclear to me.

teolemon:
- It uploads all the translations, at each sync. That creates 2 conflicting source of truths. 
- Translations have already been exported to crowdin, so no need to upload it each time. 
- If we want to upload French translations, we can upload the file manually using the web interface.

stephanegigandet (approved):

PR #5611, opened 2021-09-05
Rename 'Validate' -> 'Resize'

blazern:
**Note**: all I did was a global search and replace. I have know idea of whether I'm breaking anything.

Changes according to this dialog: https://openfoodfacts.slack.com/archives/C02KVRT2C/p1630762333049300?thread_ts=1630741111.048400&cid=C02KVRT2C
image


stephanegigandet:
So in fact we just need to change the file in /lang/en/texts + /po/common/common.pot , the rest will be translated by Crowdin, they should not be changed manually.

Issue #3426, opened 2020-05-14
Tag rules: Tag A + Tab B = Tag C


stephanegigandet:
Following a discussion on how we can apply finer categories to products (e.g. Yogurts with fruits, Artificially sweetened yogurts instead of just Yogurts) massively to products, we have discussed introducing a new feature to automatically apply some categories when products have specific tags.

For instance, we could have rules like:

+categories:en:yogurts +ingredients:en:sweetener = +categories:en:artificially-sweetened-yogurts

+categories:en:yogurts +ingredients:en:sugar -ingredients:en:sweetener = +categories:en:sugared-yogurts (yogurts with sugar in the ingredients but without sweeteners)

We could create a testing tool to display which products would be affected by a rule we want to test.

And we could then automatically apply validated rules, either in real-time, or periodically (which would have the advantage of cleanly attributing new tags to the tag rules instead of to users).

bernardlemoullec:
Good idea! Thank you Stéphane

teolemon:
+categories:fr:Sancerre = origins_of_ingredients:en:France + labels:fr:aoc
+packager-codes:msc-* = labels:en:msc
+brands:fr:carrefour = +stores:fr:carrefour

teolemon:
- Associated Pull Request with a proposal by @stephanegigandet : https://github.com/openfoodfacts/openfoodfacts-server/pull/6188

PR #5614, opened 2021-09-05
Add more organic brands

teolemon:
### What 
- Add more organic brands to feed https://github.com/openfoodfacts/openfoodfacts-server/issues/3426
### Part of 
- #5520


stephanegigandet (approved):
We can add those entries to this file, but when we do taxonomize brands, we will need to review each entry of this file one by one before adding them to the production brands taxonomy.

Issue #5591, opened 2021-09-01
Serve packaging shapes icons


teolemon:
### Who for
- Developers
- Users
### What

- [ ] Serve packaging shapes icons
- [ ] Add a mapping to the packaging shapes taxonomy

PR #5627, opened 2021-09-07
icons from the French CITEO national recycling system (dark and light mode)

teolemon:
### What
- Icons from the CITEO national recycling system (dark and light mode)
- Mapping to the packaging shapes will be done by a separate file, so that we can keep the granularity of the shapes (there are several icons for a same shape name)
- Attribution © French CITEO national recycling system is necessary

image

### Required by
- #5591
- #5525


stephanegigandet:
We should add a README and LICENCE files to acknowledge the source (ideally with a link to it) and warn that those icons are the property of CITEO. We probably need similar files as well for labels logos too.

Where do the files come from? Straight from CITEO, or there was some processing applied to them, or to their file path / file names?

teolemon:
@stephanegigandet done. I used the PDF mentioned in the README, but CITEO also has a large Adobe Illustrator file (also mentioned).
I created the dark icons with recoloring. It would be nice for the knowledge panel API to support dark mode.

stephanegigandet (approved):

Issue #5633, opened 2021-09-08
Display knowledge panels on the OFF website


stephanegigandet:
### Story

- Just like product attributes, knowledge panels are a very good way to build user features that are shared between the OFF web site, the OFF mobile apps, and even 3rd party apps.
- The Knowledge Panel API + a few panels are currently being built server side (see #5541 for the Eco-Score panel), and the knowledge panel display is being implemented in the new Flutter mobile app (Smoothie).
- Being able to also show knowledge panels on the OFF website will ease the development of new panels and make the panels highly visible to users.

### Part of
- #5493

stephanegigandet:
Panels are deployed on the web, currently for moderators.

PR #5634, opened 2021-09-08
Display knowledge panels on the web

stephanegigandet:
- See #5633

This PR:
- shows the Eco-Score panel next to the Eco-Score on product page when ?panels=1 is added to the URL.
- implements some basic functions to display panels and subpanels (currently just the text and table elements)

At this point, it is mostly for demo and development purposes, but the goal is to eventually replace the current Eco-Score + Eco-Score details display with the Eco-Score knowledge panel.

This is just a sample implementation to display the panels structure and data, the actual implementation of how we display panels on the web might be very different (e.g. using accordions, dropdowns, or modals to display panels and subpanels).

### DIFF
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -369,20 +369,22 @@ sub create_ecoscore_panel($$$) {
 
         foreach my $adjustment ("production_system", "origins_of_ingredients", "threatened_species", "packaging") {
 
+            my $plus_or_minus;
+
             if (not defined $product_ref->{ecoscore_data}{adjustments}{$adjustment}{value}) {
-                $grade = "unknown";
+                $plus_or_minus = "unknown";
             }
             elsif ($product_ref->{ecoscore_data}{adjustments}{$adjustment}{value} < 0) {
-                $grade = "minus";
+                $plus_or_minus = "minus";
             }
             else {
-                $grade = "plus";
+                $plus_or_minus = "plus";
             }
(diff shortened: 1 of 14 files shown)


stephanegigandet:
Sample URL:

https://world.openfoodfacts.dev/product/3017620422003/nutella-ferrero?panels=1

Current look:

image

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -369,20 +369,22 @@ sub create_ecoscore_panel($$$) {
 
         foreach my $adjustment ("production_system", "origins_of_ingredients", "threatened_species", "packaging") {
 
+            my $plus_or_minus;
```
jasmeet0817, line 372: difficult to follow, what this is from reading the code, could you add comments ?

jasmeet0817 (approved):
Looks good

lib/ProductOpener/KnowledgePanels.pm
stephanegigandet: @jasmeet0817 : right, I changed a few things in the API and forgot to point it out.

I separated the "plus" and "minus" values I initially planned to put in the "grade" field. But then we would have essentially one field for two different thing: an actual grade "a" to "e", and then "plus" or "minus".

So I made a new field, which I called "plus_or_minus" because I failed to find a better name. :) 

I will in fact remove this code from KnowledgePanel.pm and in fact set the value directly in the JSON template for each panel.

Issue #5640, opened 2021-09-09
Some RTL Glitches


yarons:
Unicode can handle mixed LTR and RTL text up to a certain level, above that there should be some intervention in terms of modifying some "left" or "right" CSS selectors or adding some directionality characters.

<img width="358" alt="Screen Shot 2021-08-04 at 10 48 25" src="https://user-images.githubusercontent.com/406826/128684864-af0f97df-4c56-4c0c-9a59-21c626438a08.png">
(Notice the grams out of grams line, it's incomprehensible).

<img width="578" alt="Screen Shot 2021-08-09 at 12 25 42" src="https://user-images.githubusercontent.com/406826/128685467-e114c08b-f0a5-4c34-af0d-7d94b4eb3d8d.png">
(The image overlaps the list, there's some CSS selector there that need some modification).

yarons:
How can we make sure it remains open?

teolemon:
It will. It's just the bot reminding us that no one has had RTL in his thoughts for a while 👼

PR #5655, opened 2021-09-11
i18n: Make 'X g / 100 g' translatable

hangy:
**Description:** Introduces a new translation key `g_per_100g`, which can be used to translate the text `'%.2f g / 100 g'` into other languages. This is useful if a language uses other formatting, ie. RTL languages (as opposed to LTR languages).

- **Related issues and discussion:** #5640


teolemon (approved):
LGTM

PR #5668, opened 2021-09-13
Improvements for Eco-Score knowledge panels #5541

stephanegigandet:
Improvements to the Eco-Score panels.

- Enhance the backticks feature to escape quotes in JSON templates
- Allow comments in the JSON templates
- Move the logic to decide if an adjustment is a plus or minus directly in the template for each adjustment.
- Add some warnings when packaging info is missing.
- Added origins of ingredients table
- Support for colored percent bars

### DIFF
--- a/lib/ProductOpener/Ecoscore.pm
+++ b/lib/ProductOpener/Ecoscore.pm
@@ -1407,6 +1407,20 @@ sub localize_ecoscore ($$) {
 			
 			$product_ref->{ecoscore_data}{adjustments}{origins_of_ingredients}{"transportation_value"}
 			= $product_ref->{ecoscore_data}{adjustments}{origins_of_ingredients}{"transportation_value_" . $cc};
+
+			# For each origin, we also add its score (EPI + transporation to country of request)
+			# so that clients can show which ingredients contributes the most to the origins of ingredients bonus / malus
+
+			if (defined $product_ref->{ecoscore_data}{adjustments}{origins_of_ingredients}{aggregated_origins}) {
+			
+				foreach my $origin_ref (@{$product_ref->{ecoscore_data}{adjustments}{origins_of_ingredients}{aggregated_origins}}) {
+		
+					my $origin_id = $origin_ref->{origin};
+					$origin_ref->{epi_score} = $ecoscore_data{origins}{$origin_id}{epi_score};
+					$origin_ref->{transportation_score} = $ecoscore_data{origins}{$origin_id}{"transportation_score_" . $cc};
+				}
+			}
+
 		}
 	}		
 	
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -163,6 +163,23 @@ sub create_knowledge_panels($$$$) {
 }
 
 
+=head2 convert_multiline_string_to_singleline($line)
+
+Helper function to allow to enter multiline strings in JSON templates.
+The function converts the multiline string into a single line string.
+
+=cut
+
+sub convert_multiline_string_to_singleline($) {
+    my $line = shift;
+    $line =~ s/\n/\\n/sg;
+    # Escape quotes unless they have been escaped already
+    # negative look behind to not convert \" to \\"
+    $line =~ s/(?<!\\)"/\\"/g;
+    return '"' . $line . '"';
+}
+
+
 =head2 create_panel_from_json_template ( $panel_id, $panel_template, $panel_data_ref, $product_ref, $target_lc, $target_cc )
 
 Creates a knowledge panel from a JSON template.
@@ -243,15 +260,6 @@ sub create_panel_from_json_template ($$$$$$) {
         # e.g. when we want to generate HTML
 
         # Also escape quotes " to \"
-
-        sub convert_multiline_string_to_singleline($) {
-            my $line = shift;
-            $line =~ s/\n/\\n/sg;
-            # Escape quotes unless they have been escaped already
-            # negative look behind to not convert \" to \\"
-            $line =~ s/(?<!\\)"/\\"/g;
-            return '"' . $line . '"';
-        }
 
         $panel_json =~ s/\`([^\`]*)\`/convert_multiline_string_to_singleline($1)/seg;
 
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -227,9 +227,14 @@ sub create_panel_from_json_template ($$$$$$) {
         # In the template, we use multiline strings for readability
         # e.g. when we want to generate HTML
 
+        # Also escape quotes " to \"
+
         sub convert_multiline_string_to_singleline($) {
             my $line = shift;
             $line =~ s/\n/\\n/sg;
+            # Escape quotes unless they have been escaped already
+            # negative look behind to not convert \" to \\"
+            $line =~ s/(?<!\\)"/\\"/g;
             return '"' . $line . '"';
         }
 
@@ -365,31 +370,15 @@ sub create_ecoscore_panel($$$) {
         create_panel_from_json_template("ecoscore_agribalyse", "api/knowledge-panels/ecoscore/agribalyse.tt.json",
             $panel_data_ref, $product_ref, $target_lc, $target_cc);
 
-        # TODO: add panels for the different bonuses and maluses
+        # Add panels for the different bonuses and maluses
 
         foreach my $adjustment ("production_system", "origins_of_ingredients", "threatened_species", "packaging") {
 
-            my $plus_or_minus;
-
-            if (not defined $product_ref->{ecoscore_data}{adjustments}{$adjustment}{value}) {
-                $plus_or_minus = "unknown";
-            }
-            elsif ($product_ref->{ecoscore_data}{adjustments}{$adjustment}{value} < 0) {
-                $plus_or_minus = "minus";
-            }
-            else {
-                $plus_or_minus = "plus";
-            }
-
-            $title = lang("ecoscore_" . $adjustment);
-
-            $panel_data_ref = {
-                "plus_or_minus" => $plus_or_minus,
-                "title" => $title,
+            my $adjustment_panel_data_ref = {
             };            
 
             create_panel_from_json_template("ecoscore_" . $adjustment, "api/knowledge-panels/ecoscore/" . $adjustment . ".tt.json",
-                $panel_data_ref, $product_ref, $target_lc, $target_cc);
+                $adjustment_panel_data_ref, $product_ref, $target_lc, $target_cc);
         }
 
 	}
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -169,6 +169,16 @@ Creates a knowledge panel from a JSON template.
 The template is passed both the full product data + optional panel specific data.
 The template is thus responsible for all the display logic (what to display and how to display it).
 
+Some special features that are not included in the JSON format are supported:
+
+1. Multiline strings can be included using backticks ` at the start and end of the multinine strings.
+- The multiline strings will be converted to a single string.
+- Quotes " are automatically escaped unless they are already escaped
+
+2. Comments can be included by starting a line with //
+- Comments will be removed in the resulting JSON, they are only intended to make the source template easier to understand.
+
+
 =head3 Arguments
 
 =head4 panel id $panel_id
@@ -222,6 +232,11 @@ sub create_panel_from_json_template ($$$$$$) {
     else {
 
         # Turn the JSON to valid JSON
+
+        # Remove comment lines starting with //
+        # comments are not allowed in JSON, but they can be useful to have in the templates source
+        # /m modifier: ^ and $ match the start and end of each line
+        $panel_json =~ s/^(\s*)\/\/(.*)$//mg;
 
         # Convert multilines strings between backticks `` into single line strings
         # In the template, we use multiline strings for readability
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -5696,6 +5696,22 @@ msgctxt "ecoscore_packaging_missing_information"
 msgid "The information about this product's packaging is missing."
 msgstr "The information about this product's packaging is missing."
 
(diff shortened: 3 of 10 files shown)


stephanegigandet:
Knowledge panels now support colored percent bars:

image

https://fr-en.openfoodfacts.dev/product/00434034/swiss-chocolate-extra-fine-milk-marks-spencer?panels=1

jasmeet0817:
Do you have an example of the output json, kind of hard to estimate it from the code

stephanegigandet:
> Do you have an example of the output json, kind of hard to estimate it from the code

@jasmeet0817 here is one https://fr-en.openfoodfacts.dev/api/v2/product/00434034/swiss-chocolate-extra-fine-milk-marks-spencer?fields=knowledge_panels (login and pass : off )

The JSON for the table with colored percent bars looks like this:

```
ecoscore_origins_of_ingredients: {
elements: [
{
element: {
headers: [
{
text: "Origin"
},
{
text: "% of ingredients"
},
{
text: "Impact"
}
],
rows: [
{
values: [
{
text: "France"
},
{
percent: 50,
percent_color: "green",
text: "50 %"
},
{
color: "green",
text: "Low"
}
]
},
```

jasmeet0817:
See comments inline

> > Do you have an example of the output json, kind of hard to estimate it from the code
> 
> @jasmeet0817 here is one https://fr-en.openfoodfacts.dev/api/v2/product/00434034/swiss-chocolate-extra-fine-milk-marks-spencer?fields=knowledge_panels (login and pass : off )
> 
> The JSON for the table with colored percent bars looks like this:
> 
> ```
> ecoscore_origins_of_ingredients: {
> elements: [
> {
> element: {
> headers: [
> {
> text: "Origin"
> },

will there be metadata for each header? otherwise we could just have a list like `["origin", "% of ingredients", "impact"]`

if there will be other meta then looks fine, maybe rename `text` to something more specific like `header_title`/`header_label` etc.


> {
> text: "% of ingredients"
> },
> {
> text: "Impact"
> }
> ],
> rows: [
> {
> values: [

do we need the key `values`? 
can it be rows: [{row1..}, {row2...}]

> {
> text: "France"
> },
> {
> percent: 50,

let's add a `type` to each row? here `type` can be `"percent"`. this way the client knows how to parse the field.

> percent_color: "green",

maybe just `color` ? with `type = "percent"` we don't need to specify that this is `percent_color`

> text: "50 %"

can we rename `text` to `label`

> },
> {
> color: "green",

By passing "green" we are mildly breaking the principal that the server should not decide how to render something. I feel like the color that should be used here should be decided by the client (as per their color scheme). 

I wonder if it's possible to pass something else here that indicates this row value is "good", coz I think that's what we are trying to communicate with the color here, "green" = "good" and "red" = "bad", and then client can decide to use whichever color they want for "good" (which will most likely be green) but also they may want to do something else for "good" rows, like highlight them/shadow them/italicize/bold etc, by passing "good" vs "bad" as opposed to "green" vs "red" we give more power to the client. 

> text: "Low"
> }
> ]
> },
> ```

stephanegigandet:
Thanks for all the comments @jasmeet0817 

> > headers: [
> > {
> > text: "Origin"
> > },
> > ```
> 
> will there be metadata for each header? otherwise we could just have a list like `["origin", "% of ingredients", "impact"]`
> 
> if there will be other meta then looks fine, maybe rename `text` to something more specific like `header_title`/`header_label` etc.

It's likely that we will add other column specific fields at some point. e.g. below you suggest adding a "percent" type to a row, but it should probably be a type for the column. We could also have colors, icons etc. like for each cell value.

> do we need the key `values`?
> can it be rows: [{row1..}, {row2...}]

I think an array / list makes it easier for clients to iterate without having to construct key names. Having structured data is why we use JSON, otherwise we could just make everything flat and have keys like "panel_1_element_2_row_4_colum_2"... 

> > {
> > text: "France"
> > },
> > {
> > percent: 50,
> 
> let's add a `type` to each row? here `type` can be `"percent"`. this way the client knows how to parse the field.

So that would be adding a "type" for each colum, not row, right?

> > percent_color: "green",
> 
> maybe just `color` ? with `type = "percent"` we don't need to specify that this is `percent_color`

We already have a key "color" that is applied to the "text" of the value, that's why I made a separate key for the percent bar.

> > text: "50 %"
> 
> can we rename `text` to `label`

I don't know, 'text' sounds more descriptive to me. We might have tables with images in some cells instead of text, in which case we would use an 'image_url' property.

> > },
> > {
> > color: "green",
> 
> By passing "green" we are mildly breaking the principal that the server should not decide how to render something. I feel like the color that should be used here should be decided by the client (as per their color scheme).
> 
> I wonder if it's possible to pass something else here that indicates this row value is "good", coz I think that's what we are trying to communicate with the color here, "green" = "good" and "red" = "bad", and then client can decide to use whichever color they want for "good" (which will most likely be green) but also they may want to do something else for "good" rows, like highlight them/shadow them/italicize/bold etc, by passing "good" vs "bad" as opposed to "green" vs "red" we give more power to the client.

OK, we could use the same "plus_or_minus" property we have for panels, with the same values "plus", "neutral", "minus", "unknown".

stephanegigandet:
I will rename "plus_or_minus" to "evaluation", with values "good", "neutral", "bad", "unknown".

jasmeet0817:
> Thanks for all the comments @jasmeet0817
> 
> > > headers: [
> > > {
> > > text: "Origin"
> > > },
> > 
> > 
> > will there be metadata for each header? otherwise we could just have a list like `["origin", "% of ingredients", "impact"]`
> > if there will be other meta then looks fine, maybe rename `text` to something more specific like `header_title`/`header_label` etc.
> 
> It's likely that we will add other column specific fields at some point. e.g. below you suggest adding a "percent" type to a row, but it should probably be a type for the column. We could also have colors, icons etc. like for each cell value.

> 
> > do we need the key `values`?
> > can it be rows: [{row1..}, {row2...}]
> 
> I think an array / list makes it easier for clients to iterate without having to construct key names. Having structured data is why we use JSON, otherwise we could just make everything flat and have keys like "panel_1_element_2_row_4_colum_2"...

Not sure I follow what you mean, I'm completely in favor of as much structure as humanly possible, but the question is whether we need an array here or a map. 

Current proposal is rows: [{values: [{row1_map}, {row2_map}, ...]}]
I'm proposing rows: [{row1_map}, {row2_map}, ...]

A map doesn't necessarily add more structure, in this case if we use an array it's assumed each element of the row is a "value". Not sure we gain anything from a map in this case unless there are plans to add metadata  in the map, but since this metadata would apply to all rows, I would presume it's the metadata for the table itself.

> 
> > > {
> > > text: "France"
> > > },
> > > {
> > > percent: 50,
> > 
> > 
> > let's add a `type` to each row? here `type` can be `"percent"`. this way the client knows how to parse the field.
> 
> So that would be adding a "type" for each colum, not row, right?

hmm good idea, just to confirm: we will have a column field that describes the structure of the table?

something like: 
 columns {
  [{type: text, type: percent, type: somethingelse}]
}

and then rows 
rows: [{row1_map}, {row2_map}, {row3_map}]

and then it's understood that row1 is of type text, row2 is of type percent and so on....

That's a good idea, specially because it roughly follows how how we define tables in databases (a proven way of representing tables)
> ​
> > > percent_color: "green",
> > 
> > 
> > maybe just `color` ? with `type = "percent"` we don't need to specify that this is `percent_color`
> 
> We already have a key "color" that is applied to the "text" of the value, that's why I made a separate key for the percent bar.


Why don't we just reuse 'color'? the client knows 'color' is applied on the row.

> 
> > > text: "50 %"
> > 
> > 
> > can we rename `text` to `label`
> 
> I don't know, 'text' sounds more descriptive to me. We might have tables with images in some cells instead of text, in which case we would use an 'image_url' property.

I'm just wondering if we could eventually have some row/header in which we want to pass a note or some other form of text in which case the name `text` becomes dubious. Since this is basically the `label` of the row I thought it's more specific to call it that (same for the comment on "header" field)

stephanegigandet:
@jasmeet0817 I replaced the "plus_or_minus", "color" and "percent_color" properties by a single property "evaluation" which can be "good", "neutral", "bad" or "unknown", that way as you suggested, the client can better decide how display good and bad things.

jasmeet0817:
> Thanks for all the comments @jasmeet0817
> 
> > > headers: [
> > > {
> > > text: "Origin"
> > > },
> > 
> > 
> > will there be metadata for each header? otherwise we could just have a list like `["origin", "% of ingredients", "impact"]`
> > if there will be other meta then looks fine, maybe rename `text` to something more specific like `header_title`/`header_label` etc.
> 
> It's likely that we will add other column specific fields at some point. e.g. below you suggest adding a "percent" type to a row, but it should probably be a type for the column. We could also have colors, icons etc. like for each cell value.
 
> 
> > do we need the key `values`?
> > can it be rows: [{row1..}, {row2...}]
> 
> I think an array / list makes it easier for clients to iterate without having to construct key names. Having structured data is why we use JSON, otherwise we could just make everything flat and have keys like "panel_1_element_2_row_4_colum_2"...
 
Not sure I follow, I'm completely in favor of as much structure as possible, but the question is whether we need an array here or a map. Both should be able to communicate the same information in a structured way
 
Current proposal is -> rows: [{values: [{cell1}, {cell2}}, {values: {cell3}, {cell4}}]
I'm proposing rows: [[{cell1}, {cell2}], [{cell3}, {cell4}]]

In my proposal it's assumed that [{cell1}, {cell2}] is row1 and [{cell3}, {cell4}] is row2. Not sure we gain anything from a map in this case unless there are plans to add metadata  in the map, but since this metadata would apply to all rows, I would presume it's the metadata for the table itself.
 
> 
> > > {
> > > text: "France"
> > > },
> > > {
> > > percent: 50,
> > 
> > 
> > let's add a `type` to each row? here `type` can be `"percent"`. this way the client knows how to parse the field.
> 
> So that would be adding a "type" for each colum, not row, right?
 
hmm good idea, just to confirm: we will have a column field that describes the structure of the table?
 
something like: 
 columns {
  [{type: text, type: percent, type: somethingelse}]
}
 
and then rows 
rows:  [[{cell1}, {cell2}, {cell3}], [{cell4}, {cell5}, {cell6}]]
 
and then it's understood that cell1, cell4 are of type `text`, cell2, cell5 of type `percent` and so on....
 
That's a good idea, specially because it roughly follows how how we define tables in databases (a proven way of representing tables)
> ​
> > > percent_color: "green",
> > 
> > 
> > maybe just `color` ? with `type = "percent"` we don't need to specify that this is `percent_color`
> 
> We already have a key "color" that is applied to the "text" of the value, that's why I made a separate key for the percent bar.
 
 
Why don't we just reuse 'color'? the client knows 'color' is applied on the cell.
 
> 
> > > text: "50 %"
> > 
> > 
> > can we rename `text` to `label`
> 
> I don't know, 'text' sounds more descriptive to me. We might have tables with images in some cells instead of text, in which case we would use an 'image_url' property.
 
I'm just wondering if we could eventually have some row/header in which we want to pass a note or some other form of text in which case the name `text` becomes dubious. Since this is basically the `label` of the row I thought it's more specific to call it that (same for the comment on "header" field)

stephanegigandet:
> Current proposal is -> rows: [{values: [{cell1}, {cell2}}, {values: [{cell3}, {cell4}]}]
> I'm proposing rows: [[{cell1}, {cell2}], [{cell3}, {cell4}]]

So in fact we have metadata on rows already, e.g. in the Eco-Score impact by stage table, we have ids for rows.

id: "ecoscore_lca_impacts_by_stages_table",
rows: [
{
id: "agriculture",
values: [
{
icon_url: "https://static.openfoodfacts.dev/images/icons/dist/agriculture.svg",
text: "Agriculture"
},
{
percent: 82.6683253659907,
text: "82.7 %"
}
]
},

stephanegigandet:
> hmm good idea, just to confirm: we will have a column field that describes the structure of the table?

I was just thinking to use the existing "headers" field and add a "type" to it:

headers: [
{
text: "Etape"
},
{
text: "Impact",
type: "percent"
}
],

I guess we could just rename "headers" to "columns".

jasmeet0817:
> I guess we could just rename "headers" to "columns".

Yes I think that would work

jasmeet0817:
> {
> percent: 82.6683253659907,
> text: "82.7 %"
> }



> > Current proposal is -> rows: [{values: [{cell1}, {cell2}}, {values: [{cell3}, {cell4}]}]
> > I'm proposing rows: [[{cell1}, {cell2}], [{cell3}, {cell4}]]
> 
> So in fact we have metadata on rows already, e.g. in the Eco-Score impact by stage table, we have ids for rows.
> 
> id: "ecoscore_lca_impacts_by_stages_table",
> rows: [
> {
> id: "agriculture",

hmm. that makes sense, thanks.

jasmeet0817:
> > hmm good idea, just to confirm: we will have a column field that describes the structure of the table?
> 
> I was just thinking to use the existing "headers" field and add a "type" to it:
> 
> headers: [
> {
> text: "Etape"

can you also add a type here, type could just be "text" but better to be explicit then leave it upto client's interpretation.

> },
> {
> text: "Impact",
> type: "percent"
> }
> ],
> 
> I guess we could just rename "headers" to "columns".

stephanegigandet:
@jasmeet0817 I renamed "headers" to "columns" and added a "type" property for all columns.

ok to merge?

stephanegigandet:
@teolemon:

> * [ ]  "Transportation and origins of ingredients have a medium impact." >> We lack context, unless it's correlated with 2,5%

It is not correlated to the Agribalyse average for the category. We could add "Transportation and origins of ingredients for this product have a medium impact" to make it more clear, but it becomes a bit long.

Another option is not to talk about impact, but more about locality: "Most ingredients of this product are from France", or "76% of ingredients of this product are from Europe" or something. We could make it even less abstract: "Average distance traveled by ingredients: 850 kilometers".

But that's a separate discussion, the aim of this PR is to bring us closer to serve the current Eco-Score details through a knowledge panel.

> * [ ]  For packaging, even if it has a low impact, we should make clear that recycling is NOT just an option

I don't think anyone would think "low packaging impact" is a licence for throwing the packaging in any garbage bin, but just like for the origins, the wording / messaging should be discussed, the current titles are just default titles until we brainstorm the messages.

jasmeet0817 (approved):
Looks good, thanks

PR #5677, opened 2021-09-14
Fix daily MongoDB task + add refresh_products_tags.js cron

ocervell:
This makes sure that the `mongorestore` command will not go OOM. 

Proxmox VM size has been increased from 4GB to 16GB since the process takes at least 14GB of RAM to run. Disk space has also been increased by ~60GB.

Also: 
- [x] fix for the data directory mount for MongoDB, and set restart policy to always for all deployments (advised for prod deployments).
- [x] add cron for `refresh_products_tags.js` script that creates the MongoDB collection `products_tags`

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -27,8 +27,8 @@ jobs:
 
   dev-db-sync:
     runs-on: ubuntu-latest
-    environment: off-net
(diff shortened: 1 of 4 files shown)


stephanegigandet (approved):
Looks good, thank you!

PR #5722, opened 2021-09-21
fix: solve Config.pm Config2.pm prod conflict with Docker configs

ocervell:
- [x] Rename `Config.pm` --> `Config_docker.pm` and `Config2.pm` to `Config2_docker.pm`
- [x] Add symlinks of `Config_docker.pm` --> `Config.pm` and `Config2_docker.pm` --> `Config2.pm`


stephanegigandet (approved):
Thank you!

PR #5727, opened 2021-09-21
fix: change local domain to openfoodfacts.localhost

ocervell:
Change `productopener.localhost` to `openfoodfacts.localhost` to guarantee operability between local installations of Robotoff and Product Opener (Robotoff need them to run on the same domain otherwise requests are rejected..)


stephanegigandet (approved):

teolemon:
merged, thanks @ocervell

PR #5744, opened 2021-09-22
ci: check for link rot in static resources

teolemon:
### What
- This aims at checking that no link nor image is broken in the content pages. 
- The action is currently stuck due to the sheer size of the arglist
- It should only be run periodically
- We could have another action that only runs on modified files

### Part of
- #5512


stephanegigandet:
This issue was fixed by @jnsereko in this PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/8232

So I'm closing this one.

PR #5780, opened 2021-09-29
Remove ingredients related fields when ingredients are empty

stephanegigandet:
This is to remove some fields for products that have an empty ingredient list, as those fields don't make sense for them.
e.g. the number of unknown ingredients is 0.


teolemon (approved):
Does it have any consequences for the API ?

stephanegigandet:
> Does it have any consequences for the API ?

It will remove some fields. If some clients expect those fields to be there, then they may break.

But all fields are optional, clients should not rely on any field being there.

PR #5789, opened 2021-09-29
Carbon impact knowledge panel with car driving km equivalence

stephanegigandet:
Start of the implementation of Tim's carbon impact knowledge panel
- API: new ecoscore_carbon_impact panel in the API
- API + web interface: some changes to color icons according to the evaluation of the panel (good, bad, average, neutral, unknown)

API changes:
- added a "name" field to the panel, to match the names just above the panel title
- added "icon_color_from_evaluation": true to indicate that the SVG icon needs to be colored. This makes it easy for clients to have different colors for light and dark modes. But we'll need to see if that can work in Flutter.

Current result:

image


greentoff:
I believe we currently show per 1kg of product which I think we should stick with

teolemon (approved):

PR #5792, opened 2021-09-30
test repair

aleene:
**Description:**
Repaired the test.

There is an issue with en:turmeric/kurkuma and en:E100/curcumine. The ingredients and additives interfere.

Need to move turmeric from ingredients to additives.

**Related issues and discussion:** #5791

### DIFF
--- a/t/additives.t
+++ b/t/additives.t
@@ -1474,7 +1474,7 @@ is(canonicalize_taxonomy_tag("fi", "additives", "sitruunahappo"), "en:e330");
 $product_ref = {
 	lc => "fi",
 	ingredients_text =>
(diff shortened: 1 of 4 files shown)


stephanegigandet (commented):
I don't think turmeric should be moved in E100, only curcumin is listed for E100 in https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=CELEX:02008R1333-20141106&rid=1

stephanegigandet:
e.g. for French, we have 783 products that list "curcuma" as an ingredient, but none that lists it as colouring.

So we should keep turmeric as an ingredient, and curcumine as an additive. And there should be no overlap between their synonyms.

aleene:
Is indeed better. However I guess that most turmeric is just added a colorant. And it does contain some curcumine. So if we say E100, it is not wrong.

stephanegigandet (approved):
Thanks

PR #5794, opened 2021-10-01
More ciqual and ingredients updates

aleene:
COntinuing my ingredients cleanup and ciqual integrations.

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -48267,21 +48267,45 @@ ciqual_food_name:fr:Chou chinois ou pak-choi ou pé-tsai, cru
 # description:en:Bok choy, pak choi or pok choi (Chinese:上海青; pinyin:Shànghǎi qīng; literally:"Shanghai Green"; Brassica rapa subsp. chinensis) is a type of Chinese cabbage.
 
 <en:cabbage
-en:pak choi, Bok choy, pok choi
-de:Pak Choi
-el:Μποκ τσόι
+en:pak choi, bok choy, pok choi
+ar:ملفوف صيني
+bo:ཚལ་དཀར་ཆེ་བ།
+ca:pak choi, bleda xinesa
+cs:bok čoj, pak čoj zelí, čínské
+de:Pak Choi, Chinesischer Senfkohl
+el:mποκ τσόι
+es:bok choy, pak choi
+eu:txinatar aza
+fa:کلم چینی
 fi:pinaattikiinankaali, pak choi, bok choy, paksoi, pak-soi
-fr:Bok choy, Pak choï
+fr:bok choy, pak choï, chou de Chine
+gl:col chinesa, coia chinesa, repolo chinés
+hu:bordáskel, bok choy, pak choi, bok choi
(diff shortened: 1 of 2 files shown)


taxonomies/ingredients.txt
```diff
@@ -3243,7 +3243,7 @@ wikipedia:en:https://en.wikipedia.org/wiki/Beeswax
 # usage:fr:agent antiadhérant:cire d'abeilles
 
 <en:beeswax
-en:white and yellow beeswax
+en:white and beeswax
```
stephanegigandet, line 3246: that change seems incorrect, why remove "yellow"?

taxonomies/ingredients.txt
```diff
@@ -53714,12 +53903,26 @@ zh:羽扇豆属, 羽扇豆, 鲁冰花
...
-# ingredient/fr:lupin has 390 products @2019-04-11
-# usage:fr:lupins (lupine, sel de cuisine, antioxydants:acide citrique et acide ascorbique, conservateur:E202)
-
-
+ciqual_food_code:en:20534
+ciqual_food_name:en:lupin grain, raw
+ciqual_food_name:fr:Lupin, graine crue
+# 957 in 12 @2021-10-01
+
+#processing:en: en:flour
```
stephanegigandet, line 53914: I think we should keep the "(something) flour" entries as separate ingredients.

taxonomies/ingredients.txt
aleene: no idea why that changed

taxonomies/ingredients.txt
aleene: we can, but why? If it is a separate ciqual I do keep the entry

taxonomies/ingredients.txt
```diff
@@ -3989,13 +3989,13 @@ wikidata:en:Q191381
 wikipedia:en:https://en.wikipedia.org/wiki/Saccharin
 # ingredient/fr:saccharinate-de-sodium has 30 products in french @2019-01-30
```
teolemon, line 3990:
```suggestion
# ingredient/fr:saccharinate-de-sodium has 30 products in french @2019-01-30
ciqual_food_code:en:31064
ciqual_food_name:en:Saccharin sweetener
ciqual_food_name:fr:Édulcorant à la saccharine
```

aleene: These are now in additives

taxonomies/ingredients.txt
stephanegigandet: It makes it possible to have "flour" as a parent, and to find foods that contain "(something) flour" (and not all the foods containing "(something)".

For the processings, I think we should include only processings that don't transform the ingredient into something radically different. e.g. "cut something", "dried something" etc. "something juice" is probably fine too. But for "something flour" or "something extract" for instance, I think it makes sense to keep the distinction.

stephanegigandet (approved):

taxonomies/ingredients.txt
aleene: subtle, but ok

PR #5795, opened 2021-10-01
feat: new taxonomy API

### LINKED ISSUE #5785: Create an API to get an extract of the taxonomy for specific tags and properties

We currently have JSON exports of categories like this one: https://world.openfoodfacts.org/data/taxonomies/packaging_materials.json

But for taxonomies like categories, ingredients, additives etc. those exports can be huge, as they contain all tags and all names, synonyms, descriptions etc. in all languages.

For some uses, it would be great to have an API with only the tags and fields needed for the use.

For instance, for a category chooser on a mobile app (such as @gspencergoog 's category choose for the Flutter app), the app could request only the children of a specific tag, in only one language.

Create an API to get only specific tags, with only specific fields, in only specific languages.

e.g. something like /api/v2/tags?type=categories&tags=en:fruits,en:vegetables&fields=name,description,parents,children&lc=en,fr&include_children=1

It would return fields in the same format as the exports (e.g. https://world.openfoodfacts.org/data/taxonomies/packaging_materials.json ) but with only the requested data.

include_children and include_parents would indicate that we want to retrieve data for children and parents (so that we can have all local names of all children in a single API call).

This would also be a generic way to solve #5770 

_No response_

### Additional context

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
Fixes #5785 

This is a new API to get an extract of a taxonomy for specific tags, specific fields/properties, in specific languages.

example: https://world.openfoodfacts.dev/api/v2/taxonomy?type=labels&tags=en:organic,en:fair-trade&fields=name,description,children&include_children=1&lc=en,fr

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -822,8 +822,8 @@ sub analyze_request($)
 		$request_ref->{search} = 1;
 	}
 
-	# /tags API endpoint
-	# e.g. /api/v2/tags?type=categories&tags=en:fruits,en:vegetables&fields=name,description,parents,children&lc=en,fr&include_children=1
+	# /taxonomy API endpoint
+	# e.g. /api/v2/taxonomy?type=categories&tags=en:fruits,en:vegetables&fields=name,description,parents,children,vegan:en,inherited:vegetarian:en&lc=en,fr&include_children=1
 	elsif ($components[0] eq "taxonomy") {
 		$request_ref->{taxonomy} = 1;
 	}	
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -82,6 +82,7 @@ BEGIN
 		&search_and_map_products
 		&display_recent_changes
 		&add_tag_prefix_to_link
+		&display_taxonomy_api
 
 		&display_ingredient_analysis
 		&display_ingredients_analysis_details
@@ -563,6 +564,9 @@ sub init()
 			@lcs = ($lc);
 		}
 	}
+	else {
+		@lcs = ($lc);
+	}
 	# change the subdomain if we have overrides so that links to product pages are properly constructed
 	if ($cc_lc_overrides) {
 		$subdomain = $cc;
@@ -817,7 +821,13 @@ sub analyze_request($)
 	elsif ($components[0] eq "search") {
 		$request_ref->{search} = 1;
 	}
-	
+
+	# /tags API endpoint
+	# e.g. /api/v2/tags?type=categories&tags=en:fruits,en:vegetables&fields=name,description,parents,children&lc=en,fr&include_children=1
+	elsif ($components[0] eq "taxonomy") {
+		$request_ref->{taxonomy} = 1;
+	}	
+
 	# /products endpoint (e.g. /products/8024884500403+3263855093192 )
 	# assign the codes to the code parameter
 	elsif ($components[0] eq "products") {
@@ -9944,7 +9954,7 @@ sub display_preferences_api($$)
 }
 
 
-=head2 display_attribute_groups_api ( $target_lc )
+=head2 display_attribute_groups_api ( $request_ref, $target_lc )
 
 Return a JSON structure with all available attribute groups and attributes,
 with strings (names, descriptions etc.) in a specific language,
@@ -9996,8 +10006,49 @@ sub display_attribute_groups_api($$)
 }
 
 
-sub display_product_api($)
-{
+=head2 display_taxonomy_api ( $request_ref )
+
+Generate an extract of a taxonomy for specific tags, fields and languages,
+and return it as a JSON object.
+
+Accessed through the /api/v2/taxonomy API
+
+e.g. https://world.openfoodfacts.org/api/v2/taxonomy?type=labels&tags=en:organic,en:fair-trade&fields=name,description,children&include_children=1&lc=en,fr
+
+=head3 Arguments
+
+=head4 request object reference $request_ref
+
+=cut
+
+sub display_taxonomy_api($) {
+
+	my $request_ref = shift;
+
+	my $tagtype = param('type');
+	my $tags = param('tags');
+	my @tags = split(/,/, $tags);
+
+	my $options_ref = {};
+
+	foreach my $field (qw(fields include_children include_parents)) {
+		if (defined param($field)) {
+			$options_ref->{$field} = param($field);
+		}
+	}
+
+	my $taxonomy_ref = generate_tags_taxonomy_extract($tagtype, \@tags, $options_ref, \@lcs);
+
+	$request_ref->{structured_response} = $taxonomy_ref;
+
+	display_structured_response($request_ref);
+	
+	return;
+}
+
+
+sub display_product_api($) {
+
 	my $request_ref = shift;
 
 	my $code = normalize_code($request_ref->{code});
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -1814,7 +1814,7 @@ sub generate_tags_taxonomy_extract ($$$$) {
 
 		if (defined $properties{$tagtype}{$tagid}) {
 
-			foreach my $prop_lc (sort keys %{$properties{$tagtype}{$tagid}}) {
+			foreach my $prop_lc (keys %{$properties{$tagtype}{$tagid}}) {
 				
 				if ($prop_lc =~ /^(.*):(\w\w)$/) {
 					my $prop = $1;
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -1745,27 +1745,29 @@ sub generate_tags_taxonomy_extract ($$$$) {
 	}
 
 	# For the options include_children or include_parents,
-	# we will need to include data for more tags that requested.
+	# we will need to include data for more tags than requested.
 	# @tags will hold the tags to include
 
 	my @tags = ();
 	my %requested_tags = ();
+	my %included_tags = ();
 	foreach my $tagid (@$tags_ref) {
 		push @tags, $tagid;
 		$requested_tags{$tagid} = 1;
+		$included_tags{$tagid} = 1;
 	}
 
-	my $fields_ref;
-	my @properties = ();
+	my $include_all_fields = 0;
+	my $fields_ref = {};
+	my @inherited_properties = ();
 	if ((defined $options_ref) and (defined $options_ref->{fields})) {
-		$fields_ref = {};
 		foreach my $field (split(/,/, $options_ref->{fields}) ) {
 			# Compute a list of the requested inherited properties,
 			# as we will populate them directly
 			if ($field =~  /^inherited:(.*):(\w\w)$/) {
 				my $prop = $1;
 				my $lc = $2;
-				push @properties, [$prop, $lc];
+				push @inherited_properties, [$prop, $lc];
 			}
 			# Compute a hash of the other requested fields
 			# as we will go through all existing properties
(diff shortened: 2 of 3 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -817,7 +821,13 @@ sub analyze_request($)
 	elsif ($components[0] eq "search") {
 		$request_ref->{search} = 1;
 	}
-	
+
+	# /tags API endpoint
```
gspencergoog, line 825: Is it "tags", or "taxonomy"?

gspencergoog (commented):
LGTM, modulo the docs nit.

lib/ProductOpener/Display.pm
stephanegigandet: Good catch, it's taxonomy, I changed my mind half way through coding. I'll fix it.

stephanegigandet:
Added a way to request inherited properties:
https://world.openfoodfacts.dev/api/v2/taxonomy?type=ingredients&tags=en:beans&fields=name,inherited:vegan:en&lc=fr

(the vegan:en property is not set on en:beans, but on its parent)

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+	my $taxonomy_ref = {};
+
+	while (my $tagid = shift @tags) {
+
+		if (defined $direct_parents{$tagtype}{$tagid}) {
+			
+			$taxonomy_ref->{$tagid}{parents} = [];
+			foreach my $parentid (sort keys %{$direct_parents{$tagtype}{$tagid}}) {
+				if ((not defined $fields_ref) or (defined $fields_ref->{parents})) {
+					exists $taxonomy_ref->{$tagid}{parents} or $taxonomy_ref->{$tagid}{parents} = [];
```
alexgarel, line 1787: Is that necessary considering it's already initialized at line 1784 ?

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+
+	my $taxonomy_ref = {};
+
+	while (my $tagid = shift @tags) {
+
+		if (defined $direct_parents{$tagtype}{$tagid}) {
+			
+			$taxonomy_ref->{$tagid}{parents} = [];
+			foreach my $parentid (sort keys %{$direct_parents{$tagtype}{$tagid}}) {
+				if ((not defined $fields_ref) or (defined $fields_ref->{parents})) {
```
alexgarel, line 1786: I don't understand the first part of the condition ! Does that mean we want all fields ?
In this case I would define it in a variable before : `my $all_fields = not defined $fields_ref` it would be easier to understand in condition : `if ($all_fields or (defined $fields_ref->{parents}) {`

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+					exists $taxonomy_ref->{$tagid}{parents} or $taxonomy_ref->{$tagid}{parents} = [];
+					push @{ $taxonomy_ref->{$tagid}{parents} }, $parentid;
+				}
+				# Include parents if the tag is one of the initially requested tags
+				# so that we don't also add parents of parents
+				if (($options_ref->{include_parents}) and ($requested_tags{$tagid})) {
+					push @tags, $parentid;
```
alexgarel, line 1793: Hum it took me some time to get the condition, has it's not obvious that here we modify tag list. Maybe a comment here can help.

```suggestion
                    # add parent to list of tags to process, while leaving it outside of requested_tags
					push @tags, $parentid;
```

alexgarel: I see a potential bug here if some requested tags  are parents of others or requested tags have same parents, as we may put same tag more than once in @tags.

There is the same problem with children below.

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+		}
+
+		if (defined $direct_children{$tagtype}{$tagid}) {
+			
+			foreach my $childid (sort keys %{$direct_children{$tagtype}{$tagid}}) {
+				if ((not defined $fields_ref) or (defined $fields_ref->{children})) {
+					exists $taxonomy_ref->{$tagid}{children} or $taxonomy_ref->{$tagid}{children} = [];
+					push @{$taxonomy_ref->{$tagid}{children}}, $childid;
+				}
+				if (($options_ref->{include_children}) and ($requested_tags{$tagid})) {
```
alexgarel, line 1805: I think you could repeat comments from above here.

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+					push @tags, $childid;
+				}
+			}
+		}
+
+		if (((not defined $fields_ref) or (defined $fields_ref->{name}))
```
alexgarel, line 1811: there again all_fields variable would help readability.

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+			else {
+				$fields_ref->{$field} = 1;
+			}
+		}
+	}
+
+	my $taxonomy_ref = {};
+
+	while (my $tagid = shift @tags) {
+
```
alexgarel, line 1781:
```suggestion

		# handle parents field
```

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+				}
+				# Include parents if the tag is one of the initially requested tags
+				# so that we don't also add parents of parents
+				if (($options_ref->{include_parents}) and ($requested_tags{$tagid})) {
+					push @tags, $parentid;
+				}
+			}
+		}
+
+		if (defined $direct_children{$tagtype}{$tagid}) {
```
alexgarel, line 1798:
```suggestion

		# handle children field
```

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+				if ((not defined $fields_ref) or (defined $fields_ref->{children})) {
+					exists $taxonomy_ref->{$tagid}{children} or $taxonomy_ref->{$tagid}{children} = [];
+					push @{$taxonomy_ref->{$tagid}{children}}, $childid;
+				}
+				if (($options_ref->{include_children}) and ($requested_tags{$tagid})) {
+					push @tags, $childid;
+				}
+			}
+		}
+
```
alexgarel, line 1810:
```suggestion

		# handle name and synonyms fields
```

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+					# add synonyms to the full taxonomy
+					if (((not defined $fields_ref) or (defined $fields_ref->{synonyms}))
+						and (defined $synonyms_for{$tagtype}{$lc}{$lc_tagid})) {
+						(defined $taxonomy_ref->{$tagid}{synonyms}) or $taxonomy_ref->{$tagid}{synonyms} = {};
+						$taxonomy_ref->{$tagid}{synonyms}{$lc} = $synonyms_for{$tagtype}{$lc}{$lc_tagid};
+					}
+				}
+			}
+		}
+
```
alexgarel, line 1840:
```suggestion

		# handle properties
```

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+		if (defined $properties{$tagtype}{$tagid}) {
+
+			foreach my $prop_lc (keys %{$properties{$tagtype}{$tagid}}) {
+				
+				if ($prop_lc =~ /^(.*):(\w\w)$/) {
```
alexgarel, line 1845: it would be interesting to know why we do nothing if prop_lc does not match the pattern.

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+
+	my @tags = ();
+	my %requested_tags = ();
+	foreach my $tagid (@$tags_ref) {
+		push @tags, $tagid;
+		$requested_tags{$tagid} = 1;
+	}
+
+	my $fields_ref;
+	my @properties = ();
```
alexgarel, line 1759: Maybe it's normal in perl, but I'm a bit surprised that we use same name as `$properties`, I find it a bit confusing. We could also maybe make the intention more clear here by using a name like `@inherited_props`.

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+
+		# Allow requests for inherited properties
+		foreach my $property_ref (@properties) {
+
+			my $prop = $property_ref->[0];
+			my $lc = $property_ref->[1];
+			my $property_value = get_inherited_property($tagtype, $tagid, "$prop:$lc");
+			if (defined $property_value) {
+				(defined $taxonomy_ref->{$tagid}{$prop}) or $taxonomy_ref->{$tagid}{$prop} = {};
+					$taxonomy_ref->{$tagid}{$prop}{$lc} = $property_value;
```
alexgarel, line 1880:
```suggestion
                    # we don't have to check for a previous value because there is a clear spilt between field_ref and @properties
					$taxonomy_ref->{$tagid}{$prop}{$lc} = $property_value;
```

lib/ProductOpener/Tags.pm
```diff
@@ -1702,6 +1704,188 @@ sub build_tags_taxonomy($$$) {
...
+	}
+
+	my $fields_ref;
+	my @properties = ();
+	if ((defined $options_ref) and (defined $options_ref->{fields})) {
+		$fields_ref = {};
+		foreach my $field (split(/,/, $options_ref->{fields}) ) {
+			# Compute a list of the requested inherited properties,
+			# as we will populate them directly
+			if ($field =~  /^inherited:(.*):(\w\w)$/) {
```
alexgarel, line 1765: This means that in the case of inherited properties, we must provide language ?

If it's not intentional, we could avoid forcing it and it this case, use $lcs_ref to populate several entries in @properties.

lib/ProductOpener/Display.pm
```diff
@@ -817,7 +821,13 @@ sub analyze_request($)
 	elsif ($components[0] eq "search") {
 		$request_ref->{search} = 1;
 	}
-	
+
+	# /taxonomy API endpoint
+	# e.g. /api/v2/taxonomy?type=categories&tags=en:fruits,en:vegetables&fields=name,description,parents,children,vegan:en,inherited:vegetarian:en&lc=en,fr&include_children=1
+	elsif ($components[0] eq "taxonomy") {
+		$request_ref->{taxonomy} = 1;
+	}
```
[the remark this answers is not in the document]
alexgarel: In fact, if I test it locally, http://world.openfoodfacts.localhost/api/v2/taxonomy?type=labels&tags=en:organic,en:fair-trade&fields=name,description,children&include_children=1&lc=en,fr

I think it handles it as a product api call and return a 

```
{"code":null,"status":0,"status_verbose":"no code or invalid code"}
```

alexgarel (changes requested):
It seems the method can't be called right now !

lib/ProductOpener/Display.pm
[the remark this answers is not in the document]
stephanegigandet: Oops, I did forget to check in display.pl indeed. Fixed.

lib/ProductOpener/Tags.pm
stephanegigandet: Currently the only use we have for inherited properties is for properties where the value is not really language specific. We mostly use it for things like "vegan:en: en:yes". (and only the English value is defined). e.g. We want all children of fruits and vegetables to be vegan. That's why I didn't add support for matching multiple languages.

lib/ProductOpener/Tags.pm
stephanegigandet: That's right, tags could be included multiple times. In practice the result is a hash of tags, so it's not going to make a different result, but it is a bit sloppy. I'll add a structure to keep track of tags already included.

lib/ProductOpener/Tags.pm
stephanegigandet: Agreed

lib/ProductOpener/Tags.pm
stephanegigandet: Good idea.

lib/ProductOpener/Tags.pm
stephanegigandet: It always match the pattern. I guess we could remove the if and do it differently.

Something like this should work:

# properties are of the form [property_name]:[2 letter language code]
my ($prop, $lc) = split(/:/, $prop_lc)

lib/ProductOpener/Tags.pm
stephanegigandet: Agreed.

lib/ProductOpener/Tags.pm
stephanegigandet: In fact technically there could be a previous value, if the request is like fields=vegan:en&inherited_properties=vegan:en
But that would result in overriding the previous value with exactly the same value. (in fact the query would be identical to removing fields=vegan:en)

lib/ProductOpener/Tags.pm
stephanegigandet: added with other changes

lib/ProductOpener/Tags.pm
stephanegigandet: Right, I removed the initial initialization, so that we create the parents array only when we actually have entries to put in it

lib/ProductOpener/Tags.pm
stephanegigandet: added with other changes

stephanegigandet:
Thanks a lot for the thorough review @alexgarel , I incorporated all your suggestions I believe.

lib/ProductOpener/Tags.pm
alexgarel: Ok, that would be worth noting in a comment.

alexgarel (approved):
(Quickly) Tested ok on my side.

gspencergoog:
@stephanegigandet When I try and access the API with your example URL in the description, it wants a login and password. What would I use for that?

Also, is there a corresponding Dart API change already in the works, or should I try and flesh that out?

PR #5819, opened 2021-10-05
fix: accessibility issues with personal search preferences

### LINKED ISSUE #5818: Personal Search preferences accessibility issues
### Describe the bug

Some accessibility issues reported by the Accessibility Fundation on the personal search preferences on the website:

"On version 87 of Chrome I cannot reach the "Edit your food preferences" button with the keyboard and open it with the keyboard. This also applies to the button "close" & "Delete all preferences" if I want to close the preferences.

The radio buttons with the options "Not important, Important, etc." are not provided with a label (which falls under success criteria 1.3.1) so that the screen reader only reads the options and not for which the choice can be made. See example: () https://www.w3.org/WAI/tutorials/forms/grouping/) and (https://www.w3.org/WAI/WCAG21/Techniques/html/H71.html).

The logos of the companies are not provided with an alternative text in the alt = "" attribute. Not bad in itself because it is decorative, but it could help to improve the findability of your website in the search engines. If it doesn't interest you I would advise you to include the alt = "" attribute and leave it blank."

### To Reproduce

Go to https://world.openfoodfacts.org

Try navigating with the keyboard (tab + enter)  to select preferences.

### Expected behavior

Preferences can be set with the keyboard.

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Browser

### Browser version

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
- make it possible to select and activate the buttons to select preferences
- add a fieldset around the radio buttons for each attribute
- add empty alt="" for icons (all of them have a text label besides them)

fixes #5818

### DIFF
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -694,3 +694,8 @@ html[dir="rtl"] {
 .filter-grey {
   filter: invert(52%) sepia(0%) saturate(1186%) hue-rotate(218deg) brightness(96%) contrast(84%);
 }
+
+/* Accessuibility: Adding focus styles on the outer-box of the fake checkbox used for the Foundation switch to classify products according to one's preferences */
+.switch input[type="checkbox"]:focus + label::after {
+    outline: rgb(59, 153, 252) auto 5px;
+}
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -695,7 +695,7 @@ html[dir="rtl"] {
   filter: invert(52%) sepia(0%) saturate(1186%) hue-rotate(218deg) brightness(96%) contrast(84%);
 }
 
-/* Accessuibility: Adding focus styles on the outer-box of the fake checkbox used for the Foundation switch to classify products according to one's preferences */
+/* Accessibility: Adding focus styles on the outer-box of the fake checkbox used for the Foundation switch to classify products according to one's preferences */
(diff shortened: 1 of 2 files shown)


teolemon (approved):

stephanegigandet:
I added a change to CSS for the Foundation switches, otherwise when they have focus, there's no visual difference.

The drawback is that when we click on them, we still have that visible focus:

image

scss/_off.scss
```diff
@@ -694,3 +694,8 @@ html[dir="rtl"] {
 .filter-grey {
   filter: invert(52%) sepia(0%) saturate(1186%) hue-rotate(218deg) brightness(96%) contrast(84%);
 }
+
+/* Accessuibility: Adding focus styles on the outer-box of the fake checkbox used for the Foundation switch to classify products according to one's preferences */
```
teolemon, line 698:
```suggestion
/* Accessibility: Adding focus styles on the outer-box of the fake checkbox used for the Foundation switch to classify products according to one's preferences */
```

stephanegigandet:
> But I see the "close" and "delete all preferences" buttons are not accessible.

@alexgarel : with the changes in the PR, I can now select them and activate them with the keyboard though, why do you say they are not accessible?

PR #5840, opened 2021-10-06
Update Dockerfile

ocervell:
User `www-data` needs to map `off-net` user with uid `1000`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -129,7 +129,7 @@ RUN set -x \
     incron
 
 # Run www-data user as host user 'off'
-ARG WWW_DATA_HOST_USER=1001
+ARG WWW_DATA_HOST_USER=1000
 RUN usermod -u $WWW_DATA_HOST_USER www-data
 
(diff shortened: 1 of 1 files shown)


alexgarel (changes requested):
While I think having 1000 by default is ok.

I'm not sure tweaking this way is the good way.
Isn't it the other way around ? Shouldn't we force node and nginx user to host user ?

I would say:

1. add HOST_UID in .env, defaults to 1000
2. add a NODE_UID arg in node build and nginx build, change the UID of node / www-data user in those containers
3. use build args in docker-compose.yml for backend and frontend and set all needed users uid args to $HOST_UID

If you're too much in a hurry, I may still approve that PR, and consider the above a future change.

alexgarel:
@ocervell : I agree that default USER_UID should be 1000.
In prod you normally don't have problems with that, and normally no bind mount, or not many, and you may change user id on those to be on 1000.
The problem is more for DEV environment (hence the need to be able to tweak it) where you want dev user to access all without permissions problems.

PR #5841, opened 2021-10-06
feat: refactor Eco-Score knowledge panels + accordion display on web

stephanegigandet:
- Small change to the knowledge panel API : the title / subtitle / icon_url properties are now in a title_element (discussed with @jasmeet0817 )
- Improved a bit the Eco-Score panel (the panel that will be shown when we click on Eco-Score, to explain the computation). Still not complete though.
- Changed the template to display knowledge panels on the Web to use the accordion provided by Foundation: by default we only show the title element of the panel, and users can click to see more. In the app, the behaviour will probably be different (tapping on the title will open a new page with the details).

To be added soon:
- "expanded" property to indicate if a panel should be displayed already open, with all its content elements visible.
- "name" properties to be displayed above panel title (as in Tim's mocks)

To test:

Panels API result: https://uk.openfoodfacts.dev/api/v2/product/3017620422003/nutella-ferrero?fields=knowledge_panels
Panels shown on the web: https://uk.openfoodfacts.dev/product/3017620422003/nutella-ferrero?panels=1

image

### DIFF
--- a/templates/api/knowledge-panels/ecoscore/carbon_impact.tt.json
+++ b/templates/api/knowledge-panels/ecoscore/carbon_impact.tt.json
@@ -2,7 +2,7 @@
 [% SET co2_100g = product.ecoscore_data.agribalyse.co2_total / 10 %]
 [% SET driving_100g = co2_100g  * 100 / 19.3 %]
 {
-    "parent_panel_id": "ecoscore_carbon_impact",
+    "parent_panel_id": "root",
     "type" : "score",
     "level" :"info",
     "topics": [


jasmeet0817:
Question regarding `icon_color_from_evaluation`, should the BE just send the evaluation and let the FE decide whether they want to color or not, maybe the FE doesn't want to color the icon even if `icon_color_from_evaluation` is `true`, because the client may want to do something completely different according to the evaluation. Also maybe the client also wants to color the title string, not just the icon.

Then a discrepancy I saw in the sample data you shared:
Parent of a ecoscore_carbon_impact was ecoscore_carbon_impact  
image

stephanegigandet:
> Parent of a ecoscore_carbon_impact was ecoscore_carbon_impact

Fixed, thank you!

jasmeet0817:
btw I found that the type of panel is empty sometimes:

image

jasmeet0817:
Non Environment panels still have the old title/subtitle/icon
image

stephanegigandet:
> you could rename `grade` to `score` to make the distinction more clear.

In fact for both the Nutri-Score and the Eco-Score, there are both a score and a grade. The score is numeric (e.g. from -40 to 15 for the Nutri-Score, and from 0 to 100 for the Eco-Score). And the scores are then mapped to a 5 letter grade.

So I don't think we should rename grade to score.

stephanegigandet:
> btw I found that the type of panel is empty sometimes:

yes, to be honest, I'm not sure what we will use the type for, and which types we will eventually have. So just ignore "type" for now. I will put "info" or something for now instead of an empty type.

PR #5950, opened 2021-10-07
feat: Knowledge panels for labels

stephanegigandet:
- Added new panels for the labels used for Eco-Score (e.g. organic labels)
-> Panel data comes from the taxonomy. We may want to add a way to references translations from the po/common files instead of managing translations inside the taxonomies. That would ease the translations.

- Added a way to generate strings that contain a comma separated list (e.g. a list of labels), by genering the string with trailing commas on all elements, and removing the last trailing comma. (hacky but should work)

- Fixed sample Nutella brand panel to use the new title_element structure


stephanegigandet:
Current look:

image

PR #5955, opened 2021-10-08
feat: start of Environment card knowledge panel

stephanegigandet:
Some changes discussed with Jasmeet, in order to build the Environment card as designed by Tim.

The environment card is going to be itself a knowledge panel, of a special type "card", which will make it shown expanded, and we can round corners etc.

In it we have elements that are panel groups: groups that contain multiple panels. For instance the "Labels" panel  group may contain an organic panel + a fair trade panel. This is to be able to display "section names" like we have in the mockup.

Mockup:

image

Current web display:

image


stephanegigandet:
Replaced by https://github.com/openfoodfacts/openfoodfacts-server/pull/5958

PR #5958, opened 2021-10-11
Feat: add panel_group element and environment_card panel

stephanegigandet:
Replaces #5955 with a clean commit history.

--

Some changes discussed with Jasmeet, in order to build the Environment card as designed by Tim.

The environment card is going to be itself a knowledge panel, of a special type "card", which will make it shown expanded, and we can round corners etc.

In it we have elements that are panel groups: groups that contain multiple panels. For instance the "Labels" panel  group may contain an organic panel + a fair trade panel. This is to be able to display "section names" like we have in the mockup.

Mockup:

image

Current web display:

image


jasmeet0817:
Hello, possible to see a sample json response ? Can review the PR from API perspective, but probably better I leave the code review to Alex:)

jasmeet0817:
LGTM

PR #5971, opened 2021-10-12
fix: update log.conf to match production settings

ocervell:
We generated 121 GB of logs in under a week on pre-production ProductOpener. This matches production server log configuration.


alexgarel:
Can't we have configuration depends on an environment variable ?

We could have a log.conf and log-dev.conf and choose using an env variable. See how it's done for minions with minions_log.conf.

I don't know if you prefer that @stephanegigandet does the perl part for this.

PR #5972, opened 2021-10-12
Fix: identify lecitina de girasol additive and make emulsifiers Nova 4

stephanegigandet:
We correctly recognize "soy lecithin" and "lecithine de soja" as E322 lecithin additive because we have some code to remove leading words (in English) and trailing words (in French). This adds Spanish so that "lecitina de girasol" can be matched to "lecitina".

Also make emulsifiers trigger NOVA 4 even if they are not prefixed by the "emulsifier" additive class.


alexgarel:
@stephanegigandet Is that normal that you commited taxonomies/ingredients.result.sto ?

Issue #6009, opened 2021-10-20
Update the sample products data and images used for development in Docker


stephanegigandet:
### Is your feature request related to a problem? Please describe.

make import_sample_data in docker loads 117 products that were exported once in 2018:

```
off1:/srv/off/html/exports# ls -lrt 39*
-rw-r--r-- 1 off root    829380 Nov 11  2018 39-.tar.gz
-rw-r--r-- 1 off root 343424960 Nov 11  2018 39-.images.tar.gz
```

We should update the products more regularly.

### Describe the solution you'd like

We could:

1. Create a script to create the exports for a set of products (e.g. the top scanned products)
2. Run that script regularly to update the exports

stephanegigandet:
This has been fixed, we now have a nightly export of same products.

PR #6010, opened 2021-10-20
feat: add script to export products data and images for docker dev

stephanegigandet:
New script to export product data and images, for instance to load into a development install of Product Opener.

Related to #6009

Can be run like this:

`~/openfoodfacts-server$ docker-compose run --rm backend /opt/product-opener/scripts/export_products_data_and_images.pl --query ingredients_tags=en:salt --products-file /opt/product-opener/html/exports/test-products.tar.gz --images-file /opt/product-opener/html/exports/test-products-images.tar
`

### DIFF
--- a/scripts/export_products_data_and_images.pl
+++ b/scripts/export_products_data_and_images.pl
@@ -141,6 +141,7 @@ if (defined $sample_mod) {
 use Data::Dumper;
 print STDERR "MongoDB query:\n" . Dumper($query_ref) . "\n";
 
+# harvest products'code from mongo db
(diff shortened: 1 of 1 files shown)


stephanegigandet:
Added an option to get a random sample of products, based on their creation date:

./export_products_data_and_images.pl --sample-mod 10000,0 --products-file /srv/off/html/exports/products.random-modulo-10000.tar.gz --images-file /srv/off/html/exports/products.random-modulo-10000.images.tar

This is so that we can make a random sample of products instead of using the top scanned products (which typically have hundreds of images, so it's not very suitable for a dev environment).

PR #6014, opened 2021-10-21
fix: add countries correctly in scanbot

### LINKED ISSUE #4870: Scanbot is creating bad country entries
**Describe the bug**
Scanbot is creating bad entries such as `it:Belgien` or `it:Frankreich`.
See the following example: https://world.openfoodfacts.org/product/8000500111581/kinder-cioccolato?rev=43

It seems it has created dozens of bad entries like that, see: https://world.openfoodfacts.org/editor/scanbot/countries

stephanegigandet:
fixes #4870

The add_tags_to_field() from Tags.pm is the safe way to add some tags to a field like countries, as the "countries" field could be in any language, and not necessarily in the main language of the product (it is in the language last used to update the product).


alexgarel (approved):
LGTM.

PR #6015, opened 2021-10-21
Don't run this script, fix it first

CharlesNepote:
This script is producing bad data.

See: https://github.com/openfoodfacts/openfoodfacts-server/issues/4870


stephanegigandet:
Should have been fixed by https://github.com/openfoodfacts/openfoodfacts-server/pull/6014

Issue #6006, opened 2021-10-20
Finalize dockerization


alexgarel:
### Who for

Developpers / Administrators

### What

Change docker settings and some lines of codes to ease development on off-server and pro plateform.

### Why

Because I use this, and it can ease contributors. Because we use this for preproduction.

### Issues

#6000 
#6002 
#6003 
#6007

alexgarel:
All related issues are closed, so close this.

PR #6020, opened 2021-10-24
Ag docker improvments

### LINKED ISSUE #6000: docker-compose lacks volumes for a lot of data folders / files
### Describe the bug

Currently the docker-compose.yml 

defines some volumes  for data:
- users
- products
- new_images
- orgs
- logs
- tmp (tmpfs)

And some symlinks created by entrypoint for configuration data:
- ecoscore
- emb_codes
- forest-footprint
- ingredients
- lang
- packager-codes
- po
- taxonomies
- templates

But **we lack**:

Following folders:

- debug (WIP)
- export_files
- import_files
- deleted.images
- deleted_private_products
- facebook_users
- index
- po_from_lang

and maybe some of those (used only by scripts):
- lists  (/packager-codes.uk.en.html)
- madenearme
- rdf
- taxonomies-obf
- texts

and there are also those files stored directly in data_root:
- users_emails.sto
- ambassadors.txt
- Lang.${server_domain}.sto
- orgs_glns.sto

### To Reproduce

To find all folders and files, I did a very crude:

```
grep 'data_root' -r . --include="*.pm" --include="*.pl" --exclude="scripts/**" --color=always|sed -e 's/^.*data_root[^/]*\/\([A-Za-z0-9._-]*\).*$/\1/g'|sort|uniq
```
And then sort out the list using my editor and checking one by one.

### Expected behavior

Maybe we should mount data_root directly as a data volume (but we would need a deployement action to be sure to have symlinks created correctly).

### Screenshots

_No response_

### Additional context

https://openfoodfacts.slack.com/archives/C028ENWMH9N/p1634632376280700

### Type of device

Other

### Browser version

_No response_

_No response_

### Time per product

_No response_

### LINKED ISSUE #6003: Simplify docker-entrypoint

Currently docker-entrypoint does a lot of things that it may not be doing:

- symlinking data directories for configuration data (see also )
- symlinking site specific directory
- compiling language files

Language compilation particularly takes a lot of time (about 25s).

This should not be done in entrypoint but rather done in an init procedure, that is run once at deployment or when needed on the dev setup.

Have a make init which:
- ensure symlinks are there
- compile languages

Use this in deployment.

Eventually also look if it would be easy to avoid the site-specific link and instead rely on environment variable when supplied.

Language can't be compiled in Dockerfile for the file name depends on the model.

### Additional context

_No response_

_No response_

### Time per product

_No response_

alexgarel:
**Checklist:**

- [ ] Code is well documented
- [ ] If you have multiple commits please combine them into one commit by squashing them.
- [ ] Read and understood the [contribution guidelines](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/CONTRIBUTING.md)

**Description:**

Improvments to Docker setup. Mainly:

- fixes #6000
- fixes #6003
- avoid using root in containers (security best practice)
- align users uid with host user uid: easier for dev (binds mounts better supported) and even on production as we use bind mounts.

**Related issues and discussion:** #6006


alexgarel:
:'( no review from anyone @hangy @ocervell ?

Dockerfile
```diff
@@ -158,44 +178,46 @@ RUN rm /etc/apache2/sites-enabled/000-default.conf
...
+    mkdir -p var/run/apache2/ && \
+    chown www-data:www-data var/run/apache2/ && \
+    for path in users products product_images orgs new_images logs; do \
+        mkdir -p /mnt/podata/${path}; \
+        chown www-data:www-data /mnt/podata/${path}; \
+    done && \
+    # Create symlinks of data files in /mnt/podata (because we currently mix data and conf data)
+    for path in ecoscore emb_codes forest-footprint ingredients lang packager-codes po taxonomies templates; do \
+        ln -sfT /opt/product-opener/${path} /mnt/podata/${path}; \
+    done
```
ocervell, line 192: I'm convinced this still fits better in the init script. What would be best is to not need those in the first place, i.e get rid of the /mnt/podata symlinking, and not need to change the permissions. How we are going to achieve this, I'm not sure yet.

Makefile
```diff
@@ -87,6 +88,21 @@ tail:
...
+# Services #
+#----------#
+rebuild_language:
+	@echo "🥫 Rebuild language"
+	# Run build_lang.pl
+	${DOCKER_COMPOSE} run --rm backend perl -I/opt/product-opener/lib -I/opt/perl/local/lib/perl5 /opt/product-opener/scripts/build_lang.pl
+
+reset_owner:
+	@echo "🥫 reset owner"
+	${DOCKER_COMPOSE} run --rm --user root backend chown www-data:www-data -R /opt/product-opener/ /mnt/podata
```
ocervell, line 102: in production, or reasonably big setups with lots of data, this will take forever to run

Makefile
```diff
@@ -128,11 +144,14 @@ checks: front_lint
...
 create_external_volumes:
 	@echo "🥫 Creating external volumes (production only) …"
+	# zfs replications
 	docker volume create --driver=local -o type=none -o o=bind -o device=${MOUNT_POINT}/data html_data
 	docker volume create --driver=local -o type=none -o o=bind -o device=${MOUNT_POINT}/users users
 	docker volume create --driver=local -o type=none -o o=bind -o device=${MOUNT_POINT}/products products
 	docker volume create --driver=local -o type=none -o o=bind -o device=${MOUNT_POINT}/product_images product_images
 	docker volume create --driver=local -o type=none -o o=bind -o device=${MOUNT_POINT}/orgs orgs
+	# local data
+	docker volume create --driver=local -o type=none -o o=bind -o device=${DOCKER_LOCAL_DATA}/po_data po_data
```
ocervell, line 154: Should `po_data` really be an external volume with a mount point since the data comes from a ZFS clone / backup already ? I had kept these kind of volumes for data that is mounted to ProductOpener (host --> container); not for mounting from the container to the host (what's the value added in having this data back on the host ?)

Makefile
```diff
@@ -3,6 +3,7 @@
 NAME = "ProductOpener"
 ENV_FILE ?= .env
 MOUNT_POINT ?= /mnt
+DOCKER_LOCAL_DATA ?= /srv/off/docker_data
```
ocervell, line 6: ?

Makefile
```diff
@@ -87,6 +88,21 @@ tail:
...
 	tail -f logs/**/*
 
+
+#----------#
+# Services #
+#----------#
+rebuild_language:
+	@echo "🥫 Rebuild language"
+	# Run build_lang.pl
+	${DOCKER_COMPOSE} run --rm backend perl -I/opt/product-opener/lib -I/opt/perl/local/lib/perl5 /opt/product-opener/scripts/build_lang.pl
```
ocervell, line 98: Should we call this step just `build_lang` to match the script name ?

ocervell (changes requested):
Looks good but I'm still unconvinced by the `USER_UID` and `USER_GID` arguments since they are set at build time and not run time...

Normally everyone should be able to pull a production build and run it locally, this makes it impossible without changing the user uid / gid locally ... any ideas on how to make this possible ?

Makefile
alexgarel: This is a one shot command that you have to launch manually. 

Really meant to fix dev setups. (maybe I should add a comment ?)

Makefile
alexgarel: Talking to Christian, this is tha folder where we may put volumes data.
For what I understood, this volume is not a ZFS clone, it will hold data specific to preprod (this is mostly debug and temporary data, or local data).

alexgarel:
@ocervell 
> Looks good but I'm still unconvinced by the USER_UID and USER_GID arguments since they are set at build time and not run time...
>
> Normally everyone should be able to pull a production build (or a specific release) and run it locally, this makes it impossible without changing the user uid / gid locally ... any ideas on how to make this possible ?

It's not that using the remote image is impossible, it just work if you do not want to interact with bind mounted files, and didn't already moved permissions in some way (make reset_owner may help in this case).

But yes, I understand your concern about having different uid / gid in the image in dev, but I've never found a solution to this (even asking for people with more docker experience). Ideally docker should have provided id remapping on bind mount.

Makefile
alexgarel: see https://github.com/openfoodfacts/openfoodfacts-server/pull/6020#discussion_r742082012

conf/log.conf
```diff
@@ -1,5 +1,5 @@
-log4perl.rootLogger=ERROR, LOGFILE
-log4perl.logger.mongodb=INFO, MONGODB_LOGFILE
+log4perl.rootLogger= sub {return ($ENV{LOG_LEVEL_ROOT} // "ERROR") . ", LOGFILE";}
+log4perl.logger.mongodb= sub {return ($ENV{LOG_LEVEL_ROOT} // "ERROR") . ", LOGFILE";}
```
ocervell, line 2: default was 'INFO' and 'MONGODB_LOGFILE'

Dockerfile
```diff
@@ -158,44 +178,46 @@ RUN rm /etc/apache2/sites-enabled/000-default.conf
...
 ENTRYPOINT [ "/docker-entrypoint.sh" ]
+# default command is apache2ctl start
 CMD ["apache2ctl", "-D", "FOREGROUND"]
 
-FROM runnable AS runnable-vscode
+######################
+# Image for dev
+######################
+# note: user is not root but you car run --user root in docker if you want
+FROM runnable AS vscode
```
ocervell, line 206: do we need the vscode section at all ? Are you using it ? I'm using vscode but find no need for it.

Makefile
ocervell: ok, currently all mounts are done under `/mnt` in the QEMU VM. `/srv` does not exist there which will cause an error when deploying this to pre-prod.

alexgarel: Ok, so what is your proposal, I don't have a strong opinion on that. I was just thinking this data would be local to the machine, hence put it in srv.
Do you think it's better create an external volume and mount it ?

@stephanegigandet and @cquest your opinion may be usefull here. I may provide more detail if it's too much obscure to you.

conf/log.conf
```diff
@@ -1,5 +1,5 @@
-log4perl.rootLogger=ERROR, LOGFILE
-log4perl.logger.mongodb=INFO, MONGODB_LOGFILE
+log4perl.rootLogger= sub {return ($ENV{LOG_LEVEL_ROOT} // "ERROR") . ", LOGFILE";}
+log4perl.logger.mongodb= sub {return ($ENV{LOG_LEVEL_ROOT} // "INFO") . ", MOGODB_LOGFILE";}
```
stephanegigandet, line 2:
```suggestion
log4perl.logger.mongodb= sub {return ($ENV{LOG_LEVEL_ROOT} // "INFO") . ", MONGODB_LOGFILE";}
```

Dockerfile
```diff
@@ -158,44 +178,46 @@ RUN rm /etc/apache2/sites-enabled/000-default.conf
 COPY --from=builder /tmp/local/ /opt/perl/local/
 ENV PERL5LIB="/opt/product-opener/lib/:/opt/perl/local/lib/perl5/"
 ENV PATH="/opt/perl/local/bin:${PATH}"
+# Create writable dirs and change ownership to www-data
+RUN \
+    mkdir -p var/run/apache2/ && \
+    chown www-data:www-data var/run/apache2/ && \
+    for path in users products product_images orgs new_images logs; do \
```
stephanegigandet, line 185: I think we need to add "tmp" here, as PO tries to create /mnt/podata/tmp/templates

Otherwise I get this error:
```

# Run build_lang.pl
docker-compose --env-file=.env run --rm backend ls -lrt /mnt/
Starting po_memcached_1 ... done
Starting po_postgres_1  ... done
total 4
drwxr-xr-x 10 root root 4096 Nov  5 15:20 podata
docker-compose --env-file=.env run --rm backend perl -I/opt/product-opener/lib -I/opt/perl/local/lib/perl5 /opt/product-opener/scripts/build_lang.pl
Starting po_memcached_1 ... done
Starting po_postgres_1  ... done
Status: 500
Content-type: text/html

<h1>Software error:</h1>
<pre>mkdir /mnt/podata/tmp: Permission denied at /opt/perl/local/lib/perl5/x86_64-linux-gnu-thread-multi/Template/Provider.pm line 378.

```

alexgarel:
I've done some changes and tested, it seems ok.

I struggled with products image folder for there is no simple solution that is still compatible with prod. So I mounted it directly to right folder, but this mean in dev mode, you have them in your local folder (no way to have a volume mount inside a bind mount, and I can't have a symlink to products images because it would conflict with production).

Of course we could change code + frontend + prod, but it's a longer run. I think it's ok though, this way.

stephanegigandet (approved):

Makefile
alexgarel: confirmed with @stephanegigandet, as of now there is no important data there.

Dockerfile
```diff
@@ -158,44 +170,37 @@ RUN rm /etc/apache2/sites-enabled/000-default.conf
...
+# Create writable dirs and change ownership to www-data
+RUN \
+    mkdir -p var/run/apache2/ && \
+    chown www-data:www-data var/run/apache2/ && \
+    for path in data html_data users products product_images orgs new_images logs tmp; do \
+        mkdir -p /mnt/podata/${path}; \
+    done && \
+    chown www-data:www-data -R /mnt/podata && \
+    # Create symlinks of data files that are indeed conf data in /mnt/podata (because we currently mix data and conf data)
+    for path in ecoscore emb_codes forest-footprint ingredients lang packager-codes po taxonomies templates; do \
```
stephanegigandet, line 182:
```suggestion
    for path in ecoscore emb_codes forest-footprint ingredients lang packager-codes po t taxonomies templates; do \
```

We also need the /t directory because some tests need to read expected responses from the /t/expected_test_results / dir.

stephanegigandet: Otherwise this fails:

docker-compose run --rm  backend perl /opt/product-opener/t/packaging.t

alexgarel: why do we seek it under $data_root instead of current dir, or relative to test file ?

Makefile
```diff
@@ -87,12 +90,26 @@ tail:
...
+# Services #
+#----------#
+build_lang:
+	@echo "🥫 Rebuild language"
+	# Run build_lang.pl
+	${DOCKER_COMPOSE} run --rm backend perl -I/opt/product-opener/lib -I/opt/perl/local/lib/perl5 /opt/product-opener/scripts/build_lang.pl
+
+# create some bind mounted dirs, to avoid them from being owned by root
+make_base_dirs:
+	mkdir -p logs/apache2/ logs/nginx/
```
ocervell, line 104: Could you explain on why we need this, since I don't think we did before ?

docker-compose.yml
```diff
@@ -14,53 +67,13 @@ services:
...
-        target: /mnt/podata/mnt
-    networks:
-      - webnet
+  backend: *backend-conf
+  incron:
+    <<: *backend-conf
+    # This service run the incron jobs
+    # Only root can run incron
+    user: root
+    command: ["incrond"]
```
ocervell, line 76: :'( this means we run another container with the same configuration as product opener ? PO takes in average 6GB of memory so is this really necessary just to run the `incron` command ?

docker/dev.yml
```diff
@@ -1,20 +1,37 @@
...
-  backend:
-    image: openfoodfacts-server/backend:dev
-    build:
-      context: .
-      dockerfile: Dockerfile
-      target: runnable
-    volumes:
-      - ./:/opt/product-opener # mount local folder for reload on dev changes
+  backend: *backend-conf
+  incron: *backend-conf
```
ocervell, line 24: this doesn't match what's in the docker-compose.yml setup where we override the `command` field for `incron` service.

Dockerfile.frontend
```diff
@@ -1,21 +1,41 @@
...
 
-COPY html /opt/product-opener/html
-COPY icons /opt/product-opener/icons
-COPY scss /opt/product-opener/scss
-COPY gulpfile.js /opt/product-opener/
+COPY --chown=node:node html /opt/product-opener/html
+COPY --chown=node:node icons /opt/product-opener/icons
+COPY --chown=node:node scss /opt/product-opener/scss
+COPY --chown=node:node gulpfile.js /opt/product-opener/
+COPY --chown=node:node .snyk /opt/product-opener/
```
ocervell, line 29: Why do we need the `--chown=node:node` in those `COPY` commands ?

docker/docker-entrypoint.sh
```diff
@@ -1,16 +1,5 @@
...
-for path in " " users products product_images orgs new_images logs; do
-  mkdir -p /mnt/podata/${path}
-  chown www-data:www-data /mnt/podata/${path}
-done
-
-# Create symlinks of data files to /mnt/podata
-for path in ecoscore emb_codes forest-footprint ingredients lang packager-codes po taxonomies templates; do
-  ln -sfT /opt/product-opener/${path} /mnt/podata/${path}
-done
-
```
ocervell, line 13: In some cases I find it okay for the entrypoint script to contain `chown` statements and even `usermod -uid $USER_UID` to allow changing UID/GID dynamically and not have to worry about at runtime.

Looking at how much moving those few lines from `docker-entrypoint.sh` to Dockerfile complexified the setup workflow, I'm not so sure if that was the right call:

* Dockerfile is now way more complex than it used to (I'm confused by it not being an absolute expert in Docker but I haven't seen `COPY --chown` commands in other Dockerfiles, as well as 5/6 lines of inline script containing `mkdir`, `chown`, `usermod` commands)
* We added an `init_backend` command that needs to be run before `up`

lib/ProductOpener/Lang.pm
```diff
@@ -388,11 +388,11 @@ PO
 
 # Load stored %Lang from Lang.sto
 
-my $path = "$data_root/Lang.${server_domain}.sto";
+my $path = "$data_root/data/Lang.${server_domain}.sto";
```
ocervell, line 391: this might make production deployment fail (not sure how and how often build_lang is run there)

ocervell (changes requested):
I'm worried we made this too complex for the end user, reading and understanding the `Dockerfile`'s is harder for me now. I didn't review the `lib/` and `t/` changes as I don't know the impact of those changes on the Docker setup.

Makefile
ocervell: If it **needs** to be mounted, I would just mount it to the same ${MOUNT_POINT} as the rest of PO's data ?

Dockerfile
alexgarel: Done with  __FILE__

Makefile
alexgarel: I try to explain it in the comment above. If bind mounted directories do not exist, docker-compose create them on the fly, but owned by root:root and with 0771 permissions, thus www-data won't be able to write into them (neither would be the developer, but using sudo).

Before we were using root inside container, thus this was not a problem.

docker-compose.yml
alexgarel: The container does not take 6GB, it's the apache/perl processes that does.
Here you just create a single process using incron and running the script, which takes the same size as when it was exec.

You gain the "restart" capability of your service. (using exec to run a service is not a best practice)

Dockerfile.frontend
alexgarel: This is mostly due to snyk which will try to modify libraries on the fly (this is its role) so we need node user to have write capability.

docker/docker-entrypoint.sh
alexgarel: For me the problem with previous setup, was that starting any task with docker was taking some 15s, which I find a bit too much if I can displace it in the Image creation (I really use a lot run --rm rather than exec and this is a good practice for say running test, or running a command whithout needing to start the apache service and it's 6G load). More over it is also better for the image to really contain all target structure, for me this is like a documentation of full structure.

I think the complexity here is brought by the lack of separation between code, configuration and data in product opener. I hope we will solve that slowly by changing the code.

For me entrypoint can transform files etc., but normally for configuration purpose, not finishing the install.

lib/ProductOpener/Lang.pm
alexgarel: Previously this was run at each start of the container. Now it's run as image is built. It only depends on files in the git, so this is the right time to compute it.

ocervell: I was mentioning production because the deployment there is not automated, so the file is probably located in `$data_root/Lang.${server_domain}.sto` and I'm not sure when `build_lang.pl` is ran in production (maybe a cronjob ?)

Makefile
ocervell: Can't we do this in the Dockerfile itself, similarly to the other dirs we create there ? And then the bind mount on `./` would mount them since they exist within the container ?

docker/docker-entrypoint.sh
ocervell: +1 for the lack of separation between code / configuration / data --> the ideal state would be to have a stateless ProductOpener in Docker but we're far away from that :) as long as it's not too tedious for the end user and that maintainers understand what's in there that's fine for me.

Dockerfile
ocervell: Sounds good

ocervell (approved):

docker/dev.yml
alexgarel: It's ok as we do not redefine the `command:` data in backend-conf.

conf/log.conf
alexgarel: This was fixed.

Dockerfile
alexgarel: done.

lib/ProductOpener/Lang.pm
alexgarel: as said in conf call, @stephanegigandet will run build_lang.pl on prod as he deploy (this is a necessary step), which will generate file in the right place. It might then remove the old one.

Makefile
alexgarel: @ocervell just as I was going to ask again @cquest, I realized that this is not a good option to have a zclone for this in preprod, as in prod, this folder is /srv/off which does not only contains data but also a lot of stuff like the code and so on…

So I think we have to keep it local for now, as I planned initially.

PR #6037, opened 2021-10-29
fix: product images ownership

ocervell:
- Revert "fix: the backend needs write access to product images (#6011)"
- fix: set current ${UID} by default in build arguments, so that the current user running docker-compose maps to the `WWW_DATA_HOST_USER` uid and can write to folders owned by the klatter


docker-compose.yml
```diff
@@ -45,9 +45,8 @@ services:
       # Users, products, product images and orgs files
       - users:/mnt/podata/users # .sto
       - products:/mnt/podata/products # .sto
+      - product_images:/mnt/podata/product_images # .jpg
```
alexgarel, line 48: @ocervell I don't understand why you had to revert this ?
(the concern was the permissions, isn't it ?)

PR #6050, opened 2021-11-04
Test empty fields + cors issues

CharlesNepote:


alexgarel (approved):
I approve but would prefer not to have an alert (I let you do as you think @CharlesNepote)

stephanegigandet (approved):

PR #6051, opened 2021-11-04
Update languages.result.txt

laralem:
**Description:**

- I've removed "Língua" from strings like "Língua portuguesa" in Portuguese translations to fix a issue while editing products in Open Food Facts website. Because at this moment, pressing the key "P" in keyboard, it jumps to another language and not "Portuguese" or the first one having the "P" letter.  This affects also the sort order of course. (See  #6046)

- Fixed a few Portuguese translation errors.

- In some cases (4 or 6 of them) I've changed the first existing translation in each string to show simultaneously Brazilian Portuguese and Portuguese (or European Portuguese) in the string shown in interface (where they differ) to reach all Portuguese speaking variations. For example "Thecho / Checo" (pt-br / pt). So this way no user will think "oh this website is in Brazilian Portuguese" or "it's in European Portuguese", it's in both like in Portuguese Wikipedia.

- Added a note in the beginning to help other translators.

#6046 

P.S.: I have to make this changes to https://github.com/openfoodfacts/openfoodfacts-server/blob/main/taxonomies/languages.txt also right? Can someone please explain why there are 2 files apparently about the same thing?  "languages.txt" and  "languages.result.txt"

### DIFF
--- a/taxonomies/languages.result.txt
+++ b/taxonomies/languages.result.txt
@@ -1,3 +1,10 @@
+# NOTE: 1. Just translate the name of the language, for example, in "Abkhaz language" translate only "Abkhaz"
+#          because if the resulting translation is "Língua abcázia" (example in Portuguese) when you are editing a product
+#          in Open Food Facts website and you press the key "A" in the field using this translation, it will jump right
+#          to the language translation beginning with "A" and not "Língua abcázia" because this one begins with "L".
+#       2. Only the first string until "," is shown. So if there is need to show the name of the language in other language variation
+#          use "/", for example "Color / colour", "Pastó / afegão" or "Tcheco / checo". Try to keep short and only use very common names, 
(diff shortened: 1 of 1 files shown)


stephanegigandet:
Hello @laralem , in fact the changes need to be made only in languages.txt (languages.result.txt is computed automatically from languages.txt, it will be overwritten when we rebuild the taxonomy)

laralem:
> Hello @laralem , in fact the changes need to be made only in languages.txt (languages.result.txt is computed automatically from languages.txt, it will be overwritten when we rebuild the taxonomy)

Thanks. I should have known that. So this pull request should be closed/deleted or something right?

aleene (approved):

aleene:
Guess we just close this one and move the changes over to languages.txt.

laralem:
> Guess we just close this one and move the changes over to languages.txt.

Done in #6064

PR #6054, opened 2021-11-05
Small fixes for knowledge panels

stephanegigandet:
- Fixed some icons paths (the Eco-Score ones existed in prod but not in github)
- Using the result of a FILTER to set a variable in Template::Toolkit works in the dev server, but not in production and in the docker environment. Now exposing the sprintf function directly.
- Fixed the display of packaging components by recycling types
- Fixed the display of environmental labels panels when we don't have an Eco-Score (no Agribalyse category) but when we have organic labels.
- Now showing the panels in the product page for all moderators, in order to uncover issues before we set it for all users.

### DIFF
--- a/templates/api/knowledge-panels/environment/carbon_footprint.tt.json
+++ b/templates/api/knowledge-panels/environment/carbon_footprint.tt.json
@@ -18,7 +18,7 @@
     "title_element": {
         [% SET driving_100g_rounded = sprintf('%.1f', driving_100g) %]
         "title": "[% f_lang('f_equal_to_driving_km_in_a_petrol_car',  { 'kilometers' => driving_100g_rounded } ) %]",
-        [% SET co2_100g_rounded = co2_100g * 1000 FILTER format('%.0f') %]
+        [% SET co2_100g_rounded = sprintf('%.0f', co2_100g * 1000) %]
         "subtitle": "[% f_lang('f_carbon_footprint_per_100g_of_product', { 'grams' => co2_100g_rounded }) %]",
         "icon_url": "[% static_subdomain %]/images/icons/dist/car.svg",
         "icon_color_from_evaluation": true,
--- a/templates/api/knowledge-panels/environment/carbon_footprint.tt.json
+++ b/templates/api/knowledge-panels/environment/carbon_footprint.tt.json
@@ -16,9 +16,9 @@
     "evaluation": "good",
     [% END %]
     "title_element": {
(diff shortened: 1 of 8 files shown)


templates/api/knowledge-panels/environment/carbon_footprint.tt.json
```diff
@@ -16,9 +16,9 @@
     "evaluation": "good",
     [% END %]
     "title_element": {
-        [% SET driving_100g_rounded = driving_100g FILTER format('%.1f'); %]
+        [% SET driving_100g_rounded = sprintf('%.1f', driving_100g) %]
         "title": "[% f_lang('f_equal_to_driving_km_in_a_petrol_car',  { 'kilometers' => driving_100g_rounded } ) %]",
-        [% SET co2_100g_rounded = co2_100g * 1000 FILTER format('%.0f'); %]
+        [% SET co2_100g_rounded = co2_100g * 1000 FILTER format('%.0f') %]
```
alexgarel, line 21: Why didn't you use sprintf here also ?

templates/web/pages/product/product_page.tt.html
```diff
@@ -184,7 +184,7 @@ <h4>[% lang("nova_groups_s") %]
                 <a href="/nova" title="[% a_title %]">
                 [% display_icon('info') %]</a>
             </h4>
-            <a href="/nova" title="[% a_title %]"><img src="/images/misc/nova-group-[% group %].svg" alt="[% display %]" style="margin-bottom:1rem;max-width:100%"></a><br>
+            <a href="/nova" title="[% a_title %]"><img src="/images/attributes/nova-group-[% group %].svg" alt="[% display %]" style="margin-bottom:1rem;max-width:100%"></a><br>
```
alexgarel, line 187: we are not very consistent… in donate_banner.tt we use static.openfoodfacts.org, there no. (not so important, but just to notice)

alexgarel (approved):
LGTM appart from two small comments.

templates/api/knowledge-panels/environment/carbon_footprint.tt.json
stephanegigandet: Good catch, fixed.

templates/web/pages/product/product_page.tt.html
stephanegigandet: Right, we should fix that in another PR.

PR #6055, opened 2021-11-05
Try to fix build_lang.pl in Docker

stephanegigandet:
Some changes to try to make build_lang.pl to work. This script currently writes in $data_root + $www_root/data/ which makes thing fail.

I'm creating a new $data_root/data directory to hold the resulting translations .sto files (instead of putting them in $data_root). Ideally I'd like to gradually move all the data files that the backend writes in this /data directory (e.g. users, products, taxonomies etc.), with the exception of files that the frontend is expected to serve (we could have a corresponding html/data/ directory for those)

Also added a /mnt/podata/tmp file (needed for Template::Toolkit compiled templates)


docker-compose.yml
```diff
@@ -49,6 +48,8 @@ services:
       - products:/mnt/podata/products # .sto
       - product_images:/mnt/podata/product_images # .jpg
       - orgs:/mnt/podata/orgs # .sto
+      - html_data:/mnt/podata/html_data # Static data (e.g dumps)
```
alexgarel, line 51: Do we need this to be a separate volume ? Is this a zfs clone also ?

conf/log.conf
```diff
@@ -1,5 +1,5 @@
 log4perl.rootLogger= sub {return ($ENV{LOG_LEVEL_ROOT} // "ERROR") . ", LOGFILE";}
-log4perl.logger.mongodb= sub {return ($ENV{LOG_LEVEL_ROOT} // "INFO") . ", MOGODB_LOGFILE";}
+log4perl.logger.mongodb= sub {return ($ENV{LOG_LEVEL_ROOT} // "INFO") . ", MONGODB_LOGFILE";}
```
alexgarel, line 2: Nice catch !

docker-compose.yml
alexgarel: Also if we add it here, we must add it to deployement file.

alexgarel (changes requested):
I just have the question on the separate volume for html_data.

docker-compose.yml
stephanegigandet: I don't think it's a ZFS clone, we need it to be writable by the backend.

alexgarel: ok but all /mnt/podata should be writable. So we don't need a specific mount point here.

stephanegigandet: ok, I removed it, it works.

alexgarel:
I took what was usefull on my PR #6020

PR #6059, opened 2021-11-07
Dutch specific additions

aleene:
**Description:**
I tried to add (some) dutch recycling instructions. But I am not sure I add them to the right files. The instructions often have the structure `<packaging> in/bij <recycling bin>`. Where can I add the structure? And what to do about adding logo's for Robotoff?

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -1057,7 +1057,7 @@ sub parse_ingredients_text($) {
 							# origin? (origine : France)
 
 							# try to remove the origin and store it as property
-							if ($between =~ /\s*(de origine|d'origine|origine|origin|origins|alkuperä|ursprung)\s?:?\s?\b(.*)$/i) {
+							if ($between =~ /\s*(de origine|d'origine|origine|origin|origins|alkuperä|ursprung|oorsprong)\s?:?\s?\b(.*)$/i) {
 								$between = '';
 								my $origin_string = $2;
 								# d'origine végétale -> not a geographic origin, add en:vegan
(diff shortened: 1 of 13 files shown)


aleene:
Need to add these https://kidv.nl/media/weggooiwijzer/2021_kidv_weggooiwijzer_handleiding_nl_juni_2021.pdf?1.2.2

stephanegigandet:
> The instructions often have the structure `<packaging> in/bij <recycling bin>`.

We should add some of those in the tests t/packaging.txt, to make sure that we parse them correctly.

aleene:
I'll create some tests. What about the recycling instruction part? Where can I add/test that?

stephanegigandet:
> I'll create some tests. What about the recycling instruction part? Where can I add/test that?

It's in packaging.txt as well.

We may need to do some code changes, as the code is expecting something like "[shape] in [material] to [discard/recycle]", whereas in Dutch we have something like "[shape] to recycle in [material] bin". e.g. "XYZ bij oud papier." does not necessarily mean that XYZ is in paper, it could be cardboard I guess.

aleene:
> We may need to do some code changes, as the code is expecting something like "[shape] in [material] to [discard/recycle]", whereas in Dutch we have something like "[shape] to recycle in [material] bin". e.g. "XYZ bij oud papier." does not necessarily mean that XYZ is in paper, it could be cardboard I guess.
Correct in NL you throw the cardboard items into the paper bin. So we do not know what XYZ is made of.

aleene:
I commented out the tests. The taxonomies need to be committed before I can add those.

stephanegigandet (approved):
Thank you!

PR #6061, opened 2021-11-09
New vegetarian icons

stephanegigandet:
We currently use a leaf for the vegan icon, and an egg for the vegetarian icon. But we also use the egg icon to indicate the presence of eggs as an allergen, so it can be confusing.

This PR is to change the vegetarian icon:

image


alexgarel (approved):
LGTM

PR #6062, opened 2021-11-09
Feat: Added irradiated / non-irradiated ingredient processing

stephanegigandet:
This is to parse things like "irradiated spices" in ingredients lists.


taxonomies/ingredients_processing.result.txt
```diff
@@ -1097,14 +1103,6 @@ nl:zongedroogd, zongedroogde
 pt:seco ao sol, seca ao sol, secos ao sol, secas ao sol
 sv:soltorkad, soltorkade
 
-en:superior quality
```
teolemon, line 1100: why are we removing this @stephanegigandet ?

teolemon (approved):

taxonomies/ingredients_processing.result.txt
stephanegigandet: It was removed in this PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/5983
It's not a processing, we will probably add an ingredients_quality taxonomy at some point.

aleene: I guess I removed it. Can not remember what the reason was, probably an issue with the ingredients. I often use the processing as meaning `adjectives`.

aleene (approved):

PR #6064, opened 2021-11-09
Change pt and pt-br in languages.txt +notes

laralem:
**Description:**

- I've removed "Língua" (language) from strings like "Língua portuguesa" in Portuguese translations to fix a issue while editing products in Open Food Facts website. Because at this moment, pressing the key "P" in keyboard, it jumps to another language and not "Portuguese" or the first one having the "P" letter.  This also affects sorting. (See  #6046)

- Fixed a few Portuguese translation errors.

- In some cases (4 or 6 of them) I've changed the first existing translation in each string to show simultaneously Brazilian Portuguese and Portuguese (or European Portuguese) in the string shown in interface (where they differ) to reach all Portuguese speaking variations, since the interface is currently not available in Brazilian Portuguese. For example "Thecho / checo" (pt-br / pt). So this way no user will think "oh this website is in Brazilian Portuguese" or "it's in European Portuguese", it's in both like in Portuguese Wikipedia.

- Added a note in the beginning to help other translators.

#6046

### DIFF
--- a/taxonomies/languages.txt
+++ b/taxonomies/languages.txt
@@ -1,3 +1,11 @@
+# NOTE: 1. Just translate the name of the language, for example, in "Abkhaz language" translate only "Abkhaz"
+#          because if the resulting translation is "Língua abcázia" (example in Portuguese) when you are editing a product
+#          in Open Food Facts website and you press the key "A" in the field using this translation, it will jump right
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Thank you!

PR #6071, opened 2021-11-12
better wording for ingredients that threaten species

stephanegigandet:
as suggested in #translations on Slack


alexgarel (approved):
:+1:

PR #6074, opened 2021-11-15
Dutch taxonomy finetuning

aleene:
**Description:**

I am going through new product additions in the NL, in order to finetune the taxonomies.


aleene:
@stephanegigandet  I can not get the tests working. Can you have a look?

stephanegigandet:
> @stephanegigandet I can not get the tests working. Can you have a look?

Sure.

stephanegigandet:
@aleene : I made some changes to make the test pass.

There are a number of taxonomy files that have been started but that are in fact unused. I added a warning on top of them.

I added a "recycle in glass bin" entry in the packaging_recycling.txt taxonomy, but I'll need to add some code so that it also adds the glass material.

aleene:
> I added a "recycle in glass bin" entry in the packaging_recycling.txt taxonomy, but I'll need to add some code so that it also adds the glass material.
So I have been editing the wrong file?

Remark that in NL the paper container also contains the cardboard boxes. So your change to paper is incorrect.

aleene:
Should I add the other containers there as well?

aleene:
Can't we create a taxonomy that defines the rules that define what material goes into which container per country?

stephanegigandet:
> Can't we create a taxonomy that defines the rules that define what material goes into which container per country?

Well in some cases (even in France), it's by city.

stephanegigandet:
> Should I add the other containers there as well?

Let me try to first get the current test examples you added correct (the cardboard/paper thing), and then afterwards we can add more containers / more languages, but we will have to add corresponding examples in the tests as well, otherwise there's a very high chance that some of them may not work, or will be broken by future changes.

aleene:
> Well in some cases (even in France), it's by city.

At some stage a location set by the user can be used and we add the corresponding recycling rules. Also nice for calculating a better ecoscore transportation correction. But start with support for the different countries with support for ISO 3166-2 (https://en.wikipedia.org/wiki/ISO_3166-2).

stephanegigandet:
@aleene ok I'm done with the packaging related changes, things look good now I think.

aleene:
@stephanegigandet I merged main and now the ecoscore test fails. Issue with your changes?

stephanegigandet:
> Issue with your changes?

I just fixed it. The issue was for "1 PET plastic bottle":  1 is the code for PET, so we recognized "1 PET" as the material, whereas in this specific case, it should be the number of bottles.

taxonomies/packaging_shapes.result.txt
```diff
@@ -61,16 +63,30 @@ nl:Fles
...
 
-en:Bottle cap, bottle top, cap, top, Wine cork, wine bottle cork, cork, bottle caps, bottle tops, caps, tops, wine corks, wine bottle corks, corks
-fr:Bouchon de bouteille, bouchon, bouchon de bouteille de vin, bouchons de bouteille, bouchons de bouteilles, bouchons
+en:Bottle cap, bottle top, cap, top, bottle caps, bottle tops, caps, tops
+fr:Bouchon de bouteille, bouchon, bouchons de bouteilles, bouchons
 hu:Kupak, borosdugó, dugó
+nl:Dop
 pt:Cápsula de garrafa, cápsulas de garrafa, tampa de garrafa, tampas de garrafa, rolha, rolhas, rolha de garrafa, rolhas de garrafa, rolha de cortiça, rolhas de cortiça, rolha de plástico, rolhas de plástico, carica, caricas, cápsula de rosca, cápsulas de rosca
 
+< en:Bottle cap
```
stephanegigandet, line 72: So for this case, I had not put "capsules" under bottle caps, because the term is used for the Nespresso coffee capsules.

taxonomies/packaging_shapes.result.txt
```diff
@@ -388,3 +424,9 @@ fr:Muselet, Muselet et plaque
 en:wrapper, wrapping
 fr:enveloppage
 
+nl:Folder
```
stephanegigandet, line 427: Would you have English translations for those?

aleene:
@stephanegigandet I added the test for the vegan label detection. The build fails now, but I do not see which test is failing.

taxonomies/packaging_shapes.result.txt
aleene: added, would be brochures or leaflets

stephanegigandet:
@aleene I fixed the test to put what is currently output by the ingredients analysis.

Let's merge this PR as-is, and then I can make the code change in another PR so that "something (not from animals)" is analyzed as vegan.

stephanegigandet (approved):
Looks good, thank you!

PR #6102, opened 2021-11-23
Dutch additions to taxonomies

aleene:
**Description:**
Updating the taxonomies based on newly added dutch products.


t/ingredients.t
```diff
@@ -354,15 +354,6 @@ my @tests = (
 		}
 	],
 
-	# from vegetal origin
```
stephanegigandet, line 357: Removed by error?

stephanegigandet (commented):
There's a test that has been removed:

	[
		"en-vegetal-ingredients",
		{
			lc => "en",
			ingredients_text => "Gelatin (vegetal), Charcoal (not from animals), ferments (from plants), non-animal rennet, flavours (derived from plants)",
		}
	],	

Can you add it back?

t/ingredients.t
aleene: I can not see when it was deleted in the history. I saw that you made the change. Maybe I forgot to pull the newest version and did not see any warning.

stephanegigandet (approved):
Thanks!

PR #6103, opened 2021-11-23
Fix Eco-Score when Agribalyse score is 0, fixes #6101

### LINKED ISSUE #6101: No Eco-Score for lamb leg
### Describe the bug

Lamb leg (and possibly other meat categories) do not have an Eco-Score.

### To Reproduce

e.g. https://fr.openfoodfacts.org/produit/2258940012005/gigot-d-agneau (reported by @OpenChris )

### Expected behavior

Eco-Score should be assigned.

The reason for the bug is a broken check for the presence of an Agribalyse score, which fails if the score is 0.

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Other

### Browser version

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
Eco-Score was not computed for categories that have a 0 Agribalyse score.


alexgarel (approved):
LGTM

Issue #6070, opened 2021-11-10
Various weird behaviours on the Eco-Score knowledge panels


teolemon:
Missing malus on the KP for Origin of Ingredients
![image](https://user-images.githubusercontent.com/1689815/141152778-e07a7f75-a550-4e39-b32e-967ac942a071.png)
https://world.openfoodfacts.org/product/5400141352962/olives-noires-denoyautees-boni

teolemon:
![image](https://user-images.githubusercontent.com/1689815/141154990-0847d82e-89c9-4b64-a0ff-2c9c50862606.png)
https://uk.openfoodfacts.dev/product/13297703/les-laitiers-responsables

teolemon:
![image](https://user-images.githubusercontent.com/1689815/141155214-c52acec7-e5f4-4f83-b263-d0e4fc196f00.png)
Adding a link to edit mode ?

teolemon:
![image](https://user-images.githubusercontent.com/1689815/143188998-4353329a-c381-4d1e-b86d-398a027e0b45.png)
Panel not displayed when the category is not eligible

stephanegigandet:
> Panel not displayed when the category is not eligible

That's the expected behaviour, we don't display the Nutri-Score.

stephanegigandet:
> Adding a link to edit mode ?

We'll need to think about the overall design (e.g. on mostly empty products, we might not want to display 12 buttons to complete every field) and how this work on web and mobile. Right now we have the same panels for both.

stephanegigandet:
The missing values have been fixed.

For new issues, I think it's best to have one bug per issue, as otherwise it's difficult to track what has been fixed or not.

PR #6105, opened 2021-11-23
fix: Localize Eco-Score soon enough + add 'world' Eco-Score.

stephanegigandet:
- We need to localize the Eco-Score according to the country of the request before we generate the knowledge panels
- Add a "world" Eco-Score that we can display on https://world.openfoodfacts.org (this Eco-Score has no transportation bonus). (fixes 1 of the issues listed in #6070 )


alexgarel (approved):
LGTM

PR #6107, opened 2021-11-24
fix: use /tmp for compiled templates

stephanegigandet:
Fixes issue when running dev environment in docker:

after a clean repo checkout, make clean, make dev

```
Creating po_backend_run ... done
Status: 500
Content-type: text/html
<h1>Software error:</h1>
<pre>mkdir /mnt/podata/tmp/templates: Permission denied at /opt/perl/local/lib/perl5/x86_64-linux-gnu-thread-multi/Template/Provider.pm line 378.
```

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -274,7 +274,7 @@ $tt = Template->new(
 		EVAL_PERL    => 1,
 		STAT_TTL     => 60,                              # cache templates in memory for 1 min before checking if the source changed
 		COMPILE_EXT  => '.ttc',                          # compile templates to Perl code for much faster reload
-		COMPILE_DIR  => $data_root . '/tmp/templates',
+		COMPILE_DIR  => '/tmp/templates',
 		ENCODING     => 'UTF-8',
 		RECURSION    => 1,	# Needed for the knowledge panels that contain subpanels
 	}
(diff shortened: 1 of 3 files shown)


alexgarel:
I don't think it's a good solution, for e have more than one instance on same machine in prod.

lib/ProductOpener/Display.pm
```diff
@@ -274,7 +274,7 @@ $tt = Template->new(
 		EVAL_PERL    => 1,
 		STAT_TTL     => 60,                              # cache templates in memory for 1 min before checking if the source changed
 		COMPILE_EXT  => '.ttc',                          # compile templates to Perl code for much faster reload
-		COMPILE_DIR  => $data_root . '/tmp/templates',
+		COMPILE_DIR  => '/tmp/templates',
```
VaiTon, line 277: I suggest using a subdir for the server software, like `/tmp/productopener/templates/`.

stephanegigandet:
Note: Template::Toolkit prefixes what it puts in the tmp directory with the full path of templates:

off1:/srv/off/tmp/templates/srv/off/templates/web# ls
common	pages  panels

stephanegigandet:
> I don't think it's a good solution, for e have more than one instance on same machine in prod.

Then we need to have a writable /mnt/podata/tmp dir created by docker.

alexgarel (approved):
@stephanegigandet can you look at it  also as there are now more commits by myself :-D

stephanegigandet (commented):
Looks good to me

PR #6115, opened 2021-11-25
Add missing images/misc/ images

stephanegigandet:
This PR contains images that were put on the production server, but not on GitHub.

PR #6118, opened 2021-11-25
Add missing images in html/images/misc

stephanegigandet:
Replaces #6115


alexgarel (approved):
ok, I think in the future that would be really cool to distinguish images really referenced by the app vs the one for content in openfoodfacts-web, and put them there.

alexgarel:
This is only 15M now.

PR #6119, opened 2021-11-26
Dutch recycling

aleene:
**Description:**
Dutch recycling bins and instructions and tests

### DIFF
--- a/taxonomies/packaging_recycling.txt
+++ b/taxonomies/packaging_recycling.txt
@@ -46,6 +46,7 @@ nl:Plastic - Metaal en Drankencartons, PMD, wel PMD
 <en:recycle
 en:plastics
 nl:plastic afval, bij plastic afval
+packaging_materials:en: en:plastics
 
 #packaging that is to be returned to the shop, in exchange for money, might be PET-bottles, meetal cans or glass bottles
 <en:recycle
--- a/taxonomies/packaging_recycling.txt
+++ b/taxonomies/packaging_recycling.txt
@@ -4,6 +4,9 @@ synonyms:fr:conteneur,container,poubelle,bac,benne
 
 stopwords:fr:le,la
 
+#comment:en:Instructions for the Netherlands: https://kidv.nl/media/weggooiwijzer/2021_kidv_weggooiwijzer_handleiding_nl_juni_2021.pdf?1.2.2
+#comment:en:instructions for Belgium: https://www.betersorteren.be/nl/download-de-sorteergidsen-voor-pmd-papier-karton-en-glas
+
 en:reuse, re-use, to reuse, to re-use,  to be reused, to be re-used, reusable, re-usable, Refillable, refill, to refill, multiple use
 de:wiederverwenden, wiederverwendbar, nachfüllbar, mehrfach verwendbar, zum Wiederverwenden
 fr:réutiliser, à réutiliser, réutilisable, consigné, usage multiple, réemployer, à réemployer, réemployable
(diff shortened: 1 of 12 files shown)


aleene:
@stephanegigandet The shape interference is not yet supported?

taxonomies/packaging_recycling.txt
```diff
@@ -30,8 +33,41 @@ nl:recyclen bij oud papier, bij oud papier, in papierbak, in de papierbak, papie
 # If the recycling instruction is to recycle in the paper bin, we can't assume the material is paper
 # it could also be cardboard
 
+# in the Netherlands for recycling of drink cartons
+<en:recycle
+nl:drankencartons
```
stephanegigandet, line 38: For the packaging_recycling taxonomy, we should list the action as the main name:
recycle, discard, reuse etc.

So here we could have a main name as:

en:recycle-with-drink-cartons

Also the main name must be English, so that we don't need to update tests later.

taxonomies/packaging_recycling.txt
```diff
@@ -30,8 +33,41 @@ nl:recyclen bij oud papier, bij oud papier, in papierbak, in de papierbak, papie
...
+packaging_shapes:en: en:brick
+
+# in Belgium can be used for recycling of metal, plastics and drink cartons
+<en:recycle
+en:plastics - metal and bricks
+nl:Plastic - Metaal en Drankencartons, PMD, wel PMD
+
+<en:recycle
+en:plastics
```
stephanegigandet, line 47:
```suggestion
en:recycle-with-plastics, plastics
```

taxonomies/packaging_recycling.txt
```diff
@@ -30,8 +33,41 @@ nl:recyclen bij oud papier, bij oud papier, in papierbak, in de papierbak, papie
...
+nl:plastic afval, bij plastic afval
+packaging_materials:en: en:plastics
+
+#packaging that is to be returned to the shop, in exchange for money, might be PET-bottles, meetal cans or glass bottles
+<en:recycle
+en:deposit
```
stephanegigandet, line 53:
```suggestion
en:return to store, return to shop, deposit
```

taxonomies/packaging_recycling.txt
```diff
@@ -30,8 +33,41 @@ nl:recyclen bij oud papier, bij oud papier, in papierbak, in de papierbak, papie
...
+#In the Netherlands PET bottles must be returned in exchange for the deposit, when it has the statiegeldfles logo
+<en:deposit
+en:PET deposit bottle
+nl:statiegeldfles
+packaging_materials:en: en:pet-polyethylene-terephthalate
+#packaging_shapes:en: en:bottle seems not to work
+
+#This is green waste
+<en:recycle
+en:Green waste
```
stephanegigandet, line 65:
```suggestion
en:Compost, Green waste
```

taxonomies/packaging_recycling.txt
aleene: OK I changed the names.

In NL metal is thrown away with regular waste, but recovered/recycled during incineration. How shall I add this?

stephanegigandet: This taxonomy is more about the recycling instructions than the actual recycling, so it should just be "to discard" I think.

taxonomies/packaging_recycling.txt
aleene: Added

taxonomies/packaging_recycling.txt
aleene: OK

stephanegigandet (approved):
Looks good to me. We will have to change some code for the packaging knowledge panels as currently it expects only discard or recycle.

PR #6120, opened 2021-11-26
Use new nutrients taxonomy to replace hardcoded values from Food.pm - WIP

stephanegigandet:
This is a work in progress, as those changes touch many parts of the code.

In a nutshell, the goal is to remove the hardcoded translations and properties for nutrients that are currently in Food.pm, and use a nutrients taxonomy instead.

- Moved @aleene's nutrients taxonomy to nutrients_old.txt,
- Created a script to generate a new taxonomy source file nutrients.txt, from the hardcoded values in Food.pm (that changed since the old nutrients.txt taxonomy was created) + the old nutrients taxonomy in order to get all the synonyms from wikidata.
- Started to replace the uses of %Nutriments (the hardcoded values in Food.pm) with the new nutrients taxonomy
- Had to move some code around between Food.pm, Tags.pm and the new PackagerCodes.pm module
- Simplified a bit the template for the product edit form

Note: the canonical entry for nutrients in the new taxonomy use the zz: language prefix. This is because we can't change the ids of nutrients from the ones we have today in the nutriments hash of the products, even though they do not correspond exactly to their English names. e.g. some are shortened ("fruits-vegetable-nuts" instead of "fruits, vegetables, nuts, and rapeseed, walnut and olive oil").

NOT DONE YET:
- Replacing all uses of %Nutriments by the taxonomy
- Removing %Nutriments from Food.pm
- Testing very thoroughly that all the changes did not break something


stephanegigandet:
Using a taxonomy for nutrients will ease translations, but it will also be easier to add properties. We need some new properties for the nutrition table knowledge panels. I could have added them in the hardcoded values, but that would have been one more thing to migrate later. So let's bite the bullet now, do all the needed refactoring, and then after we can add properties and features.

aleene (approved):

stephanegigandet:
replaced by https://github.com/openfoodfacts/openfoodfacts-server/pull/6123

PR #6129, opened 2021-11-29
fix: put back compiled templates dir in data_root/tmp

stephanegigandet:
Somehow I can't get compiled templates to be created in /tmp on the small dev openfoodfacts server, so putting back the compiled templates in $data_root/tmp/templates


alexgarel (approved):
LGTM

PR #6130, opened 2021-11-29
Ag docker fixes

alexgarel:
Small fix to avoid problems with init_backend, which needs to first have build (and needs to be before `up`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ goodbye:
 #-------#
 # Local #
 #-------#
-dev: hello init_backend up import_sample_data refresh_product_tags
+dev: hello build init_backend _up import_sample_data refresh_product_tags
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -33,7 +33,7 @@ goodbye:
 #-------#
 # Local #
 #-------#
-dev: hello init_backend up import_sample_data refresh_product_tags
+dev: hello build init_backend _up import_sample_data refresh_product_tags
```
stephanegigandet, line 36: Is _up a typo?

stephanegigandet: ah no, I see it's defined after. Why the "_" before?

alexgarel: Because up is "build + _up"
While dev is "build + init_backend + _up + ..."

leading _ in python means "internal"

stephanegigandet (approved):

Issue #6132, opened 2021-11-30
Percentages as mg/kg not correctly parsed


aleene:
### Describe the bug

The fractions in the ingredient list are expressed as `Hierro 30 mg/kg, Ac. Fólico 2,2 mg/kg, Tiamina 6,3 mg/kg. Riboflavina (B2) 1,3 mg/kg, Niacina 13 mg/kg`. The slash is seen as a delimiter  and breaks the `mg` and `kg`.

### To Reproduce

See product: https://world.openfoodfacts.org/product/7790174001295/nuggets-vegetalex

### Expected behavior

In this case the slash should not be seen as a delimiter.

And for the ingredients recogniser the weights should be ignored.

### Part of
- #5706

NightWinkle:
I hit this while trying to use Open Pet Food Facts: ingredient analysis almost systematically looked wrong, and `kg` even shows among the most frequent “ingredients” (from `mg/kg` being split on `/`). A quick visual sweep of most frequent ingredient also revealed similar problems for parsing of other "biologic activity" units like U.I. and UFC.

Examples:
- [mg/kg (product 7141237765577)](https://fr.openpetfoodfacts.org/produit/7141237765577/poulet-thon-saumon)
- [U.I and UFC (product 3760260843959)](https://fr.openpetfoodfacts.org/produit/3760260843959/croquettes-chien-sans-cereales-dinde-riche-en-canard-atavik)

PR https://github.com/openfoodfacts/openfoodfacts-server/pull/14163

PR #6133, opened 2021-11-30
Spanish ingredients

aleene:
**Description:**

Added spanish language ingredients

**Related issues and discussion:** #6132

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -23799,6 +23799,12 @@ fr:arôme naturel d'aneth
 # ingredient/fr:arome-naturel-d'aneth has 34 products in 1 language @2020-05-21
 
 <en:flavouring
+en:chicken flavouring
+es:sabor pollo
+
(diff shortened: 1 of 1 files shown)


taxonomies/ingredients.txt
```diff
@@ -23798,6 +23798,9 @@ de:natürliches Dill-Aroma, natürliches Dillaroma
 fr:arôme naturel d'aneth
 # ingredient/fr:arome-naturel-d'aneth has 34 products in 1 language @2020-05-21
 
+<en:flavouring
+es:sabor idéntico al natural pollo
```
stephanegigandet, line 23802: According to the legal definition, this is an artificial chicken flavouring that has the same molecules as the ones that can be found in chicken: http://www.sice.oas.org/trade/mrcsrs/resolutions/an4693.asp

So we can put it under <en:chicken flavouring

The English term is "Nature-identical flavour".

So we can create an entry "Nature-identical chicken flavour"

I think it's best to try to always add an English name for all entries we add.

stephanegigandet (approved):
Thanks!

PR #6143, opened 2021-12-01
Simplify donation banner translations

stephanegigandet:
- Remove some hardcoded translations from the donation banner template
- Make the year and the number of employees variables so that we don't have to update the translations every year


alexgarel (approved):
LGTM + testé ok.

alexgarel (approved):

PR #6147, opened 2021-12-02
ci: run dev workflow on pull requests

ocervell:
This will help validate the developer workflow on new changes

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -170,3 +170,16 @@ jobs:
         
         # Run unit tests
         prove -l --jobs 2
+
+  dev:
+    name: Run developer workflow
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/pull_request.yml
```diff
@@ -170,3 +170,24 @@ jobs:
...
+        fetch-depth: 1
+    # - uses: satackey/action-docker-layer-caching@v0.0.11
+    #   continue-on-error: true
+    - run: |
+        make dev
+        make status
+      env:
+        DOCKER_BUILDKIT: 1
+    # - run: |
+    #     make tests
```
alexgarel, line 189: Using #6155 you would add make build_taxonomies before make tests

Note that we could also change the upper part to use:

make checks
make check_perl

alexgarel (approved):
LGTM

PR #6149, opened 2021-12-03
move users_emails.sto and orgs_glns.sto to /users and /orgs

### LINKED ISSUE #6002: Move users_emails.sto to users folders
users_emails.sto is really part of users data. So we should move it to users directory so that:
- it's in the same volume as users
- it's easier to zclone from prod this file also
- we don't need a separate mount for this very file.

stephanegigandet:
fixes #6002

### DIFF
--- a/cgi/reset_password.pl
+++ b/cgi/reset_password.pl
@@ -72,7 +72,7 @@ if ($action eq 'process') {
 	# Is it an email?
 
 		if ($id =~ /\@/) {
-			my $emails_ref = retrieve("$data_root/users_emails.sto");
+			my $emails_ref = retrieve("$data_root/users/users_emails.sto");
 			if (not defined $emails_ref->{$id}) {
 				# not found, try with lower case email
 				$id = lc $id;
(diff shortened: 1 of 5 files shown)


cgi/reset_password.pl
```diff
@@ -72,7 +72,7 @@
 	# Is it an email?
 
 		if ($id =~ /\@/) {
-			my $emails_ref = retrieve("$data_root/users_emails.sto");
+			my $emails_ref = retrieve("$data_root/users/users_emails.sto");
```
alexgarel, line 75: To ease transition on production, I propose to already move the files to users/ and orgs/ but symlink them in the upper directory.

alexgarel (approved):
:clap: cool one :-)

cgi/reset_password.pl
alexgarel: pour l'info je l'ai fait en preprod, pour tester.

stephanegigandet: sounds good

PR #6150, opened 2021-12-03
Ag docker pro

alexgarel:
This should enable using pro platform for dev (no deployment yet).

There is an explanation.
Docker file are a bit complexified, but I hope not too much.

### DIFF
--- a/docs/explanations/pro-dev-setup.md
+++ b/docs/explanations/pro-dev-setup.md
@@ -4,11 +4,11 @@ This explains how we setup docker file for pro platform development.
 For explanations on how to use it, see: [how-to-guides/pro-development](../how-to-guides/pro-development.md)
 
 off is the public facing application (world.openfoodfacts.org)
-off-pro is the pro platform (world.pro.openfoodfacts.org)
+off-pro is the producers platform (world.pro.openfoodfacts.org)
 
 When we work on the pro platform for development we want:
-* off containers to talk between them, and have their own volumes
-* off-pro containers to talk between them, and, generally, have their own volumes
+* off containers to talk between each other, and have their own volumes
+* off-pro containers to talk between each other, and, generally, have their own volumes
 * minion and backend from both app access to the same postgres database
   (which stores tasks queues)
 * off and off-pro backends / minion needs to share some volumes :
@@ -33,12 +33,12 @@ we will use postgres database from off as the common database.
 
 In order to achieve that:
 * we use profiles, so we won't start postgres in pro docker-compose
-* we connect postgres, backend and minion to a shared network, called minion_db
+* we connect `postgres`, `backend` and `minion` services to a shared network, called `minion_db`
 Fortunately this works, but note that there is a pitfall:
-on minion_db network both "backend" service will respond to same name.
+on `minion_db` network both `backend` services (`off` and `off-pro`) will respond to same name.
 For the moment it is not a problem for we don't need to communicate directly
 between instances.
-If it was, we would have to define custom aliases for those services on the minion_db network.
+If it was, we would have to define custom aliases for those services on the `minion_db` network.
 
 network    OFF              network       PRO              network
 po_default containers       minion_db     containers       po_pro_default
--- a/docs/explanations/pro-dev-setup.md
+++ b/docs/explanations/pro-dev-setup.md
@@ -1,0 +1,58 @@
+# Docker Setup of pro platform for development
+
+This explains how we setup docker file for pro platform development.
+For explanations on how to use it, see: [how-to-guides/pro-development](../how-to-guides/pro-development.md)
+
+off is the public facing application (world.openfoodfacts.org)
+off-pro is the pro platform (world.pro.openfoodfacts.org)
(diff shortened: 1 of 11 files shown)


docs/how-to-guides/pro-development.md
```diff
@@ -0,0 +1,44 @@
+# develop on the pro platform
+
+Here is how to develop for the pro platform using docker.
+
+It suppose [you already have setup docker for dev](../introduction/dev-environment-quick-start-guide.md).
+
+You should have two kind of shell:
+- the shell for openfoodfacts
+- the shell for openfoodfacts-pro, this is a shell where you have source the `setenv.sh`,
```
stephanegigandet, line 9: Shouldn't it be setenv-pro.sh  ?

stephanegigandet (approved):
The approach looks good to me

docs/how-to-guides/pro-development.md
alexgarel: you're right.

alexgarel:
@hangy  and @ocervell  It would be cool if you can review.
I'll be back only on wednesday, so you have time :-)

setenv-pro.sh
```diff
@@ -0,0 +1,16 @@
...
+export PRODUCERS_PLATFORM=1
+export MINION_QUEUE=pro.openfoodfacts.localhost
+# use a different port for minion admin
+# Note: we do not change PRODUCT_OPENER_DOMAIN for Config2.pm will handle this
+# neither do we change PRODUCT_OPENER_FLAVOR, as off-pro use same contents as off
+
+set_ps1 () {
+    export PS1=${PS1:-}"(pro) "
+}
+set_ps1;
```
ocervell, line 16: maybe instead of this script I would propose a ready-to-go .env.pro file that people can switch to by simply setting `ENV_FILE=.env.pro`: we could then have an `.env.obf`, `.env.opf` etc that has the default setup for each flavor and the setup would not be too complicated.

alexgarel: the problem is that you would have to maintain two .env that are very close, and you can't combine .env.

Working on the pro platform is not a thing a lot of contributor will do. So I think it's ok.

docs/how-to-guides/pro-development.md
```diff
@@ -0,0 +1,44 @@
...
+  Your prompt, should now contains a `(pro)` to recall you you are in pro environment.
+  (this simply sets some environment variables that will overides the one in .env)
+
+To develop, on pro plateform, you can then us a shell for openfoodfacts-pro and simply do a `make dev` and everything as usual.
+
+If you need to work on product import/export, or interacting with public platform,
+you have to start postgres and the minion on off side.
+That is, in a *non pro* shell, run `docker-compose up postgres minion`.
+
+Note that the setup does not currently support running the http server for both public and pro platform at the same time.
```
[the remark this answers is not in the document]
alexgarel: yes but `openfoodfacts.localhost` and `openfoodfacts.pro.localhost` have to be different instance of product opener.
Those instances are not on the same network.
If we want to have nginx from one project serving both, it will require a specific network + more configuration. 
I think this is out of scope for:
- on preprod / prod the host nginx will do the job
- on dev not many people have engouh RAM to run two instances of productopener, and you can start / stop the one you need, to check everything is ok (import / export does not need the public instance to run).

docs/how-to-guides/pro-development.md
```diff
@@ -0,0 +1,44 @@
+# develop on the pro platform
+
+Here is how to develop for the pro platform using docker.
+
+It suppose [you already have setup docker for dev](../introduction/dev-environment-quick-start-guide.md).
+
+You should have two kind of shell:
+- the shell for openfoodfacts
+- the shell for openfoodfacts-pro, this is a shell where you have source the `setenv-pro.sh`,
```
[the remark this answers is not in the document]
alexgarel: I think it will be quite a burden to maintain a .env for producer platform just to tweak several variables, that's why I propose the script approach for now.

In the future we may change that, but maybe this would mean generating a .env from two files ? (maybe thanks to an entry in the Makefile ?)

With the script, the Makefile works (environment variables have precedence over the .env).

docs/explanations/pro-dev-setup.md
```diff
@@ -0,0 +1,58 @@
...
+
+To satisfy the access to the same database,
+we will use postgres database from off as the common database.
+
+In order to achieve that:
+* we use profiles, so we won't start postgres in pro docker-compose
+* we connect postgres, backend and minion to a shared network, called minion_db
+Fortunately this works, but note that there is a pitfall:
+on minion_db network both "backend" service will respond to same name.
+For the moment it is not a problem for we don't need to communicate directly
```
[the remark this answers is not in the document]
alexgarel: No it's not a problem. What I mean here, is that backend for producers does not need to talk to public backend. They never communicate directly, the only communication for now is through posting task in the postgres that the minions read, and by sharing files.

ocervell (approved):

PR #6162, opened 2021-12-07
Start of ingredients knowledge panel

stephanegigandet:
Just a basic panel to display the ingredients list.

image

Also include new fields that can be used to add an Edit button in the app and/or the website:

```
{
elements: [
{
element_type: "text",
text_element: {
html: "(French)"
}
},
{
element_type: "text",
text_element: {
edit_field_id: "ingredients_text_fr",
edit_field_type: "ingredients_text",
edit_field_value: "Eau, boulgour 15% (blé dur concassé) grand épeautre 9% ( contient gluten), carottes 5,3%, courgettes 5,3%, pois chiches 5,3%, poivrons rouges 4%, oignons 3,3%, huile d'olive vierge extra, raisins secs 1,6% (raisins secs, huile de coton), double concentré de tomate, farine de blé, épices 0,8% (coriandre, anis, cannelle, girofle, poivre, curcuma, cumin, fenugrec, fenouil, piment, cardamome, livèche, ail, paprika), coriandre 0,4 %, purée d'ail, sel. Fabriqué dans un atelier qui utilise du poisson, des oeufs et de la moutarde.",
html: "Eau, <span class="allergen">boulgour</span> 15% (blé dur concassé) <span class="allergen">grand épeautre</span> 9% ( contient <span class="allergen">gluten</span>), carottes 5,3%, courgettes 5,3%, pois chiches 5,3%, poivrons rouges 4%, oignons 3,3%, huile d'olive vierge extra, raisins secs 1,6% (raisins secs, huile de coton), double concentré de tomate, farine de <span class="allergen">blé</span>, épices 0,8% (coriandre, anis, cannelle, girofle, poivre, curcuma, cumin, fenugrec, fenouil, piment, cardamome, livèche, ail, paprika), coriandre 0,4 %, purée d'ail, sel. Fabriqué dans un atelier qui utilise du <span class="allergen">poisson</span>, des <span class="allergen">oeufs</span> et de la <span class="allergen">moutarde</span>."
}
}
],
level: "info",
topics: [
"health"
],
type: "inline"
},
```

### DIFF
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -626,6 +626,9 @@ sub create_health_card_panel($$$) {
     # Create the physical activities panel
     create_physical_activities_panel($product_ref, $target_lc, $target_cc);
 
+    # Create the ingredients panel
+    create_ingredients_panel($product_ref, $target_lc, $target_cc);
+
     # Create the health_card panel
     create_panel_from_json_template("health_card", "api/knowledge-panels/health/health_card.tt.json",
         $panel_data_ref, $product_ref, $target_lc, $target_cc);    
@@ -684,7 +687,7 @@ sub create_nutriscore_panel($$$) {
 
 =head2 create_nutrition_facts_table_panel ( $product_ref, $target_lc, $target_cc )
 
-Creates a knowledge panels with the nutrition facts table.
+Creates a knowledge panel with the nutrition facts table.
 
 =head3 Arguments
 
@@ -837,5 +840,56 @@ sub create_physical_activities_panel($$$) {
     }
 }
 
+
+=head2 create_ingredients_panel ( $product_ref, $target_lc, $target_cc )
+
+Creates a knowledge panels with the list of ingredients.
+
+=head3 Arguments
+
+=head4 product reference $product_ref
+
+Loaded from the MongoDB database, Storable files, or the OFF API.
+
+=head4 language code $target_lc
+
+Returned attributes contain both data and strings intended to be displayed to users.
+This parameter sets the desired language for the user facing strings.
+
+=head4 country code $target_cc
+
+=cut
+
+sub create_ingredients_panel($$$) {
+
+	my $product_ref = shift;
+	my $target_lc = shift;
+	my $target_cc = shift;
+
+	$log->debug("create ingredients panel", { code => $product_ref->{code}, nutriscore_data => $product_ref->{nutriscore_data} }) if $log->is_debug();
+
+	# try to display ingredients in the requested language if available
+
+	my $ingredients_text = $product_ref->{ingredients_text};
+    my $ingredients_text_with_allergens = $product_ref->{ingredients_text_with_allergens};
+	my $ingredients_text_lc = $product_ref->{lc};
+
+	if ((defined $product_ref->{"ingredients_text" . "_" . $target_lc}) and ($product_ref->{"ingredients_text" . "_" . $target_lc} ne '')) {
+		$ingredients_text = $product_ref->{"ingredients_text" . "_" . $target_lc};
+		$ingredients_text_with_allergens = $product_ref->{"ingredients_text_with_allergens" . "_" . $target_lc};
+		$ingredients_text_lc = $target_lc;
+	}
+
+    my $panel_data_ref = {
+        ingredients_text => $ingredients_text,
+        ingredients_text_with_allergens => $ingredients_text_with_allergens,
+        lc => $target_lc,
+        ingredients_text_lc => $ingredients_text_lc,
+        ingredients_text_language => display_taxonomy_tag($target_lc,'languages',$language_codes{$ingredients_text_lc}),
+    };
+
+    create_panel_from_json_template("ingredients", "api/knowledge-panels/health/ingredients/ingredients.tt.json",
+        $panel_data_ref, $product_ref, $target_lc, $target_cc);
+}
 
 1;
--- a/templates/api/knowledge-panels/health/ingredients/ingredients.tt.json
+++ b/templates/api/knowledge-panels/health/ingredients/ingredients.tt.json
@@ -3,13 +3,26 @@
     "topics": [
         "health"
     ],
-    "elements": [   
+    "type": "inline",
+    "elements": [
+        [% IF panel.lc != panel.ingredients_text_lc %]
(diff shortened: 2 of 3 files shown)


stephanegigandet:
> question regarding the id "ingredient_text_fr" vs type "ingredient_text". Is there a use case for passing the language ? Should the client use that somehow.

Yes, there isn't a single "ingredients_text" field: each product can have ingredients in multiple languages  (e.g. in German, Italian and French in Switzerland). So we need to know which one we are displaying, and more importantly, which one the user is changing / adding.

> Also how should the client use the `type` and `id` fields ? what other types and ids can there be ? If the type is to help the client decipher that the data is in free text form, maybe we should just set `type` as `text`

The type is to indicate to the client that it is an ingredients list. The client could then offer the user to take a picture of the ingredients, do the OCR etc.

Other potential types: "packaging_text" for the packaging / recycling instructions.

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -838,4 +841,55 @@ sub create_physical_activities_panel($$$) {
...
+        ingredients_text_with_allergens => $ingredients_text_with_allergens,
+        lc => $target_lc,
+        ingredients_text_lc => $ingredients_text_lc,
+        ingredients_text_language => display_taxonomy_tag($target_lc,'languages',$language_codes{$ingredients_text_lc}),
+    };
+
+    create_panel_from_json_template("ingredients", "api/knowledge-panels/health/ingredients/ingredients.tt.json",
+        $panel_data_ref, $product_ref, $target_lc, $target_cc);
+}
+
```
alexgarel, line 894: Minor: I see indentation problems (as displayed here).

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -838,4 +841,55 @@ sub create_physical_activities_panel($$$) {
...
+	$log->debug("create ingredients panel", { code => $product_ref->{code}, nutriscore_data => $product_ref->{nutriscore_data} }) if $log->is_debug();
+
+	# try to display ingredients in the requested language if available
+
+	my $ingredients_text = $product_ref->{ingredients_text};
+    my $ingredients_text_with_allergens = $product_ref->{ingredients_text_with_allergens};
+	my $ingredients_text_lc = $product_ref->{lc};
+
+	if ((defined $product_ref->{"ingredients_text" . "_" . $target_lc}) and ($product_ref->{"ingredients_text" . "_" . $target_lc} ne '')) {
```
alexgarel, line 877: (thinking) I find it's a pity not to have some small utils functions for this kind of pattern, that would make it more readable. Something like `is_non_empty_text($product_ref->{"ingredients_text" . "_" . $target_lc})`

templates/api/knowledge-panels/health/ingredients/ingredients.tt.json
```diff
@@ -3,13 +3,26 @@
     "topics": [
         "health"
     ],
-    "elements": [   
+    "type": "inline",
+    "elements": [
+        [% IF panel.lc != panel.ingredients_text_lc %]
```
alexgarel, line 8: This condition is not really clear to me. Is that because if lc is same as ingredients_text_lc we already displayed it ? If so, a comment could be helpful.

alexgarel (approved):
LGTM, minor comments on style.

jasmeet0817:
> > question regarding the id "ingredient_text_fr" vs type "ingredient_text". Is there a use case for passing the language ? Should the client use that somehow.
> 
> Yes, there isn't a single "ingredients_text" field: each product can have ingredients in multiple languages (e.g. in German, Italian and French in Switzerland). So we need to know which one we are displaying, and more importantly, which one the user is changing / adding.

Hmm ok, but I'm guessing 1 API response will only contain the `ingredient_text` in a single language? Based on the language that the response was requested in ? Is that correct?

If ingredient_id is there just to pass the language, should we rename the field to `language` instead? Maybe we can structure the response a bit, suggestion: 

`
text_element: {
edit_field: [
 type: 'ingredient_text",
 text: {
  language: 'fr'
  value: 'French ingredients'
 },
}
`

> 
> > Also how should the client use the `type` and `id` fields ? what other types and ids can there be ? If the type is to help the client decipher that the data is in free text form, maybe we should just set `type` as `text`
> 
> The type is to indicate to the client that it is an ingredients list. The client could then offer the user to take a picture of the ingredients, do the OCR etc.
> 
> Other potential types: "packaging_text" for the packaging / recycling instructions.

stephanegigandet:
> Hmm ok, but I'm guessing 1 API response will only contain the `ingredient_text` in a single language? Based on the language that the response was requested in ? Is that correct?

The API will return only one language, but it might not be the language requested. If we do have the language requested, we will return it, but if not, we will return another language, and the client can display it, with the language name.

stephanegigandet:
> text_element: { text: { language: 'fr' value: 'French ingredients' }, editable: true, type: 'ingredient_text", }

We use text elements just for text in many other panels though, so maybe we can create a new element type instead. e.g. "field_element".

templates/api/knowledge-panels/health/ingredients/ingredients.tt.json
stephanegigandet: It's because we display something like "Ingredients (Spanish) : leche de vaca" when the language of the ingredients is different from the language of the website / the app. If you are on the French site and we have ingredients in French, we just display "Ingrédients : lait de vache", not "Ingrédients (français) : lait de vache".

stephanegigandet:
@jasmeet0817 I changed the structure to pass the language code (lc field) and the language name in the language of the interface (language) field inside the text_element, that way the client can decide how to display the language (e.g. not at all, with a text, or with a flag icon for instance).

Here is a sample output for a knowledge panel when the client interface is in Spanish, but the product has only French ingredients:

```
{
element_type: "text",
text_element: {
edit_field_id: "ingredients_text_fr",
edit_field_type: "ingredients_text",
edit_field_value: "test d'ingrédients, oeuf, beurre, lait, _chocolat_",
html: "test d'ingrédients, <span class="allergen">oeuf</span>, <span class="allergen">beurre</span>, <span class="allergen">lait</span>, <span class="allergen">chocolat</span>",
language: "Francés",
lc: "fr"
}
}
```

jasmeet0817 (approved):

PR #6173, opened 2021-12-09
fix issue with nutrition facts edits through API

stephanegigandet:
recently deployed changes resulted in the checkboxes for "nutrition specified" to be turned off.


alexgarel (approved):
LGTM

PR #6178, opened 2021-12-10
ops: Change deploy for new pro setup

alexgarel:
Changes consequent to new setup for pro.


.github/workflows/container-deploy.yml
```diff
@@ -106,6 +108,7 @@ jobs:
           echo "PRODUCT_OPENER_FLAVOR=${{ secrets.PRODUCT_OPENER_FLAVOR }}" >> .env
           echo "PRODUCT_OPENER_FLAVOR_SHORT=${{ secrets.PRODUCT_OPENER_FLAVOR_SHORT }}" >> .env
           echo "MINION_MODE=production" >> .env
+          echo "MINION_QUEUE=openfoodfacts.net" >> .env
```
stephanegigandet, line 111: openfoodfacts.net instead of PRODUCT_OPENER_DOMAIN ?

alexgarel: oooh I see now what you meant !, I should have use the one in secret, you're right, sorry.

PR #6179, opened 2021-12-10
Ag pro ademe

alexgarel:
First perl ever :-D

I want to lower difference between csv and xls parsing, and test if official eco-score file would be parsed correctly. 

Every comment will be greatly appreciated (event style).


stephanegigandet:
It looks like we need to add ssconvert to the docker install:

Can't exec "ssconvert": No such file or directory at /home/runner/work/openfoodfacts-server/openfoodfacts-server/lib/ProductOpener/Producers.pm line 174.

stephanegigandet:
> I want to lower difference between csv and xls parsing

Good idea. :)

t/load_csv_or_excel_file.t
```diff
@@ -0,0 +1,52 @@
...
+use utf8;
+
+use Test::More;
+use Log::Any::Adapter 'TAP';
+
+use File::Basename "dirname";
+
+use ProductOpener::Producers qw(load_csv_or_excel_file);
+
+my $inputs_dir = dirname(__FILE__) . "/inputs/";
```
stephanegigandet, line 14: Good idea to add tests, they are indeed badly missing.

Maybe add the name of the test in the directory, like we do for t/expected_test_results, as we might have other types of inputs later.

t/load_csv_or_excel_file.t
```diff
@@ -0,0 +1,52 @@
...
+    "Ingrédient - 6", "Origine - 6", "Pourcentage - 6",
+    "Format", "Matériau",
+    "Format - 2", "Matériau - 2",
+    "Format - 3", "Matériau - 3",
+    "Huile de palme", "Poissons menacés"
+);
+
+# we have headers
+is_deeply(
+    \@{$results_ref->{headers}},
```
stephanegigandet, line 36:
```suggestion
    $results_ref->{headers},
```

@{$array_ref} dereferences $array_ref to make it an array, and then \ makes it a reference again.

stephanegigandet (approved):
Looks good to me, thank you!

Issue #6174, opened 2021-12-09
Rename PNNS Groups into Food Groups and refactor


stephanegigandet:
### What

We currently have 2 levels of "PNNS groups" that are derived from categories. Those PNNS groups are used by the French researchers who created the Nutri-Score for their studies. (PNNS stands for "Programme National Nutrition et Santé": the French National Nutrition and Health Program).

It is more and more used internationally though, so it would be a good thing to use a more descriptive name, like "Food groups".

In addition, the researchers want to add a 3rd level for some of the groups.

### Proposed solution

We could create a "Food groups" taxonomy, so that it's easier to navigate between the food groups, and so that we can translate them.

The food groups would still be populated from the categories. In the categories taxonomy, we currently have pnns_groups_1:en and pnns_groups_2:en properties to determine the pnns group of a product based on its categories.

stephanegigandet:
This has been done

PR #6183, opened 2021-12-10
First implementation of food groups to replace and extend PNNS groups

stephanegigandet:
See #6174 for the motivation of this PR.

This is a first implementation of food groups to replace the PNNS groups. For some time, we probably to have both running together, so that we can compare them, and make sure the new implementation works well.

- Moved all PNNS code to a new FoodGroups.pm module
- Created a food groups taxonomy used to determine food groups from categories.
- Basic implementation for assigning food groups from categories.
- Created a new food-groups facet
- Added tests

Will be done later:
- Assigning food_group_1, food_group_2, food_group_3 properties for each product.
- Reimplementing the special case handling (e.g. determining if a beverage is a beverage in the sense of the food groups (> 80% milk then not considered a beverage)


lib/ProductOpener/FoodGroups.pm
```diff
@@ -0,0 +1,437 @@
...
+	"Teas and herbal teas and coffees" => "Beverages",
+	"Plant-based milk substitutes" => "Beverages",
+
+	"Alcoholic beverages" => "Alcoholic beverages",
+
+	"unknown" => "unknown",
+
+);
+
+foreach my $group (keys %pnns) {
```
alexgarel, line 116:
```suggestion
# use string id in pnns hash
foreach my $group (keys %pnns) {
```

lib/ProductOpener/FoodGroups.pm
```diff
@@ -0,0 +1,437 @@
...
+
+=cut
+
+sub compute_food_groups($) {
+
+	my $product_ref = shift;
+
+    $product_ref->{nutrition_score_beverage} = is_beverage_for_nutrition_score($product_ref);
+
+	# Temporarily change categories
```
alexgarel, line 233:
```suggestion
	# Temporarily change categories (backup old one in original_categories_tags)
```

lib/ProductOpener/FoodGroups.pm
```diff
@@ -0,0 +1,437 @@
...
+
+	# For Open Food Facts, add special categories for beverages that are computed from
+	# nutrition facts (alcoholic or not) or the ingredients (sweetened, artificially sweetened or unsweetened)
+	# those tags are only added to categories_tags (searchable in Mongo)
+	# and not to the categories_hierarchy (displayed on the web site and in the product edit form)
+	# Those extra categories are also used to determined the French PNNS food groups
+
+	if (($product_ref->{nutrition_score_beverage}) and (not has_tag($product_ref,"categories","en:instant-beverages"))) {
+
+		if (defined $product_ref->{nutriments}{"alcohol_100g"}) {
```
alexgarel, line 308: suggestion:
I would have made two arrays: to_remove and to_add, to collect categories to eventually add / remove.
And only at the end I would have made a loop to remove / add what was to remove / add.
I think, It would make the code more easy to read.

lib/ProductOpener/FoodGroups.pm
```diff
@@ -0,0 +1,437 @@
...
+				or has_tag($product_ref, "ingredients", "sirop-de-sucre-de-canne") or has_tag($product_ref, "ingredients", "sucre-bio")
+				or has_tag($product_ref, "ingredients", "sucre-de-canne-liquide") or has_tag($product_ref, "ingredients", "sucre-de-betterave")
+				or has_tag($product_ref, "ingredients", "sucre-inverti") or has_tag($product_ref, "ingredients", "canne-sucre")
+				or has_tag($product_ref, "ingredients", "sucre-glucose-fructose") or has_tag($product_ref, "ingredients", "glucose-fructose-et-ou-sucre")
+				or has_tag($product_ref, "ingredients", "sirop-de-glucose") or has_tag($product_ref, "ingredients", "glucose")
+				or has_tag($product_ref, "ingredients", "sirop-de-fructose") or has_tag($product_ref, "ingredients", "saccharose")
+				or has_tag($product_ref, "ingredients", "sirop-de-fructose-glucose") or has_tag($product_ref, "ingredients", "sirop-de-glucose-fructose-de-ble-et-ou-de-mais")
+				or has_tag($product_ref, "ingredients", "sugar") or has_tag($product_ref, "ingredients", "sugars")
+				or has_tag($product_ref, "ingredients", "en:sugar")
```
alexgarel, line 372: Putting only one `has_tag` by line, seems more readable.

Really, I would use  `List::Util ``any`:
```perl
my @sweeteners = qw(sucre sucre-de-canne-roux sucre-caramelise ... )
if (any { has_tag($product_ref", "ingredients", $_)  } @sweeteners )
```

lib/ProductOpener/FoodGroups.pm
```diff
@@ -0,0 +1,437 @@
...
+
+=head3 Return values
+
+Categories can be added or removed in $product_ref->{categories_tags}
+
+Original categories are saved in $product_ref->{original_categories_tags}
+
+=cut
+
+sub temporarily_change_categories_for_food_groups_computation($) {
```
alexgarel, line 289: cool to have made a specific function for that.

taxonomies/food_groups.txt
```diff
@@ -0,0 +1,269 @@
+# The food groups taxonomy is a hierarchy with up to 3 levels.
+# Food groups are used by researchers (in particular the EREN team that created the Nutri-Score) for studies.
+# Each product can be in only 1 of each food group at each level.
+
+# The groups for each product are determined from its categories, thanks to properties in the categories taxonomy.
```
alexgarel, line 5: Maybe we should tell that this taxonomy entries do not need to be translated (or do they ?)

alexgarel (approved):
I propose some changes that are mostly style.

lib/ProductOpener/FoodGroups.pm
```diff
@@ -0,0 +1,437 @@
...
+
+use ProductOpener::Store qw/:all/;
+use ProductOpener::Config qw/:all/;
+use ProductOpener::Lang qw/:all/;
+use ProductOpener::Tags qw/:all/;
+use ProductOpener::Food qw/:all/;
+
+use Log::Any qw($log);
+
+my %pnns = (
```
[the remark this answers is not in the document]
stephanegigandet: It's existing code that I only moved from Food.pm and that we will get rid of, I can add a few comments but we will remove it soon hopefully.

lib/ProductOpener/FoodGroups.pm
stephanegigandet: Good idea, but same here, we will soon replace all that code because we will have a single en:added-sugar parent.

taxonomies/food_groups.txt
stephanegigandet: My plan is to make food groups much more visible than PNNS groups, so it would be a good thing to have them translated eventually.

lib/ProductOpener/FoodGroups.pm
alexgarel: yes ok, so leave it that way :-)

PR #6188, opened 2021-12-13
feat: Start of product rules: simple rule format

stephanegigandet:
An idea that was discussed with contributors during the "OFF days" this weekend is to enable the creation of rules that can be applied automatically to products to complete their data.

For instance if a product is the yoghurts category, and one of its ingredient is sugar, then we could add it to the sweetened yoghurts category if it's not in there already.

This PR is to propose a simple rule format that we can use internally in Product Opener to store rules and apply them.

I chose a simple text format (similar to the taxonomies) so that it can easily be read/written without a specific interface, and so that we can have diffs in github.

If we create a specific interface to experiment with rules, test them, validate them etc. we could of course have that tool output such a format. So it doesn't mean that people will need a text editor to propose rules.

Once we have rules defined, we can then do different things:

1. Evaluate them. e.g. count how many products would be affected.
2. Execute them and apply their actions to complete products.

The execution of the rules can be done in real time (anytime products are saved) and/or in batch.

The application of the result of the rules (e.g. a new category) can be done either on the source data (adding a category as if it was added by a user), or the resulting data (adding a category but only on the resulting data, a bit like we extract ingredients from the ingredients list each time we save the product).


rules/yoghurts.txt
```diff
@@ -0,0 +1,27 @@
...
+
+# General rules syntax:
+
+# name: a name to identify the rule
+
+# description: an explanation of what the rule does
+
+# conditions: 1 or more conditions, in the same format as the search API v2 (makes copy and pasting easy to test rules)
+# e.g. https://world.openfoodfacts.org/search?categories_tags_en=yoghurts,-sweetened%20yoghurts&ingredients_tags_en=sugar
+# - conditions are comma separated, and can also be specified on several conditions: lines
```
alexgarel, line 11: you should detail:
do we do an "and" on all conditions ?
Or do we do an "and" on conditions on same line, and "or" between lines ?
Or do we do an "and" on conditions expressed with "&" sign, and an or between those, separated by comma or line ?

alexgarel (commented):
Overall I'm ok but you should detail more, maybe.

I like the idea that we stay close to the API, such that adding more functionalities to API = adding more functionalities for filtering rules.

I wonder if we could quickly also have a mean to define conditions that are re-usable because this could helps readability (and I think this is easily done in perl).
like:

```
condition:yoghourt_without_sugar: categories_tags_en=yoghurts,-sweetened yoghurts
```

```
condition: $yoghourt_witch_sugar&ingredients_tags_en=sugar
```

rules/yoghurts.txt
stephanegigandet: We do an "and" on all conditions. Initially I thought to have only 1 conditions: line, but then I'm thinking it could be more readable to have multiple lines.

e.g. in yoghurts.txt we could have all rules targeting yoghurts, and so the first conditions: line could be just to check that the category is yoghurts.

stephanegigandet:
> condition: $yoghourt_witch_sugar&ingredients_tags_en=sugar

We could do something like that, but then we lose a bit the ability to quickly test conditions by copying and pasting them in a URL. (or we need to make sure we see all the $ signs and replace them accordingly).

teolemon:
- [ ] I can share this collaborative list of potential rules from the Regex Engine folder: https://docs.google.com/spreadsheets/d/1v62INXh8cKqyHXXKnj1nsGBb6KaICu14tV8f1uPrWHs/edit#gid=0
- [ ] Here's the folder of Regex Engine, which adopts a web-based, collaborative and iterative approach, with a rules database and rules management: https://drive.google.com/drive/u/1/folders/1YY8NM-7MIXNYPlBhI0JSjK-P1Rv36FZC

teolemon:
I reformated the taxonomies in #6196 to create space for impact for an `origins:en:` rule

teolemon (approved):

alexgarel:
@stephanegigandet why is this PR stalled ?

teolemon:
I'm adding a P1, because everyone agrees the potential of the idea/feature.

teolemon:
Associated issue: https://github.com/openfoodfacts/openfoodfacts-server/issues/3426

alexgarel:
@teolemon put it in discussion for next call maybe ?

stephanegigandet:
This was never done, so closing the PR, we can create a new one if we want to do this at some point

PR #6194, opened 2021-12-14
fix precedence issue with or statement

stephanegigandet:
I didn't understand why we had a warning:

Useless use of string eq in void context at /opt/product-opener/lib/ProductOpener/Display.pm line 9533.

That's because of the very weak precedence of the "or" operator:

https://perldoc.perl.org/perlop#Logical-or-and-Exclusive-Or

```
$x = $y or $z;              # bug: this is wrong
($x = $y) or $z;            # really means this
$x = $y || $z;              # better written this way
```


alexgarel (approved):
:+1:

PR #6202, opened 2021-12-16
chore: Vegan categories

stephanegigandet:
Added a few synonyms to vegetarian categories


blazern:
I'm a bit confused :)

Does this change mean that when a person wants to get vegan spreads / nuggets / something else, they have to:
- Request products with the "vegan XXX" category
- Ensure the returned products have the "vegan" label, or all of their ingredients are vegan
?

Because, as we've discussed in Slack, generally speaking "vegetarian XXX", "plant-based XXX" and "vegan XXX" are quite different things:
- A product can be plant-based, but not vegan or vegetarian.
- A product can be vegetarian, but not plant-based or vegan.
- etc

stephanegigandet:
Today we have a "vegetarian nuggets" category, but no "vegan nuggets" category. This change means that if someone edits a product to put "vegan nuggets", it will be added to "vegetarian nuggets" instead.

I think that's good for categories where almost all products are in fact vegan, but where the most common name for it is "vegetarian something".

The other option is to have both "vegetarian nuggets" and "vegan nuggets" categories, but it's likely that they will contain almost identical sets of products.

blazern:
Now I don't know, where the changes should be discussed, but here's a screenshot with the latest discussion from Slack, just in case.
image

teolemon (approved):
LGTM @stephanegigandet

stephanegigandet:
I made a separate entry for vegan nuggets, as discussed on Slack with @blazern and @jolesh

PR #6203, opened 2021-12-16
fix: allow unchecking checked boxes in product edit form

stephanegigandet:
This fixes an issue with my recent refactoring of the code to handle nutrition facts edit from the web product edit form and the API.

The issue is that when a checkbox is unchecked on a form, the browser actually does not send the field at all. So if the parameter is not there, we don't know if it's because it is not set on purpose (e.g. the API is changing something else), or the box is there but has been unset (through the web edit form).

The trick is to add a hidden field with the same name of the checkbox on the web form. That way we do get the hidden value if the box is unchecked, and the checked value if it is checked.

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -2576,26 +2576,27 @@ sub assign_nutriments_values_from_request_parameters($$) {
 
 	$log->debug("Nutrition data") if $log->is_debug();
 
-	# Note: browsers do not send any value for checkboxes that are unchecked.
-	# So in order to be able to uncheck values, there needs to be an hidden field with the same name with a false value before the checkbox
-	# and the server will only keep the last value: either the hidden field, or the checkbox if checked.
+	# Note: browsers do not send any value for checkboxes that are unchecked,
+	# so the web form also has a field (suffixed with _displayed) to allow us to uncheck the box.
 
-	if (defined param("no_nutrition_data")) {
-		$product_ref->{no_nutrition_data} = remove_tags_and_quote(decode utf8=>param("no_nutrition_data"));
-	}
+	# API:
+	# - check: no_nutrition_data is passed "on" or 1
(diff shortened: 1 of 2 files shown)


templates/web/pages/product_edit/product_edit_form_display.tt.html
```diff
@@ -96,12 +96,16 @@
 
         <!--nutrient fieldset-->
         <div class="fieldset" id="nutrition"><legend>[% lang('nutrition_data') %]</legend>
+            <!-- empty hidden checkbox with the same name so that the browser sends a value when the box is unchecked -->
+            <input type="hidden" name="no_nutrition_data" value="" />
```
alexgarel, line 100: Hum this does not work for me !

If I check the box, this is not taken into account.

In Food.pm, in `assign_nutriments_values_from_request_parameters`, `param("no_nutrition_data")` is `''`

alexgarel (changes requested):
Does not work for me.

stephanegigandet:
@alexgarel I updated the code to add a new hidden field, as you suggested if I remember correctly.

lib/ProductOpener/Food.pm
```diff
@@ -2576,22 +2576,27 @@ sub assign_nutriments_values_from_request_parameters($$) {
...
-	if (defined param("nutrition_data_prepared")) {
-		$product_ref->{nutrition_data_prepared} = remove_tags_and_quote(decode utf8=>param("nutrition_data_prepared"));
-	}
+		if (defined param($checkbox)) {
+			$product_ref->{$checkbox} = remove_tags_and_quote(decode utf8=>param($checkbox));
+		}
+		elsif (defined param($checkbox . "_displayed")) {
+			$product_ref->{$checkbox} = "";
+		}
+	}
```
alexgarel, line 2599: Cool

PR #6205, opened 2021-12-16
NOVA updates for starches + tests

stephanegigandet:
According to the NOVA paper, starches are NOVA 2: https://www.fao.org/3/ca5644en/ca5644en.pdf

We previously made them NOVA 3.

Also added some tests.


alexgarel (approved):
LGTM

PR #6206, opened 2021-12-16
fix: docker networks in prod after exporters crashing

### LINKED ISSUE #6204: nginx-exporter and apache-exporter down
### Describe the bug

Both exporters need to be on the same network as the backend / frontend containers in order to work properly.

The exporters were broken after https://github.com/openfoodfacts/openfoodfacts-server/pull/6150/files#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3L63 removed the belonging of the `backend` and `frontend` services to the same `webnet` network.

### To Reproduce

Include `docker/monitor.yml` in your deployment `COMPOSE_FILE` and run `make dev`: the exporters crash.

### Expected behavior

Exporters should not crash.

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Other

### Browser version

_No response_

_No response_

### Time per product

_No response_

ocervell:
This fixes #6204

### DIFF
--- a/docker/prod.yml
+++ b/docker/prod.yml
@@ -6,12 +6,16 @@ services:
     restart: always
   backend:
     restart: always
+    networks: 
+    - webnet
   incron:
     restart: always
   minion:
     restart: always
   frontend:
     restart: always
+    networks:
+    - webnet
 volumes:
   html_data:
     external: true
--- a/docker/prod.yml
+++ b/docker/prod.yml
@@ -1,17 +1,25 @@
-﻿version: "3.7"
+version: "3.7"
 services:
   memcached:
     restart: always
+    networks:
+    - webnet
   postgres:
     restart: always
+    networks:
+    - webnet
   backend:
     restart: always
     networks: 
     - webnet
   incron:
     restart: always
+    networks:
+    - webnet
   minion:
     restart: always
+    networks:
+    - webnet
   frontend:
     restart: always
     networks:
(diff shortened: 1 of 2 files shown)


docker/prod.yml
```diff
@@ -6,12 +6,16 @@ services:
...
+    networks: 
+    - webnet
   incron:
     restart: always
   minion:
     restart: always
   frontend:
     restart: always
+    networks:
+    - webnet
```
alexgarel, line 18: I think you should either:
- put `webnet` to every services
- or put `webnet` and `default` to services you want to monitor

In you config, postgres  / incron stays on default, while backend / frontend would only be on webnet

(or Am I wrong ?)

alexgarel (changes requested):

docker/prod.yml
ocervell: What does `default` add ? The idea is just to have `frontend`, `backend` and the `apache_exporter`, `nginx_exporter` on the same network so that they can communicate.

alexgarel: default is the default network, where all service without a networks directive are located :-)

alexgarel: if you want backend to be able to communicate with postgres, it will have to be in default network (as well as webnet).

alexgarel (approved):
LGTM

PR #6209, opened 2021-12-17
Localize Eco-Score data even if we don't have an Eco-Score category, for the origins of ingredients knowledge panel

stephanegigandet:
This is so that we can compute environment knowledge panels for the origins of ingredients even if we don't have an Eco-Score because we are missing a category.


alexgarel (approved):
LGTM

PR #6215, opened 2021-12-20
fix Equadis import of Nutri-Score field

stephanegigandet:
The format of the XML we get from Equadis seems to have changed, at least for Pepsico, the value of the Nutri-Score computed by the producer is now in an extra "nutritionalProgram" layer.


alexgarel (approved):
Good

PR #6217, opened 2021-12-20
Feat: Add Top Issues

teolemon:
Add Top Issues

### DIFF
--- a/.github/workflows/top-issues.yml
+++ b/.github/workflows/top-issues.yml
@@ -1,0 +1,18 @@
+name: Label Top Issues
+
+on:
+  schedule:
+  - cron:  '15 0-23 * * *'
+
+jobs:
+  labelTopIssues:
+    name: Label Top Issues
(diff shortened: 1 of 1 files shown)


.github/workflows/top-issues.yml
```diff
@@ -0,0 +1,21 @@
+name: Label Top Issues
+
+on:
+  schedule:
+  #- cron:  '15 0-23 * * *'
+  push:
+  #  branches: [top-issues]
```
alexgarel, line 7: I think cron make sense here, better than push.
Maybe once a week is engouh ?
```suggestion
  schedule:
  # every sunday at 23
  - cron:  '0 23 * * 0'
```

alexgarel (approved):
@teolemon I'm ok modulo a schedule I propose.

teolemon:
I disabled it because it was not behaving properly.

Issue #5807, opened 2021-10-03
Password reset capitalization


M123-dev:
### Describe the bug

If you want to reset your password but enter the email with the wrong capitalization, it says "There is no account with this email".

### To Reproduce

Try [requesting a password reset email](https://world.openfoodfacts.org/cgi/reset_password.pl), but make the first letter uppercase (or lowercase)

### Expected behavior

The capitalization should be ignored

### Type of device

Browser (Firefox)

alexgarel:
I'll give it a try as a first bug.

alexgarel:
Huuuum, in fact we never apply transformations on email addresses, so that ain't so easy. 

I checked in prod, we have something like 10215 out of 110262 users which registered with an email with a capital letter inside.

What I propose:

1. I will implement a fix that will work on most case: if the email is not found as is, try with lower cased email
2. We should open up a new bug to understand if we want to always lowercase email addresses in our database (or at least in user_emails.sto ?). That could be only for new registrations.

@stephanegigandet any though on second point ?

stephanegigandet:
Thanks @alexgarel. I think it would make a lot of sense to lowercase all emails in user_emails.sto (existing and new) so that we solve the issue for emails for uppercase, and not only for password reset, but also for logging in. So the approaches you suggest sound good to me.

M123-dev:
@alexgarel. just tried it again and it still does not work as expected, from what I can take out of https://github.com/openfoodfacts/openfoodfacts-server/commit/527e31760e47d240a85ffb0e48e4bb74d0754b9c 

`$id = lc $id;`

it only makes the provided one lowercase not the saved one. Mine is saved with a capital M at the beginning

PR #6221, opened 2021-12-23
Fix: Reset email capitalization

M123-dev:

Follow up for #5807 and https://github.com/openfoodfacts/openfoodfacts-server/commit/527e31760e47d240a85ffb0e48e4bb74d0754b9c

**What happend**

> If you want to reset your password but enter the email with the wrong capitalization, it says "There is no account with this email".

In https://github.com/openfoodfacts/openfoodfacts-server/commit/527e31760e47d240a85ffb0e48e4bb74d0754b9c @alexgarel updated the code to lowercase the provided email but not the one linked with the account, makes the accuracy rate greater

### DIFF
--- a/cgi/reset_password.pl
+++ b/cgi/reset_password.pl
@@ -75,7 +75,7 @@ if ($action eq 'process') {
 			my $emails_ref = retrieve("$data_root/users/users_emails.sto");
 			if (not defined $emails_ref->{$id}) {
 				# not found, try with lower case email
-				$id = lc $id;
+				lc $id = lc $id;
 			}
 			if (not defined $emails_ref->{$id}) {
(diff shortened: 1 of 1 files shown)


cgi/reset_password.pl
```diff
@@ -75,7 +75,7 @@
 			my $emails_ref = retrieve("$data_root/users/users_emails.sto");
 			if (not defined $emails_ref->{$id}) {
 				# not found, try with lower case email
-				$id = lc $id;
+				lc $id = lc $id;
```
stephanegigandet, line 78: We have to process once all emails stored in users_emails.sto to lowercase them, we can't do that here.

PR #6227, opened 2021-12-29
fix: fix daily tasks

alexgarel:
We had a problem of disk space on preprod.

This is linked to daily failing, and as we did not remove previous archive before using wget, it was uploading archive with a new name, saturating disk space…

Task are failing because of a timeout which was, it seems, too short.


teolemon (approved):
LGTM

stephanegigandet (approved):

Issue #5517, opened 2021-08-21
Nutri-Score (tracker)


teolemon:
### What

- [ ] #5478
- [ ] #5369
- [ ] #4835
- [ ] #3585
- [ ] #3157
- [ ] #2776
- [ ] #2408
- [ ] #2176
- [ ] #2171
- [ ] #1860
- [ ] #1792
- [ ] #1791
- [ ] #1214

### Part of
- Scores

adityav27:
I've submitted a PR (#13367) to resolve #8353, which addresses the Nutri-Score category coherence mentioned in this tracker.

PR #6237, opened 2022-01-03
Fix Nutri-Score panel, add reason when not computed

stephanegigandet:
The Nutri-Score panel currently has a broken image when the Nutri-Score is unknown.

This fixes it, and also displays in the panel subtitle the reason why the Nutri-Score is not computed: missing category, not applicable category, or missing nutrition data.

Also moved the Nutri-Score warnings for fiber and fruits inside the Nutri-Score panel.

### Part of
- #5517


stephanegigandet:
image

alexgarel (approved):
LGTM + tested ok

PR #5806, opened 2021-10-02
Update packaging_materials.txt

dnim:
according to https://en.wikipedia.org/wiki/Recycling_codes

**Checklist:**

Make sure you've done all the following (You can delete the checklist before submitting)

- [x] Code is well documented
- [ ] Include unit tests for new functionality
- [ ] Code passes GitHub workflow checks in your branch
- [ ] If you have multiple commits please combine them into one commit by squashing them.
- [ ] Read and understood the [contribution guidelines](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/CONTRIBUTING.md)

**Description:**

Describe the changes made and why they were made instead of how they were made.

**Related issues and discussion:** #[ISSUE NUMBER]

PR #6239, opened 2022-01-04
Clean packaging materials taxonomy, fix glass recycling codes

stephanegigandet:
Now that we have xx: entries that work for all languages, we can move all recycling codes to xx: , and remove them from the other languages. Also added some notes in the taxonomy about that.

I also fixed the glass codes, there was a mismatch between some codes and what seems to be the official list in the EU directive. I added the link in the taxonomy.

This also incorporates the change from https://github.com/openfoodfacts/openfoodfacts-server/pull/5806


alexgarel (approved):
LGTM

Issue #6242, opened 2022-01-04
Parse mentions for specific ingredients like "Prepared with 20% of fruits", "Total milk content: 23%" at the end of the ingredients list


stephanegigandet:
### What

Some products specify the % of specific ingredients at the end of the ingredient list.

e.g. https://uk.openfoodfacts.org/product/5000354912696/birds-brandy-sauce
"Ingredients list:
Skimmed Milk, Buttermilk, Sugar, Modified Maize Starch, Brandy (1.5%), Palm Oil, Whey (Milk), Flavouring. Total Milk Content 73%. "

https://fr.openfoodfacts.org/produit/8410409010142/alvalle-gazpacho-tradicional
"Tomate,concombre,poivron,oignon,eau,huile d'olive vierge extra (1,1%),vinaigre de vin,pain de riz,sel,ail,jus de citron,teneur en légumes: 89%"

Those strings are currently unrecognized and prevent accurate ingredients analysis. They also contain valuable info about the fruits or milk content which are factors for the Nutri-Score computation.

### Describe the solution you'd like

Parse the strings, record their content in a new specific_ingredients structure that can be used for the Nutri-Score algorithm.

### Part of
- #5517

stephanegigandet:
Deployed.

PR #6243, opened 2022-01-04
feat: Initial support for specific ingredients parsing

stephanegigandet:
- See #6242 

This is in particular to resolve Nutri-Score computation errors when we fail to parse the % of fruits/vegetables or milk content indicated at the end of the ingredients list.

The resulting specific_ingredients structure is now used when computing the Nutri-Score.

It is also potentially useful to later parse other types of mentions like "Origin of something: some country" at the end of ingredients list.

### Part of
- #5517


lib/ProductOpener/Ingredients.pm
```diff
@@ -4298,6 +4396,20 @@ sub preparse_ingredients_text($$) {
 	# turn & to and
 	$text =~ s/ \& /$and/g;
 
+	# number + gr / grams -> g 
+	$text =~ s/(\d\s*)(gr|gram|grams)\b/$1g/ig;
+	if ($product_lc eq 'fr') {
+		$text =~ s/(\d\s*)(gramme|grammes)\b/$1g/ig;
+	}
```
alexgarel, line 4403: would'nt it be safer to do this the other way around ?

I mean replacing `g` / `gr` with `grammes`

lib/ProductOpener/Ingredients.pm
```diff
@@ -852,17 +852,120 @@ my %min_regexp = (
...
+		}
+
+		# If we found an ingredient, save it in specific_ingredients
+		if (defined $ingredient) {
+			my $ingredient_id = canonicalize_taxonomy_tag($product_lc, "ingredients", $ingredient);
+
+			# We might have an ingredient specified multiple times (e.g. once for percent, another for origins or labels)
+			defined $product_ref->{specific_ingredients}{$ingredient_id} or $product_ref->{specific_ingredients}{$ingredient_id} = {};
+			$product_ref->{specific_ingredients}{$ingredient_id}{ingredient} = $ingredient;
+			$product_ref->{specific_ingredients}{$ingredient_id}{text} = $matched_text;
```
alexgarel, line 954: Is it ok to smash previous values ?

lib/ProductOpener/Ingredients.pm
```diff
@@ -852,17 +852,120 @@ my %min_regexp = (
...
+			}
+		}
+
+		# If we found an ingredient, save it in specific_ingredients
+		if (defined $ingredient) {
+			my $ingredient_id = canonicalize_taxonomy_tag($product_lc, "ingredients", $ingredient);
+
+			# We might have an ingredient specified multiple times (e.g. once for percent, another for origins or labels)
+			defined $product_ref->{specific_ingredients}{$ingredient_id} or $product_ref->{specific_ingredients}{$ingredient_id} = {};
+			$product_ref->{specific_ingredients}{$ingredient_id}{ingredient} = $ingredient;
```
alexgarel, line 953: On this kind of code blocs, where $product_ref->{specific_ingredients}{$ingredient_id} is repetead a lot there after, I would find it more readable to have:

```perl
my %ingredient_entry = $product_ref->{specific_ingredients}{$ingredient_id};
```
and use it in following lines.

t/ingredients_parsing.t
```diff
@@ -168,10 +168,10 @@ my @lists =(
...
 
-	["fr","Céleri - rave 21% - Eau, légumes 33,6% (carottes, céleri - rave, poivrons rouges 5,8% - haricots - petits pois bio - haricots verts - courge - radis, pommes de terre - patates - fenouil - cerfeuil tubéreux - persil plat)","Céleri-rave 21% - Eau, légumes 33,6% (carottes, céleri-rave, poivrons rouges 5,8% - haricots - petits pois bio - haricots verts - courge - radis, pommes de terre - patates - fenouil - cerfeuil tubéreux - persil plat)"],
+	["fr","Céleri - rave 21% - Eau, légumes 33,6% (carottes, céleri - rave, poivrons rouges 5,8% - haricots - petits pois bio - haricots verts - courge - radis, pommes de terre - patates - fenouil - cerfeuil tubéreux - persil plat)","Céleri-rave 21% - Eau, légumes 33.6% (carottes, céleri-rave, poivrons rouges 5.8% - haricots - petits pois bio - haricots verts - courge - radis, pommes de terre - patates - fenouil - cerfeuil tubéreux - persil plat)"],
```
alexgarel, line 171: wow, that would be clearer with a line between entry / expected ! 
I did not understand you where changing the expected result at first sight.

t/nutriscore.t
```diff
@@ -3,6 +3,8 @@
 use strict;
 use warnings;
 
+use utf8;
+
```
alexgarel, line 7: why don't we have the use `Modern::Perl '2017';` 

shan't it be in every file ?

alexgarel (commented):
LGTM, still I need to test.

lib/ProductOpener/Ingredients.pm
```diff
@@ -852,17 +852,120 @@ my %min_regexp = (
...
 my %ignore_strings_after_percent = (
-	en => "of (the )?(?:total weight|grain is wholegrain rye)",
-	es => "(en el chocolate( con leche)?)",
+	en => "of (?:the )?(?:total weight|grain is wholegrain rye)",
+	es => "(?:en el chocolate(?: con leche)?)",
 	fi => "jauhojen määrästä",
-	fr => "(dans le chocolat( (blanc|noir|au lait))?)|(du poids total|du poids)",
+	fr => "(?:dans le chocolat(?: (?:blanc|noir|au lait))?)|(?:du poids total|du poids)",
 	sv => "fetthalt",
 );
```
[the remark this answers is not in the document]
stephanegigandet: Taxonomies are great for words and simple phrases, but not for regular expressions that can easily break or have unintended consequences. That's why I think it's best to keep the values close to the code, so that the effect of changing something is clear.

lib/ProductOpener/Ingredients.pm
stephanegigandet: If we replace g by grammes, then we have to make a lot of code that analyze the resulting text language specific, which increases the complexity of the parsing (which is already quite complex). That's why I think it's best to normalize to "g".

lib/ProductOpener/Ingredients.pm
stephanegigandet: I guess we could concatenate them. I didn't do it because the case is largely hypothetical and the smashed values ("ingredient" and "text") are kept mostly for debugging purposes. But it's cleaner if we concatenate, I'll change it.

t/ingredients_parsing.t
stephanegigandet: Agreed, I'll change it.

t/nutriscore.t
stephanegigandet: I'll add it.

lib/ProductOpener/Ingredients.pm
stephanegigandet: I'll do that, but I have to keep the line 			defined $product_ref->{specific_ingredients}{$ingredient_id} or $product_ref->{specific_ingredients}{$ingredient_id} = {};
before, as otherwise I will get an undef reference if it does not exist yet.

alexgarel: yes, of course, that was what I intended ;-)

alexgarel:
The sentences parsed with the new function do not appear in *Details of the analysis of the ingredients*, so it's not easy to test manually, and I also think it's a pity as it participates to nutri-score.

stephanegigandet:
Applied changes from code review, thank you!

Also added initial support to parse things like "Origin of the milk: UK".

stephanegigandet:
> The sentences parsed with the new function do not appear in _Details of the analysis of the ingredients_, so it's not easy to test manually, and I also think it's a pity as it participates to nutri-score.

I added a sub section to list the specific ingredients in the analysis.

alexgarel (approved):
Cool !

PR #6245, opened 2022-01-04
feat: integration test on import_csv

alexgarel:
two main commits

* adding some dev tooling in docker (I still need to document usage in vscode)
* adding an integration test on import_csv_file


lib/ProductOpener/Import.pm
```diff
@@ -1921,14 +1935,7 @@ sub import_csv_file($) {
...
 
-									require LWP::UserAgent;
-
-									my $ua = LWP::UserAgent->new(timeout => 10);
-									
-									# Some platforms such as CloudFlare block the default LWP user agent.
-									$ua->agent(lang('site_name') . " (https://$server_domain)");
-
-									my $response = $ua->get($image_url);
+									my $response = download_image($image_url);
```
stephanegigandet, line 1938: Good idea to make a function, it makes things more readable, in addition to being needed for the mock.

lib/ProductOpener/Import.pm
```diff
@@ -2185,9 +2192,9 @@ sub update_export_status_for_csv_file($) {
 
 	my $args_ref = shift;
 
-	$User_id = $args_ref->{user_id};
-	$Org_id = $args_ref->{org_id};
-	$Owner_id = get_owner_id($User_id, $Org_id, $args_ref->{owner_id});
+	my $User_id = $args_ref->{user_id};
```
stephanegigandet, line 2195: This may have unwanted side effects.
It's unfortunate, but I put some global variables $User_id, $Owner_id and $Org_id in Display.pm, and some functions use them directly.
So if you make them local, there might be some functions called that uses the global values.

The best thing would be to put those variables in a request structure that we always pass, but that would require changes all over.

t/import_csv_file.t
```diff
@@ -0,0 +1,213 @@
...
+    my $cursor = execute_query(sub {
+		return get_products_collection()->query({})->sort({code =>1});
+	});
+    my @products = ();
+    while ( my $doc = $cursor->next ) {
+        push(@products, $doc);
+    }
+    # expected values
+    ## no critic (ProhibitStringyEval)
+    my $expected_products = eval(path($expected_products_path)->slurp_raw());
```
stephanegigandet, line 171: This works, but I think it would be easier to use the same mechanism that we have for the other tests: we store products as JSON files. So here we would just read all JSON files (one per product) in the directory.

The advantage is that when the expected output change, it's much easier to save the new output as JSON files, and to see diffs. We can also easily see how many products and which products have been affected just by looking at the number of JSON product files that have changed in git.

It's important to note that the expected results will change a lot over time (even if the input CSV does not change) because we keep adding new features like the Nutri-Score, Eco-Score, ingredients parsing etc. that change the resulting product data once it is imported.

So we will need a way to very easily update tests and see diffs.

For instance when I change something in the ingredients parsing, I do this:

perl ingredients.t results expected_tests_results/ingredients
-> this saves the new output product as JSON files over the expected outputs
-> then I can just do "git diff t" and I will see the corresponding changes, to check if they are expected or not.

t/import_csv_file.t
```diff
@@ -0,0 +1,213 @@
...
+    ## use critic
+    # same number of products
+    is(scalar @products, scalar @$expected_products, , "As much expected produts as in db");
+    my $items_iter = each_array(@products, @$expected_products);
+    while ( my ($product, $expected_product) = $items_iter->() )
+    {
+        # load from sto
+        my $product_obj = retrieve_product($product->{code});
+        # values are those expected
+        my $cmp_product = deep_extract_keys_from($product, $expected_product);
```
stephanegigandet, line 181: Why do we need to filter the keys? Can't we just compare the product to the expected product directly? Is it because there are a few fields that change over time? (e.g. some timestamps?)

t/inputs/import_csv_file/test.converted.csv
```diff
@@ -0,0 +1,2 @@
+code	producer_product_id	producer_version_id	lc	product_name_fr	abbreviated_product_name_fr	generic_name_fr	.2	.3	.4	.5	packaging	packaging_text_fr	brands	brand_owner	categories	labels	countries	stores	obsolete	obsolete_since_date	periods_after_opening	origins	manufacturing_places	producer_fr	emb_codes	ingredients_text_fr	allergens	traces	energy-kj_100g_value_in_kj	energy-kcal_100g_value_in_kcal	.6	.7	.8	.9	.10	.11	.12	.13	.14	nutriscore_score_producer	nutriscore_grade_producer	nova_group_producer	conservation_conditions_fr	warning_fr	preparation_fr	recipe_idea_fr	recycling_instructions_to_recycle_fr	recycling_instructions_to_discard_fr	customer_service_fr	link	image_front_url_fr	image_ingredients_url_fr	image_nutrition_url_fr	image_packaging_url_fr	image_other_url
```
stephanegigandet, line 1: Aren't you creating this file in the test? (as well as test.csv)
If so then we should not need them in git as they won't be used.

lib/ProductOpener/Import.pm
alexgarel: Hum, I see. I will revert the change document and add a TODO.

t/inputs/import_csv_file/test.converted.csv
alexgarel: My bad, they should not have been commited (and I though I did not !)

t/import_csv_file.t
alexgarel: This is because
* there are some fields like timestamps and others that may change
* I don't want the test to break too often because of some new computation, or change in computation, etc. So I just verify the output of the code I test (not of other parts). (Of course you may also argue that integration test should look at unexpected changes, but I think maintainability counts also).

t/import_csv_file.t
alexgarel: Initially I go for perl because, I wanted to be able to use ignore / bag / supersetof etc from Test::Deep (I didn't know well what to expect).

I think your methodology is also fine, so I can change and go for JSON.
I will however have to clean some fields before comparison, but that can be done in the test part.

t/import_csv_file.t
stephanegigandet: ok, fine with me.

lib/ProductOpener/Import.pm
alexgarel: Could we at least declare them "local" ?

t/import_csv_file.t
```diff
@@ -0,0 +1,193 @@
...
+            close $image;
+            return $content;
+        },
+    );
+    return $response;
+}
+
+
+# fields we don't want to check for they vary from test to test
+my @fields_ignore_content = qw(last_modified_t created_t owner_fields sources.0.import_t);
```
stephanegigandet, line 68:
```suggestion
my @fields_ignore_content = qw(last_modified_t created_t owner_fields sources.0.import_t entry_dates_tags last_edit_dates_tags);
```

Otherwise the tests won't work the next day:

```
    "entry_dates_tags" : [
-      "2022-04-04",
+      "2022-04-06",
       "2022-04",
       "2022"
    ],
@@ -364,7 +365,7 @@
       "en:1"
    ],
    "last_edit_dates_tags" : [
-      "2022-04-04",
+      "2022-04-06",
       "2022-04",
       "2022"
    ],

```

t/import_csv_file.t
```diff
@@ -0,0 +1,193 @@
...
+        "org_id" => "test-org",
+        "owner_id" => "org-test-org",
+        "csv_file" => $converted_file,
+        "exported_t" => $datestring,
+    };
+    # run
+    ($out, $err) = capture_ouputs (sub {
+        ProductOpener::Import::import_csv_file($args);
+    });
+    # get all products in db, sorted by code for perdictability
```
stephanegigandet, line 152:
```suggestion
    # get all products in db, sorted by code for predictability
```

t/import_csv_file.t
```diff
@@ -0,0 +1,193 @@
...
+}
+
+
+# fields we don't want to check for they vary from test to test
+my @fields_ignore_content = qw(last_modified_t created_t owner_fields sources.0.import_t);
+# fields that are array and need to sort to have predictable results
+my @fields_sort = qw(_keywords);
+
+
+# clean products fields that we can't check because they change over runs
```
stephanegigandet, line 73: It could be useful to move that function to the Test.pm module, for future integration tests (e.g. if we want to test the API or the web form to edit a product).

t/import_csv_file.t
```diff
@@ -0,0 +1,193 @@
...
+# we may still add some test on those fields here
+sub clean_products_fields($) {
```
stephanegigandet, line 75: Could we rename the function to something more specific? We already have some functions called clean_fields() for a different type of cleaning (e.g. remove extra spaces).

Maybe assign_unchanging_value_to_date_dependent_fields() ?

t/import_csv_file.t
```diff
@@ -0,0 +1,193 @@
...
+            @key = split(/\./, $field_s);
+            if (!deep_exists($product, @key)) {
+                push(@missing_fields, ($code, $field_s));
+            } else {
+                my @sorted = sort @{deep_get($product, @key)};
+                deep_set($product, @key, \@sorted);
+            }
+        }
+    }
+    if (@missing_fields) {
```
stephanegigandet, line 101: I'm not sure this is useful. If we are missing a field that is expected, we will see it in the diff later anyway.

Removing this check would allow the function to be used in more cases, and not having each test to specify which fields to remove.

t/import_csv_file.t
alexgarel: Done, I also made it more generic regarding going down arrays.

t/import_csv_file.t
alexgarel: Done.

lib/ProductOpener/Test.pm
```diff
@@ -159,20 +266,158 @@ sub compare_csv_file_to_expected_results($$$) {
...
+Array of products
+
+=cut
+
+# clean products fields that we can't check because they change over runs
+# we may still add some test on those fields here
+sub normalize_products_for_test_comparison($) {
+    my $array_ref = shift;
+
+    # fields we don't want to check for they vary from test to test
```
stephanegigandet, line 409: Those are now defined in normalize_product_for_test_comparison((), they can be removed from here

stephanegigandet (approved):
Thank you!

PR #6254, opened 2022-01-07
fix: ensure windows newlines don't break panels

stephanegigandet:
some producers have imported ingredients lists with \r\n sequences, and that broke the JSON used by knowledge panels.

this fixes the generation of the panels.

we probably need another fix to normalize newlines when we edit products / import data.

### DIFF
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -179,7 +179,8 @@ The function converts the multiline string into a single line string.
 
 sub convert_multiline_string_to_singleline($) {
     my $line = shift;
-    $line =~ s/\n/\\n/sg;
+    # \R will match all Unicode newline sequence
+    $line =~ s/\R/\\n/sg;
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -179,7 +179,8 @@ The function converts the multiline string into a single line string.
 
 sub convert_multiline_string_to_singleline($) {
     my $line = shift;
-    $line =~ s/\n/\\n/sg;
+    # \R will match all Unicode newline sequence
+    $line =~ s/\R/\\n/sg;
```
alexgarel, line 183: wow this \R character set is cool :-)

alexgarel (approved):
LGTM

lib/ProductOpener/KnowledgePanels.pm
stephanegigandet: just discovered it ;) We probably could look for instances of \n and replace them.

PR #6267, opened 2022-01-07
feat(producers): add link to admin manual on session sucess

alexgarel:
Very simple, just to find easily the documentation :-)


po/common/en.po
```diff
@@ -279,6 +279,10 @@ msgctxt "you_can_also_help_us"
 msgid "You can also help to fund the Open Food Facts project"
 msgstr "You can also help to fund the Open Food Facts project"
 
+msgctxt "producers_administration_manual"
```
stephanegigandet, line 282: Crowdin currently uses /po/common/common.pot as the source, so we need to copy this English entry in common.pot too.

lib/ProductOpener/Users.pm
```diff
@@ -197,6 +198,19 @@ sub delete_user($) {
 	find_and_replace_user_id_in_products($userid, $new_userid);
 }
 
+=head2 is_admin_user($user_id)
```
stephanegigandet, line 201: good idea!

cgi/session.pl
```diff
@@ -49,6 +49,7 @@
 if (defined $User_id) {
 
 	$template_data_ref->{user_name} =  $User{name};
+	$template_data_ref->{is_admin} = is_admin_user($User_id);
```
stephanegigandet, line 52: In fact there's already an "admin" field that is set automatically for all templates:

in Display.pm:

```
sub process_template($$$) {

	my $template_filename = shift;
	my $template_data_ref = shift;
	my $result_content_ref = shift;

	# Add functions and values that are passed to all templates

[..]
	$template_data_ref->{admin} = $admin;
```

VaiTon: https://github.com/openfoodfacts/openfoodfacts-server/blob/686f4fdf5d2b26209581c91050c7c7fc08845814/lib/ProductOpener/Display.pm#L380

alexgarel: ok I removed this and use admin in template now.

po/common/en.po
alexgarel: Done.

stephanegigandet (approved):
Looks good, thank you!

PR #6270, opened 2022-01-10
feat: start of additives panels

stephanegigandet:
Functional:
- list additives after the list of ingredients, as a list of small panels
- clicking on the additive name opens the additive panel
- additive panel currently consists only of the Wikipedia abstract and link (which are of very different levels of quality and relevance to food products depending on additive and language)
- to be added: a list of the additive functions (colours, raising agent etc.)
- to be added: EFSA table for the risk of overexposure (+ corresponding colors/icons in the list of additives)

Knowledge panel API changes:
- added a size="small" property to panels, so that the list of additives doesn't take a huge amount of vertical space
- added source_text / source_url / source_language to panel text elements

image

### DIFF
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -946,7 +946,33 @@ sub create_additives_panel($$$) {
 }
 
 
-sub add_taxonomy_properties_in_target_languages_to_object () {
+=head2 add_taxonomy_properties_in_target_languages_to_object ( $object_ref, $tagtype, $tagid, $properties_ref, $target_lcs_ref )
+
+This function adds to the hash ref $object_ref (for instance a data structure passed to a template) the values
+of selected properties, if they exist in one of the target languages.
+
+For instance for the panel for an additive, we can include a Wikipedia abstract in the requested language if available,
+or in English if not.
+
+=head3 Arguments
+
+=head4 object reference $object_ref
+
+=head4 taxonomy $tagtype
+
+=head4 tag id $tagoid
+
+=head4 list of properties $properties_ref
+
+Properties to add to the resulting object.
+
+=head4 language codes $target_lcs
+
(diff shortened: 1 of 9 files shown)


lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -885,4 +887,116 @@ sub create_ingredients_panel($$$) {
...
+    my $tagtype = shift;
+    my $tagid = shift;
+
+    my $properties_ref = shift;
+    my $target_lcs_ref = shift;
+
+    foreach my $property (@$properties_ref) {
+        my $property_value;
+        my $property_lc;
+        foreach my $target_lc (@$target_lcs_ref) {
```
alexgarel, line 987:
```suggestion
        # get property value for first language for which it is defined
        foreach my $target_lc (@$target_lcs_ref) {
```

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -885,4 +887,116 @@ sub create_ingredients_panel($$$) {
...
+=cut
+
+sub add_taxonomy_properties_in_target_languages_to_object ($$$$$) {
+
+    my $object_ref = shift;
+    my $tagtype = shift;
+    my $tagid = shift;
+
+    my $properties_ref = shift;
+    my $target_lcs_ref = shift;
```
alexgarel, line 982: When there is more than 3 or 4 arguments, shouldn't we use hash to pass arguments ?

I feel like this could help make the code a bit more readable (where we call such functions).

alexgarel (approved):
LGTM

lib/ProductOpener/KnowledgePanels.pm
stephanegigandet: If we use a hash, we lose the check on the number of arguments that Perl does at compile time. There isn't a very good way to have signatures in Perl unfortunately (there's an experimental "use experimental 'signatures';" and some modules).

PR #6272, opened 2022-01-11
Recommendation panels (global + country specific) for health: alcohol and pulses

stephanegigandet:
This is a first implementation of panels to make recommendations based on products (e.g. categories, ingredients, nutrients) and users (based on country, language, and maybe later on personal preferences).

The first recommendations are examples, to see how it looks on the web and in the app, experiment with ways to define them etc.
Those first recommendations are related to health and shown in the health card, below the Nutri-Score panel, but we could also have other kinds of recommendations, for instance related to the environment.

The first recommendation is for alcoholic beverages. By default, we show the WHO recommendation, and for users in France with the French language, we show instead the recommendation from the French health agency.

image

in France:

image

The second one is to suggest eating more pulses, and this one is specific to France.

image

Note: there should be an extra link in the panel to indicate the source of the recommendation, but the implementation is in another PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/6270


templates/api/knowledge-panels/health/ingredients/additive.tt.json
```diff
@@ -0,0 +1,24 @@
...
+                "html": `<strong>[% panel.wikipedia_title %][% sep %]:</strong> [% panel.wikipedia_abstract %]`,
+                "source_text": `Wikipedia`,
+                "source_url": `[% panel.wikipedia_url %]`,
+                "source_lc": "[% panel.wikipedia_url_lc %]",
+                "source_language": "[% panel.wikipedia_url_language %]"
+            }
+        },
+        [% END %]
+    ]
+}
```
alexgarel, line 24: C'est normal qu'il soit là lui ?

templates/web/panels/panel.tt.html
```diff
@@ -149,6 +149,17 @@ <h4 class="panel_title">
                     <img src="[% image_element.url %]" alt="[% image_element.alt %]">
                 [% IF image_element.link_url.defined %]</a>[% END %]
 
+                [% IF image_element.source_url.defined %]
+                <div class="image_source" style="font-style:italic">[% lang("source") %][% sep %]:
+                    <a href="[% text_element.source_url %]">[% image_element.source_text %]
```
alexgarel, line 154: isn't it `href="[% image_element.source_url %]"` ? or is it the IF above ill defined ? I'm a bit confused…

alexgarel (approved):
LGTM modulo a comment on a potential bug

templates/web/panels/panel.tt.html
stephanegigandet: ah right, my mistake, thanks!

PR #6279, opened 2022-01-12
ops: fixing daily task

alexgarel:
- have coherent timeouts
- no git checkout as this is the job of mongo-deploy

Currently daily was failing because main branch on the preprod env had diverged !
This is not the job of daily to deploy, there is another automation for that.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -45,13 +45,10 @@ jobs:
         proxy_host: ${{ secrets.PROXY_HOST }}
         proxy_username: ${{ secrets.USERNAME }}
         proxy_key: ${{ secrets.SSH_PRIVATE_KEY }}
-        command_timeout: 30m
+        command_timeout: 180m
         script_stop: false
         script: |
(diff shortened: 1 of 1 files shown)


.github/workflows/daily.yml
```diff
@@ -45,13 +45,10 @@ jobs:
...
         proxy_username: ${{ secrets.USERNAME }}
         proxy_key: ${{ secrets.SSH_PRIVATE_KEY }}
-        command_timeout: 30m
+        command_timeout: 180m
         script_stop: false
         script: |
           cd ${{ matrix.env }}
-          cp .env ../.env
-          git fetch --depth 1 && git checkout -qf ${{ startsWith(github.ref, 'refs/heads/cron-') && github.sha || 'main' }}
-          cp ../.env .env
```
alexgarel, line 54: FYI I imagine this was done before automating deployement. But now we have deployment, we do not need this.

stephanegigandet (approved):

PR #6281, opened 2022-01-12
Use Smoothie colors and styles on the web for attributes and panels

stephanegigandet:
This is to use the colors and styles defined in https://docs.google.com/presentation/d/1fVVAo888BM-rpOAdQvjFhrmWe7NeAdCO2ZRxMzkHEvg/edit?pli=1#slide=id.g10a8b8e61b0_0_274 on the Web as well.

- changed border radius from 12px to 8px
- changed colors for A to E grades and evaluation (good, average, bad) : background color + icon color + text color

Before:

image

After:

image

### DIFF
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -547,7 +547,7 @@ html[dir="rtl"] {
 // List of products with cards (e.g. search results, facets)
 
 .list_product_a {
-  border-radius:12px;
+  border-radius:8px;
   padding:10px;
   display:block;
   background-color:#eee;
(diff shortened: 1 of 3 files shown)


scss/_off.scss
```diff
@@ -721,52 +731,96 @@ html[dir="rtl"] {
...
 }
 .accordion .accordion-navigation>a:hover, .accordion dd>a:hover {
   background:#f0f0f0;
+  text-decoration:none;
 }
 
 // background colors for attributes and panels
+
+.evaluation_good_title {
+  color:rgba(33,150,83,1);
```
alexgarel, line 748: As we use scss why don't we use variables for those things ?

alexgarel (approved):
LGTM

scss/_off.scss
stephanegigandet: Just because I'm not used to it. Changed to scss variables.

teolemon:
scss/_off.scss
 761:9  ✖  Missed semicolon   CssSyntaxError

PR #6285, opened 2022-01-12
Cereals

moon-rabbitOFF:


taxonomies/categories.txt
```diff
@@ -46741,7 +46745,7 @@ fr:Céréales aux fruits
 en:Cereal flakes with fruits
 fr:Pétales de céréales aux fruits
 
-<en:Cereal flakes
+<en:Extruded cereals
```
stephanegigandet, line 46748: What is the reason for this change? Shouldn't cereal flakes with nuts be under cereal flakes?

moon-rabbitOFF: The category 'cereal flakes' contains products containing cereal flakes only, with no addition of fruits, chocolate, nuts. If we begin to add all products containing cereal flakes (mueslis, granola, porridge...) in this category, most of the cereals would be there and I don't think such a category would make sense.

stephanegigandet: Sounds good, thanks for the explanation.

Issue #6288, opened 2022-01-13
Typo in function name not caught by tests, only at runtime: Undefined subroutine &ProductOpener::Food::as_tag called at /srv/off/lib/ProductOpener/Food.pm line 1582


stephanegigandet:
### Describe the bug

We deployed a module with a typo in one of the function names, and it was not detected by the automated tests.

`Undefined subroutine &ProductOpener::Food::as_tag called at /srv/off/lib/ProductOpener/Food.pm line 1582
`
Corresponding line:

```
	if (not ((has_tag($product_ref, "categories", "en:spring-waters"))
		and not (has_tag($product_ref, "categories", "en:flavored-waters") or **as_tag**($product_ref, "categories", "en:flavoured-waters")))) {
```

It resulted in the Nutri-Score computation to crash for spring waters.

### To Reproduce

Edit a product with the en:spring-waters category.

### Expected behavior

No crash.

More importantly: tests should fail.

### Type of device

Other

stephanegigandet:
Unfortunately, this does not seem to work:

```
/srv/off/scripts$ perl -MO=Lint,undefined-subs ../lib/ProductOpener/Food.pm 
Subroutine normalize_nutriment_value_and_modifier redefined at ../lib/ProductOpener/Food.pm line 178.
Subroutine default_unit_for_nid redefined at ../lib/ProductOpener/Food.pm line 233.
Subroutine assign_nid_modifier_value_and_unit redefined at ../lib/ProductOpener/Food.pm line 262.
Subroutine unit_to_kcal redefined at ../lib/ProductOpener/Food.pm line 335.
Subroutine kcal_to_unit redefined at ../lib/ProductOpener/Food.pm line 348.
Subroutine unit_to_g redefined at ../lib/ProductOpener/Food.pm line 369.
Subroutine g_to_unit redefined at ../lib/ProductOpener/Food.pm line 415.
Subroutine unit_to_mmoll redefined at ../lib/ProductOpener/Food.pm line 453.
Subroutine mmoll_to_unit redefined at ../lib/ProductOpener/Food.pm line 476.
Subroutine canonicalize_nutriment redefined at ../lib/ProductOpener/Food.pm line 1181.
Subroutine normalize_quantity redefined at ../lib/ProductOpener/Food.pm line 1224.
Subroutine normalize_serving_size redefined at ../lib/ProductOpener/Food.pm line 1262.
[...]

Issue #6287, opened 2022-01-13
Import of some products from pro platform seem to fail


stephanegigandet:
### Describe the bug

for org-nestle-waters, this product (and maybe others) are not imported to the public platform:

https://fr.pro.openfoodfacts.org/produit/8002270446764/sanpellegrino-momenti-orange-sanguine-fleur-d-oranger-6x33cl

From the minion_log4perl log, it seems the job dies after:

[58137] ProductOpener/Ingredients.pm 5689 ProductOpener.Ingredients {} detect_allergens_from_text - done
[58137] ProductOpener/Import.pm 1720 ProductOpener.Import {} fix_salt_equivalent {code => "7613036249928",product_id => "7613036249928"}
[58137] ProductOpener/Import.pm 1723 ProductOpener.Import {} compute_serving_size_data {code => "7613036249928",product_id => "7613036249928"}
[58138] ProductOpener/Import.pm 1726 ProductOpener.Import {} compute_nutrition_score {code => "7613036249928",product_id => "7613036249928"}

### To Reproduce

Try to export the product, watch logs in /srv/off/logs/

### Expected behavior

Product is imported.

### Type of device

Other

PR #6291, opened 2022-01-13
fix: typo in function name in Food.pm 

stephanegigandet:
Related issues #6288 #6287

Also add some tests.

But it does not solve the more general issue that typos in function names are currently undetected.


teolemon (commented):
It seems more complicated, but I have found this, beyond another page which says it's impossible.
https://stackoverflow.com/questions/8499608/how-do-i-get-perl-c-to-throw-undefined-or-undeclared-function-errors?rq=1

alexgarel (approved):
Good !

PR #6292, opened 2022-01-13
add more automatic Equadis imports to public database

stephanegigandet:
This is to add more producers from Equadis to the daily exports from the producers platform to the public database.

We'll probably need to do this differently at some point. We have a checkbox in the org settings for daily exports, but it currently has no effect.


teolemon (approved):

alexgarel (approved):
LGTM

Issue #6144, opened 2021-12-02
Docker apache should use npm_prefork


alexgarel:
### Describe the bug

we currently have memory problems in preprod, this may be linked to bad config of apache

### To Reproduce

See graph on monitoring

### Expected behavior

We use default conf for npm (npm event) in apache docker. We should use the npm_prefork. 

This may be the source of memory problems we encounter in preprod.

### Type of device

Other

PR #6293, opened 2022-01-14
ops: make apache use prefork as in prod

alexgarel:
see  #6144

I tested it locally also with some load.


stephanegigandet (approved):
Let's try that.
For dev, it will be interesting to see if we can use less memory with fewer servers.

PR #6303, opened 2022-01-14
fix: Remove unary operators

teolemon:
### Why
- Because the unary ++ and -- operators are subject to automatic semicolon insertion, differences in whitespace can change semantics of source code.

### DIFF
--- a/html/js/folksonomy.js
+++ b/html/js/folksonomy.js
@@ -374,7 +374,7 @@ function displayProductsWithProperty(_property, _value) {
             '<th class="values">Values</th>' +
             '</tr>' +
             '<tbody id="free_prop_body">';
-        while (index-=) {
+        while (index-=1) {
             content += ('<tr>' +
                         '<td class="product_code">' +
                         '<a href="/product/'+ data[index].product + '">' + data[index].product + '</a>' +
--- a/html/js/folksonomy.js
+++ b/html/js/folksonomy.js
@@ -315,7 +315,7 @@ function displayFolksonomyPropertyValues() {
         let content = "";
         // Sort by property
         const d = data.sort(function(a,b){ return a.k <b.k ? 1 :-1; });
-        while (index-=) {
+        while (index-=1) {
             content += ('<tr>' +
                         '<td class="version" data-version="' + d[index].version + '"> </td>' +
                         '<td class="property"><a href="/property/' + d[index].k + '">'                      + d[index].k + '</a></td>' +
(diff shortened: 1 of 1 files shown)


html/js/folksonomy.js
```diff
@@ -315,7 +315,7 @@ function displayFolksonomyPropertyValues() {
         let content = "";
         // Sort by property
         const d = data.sort(function(a,b){ return a.k <b.k ? 1 :-1; });
-        while (index--) {
+        while (index-=) {
```
alexgarel, line 318:
```suggestion
        while (index-=1) {
```

html/js/folksonomy.js
```diff
@@ -374,7 +374,7 @@ function displayProductsWithProperty(_property, _value) {
             '<th class="values">Values</th>' +
             '</tr>' +
             '<tbody id="free_prop_body">';
-        while (index--) {
+        while (index-=) {
```
alexgarel, line 377:
```suggestion
        while (index-=1) {
```

html/js/folksonomy.js
```diff
@@ -687,7 +687,7 @@ function displayFolksonomyForm() {
         let index = data.length;
         let content = "";
         const d = data.sort(function(a,b){ return a.k <b.k ?1 :-1; });
-        while (index--) {
+        while (index-=) {
```
alexgarel, line 690:
```suggestion
        while (index-=1) {
```

alexgarel (changes requested):
I think you forgot the 1

VaiTon:
NPM test is failing because of:

```
 /home/runner/work/openfoodfacts-server/openfoodfacts-server/html/js/folksonomy.js
Error:   318:16  error    Expected a conditional expression and instead saw an assignment  no-cond-assign
Error:   377:16  error    Expected a conditional expression and instead saw an assignment  no-cond-assign
Error:   425:16  error    Expected a conditional expression and instead saw an assignment  no-cond-assign
Error:   690:16  error    Expected a conditional expression and instead saw an assignment  no-cond-assign
Warning:   991:13  warning  Unary operator '++' used                                         no-plusplus
```

alexgarel:
@VaiTon error is related to the PR (you see it in the *Files changed* tab).

Indeed the linter does not like this way of writing. So I think the `index--` was the right way of writing…

@teolemon I would suggest closing this.

teolemon:
ok

PR #6304, opened 2022-01-14
Create MongoDB indexes for dev environment in docker

stephanegigandet:
Indexes are needed if we load more than ~1000 products in the database.


alexgarel (approved):
Perfect.

PR #6305, opened 2022-01-14
add countries list to scanbot output data

stephanegigandet:
This is to add a field (list of scan countries) to the list of products created by scanbot, so that it can be easily grepped for a particular country.


alexgarel (approved):
LGTM (not tested)

PR #6311, opened 2022-01-17
fix: set language of fields during init

### LINKED ISSUE #6310: Producer platform columns selection for file import can sometimes be ignored
### Describe the bug

It's a rare thing, but if the stored columns selections are broken (as was the case with the org elle-vire) with some language specific fields not having a language set, then importing a file, and not changing a column can result in the column being ignored. (e.g. no language is sent, even though the interface displays a language selected). This result in a field like ingredients_text in the converted CSV file, instead of ingredients_text_fr

### To Reproduce

difficult to reproduce, but selecting a language field, then unselecting it may be the reason

### Expected behavior

All settings shown in column match form should be applied.

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Browser

### Browser version

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
Fixes JS to initialize column matches.

- fixes #6310

### DIFF
--- a/cgi/import_file_process.pl
+++ b/cgi/import_file_process.pl
@@ -110,7 +110,11 @@ foreach my $field (keys %$columns_fields_ref) {
 	delete $columns_fields_ref->{$field}{max};
 	delete $columns_fields_ref->{$field}{n};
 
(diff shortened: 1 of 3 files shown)


alexgarel (approved):
LGTM

PR #6315, opened 2022-01-18
ops: add marker to block daily update in preprod

alexgarel:
Adds a marker to daily to avoid refreshing the database in preprod.

We want this from time to time to be able to run long experiments.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -49,8 +49,15 @@ jobs:
         script_stop: false
         script: |
           cd ${{ matrix.env }}
-          make import_prod_data
-          make restart_db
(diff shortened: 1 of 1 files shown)


.github/workflows/daily.yml
```diff
@@ -75,5 +81,11 @@ jobs:
         script_stop: false
         script: |
           cd ${{ matrix.env }}
-          make refresh_product_tags
-          make restart_db
+          if [[ ! -e ./NO_DAILY ]]
+          then
+            make refresh_product_tags
```
stephanegigandet, line 86: Why not refresh the products_tags ? They are recomputed from the mongo products collection, not imported from prod.

PR #6319, opened 2022-01-20
feat: Fix and improve detection of apps (name and UUID) to populate data sources

### LINKED ISSUE #2195: UUID extraction from comment seems broken
UUID extraction from comment seems broken

### LINKED ISSUE #2592: file_id regex for UUIDs does not match description
While researching for #2563, I noticed that the regex https://github.com/openfoodfacts/openfoodfacts-server/blob/8cb6fd279409ebd09e72de5e2504b3e14e0b93dc/lib/ProductOpener/Store.pm#L107-L109 takes quite some time during execution (according to `Devel::NYTProf`). While trying to optimize it, it looked like the current regex actually does not match the Waistline example from the comment: https://regexr.com/4oo4c

### LINKED ISSUE #3849: UUIDs parsing is broken
UUIDs parsing is broken. As a result, we can't browse products by an anonymous user. 
https://world.openfoodfacts.org/product/9782213705996/servir-fayard
July 15, 2020 at 7:47:33 PM CEST - Data (Added: product_name, quantity, packaging, brands, categories, labels, origins, manufacturing_places, purchase_places, stores, nutrition_data_per, nutrition_data_prepared_per, traces, ingredients_text, ingredients_text_fr, product_name_fr) (Changed: lang) -- Nutriments (Added: energy-kcal, energy, fat, saturated-fat, carbohydrates, sugars, fiber, proteins, salt, sodium, alcohol, nutrition-score-fr) - (app)Official Android app 3.2.9 ( Added by 69eea1ba7ab171af9a9a62b91142d797 ) - view

### LINKED ISSUE #5195: Add hashed user id field to API, for apps that have an app user id
### What

- A few apps that use one OFF account for writes send us some kinds of hashed identifiers of users. 
- This is useful when there are problematic edits, as we can find the other edits of the same user. 
- Today we do not have a dedicated field for this, and instead we use the comments field, which is parsed in different ways depending on the app.
- [ ] We could add a hashed_user_id field in the API, to record it in the change history.

### Part of 
- #5538

stephanegigandet:
We try to identifiy product edits that are made through apps, so that we can populate the data_sources facet. We currently do it by using the comment field that can contain an app name, and in some cases a uuid (used so that we can group contributions by the same user, for instance to check for errors), and/or the specific OFF user used by some app. This has been broken for some time.

This PR:
- refactors the identification of apps through the existing specific app OFF user ids and comments, and fixes issues - fixes #2592 - fixes #3849 - fixes #2195
- adds tests so that it does not become broken again
- add more robust ways for apps to pass their name and user uuid (app_name, app_version and app_uuid) instead of using the comment field - fixes #5195
- add an alternate way to pass the user agent through a query parameter (useful for web frameworks where it's not possible to change the user-agent header), and record that user agent (currently it's savec in the nginx logs, but not in the changes, which makes it useless to identify apps that edit products)
- fixes the pages that show edits from a specific app + uuid

image

image

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -1093,6 +1093,8 @@ sub store_product($$$) {
 	push @{$changes_ref}, {
 		userid => $user_id,
 		ip => remote_addr(),
+		# Allow apps to send the user agent as a form parameter instead of a HTTP header, as some web based apps can't change the User-Agent header sent by the browser
+		user_agent => remove_tags_and_quote(decode utf8=>param("User-Agent")) || remove_tags_and_quote(decode utf8=>param("user-agent")) || user_agent(),
 		t => $product_ref->{last_modified_t},
 		comment => $comment,
 		rev => $rev,
@@ -1108,7 +1110,7 @@ sub store_product($$$) {
 
 	compute_product_history_and_completeness($new_data_root, $product_ref, $changes_ref, $blame_ref);
 
-	compute_data_sources($product_ref);
+	compute_data_sources($product_ref, $changes_ref);
 	
 	compute_main_countries($product_ref);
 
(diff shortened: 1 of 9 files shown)


teolemon (approved):

alexgarel (approved):
:+1:

alexgarel:
Did you update the documentation to explain the new parameters ?

I think you should update the wiki and the API documentation.

lib/ProductOpener/Products.pm
```diff
@@ -1088,14 +1089,34 @@ sub store_product($$$) {
...
+		$change_ref->{user_agent} = $user_agent;
+	}
+
+	# Allow apps to send app_name, app_version and app_uuid parameters
+	foreach my $field (qw(app_name app_version app_uuid)) {
+		my $value = remove_tags_and_quote(decode utf8=>param($field));
+		if ((defined $value) and ($value ne "")) {
+			$change_ref->{$field} = $value;
+		}
+	}
```
VaiTon, line 1116: So we have three string parameters:
- `app_name`
- `app_version`
- `app_uuid`

PR #6335, opened 2022-01-24
fix: changed some log levels to debug

stephanegigandet:
This is to remove logs printed for every product change, and make them debug messages instead.


alexgarel (approved):
LGTM

PR #6342, opened 2022-01-25
fix: show 'we need your help' message for ingredients analysis only when needed

### LINKED ISSUE #6341: Ingredients analysis "We need your help!" message always displayed even when there are no unrecognized ingredients
### Describe the bug

Introduced when adding the specific ingredients parsing feature.

### To Reproduce

https://world.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=fresh+potato&type=add&action=process

### Expected behavior

Message shown only when there are some unrecognized ingredients.

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Browser

### Browser version

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
fixes #6341

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -10996,8 +10996,8 @@ sub display_ingredients_analysis_details($) {
 	my $unknown_ingredients_html = '';
 	my $unknown_ingredients_help_html = '';
 
-	if ($ingredients_text . $specific_ingredients =~ /unknown_ingredient/) {
-		$template_data_ref->{ingredients_text_comp} = 'unknown_ingredient';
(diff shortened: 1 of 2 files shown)


teolemon:
- If I open a product that's both in French and Ukrainian, the main language is used for recognition. 
- We're thus removing the ability to improve Ukrainian.
- This message should ideally apply language by language

VaiTon (approved):

stephanegigandet:
"This message should ideally apply language by language"

The message is shown in context of the ingredients analysis results, so it's relevant only for the language used. Whether we should run the ingredients analysis on all available languages is a much more complex issue.

PR #6345, opened 2022-01-26
New food groups facet

stephanegigandet:
New strings for the food groups facet + new index for MongoDB.


teolemon (commented):
Given the limited avail of MongoDB indexes, how about merging with ?
db.products.createIndex({pnns_groups_1_tags:1,"last_modified_t": -1}, { background: true });
db.products.createIndex({pnns_groups_2_tags:1,"last_modified_t": -1}, { background: true });

stephanegigandet:
> db.products.createIndex({pnns_groups_1_tags:1,"last_modified_t": -1}, { background: true });
> db.products.createIndex({pnns_groups_2_tags:1,"last_modified_t": -1}, { background: true });

Those will be removed eventually, as food groups will replace them. But we need both now to verify that it works. Note that there may be some apps that use them.

PR #6349, opened 2022-01-27
Labeler improvements

teolemon:
Labeler improvements

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -93,6 +93,9 @@ ingredients:
 - lib/ProductOpener/Ingredients.pm
 - taxonomies/ingredients.txt
 
+ingredients analysis:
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -1,6 +1,12 @@
 github_actions:
 - .github/**/*
 
+nginx:
+- conf/nginx/**/*
+
+mongodb:
+- conf/mongodb/create_indexes.js
```
alexgarel, line 8: you could also add `scripts/refresh_products_tags.js`  ?

alexgarel (approved):
LGTM

.github/labeler.yml
teolemon: done

PR #6353, opened 2022-01-27
fix: turn relative links to absolute links in knowledge panels

stephanegigandet:
There's some content in the .po files where we have relative links like <a href="/ecoscore">. But now that we use some of them in knowledge panels, we need to make the links absolute.


alexgarel (approved):
LGTM

PR #6362, opened 2022-01-28
Mozzarella ingredients

stephanegigandet:
Mozzarella is not detected as an ingredient in many instances because the taxonomy contains "mozzarella cheese" as the main entry in many languages.

- Added "mozzarella" and make it the canonical entry
- Added most frequent types of mozzarella used in US products: https://us.openfoodfacts.org/ingredients?filter=mozzar


teolemon (approved):

alexgarel (approved):
Cool !

PR #6373, opened 2022-02-01
new misc facet values for Eco-Score missing packaging warning

stephanegigandet:
This is to add a /misc/en:ecoscore-missing-data-packaging facet.


teolemon (approved):
Approved, would add facets for origins

PR #6377, opened 2022-02-02
feat: Extract ingredients origins from labels and use them in Eco-Score

stephanegigandet:
### What
3 changes:

- There is now a way in the labels taxonomy to specify that a label like "French eggs" means that the origin of the ingredient "en:egg" is "en:france".

en:French eggs, Eggs from France
ingredients:en: en:egg
origins:en: en:france

This info is added to the "specific_ingredients" structure, which can also be populated through parsing the end of the ingredient list (e.g. phrases like "Origin of the milk: United Kingdom").

- The origins of specific ingredients are now used during ingredients analysis to populate the origins field of matching ingredients.

- For the Eco-Score, if we don't have an ingredients list (e.g. products that have a single ingredient like eggs often don't have an ingredient list filled), but if we have an ingredient specific origin (French eggs), we use that origin for the whole product.
### Part of
- #5510


teolemon (approved):

lib/ProductOpener/Ingredients.pm
```diff
@@ -863,13 +865,146 @@ my %ignore_strings_after_percent = (
...
+	my $property = shift;
+
+	my $value;
+
+	# If we have specific ingredients, check if we have a higher fruits / vegetables content
+	if (defined $product_ref->{specific_ingredients}) {
+		foreach my $ingredient_ref (@{$product_ref->{specific_ingredients}}) {
+			my $specific_ingredient_id = $ingredient_ref->{id};
+			if ((defined $ingredient_ref->{origins})
+				and ((not defined $ingredient_id) or (is_a("ingredients", $ingredient_id, $specific_ingredient_id)))) {
```
alexgarel, line 899:
```suggestion
            if ((defined $ingredient_ref->{origins}) # we have an origin for specific ingredient
                # and we did not target a specific ingredient, or this is equivalent to the searched ingredient
				and ((not defined $ingredient_id) or (is_a("ingredients", $ingredient_id, $specific_ingredient_id)))) {
```

lib/ProductOpener/Ingredients.pm
```diff
@@ -863,13 +865,146 @@ my %ignore_strings_after_percent = (
...
+
+				$value = $ingredient_ref->{origins};
+				last;
```
alexgarel, line 902: I'm asking myself in if this kind of treatment we should not continue for sanity checks. Something like:

```perl
defined $value and $value = $ingredient_ref->{origins} 
    or log->warning("Got more than one specific ingredient origin ", {product => $product_ref->{id}, ingredient_id => $ingredient_id});
```

lib/ProductOpener/Ingredients.pm
```diff
@@ -863,13 +865,146 @@ my %ignore_strings_after_percent = (
...
+
+Check if the product has labels that indicate properties (e.g. origins) for specific ingredients.
+
+e.g.
 
-=head2 parse_specific_ingredients_text ( product_ref, $text )
+en:French pork
+fr:Viande Porcine Française, VPF, viande de porc française, Le Porc Français, Porc Origine France, porc français, porc 100% France
+origins:en: en:france
+ingredients:en: en:pork
```
alexgarel, line 957: You should document that origins can be a list, but ingredients not (from what I understand).

alexgarel (approved):
LGTM, eventual cosmetic changes  possibles

lib/ProductOpener/Ingredients.pm
stephanegigandet: good idea. I also just noticed that I hardcoded "origins" in this function even though it's passed a $property parameter.

PR #6384, opened 2022-02-03
change import_csv_file.pl to allow imports without an org and without setting lc and lang

stephanegigandet:
This is to be able to set a field value for a list of barcodes like this:

./import_csv_file.pl --csv carrefour_fr_codes.20200303.csv --user_id driveoff --comment "carrefour.fr 20220303" --define stores="carrefour.fr" --skip_not_existing_products --no_source 

more scripts/carrefour_fr_codes.20200303.csv 
code
0000030000117
0000030005501
0000030005518
0000030043602
0000030075214
0000030182141
0000030182158
0000030182165
0000030182172
0000030182189
0000030182196
0000030182202
0000030182219
0000040122250
0000040122274
0000040144108
0000040144283
0000040198651
0000040198668

We don't want to change the lc / lang field (main language of the product), and we don't have want to add a new data source.

### DIFF
--- a/lib/ProductOpener/Import.pm
+++ b/lib/ProductOpener/Import.pm
@@ -689,7 +689,7 @@ sub import_csv_file($) {
 		}
 
 		# add data_source "Producers"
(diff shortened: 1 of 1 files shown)


alexgarel (approved):
LGTM

Issue #6382, opened 2022-02-03
Translations associated to 2 different entries in categories taxonomy


stephanegigandet:
### What
- Translations associated to 2 different entries in categories taxonomy
- #6378 adds a new check

### Errors in the categories taxonomy definition

- [ ] ERROR - fr:dessert-lacte-pour-bebe already is associated to en:dairy-dessert-for-baby - fr:dessert-lacte-pour-bebe cannot be mapped to entry en:dairy-dessert-for-baby-with-custard
- [ ] ERROR - fi:äidinmaitokorvikkeet already is associated to en:baby-milks - fi:äidinmaitokorvikkeet cannot be mapped to entry en:infant-formulas
- [ ] ERROR - es:bebidas-destiladas already is associated to en:hard-liquors - es:bebidas-destiladas cannot be mapped to entry en:distilled-beverages
- [ ] ERROR - fr:boissons-spiritueuses already is associated to en:spirit-drinks - fr:boissons-spiritueuses cannot be mapped to entry en:distilled-beverages
- [ ] ERROR - es:bebidas-destiladas already is associated to en:hard-liquors - es:bebidas-destiladas cannot be mapped to entry en:eaux-de-vie
- [ ] ERROR - fi:siiderit already is associated to en:ciders - fi:siiderit cannot be mapped to entry en:sparkling-ciders
- [ ] ERROR - he:משקאות-מוגזים already is associated to en:carbonated-drinks - he:משקאות-מוגזים cannot be mapped to entry en:sodas
- [ ] ERROR - pl:napoje-gazowane already is associated to en:carbonated-drinks - pl:napoje-gazowane cannot be mapped to entry en:sodas
- [ ] ERROR - th:น้ำอัดลม already is associated to en:carbonated-drinks - th:น้ำอัดลม cannot be mapped to entry en:sodas
- [ ] ERROR - th:น้ำผลไม้ already is associated to en:juices-and-nectars - th:น้ำผลไม้ cannot be mapped to entry en:fruit-juices
- [ ] ERROR - it:succo-di-mela already is associated to en:apple-juices - it:succo-di-mela cannot be mapped to entry en:squeezed-apple-juices
- [ ] ERROR - nl:verse-grapefruitsappen already is associated to en:fresh-grapefruit-juice - nl:verse-grapefruitsappen cannot be mapped to entry en:squeezed-grapefruit-juices
[...]

PR #6393, opened 2022-02-07
Temporarily disable checks for categories taxonomy

stephanegigandet:
This is to give us time to gradually fix errors in the categories taxonomy: #6382


alexgarel (approved):
LGTM

PR #6413, opened 2022-02-13
chore: Update nova_groups.txt

Roto15:


teolemon (approved):
Thank you very much @Roto15 👍

PR #6419, opened 2022-02-16
fix: ingredient doubles

aleene:
Started working on the doubles in the ingredients


stephanegigandet (approved):
Thank you very much @aleene !

Issue #4264, opened 2020-09-27
Add lactose support in ingredient analysis


teolemon:
Tracker bug for lactose support in ingredient analysis

stephanegigandet:
This would need to be defined more precisely. We do have a detection of dairy products for allergens already. What would we do differently for lactose? There are some dairy foods that contain little lactose (e.g. butter), but they do contain some of it. We could try to estimate the lactose content, and then have thresholds for "low lactose content", but there's some research to do. There are also some "lactose free" milks which contain low lactose (because they have added lactase).

alexgarel:
I also heard that Old hard paste cheese (fromage à pâte dure vieillis) have less lactose.

toas-koas:
Is there any Update? We need a app that can tell me, when lactose is in the food. You can use the labels low lactose or lactose. We have the labels vegan, vegeteria, gluten... why no lactose?

alexgarel:
@toas-koas would you like to add it as a label on dairy products ? 

You can already add whatever label you want. We can add to taxonomy later on. That said, maybe we can continue to discuss the plan first, to see what is more adapted.

teolemon:
@stephanegigandet @alexgarel there would be some advantage in blending:
- ingredient analysis (the current allergen thing we do)
- lactose-free labels
- category analysis

to determine a lactose status for the maximum amount of products.

Issue #4263, opened 2020-09-27
Add support for pork-derived ingredients in ingredient analysis


teolemon:
Add support for pork-derived ingredients in ingredient analysis to help people already building Halal and Kosher apps detect more reliably, and possibly for integration in the official apps at a later point.

teolemon:
@aleene @stephanegigandet I have the feeling we haven't invested as much time as we should on the topic, given its potential to bring us more users in more geographies, and more data from 1st and 3rd party apps.

Issue #1105, opened 2018-02-20
Add kosher/non kosher detection


teolemon:

omeritzics:
Hi,
I don't know whether this is still relevant, but adding kosher detection could also be done by checking whether the product has a Kosher certification (הכשר/Hechsher). Almost any product without a kosher certification is simply non-kosher. There are online databases of products with kosher certifications, some of them are by the certificate givers themselves.

Examples:
https://www.ok.org/product-search/
https://oukosher.org/product-search/
https://isitkosher.uk/

PR #6420, opened 2022-02-16
feat: knowledge panel to display the ingredients that make a product not vegan / vegetarian / palm oil free

stephanegigandet:
This PR creates a new ingredients analysis panel to show if a product is vegan, vegetarian and palm oil free. Click on the title of the panel allows to see the ingredients that cause a product to be non-vegan etc.

Other changes:
- Adding a ?field= parameter at the end of a product URL now automatically triggers the JSON API mode, without having to also add the /api/v2/ at the beginning of the URL. This is for debug purposes: I need to see the content of fields very very often. It will use the latest version of the API, so it should not be used by apps etc.
- Changed the vegetarian icon so that it's not the same as the egg allergen icon.

image


lib/ProductOpener/Display.pm
```diff
@@ -11084,19 +11099,55 @@ sub display_ingredients_analysis($) {
 				}
 			}
 
-			push @{$template_data_ref->{ingredients_analysis_tags}}, {
+			my $ingredients_title_id;
+			if ($evaluation eq "unknown") {
+				$ingredients_title_id = "unrecognized_ingredients";
+			}
+			else {
+				# convert en:non-vegetarian property to non_vegetarian_ingredients string id
```
alexgarel, line 11107:
```suggestion
				# convert analysis tag to a string id
				# eg. en:non-vegetarian property to non_vegetarian_ingredients string id
```

lib/ProductOpener/Display.pm
```diff
@@ -11084,19 +11099,55 @@ sub display_ingredients_analysis($) {
 				}
 			}
 
-			push @{$template_data_ref->{ingredients_analysis_tags}}, {
+			my $ingredients_title_id;
```
alexgarel, line 11102:
```suggestion
			# Attribute to right lists for display (to analyze classifications)
			my $ingredients_title_id;
```

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -297,7 +297,7 @@ sub create_panel_from_json_template ($$$$$$) {
         # As it is a trailing comma inside a string, it's not a terrible issue, the string will be valid,
         # but it will have an unneeded trailing comma.
         # The group (\W) at the end is to avoid removing commas before an opening quote (e.g. for "field": true, "other_field": ..)
-        $panel_json =~ s/(?<!("|'|\]|\}|\d))\s*,\s*"(\W)/"$2/g;
+        $panel_json =~ s/(?<!("|'|\]|\}|\d))\s*,\s*"(\W)/"$2/sg;
```
alexgarel, line 300: hh, this kind of stuff always looks simple at start, but you always find a bug later on :stuck_out_tongue_closed_eyes:

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -751,6 +751,9 @@ sub create_health_card_panel($$$) {
     # Create the additives panel
     create_additives_panel($product_ref, $target_lc, $target_cc);
 
+    # Create the additives panel
```
alexgarel, line 754:
```suggestion
    # Create the ingredients analysis panel
```

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -1065,6 +1068,54 @@ sub create_additives_panel($$$) {
...
+	my $target_cc = shift;
+
+	$log->debug("create ingredients analysis panel", { code => $product_ref->{code} }) if $log->is_debug();
+
+    my $ingredients_analysis_data_ref = data_to_display_ingredients_analysis($product_ref);
+
+    if (defined $ingredients_analysis_data_ref) {
+
+        my $ingredients_analysis_panel_data_ref = {
+        };
```
alexgarel, line 1103: you should find another (shorter ?) variable name, because I didn't see the difference between `ingredients_analysis_data_ref` and `ingredients_analysis_panel_data_ref`.

Isn't `panel_data_ref` sufficient, given the limited scope ?

alexgarel: Also you do not reference it but in last line ? Seems strange in this case that you need this.

alexgarel (changes requested):
I think there is a bug in the panel:

Capture d’écran de 2022-02-16 16-37-51

I think I spotted where with a comment.

lib/ProductOpener/KnowledgePanels.pm
stephanegigandet: yes... what's the meaning of "hh"?

teolemon (approved):
- It would be nice to have a nudge for ingredient extraction, in case ingredients_tags are not defined (ingredients not extracted). "Extract/Photograph the ingredients to get the vegan status"
- In the future, we can even make the unrecognized ingredients clickable, so that users can request them to be added, or add them theirselves.
- It would be great to add #4264 (new detection) and #1106 (for better explainability) (and more far fetched: #4263 / #1105)

lib/ProductOpener/KnowledgePanels.pm
stephanegigandet: We don't need it indeed.

Issue #6424, opened 2022-02-17
Refactor CSV / XSLX exports to have 1 export format (and code) instead of 3


stephanegigandet:
### Is your feature request related to a problem? Please describe.

We currently have 3 different CSV exports, each with different fields, and with different code:

1. the full database CSV export, generated nightly with scripts/export_database.pl
2. the on demand CSV / XLSX export, available through the search form (code in lib/ProductOpener/Display.pm search_and_export_products()
3. the CSV export used internally to export and import data from the producers platform to the public platform (code in lib/ProductOpener/Export.pm, called from the producers platform, and also availble from export_csv.pl)

### Describe the solution you'd like

Gradually converge to 1 common format and 1 single code base (Export.pm).

Desirable features:
- let producers download an export of their data on the producers platform, make some changes to it, and reimport it
- clear separation between columns that contain primary data (e.g. ingredients text, nutrition values, Nutri-Score computed by the producer) vs derived data (canonical ingredients from ingredients_tags, Nutri-Score computed by OFF, Eco-Score details). e.g. maybe prefix the column name with off:  off:ingredients_tags

stephanegigandet:
My plan is to first start to make the on demand search export on the website use the code from Export.pm. The search export is requested manually, so it's probably not a big issue if we change which columns are included and their names.

Once we have something that we are happy with, we can look into replacing the full database CSV export, probably in a new location, so that we don't break things.

alexgarel:
Convergence of exports is a real big plus.

When refactoring, I think it would be great to separate processing into map like structure, from rendering to xls.

My only small concern is that for communication between producer and public platform, I would go away from the csv/xls format… because it seems to me that in introduces some tweaks that are not useful.

PR #6444, opened 2022-02-25
feat: Refactor CSV exports, add tests, export Eco-Score fields

stephanegigandet:
- Refactor of the CSV exports so that export_csv_file.pl and the on demand CSV / XLSX exports on the website use the same code. See #6424
- prefixed fields that are computed by OFF with "off:". This is to make it clear that the fields are computed. If the exported file is reimported, those fields should be ignored.
- add an option to export the normalized _tags fields. e.g. "categories_tags", in addition to "categories" which is exported in the main language of the product.
- Added an integration test export.t : a CSV file is first imported, then products are exported using various export parameters
- Added ecoscore related fields to the exports

Sample output:
[openfoodfacts_export (12).xlsx](https://github.com/openfoodfacts/openfoodfacts-server/files/8142833/openfoodfacts_export.12.xlsx)


lib/ProductOpener/Export.pm
```diff
@@ -306,21 +333,42 @@ sub export_csv($) {
...
 						}
-						elsif ((defined $product_ref->{$field}) and ($product_ref->{$field} ne "")) {
-							$populated_fields{$field} = sprintf("%08d", $group_number * 1000 + $item_number);
+						# Allow returning fields that are not at the root of the product structure
+						# e.g. ecoscore_data.agribalyse.score  -> $product_ref->{ecoscore_data}{agribalyse}{score}
+
+						# Special case for ecoscore_data.adjustments.origins_of_ingredients.value
+						# which is only present if the Eco-Score fields have been localized (done only once after)
+						elsif ($field eq "ecoscore_data.adjustments.origins_of_ingredients.value") {
+							if (deep_exists($product_ref, split(/\./, "ecoscore_data.adjustments.origins_of_ingredients.values"))) {
```
alexgarel, line 360: you do this to add the "s" to `value` ? 
It's not explicit in your comment and hard to catch at first sight !

lib/ProductOpener/Export.pm
```diff
@@ -306,21 +333,42 @@ sub export_csv($) {
...
+						}
+						elsif ($field =~ /\./) {
+							if (deep_exists($product_ref, split(/\./, $field))) {
+								$populated_fields{$group_prefix . $field} = sprintf("%08d", $group_number * 1000 + $item_number);
+							}
+						}
+
+						elsif (defined $product_ref->{$field}) {
+							$populated_fields{$group_prefix . $field} = sprintf("%08d", $group_number * 1000 + $item_number);
 						}
```
alexgarel, line 372: To:
1. make the added "s" explicit and not so special
2. unite the case where you have a "." or not and always use deep_exists

I would write.

```suggestion
						else {
						  my $key = $field;
						  if ($field eq "ecoscore_data.adjustments.origins_of_ingredients.value") {
						    $key = $key . "s"
						  }
						  if ($deep_exists($product_ref, split(/\./, $key))) {
								$populated_fields{$group_prefix . $field} = sprintf("%08d", $group_number * 1000 + $item_number);
							}
						}

```

lib/ProductOpener/Export.pm
```diff
@@ -306,21 +333,42 @@ sub export_csv($) {
 
 						if (defined $tags_fields{$field}) {
 							if ((defined $product_ref->{$field . "_tags"}) and (scalar @{$product_ref->{$field . "_tags"}} > 0)) {
-								$populated_fields{$field} = sprintf("%08d", $group_number * 1000 + $item_number);
+								# Export the tags field in the main language of the product
+								$populated_fields{$group_prefix . $field} = sprintf("%08d", $group_number * 1000 + $item_number);
```
alexgarel, line 337: why compute `sprintf("%08d", $group_number * 1000 + $item_number)` once and use it all over the place there after.

```perl
my $field_id = sprintf("%08d", $group_number * 1000 + $item_number);
...
$populated_fields{$group_prefix . $field} = $field_id;
...
$populated_fields{$group_prefix . $field . "_tags"} = $field_id . "_tags";
```

lib/ProductOpener/Export.pm
```diff
@@ -360,29 +408,92 @@ sub export_csv($) {
...
+		# Send HTTP headers, unless search_and_export_products() is called from a script
+		if ($send_http_headers) {
+			require Apache2::RequestRec;
+			my $r = Apache2::RequestUtil->request();
+			$r->headers_out->set("Content-type" => "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet");
+			$r->headers_out->set("Content-disposition" => "attachment;filename=$filename");
+			print "Content-Type: application/vnd.openxmlformats-officedocument.spreadsheetml.sheet\r\n\r\n";
+
+			binmode($filehandle);
+		}
```
alexgarel, line 428: Not mandatory, but I would really advocate to put it in a separate function which is called where needed, by caller of `export_csv`

Deep nesting web processing deep inside the code is bad practice. (remember I spoke about de-mangling our code)

t/export.t
```diff
@@ -0,0 +1,221 @@
...
+    # clean database
+    execute_query(sub {
+		return get_products_collection()->delete_many({});
+	});
+    # clean files
+    remove_tree("$data_root/products", {keep_root => 1, error => \my $err});
+    if (@$err) {
+        die("not able to remove some products directories: ". join(":", @$err));
+    }
+}
```
alexgarel, line 48: This should go it a test util, but I didn't know where to put it ! (in a new lib directory ?)

t/export.t
```diff
@@ -0,0 +1,221 @@
...
+        
+        if (defined $update_expected_results) {
+            open (my $result, ">:encoding(UTF-8)", "$expected_results_dir/$code.json") or die("Could not create $expected_results_dir/$code.json: $!\n");
+            print $result $json->pretty->encode($product_ref);
+            close ($result);
+        }
+
+        # Compare the result with the expected result
+        
+        if (open (my $expected_result, "<:encoding(UTF-8)", "$expected_results_dir/$code.json")) {
```
alexgarel, line 120: No need to compare if we just saved it !

```suggestion
        elif (open (my $expected_result, "<:encoding(UTF-8)", "$expected_results_dir/$code.json")) {
```

alexgarel (approved):
LGTM (not tested yet) appart some style

lib/ProductOpener/Export.pm
stephanegigandet: Good idea, thank you!

t/export.t
stephanegigandet: Right. :)

lib/ProductOpener/Export.pm
stephanegigandet: Good point, I'll change it.

lib/ProductOpener/Export.pm
stephanegigandet: Agreed, but I can't think of a good way to do it, because the caller of export_csv does not know in advance if it will have to show an error (e.g. database issue), or if the export will succeed.

PR #6453, opened 2022-03-02
chore: Update ingredients.txt

Atos10:
Update of ingredients for bulgarian but changes for other languages are not mine strange.


stephanegigandet:
> Update of ingredients for bulgarian but changes for other languages are not mine strange.

That's strange, it looks like the commit also removes some recent changes from @aleene .

PR #6455, opened 2022-03-04
chore: Roto15 ingredients robi

stephanegigandet:
Trying to fix conflicts in https://github.com/openfoodfacts/openfoodfacts-server/pull/6453


stephanegigandet:
Unfortunately it would overwrite a lot of changes made by @aleene  :(

Roto15:
Ok, so I have to try to do the changes again I suppose because I don't see a solution @stephanegigandet

PR #6459, opened 2022-03-05
feat: Automatically label issues

teolemon:
### Checklist
- Automatically label issues


.github/issue-labeler.yml
```diff
@@ -0,0 +1,1019 @@
...
+no-pr-activity:
+    - '(no-pr-activity)'
+
+nodejs:
+    - '(nodejs)'
+
+non tech:
+    - '(non tech)'
+
+not done
```
stephanegigandet, line 581: typo?

stephanegigandet (changes requested):
issue-labeler can only be used for a very small and specific set of labels that are expected in an issue template.

It's probably not a good match for this repo that has lots of labels.

Even on other repos with fewer labels, please try it with one label at a time, and uncomment the not-before option, so that we can check that it does not break old issues.

Specifically, merging this PR as-is would remove almost all existing labels on almost all issues.

.github/issue-labeler.yml
teolemon: fixed

teolemon:
@stephanegigandet I've disabled processing of issues older than today, and enabled a revision mechanism

stephanegigandet:
@teolemon : this PR will remove most labels entered manually, unless their body also match the exact regexps (including case) listed.

stephanegigandet:
@teolemon example : I changed the title of this issue https://github.com/openfoodfacts/smooth-app/issues/1166 , and the action removed the ios label, because it is not in the body.

stephanegigandet:
Closing as issue-labeler does not work as wanted.

PR #6466, opened 2022-03-08
chore: re-apply Bulgarian translations from Roto15

stephanegigandet:
Reapplied https://github.com/openfoodfacts/openfoodfacts-server/pull/6453 changes to keep only the new Bulgarian translations from @Roto15


alexgarel (approved):
:+1: thank you @Roto15

Roto15:
Thanks @alexgarel and @stephanegigandet

PR #6472, opened 2022-03-09
fix: Cleaning ingredients

aleene:
### Description
Cleaning up the ingredients taxonomy
* removing doubling with other taxonomies


stephanegigandet (approved):
Thank you!

PR #6474, opened 2022-03-09
fix: Import serving size bug

### LINKED ISSUE #6473: Import of nutrition data specified for a serving size fail, and set the serving size to "serving"
### Describe the bug

There's a broken logic in Import.pm, when nutrition_data_per is set to "serving", then we set "serving_size" to "serving" as well.

### To Reproduce

Will add test to reproduce the problem.

### Expected behavior

serving_size is not changed

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Other

### Browser version

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
fixes #6473 , adds a test


alexgarel (approved):
LGTM

Issue #2973, opened 2020-03-05
Automatic imports make empty edits on the products from Carrefour


teolemon:
Each morning during the import sequence, empty edits are made on the products from Carrefour

```
8431876208993; March 5, 2020 at 8:25:03 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
8012666011377; March 5, 2020 at 8:25:01 AM CET - carrefour (- Editing product (import) - Import Carrefour) [Data (Changed: quantity) ] - view
3560071012236; March 5, 2020 at 8:24:58 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
3560070821204; March 5, 2020 at 8:24:55 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
3560070820306; March 5, 2020 at 8:24:52 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
3560070750238; March 5, 2020 at 8:24:50 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
3560070592630; March 5, 2020 at 8:24:47 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
3560070416127; March 5, 2020 at 8:24:44 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
3560070333127; March 5, 2020 at 8:24:41 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
4104420133747; March 5, 2020 at 8:24:40 AM CET - prepperapp (- (app)Edited by a user of https://speisekammer-app.de) [Data (Changed: product_name, categories, traces, product_name_de) -- Nutriments (Added: energy-kcal, energy, fat, carbohydrates, sugars, proteins) ] - view
3245412966156; March 5, 2020 at 8:24:37 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
3245412563331; March 5, 2020 at 8:24:35 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
3245412435263; March 5, 2020 at 8:24:32 AM CET - carrefour (- Editing product (import) - Import Carrefour) [Data (Changed: quantity) ] - view
3245412256639; March 5, 2020 at 8:24:29 AM CET - carrefour (- Editing product (import) - Import Carrefour) [] - view
[...]

PR #6476, opened 2022-03-09
fix: Add tests for Carrefour France import, + solve serving_size bug

stephanegigandet:
There's a long standing issue #2973 that results in almost daily changes being made to some products imported by Carrefour France. The corresponding input files in fact did not change, but we were mapping multiple values (for different languages) to the same field: the values were concatenated, but in a not specified order. So the values changed.

This PR:
- adds some tests so that we can catch those issues, and also so that we can more easily improve the Carrefour data conversion (if needed)
- uses only the French value for fields like the quantity (we could try to support more cases, but the extra logic may not be worth the very few cases where that might be useful)
- solves some small issues, undefined values etc.

Note: in order to be able to create a test, most of the code from convert_carrefour_data.pl was moved to a new ImportConvertCarrefourFrance.pm module, but no attempt was made to improve that code.


alexgarel (changes requested):
I think you forgot to add ImportConvertCarrefourFrance.pm @stephanegigandet  !

alexgarel (approved):
LGTM

PR #6483, opened 2022-03-14
feat: more flexible exports

stephanegigandet:
- add ingredients_tags to the full CSV export
- add options to export_csv_file.pl

### DIFF
--- a/scripts/export_database.pl
+++ b/scripts/export_database.pl
@@ -283,7 +283,21 @@ XML
 
 		foreach my $field (@export_fields) {
 
-			my $field_value = ($product_ref->{$field} // "");
+			my $field_value;
+
+			# _tags field contain an array of values
+			if ($field =~ /_tags/) {
+				if (defined $product_ref->{$field}) {
+					$field_value = join(',', @{$product_ref->{$field}});
+				}
+				else {
+					$field_value = "";
+				}
+			}
+			# other fields
+			else {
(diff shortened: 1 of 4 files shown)


scripts/export_database.pl
```diff
@@ -283,7 +283,21 @@ sub sanitize_field_content {
 
 		foreach my $field (@export_fields) {
 
-			my $field_value = ($product_ref->{$field} // "");
+			my $field_value;
+
+			# _tags field contain an array of values
+			if ($field =~ /_tags/) {
+				if (defined $product_ref->{$field}) {
+					$field_value = join(',', @{$product_ref->{$field}});
```
alexgarel, line 291: `', '` is ok because the csv lib will handle it putting thing in quotes.

But is this the most convenient representation ? Isn't `\n` more convenient for a human reader ?

alexgarel (approved):
LGTM, just one comment on list separator.

scripts/export_database.pl
stephanegigandet: a good question.. but is the CSV for human readers?

I kind of like the fact that we can have one line per product, so we can count them, grep them, head/tail them etc. If CSV uses quotes and keep the new lines, we wouldn't be able to do that anymore

alexgarel: ok

PR #6485, opened 2022-03-14
fix: fix image upload buttons

### LINKED ISSUE #173: Adding a picture button problem
When a product page is empty, clicking on the "Add a picture" button located under "Ingredients" moves us to the top of the page.

stephanegigandet:
Fixes #173

The file upload buttons (for images, and data file on producers platform) have a <a href="#"> link that encapsulate a <input> button. But the button works only in the top half, clicking on the lower half makes the page jump to the top of the page.

There were 2 issues:
- the input button was only half height, fixed by height:100%
- the href="#" makes the browser go to the top of the page (as identified by @chk1).


alexgarel (approved):
:+1:

PR #6487, opened 2022-03-15
build: make tests run in isolation

alexgarel:
Make tests run in isolation.

We now have integration tests, avoid destroying dev database each time he wants to run tests.


stephanegigandet (approved):
Looks good to me, I ran the tests and I still have my previous products afterwards. :)

PR #6492, opened 2022-03-16
fix: untranslated string for Smoothie

teolemon:
### What 
- Fix untranslated string for Smoothie


stephanegigandet (approved):
Looks good but you need to add the new string to en.po as well so that we can deploy without having to wait for the next crowdin merge

PR #6495, opened 2022-03-18
chore: ADD BG to packaging_materials.result.txt

Roto15:


Roto15:
I just saw that I shouldn't edit this file manually but if I edited it is it wrong?

stephanegigandet:
> I just saw that I shouldn't edit this file manually but if I edited it is it wrong? I already translated packaging_materials.txt but I still don't see the changes in the files like packaging_materials.result.txt for example.

That's because the corresponding taxonomies files have to been rebuilt, it's currently a manual process. The *.result.txt files have no effect, they are here only for debugging purposes.

I will rebuild the taxonomies with your changes.

stephanegigandet:
Closing this PR. PR for the built taxonomies: https://github.com/openfoodfacts/openfoodfacts-server/pull/6497

PR #6503, opened 2022-03-23
fix: release please should trigger actions

alexgarel:
We need actions to be triggered by release please this means using a personal access token instead of the default one.

The secret is already created in the repo.


stephanegigandet (approved):
That will solve the fact that tests don't seem to run for release-please releases?

PR #6509, opened 2022-03-24
chore: Meat alternatives

moon-rabbitOFF:


taxonomies/categories.txt
```diff
@@ -54777,7 +54781,7 @@ fr:Purées de sésame demi-complet, Purées de sésame semi-complet
 nl:Half-volkoren sesampastas
 
 <en:Cereals and their products
-<en:Meat analogues
+<en:Meat alternatives
```
stephanegigandet, line 54784:
```suggestion
<en:Meat alternatives, meat analogues
```

If we rename a category, it's best to keep the old name as a synonym, otherwise products that already have it will have an unexisting category.

moon-rabbitOFF: I kept the category "meat analogues" under "meat alternatives" (see the discussion in slack)

stephanegigandet (approved):
Thanks!

PR #6510, opened 2022-03-25
feat: Record the reasons for NOVA classification of a product and add a NOVA knowledge panel

stephanegigandet:
### What

- new nova_groups_markers field that contains the elements (e.g. categories, ingredients, additives) that resulted in a given NOVA group for a product
- new "unknown" value in addition to "not-applicable" for nova_groups_tags
- new NOVA group panel to show the NOVA classification and why a product was classified that way.

### Part of
- #5493 

### Sample NOVA panel

image


lib/ProductOpener/Food.pm
```diff
@@ -2239,45 +2244,53 @@ sub compute_nova_group($) {
...
-
 			}
 		}
 	}
 
-	# Also loop through categories and ingredients to see if the taxonomy has associated minimum NOVA grades
-	# Categories need to be first, so that we can identify group 2 foods such as salt, sugar, fats etc.
+
 	# Group 2 foods should then not be moved to group 3
 	# (e.g. sugar contains the ingredient sugar, but it should stay group 2)
```
alexgarel, line 2255: I think this comment should be below.

lib/ProductOpener/Food.pm
```diff
@@ -2239,45 +2244,53 @@ sub compute_nova_group($) {
...
-
-					if ($options{nova_groups_tags}{$tag} > $product_ref->{nova_group}) {
-
-						# only move group 1 product to group 3, not group 2
-						if (not (($product_ref->{nova_group} == 2) and ($options{nova_groups_tags}{$tag} == 3))) {
-							$product_ref->{nova_group_debug} .= " -- $tag : " . $options{nova_groups_tags}{$tag} ;
-							$product_ref->{nova_group} = $options{nova_groups_tags}{$tag};
-						}
-					}
+					push @matching_tags, [$tagtype, $tagid, $options{nova_groups_tags}{$tag}];
```
alexgarel, line 2247: can we fix indentation problem here ?

lib/ProductOpener/Food.pm
```diff
@@ -2219,15 +2222,17 @@ sub compute_nova_group($) {
...
-	#
-	# "categories/en:fats" => 2,
+	# We currently have 2 sources for tags that can trigger a given NOVA group:
+	# 1. tags specified in the %options of Config.pm 
+	# 2. tags in the categories, ingredients and additives taxonomy that have a nova:en property
 
+	# We first generate a list of matching tags, from the two sources
+	# Note that elements must stay in order, as we need t check first for categories that trigger group 2 (culinary ingredients)
+
+	my @matching_tags = ();
```
alexgarel, line 2235: good idea to refactor this way

lib/ProductOpener/Food.pm
```diff
@@ -2239,45 +2244,53 @@ sub compute_nova_group($) {
...
+	# Group 2 foods should then not be moved to group 3
+	# (e.g. sugar contains the ingredient sugar, but it should stay group 2)
+
+	my %seen_markers = ();
+
+	foreach my $matching_tag_ref (@matching_tags) {
+		my ($tagtype, $tagid, $nova_group) = @$matching_tag_ref;
+		if (($nova_group >= $product_ref->{nova_group})
+			# don't move group 2 to group 3
+			and not (($nova_group == 3) and ($product_ref->{nova_group} == 2))
```
alexgarel, line 2282: I would like to have clarification on this, this works because of the order in which we process things right ? config option then categories then ingredients then additives, isn't it ?

Then it is worth mentioning I think. (it twist my mind every time).

alexgarel (commented):
LGTM but I want to test it locally

lib/ProductOpener/Food.pm
alexgarel: Maybe make it more robust, by adding a condition on tag_type ?

stephanegigandet: I'm changing it to have separate lists of matching tags for each NOVA group, it will be more clear.

alexgarel:
You didn't implement the click on nova score that should bring to the panel. (I imagine you forgot to add the id).

templates/api/knowledge-panels/health/ingredients/nova.tt.json
```diff
@@ -0,0 +1,59 @@
...
+            "text_element": {
+                "html": `
+[% IF ! product.nova_group.defined %]
+    [% IF product.nova_group_error.defined %]
+       [% lang("nova_group_" _ product.nova_group_error) %]
+    [% END %]
+[% ELSE %]
+    [% IF nova_group == 1 %]
+    [% ELSIF nova_group >= 2 %]
+        [% IF product.nova_groups_markers.$nova_group.defined %]
```
alexgarel, line 28: this part seems broken, I have nothing displayed (but I enter the IF, I put a simple string to be sure)

alexgarel (changes requested):
I don't have the panel when I test locally.

templates/api/knowledge-panels/health/ingredients/nova.tt.json
stephanegigandet: I forgot to mention that the NOVA analysis has to be done again on the product for the nova markers to be computed. If you edit and save the product, that should be enough.

PR #6513, opened 2022-03-26
outreachy uninitialized value update

dipietroR:
### Description

I'm making a commit for an outreachy project

Originally I was seeing
[Sat Mar 26 20:37:58 2022] -e: Use of uninitialized value in concatenation (.) or string at /opt/product-opener/lib/ProductOpener/Display.pm line 8127.

this edit seems to resolve that.

### DIFF
--- a/taxonomies/nutrients.txt
+++ b/taxonomies/nutrients.txt
@@ -447,6 +447,7 @@ en:Caproic acid, Hexanoic acid
 xx:Caproic acid, Hexanoic acid
 ar:حمض الهكسانويك
 az:Kapron turşusu
+bg:Капронова киселина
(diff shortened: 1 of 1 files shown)


hangy (commented):
I think this just shadows https://github.com/openfoodfacts/openfoodfacts-server/blob/9eb12c3adc539c4860074a7891e83333867e0506/lib/ProductOpener/Display.pm#L109 and thus might not work as expected.

dipietroR:
Sorry I changed branches and didn't realize that would close the PR. I opened a new one here:
https://github.com/openfoodfacts/openfoodfacts-server/pull/6514

PR #6514, opened 2022-03-26
fix: outreachy uninitialized value update

dipietroR:
Sorry moved my git branch and ended up closing the old PR (https://github.com/openfoodfacts/openfoodfacts-server/pull/6513)

### Description

I'm making a commit for an outreachy project

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -8117,21 +8117,25 @@ HTML
 <meta name="twitter:creator" content="@<twitter_account>">
 <meta name="twitter:title" content="$title">
 <meta name="twitter:description" content="$description">
-<meta name="twitter:label1" content="$Lang{brands_s}{$lc}">
-<meta name="twitter:label2" content="$Lang{categories_s}{$lc}">
 HTML
 ;
 
-	if (defined $product_ref->{brand}) {
+	if (defined $product_ref->{brands}) {
+		# print only first brand if multiple exist.
+		my @brands = split(',', $product_ref->{brands});
 		$header .= <<HTML
-<meta name="twitter:data1" content="$product_ref->{brand}">
+<meta name="twitter:label1" content="$Lang{brands_s}{$lc}">
+<meta name="twitter:data1" content="$brands[0]">
 HTML
 ;
 	}
 
-	if (defined $product_ref->{category}) {
+	# get most specific category (the last one)
+	my $data2 = display_taxonomy_tag($lc, "categories", $product_ref->{categories_tags}[-1]);
+	if ($data2) {
 		$header .= <<HTML
-<meta name="twitter:data2" content="$product_ref->{category}">
+<meta name="twitter:label2" content="$Lang{categories_s}{$lc}">
+<meta name="twitter:data2" content="$data2">
 HTML
 ;
 	}
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -8111,16 +8111,6 @@ HTML
 ;
 	}
 
-	#unless (defined $product_ref->{brand}) {
-	#	$product_ref->{brand} = "";
-	#}
-	#unless (defined $product_ref->{category}) {
-	#	$product_ref->{category} = "";
-	#}
-
-	$product_ref->{brand} //= "";
-	$product_ref->{category} //= "";
-
 	$header .= <<HTML
 <meta name="twitter:card" content="product">
 <meta name="twitter:site" content="@<twitter_account>">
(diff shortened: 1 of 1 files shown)


dipietroR:
I'm not sure if this is the meaning of the comment in https://github.com/openfoodfacts/openfoodfacts-server/pull/6513

I was thinking it might have been a scope problem and https://stackoverflow.com/questions/845060/what-is-the-difference-between-my-and-our-in-perl gave me the idea for scoping to the package. Rather than using 'my'

stephanegigandet:
I'm copying the description you had in the other PR:

----
Description
I'm making a commit for an outreachy project

Originally I was seeing
[Sat Mar 26 20:37:58 2022] -e: Use of uninitialized value in concatenation (.) or string at /opt/product-opener/lib/ProductOpener/Display.pm line 8127.

this edit seems to resolve that.

---

I don't think the change you make is affecting the warning you saw.  "Use of uninitialized value in concatenation (.) " --> this is indicating that one of the variables used on the right side of the expression is unitialized.

```
	$header .= <<HTML
<meta name="twitter:card" content="product">
<meta name="twitter:site" content="@<twitter_account>">
<meta name="twitter:creator" content="@<twitter_account>">
<meta name="twitter:title" content="$title">
<meta name="twitter:description" content="$description">
<meta name="twitter:label1" content="$Lang{brands_s}{$lc}">
<meta name="twitter:data1" content="$product_ref->{brand}">
<meta name="twitter:label2" content="$Lang{categories_s}{$lc}">
<meta name="twitter:data2" content="$product_ref->{category}">
$meta_product_image_url

HTML
;
```

So the issue is not with "$header", but with one or more of the variables like $title or $description. I'm not sure if we define $product_ref->{category} somewhere.

dipietroR:
Maybe this is the fix. If it is...

https://stackoverflow.com/a/3738936 gave the idea of using //= which seems cleaner than the block, but is perhaps not obvious? I do see it used in the code in
`$series_n{$seriesid} //= 0;`
So it is at least not completely unexpected. Which is preferred for this code, the more verbose or the more compact?

Regardless is there a better place for those variables to be set to "" if they were not set?

stephanegigandet:
> https://stackoverflow.com/a/3738936 gave the idea of using //= which seems cleaner than the block, but is perhaps not obvious? I do see it used in the code in `$series_n{$seriesid} //= 0;` So it is at least not completely unexpected. Which is preferred for this code, the more verbose or the more compact?

Maybe we can use the //= and add a comment like  this, as not everyone is used to that operator.

# Set the value of undefined fields to an empty string with the defined-or operator
$product_ref->{brand} //= "";
$product_ref->{category} //= "";

> Regardless is there a better place for those variables to be set to "" if they were not set?

In fact we don't really want to set them, but we should not display them if they are undefined.

An easy way would be to do:

my $brand = $product_ref->{brand} // "";
and then use $brand.

Even better would be to not output
<meta name="twitter:label1" content="$Lang{brands_s}{$lc}">
<meta name="twitter:data1" content="$product_ref->{brand}">

if we don't have a value (instead of setting an empty value).

The best thing would be to use a template to output that HTML code, as we do for other parts of the product page.
(existing templates are in templates/web/pages/product/includes/ )

dipietroR:
That makes sense that we should not define the variable if we don't have to. If I'm understanding correctly, we don't need the whole line of HTML. Which I believe is what this commit will do. I'm not sure if the order of the lines matter, so it is a little different now:
```
<meta name="[twitter:label1]()" content="[brand]()">
<meta name="[twitter:data1]()" content="">
<meta name="[twitter:label2]()" content="[category]()">
<meta name="[twitter:data2]()" content="">
```
became:
```
<meta name="[twitter:label1]()" content="[brand]()">
<meta name="[twitter:label2]()" content="[category]()">
```
followed by any data lines, should any exist. If the order matters the code can be updated to keep the label1, data1, label2, data2 pattern.

I couldn't find where this bit of the page is generated in templates/web/pages/product/includes, though the source of the page had a helpful comment:
`<!-- start templates/web/common/site_layout.tt.html -->`
Which sure enough has a header variable! :smile: 
https://github.com/openfoodfacts/openfoodfacts-server/blob/c5cdaaa07f1d1e44036a88d80c2cb49dc65012c3/templates/web/common/site_layout.tt.html#L23

stephanegigandet:
In fact the original code has been broken for the last 6 years, the $product_ref->{brand} and $product_ref->{categories} fields are never set... so we always have empty entries if you look at the HTML code of a product page:


```
<meta name="twitter:creator" content="@OpenFoodFacts">
--
  | <meta name="twitter:title" content="Allumettes - Carrefour - 1 kg">
  | <meta name="twitter:description" content="Ingredients, allergens, additives, nutrition facts, labels, origin of ingredients and information on product Allumettes - Carrefour - 1 kg">
  | <meta name="twitter:label1" content="brand">
  | <meta name="twitter:data1" content="">
  | <meta name="twitter:label2" content="category">
  | <meta name="twitter:data2" content="">
```

$product_ref->{brand} should be $product_ref->{brands} (with an s). If the field is long, we might want to take only the first brand, before the first comma.

category is different, we have a categories field, but it should not be used, instead we should use the display_taxonomy_tags function, and then get the first category (or maybe the last, I never remember which is the most specific one).

When we don't have data1, we shouldn't have label1.

dipietroR:
How does this look? The most specific category seems to be the last in the list. If there is no data1 (or label1) does that mean data2 should be printed as data1 instead?

stephanegigandet (approved):
Looks good, thank you!

PR #6549, opened 2022-04-02
chore: Add BG to packagings.txt

Roto15:


stephanegigandet (approved):
We can merge it, but things that are in /old/ are not in use, they are kept for reference only.

PR #6553, opened 2022-04-04
fix: languages and nutrients taxonomies

stephanegigandet:
Removed or fixed some translations that were assigned to more than 1 entry.


alexgarel (approved):
:clap:

PR #6555, opened 2022-04-04
docs: Add README to avoid overwritten contribs

teolemon:
### Checklist

doc: Add a translation README to avoid translations being overwritten

### DIFF
--- a/po/README.md
+++ b/po/README.md
@@ -1,3 +1,3 @@
 ### Translations
 You can translate Open Food Facts at https://translate.openfoodfacts.org. Please do not translate directly in the code, or your translations will be overwritten.
-If you have added or modified a string, you should update the crowdin-trigger branch from main/develop. This will trigger import within Crowdin, and 1 hour later, it will create a new pull request with added or modified strings in all language files, plus all added or modified translations.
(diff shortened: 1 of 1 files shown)


po/README.md
```diff
@@ -0,0 +1,3 @@
+### Translations
+You can translate Open Food Facts at https://translate.openfoodfacts.org. Please do not translate directly in the code, or your translations will be overwritten.
+If you have added or modified a string, you should update the crowdin-trigger branch from main/develop. This will trigger import within Crowdin, and 1 hour later, it will create a new pull request with added or modified strings in all language files, plus all added or modified translations.
```
alexgarel, line 3: Wow I didn't know about this.

It's not clear to me what you mean by "update the crowdin-trigger branch from main/develop" does this mean "merging main into crowdin-trigger" ? 
Could'nt this be done by looking at po files instead of a branch (actions can be conditioned by files).

teolemon: - there are some times where you're trying to merge translations, and you don't want it auto-triggering
- we could give the file based trigger a try

stephanegigandet, line 3:
```suggestion
If you have added or modified a string in the source file (the file with the .pot extension), you should update the crowdin-trigger branch from main/develop. This will trigger import within Crowdin, and 1 hour later, it will create a new pull request with added or modified strings in all language files, plus all added or modified translations.
```

stephanegigandet (approved):

PR #6556, opened 2022-04-05
fix: Dutch next round

aleene:
Went through the new additions and improved the taxonomies accordingly.


stephanegigandet:
ERROR - nl:duurzame-visserij already is associated to en:sustainable-fishing-method - nl:duurzame-visserij cannot be mapped to entry en:sustainable-fishing

-> this one is subtle: we want to differentiate "Sustainable fishing" (which contains anything related to sustainable fishing, like the MSC label), from "Sustainable fishing methods".

PR #6566, opened 2022-04-06
feat: Improvements to GS1 imports to prepare integration of Agena3000

stephanegigandet:
This PR includes:

- changes to use the Perl XML::XML2JSON module instead of nodejs xml2json module, in order to remove the dependency on nodejs (as suggested by @alexgarel )
- support for GS1 messages that contain nested products (a pallet that contains cartons that contain the actual products sold to consumers)
- updated expected test results: as we might have more than 1 product in a GS1 message, each test now returns an array of product data
- new test cases with sample data from Agena3000 in XML converted by the Perl XML::XML2JSON module


scripts/convert_gs1_xml_to_json_in_dir.js
```diff
@@ -1,12 +1,14 @@
 // This script is used to convert GDSN data from Equadis in XML format
 // to a corresponding JSON structure
 
-const xml2json = require('xml2json')
-const fs = require("fs")
+const process = require('process')
```
alexgarel, line 4: needs a ";"

linting does not pass, see https://github.com/openfoodfacts/openfoodfacts-server/runs/5869772770?check_suite_focus=true#step:5:39

lib/ProductOpener/GS1.pm
```diff
@@ -1208,7 +1235,19 @@ sub gs1_to_off ($$$) {
 				elsif (ref($json_ref->{$source_field}) eq "ARRAY") {
 					foreach my $json_array_entry_ref (@{$json_ref->{$source_field}}) {
 
-						gs1_to_off($source_target, $json_array_entry_ref, $results_ref);
+						# We should have an array of hashes, but in some CodeOnline files we have an array with an empty hash..
```
alexgarel, line 1238:
```suggestion
						# We should have an array of hashes, but in some CodeOnline files we have an array with an empty array..
```

lib/ProductOpener/GS1.pm
```diff
@@ -1217,31 +1256,110 @@ sub gs1_to_off ($$$) {
...
+One GS1 message can include 1 or more products, typically products that contain other products
+(e.g. a pallet of cartons of products).
+
+=head3 Arguments
 
-=head3 Return value
+=head4 $json_ref Reference to a decoded JSON structure
 
-=head4 Reference to a hash of fields
+=head4 Reference to an array of product data
```
alexgarel, line 1340:
```suggestion
=head4 $product_ref - Reference to an array of product data
```

scripts/convert_gs1_xml_to_json_in_dir.pl
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/perl -w
+
```
alexgarel, line 2:
```suggestion

# Perl script to convert XML received from GS1 to JSON files, more easily processed

```

scripts/equadis-import/run_equadis_import.sh
```diff
@@ -12,11 +12,10 @@ find /home/sftp/equadis/data/ -mtime -2 -type f -exec cp {} /srv2/off-pro/equadi
 
 export NPM_CONFIG_PREFIX=~/.npm-global
 
-node equadis-xml2json.js
+node /srv/off-pro/scripts/convert_gs1_xml_to_json_in_dir.js /srv2/off-pro/equadis-data-tmp/
```
alexgarel, line 15: why not use the new .pl here ?

alexgarel (approved):
LGTM

kudos

scripts/equadis-import/run_equadis_import.sh
stephanegigandet: I want to test it out more thoroughly first, to check that we end up with the exact same results once the products are imported into OFF.

scripts/convert_gs1_xml_to_json_in_dir.js
stephanegigandet: fixed

PR #6568, opened 2022-04-07
fix: skip last header line in csv

dipietroR:
Seeing:
```
Argument "Transformation" isn't numeric in addition (+) at /opt/product-opener/lib/ProductOpener/Ecoscore.pm line 139, <$io> line 8.
Argument "Emballage" isn't numeric in addition (+) at /opt/product-opener/lib/ProductOpener/Ecoscore.pm line 139, <$io> line 8.
Argument "Transport" isn't numeric in addition (+) at /opt/product-opener/lib/ProductOpener/Ecoscore.pm line 139, <$io> line 8.
Argument "Supermarch\x{e9} et distribution" isn't numeric in addition (+) at /opt/product-opener/lib/ProductOpener/Ecoscore.pm line 139, <$io> line 8.
```
Errors. This may be from adding 0 to an undefined value rather than using a default.

There was no ecoscore popping up on the products on the local dev environment. I see them on the app. Though I'm not sure how to test there. Or quite what to test on the dev environment, as I don't see the ecoscore there to check that it is still there after this change. This does seem to clear the errors on startup.

### DIFF
--- a/lib/ProductOpener/Ecoscore.pm
+++ b/lib/ProductOpener/Ecoscore.pm
@@ -130,7 +130,8 @@ sub load_agribalyse_data() {
 
 		my $row_ref;
 
-		# Skip 3 first lines
+		# Skip 4 first lines
+		$csv->getline ($io);
 		$csv->getline ($io);
 		$csv->getline ($io);
 		$csv->getline ($io);


lib/ProductOpener/Ecoscore.pm
```diff
@@ -142,20 +142,20 @@ sub load_agribalyse_data() {
...
+				ef_distribution => $row_ref->[12] // 0, # Supermarché et distribution
+				ef_consumption => $row_ref->[13] // 0, # Consommation
+				ef_total => $row_ref->[14] // 0, # Total
+				co2_agriculture => $row_ref->[15] // 0, # Agriculture
+				co2_processing => $row_ref->[16] // 0, # Transformation
+				co2_packaging => $row_ref->[17] // 0, # Emballage
+				co2_transportation => $row_ref->[18] // 0, # Transport
+				co2_distribution => $row_ref->[19] // 0, # Supermarché et distribution
+				co2_consumption => $row_ref->[20] // 0, # Consommation
+				co2_total => $row_ref->[21] // 0, # Total
```
alexgarel, line 158: @stephanegigandet knows better but I would say that the +0 was also intended to cast to int.

So maybe you need  `co2_total => int($row_ref->[21] // 0)` (and for all lines)

alexgarel (changes requested):
This PR is a good idea, but I think you should ensure we have ints.

lib/ProductOpener/Ecoscore.pm
stephanegigandet: Yes, the +0 is to cast the value to a number (not necessarily an integer).

But if you look at the warning:
Argument "Transformation" isn't numeric in addition (+) at /opt/product-opener/lib/ProductOpener/Ecoscore.pm line 139, <$io> line 8.

The issue is not that the value is undefined, it is that it is a string: "Transformation".

My guess is that we are in fact reading one of the header lines while we think it's one of the data lines.

stephanegigandet: You can verify that guess by looking at the file that is being read:

~/openfoodfacts-server$ vi ecoscore/agribalyse/AGRIBALYSE3.0.1_vf.csv.2 

the warning says "line 8"

dipietroR: Ah, thank you for pointing me to where the data is coming from. I looked at the file and noticed that lines 1-4 didn't include any data, but the code only skipped the first three lines. How does this look?

I don't quite understand why it's talking about line 8 though; lines 5+ are data lines. I'm not sure why line 8 is special.

stephanegigandet: well I think the CSV file has 4 header rows (and not 3), but the CSV file maps those rows to 8 lines, there must be escaped carriage returns or something

stephanegigandet (approved):

lib/ProductOpener/Ecoscore.pm
dipietroR: Right, did my second commit get the fourth header line dropped? I can squash my commits if it looks alright.

Interesting libre office is seeing the newlines differently than bash. The headers do end at 8 in bash, but 4 in libre. Thank you for the pointer!

dipietroR: Oh I see the commit was approved. Thank you! I'll merge after the tests finish

alexgarel (approved):
Cool

Issue #6332, opened 2022-01-22
Make the "Unselect image" button translatable


teolemon:
### What
Make the "Unselect image" button translatable

### Code Pointer
- https://github.com/openfoodfacts/openfoodfacts-server/blob/0092e2e0a119290742574bf7e8de8d52d58ea6d9/lib/ProductOpener/Display.pm#L7467

### Part of
- #5528

yuktea:
Fixed in PR #6570

PR #6570, opened 2022-04-07
fix: Makes "Unselect Image" button translatable

yuktea:
This PR was made to fix Issue #6332

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -5482,7 +5482,7 @@ msgstr "The Eco-Score can only be calculated if the product has a sufficiently p
 msgctxt "ecoscore_add_more_precise_category"
 msgid "You can modify the product page to add a more precise category."
 msgstr "You can modify the product page to add a more precise category."
-      
+
 msgctxt "ecoscore_platform_promo"
 msgid "If you are the manufacturer of this product, you can send us the information with our <a href=\"https://world.pro.openfoodfacts.org\">free platform for producers</a>."
 msgstr "If you are the manufacturer of this product, you can send us the information with our <a href=\"https://world.pro.openfoodfacts.org\">free platform for producers</a>."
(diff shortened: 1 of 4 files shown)


po/common/common.pot
```diff
@@ -6069,4 +6069,8 @@ msgstr "The ingredients of the product must be specified in order to determine t
 
 msgctxt "nova_group_too_many_unknown_ingredient"
 msgid "We could not recognize some of the ingredients and determine the NOVA group."
-msgstr "We could not recognize some of the ingredients and determine the NOVA group."
\ No newline at end of file
+msgstr "We could not recognize some of the ingredients and determine the NOVA group."
+
+msgctxt "unselect_image"
+msgid "Unselect Image"
+msgstr ""
```
alexgarel, line 6076: @stephanegigandet to have the message at least in english if translation isn't available in target language,
 isn't it better to repeat msgid in msgstr, or is there an automatic fallback to en.po ?

alexgarel (approved):
LGTM, just added a comment to @stephanegigandet on a topic I do not master !

po/common/common.pot
stephanegigandet: Product Opener does not use common.pot at all in fact, it reads only the *.po files. The common.pot is used by Crowdin.

In practice, we probably should have only one file, and configure crowdin to use en.po instead of common.pot
@teolemon is the master of the crowdin configuration

teolemon: This is related to GNU GetText.
per phrase.com:
POT, short for “portable object template”, is a format generated by GNU gettext to streamline the [software localization](https://phrase.com/blog/localization-and-translation-glossary/localization-l10n/) and [internationalization](https://phrase.com/blog/localization-and-translation-glossary/internationalization-i18n/). While source strings are placed after “msgid”, their translations are placed after “msgtr”. 

Note: POT files can sometimes be [PO files](https://help.phrase.com/help/gettext-po-files) as well. These two files are mostly identical except that PO files are generated by calling “msginit” in CMD. It is possible to perform the [translation](https://phrase.com/blog/localization-and-translation-glossary/translation-t9n/) on a POT file first and then rename it to a PO. The final localized format should be a machine-readable [“.mo” file](https://help.phrase.com/help/gettext-compiled-mo-files) generated from the PO file. The “.mo” file can be generated by calling “msgfmt” in CMD.

**In practise, Product Opener does not support .pot or .mo files**

yuktea:
> @yuktea : I added a commit to add translation in Hindi, so that we can check that it is working.
> 
> After that you have to run make build_land and make restart, then put the interface in Hindi (using the world-hi.openfoodfacts.localhost domain for instance)

Thank you for testing this out, @stephanegigandet ! It's a mystery why it wouldn't work for me despite doing everything right (the exact same changes+ `make build_lang` + `make restart`. I'll try to look into it when I'm back

PR #6578, opened 2022-04-09
fix: Email address with space

### LINKED ISSUE #4549: Email address with space
It's possible to create a new user account which email address contains spaces.
Eg. `bibi @fricotin.org`

aadarsh-ram:
### Description

It's possible to create a new user account with an email address that contains spaces.
- Fixes #4549
- Modified email was not updated to variable after email validation.

### DIFF
--- a/lib/ProductOpener/Users.pm
+++ b/lib/ProductOpener/Users.pm
@@ -348,6 +348,9 @@ sub check_user_form($$$) {
 	if (not $address) {
 		push @{$errors_ref}, $Lang{error_invalid_email}{$lang};
 	}
+	
+	# If all checks passed, reinitialize with modified email
+	$user_ref->{email} = Email::Valid->address($user_ref->{email});
 
 	if ($type eq 'add') {
 
--- a/lib/ProductOpener/Users.pm
+++ b/lib/ProductOpener/Users.pm
@@ -348,9 +348,10 @@ sub check_user_form($$$) {
 	if (not $address) {
 		push @{$errors_ref}, $Lang{error_invalid_email}{$lang};
 	}
-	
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Users.pm
```diff
@@ -348,6 +348,9 @@ sub check_user_form($$$) {
 	if (not $address) {
 		push @{$errors_ref}, $Lang{error_invalid_email}{$lang};
 	}
+	
+	# If all checks passed, reinitialize with modified email
+	$user_ref->{email} = Email::Valid->address($user_ref->{email});
```
stephanegigandet, line 353: I think it would be better to do that in a else clause and use the $address which already has the result of Email::Valid->address( -address => $user_ref->{email}, -mxcheck => 1 );

stephanegigandet (changes requested):
Looks good, but we can use the modified email already in $address

lib/ProductOpener/Users.pm
aadarsh-ram: @stephanegigandet The $address variable becomes a boolean, if a "$@" condition is satisfied. So, I didn't want to disturb that check. Instead, I reinitalized the variable separately.

aadarsh-ram: @stephanegigandet I just realized that now. Really sorry for not noticing it earlier! 
Did a new commit with the else clause and reinitalization with $address variable.

stephanegigandet (approved):
Thank you!

lib/ProductOpener/Users.pm
stephanegigandet: Perfect, thanks!

Issue #6612, opened 2022-04-13
Ingredients percent analysis sub-optimal for missing % that are not the first or last ingredient


stephanegigandet:
### Describe the bug

For this ingredient list, we should be able to compute the percent in the middle: "Jus de pomme (57,3%), jus de carotte, jus de gingembre (2,5%)."

We find the right percent_max, but not the percent_min (which should be the same)

```
			{
				'id' => 'en:apple-juice',
				'percent' => '57.3',
				'percent_estimate' => '57.3',
				'percent_max' => '57.3',
				'percent_min' => '57.3',
				'text' => 'Jus de pomme'
			},
			{
				'id' => 'en:carrot-juice',
				'percent_estimate' => '32.025',
				'percent_max' => '42.7',
				'percent_min' => '21.35',
				'text' => 'jus de carotte'
			},
			{
				'id' => 'en:ginger',
				'percent' => '2.5',
				'percent_estimate' => '10.675',
				'percent_max' => '2.5',
				'percent_min' => '2.5',
				'processing' => 'en:juice',
				'text' => 'gingembre'
			}
```

### To Reproduce

https://fr.pro.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=Jus+de+pomme*+%2857%2C3%25%29%2C+jus+de+carotte*%2C+jus+de+gingembre*+%282%2C5%25%29.&type=add&action=process&submit=Envoyer

### Expected behavior

percent_min should be 42.7

### Type of device

Browser

stephanegigandet:
fixed

PR #6614, opened 2022-04-13
fix: more precise estimate of ingredients percents min and max

stephanegigandet:
fix for #6612

This change 2 rules to compute percent min and percent max of ingredients based on the other ingredients min and max.

You can see the difference of result in this commit as I first added a test that demonstrated the issue: https://github.com/openfoodfacts/openfoodfacts-server/commit/cba72e31ae084117f42dd8506dd9ff64c6249613

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -2462,7 +2462,7 @@ sub set_percent_max_values($$$) {
 		}
 		my $max_percent_max = $total_max - $sum_of_mins_before - $sum_of_mins_after;
 
-		if ($ingredient_ref->{percent_max} > $max_percent_max) {
+		if (($max_percent_max >= 0) and ($ingredient_ref->{percent_max} > $max_percent_max)) {
 			$ingredient_ref->{percent_max} = $max_percent_max;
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -2454,10 +2454,16 @@ sub set_percent_max_values($$$) {
...
 
-		if ($ingredient_ref->{percent_max} > $total_max - $sum_of_mins_before) {
-			$ingredient_ref->{percent_max} = $total_max - $sum_of_mins_before;
+		my $sum_of_mins_after = 0;
+		for (my $j = $i; $j < $n; $j++) {
+			$sum_of_mins_after += $ingredients_ref->[$j]{percent_min};
+		}
+		my $max_percent_max = $total_max - $sum_of_mins_before - $sum_of_mins_after;
+
+		if ($ingredient_ref->{percent_max} > $max_percent_max) {
```
alexgarel, line 2465: Shouldn't we first check that $max_percent_max is non negative (a bit more robust)

alexgarel (approved):
LGTM, just a small suggestion.

PR #6618, opened 2022-04-14
build: use local robotoff if exists for dev

alexgarel:
I never checked using robotoff with local instance. Robotoff side was already configured, but not product opener side.
This change makes it happens.


.env
```diff
@@ -36,7 +36,7 @@ MONGODB_HOST=mongodb
 MONGODB_CACHE_SIZE=8 # GB
 MONGO_INITDB_ROOT_USERNAME=root
 MONGO_INITDB_ROOT_PASSWORD=test
-ROBOTOFF_URL=http://host.docker.internal:5500 # connect to Robotoff running in separate docker-compose deployment
+ROBOTOFF_URL=http://api:5500 # connect to Robotoff running in separate docker-compose deployment
```
stephanegigandet, line 39: What is "api" exactly? an hostname that is available from the po_backend container / all product opener containers / all running docker containers? Where is it defined? is it just an equivalent to "localhost" and the robotoff container expose the port 5500 on the local machine?

Could we rename "api" to "robotoff" to make it clearer?

alexgarel: api is the name of the service (in docker-compose.yml) this is the way docker-compose works (it gives name of the service to container in the network). We should not change that.

The way we can reach it is that we say to robotoff that it's network is the same as productopener (all that meaning the default names for developers).

stephanegigandet: maybe robotoff-api could be a good name then, as we may have a separate product-opener-api service some day.

alexgarel: yes but this is to be changed on robotoff side, and I don't know what it will imply…

alexgarel: It's only a settings for dev ! I think we can keep it for now …

stephanegigandet (approved):
Looks good

PR #6619, opened 2022-04-15
fix: Wikidata entries to update

### LINKED ISSUE #6530: Wikidata entries to update
### What

- Wikidata entries get merged overtime
- This generates exceptions in the old Android app, but more importantly we need to keep taxonomies up to date
-  Q63534992
- Q2366206

aadarsh-ram:
### Description

Wkidata entries get merged overtime. So, exceptions were generated in the Android app due to this merge. 
Affected wikidata due to merge:
- Q63534992
- Q2366206

- Fixes #6530 
- Updated additives.txt and categories.txt taxonomies with new values.

### DIFF
--- a/.github/workflows/github-projects.yml
+++ b/.github/workflows/github-projects.yml
@@ -15,5 +15,5 @@ jobs:
       - uses: actions/add-to-project@main
         with:
           project-url: https://github.com/orgs/openfoodfacts/projects/12
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
           label-operator: AND
--- a/.github/workflows/github-projects.yml
+++ b/.github/workflows/github-projects.yml
@@ -15,5 +15,5 @@ jobs:
       - uses: actions/add-to-project@main
         with:
           project-url: https://github.com/orgs/openfoodfacts/projects/12
-          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
           label-operator: AND
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ We have [[templatized](https://github.com/openfoodfacts/openfoodfacts-server/tre
 ## Contributors
 
 This project exists thanks to all the people who contribute.
-<a href="https://github.com/openfoodfacts/openfoodfacts-server/graphs/contributors"><img src="https://opencollective.com/openfoodfacts-server/contributors.svg?width=890&button=false" /></a>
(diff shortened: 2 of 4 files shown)


stephanegigandet (approved):
Thanks!

.github/workflows/github-projects.yml
```diff
@@ -15,5 +15,5 @@ jobs:
       - uses: actions/add-to-project@main
         with:
           project-url: https://github.com/orgs/openfoodfacts/projects/12
-          github-token: ${{ secrets.ADD_TO_PROJECT_PAT }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
alexgarel, line 18: Why did you change that @aadarsh-ram ?

And why did @teolemon us a PAT (I imagine it is because this action can trigger other actions ?) 
see https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow (in this case we can add a comment to make it clearer)

alexgarel (changes requested):
Thanks for the fix. Just a question on you first (or second) commit.

.github/workflows/github-projects.yml
aadarsh-ram: Yep, so it for some reason failed for my first commit.

aadarsh-ram: So, since the GITHUB_TOKEN already has the required permissions of ADD_TO_PROJECT_PAT, I used that.
But, for my second commit, the action didn't run. Hence, I have no idea if it worked or not.

alexgarel: The problem we have is that we hit the limit of github projects, it's not related to the token.

In fact I think that we should remove this action for now @teolemon ? (or maybe remove old enough bug from the project to gain space in the project ?)

aadarsh-ram: @alexgarel @stephanegigandet @teolemon This was the failed workflow run: https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/2171303073
Since the error was related to a token, I wanted to see if it would work for the GITHUB_TOKEN.

teolemon: I archived fixed issues, I added all new issues, and we're below the threshold.

aadarsh-ram: For some reason, I'm not able to recreate the workflow error, since the issue has already been added to the project I think. 
So, I brought things back to how it was.

aadarsh-ram:
@teolemon @stephanegigandet @alexgarel I noticed that the contributor images weren't being updated. It seems there is an issue in the Open Collective's contributors.svg generation. So, I added a new image generator which is also open-source. 
Hope this is okay!

alexgarel (approved):

Issue #6537, opened 2022-04-01
Import GS1 product data from Agena3000


stephanegigandet:
### Is your feature request related to a problem? Please describe.

Agena3000 is a GS1 datapool (similar to Equadis) which is going to allow its customers (manufacturers) to send us product data in the GS1 format.

### Describe the solution you'd like

We already have some code to process GS1 data for Equadis and CodeOnline, but there maybe subtle differences in the format and/or the data.

Compared to other GS1 imports, we will need to generate confirmation messages in the GS1 format (CIC)

We also will need to implement some plumbing to automate the import.

PR #6630, opened 2022-04-15
feat: add support for 2 GS1 quantityContained field in nutrientDetail

stephanegigandet:
part of #6537

needed as Equadis and Agena3000 have different ways of passing the values for energy in both kj and kcal


alexgarel (approved):
:+1:

PR #6631, opened 2022-04-15
fix: limit usernames to 20 characters

stephanegigandet:


lib/ProductOpener/Users.pm
```diff
@@ -380,6 +380,9 @@ sub check_user_form($$$) {
 		elsif ($user_ref->{userid} !~ /^[a-z0-9]+[a-z0-9\-]*[a-z0-9]+$/) {
 			push @{$errors_ref}, $Lang{error_invalid_username}{$lang};
 		}
+		elsif (length($user_ref->{userid}) > 20) {
+			push @{$errors_ref}, $Lang{error_username_too_long}{$lang};
+		}
```
alexgarel, line 385: Being at it, can't we also add a check on username length < 60 characters for example (to avoid spam)

alexgarel (approved):
LGTM but I would like you to add a test on user name.

lib/ProductOpener/Users.pm
stephanegigandet: @alexgarel I added the check.

alexgarel (approved):
LGTM + testé ok

PR #6633, opened 2022-04-16
fix:  warning message

dipietroR:
Fixes "Use of uninitialized value" warning. This value didn't use a test for if it was defined but the other uses of it had a test for definition, so I added one here as well.


stephanegigandet (approved):
Looks good, thank you!

Issue #6634, opened 2022-04-16
Recognize "Dry" as an ingredient processing type synonymous to "Dried"


yuktea:
### Describe the bug

A considerable number of products used the prefix "dry" instead of "dried" when indicating the processing type.
Ingredients are not recognized correctly when "dry" is used.
Example: `Dry Mango` is not recognized as an ingredient but `Dried Mango` is
 

### To Reproduce

https://us.openfoodfacts.org/ingredients?filter=dry

### Expected behavior

We should recognize include `dry` as a secondary word for `dried`

ditsuke:
Closed in #6636

PR #6636, opened 2022-04-17
fix: "Dry" not being recognized as a processing type

yuktea:
### Description

- Added a test that would verify when the fix has been implemented.
- Added dry as a secondary word for dried

### Screenshot
image

- Test for #6634


stephanegigandet (approved):
Looks good, thank you!

PR #6637, opened 2022-04-17
fix: Add check for the definition of $user_ref->{org}

yuktea:
### Description
Implements an additional check to see if an org associated with the user is defined before executing the code block. 
This fixes an uninitialized value warning.

#### The error
`[Sun Apr 17 11:00:58 2022] -e: Use of uninitialized value in concatenation (.) or string at /opt/product-opener/cgi/user.pl line 208.
`

### DIFF
--- a/cgi/user.pl
+++ b/cgi/user.pl
@@ -196,28 +196,28 @@ if ($action eq 'display') {
 
 		# Professional account
 		push @{$template_data_ref->{sections}}, {
-				id          => "professional",
-				name        => lang("pro_account"),
-				description => "if_you_work_for_a_producer",
-				note        => "producers_platform_description_long",
-				fields      => [
-					{
-						field   => "pro",
-						type    => "checkbox",
-						label   => lang("this_is_a_pro_account"),
-						value   => "off",
-					},
-					{
-						field => "pro_checkbox",
-						type  => "hidden",
-						value => 1,
-					},
-					{
-						field => "requested_org",
-						label => lang("producer_or_brand") . ":",
-					}
-				]
-			};
+			id => "professional",
+			name => lang("pro_account"),
+			description => "if_you_work_for_a_producer",
+			note => "producers_platform_description_long",
+			fields => [
+				{
+					field => "pro",
+					type => "checkbox",
+					label => lang("this_is_a_pro_account"),
+					value => "off",
+				},
+				{
+					field => "pro_checkbox",
+					type => "hidden",
+					value => 1,
+				},
+				{
+					field => "requested_org",
+					label => lang("producer_or_brand") . ":",
+				}
+			]
+		};
 
 		# Teams section
 		# Do not display teams if it is a professional account
--- a/cgi/user.pl
+++ b/cgi/user.pl
@@ -195,8 +195,7 @@ if ($action eq 'display') {
 		};
 
 		# Professional account
-		if (defined $user_ref->{org}) {
-			push @{$template_data_ref->{sections}}, {
+		push @{$template_data_ref->{sections}}, {
 				id          => "professional",
 				name        => lang("pro_account"),
 				description => "if_you_work_for_a_producer",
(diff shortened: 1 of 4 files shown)


cgi/user.pl
```diff
@@ -195,30 +195,32 @@
...
-					type => "hidden",
-					value => 1,
-				},
-				{
-					field => "requested_org",
-					label => lang("producer_or_brand") . ":",
-				}
-			]
-		};
+		if (defined $user_ref->{org}) {
```
stephanegigandet, line 198: So that definitely will solve the warning, but then it will prevent users to create a professional account. When users create an account, $user_ref->{org} is not defined (yet), but we want to display them the checkbox option to say that it is a professional account.

The warning field should probably be removed, and instead, we can change the template to display the warning message, only  when needed.

Template is here : /templates/web/pages/user_form/user_form_page.tt.html

This part should be changed:
```

                    [% IF accepted_organization && section.id == "professional" %]
                        <p>[% field.warning %]</p>
                    [% ELSE %]
                        <label for="[% field.field %]">
                            <input type="checkbox" id="[% field.field %]" name="[% field.field %]" [% IF field.value == 'on' %]checked="checked"[% END %] />
                            [% field.label %]
                        </label>
                    [% END %]
```

the Lang.pm defines f_lang() and lang_sprintf() functions that can be used directly in the template.

stephanegigandet (changes requested):

cgi/user.pl
yuktea: Thank you for reviewing! This is certainly a great approach to fix that warning.
I've pushed another commit with an update :)

cgi/user.pl
```diff
@@ -196,29 +196,28 @@
...
-					type => "hidden",
-					value => 1,
-				},
-				{
-					field => "requested_org",
-					label => lang("producer_or_brand") . ":",
-				}
-			]
-		};
+				id          => "professional",
```
ditsuke, line 199: Could you revert the formatting to minimize the diff? I believe its because hash separators are aligned, might be helpful to adjust your editor's formatting settings.

yuktea: Thank you @ditsuke , I noticed that too!

templates/web/pages/user_form/user_form_page.tt.html
```diff
@@ -82,8 +82,8 @@
                     <label for="[% field.field %]">	[% field.label %]</label>
                     <input type="email" id="[% field.field %]" name="[% field.field %]" value = "[% field.value %]" autocomplete="email"/>
                 [% ELSIF field.type == 'checkbox' %]
-                    [% IF accepted_organization && section.id == "professional" %]
-                        <p>[% field.warning %]</p>
+                    [% IF accepted_organization && section.id == "professional" && user_org.defined %]
+                        <p>[% f_lang('this_is_a_pro_account_for_org', {'org' => user_org} ) %]</p>
```
stephanegigandet, line 86: Hi @yuktea , have you checked if it works? (you can do that by creating an account with the "this is a professional account" checkbox, and then editing the account).
I think it won't work as is, because f_lang() expects a different format than sprintf  (%s vs something like <org>).
You can look at the strings in po/common/common.pot and po/common.en.po
The best solution is indeed to use f_lang(), but then you have to change the string. 

e.g. here is an example for another string:

# variable names between { } must not be translated
msgctxt "f_move_data_and_photos_to_main_language"
msgid "Move all data and selected photos in {language} to the main language of the product: {main_language}"
msgstr "Move all data and selected photos in {language} to the main language of the product: {main_language}"

stephanegigandet (approved):
Looks good!

There's a line that I think is not used (and was not used before):

$template_data_ref->{pro_account_org} = sprintf(lang("this_is_a_pro_account_for_org"),"<b>" . $user_ref->{org} . "</b>");

stephanegigandet:
Could you also remove the old string "this_is_a_pro_account_for_org" from the common.pot file?

stephanegigandet (approved):
Thank you!

yuktea:
> Looks good!
> 
> There's a line that I think is not used (and was not used before):
> 
> $template_data_ref->{pro_account_org} = sprintf(lang("this_is_a_pro_account_for_org"),"" . $user_ref->{org} . "");

Yes! was thinking the same. I'll clean this up

stephanegigandet (approved):

PR #6654, opened 2022-04-20
fix: move h1 tags to template

stephanegigandet:
Small changes to put the generation of` <h1>` title tags for all pages in templates.


lib/ProductOpener/Display.pm
```diff
@@ -1434,10 +1433,10 @@ sub display_text($)
 	# wikipedia style links [url text]
 	$html =~ s/\[(http\S*?) ([^\]]+)\]/<a href="$1">$2<\/a>/g;
 
-
+	# Remove the title from the content to put it in the title field
 	if ($html =~ /<h1>(.*)<\/h1>/) {
```
alexgarel, line 1437: Make a non greedy expression (use ?) to handle case of multiple h1

alexgarel (approved):
LGTM but a small improvement.

PR #6655, opened 2022-04-20
[WIP] experiment(data-quality): Disallow emails in product fields

### LINKED ISSUE #6539: Prevent people from entering email addresses as values
### What
- Prevent people from entering email addresses as values
### Screenshots
Capture d’écran 2022-04-01 à 18 27 54

### Part of
- #5538

yuktea:
Closes  #6539 

## Information
This PR aims to disallow email addressed fields belonging to products new or existing. I've made an effort in accommodating the logic into, what I think is our existing validation layer. However I'm yet to figure out where we're setting up the other validation rules (if there are any others yet). for this I'm attempting to set up a remote debugger which would let me introspect and figure it out better. Apart from that, some problems I could identify with this approach:
- [ ] No client-side validation and indication.
- [ ] The form is not preserved on submission of invalid information, and there's no information on why the form failed.

I'd like reviewers to give feedback on the approach any insights into how we validate data and validation taken care of by `Product::process_product_edit_rules`.


stephanegigandet:
This would work, but as you suggest, we probably should think more about this, to decide what would be the right behaviour:

1. reject the changes completely if there is an error in one of the fields
2. ignore the fields that have errors

And how to convey this information to users so that they can correct the fields.

It's also complexified a bit by the fact that those changes can be made through the OFF website (and app), but also through third parties that use our API to add / edit products. It's relatively easy for us to introduce new behaviours on the website (such as redisplaying the form with messages about errors), but for the API, we need to find a solution that does not break existing applications that use the API.

yuktea:
Hi folks! I'm so sorry I've not been able to make progress on this PR yet. I've been occupied with work commitments and University since my academic session is ending next week (which means I'll be taking my final exams this week), but I expect to continue work on this by next week. Thanks for the support and patience!

cc: @stephanegigandet

stephanegigandet:
Hi @yuktea ,thank you and  good luck with your exams!

stephanegigandet:
Done in https://github.com/openfoodfacts/openfoodfacts-server/pull/8452

PR #6656, opened 2022-04-21
fix: undefined variable warning

dipietroR:
This fixes "Use of uninitialized value" warning.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -4952,7 +4952,7 @@ sub search_and_display_products($$$$$) {
 	push @{$template_data_ref->{sort_options}}, { value => "last_modified_t", link => $request_ref->{current_link} . "?sort_by=last_modified_t", name => lang("sort_by_last_modified_t") };
 
 	my $count;
-	my $page_count;
+	my $page_count = 0;
 
 	my $fields_ref;
 


stephanegigandet (changes requested):
Thanks for the PR!

Your change would suppress the warning, but what we need to do is to investigate the root cause for the warning: in which circumstances is $page_count not defined? Is it a sign that there is an error or some broken logic somewhere?

If you look at where $page_count is initialized, modified etc. you can find a way to solve the root issue. That's what we really want to solve.

dipietroR:
Looking closer, I'm not sure how $page_count can be undefined. I see it defined at:
https://github.com/openfoodfacts/openfoodfacts-server/blob/b3b482b931b3f4f6fe5731011ce763f4591a2550/lib/ProductOpener/Display.pm#L4955
And the line that I was seeing the warning is in the same scope, I'm not seeing anything that undefines it. Leaving me confused as to what is happening. That I also can no longer reproduce the warning message, I'm starting to wonder if I imagined it :p

Is there a situation where $page_count can be undefined on
https://github.com/openfoodfacts/openfoodfacts-server/blob/b3b482b931b3f4f6fe5731011ce763f4591a2550/lib/ProductOpener/Display.pm#L5126
?
If it indeed can't be defined, then this PR should be closed.

stephanegigandet:
my $page_count;   --> This creates the variable, with an undefined value.

Then later we may increment that undefined value   (with $page_count++ or something like that). 

The right solution is to set the $page_count variable to 0 when we create it, so that we can increment it later, without triggering a warning (undef is equal to 0 when used in a number context, but it triggers a warning)

dipietroR:
Oh! I didn't know that about definition. I thought once you had a "my $variable" bit it was simply defined! I hadn't considered that it may be defined, but it has an undefined variable. This is new and interesting! Thank you for pointing it out! I'm going to look up more about variable definition to see if I can understand it a little better.

stephanegigandet (approved):
Thank you!

PR #6658, opened 2022-04-21
feat: add userid to emails' export

CharlesNepote:
Add userid to export


stephanegigandet (approved):
Looks good. Note that it will only output users that have checked the newsletter checkbox (which may be a good thing).

PR #6663, opened 2022-04-21
ci: Tweak Perl Critic for annotations

teolemon:
### Checklist
ci: Tweak Perl Critic for annotations

### DIFF
--- a/.github/workflows/perlcritic.yml
+++ b/.github/workflows/perlcritic.yml
@@ -14,8 +14,8 @@ jobs:
     - uses: actions/checkout@v3
 
     - name: "Run Perl::Critic" 
-      uses: Difegue/action-perlcritic@2.0.0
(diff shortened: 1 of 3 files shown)


alexgarel:
I don't think we want that (@stephanegigandet ?)

We will have [37000 annotation](https://api.github.com/repos/openfoodfacts/openfoodfacts-server/check-runs/6117963778)… it will clutter the reviews (as codecov is cluttering robotoff reviews…)

Why would you want to go to "brutal ?" @teolemon

alexgarel (changes requested):
see my comment

stephanegigandet:
Where/when would those annotations be shown? They don't prevent the test from passing, so I don't really understand how they would be used?

dipietroR:
I think that PR #7024 solves this issue

Issue #5559, opened 2021-08-27
📝 Documenting Product Opener (Tracker)


teolemon:
### Documenting Product Opener
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/2203
  - [ ] Please add a list here of what's left to do.
- [x] #5568
- [ ] #7447
- [ ] #7472

### Part of 
- #5529

YashasviChaurasia:
@teolemon I would like to work on this issue.

alexgarel:
@YashasviChaurasia sorry it's now that I see your message. You are very welcome to help :-) (sorry for the delay)
Maybe you should come on #productopener channel on [slack](https://slack.openfoodfacts.org/) to discuss it.

PR #6664, opened 2022-04-21
docs: Document the I18N module using perl-pod format

yuktea:
### Description

Documented the `ProductOpener::I18N.pm` file in `perlpod` format.

- Related Issue https://github.com/openfoodfacts/openfoodfacts-server/issues/2203
- Part of https://github.com/openfoodfacts/openfoodfacts-server/issues/5559

### DIFF
--- a/lib/ProductOpener/I18N.pm
+++ b/lib/ProductOpener/I18N.pm
@@ -18,6 +18,24 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+=head1 NAME
+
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/I18N.pm
```diff
@@ -120,6 +157,24 @@ sub read_po_files {
 # Separate the singular & plural entries from a hash, as returned by
 # read_po_files(), into two hashes. Obviously returned as two hashrefs.
 #
+
+=head1 FUNCTIONS
```
stephanegigandet, line 161: I think we can remove this header, as there is already another =head1 FUNCTIONS header above

PR #6669, opened 2022-04-22
fix: URL linked to $request_ref->{world_current_link} & $world_subdomain

### LINKED ISSUE #1437: Keep the same UI language after clicking showing products from the whole world
**Summary:**
https://de.openfoodfacts.org changes to https://world.openfoodfacts.org, but better would be https://world-de.openfoodfacts.org

**Steps to reproduce:**
- Open https://de.openfoodfacts.org
- Click on "Entsprechende Produkte aus der ganzen Welt ansehen" (Show respective products from the whole world)

**Expected behavior:**
It should show international products with labels in the language previously chosen (supposedly for good reason).

**Observed behavior:**
It shows international products, but with English labels.

pradumna-gautam:
### Description

Fix https://github.com/openfoodfacts/openfoodfacts-server/issues/1437
Changed the URL linked to $request_ref->{world_current_link}
and $world_subdomain in file Display.pm

- Fixes #1437


lib/ProductOpener/Display.pm
```diff
@@ -307,7 +307,7 @@ use vars qw();
 
 $static_subdomain = format_subdomain('static');
 $images_subdomain = format_subdomain('images');
-$world_subdomain = format_subdomain('world');
+$world_subdomain = format_subdomain("world-" . $lc);
```
alexgarel, line 310: I propose that if $lc is "en" we just put "world"

pradumna-gautam:
I'm thinking of solving this issue this way:-
We can have a sub 'get_world_subdomain' instead of using a variable, and everywhere we use $world_subdomain, we can replace it with a function call.
So, I wrote the sub like this, and called the function as follows:-
$get_world_subdomain()
Is this approach correct?
Also, I replaced all $world_subdomain with $get_world_subdomain in [Display.pm]

lib/ProductOpener/Display.pm
```diff
@@ -5189,7 +5194,7 @@ sub search_and_display_products($$$$$) {
 
 	$template_data_ref->{jqm} = param("jqm");
 	$template_data_ref->{country} = $country;
-	$template_data_ref->{world_subdomain} = $world_subdomain;
+	$template_data_ref->{world_subdomain} = $get_world_subdomain;
```
alexgarel, line 5197: Sorry I misguided you I think, but you should have tested ;-) 

```suggestion
	$template_data_ref->{world_subdomain} = get_world_subdomain();
```

lib/ProductOpener/Display.pm
```diff
@@ -7465,7 +7470,7 @@ CSS
 	$request_ref->{canon_url} = product_url($product_ref);
 
 	if ($lc eq 'en') {
-		$request_ref->{canon_url} = $world_subdomain . product_url($product_ref);
+		$request_ref->{canon_url} = $get_world_subdomain . product_url($product_ref);
```
alexgarel, line 7473:
```suggestion
		$request_ref->{canon_url} = get_world_subdomain() . product_url($product_ref);
```

lib/ProductOpener/Display.pm
```diff
@@ -114,7 +114,7 @@ BEGIN
 		$formatted_subdomain
 		$images_subdomain
 		$static_subdomain
-		$world_subdomain
+		$get_world_subdomain
```
alexgarel, line 117: You have to move that in the sub section, and use &get_worldsubdomain.

alexgarel (approved):
:+1:

PR #6672, opened 2022-04-22
fix: switching languages while staying on the same product

### LINKED ISSUE #289: No easy way to switch languages while staying on the same product
There should be a list of available language codes for the product.

pradumna-gautam:
### Description

Fixed #289
Switching languages while staying on the same product
For products having fields in more than 1 language, display a drop-down to display the product field values in the other languages.

Steps
Added a language parameter that takes precedence on the lc parameter (the language of the interface)
Display a dropdown with links to the product in other languages.

File Change in 
Display.pm

- Fixes #289


stephanegigandet:
Hello,

PR #6695, opened 2022-04-26
feat: start of template for tags

stephanegigandet:
This takes a bit (but absolutely not all) of HTML generation from the Perl code to put it in a template.

This is for pages that are list of products, or list of tags.  (e.g /category/cakes  ,  /category/cakes/label/organic  and /cakes )


alexgarel (approved):
Great.

PR #6710, opened 2022-04-27
ci: autoassign PRs

teolemon:
### What
- Auto-assign PRs


alexgarel (approved):
I have nothing against it.

But why is this important ? (there is the notion of owner already). For the project view ?

teolemon:
It shows avatars on the pull request list, which makes it more human, and also PR can be abandoned or taken over/collaboratively, so assignment can be reattributed @alexgarel

PR #6712, opened 2022-04-28
feat: google anaytics 4 and matomo for OFF

alexgarel:
### Checklist

- use GA4 with as much privacy as possible
- setup Matomo to eventually replace GA4

### DIFF
--- a/lib/ProductOpener/Config_off.pm
+++ b/lib/ProductOpener/Config_off.pm
@@ -357,13 +357,13 @@ $page_size = 24;
 
 $google_analytics = <<HTML
 <!-- Global site tag (gtag.js) - Google Analytics -->
-<script async src="https://www.googletagmanager.com/gtag/js?id=G-HQX9SYHB2P"></script>
+<script async src="https://www.googletagmanager.com/gtag/js?id=G-HQX9SYHB2P&aip=1"></script>
 <script>
   window.dataLayer = window.dataLayer || [];
   function gtag(){dataLayer.push(arguments);}
   gtag('js', new Date());
 
-  gtag('config', 'G-HQX9SYHB2P');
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Config_off.pm
```diff
@@ -357,14 +357,35 @@ $page_size = 24;
 
 $google_analytics = <<HTML
 <!-- Global site tag (gtag.js) - Google Analytics -->
-<script async src="https://www.googletagmanager.com/gtag/js?id=UA-31851927-1"></script>
+<script async src="https://www.googletagmanager.com/gtag/js?id=G-HQX9SYHB2P&aip=1"></script>
```
stephanegigandet, line 360: It's not possible to keep the same id so that we have continuous stats?

alexgarel: No, this is a requirement of GA4
Moreover GA4 drastically decrease retention period to 2 monthes (but we could set 4 monthes) (@teolemon you are aware of it right ?)

stephanegigandet: ah ok, well if it retains data for only a few months anyway, then this point becomes moot.

but that means that we can't even have very aggregated metrics across years? e.g. getting a graph with monthly users and page views over a period of several years?

stephanegigandet: It seems it's only for exploration: if you want to slice and dice data accross users: https://support.google.com/analytics/answer/7667196?hl=fr

But the summary metrics are kept.

stephanegigandet: ok, anyway there's no choice, so let's go for it

stephanegigandet (approved):

PR #6724, opened 2022-05-01
fix: tests if variable defined before use

dipietroR:
On the main page I was seeing the following warning:
```
[Sun May  1 16:45:15 2022] -e: Use of uninitialized value $current_link_query in concatenation (.) or string at /opt/product-opener/lib/ProductOpener/Display.pm line 5411.
```
And on the search page I was seeing:
```
[Sun May  1 16:54:12 2022] -e: Use of uninitialized value $current_link in concatenation (.) or string at /opt/product-opener/lib/ProductOpener/Display.pm line 5411.
```
This seems to be because they weren't being tested for definition, which was happening later on in the code.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -5408,7 +5408,7 @@ sub display_pagination($$$$) {
 	}
 	my $current_link_query = $request_ref->{current_link_query};
 
-	$log->info("current link: $current_link, current_link_query: $current_link_query") if $log->is_info();
+	$log->info("current link and current link query", { current_link => $current_link, current_link_query => $current_link_query }) if $log->is_info();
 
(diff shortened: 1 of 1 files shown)


stephanegigandet (changes requested):
That would work, but we won't log the info that $current_link is undefined.

There's a way to log multiple variables in a structured way. Here is an example from elsewhere in the code:

$log->info("displaying text from file", { cc => $cc, lc => $lc, lang => $lang, textid => $textid, textlang => $text_lang, file => $file }) if $log->is_info();

So instead of having $current_link and $current_link_query inside the debug text, we can pass them in the 2nd parameter.

Could you use that instead? Thank you!

stephanegigandet (approved):
Thank you very much @dipietroR !

PR #6734, opened 2022-05-04
chore: Vegetarian sauces

moon-rabbitOFF:


aleene (commented):
We are slowly getting a problem with all these vegan variants. It is a good idea to remove these from the "normal" variants. Also for sausages, cordon bleus, etc.

stephanegigandet (approved):

PR #6748, opened 2022-05-06
feat: Improvements to Nutri-Score panel, remove extended Eco-Score panel

### LINKED ISSUE #6532: The Nutri-Score info message in the sub-knowledge panel is not correct (It forgets nutrition is needed too)
### What

- [ ] The Nutri-Score info message in the sub-knowledge panel is not correct (It forgets to mention that nutrition is needed too)
- [ ] Extra point if we get a "Add Category" "Add nutrition"  button name in the API based on the specific subcase we are in (not completed, 99% of the time, I don't believe the category could be "not specific enough"), plus a normalized version of the tag_type to edit "category" or "nutrition_table"

### Part of
- #5517 
- #5493 

### Test product
- https://world.openfoodfacts.org/product/3794840
image

### Screenshots
image
image

### LINKED ISSUE #6585: Missing summary for nutrition on products without nutrition
### What

- [ ] Missing summary for nutrition on products without nutrition. Note that the right text appears in the summary card, which makes me suspect a client side issue
- [ ] I have also opened a server side issue, since I'm not sure what's the cause of the issue here (bad API, or problem on client): https://github.com/openfoodfacts/openfoodfacts-server/issues/6223

### Screenshot
image

### LINKED ISSUE #6678: disable the customized eco-score card for the time being

stephanegigandet:
- Remove extended Eco-Score panel, fixes #6678 
- Record and display all the reasons why we couldn't compute a Nutri-Score, and only the first one, fixes #6532
- Create a Nutri-Score knowledge panel even if we could not compute it. - fixes #6585

image

image


alexgarel (approved):
LGTM

PR #6756, opened 2022-05-10
feat: generate and send GS1 CIC confirmation messages to Agena3000

stephanegigandet:
Part of #6537

- Parse the "message" part of GS1 messages (previously we ignored the message envelope and extracted only the product data)
- Generate CIC XML confirmation messages
- Added tests for the confirmation messages
- Send back the confirmation messages to Agena3000


lib/ProductOpener/GS1.pm
```diff
@@ -1461,7 +1588,56 @@ sub read_gs1_json_file($$) {
...
+
+	# local GLN + 60 random hexadecimal characters
+	my $identifier = deep_get(\%options, qw(gs1 local_gln)) . "_";
+	$identifier .= sprintf("%x", rand 16) for 1..60;
+
+	return $identifier;
+}
+
+
+sub generate_gs1_confirmation_message($$) {
```
alexgarel, line 1605: A doc would be good, it's already quite complicated this gs1 stuff, so better document !

alexgarel (approved):
LGTM - just ask for a doc

PR #6760, opened 2022-05-11
chore: add natural rice sourdough with some translations

pilou-:
### What

Add `natural rice sourdough` with some translations


stephanegigandet (approved):
Thank you!

PR #6764, opened 2022-05-12
feat: New system to show how well products match user preferences

stephanegigandet:
This is an implementation of the product scoring system described in https://forum.openfoodfacts.org/t/proposal-for-a-new-personalized-experience-in-the-open-food-facts-app-and-website/71

- Scores are now on a 0 to 100 scale

- Product match status is now either:
1. Very good match
2. Good match
3. Poor match
4. Unknown match
and if the user selected some hard restrictions (e.g. vegan, or allergens) (currently selected by marking the attribute mandatory, but we could make new settings specific for restrictions that would be binary "I'm allergic to gluten"):
5. May not match (e.g. there can be traces of gluten, or some ingredient might not be vegan)
6. Does not match (contains gluten, is not vegan).

- Product match status is now indicated at the top of the card, in a similar way to the new Flutter app.
The score is now displayed, but we could hide it. I display it currently for debugging purposes.

- I removed the pagination at the top, it's now only on the bottom.

- I removed the tabs for each match status, to make the interface cleaner.

Screenshot:
image

2 minute video demo:
https://www.loom.com/share/33ca5a6179bd49b29ad2404a1b1af2ff


html/js/product-search.js
```diff
@@ -33,56 +47,82 @@ function match_product_to_preferences (product, product_preferences) {
...
-							if (attribute.match <= 20) {
-								status = "no";
-							}
+						// Mandatory attribute with a very bad score (e.g. contains an allergen) -> status: does not match
+						if ((attribute_preference == "mandatory") && (attribute.match <= 10)) {
+							status = "does_not_match";
+						}
+						// Mandatory attribute with a bad score (e.g. may contain traces of an allergen) -> status: may not match
+						else if ((attribute_preference == "mandatory") && (attribute.match <= 50)
+							&& ((status = "match") || (status = "may-not-match"))) {
```
alexgarel, line 85: You mean `===`, don't you ?

Where does the "may-not-match" comes from ?

html/js/product-search.js
```diff
@@ -16,7 +16,7 @@
 function match_product_to_preferences (product, product_preferences) {
 
 	var score = 0;
-	var status = "yes";
+	var status = "compatible";
```
alexgarel, line 19: Document possible values of status here.

html/js/product-search.js
```diff
@@ -33,56 +47,82 @@ function match_product_to_preferences (product, product_preferences) {
...
+		else if (score >= 50) {
+			status = "good_match";
+		}
+		else {
+			status = "poor_match";
+		}
+	}
+	else if (status == "does_not_match") {
+		score = 0;
 	}
```
alexgarel, line 122: I would do this above (as suggested) for this is not logical to have it here…

html/js/product-search.js
```diff
@@ -33,56 +47,82 @@ function match_product_to_preferences (product, product_preferences) {
...
 		});		
 	}
 	else {
 		// the product does not have the attribute_group field 
-		status = "unknown";
+		status = "unknown_match";
+	}
+
+	// Normalize the score from 0 to 100
+	score /= sum_of_factors;
```
alexgarel, line 101:
```suggestion
	if (status !== "does_not_match") {
	  // Normalize the score from 0 to 100
	  score /= sum_of_factors;
	else {
	  score = 0;
	}
```

alexgarel: Also, I'm not sure why we don't have `"may_not_match"` and `"unknown_match"` in the condition.

html/js/product-search.js
```diff
@@ -161,14 +206,17 @@ function display_products(target, product_groups, user_prefs ) {
...
-			var product_html = "";
+			var product_html = `<li><a href="${product.url}">`;
 
-			// Show the green / grey / colors for matching products only if we are using the user preferences
-			let css_classes = 'list_product_a';
 			if (user_prefs.use_ranking) {
-				css_classes += ' list_product_a_match_' + product.match_status;
+				product_html += '<div class="list_product_banner list_product_banner_' + product.match_status + '">'
+				+ lang()["products_match_" + product.match_status] + ' ' + Math.round(product.match_score) + '/100</div>'
+				+ '<div class="list_product_content">';
```
alexgarel, line 214: why did you remove the use of javascript template ? it's far less readable IMO.

lib/ProductOpener/Attributes.pm
```diff
@@ -813,7 +813,7 @@ sub compute_attribute_nova($$) {
 			$match = 100;
 		}
 		elsif ($nova_group == 3) {
-			$match = 50;
+			$match = 75;
 		}
```
alexgarel, line 817: Optional: This part could be better written as a lookup in a hash with scores.

html/js/product-search.js
```diff
@@ -109,10 +152,12 @@ function rank_products(products, product_preferences, use_user_product_preferenc
 	
 	if (use_user_product_preferences_for_ranking) {
 	
-		// Rank all products, and return them in 3 arrays: "yes", "no", "unknown"
+		// Rank all products
 		
 		products.sort(function(a, b) {
-			return (b.match_score - a.match_score) || (a.initial_order - b.initial_order);
+			return (b.match_score - a.match_score)	// Highest score first
+				|| ((a.match_status == "does_not_match" ? 1 : 0) - (b.match_status == "does_not_match" ? 1 : 0)) // Matching products second
```
alexgarel, line 159: You should use `===` for matching in js.
Why "does_not_match" gives me 1 point instead of 0 ?

alexgarel: Also is it normal that we do not consider the other status like may_not_match ?

alexgarel:
> I removed the tabs for each match status, to make the interface cleaner.

@stephanegigandet we could add a selector on the top to hide some status (eg. unknown etc)… but for a future version ;-)

alexgarel:
@stephanegigandet 

1. I'm not sure the /100 needs to be displayed, it makes even more data for eyes to process, without any information gain (maybe a % is ok, or nothing ?)

2. color choices do not really feat for color blind people.  They have the text though (so it's already functional), maybe use a ligther color for unknown match ?
Capture d’écran de 2022-05-17 12-12-41

html/js/product-search.js
alexgarel: hum I see, it's because you reversed a bnd b order compared to above.

I would better have:
```javascript
((b.match_status == "does_not_match" ? 0 : 1) - (a.match_status == "does_not_match" ? 0 : 1)) // Matching products second
```
but it's ok.

stephanegigandet: Ok, I'll change, and use === instead of ==

lib/ProductOpener/Attributes.pm
stephanegigandet: ok

stephanegigandet:
> 1. I'm not sure the /100 needs to be displayed, it makes even more data for eyes to process, without any information gain (maybe a % is ok, or nothing ?)

Right, in fact it's mostly for debugging purposes, I'm not sure we will show the score at all in the final version, maybe just the status.

> 2. color choices do not really feat for color blind people. They have the text though (so it's already functional), maybe use a ligther color for unknown match ?

Sure, we can put a lighter grey.

html/js/product-search.js
```diff
@@ -25,72 +32,128 @@ function match_product_to_preferences (product, product_preferences) {
...
+					product.attributes_for_status[match_status_for_attribute].push(attribute);
+
+					product.match_attributes[attribute_preference].push(attribute);
 				}
 			});
-		});		
+		});
+
+		// Normalize the score from 0 to 100
+		score /= sum_of_factors;
```
alexgarel, line 112: I would verify that sum_of_factors is not zero, to avoid throwing an exception.

```suggestion
		if (sum_of_factors === 0) {
		  score /= sum_of_factors;
		} else {
		  score = 0;
		}
```

html/js/product-search.js
```diff
@@ -25,72 +32,128 @@ function match_product_to_preferences (product, product_preferences) {
...
+		// If the product matches, check how well it matches user preferences
+		else if (score >= 75) {
+			product.match_status = "very_good_match";
+		}
+		else if (score >= 50) {
+			product.match_status = "good_match";
+		}
+		else {
+			product.match_status = "poor_match";
+		}
```
alexgarel, line 136: Cool refactor :+1:

alexgarel (approved):
Good, just a potential bug, that is better to fix ;-)

PR #6778, opened 2022-05-15
ci: Make dependabot GH project compatible

teolemon:
### What

ci: Make dependabot GH project compatible
Not finished, but someone gave us a solution: https://github.com/actions/add-to-project/issues/93


alexgarel:
@teolemon is this ready or not ? I don't understand. If it's not, put it as "draft".

stephanegigandet:
Another question is: do we really want to have dependabot PRs in GitHub project?

alexgarel:
> Another question is: do we really want to have dependabot PRs in GitHub project?

Yes I also strongly disagree, it's already so much noise !

alexgarel:
@teolemon can we close ?

PR #6784, opened 2022-05-16
chore: Bulgarian additions

Roto15:


taxonomies/packaging_materials.txt
```diff
@@ -163,7 +163,7 @@ pt:PET opaco
 <en:Plastic
 en:HDPE - High-density polyethylene, High-density polyethylene
```
alexgarel, line 164: why don't we have HDPE as the main word @teolemon or @stephanegigandet  (shan't we do it for every translation below ?)

taxonomies/packaging_materials.txt
```diff
@@ -163,7 +163,7 @@ pt:PET opaco
...
 xx:HDPE, HD-PE, RC 2, RC 02, RC2, RC02, HDPE 2, 2 HDPE, HDPE 02, 02 HDPE, PEHD, 2 PEHD, PEHD 2, 02 PEHD, PEHD 02, PE-HD, ♴
-bg:HDPE - Полиетилен висока плътност
+bg:HDPE - Полиетилен с висока плътност
```
alexgarel, line 166: Maybe add the form with "c" as a variant (to capture both) ? (same for LDPE)

taxonomies/packaging_materials.txt
teolemon: Because the main word is shown, and it's more pedagogical to see both

alexgarel (commented):
Awesome, I just have a doubt in packaging, let's see what @teolemon and @stephanegigandet says.

taxonomies/packaging_materials.txt
Roto15: I checked and officially they write it with the "c" but if its necesseray to have both variants I will edit the file.

Roto15:
Can we merge the request or should I make changes? Thanks @stephanegigandet

stephanegigandet (approved):
Looks good, thank you!

PR #6791, opened 2022-05-17
chore: Git ignore settings for JetBrains editors

k2s:
### What

Exclude files generated by JetBrains IDEA, WebStorm and others from adding to GIT.


stephanegigandet (approved):
Thanks!

PR #6797, opened 2022-05-19
fix: use mandatory food preferences for score

stephanegigandet:
Food preferences set to mandatory are now counted in the score of products (same weight as "very important").


alexgarel (approved):
:+1:

Issue #6771, opened 2022-05-13
When no avail translations, We should display English text rather than no text at all or taxonomy formatted text


teolemon:
When no available translations in a language, we should display English text rather than no text at all (in the case of the Bulgarian knowledge subpanel for AB Bio) or taxonomy formatted text  (in the case of the Bulgarian knowledge panel title for AB Bio)

alexgarel:
This is done.

PR #6804, opened 2022-05-20
fix: fallback to english for knowledge panels

alexgarel:
fixes #6771

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -407,6 +407,11 @@ sub process_template($$$) {
 
 	# Display one taxonomy entry in the target language
 	$template_data_ref->{display_taxonomy_tag} = sub ($$) {
+		return display_taxonomy_tag($lc, $_[0], $_[1]);
+	};
+
+	# Display one taxonomy entry in the target language, without language prefix
+	$template_data_ref->{display_taxonomy_tag_name} = sub ($$) {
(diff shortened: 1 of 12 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -407,7 +407,7 @@ sub process_template($$$) {
 
 	# Display one taxonomy entry in the target language
 	$template_data_ref->{display_taxonomy_tag} = sub ($$) {
-		return display_taxonomy_tag($lc, $_[0], $_[1]);
+		return display_taxonomy_tag_name($lc, $_[0], $_[1]);
```
stephanegigandet, line 410: This is ok I guess, I don't think we have templates where we do want the tag with the leading language code. We might in the future though, e.g. if we want to use panels to link to a list of products with a specific tag.

Another option is to change display_taxonomy_tag to display_taxonomy_tag_name in all templates.

stephanegigandet (approved):

lib/ProductOpener/Display.pm
alexgarel: I will change to have a specific display_taxonomy_tag_name available in templates

alexgarel: I've change that, to instead have two function and use appropriate one in knowledge panels.

stephanegigandet (approved):
Looks good, thank you!

PR #6817, opened 2022-05-24
fix: defining current_link

dipietroR:
In search.pl search_and_display_products is called without defining $request_ref->{current_link} resulting in a warning about an undefined variable. current_link and current_link_query appear to be the same where I see them in the logs. So defining it this way may be alright, though I haven't looked at what process_template ends up doing with template_data_ref which is where the error is coming from. I have yet to understand what current_link and current_link_query are and how they are different from another. They appear to be the path and the arguments, but I'm not sure and could use some guidance.

### DIFF
--- a/cgi/search.pl
+++ b/cgi/search.pl
@@ -643,6 +643,7 @@ elsif ($action eq 'process') {
 	}
 
 	$request_ref->{current_link_query} = $current_link;
+	$request_ref->{current_link} = $current_link;
 
 	my $html = '';
 	#$query_ref->{lc} = $lc;


stephanegigandet:
> In search.pl search_and_display_products is called without defining $request_ref->{current_link} resulting in a warning about an undefined variable. current_link and current_link_query appear to be the same where I see them in the logs. So defining it this way may be alright, though I haven't looked at what process_template ends up doing with template_data_ref which is where the error is coming from. I have yet to understand what current_link and current_link_query are and how they are different from another. They appear to be the path and the arguments, but I'm not sure and could use some guidance.

I wrote that code a long time ago, and quite frankly I don't remember what was the intended difference.

It's probably something we should refactor and document better, by first analyzing the different uses of $current_link and $current_link_query (e.g. creating links for pagination, creating a link to the same page but on the world subdomain etc.).

Most of the occurences of current_link/current_link_query are in Display.pm, and it is also used in templates/web/common/includes/list_of_products.tt.html

stephanegigandet:
@dipietroR I opened an issue for the refactor: https://github.com/openfoodfacts/openfoodfacts-server/issues/6819

stephanegigandet (approved):

PR #6830, opened 2022-05-29
feat: refresh the new Eco-Score and Nutri-Score logos

teolemon:
### What
feat: refresh the new Eco-Score and Nutri-Score logos, in line with the graphic charter. For the Nutri-Score, it's a very slight adjustment in radius, apparently, but a fix to avoid issues when scaling


stephanegigandet:
Looks good, but this will need to be tested before being merged as the SVG size has changed, it's possible that it might not be displayed correctly in some places (e.g. on the current version of the website).

stephanegigandet:
We have a few constraints for the logos, for the Nutri-Score ones I will update them with the latest version from https://www.santepubliquefrance.fr/media/files/02-determinants-de-sante/nutrition-et-activite-physique/nutri-score/annexe2-charte-graphique , and make sure there are no size issues (for instance all logo files need to be the same size and ratio, to avoid displaying some of the Nutri-Score (like unknown) bigger or smaller than the others.

stephanegigandet:
Replaced by https://github.com/openfoodfacts/openfoodfacts-server/pull/6848

Issue #6819, opened 2022-05-25
Refactor and document current_link / current_link_query 


stephanegigandet:
In many part of the code (in particular Display.pm, search.pl and templates/web/common/includes/list_of_products.tt.html) we have variables like $current_link and $current_link_query that are passed through $request_ref and used in some templates.

The use of those variables is often to construct an URL to point to the current page, but with all parameters included either in the path or in the query part of the URL. So for instance if we have a search request with POST parameters, we reconstruct a URL with GET parameters that can be shared, and that the templates can extend with extra parameters (e.g. adding a parameter to download the search results, or changing the subdomain to make the query on world.openfoodfacts.org).

The distinction between current_link and current_link_query is probably not useful. Also it's not very clear / consistent whether those links are / should be absolute or relative URLs.

So it would definitely make a lot of sense to rethink this, and use a consistent, simple, and documented method to do this.

See original discussion on https://github.com/openfoodfacts/openfoodfacts-server/pull/6817

dipietroR:
I'd like to work on this, I'll start looking at what the variables are doing and how we might clean them up and document them.

yuktea:
Closed in #6832

PR #6832, opened 2022-05-29
refactor: Refactor and document current_link / current_link_query

dipietroR:
Potential fix for https://github.com/openfoodfacts/openfoodfacts-server/issues/6819

current_link and current_link_query appear to be the same when I saw them. The primary difference appears to have been that current_link was the current link without arguments however this was applied inconsistently. And where I saw it, it usually had arguments.

The most clear place that this difference was evident that I found was
https://github.com/openfoodfacts/openfoodfacts-server/blob/f63cb9cc0bd75ffa5526d2c1ffe63123d4633f7e/lib/ProductOpener/Display.pm#L5469
As pages can be reached both by url/<number> or url?page=<number>. The former was used when "current_link" was defined and the latter when "current_link_query" was defined. Since both were possible, it seemed reasonable to drop the index function of using /<number> and to use page=<number> exclusively. Though I may be missing some details. Also if this is the case, there may be other code that can be removed as the /<number> might not be needed anymore...Or might if anyone has previously made a link to it, and changing the code would break their link.

Since the two variables were mostly confusing because they appeared to be offering the same content, further documentation of the variable merged into one may not be necessary, though if it is, where should it be put?

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -4311,8 +4311,6 @@ JS
 
 		my $query_ref = {};
 
-		$request_ref->{current_link_query} = $current_link;
-
 		if (defined param('parent_ingredients')) {
 			$html .= search_and_analyze_recipes($request_ref, $query_ref);
 		}
@@ -5211,12 +5209,12 @@ sub search_and_display_products($$$$$) {
 	$template_data_ref->{jqm} = param("jqm");
 	$template_data_ref->{country} = $country;
 	$template_data_ref->{world_subdomain} = get_world_subdomain();
-	$template_data_ref->{current_link_query} = $request_ref->{current_link_query};
+	$template_data_ref->{current_link} = $request_ref->{current_link};
 	$template_data_ref->{sort_by} = $sort_by;
 
 	# Query from search form: display a link back to the search form
-	if (defined($request_ref->{current_link_query}) && $request_ref->{current_link_query} =~ /action=process/) {
-		$template_data_ref->{current_link_query_edit} = $request_ref->{current_link_query};
+	if (defined($request_ref->{current_link}) && $request_ref->{current_link} =~ /action=process/) {
+		$template_data_ref->{current_link_query_edit} = $request_ref->{current_link};
 		$template_data_ref->{current_link_query_edit} =~ s/action=process/action=display/;
 	}
 
@@ -5227,8 +5225,8 @@ sub search_and_display_products($$$$$) {
 		# Show a download link only for search queries (and not for the home page of facets)
 
 		if ($request_ref->{search}) {
-			$request_ref->{current_link_query_download} = $request_ref->{current_link_query};
-			if ($request_ref->{current_link_query} =~ /\?/) {
+			$request_ref->{current_link_query_download} = $request_ref->{current_link};
+			if ($request_ref->{current_link} =~ /\?/) {
 				$request_ref->{current_link_query_download} .= "&download=on";
 			}
 			else {
@@ -5413,6 +5411,15 @@ JS
 	return $html;
 }
 
+=head2 display_pagination( $request_ref , $count , $limit , $page )
+
+This function is used for page navigation and gets called when there is more
+than one page of products.  The URL can be different, either page=<number> , or
+/<number> . page=<number> is used for search queries. /<number> is used for
+facets.
+
+=cut
+
 sub display_pagination($$$$) {
 	my $request_ref = shift;
 	my $count = shift;
@@ -5427,12 +5434,11 @@ sub display_pagination($$$$) {
 	if (not defined $current_link) {
 		$current_link = $request_ref->{world_current_link};
 	}
-	my $current_link_query = $request_ref->{current_link_query};
 
-	$log->info("current link and current link query", { current_link => $current_link, current_link_query => $current_link_query }) if $log->is_info();
+	$log->info("current link", { current_link => $current_link }) if $log->is_info();
 
 	if (param("jqm")) {
-		$current_link_query .= "&jqm=1";
+		$current_link .= "&jqm=1";
 	}
 
 	my $next_page_url;
@@ -5444,7 +5450,7 @@ sub display_pagination($$$$) {
 		$nofollow = ' nofollow';
 	}
 
-	if ((($nb_pages > 1) and ((defined $current_link) or (defined $current_link_query))) and (not defined $request_ref->{product_changes_saved})) {
+	if ((($nb_pages > 1) and (defined $current_link)) and (not defined $request_ref->{product_changes_saved})) {
 
 		my $prev = '';
 		my $next = '';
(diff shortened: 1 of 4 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -5467,30 +5464,12 @@ sub display_pagination($$$$) {
...
-						if ($i > 1) {
-							$link .= "/$i";
-						}
-						if ($link eq '') {
-							$link  = "/";
-						}
-						if (defined $request_ref->{sort_by}) {
-							$link .= "?sort_by=" . $request_ref->{sort_by};
-						}
-					}
```
alexgarel, line 5486: Won't it be a regression to remove that ?  I would think that previously we were going through this block ? (because $current_link was defined in most cases). Although I agree this is not compatible with current_link in search.pl (which did already have parameter, while this block clearly applies to an url without parameters).

templates/web/common/includes/list_of_products.tt.html
```diff
@@ -2,11 +2,11 @@
 <!-- start templates/[% template.name %] -->
 
 <p>[% error %]</p>
-[% IF (current_link_query.defined) && !(jqm.defined) %]
+[% IF (current_link.defined) && !(jqm.defined) %]
     [% IF country != 'en:world' %]
-        <p>&rarr; <a href= "[% world_subdomain %][% current_link_query %]"> [% lang('view_results_from_the_entire_world') %]</a></p>
+        <p>&rarr; <a href= "[% world_subdomain %][% current_link %]"> [% lang('view_results_from_the_entire_world') %]</a></p>
```
alexgarel, line 7: I'm not sure if in that kind of cases the existence of $current_link_query, was not a marker that we were on a search results page (but maybe this template is only used in the context of a search result, in which case this is ok)

alexgarel: It might be a false alarm, as it seems list_of_products.tt.html is used only for search results.

lib/ProductOpener/Display.pm
alexgarel: I think this correspond to pages like https://world.openfoodfacts.net/brand/ferrero

alexgarel (commented):
I will let @stephanegigandet review this, I have made some comments though

templates/web/common/includes/list_of_products.tt.html
stephanegigandet: The template is used whenever we display a list of products (e.g. the front page, facets, search results etc.).
e.g. https://uk.openfoodfacts.org/label/organic

Currently we do not display the link to "world" when we are on the home page of a country.

dipietroR: http://world.openfoodfacts.localhost/label/organic seems to work on this patch. Though my testing is entirely based off of looking at the dev env after applying the patch. It is surely not a full test, and as mentioned here and above, I'm missing links. to look at, and am likely to be missing more.

How should this PR proceed, does it need changes, or it is too unclear on if it could introduce a problem and should be closed?

lib/ProductOpener/Display.pm
stephanegigandet: Hi @dipietroR , we do not have tests for the display_pagination() function yet, so you have to test your changes manually.

I think the first thing you could do is to add a comment in the POD format above the function to explain what it does (you can see examples of that for other functions that already have documentation above them).

display_pagination() generates a list of links to the different pages that correspond to a query. 
e.g. if you go to the bottom of https://world.openfoodfacts.org/
you will see:

image

Notice that the links have page numbers at the end like this: https://world.openfoodfacts.org/4

If you do a search query: https://world.openfoodfacts.org/cgi/search.pl?search_terms=coke&search_simple=1&action=process

Then you get page numbers, but the links are currently broken:
https://world.openfoodfacts.org/cgi/search.pl?action=process&search_terms=coke&sort_by=unique_scans_n&page_size=24/3?sort_by=unique_scans_n

They should be like this instead:

https://world.openfoodfacts.org/cgi/search.pl?action=process&search_terms=coke&sort_by=unique_scans_n&page_size=24&page=3

That was the intent of the code that you removed: if the link was a search query, then page numbers are to be passed as an extra parameter : &page=3.  If it's a facet, like https://world.openfoodfacts.org/label/organic , then we just add it like this : https://world.openfoodfacts.org/label/organic/3

To test it locally, you have very few products, so you won't get pagination unless you make an empty query:
http://world.openfoodfacts.localhost/cgi/search.pl?search_terms=&search_simple=1&action=process

The fact that it is currently broken in production today clearly shows that we should add some tests to catch this.

Could you try to add a test in display.t to test the result of the function display_pagination() with different types of $current_link and more than 1 page?

dipietroR: I tried setting up 109 of the products in the test environment with the Organic label, though I couldn't get it to show the pagination that is shown on the production page (https://world.openfoodfacts.org/label/organic), Rather it only shows 100, then I can reach the second page by manually adding a /2 or ?page=2 at which point I see the remaining 9 products. Not sure why this is happening, but it is all happening on the main branch, rather than my branch. I'm pulling the full DB to test with that, but it is going to take awhile. Looks like it should be done in a few hours, so hopefully I can have a look with that tomorrow.

I can look into making a test for the display_pagination() function, though to verify, that should be part of a separate PR, correct?

stephanegigandet: > I can look into making a test for the display_pagination() function, though to verify, that should be part of a separate PR, correct?

If you can it's a good thing to make separate PRs if things are independent, so that we can merge things incrementally without having to wait for every part to be perfect.

dipietroR: Neat! no_cache=1 was just what I needed.

After applying no_cache=1 once. I don't seem to need to run it again, at which point on this PR I can go to:
http://world.openfoodfacts.localhost/label/organic?page=2&page_size=10
And the link to the first page will show:
http://world.openfoodfacts.localhost/label/organic?page=1&page_size=10
which indeed will go to the first page when clicked.

So far as I understand this PR is working, and fixes the bug that it, inadvertently, uncovered. Is there anything else that should be done on this before merging it?

stephanegigandet: @dipietroR : there are 2 types of pages where we display a pagination:
- what we call "facets", with urls like /label/organic
- results of search queries with urls like /cgi/search.pl?...  or like https://world.openfoodfacts.org/search?labels_tags_fr=bio 

For the facets, we want the URLs of the different pages to use /[page number] at the end.
e.g. if you go to https://world.openfoodfacts.org/label/organic in production, the links in the pagination are to https://world.openfoodfacts.org/label/organic/2 etc.

This was done by the code removed above, we need to add back this logic. One way could be to check if $current_link contains a question mark, in which case we use the format &page=[page number], otherwise we add /[page number] when the page number is greater than 1.

stephanegigandet: Also before merging the PR, it would be great to:

1. document the function in the POD format, to explain what it does, and the difference for the different page types (e.g. a summary of the info in this ticket).

2. add a few unit tests in display.t to test those different cases

dipietroR:
Ok I've made some of the updates recommended above. I think I'll add the testing in another PR. I couldn't test everything as I didn't have the full collection to work with. I'm pulling that down now. And will test the labels when I have that.

Is there a way for me to re-use an existing openfoodfacts-mongodbdump.tar.gz file that is already on my laptop?

PR #6833, opened 2022-05-30
fix: make NOVA panel clearer

teolemon:
### What

fix: make NOVA panel clearer

### Part of
- #5519

### DIFF
--- a/templates/api/knowledge-panels/health/ingredients/nova.tt.json
+++ b/templates/api/knowledge-panels/health/ingredients/nova.tt.json
@@ -45,12 +45,15 @@
             "text_element": {
                 "html": `
 <p>[% lang("nova_classification_description") %]</p>
+
 <ol>
-<li>[% IF nova_group == 1 %]<b>[% END %][% lang('attribute_nova_1_description_short') %][% IF nova_group == 1 %]</b>[% END %]</li>
-<li>[% IF nova_group == 2 %]<b>[% END %][% lang('attribute_nova_2_description_short') %][% IF nova_group == 2 %]</b>[% END %]</li>
-<li>[% IF nova_group == 3 %]<b>[% END %][% lang('attribute_nova_3_description_short') %][% IF nova_group == 3 %]</b>[% END %]</li>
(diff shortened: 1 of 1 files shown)


templates/api/knowledge-panels/health/ingredients/nova.tt.json
```diff
@@ -44,10 +46,10 @@
                 "html": `
 <p>[% lang("nova_classification_description") %]</p>
 <ol>
-<li>[% lang('attribute_nova_1_description_short') %]</li>
-<li>[% lang('attribute_nova_2_description_short') %]</li>
-<li>[% lang('attribute_nova_3_description_short') %]</li>
-<li>[% lang('attribute_nova_4_description_short') %]</li>
+<li>[% IF nova_group == 1 %]<b>[% END %][% lang('attribute_nova_1_description_short') %][% IF nova_group == 1 %]</b>[% END %]</li>
```
stephanegigandet, line 49: We can use a style, it will be cleaner and will halve the number of IF conditions.

templates/api/knowledge-panels/health/ingredients/nova.tt.json
```diff
@@ -35,6 +35,8 @@
         [% END %]
     [% END %]
 [% END %]
+            <br>
```
stephanegigandet, line 38: Adding <br> at the end of panels is probably not the right way, clients should decide how much spaces to leave.

stephanegigandet:
Good idea to bold the explanantion for the product NOVA group. It probably makes sense to invert the panels, and put the matching ingredients after the 4 groups.

teolemon:
Not fixed yet

alexgarel (approved):
LGTM, although it would be better not to introduce <br> as asked by @stephanegigandet

teolemon:
@alexgarel how about we merge that one ?

alexgarel:
Can you fix things proposed by @stephanegigandet, @teolemon ?

teolemon:
@alexgarel is it better ?

PR #6834, opened 2022-05-30
feat: expand some panels only on large devices

### LINKED ISSUE #6786: Add a way to have knowledge panels collapsed in the app but not on the website (list of ingredients + nutrition table)
### What
Add a way to have knowledge panels collapsed in the app but not on the website (list of ingredients + nutrition table)
### Part of
- #5493

alexgarel:
for now web is always considered large, but we have a specific class we could play upon.
Applied to nutrition and ingredients panels

fixes #6786


stephanegigandet (approved):
Looks good, thank you

PR #6840, opened 2022-05-31
build: speed up build actions

alexgarel:
This is an attempt at speeding up build time, by uploading docker images as artifacts

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -35,15 +35,48 @@ jobs:
     - name: lint
       run: make front_lint
 
-  test:
-    name: Perl unit tests
+  # this will build the docker image and upload as an artifact for following jobs
+  build_backend:
+    name: Build backend dev image for tests
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


alexgarel:
@stephanegigandet this take down checks to 18mn (5 min for build + 12 min for unit tests), make dev and check perl are made in parallel.

alexgarel:
I think build_taxonomies  is still slow but I'm not sure why…

stephanegigandet:
> @stephanegigandet we maybe should investigate rebuilding taxonomies on a regular basis, maybe with a reccuring action ?

Well could a GitHub action check if a taxonomy is changed, and in that case not only build the taxonomy, but also commit the result in the same branch?

alexgarel:
> Well could a GitHub action check if a taxonomy is changed, and in that case not only build the taxonomy, but also commit the result in the same branch?

The problem is that on multiple commits, it will rebuild and commit. I do not find it very practical because you then have to pull before adding commit, or pull with rebase)

But maybe we can do a check and block merging if you did not rebuild all the taxonomies ?

Issue #6092, opened 2021-11-21
Return an explanation for Eco-Score exemptions


teolemon:
## What
The eco-score does not grade bottled waters, and there's no way to know why.

## Describe the solution you'd like

- [ ] Return an explanation for Eco-Score exemptions, and a suggestion to drink tap water or self-carbonated water.
- [ ] Expose this explanation in knowledge panels and in the legacy HTML methodology

## Part of
- #5520 
- #5493

teolemon:
"L'Eco-Score n'est pas encore dispo pour cette catégorie."

alexgarel:
@teolemon it's not in current sprint ?

alexgarel:
@teolemon we got two options here:

- continue to consider this as "nutriscore unknown" but adapt messages
- consider it as a new status: "nutriscore not applicable"
- a mix of those (continue to put it as unknown but use a specific panel instead of ecoscore_unknown.tt.json)

Would you also like a different icon ?

stephanegigandet:
@alexgarel It should be "Eco-Score not applicable", as we do for "Nutri-Score not applicable".

In fact we already compute it:

https://uk.openfoodfacts.org/eco-score/not-applicable

ecoscore_data: {
adjustments: { },
ecoscore_not_applicable_for_category: "en:sodas",
status: "unknown"
},
ecoscore_grade: "not-applicable",
ecoscore_tags: [
"not-applicable"
],

And we will need a different icon, similar to the Nutri-Score not applicable icon.

stephanegigandet:
Discussed with @teolemon and @alexgarel:

- New messages : "Eco-Score not applicable", "Nutri-Score not applicable".
- Color should be grey.
- Icon to be discussed. (maybe just an outline).

Issue #6774, opened 2022-05-13
Add a knowledge panel to explain why a category is exempted from Nutri-Score


teolemon:
### What

- [ ] Add a knowledge panel to explain why a category is exempted from Nutri-Score

### Screenshot

![Image](https://user-images.githubusercontent.com/1689815/168380447-5218e1bb-0a49-4e23-96d6-1a7188e80081.png)

### Part of
- #5493

teolemon:
![image](https://user-images.githubusercontent.com/1689815/169710471-ad3d2c3b-d2be-4072-9100-1abc61c6a9e5.png)

teolemon:
![image](https://user-images.githubusercontent.com/1689815/169710520-a3b9bf22-0088-442a-9331-1fc5c15d1445.png)

PR #6844, opened 2022-06-01
feat: add explanation with category in the Nutri-Score and Eco-Score knowledge panels when not applicable

stephanegigandet:
When a product does not have a Nutri-Score or an Eco-Score because the product category is excluded from the Nutri-Score and Eco-Score, we now explain why in the corresponding knowledge panel, and we list the category.

image

Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/6092 and https://github.com/openfoodfacts/openfoodfacts-server/issues/6774

### DIFF
--- a/lib/ProductOpener/Attributes.pm
+++ b/lib/ProductOpener/Attributes.pm
@@ -75,6 +75,9 @@ use ProductOpener::Lang qw/:all/;
 use ProductOpener::Display qw/:all/;
 use ProductOpener::Ecoscore qw/:all/;
 
+use Data::DeepAccess qw(deep_get);
+
+
 =head1 CONFIGURATION
 
 =head2 Attribute groups and attributes
@@ -463,7 +466,7 @@ sub compute_attribute_nutriscore($$) {
 	my $attribute_ref = initialize_attribute($attribute_id, $target_lc);
 	
 	# Nutri-Score A, B, C, D or E
-	if (defined $product_ref->{nutriscore_data}) {
+	if ((defined $product_ref->{nutriscore_grade}) and ($product_ref->{nutriscore_grade} =~ /^[a-e]$/)) {
 		$attribute_ref->{status} = "known";
 		
 		my $nutriscore_data_ref = $product_ref->{nutriscore_data};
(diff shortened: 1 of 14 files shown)


lib/ProductOpener/Attributes.pm
```diff
@@ -656,14 +663,30 @@ sub compute_attribute_ecoscore($$$) {
 		}
 		$attribute_ref->{icon_url} = "$static_subdomain/images/attributes/ecoscore-$grade.svg";
 	}
+	# Eco-Score is not-applicable
+	elsif ((defined $product_ref->{ecoscore_grade}) and ($product_ref->{ecoscore_grade} eq "not-applicable")) {
+		$attribute_ref->{status} = "unknown";
+		$attribute_ref->{icon_url} = "$static_subdomain/images/attributes/ecoscore-unknown.svg";
```
alexgarel, line 669: you did not include the image, did you ?

lib/ProductOpener/Lang.pm
```diff
@@ -210,14 +211,55 @@ If a translation is not available, the function returns English.
...
+
+e.g. "This is a string with {a_variable} and {another_variable}."
+
+Variables between { } are interpolated with the corresponding entry
+in the $variables_ref hash reference.
+
+If a translation is not available, the function returns English.
+
+=head3 Arguments
+
```
alexgarel, line 245:
```suggestion

 =head4 target language $target_lc
 
```

alexgarel (changes requested):
LGTM but you forgot an image. Also I was expecting a new nutriscore-not-applicable logo ;-)

alexgarel:
Also, @stephanegigandet, could be good to add a test for eco-score not applicable…

alexgarel:
proposal for logo:
nutriscore-notapplicable

stephanegigandet:
@alexgarel for the logos, I made a separate PR:
https://github.com/openfoodfacts/openfoodfacts-server/pull/6848/files

Santé publique France has an official logo for "Nutri-Score unknown", so I think we should use it.

I think we can use the same logo for "Nutri-Score not-applicable" as well.

Same for Eco-Score unknown and not-applicable (yet), we can use the same logo.

stephanegigandet:
> Also, @stephanegigandet, could be good to add a test for eco-score not applicable…

Good point, I added a test.

lib/ProductOpener/Attributes.pm
alexgarel: will come with #6848

alexgarel (approved):
Good.

PR #6848, opened 2022-06-02
feat: refresh the Nutri-Score and Eco-Score logos

stephanegigandet:
This PR replaces https://github.com/openfoodfacts/openfoodfacts-server/pull/6830

The Nutri-Score logos are exactly the ones from https://www.santepubliquefrance.fr/media/files/02-determinants-de-sante/nutrition-et-activite-physique/nutri-score/annexe2-charte-graphique

Their default size is set to 240x130 px (like the previous logos).

The new Eco-Score logos are also 240x130. They are almost the same as the ones created by Quentin in https://www.figma.com/file/nFMjewFAOa8c4ahtob7CAB/Mobile-App-Design-(Quentin) 
The difference is the size, they are slightly longer, so that we can have the exact same size as the Nutri-Score logos.

Small view:

image

Large view:

image


stephanegigandet:
For the not applicable Eco-Score and Nutri-Score logos, I made them lighter (so less visible) than the unknown versions.

alexgarel (changes requested):
All good, but the nutriscore unknown, I prefer the old version with "?" (on eco-score we have the "?"). Why not keep the "?".

I like the not applicable versions.

stephanegigandet:
I liked the version with the ? too (I'm biased, I created it ;) )

But the Nutri-Score charter states that we should use their neutral logos without a question mark:

image

PR #6856, opened 2022-06-04
Diverse meat analogue categories

moon-rabbitOFF:


alexgarel:
@aleene and @stephanegigandet, I let you review that :-)

PR #6860, opened 2022-06-06
test: Add unit test for bug report on normalizing quantities

yuktea:
A test PR

### DIFF
--- a/t/food.t
+++ b/t/food.t
@@ -72,6 +72,9 @@ is( normalize_serving_size("42\N{U+516C}\N{U+5347}"), 42000 );
 is( unit_to_g(42, "\N{U+516C}\N{U+5347}"), 42000 );
 is( g_to_unit(42000, "\N{U+516C}\N{U+5347}"), 42 );
 
+#test for a bug
+is( normalize_serving_size("100grams"), 100 );
+
 # Russian units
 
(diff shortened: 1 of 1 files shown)


t/food.t
```diff
@@ -72,6 +72,9 @@ is( normalize_serving_size("42\N{U+516C}\N{U+5347}"), 42000 );
 is( unit_to_g(42, "\N{U+516C}\N{U+5347}"), 42000 );
 is( g_to_unit(42000, "\N{U+516C}\N{U+5347}"), 42 );
 
+#test for a bug
```
stephanegigandet, line 75: Maybe we could add the link to the actual bug.

stephanegigandet:
2 unrelated files were deleted in this PR.

t/food.t
yuktea: > 2 unrelated files were deleted in this PR.

@stephanegigandet This is strange. Not sure how that happened but I'll be more careful now. I don't remember staging these changes.

yuktea: Done and done

yuktea:
> 2 unrelated files were deleted in this PR.

Reverted. Wasn't mindful enough while staging.

stephanegigandet:
The test looks good, and it is failing which shows it's working. Once the bug is fixed, we can merge the test.

PR #6867, opened 2022-06-07
fix: add back nutrition facts table title in knowledge panel

stephanegigandet:
The openfoodfacts-dart package expects a non null title for the table elements in knowledge panels. This adds back the table title that was removed in https://github.com/openfoodfacts/openfoodfacts-server/pull/6839

Hot fix deployed in production as it was breaking scanning and searching in Smoothie.

Related:
https://github.com/openfoodfacts/smooth-app/issues/2203
https://github.com/openfoodfacts/openfoodfacts-dart/issues/477

### DIFF
--- a/templates/api/knowledge-panels/health/nutrition/nutrition_facts_table.tt.json
+++ b/templates/api/knowledge-panels/health/nutrition/nutrition_facts_table.tt.json
@@ -14,6 +14,7 @@
             "element_type": "table",
             "table_element": {
                 "id": "nutrition_facts_table",
+		"title": "[% lang('nutrition_data_table') %]",
                 // Create a structure to check if we already have seen a column for a given scope
                 // so that we can mark the first one as the default column for the group
                 [% SET scopes = {} %]              
--- a/templates/api/knowledge-panels/health/nutrition/nutrition_facts_table.tt.json
+++ b/templates/api/knowledge-panels/health/nutrition/nutrition_facts_table.tt.json
@@ -14,7 +14,7 @@
             "element_type": "table",
             "table_element": {
                 "id": "nutrition_facts_table",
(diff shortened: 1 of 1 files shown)


teolemon (approved):

templates/api/knowledge-panels/health/nutrition/nutrition_facts_table.tt.json
```diff
@@ -14,6 +14,7 @@
             "element_type": "table",
             "table_element": {
                 "id": "nutrition_facts_table",
+		"title": "[% lang('nutrition_data_table') %]",
```
teolemon, line 17:
```suggestion
                "id": "nutrition_facts_table",
                "title": "[% lang('nutrition_data_table') %]",
```

M123-dev (approved):

VaiTon:
Maybe adding a TODO "remove later" would make us remember to change it?

stephanegigandet:
> Maybe adding a TODO "remove later" would make us remember to change it?

@VaiTon : What would need to be changed? It's a good thing to have table titles. They don't have to be displayed by the client, but they are useful for screen readers.

On the web, we use it to set an aria-label:

`<table[% IF element.table_id %] id="[% table_element.table_id %]"[% END %] aria-label="[% table_element.title %]">`

VaiTon:
I thought we moved them here
https://github.com/openfoodfacts/openfoodfacts-server/blob/d9f898f17b4b02437485dc40aaae4ab4d12ef09e/templates/api/knowledge-panels/health/nutrition/nutrition_facts_table.tt.json#L9

If it's still useful, let's keep it!

stephanegigandet:
> I thought we moved them here

Right, we added the title elsewhere, but we still need a table title, even if it's shown only for screen readers.

PR #6869, opened 2022-06-08
feat: small improvements to ingredients and nutrition table panels

stephanegigandet:
Some small changes:
- made barcode, ingredients and nutrition table icons colorless (removing the fill completely instead of using current color) so that they can be colored more easily
- made the icon color for the ingredients and nutrition facts knowledge panel dependent on the evaluation, and the evaluation is currently unknown, which makes the icon grey: this should solve the issue that the icons were not visible in Smoothie in dark mode
- display the Eco-Score not applicable icon in the Eco-Score facet


alexgarel (approved):
Cool.

PR #6872, opened 2022-06-08
fix: flatten Nutri-Score and Eco-Score SVG icons

stephanegigandet:
The new Nutri-Score and Eco-Score icons did not display correctly in Smoothie (probably an issue with how the flutter_svg package deals with clipPath and/or transforms).

See: https://github.com/openfoodfacts/smooth-app/issues/2205

In order to solve the issue, the corresponding SVG icons were flattened (removing clipPath and transforms, so that the paths have absolute coordinates).

It's unfortunately not possible to do automatically through gulp-svg / SVGO:

The method that worked for me:
- using the Flatten SVG option of https://lean-svg.netlify.app/
- removing the clipPath manually in Inkscape for the 2 Nutri-Score B and D icons that appeared broken in lean-svg when the Flatten SVG option was active


alexgarel (approved):
Good !

Issue #6845, opened 2022-06-01
Undefined variable in search.pl


dipietroR:
### Describe the bug

When a blank search is made the following warnings appear:
[Wed Jun  1 20:43:53 2022] -e: Use of uninitialized value in concatenation (.) or string at /opt/product-opener/cgi/search.pl line 661.
[Wed Jun  1 20:43:53 2022] -e: Use of uninitialized value in concatenation (.) or string at /opt/product-opener/cgi/search.pl line 661.
[Wed Jun  1 20:43:53 2022] -e: Use of uninitialized value in concatenation (.) or string at /opt/product-opener/cgi/search.pl line 661.
[Wed Jun  1 20:43:53 2022] -e: Use of uninitialized value in concatenation (.) or string at /opt/product-opener/cgi/search.pl line 661.
[Wed Jun  1 20:43:53 2022] -e: Use of uninitialized value in concatenation (.) or string at /opt/product-opener/cgi/search.pl line 747.

This appears to be because $request_ref->{count} is not defined at those lines.

In looking at search.pl and Display.pm I'm not seeing a point where $request_ref->{count} is set in Display.pm where search.pl is or was calling it. I'm seeing a few instances of the element being set in the hash using:
git log -S'equest_ref->{count}' -- Display.pm

96d31df70a4914b240c66a73d35f436f63093af7
shows the element being removed from search_and_export_products, which is called in search.pl, though adding a static
$request_ref->{count} = 5;
Does not remove the warning.

8f2cfcdee1b03c60530babb225cd6516b08ddab8
shows the element added to the display_ecoscore_calculation_details_simple_html subroutine, though it does not appear that this was ever used in search.pl
`git log -S'display_ecoscore_calculation_details_simple_html' -- search.pl`

2a202e5cdae1da9683d5feccaa69772c8f2883cc
shows the element removed from search_and_display_products. This is used in search.pl. However adding it back in with a static value does not seem to remove the warnings.

926843164eed8f68fe439b9c65935aecb4f6dac4
shows up in the git log search, but I'm not finding $request_ref->{count} in it, perhaps my git search is not correct?

Expanding the search to `git log -S'equest_ref->{count}' -- .` from lib/ProductOpener does not yield any more results

In Display.pm $request_ref->{count} is defined in:
search_and_graph_products
search_and_map_products
search_and_analyze_recipes

of which:
search_and_graph_products
search_and_map_products

[...]

PR #6877, opened 2022-06-09
fix: Undefined variable in search.pl

dipietroR:
Potential fix for https://github.com/openfoodfacts/openfoodfacts-server/issues/6845

### DIFF
--- a/cgi/search.pl
+++ b/cgi/search.pl
@@ -654,18 +654,10 @@ elsif ($action eq 'process') {
 
 	my $share = lang('share');
 
-	my $map = '';
-	if(defined param("generate_map")) {
-		$map = param("generate_map");
-	}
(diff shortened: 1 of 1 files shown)


cgi/search.pl
```diff
@@ -654,11 +654,21 @@
 
 	my $share = lang('share');
 
-
+	my $map = '';
```
stephanegigandet, line 657: This works, but you can do it in one line instead of 4 like this:

my $map = param("generate_map") || '';

If the first expression is true (it is defined and the value is not false), then it return the first expression, otherwise it returns the right expression.

dipietroR: How's this?

stephanegigandet (approved):
Perfect, thank you!

PR #6878, opened 2022-06-09
feat: Support for unit name normalization

### LINKED ISSUE #6843: Unit names (e.g. grams) are not recognized for quantity and serving size normalization (only unit symbol e.g. g)
### Describe the bug

Nutri-Score not computed when per serving = "100grams".

### To Reproduce

https://world.openfoodfacts.org/product/7350002400104/kikartor-zeta?rev=7

### Expected behavior

Nutri-Score should be computed.

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Browser

### Browser version

_No response_

_No response_

### Time per product

_No response_

yuktea:
### What

Enables recognition of unit names by incorporating unit normalization in our normalization logic. Will enable OFF to compute nutriscores
when serving sizes have unit names etc.

- Fixes #6843


lib/ProductOpener/Food.pm
```diff
@@ -1260,21 +1265,45 @@ normalize_serving_size(2.5kg)              returns 2500
...
 
-	my $q = 0;
-	my $u;
-
-	if ($serving =~ /((\d+)(\.|,)?(\d+)?)( )?($units)\b/i) {
-		$q = lc($1);
-		$u = $6;
+	# Regex captures any <number>( )?<unit-identifier> group, but leaves allowances for a preceding
+	# token to allow for patterns like "One bag (32g)", "1 small bottle (180ml)" etc
+	if ($serving =~ /^(.*[ \(])?(?<quantity>(\d+)(\.|,)?(\d+)?)( )?(?<unit>\w+)\b/i) {
```
stephanegigandet, line 1272: Hi @yuktea , we need to be careful: if we don't recognize the unit, we don't want to return a serving size (in fact today we return the value 0).
But the proposed change would mean that something like "43 somethingwedonotunderstand" would result in a serving size of 43 (we would assume the unit is g).

One way to catch this would be to add tests like those:

is( normalize_serving_size("43 unknownthingorunit"), 0 );
is( normalize_serving_size("43 unknownthingorunit (200g)"), 200 );

lib/ProductOpener/Food.pm
```diff
@@ -1260,21 +1265,45 @@ normalize_serving_size(2.5kg)              returns 2500
 =cut
 
 sub normalize_serving_size($) {
-
 	my $serving = shift;
 
-	my $q = 0;
-	my $u;
-
-	if ($serving =~ /((\d+)(\.|,)?(\d+)?)( )?($units)\b/i) {
```
stephanegigandet, line 1269: One easy solution would be to add grams to $units, which is defined at the top of Food.pm:

my $international_units = qr/kg|g|mg|µg|oz|l|dl|cl|ml|(fl(\.?)(\s)?oz)/i;
# Chinese units: a good start is https://en.wikipedia.org/wiki/Chinese_units_of_measurement#Mass
my $chinese_units = qr/
	(?:[\N{U+6BEB}\N{U+516C}]?\N{U+514B})|  # 毫克 or 公克 or 克 or (克 kè is the Chinese word for gram)
	                                        #                      (公克 gōngkè is for "metric gram")
	(?:\N{U+516C}?\N{U+65A4})|              # 公斤 or 斤 or         (公斤 gōngjīn is a "metric kg")
	(?:[\N{U+6BEB}\N{U+516C}]?\N{U+5347})|  # 毫升 or 公升 or 升     (升 is liter)
	\N{U+5428}                              # 吨                    (ton?)
	/ix;
my $russian_units = qr/г|мг|кг|л|дл|кл|мл/i;
my $units = qr/$international_units|$chinese_units|$russian_units/i;

yuktea: Are we going to support names for other units besides grams? I feel this easier solution should be passable with just grams but maybe hard to maintain with aliases for other units too.

stephanegigandet: Ideally, we should support all units, in all languages. eg. "50 grammes" in French. We could add the unit names in common/common.pot so that they get translated in all languages.

lib/ProductOpener/Food.pm
```diff
@@ -1260,21 +1265,45 @@ normalize_serving_size(2.5kg)              returns 2500
...
+		return unit_to_g($q, $u);
 	}
 
 	#$log->trace("serving size normalized", { serving => $serving, q => $q, u => $u }) if $log->is_trace();
-	return $q;
+	return 0;
+}
+
+# @todo we should have equivalences for more units if we are supporting this
+my %unit_equivalence_map = (
```
stephanegigandet, line 1285: Maybe instead of a hash, we could do a list, as we will iterate over all values anyway.
That way we could have multiple regexps that correspond to a single unit.
That could be useful if we automatically add regexps from translations.

Something like:

```
my @unit_equivalences = (
('g', qr/gram(s)?/),
('g',qr/gramme(s)?/), # French
);
```

lib/ProductOpener/Food.pm
```diff
@@ -399,8 +399,14 @@ sub unit_to_g($$) {
 	$unit eq 'fl oz' and return $value * 30;
 
 	# return value without modification if it's already grams or 克 (kè) or 公克 (gōngkè) or г
-	return $value + 0; # + 0 to make sure the value is treated as number
+	# We return with + 0 to make sure the value is treated as number
 	# (needed when outputting json and to store in mongodb as a number)
+	(($unit eq 'g') or ($unit eq '克') or ($unit eq '公克') or ($unit eq 'г')
+		or ($unit eq 'мл') or ($unit eq '\N{U+6BEB}\N{U+5347}') or ($unit eq 'ml')
```
stephanegigandet, line 405:
```suggestion
		or ($unit eq 'мл') or ($unit eq "\N{U+6BEB}\N{U+5347}") or ($unit eq 'ml')
```

Strings within single quotes will not convert unicode characters sequence to the corresponding character, so we need to use double quotes here.

yuktea: Thanks for catching! I'll add some testcase so we can catch this.
/todo

lib/ProductOpener/Food.pm
yuktea: Agreed. I think this is more scalable and optimal.

lib/ProductOpener/Food.pm
yuktea: Apologies for my absence over the last day, I'll work on adding the translations and then we could maybe also work on an autogeneration logic for the unit equivalence table/list.

lib/ProductOpener/Food.pm
```diff
@@ -1255,27 +1259,51 @@ sub normalize_quantity($) {
...
-	my $u;
-
-	if ($serving =~ /((\d+)(\.|,)?(\d+)?)( )?($units)\b/i) {
-		$q = lc($1);
-		$u = $6;
+	# Regex captures any <number>( )?<unit-identifier> group, but leaves allowances for a preceding
+	# token to allow for patterns like "One bag (32g)", "1 small bottle (180ml)" etc
+	if ($serving =~ /^(.*[ \(])?(?<quantity>(\d+)(\.|,)?(\d+)?)( )?(?<unit>\w+)\b/i) {
+		my $q = $+{quantity};
+        my $u = normalize_unit($+{unit});
```
stephanegigandet, line 1274: Do we still need a normalize_unit() function if the translations are in %unit_conversion_map ?

lib/ProductOpener/Food.pm
```diff
@@ -367,6 +366,39 @@ unit_to_g(520,mg) => returns 0.52
 
 =cut
 
+# This is a key:value pairs
+# The keys are the unit names and the values are the multipliers we can use to convert to a standard unit. 
+# We can divide by these values to do the reverse ie, Convert from standard to non standard
+my %unit_conversion_map = (
```
stephanegigandet, line 372: I like this approach. I think it could be improved by separating the translation data from the conversion data.

e.g. we could have two maps:

%unit_translation_map = (
"gram" => "g",
"grams" => "g",
"gramme" => "g",
"grammes" => "g",
);

And then a conversion map:

%unit_conversion_map = (
"g" => 1,
"kg" => 1000,
);

But we can do it in another PR.

lib/ProductOpener/Food.pm
```diff
@@ -430,24 +451,16 @@ sub g_to_unit($$) {
...
+	if (exists($unit_conversion_map{$unit})) {
+		return $value/$unit_conversion_map{$unit};
+	}
 
-	# l = 公升 - gōngshēng = л = liter
-	(($unit eq 'l') or ($unit eq "\N{U+516C}\N{U+5347}") or ($unit eq 'л')) and return $value / 1000;
-	(($unit eq 'dl') or ($unit eq 'дл')) and return $value / 100;
-	(($unit eq 'cl') or ($unit eq 'кл')) and return $value / 10;
-	$unit eq 'fl oz' and return $value / 30;
+	(($unit eq 'kcal') or ($unit eq 'ккал')) and return int($value / 4.184 + 0.5);
```
ditsuke, line 459: You can avoid exceptional handling of these units by mapping them to lambdas in the `%unit_conversion_map`

lib/ProductOpener/Food.pm
yuktea: I agree this approach is nice, will do it in another PR

lib/ProductOpener/Food.pm
```diff
@@ -380,28 +412,17 @@ sub unit_to_g($$) {
...
-	(($unit eq 'cl') or ($unit eq 'кл')) and return $value * 10;
-	$unit eq 'fl oz' and return $value * 30;
+	(($unit eq 'kcal') or ($unit eq 'ккал')) and return int($value * 4.184 + 0.5);	
 
-	# return value without modification if it's already grams or 克 (kè) or 公克 (gōngkè) or г
-	return $value + 0; # + 0 to make sure the value is treated as number
-	# (needed when outputting json and to store in mongodb as a number)
+	# We return with + 0 to make sure the value is treated as number (needed when outputting json and to store in mongodb as a number)
+	# lets not assume that we have a valid unit
+	return 0;
```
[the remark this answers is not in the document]
yuktea: @stephanegigandet The updated PR has tests for these units

stephanegigandet (approved):
Looks good, thank you

PR #6883, opened 2022-06-10
fix: use correct unit for nutrients expressed in IU or %DV 

### LINKED ISSUE #6882: Nutrient values specified in %DV are not converted correctly
### Describe the bug

Minerals and vitamins that have a %DV value are not correctly converted to grams.

This is a regression introduced when we switched to using the nutrients taxonomy in Food.pm

https://github.com/openfoodfacts/openfoodfacts-server/commit/86d1247288cfe6ce64b1fccbd9d2d3aff16dac96#diff-351be5a181a0bf67d0ff930eae800771588040be0c2a334a98412e378da0b37aL245

### To Reproduce

Go to https://us.openfoodfacts.org
Add a value for calcium with the % DV unit.

### Expected behavior

The correct value in grams should be computed.

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Other

### Browser version

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
I made a mistake 6 months ago when removing nutrients from Food.pm to use the nutrients taxonomy instead, and the nutrients added with %DV (daily values) or IU (International Units) were not converted correctly to grams.

This fixes it and adds some tests.

Fixes #6882

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -288,13 +288,14 @@ sub assign_nid_modifier_value_and_unit($$$$$) {
 		$product_ref->{nutriments}{$nid . "_unit"} = $unit;
 		$product_ref->{nutriments}{$nid . "_value"} = $value;
 		
+		# Convert values passed in international units IU or % of daily value % DV to the default unit for the nutrient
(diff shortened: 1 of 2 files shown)


alexgarel (approved):
Cool that you added a test.

PR #6887, opened 2022-06-10
feat: suggest fuzzy + limit as a param

alexgarel:
fixes: https://github.com/openfoodfacts/openfoodfacts-dart/issues/465

Also add fuzzy searching.

### DIFF
--- a/cgi/suggest.pl
+++ b/cgi/suggest.pl
@@ -151,20 +151,19 @@ else {
 		}
 
 		if (defined $tag) {
-			# matching at start
+			# matching at start, best matches
 			if ($tagid =~ /^$stringid/) {
 				push @suggestions, $tag;
+				# only matches at start are considered
 				$suggestions_count++;
 			}
 			# matching inside
 			elsif ($tagid =~ /$stringid/) {
 				push @suggestions_c, $tag;
-				$suggestions_count++;
 			}
 			# fuzzy match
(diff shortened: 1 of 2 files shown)


alexgarel:
It would be cool to add tests, but I don't know if I have the time because it will require making a sub to be able to test with  a fake taxonomy.

monsieurtanuki (approved):
Thank you @alexgarel!
If I sum up, now there's
* the fuzzy search
* a new optional `limit` parameter (default is `25`, max is `400`)

cgi/suggest.pl
```diff
@@ -42,84 +42,146 @@
...
+			# matching at start
+			if ($tagid =~ /^$stringid/) {
 				push @suggestions, $tag;
+				$suggestions_count++;
 			}
-			else {
+			# matching inside
+			elsif ($tagid =~ /$stringid/) {
 				push @suggestions_c, $tag;
+				$suggestions_count++;
```
stephanegigandet, line 162: One issue with doing that (which was already present before) is that we will hit the limit with entries that don't start with the searched string, even though we may have some.

e.g. just searching with one letter:

image

So we could create a separate counter for suggestions matching inside instead.

alexgarel:
PR updated to only count for "start with" match to limit traversal.

stephanegigandet (approved):

PR #6888, opened 2022-06-10
taxonomy: Arancini

moon-rabbitOFF:


taxonomies/categories.txt
```diff
@@ -5172,21 +5172,27 @@ ciqual_food_name:en:Fruit juice with milk
 ciqual_food_name:fr:Boisson au jus de fruit et au lait
 
 <en:Fruit-based beverages
+<en:Plant milks
```
stephanegigandet, line 5175: I'm not sure we want beverages that have some plant milks to be under plant milks. It would be like putting "Chocolate milks" under "Milks".

moon-rabbitOFF: We currently have "Chocolate soy milks" under "Soy milks" and it's the same for the other plant milk. The issue is solved by having categories like "Unsweetened natural soy milks". The difference with real milks is that plant milks contain two ingredients at least (water and the plant) and almost always contain another ingredient such as salt and in a lesser extend sugar, algae, or additives. So it is more difficult to separate the plain and flavoured plant milk since the plain ones are already composed ingredient.

moon-rabbitOFF: I am also wondering if we should rename all the "natural" plant milks into "plain" plant milks

stephanegigandet: OK, you're the expert. :)

PR #6890, opened 2022-06-12
fix: Invalid conversion

dipietroR:
I would get a warning message:
`[Sun Jun 12 19:52:11 2022] -e: Invalid conversion in sprintf: end of string at /opt/product-opener/lib/ProductOpener/Display.pm line 8853.`
When clicking on http://world.openfoodfacts.localhost/product/3661344653573/yaourt-cremeuh-cafe-les-2-vaches

I created a makeshift test and realized that the second variable, 
`$product_ref->{nutrition_score_warning_fruits_vegetables_nuts_estimate_from_ingredients_value});`
didn't seem to be printed with the current syntax, so I removed it. 

`lang("nutrition_grade_fr_fruits_vegetables_nuts_estimate_from_ingredients_warning"), $product_ref->{nutrition_score_warning_fruits_vegetables_nuts_estimate_from_ingredients_value}` 
Sometimes wouldn't have anything in it, which is why the warning message appeared, so I made so that if the variable consists of nothing, it would be added in blank.


dipietroR:
How's this?

stephanegigandet (approved):
Looks good, thank you. I'm just going to add the corresponding changes in en.po (and fr.po so that we don't have to wait for translations from Crowdin).

PR #6905, opened 2022-06-15
feat: extend ingredients analysis panels with details and call for help

### LINKED ISSUE #6904: Extend ingredients analysis panel to add details of the analysis + info on how to improve it

We need to turn the information that is in the "Details of the ingredients analysis" dropdown into panels so that it can also be displayed inside the app.

image

Extend the ingredients analysis panel.

_No response_

### Additional context

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
Fixes #6904

- Created an ingredients_analysis_details panel to show how we analyzed the list of ingredients
- Created an ingredients_analysis_help panel with a call for help, displayed in the subpanels for vegan / vegetarian / palm oil free when we have unknown ingredients, and in the ingredients_analysis_details panel.
- Minor changes on ingredients / nutrition icons
- Added new types for the text element of knowledge panels: callout_info + callout_warning

Display on the website:

image

Display on the app:

image

image

image


templates/api/knowledge-panels/health/ingredients/ingredients_analysis_property.tt.json
```diff
@@ -1,4 +1,6 @@
 [% SET property = panel.property %]
+[% property_unknown_ingredients = property %]
+[% property_unknown_ingredients = property_unknown_ingredients.replace('-free', '-content-unknown') %]
```
alexgarel, line 3: why not directly ?

```suggestion
[% # name of the unknown ingredients property %]
[% property_unknown_ingredients = property.replace('-free', '-content-unknown') %]
```

alexgarel (approved):
LGTM but some minor remarks

templates/api/knowledge-panels/health/ingredients/ingredients_analysis_property.tt.json
stephanegigandet: You are right, but in fact while I was testing this change, I found that the feature did not output unknown ingredients when the property is otherwise known (e.g. if there's meat, we just display the meat, but not other unknown ingredients that might also be meat).

So I'm changing more things, and this logic will be removed.

Issue #558, opened 2016-12-07
Support for the webshare API


teolemon:
Web sharing

Chrome 55 supports the new Web Share API, which webpages can use to directly share content with apps installed on your device. This means websites can implement their own Share button, which would work exactly like the share button on native Android applications.

However, this API is very much in the early stages. It is only available as an Origin Trial, which means sites have to opt-in to having this feature enabled until the API is complete. To prevent malicious pages tricking the user, only HTTPS links can be shared and can only be triggered via a user gesture (like pressing a button).

PR #6912, opened 2022-06-15
fix: uninitialized value

dipietroR:
Add conditional as without it this will make a dead link resulting in a 404

It is not clear to me what the link in this section does and it is not visible on the page. It may be used for search engines? 
If `$request_ref->{current_link_query_display}` isn't defined, and we select the link in the source, it leads to a 404. As it stands without that defined it appears to be broken, though I'm not certain that we don't want this information here even with a dead link. Perhaps we want something more like `$request_ref->{current_link}`?

### DIFF
--- a/cgi/search.pl
+++ b/cgi/search.pl
@@ -743,7 +743,7 @@ HTML
 		$request_ref->{title} = lang("search_results") . " - " . display_taxonomy_tag($lc,"countries",$country);
 
 
-		#This is uesed to have a special share button on some browsers
+		#This is used to have a special share button on some browsers
 		if (not defined $request_ref->{jqm}) {
 			${$request_ref->{content_ref}} .= <<HTML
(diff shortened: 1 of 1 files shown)


stephanegigandet:
This code was to have a special share button on some browsers, it was a feature @hangy added 6 years ago.
See: https://github.com/openfoodfacts/openfoodfacts-server/issues/558

On Android on Chrome, it displays a "Share" button at the bottom of the screen.

We should fix the link though. It would be good to add some comments so that we remember what it is, I had completely forgotten about it.

image

dipietroR:
OK, this should give the code a functional link. Does this seem right to you?

stephanegigandet:
> OK, this should give the code a functional link. Does this seem right to you?

Yes, definitely!

stephanegigandet (approved):
Thanks!

PR #6913, opened 2022-06-16
refactor: comments and small refactor in build_taxonomy

alexgarel:
This PR is a documentation PR, but I couldn't refrain me from doing some refactor (code that was repeated 3 times !).


alexgarel:
@stephanegigandet, I tested it comparing to old version (I had introduced two bugs, indeed).

Can you do a final review. Also note there are todos and fixmes that you might be able to answer.

lib/ProductOpener/Tags.pm
```diff
@@ -807,33 +897,73 @@ sub build_tags_taxonomy($$$) {
...
 	$properties{$tagtype} = {};
 
 	my $errors = '';
 
 	if (open (my $IN, "<:encoding(UTF-8)", "$data_root/taxonomies/$file")) {
 
-		my $current_tagid;
-		my $current_tag;
-		my $canon_tagid;
+		my $line_tagid;  # tracks the tag normalized form
```
stephanegigandet, line 964: $line_tagid is the normalized form of $current_tag, so they should have the same prefix.

Maybe we can be more specific:

my $lc_tag; # Main name of a tag in a specific language (display form) - e.g. "Café au lait"
my $lc_tagid; # Main id of a tag in a specific language (normalized form) - e.g. "cafe-au-lait"
my $canon_tagid; # Canonical id of the tag (main language prefix + normalized form in the main language) - e.g. "en:coffee-with-milk"

stephanegigandet: I would not use "line" in the variable name.

lib/ProductOpener/Tags.pm
```diff
@@ -782,11 +782,101 @@ sub remove_plurals($$) {
...
+	$line =~ s/\(((i|v|x)+)\)/$1/i;
+
+	# strip spaces at end of line
+	$line =~ s/\s+$//;
+
+	return $line
+}
+
+
+=head2 search_tag_c( $tag, $synonyms, $tagtype, $lc, $warning )
```
stephanegigandet, line 824: I would rename the function to get_lc_tagid(), and change the order of parameters to ($lc, $tagtype, $tag) as it is the order we have in all the other functions like canonicalize_taxonomy_tag()

lib/ProductOpener/Tags.pm
```diff
@@ -782,11 +782,101 @@ sub remove_plurals($$) {
...
+
+If empty, no warning will be displayed.
+
+=head3 return str - found current tagid or undef
+
+=cut
+sub search_tag_c($$$$$)
+{
+	my $tag = shift;
+	my $synonyms = shift;
```
stephanegigandet, line 850: The naming convention we have is to suffix all references with _ref, so it would be $synonyms_ref.

lib/ProductOpener/Tags.pm
```diff
@@ -807,33 +897,73 @@ sub build_tags_taxonomy($$$) {
...
 	defined $tags_images{$lc}{$tagtype} or $tags_images{$lc}{$tagtype} = {};
 
+	# we ofen use the term *tag* in the code to indicate a single entry between commas
+	# that is most lines, are tags separated by commas.
+
+	# when we speak about normalized entry, or tagid,
+	# it's the tag value where we lowercased, replace separators by dash etc.,
+	# see get_string_id_for_lang
+
+	# line tagid (or tagid_c) in particular is the first entry in a line
```
stephanegigandet, line 907: I would change both $line_tagid and $tagid_c to $lc_tagid, as the important part is that it's about a specific language.

lib/ProductOpener/Tags.pm
```diff
@@ -1048,176 +1148,207 @@ sub build_tags_taxonomy($$$) {
 		# en:banana yogurts
 		#
 		# --> also compute banana yoghurts
+		# Note that tris does not happens on tag string but on tagid (banana-yoghurts)
```
stephanegigandet, line 1151:
```suggestion
		# Note that this does not happen on tag string but on tagid (banana-yoghurts)
```

lib/ProductOpener/Tags.pm
```diff
@@ -1048,176 +1148,207 @@ sub build_tags_taxonomy($$$) {
...
-		}
+		}  # on of pass loop
 
 
 		# add more synonyms: remove stopwords and deal with simple plurals
 		# -> should not be done on some taxonomies that contain only proper names
+		# TODO we could mark this kind of thing in a header for taxonomy
 		if (($tagtype ne "countries") and ($tagtype ne "origins")) {
 
+			# TODO: explan why we do it in synonyms and not in extended synonyms
```
stephanegigandet, line 1334: In fact %synonyms (used for matching) contains all keys from %synonyms_for and %synonyms_for_extended

lib/ProductOpener/Tags.pm
alexgarel: ok, I think clarity is important.

lib/ProductOpener/Tags.pm
alexgarel: Ooooh this is good to note above, I will do.

stephanegigandet (approved):
Thanks a lot for all the commenting and refactoring!

Overall it looks very good to me, but I suggested names that I think are more logical for some variables and functions.

It might also be useful to add unit tests for the new function that returns the tagid of a tag in a specific language.

lib/ProductOpener/Tags.pm
alexgarel: done (next commit)

lib/ProductOpener/Tags.pm
alexgarel: done (next commit)

lib/ProductOpener/Tags.pm
alexgarel: done (next commit)

lib/ProductOpener/Tags.pm
alexgarel: Haha, it was already explained :-D

lib/ProductOpener/Tags.pm
alexgarel: done (next commit)

t/tags_unit.t
```diff
@@ -0,0 +1,95 @@
...
+use ProductOpener::Tags qw/get_lc_tagid/;
+
+=head1 Some unit tests for Tags.pm module
+
+=cut
+
+=head2 Unit testing get_lc_tagid
+=cut
+
+# a fake remove stopwords
```
stephanegigandet, line 18: So this is just so that we don't have to load Tags.pm to run the test, correct? Maybe we could say so in a comment if that's the case.

t/tags_unit.t
```diff
@@ -0,0 +1,95 @@
...
+        "en" => {
+            "salted-snacks" => "en:salted-snacks",
+            "salty-snacks" => "en:salted-snacks",
+            "salty-snack" => "en:salted-snacks",
+        }
+    };
+
+    my $lc_tagid;
+
+    $lc_tagid = get_lc_tagid($synonyms_ref, "en", "category", "salted-snacks", "");
```
stephanegigandet, line 46: Not very important, but "category" should be "categories".

t/tags_unit.t
alexgarel: It's also because I won't my test to be predictable. While here it would depend on the taxonomy file I don't have the hand over.
But yes I should explain why :-)

t/tags_unit.t
alexgarel: ok, better stay close to reality, it helps understand better.

PR #6914, opened 2022-06-16
fix: $request_ref->{count} not used

dipietroR:
$request_ref->{count} shows up 4 times in the code. Though it never appears to be used as three of the times that it appears it is only to define it. The remaining instance that it appears such that it is referenced, does not seem to benefit from the previous three instances where the value is defined. As such this appears to be unused. This commit removes all four instances.


stephanegigandet (approved):
Looks good!

PR #6917, opened 2022-06-17
test: unit test for url_for_text

dipietroR:
$lc appears to default to German (de), though how would we test other language codes? Changing $lc in display.t doesn't seem to have an effect. How might we test other languages?


stephanegigandet:
> Changing $lc in display.t doesn't seem to have an effect. How might we test other languages?

$lc is a global variable from the Lang.pm module, it's usually initialized by Display.pm init() function.

In display.pm, you can set it directly: $lc = "en"; for instance.

PR #6923, opened 2022-06-18
refactor: Move inline javascript in export_products.pl

yuktea:
### What
There was a lot of inline JS in export_products.pl, this PR moves that.

### DIFF
--- a/templates/web/pages/export_products/export_products.tt.js
+++ b/templates/web/pages/export_products/export_products.tt.js
@@ -1,0 +1,72 @@
+var poll_n1 = 0;
+var timeout1 = 5000;
+var job_info_state1;
+
+var poll_n2 = 0;
+var timeout2 = 5000;
+var job_info_state2;
+
+var poll_n3 = 0;
+var timeout3 = 5000;
+var job_info_state3;
+
+var minion_status = {
+	"inactive" : "$Lang{minion_status_inactive}{$lc}",
+	"active" : "$Lang{minion_status_active}{$lc}",
+	"finished" : "$Lang{minion_status_finished}{$lc}",
+	"failed" : "$Lang{minion_status_failed}{$lc}"
+};
+
+(function poll1() {
+    \$.ajax({
+      url: '/cgi/minion_job_status.pl?job_id=$local_export_job_id',
+      success: function(data) {
+        \$('#result1').html(minion_status[data.job_info.state]);
+        job_info_state1 = data.job_info.state;
+      },
+      complete: function() {
+        // Schedule the next request when the current one's complete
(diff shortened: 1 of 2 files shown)


templates/web/pages/export_products/export_products.tt.js
```diff
@@ -0,0 +1,72 @@
...
+var poll_n2 = 0;
+var timeout2 = 5000;
+var job_info_state2;
+
+var poll_n3 = 0;
+var timeout3 = 5000;
+var job_info_state3;
+
+var minion_status = {
+	"inactive" : "$Lang{minion_status_inactive}{$lc}",
```
stephanegigandet, line 14: Inside the templates, there isn't a direct access to variables like $Lang{minion_status_inactive}{$lc} but we can access a few functions available to templates: [% lang{minion_status_inactive") %]

templates/web/pages/export_products/export_products.tt.js
```diff
@@ -0,0 +1,72 @@
...
+	"active" : "$Lang{minion_status_active}{$lc}",
+	"finished" : "$Lang{minion_status_finished}{$lc}",
+	"failed" : "$Lang{minion_status_failed}{$lc}"
+};
+
+(function poll1() {
+    \$.ajax({
+      url: '/cgi/minion_job_status.pl?job_id=$local_export_job_id',
```
stephanegigandet, line 22: Same here, $local_export_job_id won't be accessible through the template, unless it has been added to the data $template_data_ref that was passed to the template.

stephanegigandet:
I'm not sure if you have tried running the cgi and if it works? It is specific to the "producers platform", which is a separate instance of Product Opener. To run it, you have to set up some environment variables (it's described in https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/how-to-guides/docker-developer-guide.md )

Also as discussed this morning, it would probably be better to use a static JS file instead of generating JS code with a template, as we only want to pass a few parameters.

yuktea:
> I'm not sure if you have tried running the cgi and if it works? It is specific to the "producers platform", which is a separate instance of Product Opener. To run it, you have to set up some environment variables (it's described in https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/how-to-guides/docker-developer-guide.md )
> 
> Also as discussed this morning, it would probably be better to use a static JS file instead of generating JS code with a template, as we only want to pass a few parameters.

Working on the fixes!

yuktea:
not working as intended :(

teolemon:
@yuktea should I close ? Do you want to try rebasing ?
ping @stephanegigandet @alexgarel

stephanegigandet:
Hello Yukti,

PR #6935, opened 2022-06-21
fix: syntax on display_icon

dipietroR:
This function was using a different syntax than what appears to be used in most other functions. This change updates to both require one argument (this seems to be a requirement). And was using @_ rather than shift.


hangy (commented):
AFAIK, sub prototypes are discouraged, and `@_` is faster than `shift`

stephanegigandet (commented):
Looks good, but could you also add a documentation header for the function, to explain what it does, what it returns etc. ?

dipietroR:
How does this look?

stephanegigandet:
> AFAIK, sub prototypes are discouraged, and `@_` is faster than `shift`

I've read the discussions like https://stackoverflow.com/questions/297034/why-are-perl-5s-function-prototypes-bad but the prototypes we currently have to specify the number of arguments have been useful (at least to me) in the past. e.g. when I added a parameter to a function, but forgot to change all the calls to add the parameter.

It's clearly not perfect, but making sure we call functions with the right number of parameters is better than nothing I think.

Well maybe it's time to start to use Perl's function signatures. :)

dipietroR:
I've tried reading about prototypes and they seem to cause specific argument setups to be passed to another function, but I ultimately don't understand it. It seems like I should revert the syntax changes in this PR and leave the comments. Does that sound right?

stephanegigandet:
@dipietroR It's a complex topic indeed.

@hangy I'm going to merge this PR as-is, as it makes the code more consistent. But I opened a new bug to discuss prototypes and signatures: https://github.com/openfoodfacts/openfoodfacts-server/issues/6956

PR #6936, opened 2022-06-21
test: display_icon unit tests

dipietroR:
unit tests for display_icon

### DIFF
--- a/t/display.t
+++ b/t/display.t
@@ -115,15 +115,19 @@ my $nutriscore_data_ref = {
           'saturated_fat_points' => 3
         };
 
-print("-----------\n");
 like ( display_nutriscore_calculation_details($nutriscore_data_ref), qr/Score nutritionnel: 6/);
 like ( display_nutriscore_calculation_details($nutriscore_data_ref), qr/Proteins:\n2&nbsp;<\/strong>\/&nbsp;5/);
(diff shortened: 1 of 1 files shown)


t/display.t
```diff
@@ -115,15 +115,19 @@ my $nutriscore_data_ref = {
...
 like ( display_nutriscore_calculation_details($nutriscore_data_ref), qr/Proteins:\n2&nbsp;<\/strong>\/&nbsp;5/);
 like ( display_nutriscore_calculation_details($nutriscore_data_ref), qr/Positive points: 2/);
 like ( display_nutriscore_calculation_details($nutriscore_data_ref), qr/Negative points: 8/);
 like ( display_nutriscore_calculation_details($nutriscore_data_ref), qr/<strong>Nutri-Score: C<\/strong>/);
-print("-----------\n");
 
+# Test that display_icon successfully creates the cc and info icons
+use Digest::MD5  qw(md5_hex);
+my $cc_string = display_icon('cc');
+is(md5_hex($cc_string), '64b2e2b0a3f8a7a27f7e851c43333a8c');
```
stephanegigandet, line 127: This test is likely to break if the icon gets updated, or if the SVG optimizer that is run automatically on icons is updated. But it won't be very useful information and might in fact be very confusing.

I'm not sure having a test for the display_icon function is really useful, as the function itself is unlikely to break: there's no real logic in it, it just reads a file and outputs it.

PR #6938, opened 2022-06-22
refactor: display_form not used

dipietroR:
display_form doesn't seem to be used, no other functions call for it.


stephanegigandet (approved):
Great, we should definitely remove dead code when we find it. Thank you!

Issue #6540, opened 2022-04-01
India (tracker)


teolemon:
### What
- [ ] #1107 
- [ ] #6536 
### Why
- Expand the Open Food Facts community in India

### Part of

borrokk:
would love to work on this issue, commenting for future contributions :)

PR #6939, opened 2022-06-22
fix: "refined wheat flour" and "maida" should be recognized as a type of flour

yuktea:
### What
- Refined wheat flour and _maida_ are two terms that are used interchangeably to refer to white colored/all-purpose flour in the Indian subcontinent.

- Related Discussion #6540


stephanegigandet (approved):
Thank you!

PR #6950, opened 2022-06-23
taxonomy: Finnish products

moon-rabbitOFF:


taxonomies/categories.txt
```diff
@@ -44294,6 +44299,14 @@ sv:Skyr
 zh:Skyr
 wikidata:en:Q742127
 
+<en:Skyr
```
stephanegigandet, line 44302: Can we put "Skyrs" in plural, and then Skyr in singular in all the related entries?

Same for Quark?

moon-rabbitOFF: I modified it

stephanegigandet (approved):

PR #6953, opened 2022-06-24
refactor: remove canonicalize_saint

dipietroR:
This function doesn't seem to be used, it doesn't get called nor does it call other functions.


stephanegigandet (approved):
Thanks!

PR #6960, opened 2022-06-24
refactor: remove dims

dipietroR:
This function appears to only be defined, it isn't used anywhere and it doesn't call other functions.

### DIFF
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -500,12 +500,6 @@ sub process_search_image_form($) {
 }
 
 
-sub dims {
-	my ($image) = @_;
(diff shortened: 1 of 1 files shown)


dipietroR:
Commit 419bcb7ab5af16952dd23fd345f3b0cfbbe051d3 seems to be when this function was added and 617d7a12f26cc6c8b7016d8f6bbe47eb0e2af198 deletes the call to this function

stephanegigandet:
Hi @hangy , do you remember why this function was added, and whether we should keep it or remove it? Thank you!

stephanegigandet:
I completely forgot about that. Let's remove it indeed. :)

stephanegigandet (approved):

PR #6258, opened 2022-01-07
(WIP) feat: Samples for key countries and backport

teolemon:
- [ ] Work in progress
- [ ] Samples for key countries and backport from typical products spreadsheet.
- [ ] TODO: add ways for the community to get hold of that topic (wiki links, post…)

PR #6962, opened 2022-06-24
feat: add perltidy

dipietroR:
I was thinking of adding PerlTidy to the code because the code doesn't appear to have a consistent formatting. It seems to change depending on who works on it and when they worked on it. In this initial commit I have an example of how the code looks with the default settings of perltidy. 

If are tweaks to the formatting that are desired this site has a good tutorial on what options perltidy has:
http://perltidy.sourceforge.net/tutorial.html


stephanegigandet:
The output is interesting. :)

Maybe we could first try to put settings that more or less match what we have in most of the code.

e.g.

```
-bt,-pt,-sbt: Container tightness
These are parameters for controlling the amount of space within containing parentheses, braces, and square brackets. The example below shows the effect of the three possible values, 0, 1, and 2, for the case of parentheses:

 if ( ( my $len_tab = length( $tabstr ) ) > 0 ) {  # -pt=0
 if ( ( my $len_tab = length($tabstr) ) > 0 ) {    # -pt=1 (default)
 if ((my $len_tab = length($tabstr)) > 0) {        # -pt=2
```

--> our current style is the last one, with no extra spaces

We have some settings for editing software that are in the .editorconfig file:

```
# Default for all files:
# * UTF8 charset, unix-style newlines (lf), 2 spaces indentation
# * ensure files ends with a newline when savings (insert_final_newline)
# * remove any whitespace characters preceding newline characters (trim_trailing_whitespace)
[*]
charset = utf-8
end_of_line = lf
indent_style = space
indent_size = 2
insert_final_newline = true
trim_trailing_whitespace = true

# Perl files: tab indentation (4 spaces)
[*.{pm,pl,t}]
indent_style = tab
indent_size = 4
```

So it would probably be useful to have the same for perltidy.

Then it would be good to have other people's opinion, so that we can decide on which options we want. cc @alexgarel @yuktea @hangy

dipietroR:
Tightened up some of the block spacing. Tinkered around with line column limits, 80, or 78 are pretty standard, but there are some long lines here that don't look very good when we limit it to that. Since most screens handle more than 80 columns now, 120 seemed like an ok balance. Though at the same time the large column limit results in some lists looking a little less tidy, still, in my view, overall better than before. One potential way to improve this is to move from tabs to spaces, is there a reason that we are using tabs over spaces?

stephanegigandet:
Thanks @dipietroR , it looks better to me like this. :)

Tabs vs spaces: well the debate has been raging for the last 30 years or so... ;)

If there's a strong case (or even just a case) to switch to spaces, why not.

Let's also wait for @alexgarel to come back.

.perltidyrc
```diff
@@ -0,0 +1,27 @@
+-et=4
+-dt=4
+-tabs
```
Ban3, line 3:
```suggestion
```
The documentation states that `-et` is preferred over `--tabs`, so I think it's safe to remove this. (I don't know if it ever even did anything because of the missing `-` :))

vivian-rook:
Would appear that http://security.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1-1ubuntu2.1~18.04.17_amd64.deb
no longer exists. Will track down another copy.

dipietroR:
This file:  ./scripts/obsolete/import_us_ndb.pl
Seems to have formatting errors that cause perltidy to fail. It's in a directory called obsolete, can we just delete it?

dipietroR:
We can edit the formatting rules more as we like @alexgarel could use your input here.

When we're happy with the formatting style. I think to update all the perl files we should do the following:
1. Have a list of all the perl files. (I've put this list in this PR)
2. Uncomment a file one at a time to reformat it, merge it in.
3. If a file is being worked on (has a PR), merge the reformatted version of said file and fix formatting on the existing PR
4. When the list of files is all reformatted and updated, delete the list and put the final code (the one commented out at the bottom)
5. The final code will be searching for all the perl files, it will find any new files that weren't part of the list; the formatting in those files will be fixed at this point.
      One example could be Module.pm from https://github.com/openfoodfacts/openfoodfacts-server/pull/7009 which isn't in the list of files.

I've made a bit of code to identify when a file is open in another PR. There are relatively few (40) PRs that are open with edits to a perl file. With 38 different files being. For instance 
lib/ProductOpener/Config_off.pm is only open in [pr-6258 ](https://github.com/openfoodfacts/openfoodfacts-server/pull/6258) . As such Config_off.pm could be updated in pr-6258 when the Config_off.pm file is updated. We might want to discuss further which PRs we want to update, as:
- some have existing merge conflicts that could be difficult to understand
- some, like 6258, are marked as a draft
- PRs from forks that do not allow edits.
- and some are years old

I've left a line for perltidy to find in search.pl to show that the testing is indeed finding it, before merging this, that line would need updated.

.github/workflows/pull_request.yml
```diff
@@ -50,6 +50,360 @@ jobs:
...
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 50
+    - name: install cpanminus to install perltidy
+      run: sudo apt-get install -y cpanminus
+    - name: install perltidy
+      run: sudo cpanm Perl::Tidy
+    - name: temp list of perl files
+      run: |
+        cat << EOF > /tmp/perlfiles
```
alexgarel, line 66: I would instead use an exclusion list, rather than a list, we will forgot to update it.

I think of (taken from https://stackoverflow.com/a/22558474/2886726)
```
find . -name "*.pm" -or -name "*.pl" | grep -vFf .perltidy_skip
```

`.perltidy_skip` would be a file in the gitrepo

At the end you can have a single line:
```
find  . -name "*.pm" -or -name "*.pl" | grep -vFf .perltidy_skip|xargs perltidy --assert-tidy 
```

cgi/search.pl
```diff
@@ -24,18 +24,18 @@
...
+use ProductOpener::Store         qw/:all/;
+use ProductOpener::Index         qw/:all/;
+use ProductOpener::Display       qw/:all/;
+use ProductOpener::Users         qw/:all/;
+use ProductOpener::Products      qw/:all/;
+use ProductOpener::Food          qw/:all/;
+use ProductOpener::Tags          qw/:all/;
 use ProductOpener::PackagerCodes qw/:all/;
-use ProductOpener::Text qw/:all/;
+use ProductOpener::Text          qw/:all/;
```
alexgarel, line 38: I'm really not keen on this kind of tabular alignment as it is really hard to maintain for a human, and it does not bring much in my opinion.

cgi/search.pl
```diff
@@ -55,20 +55,23 @@
 
 if (user_agent() =~ /apps-spreadsheets/) {
 
-	display_error("Automated queries using Google Spreadsheet overload the Open Food Facts server. We cannot support them. You can contact us at contact\@openfoodfacts.org to tell us about your use case, so that we can see if there is another way to support it.", 200);
+	display_error(
+		"Automated queries using Google Spreadsheet overload the Open Food Facts server. We cannot support them. You can contact us at contact\@openfoodfacts.org to tell us about your use case, so that we can see if there is another way to support it.",
+		200
+	);
```
alexgarel, line 61: I like this kind of parenthesis handling :+1:

cgi/search.pl
```diff
@@ -98,93 +101,118 @@
...
+	countries            => 1,
+	additives            => 1,
+	states               => 1,
+	editors              => 1,
+	languages            => 1
+);
+
+my @search_ingredient_classes = (
+	'additives',                             'ingredients_from_palm_oil',
+	'ingredients_that_may_be_from_palm_oil', 'ingredients_from_or_that_may_be_from_palm_oil'
```
alexgarel, line 131: There again I find it really hard to maintain for a human.

.perltidyrc
```diff
@@ -0,0 +1,26 @@
...
+# Tight square brackets
+--square-bracket-tightness=2
+# Tight block braces
+--block-brace-tightness=2
+# No space before semicolons
+--nospace-for-semicolon
+# Don't outdent long quoted strings
+--no-outdent-long-lines
+# Break before all operators
+--break-before-all-operators
```
alexgarel, line 26: As said below, 

> I'm really not keen on this kind of tabular alignment as it is really hard to maintain for a human, and it does not bring much in my opinion.

```suggestion
--break-before-all-operators
# no vertical align
-novalign
```

.github/workflows/pull_request.yml
```diff
@@ -50,6 +50,360 @@ jobs:
       with:
         image: "openfoodfacts-server/backend:dev"
 
+  check_perl_formatting:
```
alexgarel, line 53: It's cool to have written this check.

That said it may be better if all this is part of a Make command and it uses docker, so that we can check it locally and also lint, because if not debugging will be hard !

So we need to add perltidy to cpanfile (in dev section). and add the right targets to Makefile.

alexgarel (changes requested):
Yop, I'm really thrilled to have it, I find automatic formatting is a cool thing to have.

There is my preference on no vertical align, I don't know know what others feel like.

For the make command, I can do it, if you prefer.

alexgarel:
@stephanegigandet ready for review.

stephanegigandet (approved):
Looks good to me, thank you!

alexgarel (approved):

alexgarel:
@dipietroR we did it !

Now next step will be to perltidy massively files and remove them from the .perltidy_exclude file.

Issue #400, opened 2016-08-06
Add facet for "Units of Alcohol"


teolemon:
5 units of alcohol in $ProductName

aleene:
Do not know what you talk about.

I added Alcohol units to the nutrition taxonomy a while ago (UK, FR, AUS, SA).

teolemon:
@hangy recently added units of alcohol calculations to Product Opener. It's related to that, in order to see if we can leverage it to create new pages

hangy:
I added the calculation for #241 the in cc7e7e4b72acbfda1d3bef557d406730f4942c1e. It is not used anywhere right now, because we are missing #247. The UK unit of alcohol is calculated using the quantity in milliliters, so a better understanding of what a quantity like "6 bottles à 33 cl" could be needs to be implemented first.

Issue #241, opened 2016-02-28
Calculate Alcohol units in wines and beers


teolemon:
Based on Volume and Alcohol concentration, we can do a bunch of interesting calculations.

https://en.wikipedia.org/wiki/Unit_of_alcohol

teolemon:
We can calculate it for a glass and a whole bottle.

joeyd15:
hello, can I help assist with this issue?

PR #6972, opened 2022-06-27
refactor: remove compute_units_of_alcohol

dipietroR:
This function doesn't appear to be used. It only seems to get defined and exported, never called.


stephanegigandet:
The function is not used, but it seems useful :) Maybe we could use it to display a "unit of alcohol" knowledge panel.

Do you remember what was the intent behind it @hangy , I think you added it: https://github.com/openfoodfacts/openfoodfacts-server/commit/cc7e7e4b72acbfda1d3bef557d406730f4942c1e

hangy:
> Do you remember what was the intent behind it @hangy , I think you added it: [cc7e7e4](https://github.com/openfoodfacts/openfoodfacts-server/commit/cc7e7e4b72acbfda1d3bef557d406730f4942c1e)

I added it, so it can be used for #241 and #400 later on.

dipietroR:
Neat! Instead of removing it I added documentation for it. :slightly_smiling_face:

hangy (approved):

stephanegigandet (approved):
Thanks!

PR #6973, opened 2022-06-27
refactor: remove create_user

dipietroR:
This function gets defined and exported, however, it's never called.


stephanegigandet:
Hi @dipietroR , thank you! I merged @yuktea 's PR which includes this change, so I will close this PR. Thanks for removing all the dead code!

PR #6974, opened 2022-06-27
refactor: clean up redundant methods from `User.pm`

yuktea:
### What
Cleans up
- `save_user()`
- ` create_user()`
- `user_path()`

These methods are still declared and in the code but they are not being used currently due to revisions to the code over time and creation of some methods like `user_init()`


stephanegigandet (approved):
Thanks

PR #6980, opened 2022-06-29
feat: Nutrient levels knowledge panels with facts and recommendations

stephanegigandet:
This adds the nutrient levels (aka nutrition traffic lights) for fat, saturated fat, sugars and salt, as knowledge panels just below the Nutri-Score panel.

It's a very good place to educate users about those nutrients: why they matter, what to do about them.

As an example, I added facts + recommendations for salt, based on the WHO and the UK NHS recommendations.

We could use default recommendations from the WHO, and then try to work with local health agencies to have more country specific health messages. (for instance in some countries, MSG / Monosodium Glutamate is a huge source of sodium, and something that should be highlighted).

Here is what it would look like on the OFF website:

image


templates/api/knowledge-panels/health/nutrition/nutrient_level.tt.json
```diff
@@ -0,0 +1,142 @@
...
+<li>[% lang('nutrient_info_salt_high_consumption') %]</li>
+</ul>
+<br>
+<b>[% lang("knowledge_panels_recommendation") %]</b>[% sep %]:
+<b>[% lang('nutrient_recommendation_salt_title') %]</b>
+<ul>
+<li>[% lang('nutrient_recommendation_salt_cooking_and_table') %]</li>
+<li>[% lang('nutrient_recommendation_salt_limit_salty_snacks') %]</li>
+</ul>
+`,
```
alexgarel, line 37: I find it a bit annoying that this is defined here so deeply in a technical construct. Shouldn't we put it in an external template (that we could include here) ?

alexgarel (approved):
LGTM but a minor remark (also tested ok)

templates/api/knowledge-panels/health/nutrition/nutrient_level.tt.json
stephanegigandet: I'm not really sure if a new level of indirection would make things simpler or more complex. The initial idea was that the knowledge panels templates was the place where we could define what information is presented and how.

I'll merge as-is and we can refine

PR #6984, opened 2022-06-29
docs: add perl pod documentation to methods in `User.pm`

yuktea:
### What
Added documentation in the Perl pod format complete with the return/argument types


stephanegigandet (approved):
Thanks!

PR #6989, opened 2022-06-30
refactor: remove generate_po_files

dipietroR:
This function looks like it was introduced in 4e142a51831
And hasn't been edited since. It is touched in f3d38267719 and ded3dec42bd though both of those look like syntax cleanup commits that touch many things. The one place I'm finding it being called was commented out shortly after being added in 8f12c82bc49. As such I suspect that this is unused.


hangy (approved):
ACK. The `.po` files are now the blueprint that `Lang.pm` used to be.

stephanegigandet (approved):
This function was used to generate the initial set of .po files, but we won't need to do this again, so we can remove it.

PR #6991, opened 2022-06-30
docs: add documentation for some methods in `user.pm` and `products.pm` 

yuktea:
### What
- Documenting `products.pm` and `user.pm` methods
- removal of a possibly unused sub in `Index.pm`
- proposal to remove[ this ](https://github.com/openfoodfacts/openfoodfacts-server/compare/docs/po-modules?expand=1#diff-d6b9a7ba08e9ecb060d8284b1bd62bcd2861eccd7d3f65f479830b915c679761R2201) (has only been utilized in an obsolete script)

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -241,7 +241,7 @@ sub assign_new_code() {
 =head2 normalize_code()
 
 C<normalize_code()> this function normalizes the product code by:
-- Keeps only digits and removes spaces/dahes etc.
+- Keeps only digits and removes spaces/dashes etc.
 - Normalizes the length by adding leading zeroes or removing the leading zero (in case of 14 digit codes)
 
 =head3 Arguments
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -238,12 +238,30 @@ sub assign_new_code() {
 	return ($code, $product_id);
 }
 
+=head2 normalize_code()
+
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Products.pm
```diff
@@ -304,17 +341,13 @@ e.g. off:[code]
 
 =head3 Parameters
 
-=head4 Owner id
+$Owner id & $code [Product barcode]
```
stephanegigandet, line 344: What is the reason for the removal of the =head4 header for each parameter?

yuktea: @stephanegigandet: I could be wrong but from what I have seen the documentation format is like- 
=`head_N` "the header"
"the content for the heading"

so, 
= `head3` Parameters
`$Parameter1 Parameter2 ` (the parameters)

stephanegigandet: so the idea in the current doc is to have a header3 for the Parameters, and then header4 for each parameter, to separate them.

e.g. you can see how it looks when changed to HTML: (the file is old, I need to regenerate it)

https://world.openfoodfacts.org/files/doc/perl/ProductOpener/Ecoscore.html

In markdown, it would be like this:

### Parameters

#### First parameter

The first parameter is blabla

#### Second parameter

...

yuktea: thank you, fixed it!

stephanegigandet (approved):
Thanks!

PR #6992, opened 2022-06-30
refactor: refactor repeated `delete` calls 

yuktea:
### What
Refactored some repeated` delete `calls in:
`Display.pm`
`Ingredients.pm`
`Food.pm`

There are more of these but a lot of them are already in loops, so a nested loop really wouldn't make anything better.


yuktea (commented):
I believe the jobs are failing because of no double quotes

yuktea:
> I believe the jobs are failing because of no double quotes

The checks pass now!

lib/ProductOpener/Display.pm
```diff
@@ -10329,9 +10329,10 @@ sub display_structured_response($)
 		# remove the languages field which has keys like "en:english"
 		# keys with the : character break the XML export
 
-		delete $request_ref->{structured_response}{product}{languages};
-		delete $request_ref->{structured_response}{product}{category_properties};
-		delete $request_ref->{structured_response}{product}{categories_properties};
+		my @foo = ("languages", "category_properties", "categories_properties");
```
stephanegigandet, line 10332: Instead create a function that we can call like this:

delete_fields($product_ref, ["languages", "category_properties", "categories_properties"]);

yuktea: Done! Looks much righter now.

yuktea: Thanks for the feedback! just made the changes

Issue #4461, opened 2020-11-02
Recognize ingredient + origin (e.g. "lait origine France") in origins field, use it to attribute origin to ingredient


stephanegigandet:
The origins field can contain origins for specific ingredients: "Cocoa from Peru", "Lait origine France" etc.

We need to recognize the ingredient + origin combination, store it in a specific way, display it correctly, and also take it into account during ingredient parsing so that we can attribute the proper origin to the ingredient.

aleene:
This could also be used for a free format origins field.

manon-corneille:
Hi everyone, 
A producer having trouble with this issue got back to me :-)  Any idea when this is going to be fix? 

Thanks

PR #6995, opened 2022-07-01
feat: parse origins of ingredients field

stephanegigandet:
Fixes partially #4461 (we could add support for more ways to describe ingredients, like "Strawberries from Spain").

Needed in particular for one producer who sent us data in the origins of ingredients field, in order to have it used for the Eco-Score computations.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -1016,12 +1016,26 @@ sub add_specific_ingredients_from_labels($) {
 }
 
 
-=head2 parse_specific_ingredients_from_text ( product_ref, $text )
+=head2 parse_specific_ingredients_from_text ( product_ref, $text, $percent_regexp )
 
 Lists of ingredients sometime include extra mentions for specific ingredients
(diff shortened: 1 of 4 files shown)


alexgarel (changes requested):
If it's not too much I would like a refactor to avoid duplicating.

alexgarel:
Also @stephanegigandet  what about using more verbose regexp in our code: https://perldoc.perl.org/perlre#/x-and-/xx

This is very handy to comment what you are doing in an expression.

alexgarel (approved):
Let's merge and refactor later on as proposed by @stephanegigandet

stephanegigandet:
> Also @stephanegigandet what about using more verbose regexp in our code: https://perldoc.perl.org/perlre#/x-and-/xx
> 
> This is very handy to comment what you are doing in an expression.

Yes, we should do that, definitely.

stephanegigandet:
> 

Thanks, opened an issue for not forgetting it: https://github.com/openfoodfacts/openfoodfacts-server/issues/7028

And start of refactor is here: https://github.com/openfoodfacts/openfoodfacts-server/pull/7026

PR #6997, opened 2022-07-01
fix: uninitialized value

dipietroR:
When clicking on the Advanced Search button, two warnings would appear:
```
[Fri Jul  1 16:34:41 2022] -e: Use of uninitialized value $title in substitution (s///) at /opt/product-opener/lib/ProductOpener/Display.pm line 6825.
[Fri Jul  1 16:34:41 2022] -e: Use of uninitialized value $title in substitution (s///) at /opt/product-opener/lib/ProductOpener/Display.pm line 6827.
```
This was because $title was empty, so I put the lines that were causing the error:
```
$title =~ s/<SITE>/$site/g;
$title =~ s/<([^>]*)>//g;
```
In the conditional below them:
```
if (defined $title) {`

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -6822,10 +6822,6 @@ sub display_page($) {
 
 	${$content_ref} =~ s/<SITE>/$site/g;
 
-	$title =~ s/<SITE>/$site/g;
-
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Looks good, thank you!

PR #6998, opened 2022-07-01
refactor: remove kcal_to_unit

dipietroR:
This function gets defined and exported, but it doesn't get called nor does it call another function.


dipietroR:
This function doesn't get used, but it seems pretty useful :)

stephanegigandet (approved):
We can remove it if it's not used.

Issue #6927, opened 2022-06-20
Refactoring HTML/CSS and JS out of perl files


yuktea:
### The files to be refactored should be added here

`export_products.pl` - [ ] 

`ProductOpener::Display::display_product_api()` - [ ]

Raghavisdev:
Hi! I’d like to work on this part of the refactoring task.

For `cgi/import_photos_upload.pl`, I noticed that the `template-upload` and `template-download` HTML templates are currently embedded inside the Perl file using a `<<JS` block.

My plan is to:

1. Move these two templates into `templates/web/pages/import_photos_upload/import_photos_upload.tt.html`.
2. Keep the JavaScript library `<script src=...>` imports in the Perl file unchanged.
3. Ensure the templates still work with the existing `import_photos_upload.tt.js` logic.

This should separate the UI templates from the Perl logic and follow the current template structure used in the project.

Please let me know if this approach looks correct before I open a PR.

PR #7003, opened 2022-07-03
refactor: templatize display_search_image_form from 'Images.pm`

yuktea:
### What
Templatizing the `display_search_image_form` method from `Images.pm`

**Related issues and Discussion**
Issue https://github.com/openfoodfacts/openfoodfacts-server/issues/6927

### DIFF
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -440,7 +440,7 @@ JS
 ;
 
 	process_template('web/common/includes/display_search_image_form.tt.html', 
-	$template_data_ref_tags_translate, \$html) || return "template error: " . $tt->error();
+	$template_data_ref, \$html) || return "template error: " . $tt->error();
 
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Images.pm
```diff
@@ -359,31 +358,12 @@ sub display_search_image_form($) {
...
-  $Lang{send_image_error}{$lang}
-  <a href="#" class="close">&times;</a>
-</div>
-
-HTML
-;
+	my $template_data_ref = {
+		product_image_with_barcode => $product_image_with_barcode,
+		id => $id,
+		lang => \&lang,
```
stephanegigandet, line 364: lang and display_icon are now added automatically by process_template(), so we don't need them here

yuktea: done!

stephanegigandet (approved):

PR #7004, opened 2022-07-03
refactor: refactor repeated code and add documentation

yuktea:
### What
- Refactored `default_unit_for_nid()`  (had a lot of `return` statements with conditionals)
- Refactored `compute_serving_size_data()`  (had a lot of `delete` calls)
- documented `default_unit_for_nid()` and `canonicalize_nutriment()` in the POD format

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -250,7 +250,7 @@ sub default_unit_for_nid($) {
 
 	my %default_unit_for_nid_map = (
     "energy-kj" => "kJ", "energy-kcal" => "kcal", "energy" => "kJ", "alcohol" => "% vol", 
-    'water-hardness' => "mmol/l"
+    "water-hardness" => "mmol/l"
 	);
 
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Food.pm
```diff
@@ -1970,11 +1989,12 @@ sub compute_carbon_footprint_infocard($) {
 
 	if (not ((has_tag($product_ref, "countries", "en:france")) and (defined $product_ref->{ingredients_text})
 		and (length($product_ref->{ingredients_text}) > 5))) {
-		delete $product_ref->{environment_impact_level};
-		delete $product_ref->{environment_impact_level_tags};
-		delete $product_ref->{environment_infocard};
-		delete $product_ref->{environment_infocard_en};
-		delete $product_ref->{environment_infocard_fr};
+		
+		state @product_fields_to_delete = ("environment_impact_level", "environment_impact_level_tags",
```
stephanegigandet, line 1993: I don't think we need @product_fields_to_delete to be a persistent variable, why not use my instead of state?

PR #7005, opened 2022-07-04
refactor: add a single method to display Ecoscore calculation details for both standard and simple html

yuktea:
#### Rationale for this proposal:
We have two subs to display the html for detailed eco-score calculation information
`display_ecoscore_calculation_details()
`
`display_ecoscore_calculation_details_simple_html()
` for mobile devices

Instead of these, we could have just one where we use an extra Boolean parameter `$simple_html_mobile` 
- if `$simple_html_mobile` is set to 1 (true) then we render the simple Eco score details page


lib/ProductOpener/Display.pm
```diff
@@ -4730,7 +4730,7 @@ sub customize_response_for_product($$) {
 		# The environment infocard now displays the Eco-Score details
 		elsif (($field =~ /^environment_infocard/) or ($field eq "ecoscore_details_simple_html")) {
 				if ((1 or $show_ecoscore) and (defined $product_ref->{ecoscore_data})) {
-						$customized_product_ref->{$field} = display_ecoscore_calculation_details_simple_html($cc, $product_ref->{ecoscore_data});
+						$customized_product_ref->{$field} = display_ecoscore_calculation_details_simple_html($cc, $product_ref->{ecoscore_data}, 1);
```
stephanegigandet, line 4733: I think it needs to be changed:

```suggestion
						$customized_product_ref->{$field} = display_ecoscore_calculation_details($cc, $product_ref->{ecoscore_data}, 1);
```

stephanegigandet:
Hi @yuktea , could you test it and verify that you get the correct output when displaying the simple / normal Eco-Score details?

yuktea:
PR closed because these changes will be obsolete owing to the redesign for OFF https://github.com/openfoodfacts/openfoodfacts-server/pull/7414

PR #7007, opened 2022-07-04
refactor: add helper function to identify if a product bears a certain tag from a passed list

yuktea:
### Proposal
Having helper functions where ever we have multiple calls to a function (let's say to `has_tag` like in this case).
I think it makes the if statements look much cleaner and keep the code organized. Also, this offers somewhat better reusability because we can simply add another parameter to the list (in this case, of brands)

_Not sure if I've declared it at the appropriate place. Should it be in `tags.pm`?_

### DIFF
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -53,6 +53,7 @@ BEGIN
 		&canonicalize_tag_link
 
 		&has_tag
+		&has_baby_brand_tag
 		&add_tag
 		&remove_tag
 		&is_a
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -319,15 +319,15 @@ sub has_tag($$$) {
 	return $return;
 }
 
-# Helper function to tell if a product is a "baby product"
-sub has_baby_brand_tag($) {
-
-	my $product_ref = shift;
+# Helper function to tell if a product has a certain tag from the passed list
+sub has_one_of_the_tags_from_the_list {
+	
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Tags.pm
```diff
@@ -318,6 +319,22 @@ sub has_tag($$$) {
 	return $return;
 }
 
+# Helper function to tell if a product is a "baby product"
+sub has_baby_brand_tag($) {
```
stephanegigandet, line 323: Maybe create a more generic function  has_one_of_tags($product_ref, $tagtype, $reference_to_a_list_of_tags)

lib/ProductOpener/Tags.pm
```diff
@@ -318,6 +319,22 @@ sub has_tag($$$) {
...
+sub has_one_of_the_tags_from_the_list {
+	
+	my($product_ref, $tagtype, $tag_list_ref) = @_;
+	
+	my $flag = 0;
+
+	foreach my $tag_name (@$tag_list_ref) {
+			if ( has_tag($product_ref, $tagtype, $tag_name) ) {
+				$flag = 1;
+				return $flag;
```
stephanegigandet, line 332: It looks good, but if you have a return statement here, we don't need a $flag variable, you can just put "return 1;" here, and "return 0;" at the end of the function.

Or if we want to have only 1 return statement at the end, you could replace "return $flag;" by "last;" to exit from the loop.

stephanegigandet (approved):
Thank you!

stephanegigandet (approved):

stephanegigandet:
Perfect, thanks!

Issue #6956, opened 2022-06-24
Decide on how to use (or not to use) Perl function prototypes or signatures


stephanegigandet:
### Is your feature request related to a problem? Please describe.

In most of the Product Opener code, we use Perl prototypes to specify how many arguments functions have.

e.g.

```
sub product_url($) {

	my $code_or_ref = shift;

```

This is something I believe is useful, because if the function is called with a different number of arguments, it will fail at compile time.

e.g. if I add this function in Display.pm

```
sub print_my_product_url() {

	print(product_url(1234, 12));
}
```

Then when I do "make restart", the container stops and I have this error in the logs:

```
[Fri Jun 24 09:48:12.380196 2022] [perl:error] [pid 16] Too many arguments for ProductOpener::Products::product_url at /opt/product-opener/lib/ProductOpener/Products.pm line 2436, near "12)"\nCompilation failed in require at /opt/product-opener/lib/ProductOpener/Users.pm line 91.\nBEGIN failed--compilation aborted at /opt/product-opener/lib/ProductOpener/Users.pm line 91.\nCompilation failed in require at /opt/product-opener/lib/ProductOpener/Display.pm line 147.\nBEGIN failed--compilation aborted at /opt/product-opener/lib/ProductOpener/Display.pm line 147.\nCompilation failed in require at /opt/product-opener/lib/startup_apache2.pl line 66.\nBEGIN failed--compilation aborted at /opt/product-opener/lib/startup_apache2.pl line 66.\nCompilation failed in require at (eval 2) line 1.\n
[Fri Jun 24 09:48:12.380239 2022] [perl:error] [pid 16] Can't load Perl file: /opt/product-opener/lib/startup_apache2.pl for server openfoodfacts.localhost:0, exiting...
```

But there are lots of threads on the Web that explain that Perl's prototypes are harmful / should not be used. e.g. https://stackoverflow.com/questions/297034/why-are-perl-5s-function-prototypes-bad

### Describe the solution you'd like

I think we could just use the "experimental" signatures of Perl (that are in fact not experimental anymore in the just released Perl 5.36):

https://perldoc.pl/perlsub#Signatures

Any thoughts?

cc @hangy @alexgarel @zigouras

smonff:
Hi Stéphane,

Answering here after reading the message on FR Perl Mongers mailing list.

```
use feature qw(signatures);
no warnings qw(experimental::signatures);
```

[...]

PR #7009, opened 2022-07-04
refactor: enable Perl signatures 

stephanegigandet:
Based on the discussion in #6956, this is a new ProductOpener::Module to put boilerplate code that we want in every module. 

It replaces Modern::Perl so that we can enable signatures.

Also switched one module, Nutriscore.pm, to signatures.


dipietroR (approved):
So we would replace the prototype:
```
sub something($) {
  my $var1 = shift;
```

with a signature:
```
sub something($var) {
```
?
I think that would be great!

Ban3:
Functionally, it's fine, but I'd go for a more descriptive name. How about `PerlStandards` suggested by @hangy? It would be clear from the name what it does.

hangy (approved):

alexgarel (approved):
Really cool.

stephanegigandet:
> So we would replace the prototype:
> 
> ```
> sub something($) {
>   my $var1 = shift;
> ```
> 
> with a signature:
> 
> ```
> sub something($var) {
> ```
> 
> ?

Yes, exactly.

PR #7018, opened 2022-07-05
refactor: headers_ref unused variable

dipietroR:
This variable appears to be defined, but not used.


stephanegigandet (approved):
Thanks!

PR #7019, opened 2022-07-06
fix: add missing subdomain to redirect url #6993

### LINKED ISSUE #6993: 301 response makes it difficult to implement a feature
@teolemon @stephanegigandet Actually there's a little bug I've just discovered, on the server side.

With the latest version of off-dart, we also put the language in the query. Which is a good idea, isn't it?

It works well for the first 3 queries (contributor, informer, photographer).
But for the last one ("to be completed") the server translates the URL, and that fails.
* When I add no language to https://world.openfoodfacts.org/informer/monsieurtanuki/state/to-be-completed.json, everything is fine
* When I put the language
  * https://world.openfoodfacts.org/informer/monsieurtanuki/state/to-be-completed.json?lc=fr
  * get translated into
  * https://world.openfoodfacts.org/informateur/monsieurtanuki/etat/a-completer.json
  * and instead of json results I get this response:
```html
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="/informateur/monsieurtanuki/etat/a-completer.json">here</a>.</p>
<hr>
<address>Apache/2.4.25 (Debian) Server at world.openfoodfacts.org Port 80</address>
</body></html>
```

What shall we do? Should we handle that in off-dart? Is that a bad redirection code on the server side?

_Originally posted by @monsieurtanuki in https://github.com/openfoodfacts/smooth-app/issues/2315#issuecomment-1170918565_

stephanegigandet:
This fixes an issue reported by @monsieurtanuki

When accessing URLs like https://world.openfoodfacts.org/informer/monsieurtanuki/state/to-be-completed.json?lc=fr  --> the tags and tagtypes are translated to French, and we are redirected to:

https://world.openfoodfacts.org/informateur/monsieurtanuki/etat/a-completer.json

This works only is the subdomain is also changed.

Fixes #6993 

Also helps https://github.com/openfoodfacts/smooth-app/issues/2315 but those URLs are very tied to the website URLs, they are mostly for debugging purposes, apps should use the search API instead to get products matching facets.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -2989,7 +2989,7 @@ sub display_points($) {
 	$request_ref->{current_link} .= "/points";
 
 	if ((defined $tagid) and ($newtagid ne $tagid) ) {
-		$request_ref->{redirect} = $request_ref->{current_link};
+		$request_ref->{redirect} = $formatted_subdomain . $request_ref->{current_link};
 		$log->info("newtagid does not equal the original tagid, redirecting", { newtagid => $newtagid, redirect => $request_ref->{redirect} }) if $log->is_info();
 		return 301;
 	}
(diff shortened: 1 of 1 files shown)


alexgarel (approved):
Good ! I tested it ok on my host.

Issue #6462, opened 2022-03-07
CORS add hunger.openfoodfacts.org to authentication url


alexfauquette:
### Is your feature request related to a problem? Please describe.

- In Hunger Games, requests to verify if a user is logged in or not fail due to a CORS issue

- Tod do so I tried `axios.get("https://world.openfoodfacts.org/cgi/auth.pl");`
  
- More information in the following PR I merge to be sure it was falling under hunger.openfoodfacts.org domain name
https://github.com/openfoodfacts/openfoodfacts-hungergames/pull/290

- In brief, I need to set `withCredentials=true` to let send the cookies, to `world.openfoodfacts.org` but I can not read the answer if the response header does not contain `Access-Control-Allow-Credentials=true`

### Describe the solution you'd like

- add `Access-Control-Allow-Credentials=true` when the request comes from `hunger.openfoodfacts.org`

stephanegigandet:
@alexfauquette I deployed https://github.com/openfoodfacts/openfoodfacts-server/pull/7022 in production, can you check if it works for you?

alexfauquette:
@stephanegigandet I added a console log to verify in production what JS have access to: The following script is run when loading the page https://hunger.openfoodfacts.org/logos/search

```js
axios.get("https://world.openfoodfacts.org/cgi/auth.pl", { withCredentials: true })
      .then((rep) => console.log({ rep }))
      .catch((err) => console.log({ err }));
```

The line you added allows sending the cookies. What is missing is the `Access-Control-Allow-Origin`

since for know it seems I'm the only JS app using it you can set `Access-Control-Allow-Origin: https://hunger.openfoodfacts.org`

I tried with the extension [Moesif Origin & CORS Changer](https://chrome.google.com/webstore/detail/moesif-origin-cors-change/digfbfaphojjndkpccljibejjbppifbc) which allows to modify the response header.

With `Access-Control-Allow-Origin: https://hunger.openfoodfacts.org` or `Access-Control-Allow-Origin: *` I can see in the axios response if I'm logged in (code 200) or not (code 403) compared to now where I always get a message "network error"

stephanegigandet:
> since for know it seems I'm the only JS app using it you can set Access-Control-Allow-Origin: https://hunger.openfoodfacts.org

mmmh. I'm not sure about that. Folksonomy Engine uses /cgi/auth.pl as well. cc @CharlesNepote  @cquest 

[...]

PR #7022, opened 2022-07-06
fix: set access-control-allow-credentials=true for auth.pl

stephanegigandet:
Fix for https://github.com/openfoodfacts/openfoodfacts-server/issues/6462

### DIFF
--- a/cgi/auth.pl
+++ b/cgi/auth.pl
@@ -57,9 +57,9 @@ print header(-status => $status);
 my $r = Apache2::RequestUtil->request();
 my $origin = $r->headers_in->{Origin} || '';
 
-# Only allow requests from hunger.openfoodfacts.org to see if a user is logged in or not
+# Only allow requests from one of our subdomains to see if a user is logged in or not
 
-if ($origin =~ /^https:\/\/hunger\.openfoodfacts\.org\//) {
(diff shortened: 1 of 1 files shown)


cgi/auth.pl
```diff
@@ -47,7 +47,14 @@
...
 }
 
-print header ( -status => $status );
+print header(-status => $status);
+
+# We need to send the header Access-Control-Allow-Credentials=true so that websites
+# such has hunger.openfoodfacts.org that send a query to world.openfoodfacts.org/cgi/auth.pl
+# can read the resulting response.
+
+print header(-Access_Control_Allow_Credentials => "true");
```
alexgarel, line 56: As we are sending the header from here, couldn't we first check for referer and limit this to:

`defined env["HTTP_REFERER"] and env["HTTP_REFERER"] =~ /^(?:https?://)\w+$main_domain/`

assuming $main_domain contains openfoodfacts.org/net/localhost

alexgarel (approved):
LGTM - proposed an improvement, but not mandatory

cgi/auth.pl
stephanegigandet: Good idea, I added a change like this. The browser will send an Origin header.

alexgarel:
ok, I imagine you will test before merging ?

alexgarel:
@stephanegigandet is this ready to merge ?

PR #7024, opened 2022-07-06
refactor: updating perlcritic

dipietroR:
I changed the perlcritic config file to make it easier to pick and choose what tests will run :)


.github/workflows/perlcritic.yml
```diff
@@ -12,10 +12,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-
-    - name: "Run Perl::Critic" 
-      uses: Difegue/action-perlcritic@2.1.0
```
stephanegigandet, line 17: What is the motivation for not using the GitHub action https://github.com/Difegue/action-perlcritic ?

.github/workflows/perlcritic.yml
```diff
@@ -12,10 +12,23 @@ jobs:
...
+    - name: "Run Perl::Critic"
+      run: |
+        for i in $(find . -type f \( -name "*.t" -o -name "*.pl" -o -name "*.pm" \))
+        do
+          if result=$(perlcritic ${i})
+          then    
+                  echo $result
+          else    
+                  echo -e "Problem found with ${i}:\n${result}"
+                  exit 1
```
stephanegigandet, line 32: The test will exit for the first file that has problems, so if multiple files have problems, we won't see it right away. (e.g. right now the tests fail for food.t, but I don't know if other files will fail once food.t is fixed)

.github/workflows/perlcritic.yml
dipietroR: It didn't seem to be working, it would pass my code when it has violations in it.

.github/workflows/perlcritic.yml
dipietroR: This is true, if I can't get the GitHub action working, I'll update this.

.github/workflows/perlcritic.yml
```diff
@@ -12,10 +12,23 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-
-    - name: "Run Perl::Critic"
```
alexgarel, line 16: There again could we go for a Make implementation using docker ? It will enable running the check locally to reproduce errors.

dipietroR: Adding it to Make seems like a good idea! Especially since I can't seem to get GitHub action to work.

stephanegigandet (approved):
Looks good to me, thank you!

dipietroR:
The perlcritic test would fail here, but it seemed to work when ran locally, I'm not quite sure why. I set a default file for perlcritic to look at in case there were no files for it to check in a PR (like if the PR didn't edit any perl files). This should keep it from failing

alexgarel:
@dipietroR It did not work, maybe we should add a `git fetch origin main` before running perl tidy and perl critic.

If you want to test locally, you can try to go in a fresh directory and run:

```bash
git init .
git remote add origin https://github.com/openfoodfacts/openfoodfacts-server
git fetch --no-tags --prune --progress --no-recurse-submodules --depth=50 origin perlcritic
git checkout perlcritic
```

It seems that adding:
```
git fetch origin/main
```
makes the `git d/iff origin main` to pass.

dipietroR:
Alright, I think this should work now.

alexgarel:
Great @dipietroR, but I added a commit because I would prefer the fetch to bit in the github action (normally locally, it should be already there).

alexgarel:
I rebased the branch to be sure we do not merge in previous commits

PR #7025, opened 2022-07-06
refactor: replace perl "shift" prototype in `Display.pm`

dipietroR:
Since https://github.com/openfoodfacts/openfoodfacts-server/pull/7009/files was merged, I figured we should start replacing the Perl prototypes with Perl signatures. I replaced a few of the "shift" prototypes in Display.pm

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -200,8 +200,7 @@ use Apache2::Const ();
 
 use URI::Find;
 
-my $uri_finder = URI::Find->new(sub {
-      my($uri, $orig_uri) = @_;
+my $uri_finder = URI::Find->new(sub($uri, $orig_uri) {
 	  if ($uri =~ /\http/) {
 		return qq|<a href="$uri">$orig_uri</a>|;
 	  }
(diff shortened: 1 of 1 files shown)


Ban3:
Change
https://github.com/openfoodfacts/openfoodfacts-server/blob/e41e2c385bdc1b3acb994b16ce3c30a6019956db/lib/ProductOpener/Display.pm#L38-L39
to

```perl
use ProductOpener::PerlStandards
```
when refactoring modules to use signatures.

dipietroR:
> Change
> 
> https://github.com/openfoodfacts/openfoodfacts-server/blob/e41e2c385bdc1b3acb994b16ce3c30a6019956db/lib/ProductOpener/Display.pm#L38-L39
> 
> to
> 
> ```perl
> use ProductOpener::PerlStandards
> ```
> 
> when refactoring modules to use signatures.

Thank you!

Ban3:
> Though, there were a few that had shifts or `@_`'s but no $'s so I left them alone; ie:

Maybe we should convert those regular subroutines to use signatures as well for uniformity.

Ban3:
> Illegal character following sigil in a subroutine signature

These seem to come from nested prototypes in that region, perl thinks they have signatures. [perlsub](https://perldoc.perl.org/perlsub#Signatures) might aid further debugging.

dipietroR:
> > sub display_structured_response_opensearch_rss {
> > my ($request_ref) = @_;
> 
> Subroutines like this are fine to convert because we are assigning `_` to `$request_ref`, but be careful if `_` is on the left side. 

Alright, thanks!

alexgarel (approved):
LGTM

stephanegigandet (approved):
Thank you!

PR #7026, opened 2022-07-06
refactor: parse origins of ingredients field

stephanegigandet:
Partial refactor based on suggestions from @alexgarel in https://github.com/openfoodfacts/openfoodfacts-server/pull/6995


alexgarel:
@stephanegigandet I renamed you PR. There is a "refactor" prefix. For refactor are not features.

PR #7027, opened 2022-07-06
refactor: replace perl "shift" prototype in `Recipes.pm`

dipietroR:
Replacing the "shift" perl prototypes to perl signatures in Recipes.pm


stephanegigandet (approved):
Thank you!

PR #7030, opened 2022-07-07
refactor: enable perl function signatures

yuktea:
### What 
PR to enable functional signatures
### Covers
- [DataQuality.pm](https://github.com/openfoodfacts/openfoodfacts-server/commit/6962f00ae72601129769bef1492c869e1d7789ec)
- [DataQualityCommon](https://github.com/openfoodfacts/openfoodfacts-server/commit/04c95ca5b595b0ce502470b36c9d6606a3ea8146)
- [DataQualityFood.pm](https://github.com/openfoodfacts/openfoodfacts-server/commit/0e3b4ff0d1b78f29314622670bc67977500a8100)
-  [Web.pm](#)
- [FoodGroups.pm](#)
- [Mail.pm](#)
- [Numbers.pm](#)


lib/ProductOpener/DataQuality.pm
```diff
@@ -70,6 +69,7 @@ The type of product is specified through Config.pm
 
 package ProductOpener::DataQuality;
 
+use ProductOpener::PerlStandards;
 use utf8;
```
stephanegigandet, line 73: We can remove:

use utf8;
use Modern::Perl '2017';

As their effect is included in ProductOpener::PerlStandards;

stephanegigandet (approved):
Looks good, the only thing is that we should also remove:

use utf8;
use Modern::Perl '2017';

As their effect is now covered by:

ProductOpener::PerlStandards;

PR #7035, opened 2022-07-07
taxonomy: Fixed paprika issue

moon-rabbitOFF:


stephanegigandet:
Can you take a look @aleene ? Should we add a "paprika" entry in Dutch? to "bell peppers"?

moon-rabbitOFF:
We can indeed. It was in plural form only "nl:Paprikas", I added the singular.

aleene:
I probably removed the singular paprika, as it is ambiguous. It can also men the spice.

moon-rabbitOFF:
Ah ok, so I remove it again

stephanegigandet (approved):

PR #7036, opened 2022-07-07
refactor: replace perl shift prototype in Test.pm

dipietroR:
Switching "shift" prototypes to perl signature

### DIFF
--- a/lib/ProductOpener/Test.pm
+++ b/lib/ProductOpener/Test.pm
@@ -28,8 +28,7 @@ ProductOpener::Test - utility functions used by unit and integration tests
 
 package ProductOpener::Test;
 
-use utf8;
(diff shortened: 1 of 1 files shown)


dipietroR:
This line seems to be what causes the tests to fail:
```
sub _sub_items($item_ref, $subfields_ref);  # declare for recursivity
```
Before it was:
```
sub _sub_items($$);  # declare for recursivity
```
I'm not sure how to fix this and why it was declared here.

Ban3:
That sort of forward declaration was needed (and possible) when it was a prototype. That can be safely removed.

dipietroR:
> That sort of forward declaration was needed (and possible) when it was a prototype. That can be safely removed.

Thanks for explaining!

PR #7038, opened 2022-07-08
fix: better wording for adding a precise category

stephanegigandet:
Better wording to indicate that we need a precise category for the Eco-Score.

Fixes https://github.com/openfoodfacts/smooth-app/issues/2542


alexgarel (approved):
LGTM

M123-dev:
Just curious what is exactly a precise category, what would it be here  https://de-de.openfoodfacts.org/product/8027547000012

PR #7040, opened 2022-07-08

dipietroR:
Adding perltidy formatting and checking to `cgi/change_password.pl`, `cgi/generate_sample_import_file.pl`, and `cgi/i18n/countries.pl`


alexgarel (approved):
LGTM. Thanks

PR #7042, opened 2022-07-08

dipietroR:
Implementing perltidy to `cgi/import_file_job_status.pl`, `cgi/import_file_process.pl`, and `cgi/import_file_select_format.pl`


alexgarel:
@dipietroR, as for #7045, because of conflicts, it becomes harder to resolve those for a low added value (as told we could tidy all files at the end of summer).

PR #7045, opened 2022-07-08

dipietroR:
Implementing perltidy to `cgi/import_file_upload.pl`, `cgi/import_photos_upload.pl`, and `cgi/import_products_categories_from_public_database.pl`


alexgarel:
@dipietroR because of conflicts, it becomes harder to resolve those for a low added value (as told we could tidy all files at the end of summer).

Do you agree that we close this as well as  #7051 ?

PR #7055, opened 2022-07-11
refactor: adding perltidy to cgi/css.pl

dipietroR:
Implementing perltidy to `cgi/css.pl`, which will cause a merge conflict with PR #6663 

Please approve this PR, then I will merge it and fix the conflicts


dipietroR:
I looked more on PR #6663 and I think that it may be solved by the other perlcritic PR (#7024 ), so I think that this is safe to merge as is.

stephanegigandet (approved):

PR #7066, opened 2022-07-13

dipietroR:
Implementing perltidy to `cgi/org.pl`, `cgi/packaging.pl`, and `cgi/product_image.pl`

### DIFF
--- a/cgi/org.pl
+++ b/cgi/org.pl
@@ -35,7 +35,6 @@ use ProductOpener::Orgs qw/:all/;
 use ProductOpener::Tags qw/:all/;
 use ProductOpener::Text qw/:all/;
 
-
 use CGI qw/:cgi :form escapeHTML charset/;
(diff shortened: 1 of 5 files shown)


cgi/org.pl
```diff
@@ -218,29 +227,34 @@
...
+				},
+				{
+					field => "gs1_nutrients_are_unprepared",
+					type => "checkbox",
+				},
+			)
+		);
+
+		push @{$template_data_ref->{sections}},
+		  {
```
stephanegigandet, line 248: I wonder why there are 2 spaces before the { (and not a tab)

stephanegigandet (approved):
Looks good to me

cgi/org.pl
dipietroR: I added another rule to try to fix the tabs, how does this look?

PR #7069, opened 2022-07-14

dipietroR:
Implementing perltidy to `t/additives.t`, `t/additives_tags.t`, and `t/allergens_tags.t`

### DIFF
--- a/.perltidy_excludes
+++ b/.perltidy_excludes
@@ -204,10 +204,7 @@ scripts/update_recent_changes.pl
 scripts/update_top_translators.pl
 tests/integration/import_csv_file.t
 tests/integration/export.t
-tests/unit/additives.t
-tests/unit/additives_tags.t
 tests/unit/allergens.t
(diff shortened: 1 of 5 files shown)


t/allergens_tags.t
```diff
@@ -5,6 +5,7 @@ use utf8;
 
 use Test::More;
 use Log::Any::Adapter 'TAP';
+
```
Ban3, line 8: Inserting blanks before comments like this doesn't make much sense to me. Maybe we should consider changing the comment to static block comment (`##` by default) to keep it in place, or use `--noblanks-before-comments` to completely disable the feature.

dipietroR: I added `--noblanks-before-comments` which seemed to fix that, thanks!

alexgarel:
I have rebased to resolve conflict.

@dipietroR if you want to work on this branch (but you should not need to), be aware that you should first remove you local branch and check it out again.

PR #7071, opened 2022-07-14
build: configure Ory Kratos

LandonPattison:
### What
- Configure kratos by env variables
- Added redirects from OFF to kratos
- Fix users not being signed in after logging in/creating an account
- Added kratos_auth.pl to create new users after being created in kratos
- Added kratos_update_settings.pl to update a users settings after being updated in kratos
- Added kratos_import_users_file.pl and kratos_import_users_post.pl to import users into kratos
- Wrote documentation on how to develop with ory


Ory/kratos/kratos.yml
```diff
@@ -80,6 +81,13 @@ secrets:
   cipher:
     - 32-LONG-SECRET-NOT-SECURE-AT-ALL
 
+session:
+  cookie:
+    domain: http://openfoodfacts.localhost/
```
alexgarel, line 86: See https://www.ory.sh/docs/kratos/guides/configuring-cookies#session-cookies you should not put the `http://` neither a final `/`:
```suggestion
    domain: openfoodfacts.localhost
```

Ory/kratos/kratos.yml
```diff
@@ -80,6 +81,13 @@ secrets:
...
+
+cookies:
+  domain: http://openfoodfacts.localhost/
```
alexgarel, line 89:
```suggestion
  domain: openfoodfacts.localhost
```

Ory/kratosauth.yml
```diff
@@ -27,8 +27,8 @@ services:
     environment:
       - DSN=postgres://kratos:secret@postgres-kratos:5432/kratos?sslmode=disable&max_conns=20&max_idle_conns=4
     ports:
-      - '127.0.0.1:4433:4433'
-      - '127.0.0.1:4434:4434'
+      - '4433:4433'
+      - '4434:4434'
```
alexgarel, line 31: Please keep as it was here, it's not linked and it's better for security to keep exposed only on localhost.

Ory/kratosauth.yml
```diff
@@ -38,10 +38,10 @@ services:
   kratos-selfservice-ui-node:
     image: oryd/kratos-selfservice-ui-node:latest
     environment:
-      - KRATOS_PUBLIC_URL=http://kratos:4433/
-      - KRATOS_BROWSER_URL=http://127.0.0.1:4433/
+      - KRATOS_PUBLIC_URL=http://kratos.openfoodfacts.localhost:4433/
+      - KRATOS_BROWSER_URL=http://kratos.openfoodfacts.localhost:4433/
     ports:
-      - '127.0.0.1:4455:3000'
+      - '4455:3000'
```
alexgarel, line 44: Please keep as it was here (this line), it's not linked and it's better for security to keep exposed only on localhost.

cgi/kratos_auth.pl
```diff
@@ -0,0 +1,71 @@
...
+use Modern::Perl '2017';
+use utf8;
+
+use ProductOpener::Users qw/:all/;
+
+use LWP::UserAgent;
+use JSON;
+use Log::Any qw($log);
+
+my $myCookie = cookie("ory_kratos_session");
```
alexgarel, line 32: `$kratos_cookie" might be a better name ;-)

cgi/kratos_auth.pl
```diff
@@ -0,0 +1,71 @@
...
+print $myCookie;
+
+if(defined $myCookie){
+    my $url = "http://world.openfoodfacts.localhost/cgi/kratos_auth.pl/session/whoami";
```
alexgarel, line 36:
```suggestion
    my $url = "http://kratos.openfoodfacts.localhost/cgi/kratos_auth.pl/session/whoami";
```

cgi/kratos_auth.pl
```diff
@@ -0,0 +1,71 @@
...
+
+    my $ua = LWP::UserAgent->new;
+
+    # set custom HTTP request header fields, must include cookie for /session/whoami
+    my $req = HTTP::Request->new(GET => $url);
+    $req->header('content-type' => 'application/json');
+    $req->header('Cookie' => $myCookie);
```
alexgarel, line 43: No, you should not set the cookie as cookie (and note that this would be an invalid cookie specification).
The [documentation](https://www.ory.sh/docs/guides/session-management/check-session) tells you to use a header named `X-Session-Cookie`
```suggestion
    $req->header('X-Session-Cookie' => $myCookie);
```

Ory/kratos/kratos.yml
LandonPattison: fixed

Ory/kratos/kratos.yml
LandonPattison: fixed

Ory/kratosauth.yml
LandonPattison: fixed

cgi/kratos_auth.pl
LandonPattison: I agree, fixed

cgi/kratos_auth.pl
LandonPattison: fixed

cgi/kratos_auth.pl
LandonPattison: Fixed, thought it could be either or

LandonPattison: Made back to Cookie [documentation](https://www.ory.sh/docs/kratos/reference/api#operation/toSession), here the docs say the `X-Session_Cookie` is for non browser clients, `Cookie` is for browser clients

alexgarel: Oh ok, fine :-)

cgi/kratos_auth.pl
```diff
@@ -45,27 +47,20 @@
     my $resp = $ua->request($req);
 
     if ($resp->is_success) {
+        #decode json to a hash
         my $json = $resp->decoded_content;
-        log->debug($json);
+        my $content = decode_json($json);
+        my %contentdecoded = %$content;
```
alexgarel, line 53: your names are a bit misleading.

```suggest
        my $content_ref = decode_json($json);
        my %content = %$content_ref;
```
Also you don't really need to do that. If you keep `$content_ref`, you can access data adding a `->`.
Like `$content_ref->{identity}{traits}{UserID}`

LandonPattison: Fixed, did not know you could do this. Thanks!

Ory/kratos/kratos.yml
```diff
@@ -23,47 +19,32 @@ serve:
...
         - Content-Type
         - Set-Cookie
-  admin:
-    base_url: http://kratos.openfoodfacts.localhost:4434/
 
 selfservice:
-  default_browser_return_url: http://world.openfoodfacts.localhost/
-  allowed_return_urls:
-    - http://world.openfoodfacts.localhost/
+  default_browser_return_url: http://world.openfoodfacts.localhost/cgi/kratos_auth.pl
```
alexgarel, line 24: Is this the last "localhost" hardcoded url ? Is it harder to configure through environment ?

cgi/kratos_import_users.pl
```diff
@@ -0,0 +1,107 @@
...
+
+    my $post_resp = $ua->request($post_req);
+
+    if($post_resp->is_success){
+        $log->debug("User Created");
+    }
+    else{
+        $log->debug("HTTP POST error code: ", $post_resp->code);
+        $log->debug("HTTP POST error message: ", $post_resp->message);
+    }
```
alexgarel, line 104: My long "migrations" experience, tells me it's better if we separate json generation and posting.

I would rather dump all json in a file (eg. one json per line)
and have a second script that read this file and post !

This has several advantage like being able to check before launching script, or splitting the json file, or just replaying failing lines, etc.
When posting, Instead of just logging failure, you cloud create a file with failing json (one by line again), for check and replay.

cgi/kratos_import_users.pl
```diff
@@ -0,0 +1,107 @@
...
+
+
+}
```
alexgarel, line 107: And kudos for your migration script in general :-)

cgi/kratos_logout.pl
```diff
@@ -0,0 +1,77 @@
...
+
+        #Unset OFF cookie
+        my $cookie_name   = 'session';
+        my $cookie_domain = "." . $server_domain;
+        my $request_ref = ProductOpener::Display::init_request();
+        my $session = {} ;
+		$request_ref->{cookie} = cookie(-name=>$cookie_name, -expires=>'-1d',-value=>$session, -path=>'/', -domain=>"$cookie_domain") ;
+
+        #Go to the given logout url, this logout url redirects to the logout.after url in kratos.yml
+        print redirect(-url=>$logout_url);
```
alexgarel, line 70: I think you should use: 

```perl
redirect_to_url($request_ref, 302, $logout_url);
```

cgi/kratos_update_settings.pl
```diff
@@ -0,0 +1,131 @@
...
+        #store the updated sto file
+        store($user_ref, $user_file);
+    }
+    else {
+        $log->debug("HTTP GET error code: ", $resp->code, "n");
+        $log->debug("HTTP GET error message: ", $resp->message, "n");
+    }
+}
+
+print redirect(-url=>'http://world.openfoodfacts.localhost/');
```
alexgarel, line 131: I think you should use: 

```perl
redirect_to_url($request_ref, 302, 'http://world.openfoodfacts.localhost/');
```
But of course the url should normally be built (but I imagine it was for testing).

cgi/reset_password.pl
```diff
@@ -61,11 +61,9 @@
 
 my $html = '';
 
-my $ORY_ENABLED = 0;
-
 if($ORY_ENABLED){
 	if(not defined param('flow')){
-		print redirect(-url=>'http://127.0.0.1:4433//self-service/recovery/browser');
+		print redirect(-url=>'http://kratos.openfoodfacts.localhost:4455/recovery');
```
alexgarel, line 66: it may be better to use redirect_to_url here also.

Ory/kratos/kratos.yml
LandonPattison: This does not matter as it is just the default, as long as they are configured through env in docker file all is good.

cgi/kratos_import_users.pl
LandonPattison: Thank you :)

cgi/kratos_logout.pl
LandonPattison: fixed

cgi/kratos_import_users.pl
LandonPattison: Created all users file [here](https://github.com/openfoodfacts/openfoodfacts-server/pull/7071/files#diff-b1f6f17300c7b56c6c26323639c01e12f3a6b30fd8b4d24a9d9a5e7a9fd3a9bf) and separated into two files users imported and not imported [here](https://github.com/openfoodfacts/openfoodfacts-server/pull/7071/files#diff-de5838131f059d46c356d13e429de2ba5877a032e59be0dc2bbd160504c2a57d)

PR #7074, opened 2022-07-15

dipietroR:
Implementing perltidy to `t/attributes.t`, `t/food_groups.t`, and `t/forest_footprint.t`


alexgarel:
@dipietroR, as for #7045, because of conflicts, it becomes harder to resolve those for a low added value (as told we could tidy all files at the end of summer).

Issue #6838, opened 2022-05-31
Create one integration test for the API to read the data of one product


stephanegigandet:

alexgarel:
[Test.pm](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Test.pm) is test helpers.

And you have two integrations tests:
1. [t/export.t](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/t/export.t) 
2. [t/import.t](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/t/import.t)

Pitfall:
when you run test, look at what we do in Makefile: we add `DOCKER_COMPOSE_TEST=COMPOSE_PROJECT_NAME=po_test PO_COMMON_PREFIX=test_`  in front of docker-compose command, it makes docker-compose run test in a completely isolated environment. It avoid deleting the sample products from your dev environment.

alexgarel:
You can run a single test, by simply issuing:

`DOCKER_COMPOSE_TEST=COMPOSE_PROJECT_NAME=po_test PO_COMMON_PREFIX=test_ docker-compose run --rm backend perl t/my_test.pl`

stephanegigandet:
import.t is not an integration test, it's import_csv_file.t and import_gs1.t that are more like integration tests

stephanegigandet:
For the API read test, we could do it like this:

the .t file first:
- uses the internal ProductOpener methods to write specific product data for some products
- calls the HTTP API of ProductOpener to retrieve the product data for those products, and stores the resulting JSONs
- compares the resulting JSONs to the expected JSONs
Using JSONs to store the expected results enables us to clearly see what has changed when we do PR reviews.

stephanegigandet:
We now have a generic way to do integration tests for the API: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/tests/integration/api_v2_product_read.t

PR #7076, opened 2022-07-18
test: read integration test Open Food Facts api

yuktea:
- Mentioned in https://github.com/openfoodfacts/openfoodfacts-server/issues/6838


stephanegigandet:
Building on @yuktea's work, we are now doing integration tests for the API in a more generic way, so I'm closing this PR.

See https://github.com/openfoodfacts/openfoodfacts-server/blob/main/tests/integration/api_v2_product_read.t

PR #7082, opened 2022-07-18
fix: return statements to `KnowledgePanels.pm`

dipietroR:
Used Perl::Critic to look for functions without return statements. I'm planning on making more PRs similar to this (for all of the other files) and then adding `[Subroutines::RequireFinalReturn]` to .perlcriticrc

### DIFF
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -172,6 +172,7 @@ sub create_knowledge_panels($$$$) {
     # Create the root panel that contains the panels we want to show directly on the product page
     create_panel_from_json_template("root", "api/knowledge-panels/root.tt.json",
         {}, $product_ref, $target_lc, $target_cc);
+    return;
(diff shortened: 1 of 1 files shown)


stephanegigandet:
> I saw that it was being done in [ded3dec](https://github.com/openfoodfacts/openfoodfacts-server/commit/ded3dec42bd2bd1171f48a1c96e096b295514624) so I assumed that we would like to add it in. Of course, if you don't want it done, I could stop and focus on a different rule that we may want added?

Let's implement this rule as you have already done it for many files, as did @Ban3 . I'll approve all the related PRs and we can merge them.

But for the future rules to be added, let's discuss them first to see which ones we think will bring the most value. Could you for instance start an issue on GitHub to list a few new rules you think would be good candidates, with maybe 1 or 2 examples of the issues they would trigger in our code? Then we can review them and discuss the benefits.

stephanegigandet (approved):
Thank you!

PR #7099, opened 2022-07-20
fix: columns synonyms for Packtic import

stephanegigandet:
Recognize some columns automatically in file provided by Packtic


lib/ProductOpener/Producers.pm
```diff
@@ -538,7 +539,7 @@ my %fields_synonyms = (
 
 en => {
```
alexgarel, line 540: should'nt we migrate that to a taxonomy at some point ? ;-)

stephanegigandet: at some point, yes :)

PR #7103, opened 2022-07-20
fix: explicit return statements in `ForestFootprint.pm`

dipietroR:
Adding explicit return statements to subroutines without them. I'm planning on putting the rest of the return statements all in one PR


alexgarel (approved):
LGTM

PR #7104, opened 2022-07-20
fix: return statements

dipietroR:
Adding return statements to subroutines without them. There should be no more subroutines without return statements at the moment.


stephanegigandet (approved):

alexgarel (approved):
LGTM ! Thank you @dipietroR

PR #7106, opened 2022-07-20
refactor: create a request object to replace global variables 

### LINKED ISSUE #7105: Create a request object and move all page / user specific global variables from Display.pm and Users.pm into it

Display.pm and Users.pm make heavy use of global variables like $cc, $lc, $User, $User_id, $cookies etc.

Create a request object that is used to keep all the data about the request and its response.

_No response_

### Additional context

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
This started simple: moving some code out of User.pm init_user() into a new function open_user_session() so that we could open sessions from other places (e.g. when we identify through Kratos).

But then I thought that I could replace the $cookie global variable by a field in the request object: $request_ref->{cookie}.

Then I found out that we didn't really have a request object,  at least not when we were calling Display.pm init() which calls init_user().

So I'm thinking to change init() by init_request() that would return a request object.

That means we need to change all the scripts in cgi, but I think it will be useful. When we have a request object, then we can gradually move most global variables into it.

Will eventually fix #7105


lib/ProductOpener/Users.pm
```diff
@@ -615,19 +621,108 @@ sub check_edit_owner($user_ref, $errors_ref) {
...
+The cookie is returned in $request_ref
+
+=cut
+
+sub open_user_session($user_ref, $request_ref) {
+
+	my $user_id = $user_ref->{'userid'};
+
+	# Maximum of sessions for a given user
+	my $max_session = 10 ;
```
alexgarel, line 646: Is that for the number of sessions we store in `<user>.sto` ? 

Because at first read, I though it was the maximum number of time we use a cookie for a session (which does not make sense, but it's hard to guess as you read the code).

```suggestion
	# Maximum number sessions to store for a given user in sto file
	my $max_session = 10 ;
```

lib/ProductOpener/Users.pm
```diff
@@ -615,19 +621,108 @@ sub check_edit_owner($user_ref, $errors_ref) {
...
+		# Find the older session and remove it
+		my @session_by_time = sort { $user_session_stored{$a}{'time'} <=>
+						$user_session_stored{$b}{'time'} } (keys %user_session_stored);
+
+		while (($#session_by_time + 1)> $max_session)
+		{
+			my $oldest_session = shift @session_by_time;
+			delete $user_ref->{'user_sessions'}{$oldest_session};
+		}
+	}
```
alexgarel, line 668: I think it would be better to have that kind of stuff in a separate sub, for clarity when reading, because it puts the reader out of the flow, while it's easy to isolate (and would ease testing).

lib/ProductOpener/Users.pm
```diff
@@ -615,19 +621,108 @@ sub check_edit_owner($user_ref, $errors_ref) {
...
+
+	$user_ref->{'user_sessions'}{$user_session} = {};
+
+	# Store the ip and time corresponding to the given session
+	$user_ref->{'user_sessions'}{$user_session}{'ip'} = remote_addr();
+	$user_ref->{'user_sessions'}{$user_session}{'time'} = time();
```
alexgarel, line 674: Easier to read, in my opinion:

```suggestion
	# Store the ip and time corresponding to the given session
	$user_ref->{'user_sessions'}{$user_session} = {
	    ip => remote_addr(),
	    time => time()
	};
```

lib/ProductOpener/Users.pm
```diff
@@ -615,19 +621,108 @@ sub check_edit_owner($user_ref, $errors_ref) {
...
+	my $session_ref = { 'user_id'=>$user_id, 'user_session'=>$user_session };
+
+	# Upgrade hashed password to scrypt, if it is still in crypt format
```
alexgarel, line 677:
```suggestion
	# Migration: take the occasion of having password to upgrade to scrypt, if it is still in crypt format
```

lib/ProductOpener/Users.pm
```diff
@@ -615,19 +621,108 @@ sub check_edit_owner($user_ref, $errors_ref) {
...
+	if ($user_ref->{'encrypted_password'} =~ /^\$1\$(?:.*)/) {
+		$user_ref->{'encrypted_password'} = create_password_hash(encode_utf8(decode utf8=>param('password')) );
+		$log->info("crypt password upgraded to scrypt_hash") if $log->is_info();
+	}
+
+	my $user_file = "$data_root/users/" . get_string_id_for_lang("no_language", $user_id) . ".sto";
```
alexgarel, line 683:
```suggestion
        # Store user data
	my $user_file = "$data_root/users/" . get_string_id_for_lang("no_language", $user_id) . ".sto";
```

lib/ProductOpener/Users.pm
```diff
@@ -615,19 +621,108 @@ sub check_edit_owner($user_ref, $errors_ref) {
...
+	}
+	elsif ((defined param('remember_me')) and (param('remember_me') eq 'on')) {
+		$length = 31536000 * 10;
+	}
 
-	my $cookie_name   = 'session';
-	my $cookie_domain = "." . $server_domain;    # e.g. fr.openfoodfacts.org sets the domain to .openfoodfacts.org
-	if ( defined $server_options{cookie_domain} ) {
-		$cookie_domain = "." . $server_options{cookie_domain};    # e.g. fr.import.openfoodfacts.org sets domain to .openfoodfacts.org
+	my $cookie_ref = {
```
alexgarel, line 697:
```suggestion
	# generate session cookie
	my $cookie_ref = {
```

lib/ProductOpener/Users.pm
```diff
@@ -680,7 +775,7 @@ sub init_user() {
 		# If the user exists
 		if (defined $user_id) {
 
-           my  $user_file = "$data_root/users/" . get_string_id_for_lang("no_language", $user_id) . ".sto";
+           my $user_file = "$data_root/users/" . get_string_id_for_lang("no_language", $user_id) . ".sto";
```
alexgarel, line 778: We use it at several places, I would find it better to have it in a function (get_user_file).

But that's for another refactor ;-)

alexgarel (approved):
Really cool move.

I suggest some changes, if you want.


A general comment:

Generally when there are complex workflow, I really prefer if the higher level function just describe the scenario (more easy if steps are sufficiently isolated).

Like:

```
sub open_user_session(...)
   my $user_session = generate_token(64);
   remove_old_sessions()
   add_user_sessipon()
   migrate_password_hash()
   return generate_session_cookie()
```

Which such code, you can write general test for main function, and then write more "corner cased" tests for the indivdual functions at a very low cost.


I think when we refactor code this is a good occasion to do that kind of thing.

lib/ProductOpener/Users.pm
stephanegigandet: > But that's for another refactor ;-)

Yes :) This refactor broke a lots of things, I need to update all cgi scripts to use my $request_ref = init_request() instead of init(). I'll first repair everything I broke.

stephanegigandet:
> Generally when there are complex workflow, I really prefer if the higher level function just describe the scenario (more easy if steps are sufficiently isolated).
> 
> Like:
> 
> ```
> sub open_user_session(...)
>    my $user_session = generate_token(64);
>    remove_old_sessions()
>    add_user_sessipon()
>    migrate_password_hash()
>    return generate_session_cookie()
> ```
> 
> Which such code, you can write general test for main function, and then write more "corner cased" tests for the indivdual functions at a very low cost.
> 
> I think when we refactor code this is a good occasion to do that kind of thing.

Thanks for all the suggestions, I refactored a bit more along those lines.

stephanegigandet:
I changed all the init() calls by init_request() calls. Everything I tested seems to work well.

lib/ProductOpener/Users.pm
```diff
@@ -616,19 +622,163 @@ sub check_edit_owner($user_ref, $errors_ref) {
...
+    # Store user data
+	my $user_file = "$data_root/users/" . get_string_id_for_lang("no_language", $user_id) . ".sto";
+	store($user_file, $user_ref);
+
+	$log->debug("session initialized and user info stored") if $log->is_debug();
+
+	$request_ref->{cookie} = generate_session_cookie($user_id, $user_session);
+
+	return;
+}
```
alexgarel, line 774: Waow thank you so much to take time to refactor ! :tada:

alexgarel:
That was tuff, but really a good move, thanks @stephanegigandet 

@Landone, you are all set :-)

LandonPattison:
Thank you @stephanegigandet !

PR #7108, opened 2022-07-20
chore: add construct url sub for tests

yuktea:
### What

Manually making requests to urls in code could be error prone.
We can pass a prefix and the target path to construct a url to make an http request

### DIFF
--- a/lib/ProductOpener/APITest.pm
+++ b/lib/ProductOpener/APITest.pm
@@ -154,5 +154,4 @@ sub construct_test_url () {
 	return $url;
 }
 
-
 1;
--- a/lib/ProductOpener/APITest.pm
+++ b/lib/ProductOpener/APITest.pm
@@ -124,4 +124,35 @@ sub create_user ($ua, $args_ref) {
 	return;
 }
 
+=head2 construct_test_url()
+
+Constructs the URL to send the HTTP request to for the API.
+
+=head3 Arguments
+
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/APITest.pm
```diff
@@ -122,4 +122,34 @@ sub create_user ($ua, $args_ref) {
...
+
+For the example cited above this returns: "http://world-fr.openfoodfacts.localhost/product/35242200055"
+
+=cut
+
+sub construct_test_url () {
+	my $target = "/xxx/yyy";
+	my $prefix = "world-en";
+
+	my $url = "http://${prefix}.openfoodfacts.localhost${target}";
```
stephanegigandet, line 150: Maybe we could replace openfoodfacts.localhost by $server_domain

lib/ProductOpener/APITest.pm
```diff
@@ -122,4 +122,34 @@ sub create_user ($ua, $args_ref) {
 	$response->is_success or die("Couldn't create user with " . dump(\%fields) . "\n");
 }
 
+=head2 construct_test_url()
```
alexgarel, line 125: change current tests to use it, to verify it works :-)

stephanegigandet (approved):

alexgarel:
hum @yuktea I have made "update with rebase" so if you want to continue on this branch, you should delete and re-checkout you local branch.

PR #7114, opened 2022-07-21
refactor: require grep and map blocks

dipietroR:
Adding `BuiltinFunctions::RequireBlockGrep` and `BuiltinFunctions::RequireBlockMap` to .perlcriticrc


alexgarel (approved):
Great !

PR #7115, opened 2022-07-21
fix: `Ingredients.pm` reused variables

dipietroR:
I'm planning on adding `[Variables::ProhibitReusedNames]` after fixing all of its warnings. There are 2 more reused variable warnings, however I wanted to make sure that I was doing this correctly before working on them

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -1262,7 +1262,6 @@ sub parse_origins_from_text($$) {
 		# Initialize values
 		$matched_ingredient = undef;
 		my $matched_ingredient_ref = {};
-		my $matched_text;
 		my $origins;
 
 		# Call match functions to look for different ways to specify origins etc.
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -1270,7 +1270,7 @@ sub parse_origins_from_text($$) {
 		foreach my $match_function_ref (\&match_origin_of_the_ingredient_origin, \&match_ingredient_origin) {
 			if ($match_function_ref->($product_lc, \$text, $matched_ingredient_ref)) {
 
-				my $matched_text = $matched_ingredient_ref->{matched_text};
+				$matched_text = $matched_ingredient_ref->{matched_text};
```
stephanegigandet, line 1273: Hi @dipietroR ,

For warnings triggered by this rule, we need to analyze what is the behaviour we want:

1. Will we be using the value outside of the loop of the "my $matched_text" that is defined inside the loop?
- In that case, we had a bug before, because "my $matched_text" will create a separate variable, and the value inside the loop will not be affected to the "my $matched_text" variable from outside the loop.

2. If we don't reuse the value outside of the loop, then the variable should be local to the loop. We could use another name for it.

3. Are we using outside of the loop the first variable defined outside of the loop?
- If not, then we don't need the variable from outside of the loop.

dipietroR: Is this what you mean?

stephanegigandet (approved):

PR #7116, opened 2022-07-22
test: add create_product sub and test

yuktea:
- Adds a create_product sub that makes a post request to create a new product.
- Adds a test to create and and read a product.


lib/ProductOpener/APITest.pm
```diff
@@ -124,4 +124,31 @@ sub create_user ($ua, $args_ref) {
 	return;
 }
 
+sub create_product ($ua, $args_ref) {
+	
+	my %fields = (
```
stephanegigandet, line 129: I would remove all the product specific data (from "code" to "packaging_text"), this data should be provided by callers of the function.

e.g. when in the test file you do:

create_product($ua, {});

# edit preference accessible
my $response = $ua->get("http://frontend/cgi/product.pl?type=edit&code=2000000000099");

Anything you test for should be defined in the arguments of the create_product() call, so that's it's possible to understand the test file. Here for instance, it's not obvious that the code 2000000000099 is set by default by create_product()

yuktea: @stephanegigandet thank you! i committed the suggestions

tests/integration/product_create.t
```diff
@@ -0,0 +1,41 @@
...
+
+use ProductOpener::APITest qw/:all/;
+use ProductOpener::Test qw/:all/;
+
+remove_all_products();
+wait_dynamic_front();
+my $ua = new_client();
+create_user($ua, {});
+
+my %args_ref = (
```
stephanegigandet, line 16: As a naming convention, we use the _ref suffix to indicate that a variable is a reference.

If it's a hash, then you can do %args = ( code => '1234' );

If it's a reference, then you need $ instead of % and  {} instead of() : $args_ref = { code => '1234' }

tests/integration/product_create.t
```diff
@@ -0,0 +1,41 @@
...
+	product_name => "Test-75ml",
+	generic_name => "Tester",
+	quantity => "75 ml",
+	link => "https://github.com/openfoodfacts/openfoodfacts-server",
+	expiration_date => "test",
+	ingredients_text => "apple, milk",
+	origin => "france",
+	serving_size => "10g",
+	packaging_text => "no",
+	action => "process",
```
stephanegigandet, line 28: So in the arguments to create_product(), we should pass only data that is specific to the product.

The "action" and "type" can be added in create_product()

".submit" is not needed, it's probably something added by the cgi script

lib/ProductOpener/APITest.pm
```diff
@@ -124,4 +125,16 @@ sub create_user ($ua, $args_ref) {
 	return;
 }
 
+sub create_product ($ua, $args_ref) {
+
+	my $fields;
+
+	while (my ($key, $value) = each %{$fields}) {
```
stephanegigandet, line 132: $fields is undef, so %{$fields} will give an error I think.

tests/integration/product_create.t
```diff
@@ -0,0 +1,41 @@
...
+	type => "add",
+	".submit" => "submit"
+);
+
+create_product($ua, \%args_ref);
```
alexgarel, line 33: As discussed. We should make things more explicit, without loosing the concision in tests.

Your big hashmap could be in a TestDefaults.pm, name `$default_product_form_ref`

(I prefer the term `form` because it's what it is here: the values to fill a form)

```
my $form_ref = create_from_defaults($default_product_form_ref, {code => '200000000099'});
create_product($ua, $form_ref);
```

yuktea: Yup after I get this working on the CI once, will commit these changes

tests/integration/product_create.t
```diff
@@ -0,0 +1,37 @@
...
+	link => "https://github.com/openfoodfacts/openfoodfacts-server",
+	expiration_date => "test",
+	ingredients_text => "apple, milk",
+	origin => "france",
+	serving_size => "10g",
+	packaging_text => "no",
+	".submit" => "submit"
+);
+
+create_product($ua, \%product_feilds);
```
alexgarel, line 29: To create a product you have to be authenticated and you are not !
You should first use create_user (which create user and log you in, because productopener log you in at creation)

alexgarel: maybe you can create a specific user for that:
```suggestion
$admin_ua = new_client();
create_user($admin_ua, {userid => "stephane", email => "stephane@test.com"})
create_product($admin_ua, \%product_feilds);
```

lib/ProductOpener/APITest.pm
```diff
@@ -125,6 +126,32 @@ sub create_user ($ua, $args_ref) {
...
+	serving_size => "10g",
+	packaging_text => "no",
+	".submit" => "submit"
+	);
+
+	while (my ($key, $value) = each %{$product_feilds}) {
+		$fields{$key} = $value;
+	}
+	
+	my $response = $ua->post("http://world.openfoodfacts.localhost/cgi/product_jqm2.pl", Content => \%fields,);
```
alexgarel, line 150: I think this call is to edit a product.

To add a product you should  first add it with a "code" parameter using a post on "cgi/product.pl".

Use you dev instance to add a product (see left column on the search page) with developpers tools opened and see the request made in "network".

stephanegigandet: product.pl is the web form, product_jqm2.pl is the API (we'll have to rename it to something better someday!)

tests/integration/product_create.t
yuktea: That was an important step i was missing :(

tests/integration/product_create.t
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/perl -w
+
+use ProductOpener::PerlStandards;
+
+use Test;
```
stephanegigandet, line 5: This should be use Test::More; I think, unless you wanted to use Test ?

tests/integration/product_create.t
```diff
@@ -0,0 +1,39 @@
...
+use ProductOpener::APITest qw/:all/;
+use ProductOpener::Test qw/:all/;
+
+remove_all_products();
+wait_dynamic_front();
+
+my $ua = new_client();
+
+my %product_feilds = (
```
stephanegigandet, line 14: typo:  feilds instead of fields

tests/integration/product_create.t
```diff
@@ -0,0 +1,39 @@
...
+	link => "https://github.com/openfoodfacts/openfoodfacts-server",
+	expiration_date => "test",
+	ingredients_text => "apple, milk",
+	origin => "france",
+	serving_size => "10g",
+	packaging_text => "no",
+	".submit" => "submit"
+);
+
+$admin_ua = new_client();
```
stephanegigandet, line 29:
```suggestion
my $admin_ua = new_client();
```

tests/integration/create_product.t
```diff
@@ -23,15 +23,16 @@ my %product_feilds = (
...
 	".submit" => "submit"
 );
 
-create_product($ua, \%product_feilds);
+create_user($admin_ua, {});
+create_product($admin_ua, \%product_fields);
 
-# edit preference accessible
-my $response = $ua->get("http://world.openfoodfacts.localhost/cgi/product.pl?type=edit&code=200000000099");
+my $response = $admin_ua->get("http://world.openfoodfacts.localhost/cgi/product.pl?type=edit&code=200000000099");
```
alexgarel, line 34: It would be nice to also test an anomymous user can see it (with a new client)

tests/integration/create_product.t
```diff
@@ -2,16 +2,16 @@
...
-use Test;
+use Test::More;
 use ProductOpener::APITest qw/:all/;
 use ProductOpener::Test qw/:all/;
 
 remove_all_products();
 wait_dynamic_front();
 
-my $ua = new_client();
+my $admin_ua = new_client();
```
alexgarel, line 12: if we keep test, it's not an admin but a simple contributor ;-)

So $contrib_ua might be more adapted.

tests/integration/create_product.t
yuktea: yes! added this check, it works

tests/integration/create_product.t
```diff
@@ -0,0 +1,41 @@
...
+	product_name => "Testttt-75ml",
+	generic_name => "Tester",
+	quantity => "75 ml",
+	link => "https://github.com/openfoodfacts/openfoodfacts-server",
+	expiration_date => "test",
+	ingredients_text => "apple, milk",
+	origin => "france",
+	serving_size => "10g",
+	packaging_text => "no",
+	action => "process",
```
stephanegigandet, line 27: action and type fields can be added by the create_product() function as we always need them.

lib/ProductOpener/APITest.pm
```diff
@@ -125,6 +126,17 @@ sub create_user ($ua, $args_ref) {
 	return;
 }
 
+sub create_product ($ua, $product_fields) {
+	my %fields;
+	while (my ($key, $value) = each %{$product_fields}) {
```
stephanegigandet, line 131: Alex suggested a better way to concatenate one hash into another, something like %a = (%b, %c);

stephanegigandet (approved):
We can merge as-is and you can improve it later.

tests/integration/create_product.t
stephanegigandet: in fact as @alexgarel suggested, we can use the %default_product_form approach to add action and type, so you can disregard this comment

PR #7124, opened 2022-07-22
feat: Deep nest limit 12

dipietroR:
Adding `[ControlStructures::ProhibitDeepNests]` to .perlcriticrc and limiting nesting to 12 (12 appears to be the most amount of nesting done). I'm planning on adding this for now so that no one can add deeper nesting


alexgarel (approved):
LGTM

PR #7129, opened 2022-07-25
chore: built taxonomies

stephanegigandet:
built all taxonomies so that they can be deployed to production in the next release


alexgarel (approved):
:+1:

PR #7130, opened 2022-07-25
feat: Cascading elsif limit 13

dipietroR:
Adding `[ControlStructures::ProhibitCascadingIfElse]` to .perlcriticrc and limiting the amount of if's/elsif's to 13 (13 appears to be the highest amount of if's/elsif's). I'm planning on adding this for now so that no one can add even more than 13


alexgarel (approved):
LGTM ! (sorry I though I had reviewed this morning).

PR #7133, opened 2022-07-25
docs: repl and docker dev

alexgarel:
- Document how to use re.pl
- Update docker dev  docs

### DIFF
--- a/docs/how-to-guides/use-repl.md
+++ b/docs/how-to-guides/use-repl.md
@@ -1,0 +1,46 @@
+# Using Repl
+
+On your local dev instance, the "backend" container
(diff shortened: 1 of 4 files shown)


docs/how-to-guides/use-repl.md
```diff
@@ -0,0 +1,46 @@
+# Using Repl
+
+On your local dev instance, the "backend" container
+comes with [Devel::REPL](https://metacpan.org/pod/) installed.
```
stephanegigandet, line 4:
```suggestion
comes with [Devel::REPL](https://metacpan.org/pod/Devel::REPL) installed.
```

stephanegigandet (approved):
Thanks for the extra documentation!

PR #7134, opened 2022-07-25
fix: Spelling correction

dipietroR:
I'm planning on correcting all of the misspelled comments in perl files and then implementing `[Documentation::PodSpelling]`


stephanegigandet (approved):
Thanks!

PR #7135, opened 2022-07-26
fix: activate the producers platform with PRODUCER_PLATFORM env variable

stephanegigandet:
This fixes the activation of the producers platform (using setenv-pro.sh).

It also removes a lot of warnings when the producers platform is not activated (the variable is expected to have a value, we don't check if it's defined when we use it in the code to activate specific producers platform features.


alexgarel (approved):
LGTM

PR #7136, opened 2022-07-26
fix: fixes redirects

stephanegigandet:
The recent refactor of init() -> init_request() broke the way we did redirects (e.g. when someone make a request for https://fr-fr.openfoodfacts.org/categorie/cafe , we redirect to https://fr.openfoodfacts.org/categorie/cafes )

In fact previously it only worked for requests proceessed by display.pl (so everything except URLs starting by /cgi/)

This fixes it and ensure the request is terminated as soon as we have determined that we should redirect to a different url.


lib/ProductOpener/Display.pm
```diff
@@ -435,6 +435,33 @@ sub process_template($template_filename, $template_data_ref, $result_content_ref
 }
 
 
+=head2 redirect($status_code, $redirect_url)
```
alexgarel, line 438: to @yuktea  : typically a function that, when we will split is part of "http utils" more than display.

alexgarel, line 438: didn't you forget to add it to export ?

lib/ProductOpener/Display.pm
```diff
@@ -435,6 +435,33 @@ sub process_template($template_filename, $template_data_ref, $result_content_ref
...
+
+=cut
+
+sub redirect($status_code, $redirect_url) {
+		
+	my $r = Apache2::RequestUtil->request();
+	
+	$r->headers_out->set(Location => $redirect_url);
+	$r->status($status_code);
+	# note: under mod_perl, exit() will end the request without terminating the Apache mod_perl process
```
alexgarel, line 460: thanks for the comment :-)

lib/ProductOpener/Display.pm
```diff
@@ -917,7 +940,7 @@ sub analyze_request($request_ref) {
 	elsif ((defined $options{redirect_texts}) and (defined $options{redirect_texts}{$lang . "/" . $components[0]})) {
 		$request_ref->{redirect} = $formatted_subdomain . "/" . $options{redirect_texts}{$lang . "/" . $components[0]};
 		$log->info("renamed text, redirecting", { textid => $components[0], redirect => $request_ref->{redirect} }) if $log->is_info();
-		return 301;
+		redirect(302, $request_ref->{redirect});
```
alexgarel, line 943: the change from 301 to 302 is intentional, right ?

lib/ProductOpener/Display.pm
```diff
@@ -472,7 +499,7 @@ sub init_request() {
 	$bodyabout = '';
 	$admin = 0;
 
-	my $r = shift;
+	my $r = Apache2::RequestUtil->request();
```
alexgarel, line 502: you can remove the code below:

```perl
	if (not defined $r) {
		$r = Apache2::RequestUtil->request();
	}
```

alexgarel (approved):
I approve, but I think you forgot to export the function !

lib/ProductOpener/Display.pm
stephanegigandet: yes, 301s make things harder to test and debug and to fix mistakes

lib/ProductOpener/Display.pm
stephanegigandet: thanks

lib/ProductOpener/Display.pm
stephanegigandet: right, I'll add it. I just noticed I was also exporting init which has been renamed to init_request

PR #7144, opened 2022-07-26
refactor: add module for default hashmaps used for integration testing

yuktea:
As recommended, I added a module which will have the default form values to be overridden by the integration tests.


lib/ProductOpener/TestDefaults.pm
```diff
@@ -0,0 +1,74 @@
...
+	pro_checkbox => 0,
+	requested_org => "",
+	team_1 => "",
+	team_2 => "",
+	team_3 => "",
+	action => "process",
+	type => "add",
+	".submit" => "Register"
+);
+
```
alexgarel, line 71: I was expecting that you also add the overriding method.

Because the goal is to remove overriding from the create_xxx methods.

But a google search lead me to this cool pattern, to update a hash with another hash:
https://perlmaven.com/how-to-insert-a-hash-in-another-hash

So really in tests we can do:

```perl
my %params = (%default_user_form, (userid => "bob", email => 'bob@test.com'));
create_xxx($ua, \%params);
```

lib/ProductOpener/TestDefaults.pm
```diff
@@ -0,0 +1,74 @@
...
+	ingredients_text => "apple, milk",
+	origin => "france",
+	serving_size => "10g",
+	packaging_text => "no",
+	action => "process",
+	type => "add",
+	".submit" => "submit"
+);
+
+my %default_user_form = (
```
alexgarel, line 56: I think you should not use "my" to be able to export a variable !

alexgarel (changes requested):
I think you should also 
- remove the default values in APITest.pm
- modify the create_user.t test to see the pattern in action and validate it works.

lib/ProductOpener/TestDefaults.pm
alexgarel: @stephanegigandet might confirm, but we struggled with that yesterday with @LandonPattison !

lib/ProductOpener/TestDefaults.pm
yuktea: this is a very cool one @alex

yuktea: Working on modifying the user creation test

lib/ProductOpener/APITest.pm
```diff
@@ -100,21 +101,8 @@ Call API to create a user
...
-		requested_org => "",
-		team_1 => "",
-		team_2 => "",
-		team_3 => "",
-		action => "process",
-		type => "add",
-		".submit" => "Register"
-	);
+
+	my %fields = (%default_user_form, (userid => "tea", email => 'tea@boba.com'));
```
alexgarel, line 105: Hum maybe you are just testing, but this was not meant to be done here, but outside the create_user function.

The idea is that create_user should really only receive data, The "apply overrides" part below should also be removed.

It's the caller who should construct its arg by attending at the provide default hash map.

```perl
# in create_product.t

my %admin_user_args = (%default_user_form, (userid => "stephane", email => 'stephane@test.com'))
create_user($ua, \%admin_user_args)
```

This is far more flexible, for we maybe could also provides other default hashmap, or hashmap that can be combined.

Eg:

```perl
my %admin_user_args = (%default_user_form, %search_preferences_user_args, (name => "Paul"))
create_user($ua, \%admin_user_args)
```

lib/ProductOpener/TestDefaults.pm
```diff
@@ -0,0 +1,55 @@
...
+	password => "testtest",
+	confirm_password => "testtest",
+	pro_checkbox => 0,
+	requested_org => "",
+	team_1 => "",
+	team_2 => "",
+	team_3 => "",
+	action => "process",
+	type => "add",
+	".submit" => "Register"
```
stephanegigandet, line 52: .submit is not needed, it can be removed

stephanegigandet (approved):

lib/ProductOpener/TestDefaults.pm
yuktea: Yup, this was incorrect, verified and corrected

alexgarel:
Kudos @yuktea !

PR #7148, opened 2022-07-27
Open API specification | Get A product By Barcode

Roxie-32:
- Defined the open API specification file structure
- Included the necessary general information details of the API
- Included the first endpoint: Get A product By Barcode

### DIFF
--- a/docs/reference/api.yml
+++ b/docs/reference/api.yml
@@ -1,0 +1,3204 @@
+---
+openapi: 3.0.1
+info:
+  title: Open Food Facts Open API
+  description: As a developer, the Open Food Facts API allows you to get information and contribute to the products database. You can create great apps to help people make better food choices and also provide data to enhance the database.
+  termsOfService: https://openweathermap.org/terms
+  contact:
+    name: Open Food Facts
+    url: https://slack-ssl-openfoodfacts.herokuapp.com/
+    email: contact@openfoodfacts.org
+  license:
+    name: License (MIT, Apache 2.0, etc)
+    url: https://opendatacommons.org/licenses/odbl/summary/index.html
+  version: "2"
+servers:
+- url: https://world.openfoodfacts.org
+  description: prod
+paths:
+  /api/v2/product/{barcode}:
+    get:
+      tags:
+      - Read Requests
+      summary: Get for a specific product by barcode
+      parameters:
+      - name: barcode
+        in: path
+        description: The barcode of the product to be fetched
+        required: true
+        style: simple
+        explode: false
+        schema:
+          type: string
+          example: "04963406"
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/inline_response_200'
+components:
+  schemas:
+    inline_response_200:
+      type: object
+      properties:
+        code:
+          type: string
+        product:
+          type: object
+          properties:
+            _id:
+              type: string
+            _keywords:
+              type: array
+              items:
+                type: string
+            added_countries_tags:
+              type: array
+              items:
+                type: object
+            additives_debug_tags:
+              type: array
+              items:
+                type: object
+            additives_n:
+              type: integer
+            additives_old_n:
+              type: integer
+            additives_old_tags:
+              type: array
+              items:
+                type: string
+            additives_original_tags:
+              type: array
+              items:
+                type: string
+            additives_prev_original_tags:
+              type: array
+              items:
+                type: string
+            additives_tags:
+              type: array
+              items:
+                type: string
+            additives_tags_n:
+              nullable: true
+            allergens:
+              type: string
+            allergens_from_ingredients:
+              type: string
+            allergens_from_user:
+              type: string
+            allergens_hierarchy:
+              type: array
+              items:
+                type: object
+            allergens_lc:
+              type: string
+            allergens_tags:
+              type: array
+              items:
+                type: object
+            amino_acids_prev_tags:
+              type: array
+              items:
+                type: object
+            amino_acids_tags:
+              type: array
+              items:
+                type: object
+            brands:
+              type: string
+            brands_tags:
+              type: array
+              items:
+                type: string
+            categories:
+              type: string
+            categories_hierarchy:
+              type: array
+              items:
+                type: string
+            categories_lc:
+              type: string
+            categories_old:
+              type: string
+            categories_properties:
+              type: object
+              properties:
+                agribalyse_proxy_food_code:en:
+                  type: string
+                ciqual_food_code:en:
+                  type: string
+            categories_properties_tags:
+              type: array
+              items:
+                type: string
+            categories_tags:
+              type: array
+              items:
+                type: string
+            category_properties:
+              type: object
+              properties:
+                ciqual_food_name:en:
+                  type: string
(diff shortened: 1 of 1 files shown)


docs/reference/api.yml
```diff
@@ -0,0 +1,3204 @@
...
+      responses:
+        "200":
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/inline_response_200'
+components:
+  schemas:
+    inline_response_200:
```
alexgarel, line 43: we should call it "product"

This is the schemas of a product. We will reuse it a lot.

docs/reference/api.yml
```diff
@@ -0,0 +1,3204 @@
...
+servers:
+- url: https://world.openfoodfacts.org
+  description: prod
+paths:
+  /api/v2/product/{barcode}:
+    get:
+      tags:
+      - Read Requests
+      summary: Get for a specific product by barcode
+      parameters:
```
alexgarel, line 24: there are a lot more of possible parameters.

It's ok not to document them now, but:

- we will have to use the equivalent of schema to avoid repeting ourselves (see https://davidgarcia.dev/posts/how-to-split-open-api-spec-into-multiple-files/)
- it's better to add a comment 
 ```suggestion
      # TODO: add all optional parameters
      parameters:
```

File: docs/reference/api.yml [hunk 3]
```diff
@@ -0,0 +1,3204 @@
...
+            vitamins_tags:
+              type: array
+              items:
+                type: object
+        status:
+          type: integer
+        status_verbose:
+          type: string
+      x-examples:
+        example-1:
```
alexgarel, line 1992: We really have to split thing (see https://davidgarcia.dev/posts/how-to-split-open-api-spec-into-multiple-files/) because it wont be manageable to have such big files (impossible to review, because it's very hard to know where you are in the file). If we give meaningful titles to files it will help a lot. Eg. this could be "Product_example_1.yml"

alexgarel: Also in the future we might want to auto-generate this.

docs/reference/api.yml
```diff
@@ -0,0 +1,3204 @@
...
+          type: string
+        product:
+          type: object
+          properties:
+            _id:
+              type: string
+            _keywords:
+              type: array
+              items:
+                type: string
```
alexgarel, line 56: What is the most important for each property is to add a description…

see https://wiki.openfoodfacts.org/API_Fields
https://wiki.openfoodfacts.org/Data_fields

Keep descriptions short, but you can put links to https://wiki.openfoodfacts.org/Data_fields (use url fragment to point to the right paragraph) and eventually add a description in the wiki if needed.

You can ask about any field you are not sure of in the slack.

docs/reference/api.yml
stephanegigandet: in fact /api/v2/{barcode} returns a response that contains a product object:

{
code: "3270160503070",
product: {  ... },
status: 1,
status_verbose: "product found"
}

Can we have a schema that we can call get_product_response, that includes a schema for product ?
That way we can reuse product is search_products_response.

docs/reference/api.yml
```diff
@@ -0,0 +1,3204 @@
...
+            added_countries_tags:
+              type: array
+              items:
+                type: object
+            additives_debug_tags:
```
stephanegigandet, line 61: There are quite a few of those response fields that don't exist anymore, or only on some old products.

Can we remove all the fields that containt "debug" or "old" from the documentation? We don't want anyone to use them.

docs/reference/api.yml
```diff
@@ -0,0 +1,3204 @@
...
+            ecoscore_data:
+              type: object
+              properties:
+                adjustments:
+                  type: object
+                ecoscore_not_applicable_for_category:
+                  type: string
+                status:
+                  type: string
+            ecoscore_extended_data:
```
stephanegigandet, line 232: This is experimental and likely to be removed, I would remove this from the documentation.

docs/reference/api.yml
```diff
@@ -0,0 +1,3204 @@
...
+              type: array
+              items:
+                type: string
+            editors_tags:
+              type: array
+              items:
+                type: string
+            emb_codes:
+              type: string
+            emb_codes_20141016:
```
stephanegigandet, line 313: old, should be removed

docs/reference/api.yml
```diff
@@ -0,0 +1,3204 @@
...
+            image_small_url:
+              type: string
+            image_thumb_url:
+              type: string
+            image_url:
+              type: string
+            images:
+              type: object
+              properties:
+                "1":
```
stephanegigandet, line 368: Here we have a number of images identified by their id "1", "2" etc. we can have any number of them.

docs/reference/api.yml
```diff
@@ -0,0 +1,3204 @@
...
+                      nullable: true
+                    y1:
+                      nullable: true
+                    y2:
+                      nullable: true
+            informers_tags:
+              type: array
+              items:
+                type: string
+            ingredients:
```
stephanegigandet, line 1333: ingredients is a nested array, as ingredients can have sub ingredients

docs/reference/api.yml
Roxie-32: Fields with debug

additives_debug_tags, debug_param_sorted_langs, debug_tags, ingredients_debug, ingredients_ids_debug, ingredients_text_debug, nova_group_debugnutrition_score_debug, additives_debug_tags, debug_param_sorted_langs, debug_tags,  ingredients_debug, ingredients_ids_debug, ingredients_text_debug, nova_group_debug, nutrition_score_debug, 

Fields with old
additives_old_n, additives_old_tags , categories_old, labels_old, origins_old, packaging_old, packaging_old_before_taxonomization, additives_old_n, additives_old_tags, categories_old, labels_old, origins_old, packaging_old, packaging_old_before_taxonomization

@stephanegigandet before I make a push for this, I want to confirm that I am talking all these fields from the get_product_response schema. Removing them entirely.

docs/reference/api.yml
alexgarel: We have to look how to document "dynamic" map with json schema, there seems to be support for this.

docs/reference/api.yml
teolemon: https://openfoodfacts.github.io/api-documentation/#6Understandingresponses 
The source of truth

docs/reference/requestBodies/add_new_photo.yaml
```diff
@@ -0,0 +1,13 @@
...
+                  type: string
+                user_id:
+                  type: string
+                password:
+                  type: string
+              x-examples:
+                example-1:
+                  code: '0074570036004'
+                  user_id: myusername
+                  password: mypassword
```
alexgarel, line 13: I would put examples on the properties directly, it's more readable.

alexgarel:
@Roxie-32 :
This is an OpenAPI 3.0 export of our Postman documentation in order for you to copy paste some parts
[off-openapi3.zip](https://github.com/openfoodfacts/openfoodfacts-server/files/9567134/off-openapi3.zip)

alexgarel (approved):
Let's merge, we will continue with the doc in specific PRs.

PR #7149, opened 2022-07-27
fix: set cookie when redirecting

stephanegigandet:
This should allow the redirect function to set a cookie.

Renamed to redirect_to_url as it was clashing with another function imported in Images.pm


alexgarel (approved):
:rocket:

PR #7150, opened 2022-07-27
refactor: Prohibit mixed boolean operators policy

dipietroR:
Adding `[ValuesAndExpressions::ProhibitMixedBooleanOperators]` to .perlcriticrc

This policy doesn't allow low-precedence booleans to be combined with high-precedence boolean operators in the same expression.

### DIFF
--- a/lib/ProductOpener/KnowledgePanels.pm
+++ b/lib/ProductOpener/KnowledgePanels.pm
@@ -598,7 +598,7 @@ sub create_ecoscore_panel($$$) {
             # Add properties of interest
             foreach my $property (qw(environmental_benefits description)) {
                 my $property_value = get_inherited_property("labels", $labelid, $property . ":" . $target_lc);
-                if (!(defined $property_value) && ($target_lc != "en")) {
+                if (!(defined $property_value) && ($target_lc ne "en")) {
                     # fallback to english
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -598,7 +598,7 @@ sub create_ecoscore_panel($$$) {
             # Add properties of interest
             foreach my $property (qw(environmental_benefits description)) {
                 my $property_value = get_inherited_property("labels", $labelid, $property . ":" . $target_lc);
-                if (not defined $property_value && $target_lc != "en") {
+                if (!(defined $property_value) && ($target_lc != "en")) {
```
stephanegigandet, line 601: I know it was like this before, but ($target_lc != "en") is a bug, it should be ($target_lc ne "en"), as != is for numeric comparison, not string comparison.

e.g. try this:

perl -e "if ('fr' != 'en') { print 'never reached'}"

dipietroR: How does this look?

alexgarel (approved):
I like this rule ;-)

alexgarel:
@dipietroR I changed the prefix to "refactor" because while adding the rule per se is tooling, but as you did change some code, it's a refactor :-)

PR #7155, opened 2022-07-28
refactor: repeated lines typo

dipietroR:
I was checking for some reused variables and found these two variables defined twice. I assume that this is a typo?


alexgarel (approved):
Nice catch :lizard:

PR #7157, opened 2022-07-29
refactor: new function: `apply_organization_settings_to_product()`

dipietroR:
Moving part of `import_csv_file` to decrease the length of the function and putting it in a new function, `apply_organization_settings_to_product`

### DIFF
--- a/lib/ProductOpener/Import.pm
+++ b/lib/ProductOpener/Import.pm
@@ -413,7 +413,14 @@ my %no = (
 
 
 
-sub apply_organization_settings_to_product() {
+sub apply_organization_settings_to_product($$$$$$) {
+	
+	my $imported_product_ref = shift;
+	my $org_id = shift;
+	my $org_ref = shift;
+	my %stats = shift;
+	my $args_ref = shift;
+	my $glns_ref = shift;
 
 	$org_id = get_string_id_for_lang("no_language", $imported_product_ref->{org_name});
 	if ($org_id eq "") {
@@ -546,7 +553,6 @@ sub apply_organization_settings_to_product() {
 
 	store_org($org_ref);
 	}
-
 	return;
 }
 
@@ -673,7 +679,7 @@ sub import_csv_file($) {
 		# If the CSV includes an org_name (e.g. from GS1 partyName field)
 		# set the owner of the product to the org_name
 		if ((defined $imported_product_ref->{org_name}) and ($imported_product_ref->{org_name} ne "")) {
-			apply_organization_settings_to_product()
+			apply_organization_settings_to_product($imported_product_ref, $org_id, $org_ref, $args_ref, $glns_ref, %stats)
 		}
 
 		$Org_id = $org_id;
--- a/lib/ProductOpener/Import.pm
+++ b/lib/ProductOpener/Import.pm
@@ -412,38 +412,6 @@ my %no = (
 );
 
 
-sub apply_organization_settings_to_product($$$$) {
-
-	my %stats = shift;
-	my $org_id = shift;
-	my $org_ref = shift;
-	my $args_ref = shift;
-
-	# Make sure the source as the authorization to load data to the org
-	# e.g. if an org has loaded fresh data manually or through Equadis,
-	# don't overwrite it with potentially stale CodeOnline or USDA data
-
-	if (defined $args_ref->{source_id}) {
-		if (not $org_ref->{"import_source_" . $args_ref->{source_id}}) {
-			$log->debug("skipping import for org without authorization for the source", { org_ref => $org_ref, source_id => $args_ref->{source_id} }) if $log->is_debug();
-			$stats{orgs_without_source_authorization}{$org_id} or $stats{orgs_without_source_authorization}{$org_id} = 0;
-			$stats{orgs_without_source_authorization}{$org_id}++;
-			next;
-		};
-	}
-
-	# The do_not_import_codeonline checkbox will be replaced by the new system above that will work for all sources
-
-	# Check if it is a CodeOnline import for an org with do_not_import_codeonline
-	if ((defined $args_ref->{source_id}) and ($args_ref->{source_id} eq "codeonline")
-		and (defined $org_ref->{do_not_import_codeonline}) and ($org_ref->{do_not_import_codeonline})) {
-		$log->debug("skipping codeonline import for org with do_not_import_codeonline", { org_ref => $org_ref }) if $log->is_debug();
-		next;
-	}
-	return;
-}
-
-
 
 sub get_org_id_for_product($$$$$$) {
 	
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Import.pm
```diff
@@ -413,6 +413,141 @@ my %no = (
 
 
 
+sub apply_organization_settings_to_product() {
+
+	$org_id = get_string_id_for_lang("no_language", $imported_product_ref->{org_name});
+	if ($org_id ne "") {
```
VaiTon, line 419: Maybe also splitting these two if branches into two different methods could improve readability.

lib/ProductOpener/Import.pm
```diff
@@ -534,135 +673,7 @@ sub import_csv_file($) {
...
-							$org_ref->{sources_field}{"org-gs1"}{"partyName"} = $imported_product_ref->{"sources_fields:org-gs1:partyName"};
-						}
-						set_org_gs1_gln($org_ref, $imported_product_ref->{"sources_fields:org-gs1:gln"});
-						$glns_ref = retrieve("$data_root/orgs/orgs_glns.sto");
-					}
-			
-					store_org($org_ref);
-				}
-			}
+			apply_organization_settings_to_product()
```
stephanegigandet, line 676: You will need to pass $imported_product_ref and anything the function changes as parameter.

Same thing for $org_id, it's declared in import_csv_file() .

So maybe a first function could be get_org_id_for_product($imported_product_ref);

then apply_organization_settings_to_product($imported_product_ref, $org_id)

stephanegigandet:
Hi @dipietroR , it passes the tests but there are some issues with parameters passing that cause bugs.

I added more tests so that we can catch those problems. You can merge this PR into your branch to get the new tests:
import_csv_file_test https://github.com/openfoodfacts/openfoodfacts-server/pull/7168

lib/ProductOpener/Import.pm
```diff
@@ -412,6 +412,167 @@ my %no = (
 );
 
 
+sub apply_organization_settings_to_product($$$$) {
```
stephanegigandet, line 415: I think there is a naming issue, as what the code below does is to check if the source is authorized to load data for the organization.

The apply_organization_settings_to_product is good name, but for the part of the code that was below that check, starting after # If it is a GS1 import (Equadis, CodeOnline, Agena3000), check if the org is associated with known issues

lib/ProductOpener/Import.pm
```diff
@@ -412,6 +412,167 @@ my %no = (
...
+	# Make sure the source as the authorization to load data to the org
+	# e.g. if an org has loaded fresh data manually or through Equadis,
+	# don't overwrite it with potentially stale CodeOnline or USDA data
+
+	if (defined $args_ref->{source_id}) {
+		if (not $org_ref->{"import_source_" . $args_ref->{source_id}}) {
+			$log->debug("skipping import for org without authorization for the source", { org_ref => $org_ref, source_id => $args_ref->{source_id} }) if $log->is_debug();
+			$stats{orgs_without_source_authorization}{$org_id} or $stats{orgs_without_source_authorization}{$org_id} = 0;
+			$stats{orgs_without_source_authorization}{$org_id}++;
+			next;
```
stephanegigandet, line 431: this "next" statement was used to get out of the loop that goes through each product, but as it is now in a separate function, it's not going to work.

instead the function should return a value to indicate if the source is authorized to update products from the organization, and if not, we would then call next.

e.g. something like 

if (not source_is_authorized_to_update_products_from_the_org(..) ) {
  next;
}

lib/ProductOpener/Import.pm
```diff
@@ -412,6 +412,167 @@ my %no = (
 );
 
 
+sub apply_organization_settings_to_product($$$$) {
+
+	my %stats = shift;
```
stephanegigandet, line 417: hashes need to be passed by references, so it should be my $stats_ref, in the function should be called with \%stats as a parameter.

lib/ProductOpener/Import.pm
```diff
@@ -412,6 +412,167 @@ my %no = (
...
+	}
+	return;
+}
+
+
+
+sub get_org_id_for_product($$$$$$) {
+	
+	my $imported_product_ref = shift;
+	my $org_id = shift;
```
stephanegigandet, line 451: if $org_id is passed as a parameter, if you change it inside the function, it won't change its value outside the function.

So instead, the function should return the $org_id, and it should be called like this:

my $org_id = get_org_id_for_product(...)

lib/ProductOpener/Import.pm
```diff
@@ -412,6 +412,167 @@ my %no = (
...
+	my $org_ref = shift;
+	my %stats = shift;
```
stephanegigandet, line 453: the hash needs to be passed by reference

lib/ProductOpener/Import.pm
```diff
@@ -412,6 +412,167 @@ my %no = (
...
+		defined $stats{orgs_existing}{$org_id} or $stats{orgs_existing}{$org_id} = 0;
+		$stats{orgs_existing}{$org_id}++;
+
+		# Make sure the source as the authorization to load data to the org
+		# e.g. if an org has loaded fresh data manually or through Equadis,
+		# don't overwrite it with potentially stale CodeOnline or USDA data
+
+		apply_organization_settings_to_product(%stats, $org_id, $org_ref, $args_ref);
+
+		# If it is a GS1 import (Equadis, CodeOnline, Agena3000), check if the org is associated with known issues
```
stephanegigandet, line 508: The code below should go in a function named apply_organization_settings_to_product() (different from the one that you added where it checks the authorization).

dipietroR:
@stephanegigandet I'm going to make this PR just for one function and when this PR gets merged, I'll work on splitting it into another function because it's getting a little confusing.

lib/ProductOpener/Import.pm
dipietroR: Why would we change it? It works fine as is. 

And when I try to change it I get this error:
```
[Tue Aug  2 14:43:10 2022] import_file_process.pl: Global symbol "%stats_ref" requires explicit package name (did you forget to declare "my %stats_ref"?) at lib/ProductOpener/Import.pm line 524.
```

Ban3: Perl passes arguments as a list of scalars. The prototype is also coercing the arguments to scalars with the `$` sigil anyway.

Try this:

``` perl
use Data::Dumper;

sub get_org_id_for_product($$$$$$) {

    my $imported_product_ref = shift;
    my $org_id               = shift;
    my $org_ref              = shift;
    my %stats                = shift;
    my $args_ref             = shift;
    my $glns_ref             = shift;

    print Data::Dumper->Dump(
        [   $imported_product_ref,
            $org_id, $org_ref, \%stats, $args_ref, $glns_ref
        ],
        [qw(imported_product_ref org_id org_ref *stats args_ref glns_ref)]
    );
}

my %stats = (
    'products_in_file'                 => {},
    'products_already_existing'        => {},
    'products_created'                 => {},
    'products_data_updated'            => {},
    'products_data_not_updated'        => {},
    'products_info_added'              => {},
    'products_info_changed'            => {},
    'products_info_updated'            => {},
    'products_info_not_updated'        => {},
    'products_nutrition_added'         => {},
    'products_nutrition_changed'       => {},
    'products_nutrition_updated'       => {},
    'products_nutrition_not_updated'   => {},
    'products_images_added'            => {},
    'products_with_images'             => {},
    'products_with_data'               => {},
    'products_with_info'               => {},
    'products_with_ingredients'        => {},
    'products_with_nutrition'          => {},
    'products_with_nutrition_prepared' => {},
    'products_without_images'          => {},
    'products_without_data'            => {},
    'products_without_info'            => {},
    'products_without_info'            => {},
    'products_without_nutrition'       => {},
    'products_updated'                 => {},

    # Keep track of the number of products that are not imported
    # because the source does not have authorization for the org
    'orgs_without_source_authorization' => {},
    'orgs_created'                      => {},
    'orgs_existing'                     => {},
    'orgs_in_file'                      => {},
);

get_org_id_for_product( "product_ref", "org_id", "org_ref", %stats,
    "args_ref", "glns_ref" );
```

Output:
``` shell
Odd number of elements in hash assignment at hashpass.pl line 11.
$imported_product_ref = 'product_ref';
$org_id = 'org_id';
$org_ref = 'org_ref';
%stats = (
           '29' => undef
         );
$args_ref = 'args_ref';
$glns_ref = 'glns_ref';
```

As you can see, it's not working fine because `%stats` gets evaluated in scalar context, resulting in the number of items in the hash.
You need to take reference of the hash with `\` on the caller side and change the subroutine to work with the reference instead

``` perl
...
my $stats_ref = shift # Note the sigil $, references are scalars.
...
```

The rest of the subroutine also needs to be modified to work with the reference instead of the hash.

PR #7166, opened 2022-08-01
refactor: enable perl signatures in ImportConvert.pm

dipietroR:
Replacing "shift" prototypes with perl signatures.

There are a few lines like this:
```
                shift @{$xml_fields_mapping_ref};
```
And I'm not sure what to do with them. Could they be deleted?

### DIFF
--- a/lib/ProductOpener/ImportConvert.pm
+++ b/lib/ProductOpener/ImportConvert.pm
@@ -43,8 +43,7 @@ convert the product data they contain to a format that can be imported on Open F
 
 package ProductOpener::ImportConvert;
 
-use utf8;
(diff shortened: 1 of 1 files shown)


Ban3:
> And I'm not sure what to do with them. Could they be deleted?

No. Those are modifying `xml_fields_mapping` so they are needed.

stephanegigandet (approved):

stephanegigandet:
> > And I'm not sure what to do with them. Could they be deleted?
> 
> No. Those are modifying `xml_fields_mapping` so they are needed.

Right, those shift are different from the shift we use to get the function parameters.

PR #7168, opened 2022-08-02
refactor: compare_to_expected_result function + more import_csv_file tests

stephanegigandet:
This PR:

Moves some code that we have in many tests (like in nutriscore.t) to a new function in Test.pm:
	compare_to_expected_results($product_ref, "$expected_dir/$testid.json", $update_expected_results);

Adds more tests to import_csv_file.t:
- 3 products instead of 1
- products with different organizations
- test the result of the stats that are returned by the import_csv_file function.


stephanegigandet:
Hi @dipietroR , it passes the tests but there are some issues with parameters passing that cause bugs.

I added more tests so that we can catch those problems. You can merge this PR into your branch to get the new tests:
import_csv_file_test

yuktea (approved):

yuktea:
Thank you, this will be a very useful change for the API tests.

PR #7170, opened 2022-08-02
test: add search integration test for the Open Food Facts API 

yuktea:
Integration test for searching using the API and different parameters


tests/integration/search_test.t
```diff
@@ -0,0 +1,126 @@
...
+use JSON::PP;
+use ProductOpener::TestDefaults qw/:all/;
+use ProductOpener::Test qw/:all/;
+use Getopt::Long;
+
+my $test_name = "search_test";
+my $tests_dir = dirname(__FILE__);
+my $expected_dir = $tests_dir . "/expected_test_results/" . $test_name;
+
+my %product_fields_1 = (
```
stephanegigandet, line 21: Instead of m%products_fields_1, 2, 3 etc. you could declare an array of products.

e.g.

my @products = ( 
{
},
{
}
...

That way it's easy to add more products without having to declare new variables.

tests/integration/search_test.t
```diff
@@ -0,0 +1,126 @@
...
+
+my %create_user_args = (%default_user_form, (email => 'bob@gmail.com'));
+create_user($ua, \%create_user_args);
+
+create_product($ua, \%product_fields_1);
+create_product($ua, \%product_fields_4);
+create_product($ua, \%product_fields_5);
+create_product($ua, \%product_fields_6);
+
+my $query_1_json = get("http://world.openfoodfacts.localhost//cgi/search.pl?action=process&json=1");
```
stephanegigandet, line 107: Same thing for queries, you could just declare an array of queries. In fact it would be good to list a name as well, so that we know what we test.

e.g.

my @tests = (
["basic_query", "http://world.openfoodfacts.localhost//cgi/search.pl?action=process&json=1"],
["products_without_palm_oil", "http://world.openfoodfacts.localhost/api/v2/search?code=200000000039,200000000038&fields=code,product_name"],
);

and then iterate on the array. There are many tests that declare an array of @tests like that, for instance nutriscore.t

tests/integration/search_test.t
```diff
@@ -0,0 +1,126 @@
...
+my $query_1_json_decoded = decode_json($query_1_json);
+
+my $query_2_json = get("http://world.openfoodfacts.localhost/cgi/search.pl?action=process&json=1&ingredients_from_palm_oil=without");
```
stephanegigandet, line 110: I think the 2 queries have the exact same result. It would be great to put queries that return different results. (which means the products might need to change a bit)

stephanegigandet (approved):
Looks good to me, but could be refactored a bit.

tests/integration/search_test.t
yuktea: Did you mean an array of all these hashmaps?

tests/integration/search_test.t
yuktea: thanks for suggesting this :)

tests/integration/search_test.t
yuktea: yes!

tests/integration/search_test.t
```diff
@@ -0,0 +1,167 @@
...
+	[
+		"q4",
+		construct_test_url(
+			"/cgi/search.pl?action=process&tagtype_0=categories&tag_contains_0=contains&tag_0=breakfast_cereals&tagtype_1=nutrition_grades&tag_contains_1=contains&ingredients_from_palm_oil=without&json=1",
+			"world"
+		)
+	]
+);
+
+## TODO : find out whats up with these quesries below :(
```
stephanegigandet, line 140: I think it would be a good idea to print more info in the test logs if some query fails.
e.g. like this: https://github.com/openfoodfacts/openfoodfacts-server/pull/7294

tests/integration/search_test.t
```diff
@@ -0,0 +1,167 @@
...
+	my $decoded_json = decode_json($json);
+
+	my $length = @{$decoded_json->{'products'}};
+	my $count;
+	for ($count = 0; $count < $length; $count++) {
+		foreach my $field ("created_t", "last_modified_t") {
+			delete($decoded_json->{'products'}[$count]{created_t});
+		}
+	}
+	is(compare_to_expected_results($decoded_json, "$expected_dir/$testid.json", $update_expected_results), 1);
```
stephanegigandet, line 164: We need to call the functions to remove dates from fields, if I run the tests today, I get different values from the ones you had yesterday:

```
          "entry_dates_tags" : [
-            "2022-09-04",
+            "2022-09-05",
             "2022-09",
             "2022"
          ],
@@ -495,13 +495,13 @@
             "en:1"
          ],
          "last_edit_dates_tags" : [
-            "2022-09-04",
+            "2022-09-05",
             "2022-09",
             "2022"
          ],
          "last_editor" : "tests",
          "last_modified_by" : "tests",
-         "last_modified_t" : 1662334632,
+         "last_modified_t" : 1662364495,

```

yuktea: I'll add these fields to the loop where we are sanitizing the hash

tests/integration/search_test.t
```diff
@@ -0,0 +1,178 @@
...
+
+wait_dynamic_front();
+
+my $ua = new_client();
+
+my %create_user_args = (%default_user_form, (email => 'bob@gmail.com'));
+create_user($ua, \%create_user_args);
+
+foreach my $ref (@products) {
+	create_product($ua, $ref);
```
alexgarel, line 119: why don't you use a %default_product_form here so that your array above could be lighter ?

teolemon:
Per a discussion in https://docs.google.com/document/d/1LL8-aiSF482xaJ1o0AKmhXB5QWfVE0_jzvYakq3VXys/edit# (2022-09-05, this is blocked on the CGI route changes)

lib/ProductOpener/APITest.pm
```diff
@@ -146,6 +146,14 @@ For the example cited above this returns: "http://world-fr.openfoodfacts.localho
 sub construct_test_url ($target, $prefix) {
 	my $link = "openfoodfacts.localhost";
 	my $url = "http://${prefix}.${link}${target}";
+
+	my $api_end = "/api/v2/search?";
+	my $cgi_end = "/cgi/search.pl?action=process&";
```
stephanegigandet, line 151: So in fact we want to replicate what nginx does: all URLs that do not start with /cgi/ are prepended with /cgi/display.pl? :

e.g. https://world.openfoodfacts.org/api/v2/search?brands_tags=nestle --> https://world.openfoodfacts.org/cgi/display.pl?/api/v2/search?brands_tags=nestle

stephanegigandet: See https://github.com/openfoodfacts/openfoodfacts-server/blob/main/conf/nginx.conf#L110

tests/integration/search_test.t
```diff
@@ -0,0 +1,178 @@
...
+		fail($testid);
+		next;
+	};
+
+	my $length = @{$decoded_json->{'products'}};
+	my $count;
+	for ($count = 0; $count < $length; $count++) {
+		foreach my $field ("created_t", "last_modified_t", "last_edit_dates_tags", "entry_dates_tags") {
+			delete($decoded_json->{'products'}[$count]{$field});
+		}
```
alexgarel, line 173: there is a helper function to do that in Test.pm: normalize_product_for_test_comparison

tests/integration/search_test.t
```diff
@@ -0,0 +1,178 @@
...
+		ingredients_text => "apple, water",
+		origin => "France",
+		serving_size => "5g",
+		packaging_text => "no",
+		action => "process",
+		categories => "breakfast cereals",
+		type => "add",
+		".submit" => "submit"
+	}
+);
```
alexgarel, line 105: you did not take into account [this comment](https://github.com/openfoodfacts/openfoodfacts-server/pull/7170/files/aafebe3ee9c499ddbd42aaafdc86d871bfd1c34c#r962885023)

You should have a %default_product (in TestDefault.pm) and use it here to avoid having big json.

tests/integration/search_test.t
yuktea: let me check that out

tests/integration/search_test.t
yuktea: @alexgarel okay :)

lib/ProductOpener/APITest.pm
yuktea: I'll add one for `display.pl` too, not sure if there could be more, thanks for reviewing the pr!

yuktea:
So, this is clearly not right. The hashmaps are not being updated with the `$product_form_override`

tests/integration/search_test.t
yuktea: @alexgarel : I tried to make the array lighter, it's somehow working in a weird way where it only takes the value of the default hash map for all the products.
I suspect it's a problem with the form this over-riding hashmap is passed in edit/create_product.

tests/integration/search_test.t
yuktea: Looks like we can merge it now!

alexgarel:
@yuktea I will maybe add a commit to re-use existing method to clean data (I think it's worth it as this test will be used as a model to others)

PR #7171, opened 2022-08-02
refactor: enable perl signature Images.pm

dipietroR:
Replacing "shift" perl protoypes with signatures.


stephanegigandet (approved):
Thank you!

PR #7179, opened 2022-08-04
refactor: fixing typos

dipietroR:
Correcting spelling errors


stop_words.txt
```diff
@@ -156,10 +156,13 @@ submodules
 sucre
 tagid
 TAGID
+tagtype
```
VaiTon, line 159: I don't think this are part of the fixing typos PR, are they?

stephanegigandet: I'm guessing that's a list of words that should be ignored by the tool used to check for typos. Is that correct @dipietroR ? Is it possible to add a comment at the top of this file to indicate what it is / how it works? e.g. how does one check for typos?

Ban3: The file was added for [PodSpelling](https://metacpan.org/pod/Perl::Critic::Policy::Documentation::PodSpelling) policy. Comments should be fine.

stop_words.txt
```diff
@@ -1,3 +1,7 @@
+This file consists of words that the PodSpelling perlcritic policy thinks are incorrect,
+but are spelled correctly (e.g acronyms, words in other languages, etc)
```
alexgarel, line 2: Commenting follows perl rules, see https://metacpan.org/pod/Pod::Spell 

> The latter is looked at line by line, with anything after an octothorp ("#") removed and then leading and trailing whitespace removed

```suggestion
# This file consists of words that the PodSpelling perlcritic policy thinks are incorrect,
# but are spelled correctly (e.g acronyms, words in other languages, etc)
```

I'm applying this immediatly

alexgarel (approved):
Good

PR #7180, opened 2022-08-04
refactor: reduce nesting in Export.pm

dipietroR:
Taking the most deeply nested part of Export.pm and putting it into a separate function.

### DIFF
--- a/lib/ProductOpener/Export.pm
+++ b/lib/ProductOpener/Export.pm
@@ -279,7 +279,7 @@ sub export_csv($args_ref) {
 				elsif ($group_id eq "images") {
 					if ($args_ref->{include_images_paths}) {
 						if (defined $product_ref->{images}) {
-							(%other_images, %populated_fields) = include_image_paths($product_ref, %populated_fields, %other_images);
+							(%other_images, %populated_fields) = include_image_paths($product_ref, \%populated_fields, \%other_images);
 						}
 					}
 				}
@@ -592,10 +592,8 @@ sub export_csv($args_ref) {
 }
 
 
-sub include_image_paths($$$) {
-	my $product_ref = shift;
-	my %populated_fields = shift;
-	my %other_images = shift;
+
+sub include_image_paths($product_ref, $populated_fields_ref, $other_images_ref) {
 
 	# First list the selected images
 	my %selected_images = ();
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Export.pm
```diff
@@ -624,5 +594,43 @@ sub export_csv($) {
 }
 
 
+sub include_image_paths($$$) {
+	my $product_ref = shift;
+	my %populated_fields = shift;
+	my %other_images = shift;
```
Ban3, line 600: Same issue as in [#7157](https://github.com/openfoodfacts/openfoodfacts-server/pull/7157#discussion_r935369936), hashes can't be passed like that.

dipietroR: Would it work if we used the perl signature instead?
Like:
```
sub include_image_paths($product_ref, %populated_fields, %other_images) {
```
Or is there no way around it?

Ban3: With signatures, there can be a single hash as a final argument that slurps remaining arguments as key-value pairs, but not two.

References are the way to go with passing a hash to a subroutine. See [perlreftut](https://perldoc.perl.org/perlreftut) and [perlref](https://perldoc.perl.org/perlref) for how to work with refs.

There's also no need to return the values when they're references.

stephanegigandet: @Ban3 is right, you need to pass references to the hashes, instead of the hashes themselves.
The first argument works like this, we pass $product_ref (the _ref at the end is a naming convention to indicate that it is a reference) which a reference to a product hash. So instead of passing %some_hash, we pass $some_hash_ref. You can get a reference for a hash like this:  $my_hash_ref = \%my_hash.
Then when you read or write $my_hash_ref->{key}, it's like reading or writing %my_hash{key}.

Ban3: > You can get a reference for a hash like this: $my_hash_ref = %my_hash.

The backslash probably got interpreted as escape character by Markdown, should be `$my_hash_ref = \%my_hash`.

dipietroR: Would it be like this?
```
sub include_image_paths($product_ref, $populated_fields_ref, $other_images_ref) {
        my $populated_fields_ref = %populated_fields;
        my $other_images_ref = %other_images;
```
Or would it be outside of the function?

dipietroR: Ok, thank you

lib/ProductOpener/Export.pm
```diff
@@ -279,37 +279,7 @@ sub export_csv($args_ref) {
...
-							foreach my $imageid (sort keys %{$product_ref->{images}}) {
-
-								if (($imageid =~ /^(\d+)$/) and (not defined $selected_images{$imageid})) {
-									$other++;
-									$populated_fields{"image_" . "other_" . $other . "_file"} = sprintf("%08d", 10 * 1000 ) . "_" . "other_" . $other;
-									# Keep the imgid for second loop on products
-									$other_images{$product_ref->{code} . "." . "other_" . $other} = { imgid => $imageid};
-								}
-							}
+							(%other_images, %populated_fields) = include_image_paths($product_ref, \%populated_fields, \%other_images);
```
Ban3, line 282:
```suggestion
							include_image_paths($product_ref, \%populated_fields, \%other_images);
```
Now that it's working with references, `%populated_fields` and `%other_images` get modified in the subroutine. No need to return anything from the subroutine.

lib/ProductOpener/Export.pm
```diff
@@ -622,5 +592,41 @@ sub export_csv($args_ref) {
...
+	foreach my $imageid (sort keys %{$product_ref->{images}}) {
+
+		if (($imageid =~ /^(\d+)$/) and (not defined $selected_images{$imageid})) {
+			$other++;
+			$populated_fields_ref->{"image_" . "other_" . $other . "_file"} = sprintf("%08d", 10 * 1000 ) . "_" . "other_" . $other;
+			# Keep the imgid for second loop on products
+			$other_images_ref->{$product_ref->{code} . "." . "other_" . $other} = { imgid => $imageid};
+		}
+	}
+	return ($other_images_ref, $populated_fields_ref);
```
Ban3, line 628: Same as above. No need to return the refs.

lib/ProductOpener/Export.pm
dipietroR: Sounds good!

lib/ProductOpener/Export.pm
dipietroR: Alright, I'll change that

stephanegigandet (approved):
Looks good, thank you!

PR #7184, opened 2022-08-05
refactor: enable perl signature Import.pm

dipietroR:
Replacing "shift" perl prototypes with signatures in Import.pm

### DIFF
--- a/lib/ProductOpener/Import.pm
+++ b/lib/ProductOpener/Import.pm
@@ -54,8 +54,7 @@ It is also used in the C<scripts/import_csv_file.pl> script.
 
 package ProductOpener::Import;
 
-use utf8;
-use Modern::Perl '2017';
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Looks good, thank you!

lib/ProductOpener/Import.pm
```diff
@@ -54,8 +54,7 @@ It is also used in the C<scripts/import_csv_file.pl> script.
 
 package ProductOpener::Import;
 
-use utf8;
-use Modern::Perl '2017';
+use ProductOpener::PerlStandards
```
stephanegigandet, line 57: The test is failing, probably because this line is missing a ; at the end

dipietroR: Oops! I'll fix that

PR #7185, opened 2022-08-06
taxonomy: add breading and whole grain flour blend

advaitathreya:
### What
Made two edits to ingredients list, to add breading as a synonym for bread crumbs, and added whole grain flour blend as a child of flour blend. Fixes issue on entry for [Trader Joe's chickenless crispy tenders](https://world.openfoodfacts.org/product/00978996/chickenless-crispy-tenders-trader-joe-s)

### Screenshot
image


stephanegigandet (approved):
Perfect, thank you!

PR #7187, opened 2022-08-08
refactor: enable perl signature Product.pm

dipietroR:
Replacing "shift" perl prototypes with perl signatures


stephanegigandet (approved):
Thanks!

PR #7188, opened 2022-08-09
refactor: enable perl signature `Data.pm`

dipietroR:
Replacing "shift" prototypes with perl signatures

### DIFF
--- a/lib/ProductOpener/Data.pm
+++ b/lib/ProductOpener/Data.pm
@@ -43,8 +43,7 @@ GitHub, where some additional context is available.
 
 package ProductOpener::Data;
 
-use utf8;
-use Modern::Perl '2017';
+use ProductOpener::PerlStandards;
 use Exporter    qw< import >;
 
 BEGIN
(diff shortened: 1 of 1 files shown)


stephanegigandet:
Failing test:

Too few arguments for subroutine 'ProductOpener::Data::get_products_collection' at /opt/product-opener/lib/ProductOpener/Test.pm line 81.
tests/unit/import_convert_carrefour_france.t .. 
Dubious, test returned 255 (wstat 65280, 0xff00)

alexgarel (dismissed):
Cool :-)

alexgarel (changes requested):
Sorry I spotted a problem:
The problem you had [here](https://github.com/openfoodfacts/openfoodfacts-server/pull/7188#issuecomment-1218131799) about too few arguments, will happen for other functions in scripts. This is because many scripts still uses the old standard of verifying the number of arguments.

We can dismiss this problem by using an explicit $timeout=undef on all functions that use $timeout parameter.

alexgarel (approved):
Good.

PR #7194, opened 2022-08-10
refactor: enable perl signature Producers.pm

dipietroR:
Replacing "shift" prototypes with perl signatures.


stephanegigandet (approved):
Thanks!

PR #7195, opened 2022-08-10
refactor: enable perl signature `Store.pm`

dipietroR:
Replacing "shift" prototypes with perl signatures

### DIFF
--- a/tests/unit/store.t
+++ b/tests/unit/store.t
@@ -8,7 +8,7 @@ use Log::Any::Adapter 'TAP';
 
 use ProductOpener::Store qw/:all/;
 
-is( get_fileid('Do not challenge me!', 0, 'en'), 'do-not-challenge-me' );
+is( get_fileid('Do not challenge me!'), 'do-not-challenge-me' );
 
 my @tests = ("Bonjour !", "Café Olé! 3€ -10%", "No hablo Español, señor", "สำนักงานคณะกรรมการกลางอิสลามแห่งประเทศไทย, คณะกรรมการกลางอิสลามแห่งประเทศไทย", "예네버르", "ラム酒", "DLG Jährlich Prämiert", "fr:Bœuf");
 foreach my $test (@tests) {
--- a/tests/unit/store.t
+++ b/tests/unit/store.t
@@ -8,7 +8,7 @@ use Log::Any::Adapter 'TAP';
 
 use ProductOpener::Store qw/:all/;
 
-is( get_fileid('Do not challenge me!'), 'do-not-challenge-me' );
+is( get_fileid('Do not challenge me!', 0, 'en'), 'do-not-challenge-me' )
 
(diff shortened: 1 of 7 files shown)


Ban3:
Actually, an additional way to handle too few arguments would be giving the parameters default values, if we don't want to touch the calls. 

```perl
sub get_fileid($file, $unaccent=undef, $lc=undef)
```
Would match the old behavior when called with one argument.

I guess too many arguments could be handled in the signature with a dummy `@`, but I think correcting the calls makes more sense in that case.

Ban3:
> Would making it like this be the same?

No, that would still require 3 arguments and overwrite `$unaccent` and `$lc` with undef even when they are provided. The only way to make them optional is to give them a value in the signature.

tests/unit/store.t
```diff
@@ -8,7 +8,7 @@ use Log::Any::Adapter 'TAP';
 
 use ProductOpener::Store qw/:all/;
 
-is( get_fileid('Do not challenge me!'), 'do-not-challenge-me' );
+is( get_fileid('Do not challenge me!', 0, 'en'), 'do-not-challenge-me' )
```
Ban3, line 11:
```suggestion
is( get_fileid('Do not challenge me!', 0, 'en'), 'do-not-challenge-me' );
```
It should also work with just one argument now that the rest of the arguments are optional.

Ban3:
Next failure:

> Too few arguments for subroutine 'ProductOpener::Store::get_urlid' at tests/unit/store.t line 16.

https://github.com/openfoodfacts/openfoodfacts-server/blob/76395069b4c75edfa465fae41517ba8d98af3e0e/tests/unit/store.t#L16

This can be solved similarly to `get_fileid`, make only one argument required and the rest optional.

Ban3 (approved):

tests/unit/store.t
```diff
@@ -8,7 +8,7 @@ use Log::Any::Adapter 'TAP';
 
 use ProductOpener::Store qw/:all/;
 
-is( get_fileid('Do not challenge me!'), 'do-not-challenge-me' );
+is( get_fileid('Do not challenge me!', 0, 'en'), 'do-not-challenge-me' );
```
alexgarel, line 11: We have to keep this test as is, because there are many places where we call get_fileid with only first arguments, and this has to work. (The real fix is that you put `undef` as defaults in function signature).

```suggestion
is( get_fileid('Do not challenge me!'), 'do-not-challenge-me' );
```

dipietroR:
@stephanegigandet 2e41f9720183a6e465318860a9f720aef15e6ce7 doesn't seem related to this PR, did you perhaps commit it to the wrong branch? And if it is related, could you please explain how? :-)

stephanegigandet:
> @stephanegigandet [2e41f97](https://github.com/openfoodfacts/openfoodfacts-server/commit/2e41f9720183a6e465318860a9f720aef15e6ce7) doesn't seem related to this PR, did you perhaps commit it to the wrong branch? And if it is related, could you please explain how? :-)

ah right I made a mistake I think, sorry

PR #7197, opened 2022-08-10
refactor: enable perl signatures `Ingredients.pm`

dipietroR:
Replacing "shift" prototypes with signatures in Ingredients.pm

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -51,8 +51,7 @@ ingredients (is the product vegetarian, vegan, does it contain palm oil etc.)
 
 package ProductOpener::Ingredients;
 
-use utf8;
-use Modern::Perl '2017';
+use ProductOpener::PerlStandards;
 use Exporter    qw< import >;
 
 BEGIN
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -5822,13 +5688,7 @@ sub replace_allergen_in_caps($$$$) {
...
 
 
-sub replace_allergen_between_separators($$$$$$) {
-	my $language = shift;
-	my $product_ref = shift;
-	my $start_separator = shift;
-	my $allergen = shift;
-	my $end_separator = shift;
-	my $before = shift;
+sub replace_allergen_between_separators($language, $product_ref, $before, $start_separator, $end_separator, $allergen) {
```
stephanegigandet, line 5691: The arguments are not in the same order, so they will be mistmatched.

stephanegigandet (approved):
I looked quickly and found 2 issues that may be responsible for the failing tests. It would be good to review the changes to make sure all the arguments are in the same order as before, otherwise we might introduce bugs.

lib/ProductOpener/Ingredients.pm
dipietroR: Alright, sorry I didn't know that would/might cause errors

PR #7198, opened 2022-08-10
taxonomy: Add new packaging materials and translations

helloworldtest123:
### What
Many new packaging materials and improvements to the recycling codes. I am very open to improvements of my changes and I hope I did not make a stupid mistake.


stephanegigandet (approved):
Looks good to me, thank you very much!

helloworldtest123:
Looks like the packaging test still expects the old definition of 84 C/PAP and therefore fails. I could possibly modify the test. But it would be better if someone who knows more about it could do it.

stephanegigandet (approved):
Thank you!

PR #7200, opened 2022-08-11
taxonomy: Remove repeated entries

BryanH01:
Removed repeated entries
Added some missing elements : language code, colon...


stephanegigandet (approved):
Thank you very much @BryanH01

PR #7208, opened 2022-08-15
refactor: enable perl signature Food.pm

dipietroR:
Replacing "shift" with signatures

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -176,7 +176,7 @@ Output modifier reference.
 
 sub normalize_nutriment_value_and_modifier($value_ref, $modifier_ref) {
 
-	$modifier_ref = undef;
+	${$modifier_ref} = undef;
 
 	return if not defined ${$value_ref};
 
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Food.pm
```diff
@@ -175,12 +174,9 @@ Output modifier reference.
...
 =cut
 
-sub normalize_nutriment_value_and_modifier($$) {
+sub normalize_nutriment_value_and_modifier($value_ref, $modifier_ref) {
 
-	my $value_ref = shift;
-	my $modifier_ref = shift;
-
-	$$modifier_ref = undef;
+	$modifier_ref = undef;
```
Ban3, line 179:
```suggestion
	${$modifier_ref} = undef;
```
I believe the tests are failing because the reference instead of the value is set to undef here.

dipietroR: Oh, thanks for fixing it! I thought the $$ was a typo :sweat_smile:

stephanegigandet (approved):
Looks good, thank you!

PR #7209, opened 2022-08-15
test: Integration test for editing a product

yuktea:


stephanegigandet:
Closing as this is now done in https://github.com/openfoodfacts/openfoodfacts-server/blob/main/tests/integration/api_v2_product_write.t

Issue #5684, opened 2021-09-15
Ideas for new tests


teolemon:
### What
- [ ] Periodic tests
  - [ ] Test that OCR is working well on image upload, so that we make sure that Robotoff can work well
  - [ ] Test we receive pings from Robotoff
- [ ] …

### Part of
- #5512

PR #7216, opened 2022-08-16
test: add test for deleting product with internal call

yuktea:
### What

- This test creates a user, and a product and then uses the call to remove products and stores the results to the search query for the product code in a json. 

### Part of
- #5684

### DIFF
--- a/tests/integration/delete_product.t
+++ b/tests/integration/delete_product.t
@@ -1,0 +1,62 @@
+#!/usr/bin/perl -w
+
+use ProductOpener::PerlStandards;
+
+use Test::More;
+use ProductOpener::APITest qw/:all/;
+use ProductOpener::Test qw/:all/;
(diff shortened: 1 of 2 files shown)


stephanegigandet (commented):
So the goal of the test is to verify that remove_all_products() removes products? In that case maybe the test could be renamed to remove_all_products.t, because initially I thought it tested the API to remove a product (which exists but is enabled only for moderators: it would be a good thing to test)

yuktea:
> So the goal of the test is to verify that remove_all_products() removes products? In that case maybe the test could be renamed to remove_all_products.t, because initially I thought it tested the API to remove a product (which exists but is enabled only for moderators: it would be a good thing to test)

I agree, unless we test the api, this looks like a pretty useless test

alexgarel:
Closing for now, as delete is not available in the API.

PR #7239, opened 2022-08-23
[DO NOT MERGE YET] Connect to Redis

simonj2:
### What
This PR connects the Server to the search service so that new writes are populated. See the search service here: https://github.com/openfoodfacts/openfoodfacts-search

The writes are contained in an `eval` statement, so there shouldn't be any effect if the Redis instance is down (such as now). However, there is no need to merge this until the search service is deployed. 

The search Redis host is located via an env config, just like Robotoff. 

This has been tested locally.


alexgarel:
@simonj2 really good. I put it on "draft" state to be clear it's not yet ready.

simonj2:
@stephanegigandet @teolemon - this is ready for review now :)

lib/ProductOpener/Products.pm
```diff
@@ -2414,6 +2419,23 @@ sub index_product($product_ref)
 	return;
 }
 
+sub index_search_service($product_ref) {
```
alexgarel, line 2422: The name is not very explicit to me.

I would name it `push_to_search_service` or something like that.

lib/ProductOpener/Products.pm
```diff
@@ -2414,6 +2419,23 @@ sub index_product($product_ref)
...
+	if ($ProductOpener::Config2::redis_url eq "")
+	{
+		$log->warn("Redis URL not provided for search indexing", { error => $@ }) if $log->is_warn();
+		return;
+	}
+	# Now send the barcode to Redis so that it can be indexed
+	eval {
+		my $redis_client = Redis::Client->new(host => $ProductOpener::Config2::redis_url);
```
alexgarel, line 2430: Shouldn't we reuse the redis connection ?

I would rather (even if it's short) put that in a specific file: Redis.pm and have a global for $redis_client.
Then use get_redis_client() to initialize it, if it's not already initialized.

This sub could also be in Redis.pm, for the moment.

lib/ProductOpener/Products.pm
```diff
@@ -2414,6 +2419,23 @@ sub index_product($product_ref)
 	return;
 }
 
+sub index_search_service($product_ref) {
+	if ($ProductOpener::Config2::redis_url eq "")
+	{
+		$log->warn("Redis URL not provided for search indexing", { error => $@ }) if $log->is_warn();
```
alexgarel, line 2425: We should only warn about this once, eg at startup (at least not on every request).
(easier to do if we keep the connection, as proposed in another comment)

alexgarel (changes requested):
@simonj2 thanks a lot for this PR, nice work.

I would just prefer we keep connection, and move it to a specific file (but you may argue against it :-) )

lib/ProductOpener/Redis.pm
```diff
@@ -0,0 +1,43 @@
+use Redis::Client;
+
+BEGIN
+{
+	use vars       qw(@ISA @EXPORT_OK %EXPORT_TAGS);
+	@EXPORT_OK = qw(
+		&push_to_search_service
```
oalders, line 7: I think the ampersand can safely be omitted here. See https://metacpan.org/pod/Exporter#How-to-Export

> If you are only exporting function names it is recommended to omit the ampersand, as the implementation is faster this way.

lib/ProductOpener/Redis.pm
```diff
@@ -0,0 +1,43 @@
...
+
+		);    # symbols to export on request
+	%EXPORT_TAGS = (all => [@EXPORT_OK]);
+}
```
oalders, line 11: Please excuse the drive-by review. I was just doing some hacktoberfest browsing. 😄 
I'm not too familiar with this project, but I think the beginning of the package can be shortened to something like:

```perl
package ProductOpener::Redis;

use strict;
use warnings;

use Exporter qw( import );
use Redis::Client ();

our @EXPORT_OK = qw( push_to_search_service );
our %EXPORT_TAGS = ( all => [@EXPORT_OK] );
```

alexgarel: @oalders advice are always welcome ! But for now it's better to be consistent with the rest of the code. If you could write an issue to describe how we should do instead, so that we apply it to all modules at once, it would be great :-) We are kin to modernize our code base one step at a time.

alexgarel, line 11: I think you should first declare your package and declare you use Export and ProductOpener::PerlStandards

@stephanegigandet knows better than me on this kind of matter :-D

```suggestion
use ProductOpener::PerlStandards;
use Exporter qw< import >;

BEGIN
{
	use vars       qw(@ISA @EXPORT_OK %EXPORT_TAGS);
	@EXPORT_OK = qw(
		&push_to_search_service
		);    # symbols to export on request
	%EXPORT_TAGS = (all => [@EXPORT_OK]);
}

use Redis::Client;
```

oalders: @alexgarel sure. I'll put something together and people can decide if there's anything there worth exploring. I had seen a talk on your site given at The Perl Conference last year, so I had been meaning to have a closer look.

simonj2:
Updating to draft since I'm having some trouble getting this to work, and want to use this as a platform to ask for some 👀

lib/ProductOpener/Redis.pm
```diff
@@ -0,0 +1,50 @@
...
+		return $redis_client;
+	
+	};
+	if ($@) {
+		$log->warn("Error connecting to Redis", { error => $@ }) if $log->is_warn();
+	}
+	return undef;
+}
+
+my $redis_client = init_redis();
```
simonj2, line 40: @alexgarel @oalders - Hey there, since I restructured Redis to be a standalone file, this is no longer working. The problem is that the `$redis_client` variable is not getting set, and then isn't visible inside `push_to_search_service`. I am relearning Perl, so the solution is likely obvious.

From Googling I've tried a few things like:
- Using `our`, but this results in other unrelated errors, and I don't see it used in our codebase
- Putting in `use vars`, `use names`, etc.

Curious what I'm missing, and what the right way to do this is :)

simonj2: Also cc @stephanegigandet incase you have any idea

stephanegigandet: Hi @simonj2 , this was difficult to debug, but I think the issue is the "return $redis_client;" which is inside the eval block: since it's in the eval, it's not returned to the caller of the init_redis() function. So you could do something like this:

```
sub init_redis() {

	$log->debug("init_redis", { redis_url_env => $ENV{REDIS_URL}, redis_url => $ProductOpener::Config2::redis_url } ) if $log->is_debug();
	
	$log->warn("REDIS_URL env", {redis_url => $ENV{REDIS_URL} }) if $log->is_warn();
	if ($ProductOpener::Config2::redis_url eq "")
	{
		$log->warn("Redis URL not provided for search indexing", { error => $@ }) if $log->is_warn();
		return undef;
	}
	my $redis_client;
	eval {
		$redis_client = Redis::Client->new(host => $ProductOpener::Config2::redis_url);
	};
	if ($@) {
		$log->warn("Error connecting to Redis", { error => $@ }) if $log->is_warn();
	}
	else {
		return $redis_client;
	}
	return undef;
}
```

But since you have my $redis_client = init_redis() after, I woud just declare "my $redis_client;" at the top of  the file (before init_redis()), and init_redis() could just set it directly, without returning it.

lib/ProductOpener/Redis.pm
```diff
@@ -0,0 +1,50 @@
...
+
+sub push_to_search_service($product_ref) {
+	if (defined($redis_client)) {
+                $redis_client->rpush('search_import_queue', $product_ref->{code});
```
stephanegigandet, line 44: I think we need an eval block here, as otherwise we get an error when redis is not reachable.

Software error:
Can't connect to Redis host at searchredis:6379:6379: IO::Socket::INET: Bad hostname 'searchredis:6379' at /opt/perl/local/lib/perl5/Redis/Client/Role/URP.pm line 21.
For help, please send mail to the webmaster ([contact@openfoodfacts.localhost](mailto:contact@openfoodfacts.localhost)), giving this error message and the time and date of the error.

lib/ProductOpener/Redis.pm
simonj2: Oh amazing, thank you. I'll give this a shot. Appreciate your help!

simonj2: This worked - thank you!

lib/ProductOpener/Redis.pm
simonj2: Sure, done!

simonj2:
This is ready for review again and passing all manual testing.

alexgarel (approved):
LGTM

alexgarel:
Thanks a lot @simonj2 :-)

stephanegigandet (approved):
Looks good, thank you!

alexgarel:
I will merge, but for it to be effective we will have to change deployment CI to add redis parameter, but I'll do it on a separate PR.

PR #7240, opened 2022-08-23
refactor: Reduce nesting processing method

dipietroR:
Removing the most deeply nested part of Ingredients.pm and putting it in a separate function. 

I did this in two commits. The first one will work by itself. Though I felt like it might be reasonable to do the same on the rest of the original function. $debug_ingredients looks like a flag for enabling or disabling debug messages in this function, it seems a little awkward to use, as it requires modifying the code directly. I'm not use to seeing this kind of logic elsewhere in the codebase, so I'm not sure if it is an old way of managing debug messages and can clean up, or debug messages for this function are intentionally suppressed. How should I proceed with this?

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -1304,8 +1304,6 @@ Nested structure of ingredients and sub-ingredients
 
 sub parse_ingredients_text($product_ref) {
 
-	my $debug_ingredients = 0;
-
 	delete $product_ref->{ingredients};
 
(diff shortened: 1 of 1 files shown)


Ban3:
> or debug messages for this function are intentionally suppressed.

Probably this (@stephanegigandet knows for sure). [Log categories](https://metacpan.org/pod/Log::Log4perl#Turn-off-a-component) would be a better solution than hand editable variables.

teolemon:
Could you solve the merge conflict @dipietroR ?

alexgarel:
I think we can close…

PR #7246, opened 2022-08-24
refactor: reducing nesting GS1.pm

dipietroR:
Reverting a few if statements to reduce nesting.


dipietroR:
This PR won't work with the if statements reverted, I'm going to just put it in a separate function instead.

dipietroR:
I'm going to close this PR and open a new one since this doesn't work.

PR #7251, opened 2022-08-25
refactor: reduce nesting `GS1.pm`

dipietroR:
Turning the most deeply nested part of GS1.pm (`function gs1_to_off`) to a separate function (`nutrient_per_quantity`)

Suggestions for the new function name is welcomed :-) I wasn't exactly sure what it should be called.

### DIFF
--- a/lib/ProductOpener/GS1.pm
+++ b/lib/ProductOpener/GS1.pm
@@ -852,7 +852,7 @@ sub assign_field($results_ref, $target_field, $target_value) {
 }
 
 
-sub nutrient_per_quantity ($type, $per, $results_ref, $nid=undef, $nutrient_detail_ref=undef) {
+sub extract_nutrient_quantity_contained($type, $per, $results_ref, $nid, $nutrient_detail_ref ) {
 
 	my $nutrient_field = $nid . $type . "_" . $per;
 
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/GS1.pm
```diff
@@ -852,6 +852,60 @@ sub assign_field($results_ref, $target_field, $target_value) {
 }
 
 
+sub nutrient_per_quantity ($type, $per, $results_ref, $nid=undef, $nutrient_detail_ref=undef) {
```
stephanegigandet, line 855: We should never call the function without arguments for $nid and $nutrient_detail_ref, so we should not set default values for them.

lib/ProductOpener/GS1.pm
```diff
@@ -852,6 +852,60 @@ sub assign_field($results_ref, $target_field, $target_value) {
...
+	# quantityContained may be a single hash, or an array of hashes
+	# e.g. for the energy ENER- field, there are values in kJ and kcal that can be specified in different ways:
+	# - Equadis has 2 ENER- nutrientDetail, each with a single quantityContained hash
+	# - Agena3000 has 1 ENER- nutrientDetail with an array of 2 quantityContained
+	# --> convert a single hash to an array with a hash
+	if ((defined $nutrient_detail_ref->{quantityContained}) and (ref($nutrient_detail_ref->{quantityContained}) ne "ARRAY")) {
+		$nutrient_detail_ref->{quantityContained} = [$nutrient_detail_ref->{quantityContained}];
+	}
+
+	foreach my $quantity_contained_ref (@{$nutrient_detail_ref->{quantityContained}}) {
```
stephanegigandet, line 871: For the name of the function, as it is processing the quantityContained field for a nutrient, I would suggest extract_nutrient_quantity_contained()

dipietroR: Alright thank you!

stephanegigandet (approved):
Looks good, thank you!

PR #7258, opened 2022-08-26
fix: send set-cookie headers in redirects

stephanegigandet:
the session cookie was not sent in 302 redirects

This is needed in particular for Kratos auth: https://github.com/openfoodfacts/openfoodfacts-server/pull/7071


alexgarel (approved):
LGTM

PR #7268, opened 2022-08-29
taxonomy: Adding Dutch translations for taxonomies

Annieieieie:
### What

Added Dutch taxonomies

- Fixes #[ISSUE NUMBER]


taxonomies/categories.txt
```diff
@@ -68,6 +68,7 @@ fr:Gressins artisanaux
 hu:Kézzel készített termékek
 it:Grissini artigianali, Grissini fatti a mano
 nl:Handgemaakte broodstengels
+nl_be:Handgemaakte broodstengels
```
stephanegigandet, line 71: Currently we do not support language variants, so only nl: is used, nl_be: will not have an effect (it's ok to add it though).

aleene:
I see you have issues "conflicting files". Usually this happens when other persons have been working on the same files. So it depends when you downloaded your versions. I'll have a look.
Start the title of your change with` chore:` to suppress one error.
For the language versions, I started to add those some x years ago in the hope it would be supported. I now only add an extra line if the flemish variant is different, and I add the flemish variant as synonym of the dutch variant (unless it is a really flemish product (waterzooi?).

aleene:
Only the primary files need to have translations added. So ending in `.txt` without something before like `wip`, `result`, etc.

aleene:
in `en:Sangria` you added a dutch variant. As `sangria` is the same in many languages, a line `xx:Sangria` is added to indicate this. So if `nl: `is not found the line` xx: `is used. Very useful to limit lines that are the same.

aleene:
For category entries you noticed already an issue: plural or singular? I tend to use plural, so you can read an entry as: "Dit product behoort in het vak Spaanse likeuren". But sometimes this does not work well. In any case I also add the singular form as synonym.

aleene:
To check the conflicts do `update from main` (under Branch). Github will show you the files with issues.

In the file you will find something like:
```
nl:Vruchtensappen, Fruitsappen
<<<<<<< HEAD
nl_be:Vruchtensappen, Fruitsappen
=======
>>>>>>> main
pl:Soki owocowe
```
Under HEAD is your change and above main what is now the file. In this case there is real issue: so edit out the line with HEAD and everything below your change (=== and >>> main). If there is an issue you can merge the new and old versions and remove the <<<,, === and >>> lines.

aleene:
Thanks for your work!!!

taxonomies/categories.txt
Annieieieie: > Currently we do not support language variants, so only nl: is used, nl_be: will not have an effect (it's ok to add it though).

Ohh, I didn't know! Thanks for letting me know!

Annieieieie:
> Only the primary files need to have translations added. So ending in `.txt` without something before like `wip`, `result`, etc.

Okay, I will remember!

Annieieieie:
> Thanks for your work!!!

Thanks for the feedback, just did some taxonomies, but I didn't know how to save them!

stephanegigandet:
@Annieieieie @aleene : I don't speak Dutch but this one looks good to me, can we merge it? Or are other changes needed?

aleene:
Ok with me

stephanegigandet (approved):
Thank you!

Issue #7270, opened 2022-08-29
Replace all instances of param() by scalar param()


stephanegigandet:
### Describe the bug

The param() function in CGI.pm has a weird behaviour : when called in list context, it returns an array of values. But that means that if no value is provided for the parameter, it returns an empty array. That causes problems when param("something") is passed itself as a parameter for a function.

See https://metacpan.org/pod/CGI#Fetching-the-value-or-values-of-a-single-named-parameter

This caused issues in this PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/7212

### To Reproduce

And we currently have a bug for image upload because of this:

my $previous_code = normalize_code(param('previous_code'));

If "previous_code" is not passed, this is like calling normalize_code() with an empty array, which makes the prototype of normalize_code() fail : it expects 1 and only 1 parameter.

### Expected behavior

Replace all instances of param("something") by scalar param("something").

### Type of device

Browser

stephanegigandet:
We have a lot of instances of code like:  decode utf8=>param("something")
for reading text parameters. The decode utf8=> will also force the scalar context.

Ban3:
Would writing our own wrapper for `scalar param($s)` make sense? CGI already provides `multi_param` for always getting a list. Something like:

```perl
sub single_param($s) {
    scalar param($s)
}
```
Then you wouldn't need to remember the `scalar` every time and the intention would be clear.

stephanegigandet:
@Ban3 that's a good idea, I will do that.

alexgarel:
fixed by pr.

PR #7273, opened 2022-08-30
refactor: specifically ask parameters as scalars

stephanegigandet:
See discussion in https://github.com/openfoodfacts/openfoodfacts-server/issues/7270

I added scalar before param() calls in many places where there was already a scalar context, for instance:

my $field_value = param("field");  -> my $field_value = scalar param("field");

This is so that we keep the scalar if we add a wrapping function (e.g. my $field_value = normalization_function(scalar param("field"));

Otherwise we may get bitten by the signature bug again if the parameter is not passed and we get an empty list from param() in a list context.

### DIFF
--- a/cgi/change_password.pl
+++ b/cgi/change_password.pl
@@ -70,11 +70,11 @@ if ($ENV{'REQUEST_METHOD'} eq 'POST') {
 		push @errors, lang('error_bad_login_password');
 	}
 
-	if (length(param('password')) < 6) {
+	if (length(scalar param('password')) < 6) {
 		push @errors, lang('error_invalid_password');
 	}
 
-	if (param('password') ne param('confirm_password')) {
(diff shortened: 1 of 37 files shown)


stephanegigandet:
Refactored using @Ban3 's suggestion: https://github.com/openfoodfacts/openfoodfacts-server/issues/7270#issuecomment-1230423857

alexgarel (changes requested):
I still see some. 
for example:
- change_password.pl line 62 and 85
- export_products.pl line 85 and line 150
- import_file_upload.pl line 68 and 69
- ...

Most of them have "decode utf8 =>" in front

I find them with `\bparam\s*\(` regexp (there are some false positive)

alexgarel (approved):
LGTM

PR #7291, opened 2022-09-02
docs: warn about strings with an impact on tests

teolemon:
doc: warn about strings with an impact on tests


stephanegigandet (approved):
I believe Crowdin will display that comment to translators, maybe we could provide a bit more context, it may be unclear to translators if they should worry about this or not.

PR #7294, opened 2022-09-05
refactor: show more info in test logs

stephanegigandet:
@yuktea : I added error messages in the test so that if something fails, we can see in the test logs what failed exactly.

Here the test shows that we can't decode the JSON response of the API because in fact we did not get a response.

You could do something similar when you call get() in order to catch errors.

See: https://metacpan.org/pod/LWP::UserAgent#get
Instead of calling get() directly, you can call 
my $res = $ua->get( $url );

That way you can check the status code of the response etc. in the test, and print it in the test log if it fails.

### DIFF
--- a/tests/integration/search_test.t
+++ b/tests/integration/search_test.t
@@ -125,7 +125,7 @@ my @tests = (
 	[
 		"q3",
 		construct_test_url(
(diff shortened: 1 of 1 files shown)


yuktea:
Thank you @stephanegigandet, I'll use this pattern now

PR #7299, opened 2022-09-05
taxonomy: adding croatian translation for ingredient of choco delicia

benbenben2:
added croatian translation of ingredients of choco product from delicia brand


stephanegigandet (approved):
Thank you!

PR #7302, opened 2022-09-06
fix: put back symlinks for cgi/product.pl

stephanegigandet:
I made a mistake while changing param() to single_param(). I used a sed command which replaced the symlinks by copies of the files... (in this commit: https://github.com/openfoodfacts/openfoodfacts-server/commit/e04d343941064f8a7c8f056b9914a27644672be1 )

This PR puts back symlinks.


alexgarel:
@stephanegigandet you have perltidy issues, a `make lint_perltidy` should fix that

alexgarel (approved):
LGTM, but you have to fix you perltidy errors :-) `make perltidy_lint`

Issue #6259, opened 2022-01-07
Send edit events to openfoodfacts-events


teolemon:
### Who for
- Contributors
### What

- [ ] Integrate [openfoodfacts-events](https://github.com/openfoodfacts/openfoodfacts-events) with product opener
  - [ ] Send 1 event as an example

### Why
- This will enable badges, gamification, time-based analysis

### Codepointer
- Robotoff pinging code (similar system): https://github.com/openfoodfacts/openfoodfacts-server/blob/6e21afb1d89618185e1bed8730c2a8189b95fea5/scripts/run_cloud_vision_ocr.pl

### Part of
- #5516

alexgarel:
Start with:
* product edit
* image upload 
* if possible: insight validation (product edit coming from robotoff)

PR #7303, opened 2022-09-06
feat: send events when editing products

stephanegigandet:
Some code to send events to  https://github.com/openfoodfacts/openfoodfacts-events : https://github.com/openfoodfacts/openfoodfacts-server/issues/6259


stephanegigandet:
Currently not working in a dev environment, as the EVENTS_URL environment variable seems not passed down, even after a make dev. Probably a bug somewhere.

www-data@d7be0fbd4116:/opt/product-opener/lib/ProductOpener$ env | grep EVEN
EVENTS_URL=https://events.openfoodfacts.org

[6242] /opt/product-opener/lib/ProductOpener/Events.pm 101 ProductOpener.Events {action => 'process',events_url => undef,request => 'lYrg7I3BWPFipE38',type => 'edit'} send_event EVENTS_URL not defined

stephanegigandet:
This is now ready for review.

There is a slight discrepancy as the events api except an integer for the barcode, while we use strings in product opener: https://github.com/openfoodfacts/openfoodfacts-events/issues/137

lib/ProductOpener/Events.pm
```diff
@@ -0,0 +1,143 @@
...
+
+Barcode of the product.
+
+=head4 user id - required
+
+=cut
+
+sub send_event ($event_ref) {
+
+	if ((defined $events_url) and ($events_url ne "") > 0) {
```
alexgarel, line 86: style: why do you have to put > 0 ?

lib/ProductOpener/Events.pm
```diff
@@ -0,0 +1,143 @@
...
+
+=head4 barcode - required - string
+
+Barcode of the product.
+
+=head4 user id - required
+
+=cut
+
+sub send_event ($event_ref) {
```
alexgarel, line 84: digression: finally passing arguments as hash we loose the benefits of signatures ... that's a pity.
I would love if perl had a way to "unroll" the hash to use it to call a function with parameter (what we can do with "**" in python), so we could get best on both side.

lib/ProductOpener/Events.pm
```diff
@@ -0,0 +1,143 @@
...
+				"send_event response not ok",
+				{
+					endpoint => $endpoint,
+					event => $event_ref,
+					is_success => $response->is_success,
+					code => $response->code,
+					status_line => $response->status_line,
+					response => $response
+				}
+			) if $log->is_debug();
```
alexgarel, line 133: I think we should at least log a warning for this case.

alexgarel (approved):
LGTM but a log level to change.

teolemon:
FYI @ocervell  :-)

lib/ProductOpener/Events.pm
stephanegigandet: ah, my mistake, I replaced a length($events_url) > 0 by $events_url ne "". The length test is supposed to be faster so we have it in a few places in our code, but I find it more difficult to quickly understand what the intent is (checking that the value is not empty).

cgi/product_multilingual.pl
```diff
@@ -1500,6 +1501,9 @@ ($product_ref, $field, $language)
 		# Notify robotoff
 		send_notification_for_product_change($product_ref, "updated");
 
+		# Create an event
+		send_event( { user_id => $User_id, event_type => "product_edited", barcode => $code});
```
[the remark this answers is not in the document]
stephanegigandet: I find it a bit akward to specify points in the client that submit events. But it has the advantage of allowing us to change the points awarded for each event type without the change having a retroactive effect. It could also be used to have special operations like "double points on candies during Halloween!".

lib/ProductOpener/Events.pm
stephanegigandet: fixed

cgi/product_multilingual.pl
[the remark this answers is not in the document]
stephanegigandet: I put 5. At this point it's only a first test, we can think of a better structure to define points awarded (maybe with rules like if the product has specific traits, then it's worth more points).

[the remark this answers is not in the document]
alexgarel: yep, we could use the quality facet changed before / after edit, to give less or more points.
But I propose 5 in the sense that it at least reward more product edition than robotoff question answering. Fine tuning is for later :-) (and might require a game design expert :-P )

alexgarel (approved):
Cool !

PR #7321, opened 2022-09-10
fix: Label parenting

aleene:
### What
Repaired some wrong parenting.

Should these not have been intercepted by a test?


stephanegigandet:
> Repaired some wrong parenting.
> 
> Should these not have been intercepted by a test?

For parenting, the taxonomy building still works if you use a synonym or a translated name for the parent.

taxonomies/labels.txt
```diff
@@ -798,7 +798,7 @@ bg:Без пшеница
 de:Ohne Weizen
 nl:Zonder graan
 
-en:No gluten, Gluten-free, Without gluten, Gluten-free-food, free of gluten, Naturally gluten free, Gluten free, ni gluten
```
stephanegigandet, line 801: For allergens, antibiotics etc. the ingredients parsing code expects the "No something" to be the canonical entry.

There's a comment about that at the top of the file:

# For labels that indicate the absence of something, in English, always make "No something" the first entry, and make "Something-free", "Without something" synonyms

stephanegigandet: Maybe we can add that it's required by the ingredients parsing code.

alexgarel:
@stephanegigandet what is the status on this PR ?

aleene:
I think I have to revert some changes and adapt the children instead.

Issue #7288, opened 2022-09-02
Change status code to 403 for API and web pages requested with a bad userid / password combination


stephanegigandet:
### Problem description

Apps can include user credentials in API requests (userid + password field). But if the credentials are incorrect, the server returns a 200 response with HTML code saying the credentials are incorrect. Apps expect a JSON response, try to decode the HTML as JSON, and fail.

### Desired solution

When user credentials are incorrect, return a 403 so that apps know the credentials are incorrect, and that they should not try to decode the body:

https://github.com/openfoodfacts/smooth-app/issues/2793#issuecomment-1234184480

PR #7331, opened 2022-09-12
fix: prevent default mod_perl error message to be appended

### LINKED ISSUE #7330: mod_perl is appending "Not found" error messages to Product Opener 404 error messages
### Describe the bug

When Product Opener returns a 404, mod_perl appends an error message at the end of the HTML body.

image

### To Reproduce

https://world.openfoodfacts.org/urlthatdoesnotexist

### Expected behavior

No extra error message.

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Browser

### Browser version

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
Fixes #7330 .

Also fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/7288 which should help https://github.com/openfoodfacts/smooth-app/issues/2793#issuecomment-1234184480

I renamed display_error() to display_error_and_exit() to make it more clear what the function actually does.

I tried a bunch of different ways to make Apache / mod_perl send back a non 200 status code while not appending a default error body, and the only solution that seems to work is to output the 404 status code and tell Apache it is a 200.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -63,6 +63,7 @@ BEGIN
 		&display_tag
 		&display_search_results
 		&display_error
+		&display_error_and_exit
 
 		&add_product_nutriment_to_stats
 		&compute_stats_for_products
@@ -199,7 +200,7 @@ our $mongodb_log = Log::Log4perl->get_logger('mongodb');
 $mongodb_log->info("start") if $mongodb_log->is_info();
 
 use Apache2::RequestRec ();
-use Apache2::Const ();
+use Apache2::Const qw(:http :common);
 
 use URI::Find;
 
@@ -691,9 +692,7 @@ sub init_request() {
 
 	my $error = ProductOpener::Users::init_user($request_ref);
 	if ($error) {
-		if (not single_param('jqm')) { # API
-			display_error($error, undef);
-		}
+		display_error_and_exit($error, 403);
 	}
 
 	# %admin is defined in Config.pm
@@ -997,7 +996,8 @@ sub analyze_request($request_ref) {
 			$request_ref->{redirect} = $formatted_subdomain . '/' . $tag_type_singular{products}{$lc} . '/' . $components[1];;
 		}
 		else {
-			display_error(lang("error_invalid_address"), 404);
+			$request_ref->{error_status} = 404;
+			$request_ref->{error_message} = lang("error_invalid_address");
 		}
 	}
 
@@ -1017,7 +1017,8 @@ sub analyze_request($request_ref) {
 			}
 		}
 		else {
-			display_error(lang("error_invalid_address"), 404);
+			$request_ref->{error_status} = 404;
+			$request_ref->{error_message} = lang("error_invalid_address");
 		}
 	}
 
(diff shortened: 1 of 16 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -1325,9 +1336,25 @@ sub display_error($error_message, $status) {
...
+Display an error message using the site template, and terminate the request immediately.
+
+Any code after the call to display_error_and_exit() will not be executed.
+
+=cut
+
+sub display_error_and_exit($error_message, $status) {
+
+	display_error($error_message, $status);
 	exit();
```
alexgarel, line 1354: Maybe we could set whatever status if we do `exit(Apache2::Const::OK)` ? (or is it the default already ?)

See https://perl.apache.org/docs/2.0/api/ModPerl/Util.html#C_exit_

lib/ProductOpener/Display.pm
```diff
@@ -1316,6 +1318,15 @@ sub display_date_iso($t) {
...
+
+=head2 display_error ( $error_message, $status )
+
+Display an error message using the site template.
+
+The request is not terminated by this function, it will continue to run.
+
+=cut
+
 sub display_error($error_message, $status) {
```
alexgarel, line 1330: we should take care of the context.
An API request should return a json instead.
I would use a parameter for that, but also have a function is_api_request($request_ref) to determine that.

But maybe this is a separate PR.

lib/ProductOpener/Display.pm
```diff
@@ -7133,14 +7163,13 @@ HTML
...
-		$r->rflush;
-		$r->status($status);
-	}
+
+	my $r = Apache2::RequestUtil->request();
+	$r->rflush;
+	# Setting the status makes mod_perl append a default error to the body
+	# $r->status($status);
+	# Send 200 instead.
+	$r->status(200);
```
alexgarel, line 7172: For me we should return Apache::Const::OK and propagate this response.

Or maybe in case of $status != 200, we should just use exit() but keep status in request.

alexgarel (commented):
Some point of discussion.

Otherwise it's ok to me globally.

lib/ProductOpener/Display.pm
alexgarel: sorry, it seems return code is ignored :-D

lib/ProductOpener/Display.pm
stephanegigandet: I tried returning Apache::Const::OK but it does not work, Apache still adds a default error body if $r's status is set to anything else than 200.

To test it, I wrote a simple test:

```

use CGI::Carp qw(fatalsToBrowser);

use Apache2::Const -compile => qw(OK);
use CGI qw/:cgi :form escapeHTML/;


my $status = param("status") || "200";

print header ( -status => $status );

my $body = "Status is set to $status";

binmode(STDOUT, ":encoding(UTF-8)");

print $body;

my $r = Apache2::RequestUtil->request();
$r->rflush;
$r->status($status);

return Apache2::Const::OK;
```

If I set the status to anything than 200, I get the default message added:

image

lib/ProductOpener/Display.pm
stephanegigandet: > An API request should return a json instead.

Agreed, but let's make it in another PR, as there may be side effects, and we need to discuss what the right format of that JSON is with the API users.

alexgarel: agree, opened #7365

lib/ProductOpener/Display.pm
alexgarel: ok

alexgarel (approved):
Let's go !

PR #7337, opened 2022-09-13
taxonomy: added some ingredients for madarica from pan-pek in HR

benbenben2:
taxonomy: added some ingredients for madarica from pan-pek in HR

### DIFF
--- a/taxonomies/additives.txt
+++ b/taxonomies/additives.txt
@@ -3645,6 +3645,7 @@ fr:E200, Acide sorbique, C6H8O2, 110-44-1
 ga:E200, Aigéad sorbach
 he:E200, חומצה סורבית
 hi:E200, सोर्बिक अम्ल
+hr:E200, konzervans E200
 hu:E200, Szorbinsav
 id:E200, Asam sorbat
 is:E200, sorbic acid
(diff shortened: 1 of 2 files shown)


taxonomies/additives.txt
```diff
@@ -3645,6 +3645,7 @@ fr:E200, Acide sorbique, C6H8O2, 110-44-1
 ga:E200, Aigéad sorbach
 he:E200, חומצה סורבית
 hi:E200, सोर्बिक अम्ल
+hr:E200, konzervans E200
```
stephanegigandet, line 3648: For additives, we should not add additive classes (e.g. "colouring") in the synonyms. Instead we need to add them to the additives classes taxonomy: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/taxonomies/additives_classes.txt

That way, "konservans E-something" will match "E-something", and we will know it's a preservative.

And the second name should be the name of the additive, so something like this:

hr:E200, Sorbinska kiselina

benbenben2: @stephanegigandet thank you for the comment, I'm still learning :-)

I took this comment into account and made a new commit (my IDE added some blanks at the end of some lines. Sorry about that. I think it can be ignored)

stephanegigandet (approved):
Perfect, thank you very much!

PR #7344, opened 2022-09-15
taxonomy: create Pâtisson (squash)

OpenChris:
• Create **Pâtisson** (squash) with Agribalyse proxy on Courge melonnette
• Create **Echine de porc** (Agribalyse 28302)
• Link **Aiguillettes de canard** to Duck breasts (vs Duck today) to give a better Agribalyse reference (Aiguillettes are a part of the Magret)
• Create **Batavia** type of salad and use Iceberg as an Abribalyse proxy (Iceberg is from the _laitues pommées batavias_ family)

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -89294,6 +89294,14 @@ ciqual_food_name:en:Squash, melonnette jaspée from Vendée, pulp, raw
 ciqual_food_name:fr:Courge melonnette, pulpe, crue
 
 <en:Pumpkins
+en:Pattypan squash
+fr:Pâtisson, Patisson
+de:Patisson
+wikidata:en:Q876892
+agribalyse_proxy_food_code:en:20136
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Thank you!

taxonomies/categories.txt
```diff
@@ -71372,6 +71372,12 @@ ciqual_food_code:en:28451
 ciqual_food_name:en:Pork, loin, roasted/baked
 ciqual_food_name:fr:Porc, échine, rôtie/cuite au four
 
+<en:Pork meat
+fr:échine de porc
```
stephanegigandet, line 71376: it would be great to have an English translation. But it seems Ciqual used the same name for both 28302 and 28003.

OpenChris: Yes, it seems that the English language doesn't have a distinction between Longe de porc and Echine de porc. They both fall into the "loin" category. Or they call the Echine "spare ribs", which should really be the Travers de porc.

PR #7367, opened 2022-09-19
feat: reverting changes from a specific user

stephanegigandet:
I had written a script to remove all changes from a specific user (used when something really bad happens), but it was not checked in.

In practice, this remove all edits done after an user edited a product (including changes from other users after that). So it can be used only shortly after a mistake has been spotted.


alexgarel (approved):
I let you review my changes before merging :-)
- Added comment
- userid is a parameter
- better pretend

scripts/revert_changes_from_user.pl
```diff
@@ -23,20 +23,19 @@
...
 This script will revert products to the most recent version before a given user changed the product.
 Products that did not exist before will be deleted.
 All changes done after the first edit of the user will also be deleted, even if done by other users.
 
+You may limit to a series of codes
+
 Usage:
 
-update_all_products.pl --pretend
+revert_changes_from_user.pl --userid user_id --codes codes --pretend
```
stephanegigandet, line 37: @alexgarel : the --codes parameter is not coded yet I believe?

alexgarel: Right, we should remove it ! (I saw it was too much time to implement it).

PR #7370, opened 2022-09-19
taxonomy: added varietal for croatians wines

benbenben2:
added varietal for croatians wines
reference: https://en.wikipedia.org/wiki/Category:Grape_varieties_of_Croatia


benbenben2:
Why did the Perl unit tests fail?

stephanegigandet (approved):
Thank you!

PR #7376, opened 2022-09-20
fix: small changes in product files sent by Carrefour

stephanegigandet:
- XML files are in UTF16 instead of UTF8
- *_valNut.xml files renamed to *_valnut.xml

Added some tests to check that we accept both the old and new formats / names.


alexgarel (approved):
LGTM

Issue #3158, opened 2020-03-29
Show Robotoff questions to non-logged users, with a prompt to sign-up or login


teolemon:
### What
Show Robotoff questions to non-logged users, with a prompt to sign-up or login.
Code pointer (original PR by @hangy) : https://github.com/openfoodfacts/openfoodfacts-server/pull/1705/files

### Part of
- #5708 

### Mockup
![image](https://user-images.githubusercontent.com/1689815/87070744-89cb6280-c219-11ea-9ae0-a101af7b521e.png)

teolemon:
@hangy @VaiTon @raphael0202 added a mockup of an improved web experience for Robotoff that would enable to convert part of the 1,4 million monthly visitors into contributors

teolemon:
display_robotoff(barcode):
on_button_click():
>>>>if is_logged_in:
>>>>>>>>current_behaviour(yes,barcode,user)
>>>>else:
>>>>>>>>login_or_signup_modal(yes, barcode) > redirect_to_the_product_page > current_behaviour(yes,barcode,user)

teolemon:
Relevant integration points: 
Javascript that displays Robotoff: https://github.com/openfoodfacts/openfoodfacts-server/blob/9af7d8767d3e2a84b055f96b9fa18dc3a45988e6/html/js/display-product.js#L33
User signup: https://github.com/openfoodfacts/openfoodfacts-server/blob/b30b6152c0d970de44ad1f2f1bfdf124489c45e3/cgi/user2.pl
Template for user signup: user_form2.tt.html

CharlesNepote:
The call to Robotoff questions is only available for registered users. Before rewrriting JS code we have to display it for every users:
https://github.com/openfoodfacts/openfoodfacts-server/blob/319b66b14a8717ddd3e8123eaaf6a30398d58d0c/templates/web/pages/product/product_page.tt.html#L50
```diff
- [% IF (user_id.defined) && (robotoff_url.defined) && (robotoff_url.length > 0) %]
+ [% IF (robotoff_url.defined) && (robotoff_url.length > 0) %]
```
Then test if user is connected or not. We can start with a simple check to the HTML id `logged_in_user_id` as we are doing in Power User Script.
```js
    /**
     * getConnectedUserID: returns user id of the current connected user
     *
     * @param    none
     * @returns  {String} user id; Example: "charlesnepote"
     */
    function getConnectedUserID() {
[...]

PR #7379, opened 2022-09-21
feat: robotoff anonymous questions

alexgarel:
### What

feat: Make it possible for anonymous users to see and answer robotoff questions
build: some changes to be compatible with a local robotoff instance

fixes: 
- #3158


templates/web/pages/product/product_page.tt.html
```diff
@@ -47,7 +47,7 @@
     <p>→ <a href="[% public_product_url %]">[% lang("product_page_on_the_public_database") %]</a></p>
 [% END %]
 
-[% IF (user_id.defined) && (robotoff_url.defined) && (robotoff_url.length > 0) %]
+[% IF (robotoff_url.defined) && (robotoff_url.length > 0) %]
```
alexgarel, line 50: yes it's that easy, because robotoff takes care of computing device_id using url.

raphael0202, line 50: Maybe it could it a good idea to add a comment in the code to explain this

stephanegigandet: so for a non logged in OFF website user, how is the device_id computed? is it a hash of the user ip address or something?

alexgarel: yes it's the sha1 of user ip, see https://github.com/openfoodfacts/robotoff/blob/e46942f0055b54183ef5570b7c39215ed312309c/robotoff/app/api.py#L235

alexgarel: done.

raphael0202 (approved):

PR #7382, opened 2022-09-21
taxonomy: fix Pork ham escalope

OpenChris:
• fix Pork ham escalope redundant categs (x3 😱) and differentiate cooked vs raw
• add proxy to Boudins blancs : copy from Boudins blancs with truffle


stephanegigandet (approved):
Thank you!

PR #7388, opened 2022-09-22
fix: Change the URL of the blog in the footer

teolemon:
### What
- fix: Change the URL of the blog in the footer
### Why
- The URL in the footer is still en.blog.openfoodfacts.org
- The SSL for en.blog.openfoodfacts.org has not been renewed https://github.com/openfoodfacts/openfoodfacts-infrastructure/issues/137
- We're loosing visitors due to the combination of those 2 bugs


alexgarel (approved):
LGTM

PR #7398, opened 2022-09-26
feat: improved parsing of origins of ingredients

stephanegigandet:
- refactored the construction of the list of regexps that we use to match entries of a taxonomy in a string. This is used in particular for ingredients parsing, and packaging parsing. There was a lot of code that I had copy/pasted and modified. The common code is now in Tags.pm.
- we now match things like "Origins of the ingredients: Spain, Italy and South America" or "Strawberries from France or Belgium".
- added tests to cover the new matches
- moved some tests out of /t and /test to /tests


alexgarel (approved):
LGTM

PR #7401, opened 2022-09-26
fix: changes needed for Lactalis Fromages import through Agena3000

stephanegigandet:
- solves an issue when servingSizeDescriptions has multiple values (for different languages)
- allow imports even if the partyName field is empty, if we have an organization with a matching GLN in OFF


alexgarel (changes requested):
LGTM but it seems you forgot to add some tests files ?

alexgarel (approved):
LGTM (didn't thoroughly dive into details, I trust the tests ;-) )

PR #7406, opened 2022-09-27
docs: start of documentation and schema for knowledge panels

stephanegigandet:
I started to create the JSON schemas for the knowledge_panels field returned by the product API.
e.g. https://world.openfoodfacts.org/api/v2/product/20047252/cornichons-en-tranches-freshona?fields=knowledge_panels

It is not complete yet, but your feedback would be very useful to check that I am going in the right direction.

image


alexgarel (approved):
Let's merge and improve later :-)

PR #7416, opened 2022-09-28
ci: Add a check that the OpenAPI doc is valid

teolemon:
### What
ci: Add a check that the OpenAPI doc is valid


alexgarel:
@teolemon maybe we should try https://github.com/stoplightio/spectral which is open-source and should already support  3.1

alexgarel:
I'll close this for now, I have to check on my side if swagger check can work, at least in the CI.

PR #7417, opened 2022-09-28
taxonomy: added ingredients for ice-cream in HR

benbenben2:
added some ingredients for an ice-cream (https://hr.openfoodfacts.org/product/3850116036113/quattro-joy-ledo), in HR


stephanegigandet (approved):
Thanks!

PR #7428, opened 2022-09-30
taxonomy: added some missing countries in HR

benbenben2:
Added missing countries in HR

FYI, I used wikidata, wikipedia and the Croatian encyclopedia (https://www.enciklopedija.hr/, from the Miroslav Krleža Lexicographic Institute) as sources.


stephanegigandet (approved):
Thank you!

PR #7441, opened 2022-10-04
taxonomy: Fi and de ingredients

moon-rabbitOFF:


aleene:
This should not be necessary as the entries are already covered by xx:

taxonomies/ingredients.txt
```diff
@@ -1153,6 +1153,7 @@ ar:عصية لبنية حمضية
 bg:Lactobacillus acidophilus, L. acidophilus
 xx:Lactobacillus acidophilus, L. acidophilus
 fa:لاکنوباسیلوس اسیدوفیلوس
+fi:Lactobacillus acidophilus, L. acidophilus
```
stephanegigandet, line 1156: In practice, the xx:Lactobacillus acidophilus, L. acidophilus means that the xx: translation will be used for all languages that don't have another translation, so we do not need to add it. It does not hurt though.

stephanegigandet (approved):

taxonomies/ingredients.txt
aleene: In fact a few iterations ago I removed all these.

PR #7451, opened 2022-10-06
refactor: tests refactor to use same pattern

alexgarel:
### What
Refactor tests to avoid code duplication for expected results pattern.


stephanegigandet (approved):
Looks good, thanks for the refactor

PR #7454, opened 2022-10-06
chore: Extruded crispbread

moon-rabbitOFF:
Differentiating swedish knäckebröd from extruded products.


stephanegigandet (approved):
Thanks!

Issue #7361, opened 2022-09-17
Allow to get the username by email


M123-dev:
### Problem description
In the mobile-app we display something like "Hello Teolemon". (Not as a one time dialog, at the top of the settings).

This doesn't work when a user uses their email to login.

### Desired solution

A simple endpoint to recieve the username.

### Alternatives you have considered

A structured answer from a login endpoint, but thats out of the scope of this what seems to me very simple issue.

### Additional context

[#1833](https://github.com/openfoodfacts/smooth-app/issues/1833)

### Number of products impacted

1, the official open food facts mobile app. 😄

PR #7455, opened 2022-10-07
feat: API to login and get user username, name and email 

stephanegigandet:
This is to address #7361 so that the Flutter app can get the username and name of an user who authenticated with an email address.

but it could also be a step to have a more uniform handling of errors in the API (specifically errors related to users not identified or bad logins and passwords supplied).

I'm proposing that we extend the /cgi/auth.pl API which currently:

- checks if an user is identified (either through a session cookie on the *.openfoodfacts.org domain, or by passing user_id and password)
- returns a 200 status code if the user is logged in, or a 403 otherwise

The extension would be to add a JSON body:
- when the user is logged in:

```json
{
    "status": 1,
    "status_verbose": "user signed-in",
    "user": {
        "email": "my email",
        "name": "Stéphane Gigandet"
    },
    "user_id": "test2"
}

```
- when the user is not logged in:
```json
{
    "status": 0,
    "status_verbose": "user not signed-in"
}
```

With the proposed code, this behaviour would only apply to /cgi/auth.pl

But we could decide to extend it to all other API calls (e.g. product edit). If invalid userid / password are supplied, we return a JSON body to explain why (instead of returning an HTML content today).

I think we could keep the existing status: 0 that we have in most APIs I think (get and search product, product edit, auth), and maybe add an "error" id + and "error_message" in the language requested.

cc @VaiTon @M123-dev @monsieurtanuki @g123k

### DIFF
--- a/cgi/auth.pl
+++ b/cgi/auth.pl
@@ -40,13 +40,35 @@ $log->info('start') if $log->is_info();
 
 my $request_ref = ProductOpener::Display::init_request();
 
-my $status = 403;
+my $status;
+my $response_ref;
 
 if (defined $User_id) {
 	$status = 200;
+
+	# Return basic data about the user
+	$response_ref = {
+		status => 1,
+		status_verbose => 'user signed-in',
+		user_id => $User_id,
+		user => {
+			email => $User{email},
+			name => $User{name},
+		},
+	};
+
+	use JSON::PP;
+
+}
+else {
+	$status = 403;
+	$response_ref = {
+		status => 0,
+		status_verbose => 'user not signed-in',
+	};
 }
 
-print header(-status => $status);
+my $json = JSON::PP->new->allow_nonref->canonical->utf8->encode($response_ref);
 
 # We need to send the header Access-Control-Allow-Credentials=true so that websites
 # such has hunger.openfoodfacts.org that send a query to world.openfoodfacts.org/cgi/auth.pl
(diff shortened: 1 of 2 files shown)


M123-dev:
I love the response body, as long as the Status Codes don't change and so break the older versions I really like this feature.

Same goes for the other api calls. It will be a relieve if we can just foreward a consistent error to the sdk user instead of manually parsing the html and extracting errors out of it. Especially if the messages are translated.

alexgarel:
@stephanegigandet would you add it to the api.yml file ?

cgi/auth.pl
```diff
@@ -63,5 +85,12 @@
...
+print header(-status => $status, -type => 'application/json', -charset => 'utf-8');
+print $json;
+
 $r->rflush;
-$r->status($status);
+
+# Setting the status makes mod_perl append a default error to the body
+# $r->status($status);
+# Send 200 instead.
+$r->status(200);
```
alexgarel, line 96: Hum ain't it too big a change for third party app ? (javascript app etc.).

Maybe we should return json only if the user add a json=1 parameter ? (and in this case return a 200).

stephanegigandet: The only apps that use /cgi/auth.pl are Folksonomy Engine and Hunger Games I believe (it's not documented anywhere). And I think they only look at the status code, as we were only returning a status code, with an empty body.

stephanegigandet: Note that this $r->status(200) actually does not change the true return status that is sent to the client, which is set in the header above.

image

stephanegigandet:
> @stephanegigandet would you add it to the api.yml file ?

Yes, I wanted to discuss it a bit before writing the documentation.

hangy:
What's the integration story for Ory's Hydra & Kratos? It seems like there might be some feature overlap with OIDC profile claims.

alexgarel:
> What's the integration story for Ory's Hydra & Kratos? It seems like there might be some feature overlap with OIDC profile claims.

We might take info from Ory Kratos when it's integrated, but it will take some times.

alexgarel (approved):
LGTM.

Maybe, if you can, document before merging :-) (or after then)

monsieurtanuki:
Cool feature! Ready to implement it on off-dart when it's merged here.

PR #7456, opened 2022-10-07
taxonomy: improve category taxonomy

raphael0202:
Improvements of the category taxonomy, spotted when working on the category matching algorithm.

- en:Teas should be below en:Tea-based beverages instead of the more generic category en:Beverages
- en:Cereal pastas are also en:Pastas
- add italian translations

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -5339,7 +5339,7 @@ fi:omenamehut
 fr:Jus de pomme, jus de pommes
 he:מיצי תפוחים
 hu:Alma levek
-it:Succo di mela
+it:Succo di mela, Spremuta di mela
 nl:Appelsappen, Appelsap
 nl_be:Appelsappen, Appelsap
 pl:Soki jabłkowe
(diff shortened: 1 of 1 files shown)


alexgarel:
@raphael0202 FYI changed title because we have a specific prefix for taxonomy changes (so that they go in the right section in release please)

taxonomies/categories.txt
```diff
@@ -48379,6 +48379,7 @@ nl:Kokosboter, Kokosboters
 
 #do not put in nut butters, peanut is not a nut
 <en:Legume butters
+<en:Nut butters
```
alexgarel, line 48382: What about the comment just above ?

aleene: Well peanut is not a nut. For allergens they differentiate between peanuts and tree nuts (hazel, wal, ...)

aleene (commented):
The easter chocolate eggs can be different from chocolate eggs. Maybe introduce a parent for chocolate eggs?

taxonomies/categories.txt
moon-rabbitOFF: The distinction is also important for the calculation of the nutri-score I think

aleene:
Need to add a test so that this can not be changed.

taxonomies/categories.txt
raphael0202: Indeed, I didn't see the comment. Many contributors make this mistake though, so it's quite frequent that "Nut butters" is suggested as a category for peanut butters.

stephanegigandet (approved):

PR #7457, opened 2022-10-07
build: make mail log

alexgarel:
Instead of sending emails in dev / test

This makes it possible then to:
- see mails when you are in dev, as they are visible in logs
- use the logs to test a mail is sent

Also adds a log if sending emails fail.


stephanegigandet (approved):
Cool, looks good to me

PR #7460, opened 2022-10-09
taxonomy: added materials name in hr

benbenben2:
added croatian words for packaging


stephanegigandet:
The tests are failing because we have the same translation for 2 different entries: cardboard and paperboard...

ERROR - hr:karton already is associated to en:cardboard - hr:karton cannot be mapped to entry en:paperboard

benbenben2:
> The tests are failing because we have the same translation for 2 different entries: cardboard and paperboard...
> 
> ERROR - hr:karton already is associated to en:cardboard - hr:karton cannot be mapped to entry en:paperboard

Sorry! It is fixed now.

stephanegigandet (approved):
Thank you!

PR #7490, opened 2022-10-11
taxonomy: Labels organisation

aleene:
### What
Reorganising and structuring of the labels taxonomy.

For details see: https://wiki.openfoodfacts.org/Labels_hierarchy#Properties


aleene:
This does not work well,

PR #7517, opened 2022-10-12
docs: explanation about packaging data

stephanegigandet:
This document explains how packaging data is currently added, updated and structured in the Open Food Facts database, and how it could be improved. (especially as part of the packaging project which is just starting)

### DIFF
--- a/docs/explanations/packaging-data.md
+++ b/docs/explanations/packaging-data.md
@@ -1,0 +1,138 @@
+# Packaging data
+
+This document explains how packaging data is currently added, updated and structured in the Open Food Facts database, and how it could be improved.
+
+## Introduction
+
+## Types of packaging data
+
+Food products typically have 1 or more packaging components (e.g. milk may have a bottle and a cap).
+
+For each product, we aim to have a comprehensive list of all its packaging components, with detailed information about each packaging component.
+
+### Data about packaging components
+
+For each packaging component, we want data for different attributes, like its shape (e.g. a bottle) and its size (e.g. plastic).
+
+There are many different attributes that can be interesting for specific uses. For instance, researchers in epidemiology are interested in knowing which packaging component is in contact with the food itself, and which one can be put in the microwave oven, so that they can study the long term effects of some plastics on health.
+
+## Sources of packaging data
+
+We can get packaging data from different sources:
+
+### Users
+
+Users of the Open Food Facts website and app, and users of 3rd party apps, can enter packaging data.
+
+### Manufacturers
+
+### Product photos and machine learning
+
+## How packaging data is currently added, updated and structured in Open Food Facts
+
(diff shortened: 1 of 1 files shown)


docs/explanations/packaging-data.md
```diff
@@ -0,0 +1,158 @@
...
+- shape: the general shape of the packaging component (e.g. "bottle", "box")
+- material: the material of the packaging component
+- quantity: how much product the packaging component contains (e.g. "25 cl")
+- recycling: whether the packaging component should be recycled, discarded or reused
+
+The "shape" and "material" fields are taxonomized using the packaging_shapes and packaging_materials taxonomies.
+
+### How the the resulting packagings data structure is created
+
+The values for each input field ("packaging" tag field and "packaging_text_[language code]" packaging information text field) are analyzed to recognize packaging components and their attributes.
```
alexgarel, line 71: Can you add a footnote on which is the corresponding function in ProductOpener ?

alexgarel, line 71: You don't explain how we deal with entries in more than one language. Do we analyze them all ?

docs/explanations/packaging-data.md
```diff
@@ -0,0 +1,158 @@
...
+
+We could discard the existing "packaging" tags field, and replace it with an API to allow clients to add partial information about packaging components.
+
+For instance, if Robotoff detects that the product is in plastic bottle by analyzing a product photo, it could send {shape:"bottle", material:"en:plastic"} and it would be added / combined with the existing "packagings" data.
+
+### Keep the "packaging_text_[language code]" field
+
+It is important to keep this field, as we can display it as-is, use it as input data, and it may contain interesting data that we do not analyze yet.
+
+When filled, the values for this field can be analyzed and added to / combined with the "packagings" data structure.
```
alexgarel, line 143: Explain if there would be a round trip between this field and packaging_text_[language code].
That is if writing in packagings should affect packaging_text ?

Could we also keep a track of which characters (span) of packaging_text mapped to an entry in packagings ?

How do you deal with eventual parts of packaging_text that you are not able to parse ? Do you reject the input in this case ?

alexgarel (approved):
I have some remarks. But overall it's ok.

docs/explanations/packaging-data.md
stephanegigandet: Good idea, I added footnotes.

stephanegigandet: Currently we only analyze the entry corresponding to the main language of the product. Added a note about that.

docs/explanations/packaging-data.md
stephanegigandet: Writing packagings will not affect packaging_text.

We can keep track of what part of the text was matched to what. What we don't recognize is ignored.

Added notes about that in the doc.

PR #7540, opened 2022-10-13
fix: default redis url to empty and handle reconnects

alexgarel:
As in dev people don't necessarily have openfoodfacts-search running

We also changed lib to Redis which is better maintained, and handle connection timeout.

### DIFF
--- a/lib/ProductOpener/Redis.pm
+++ b/lib/ProductOpener/Redis.pm
@@ -34,6 +34,9 @@ The connection to redis
 =cut
 my $redis_client;
 
+# tracking if we already displayed a warning
+my $sent_warning_about_missing_redis_url = 0;
+
 
 =head2 init_redis($is_reconnect=0)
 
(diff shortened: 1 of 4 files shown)


alexgarel:
FYI @simonj2

lib/ProductOpener/Redis.pm
```diff
@@ -1,4 +1,15 @@
 
+=head1 NAME
+
+ProductOpener::Redis - functions to push informations to redis
+
+=head1 DESCRIPTION
+
+C<ProductOpener::Redit> is handling pushing info to Redis
```
stephanegigandet, line 8:
```suggestion
C<ProductOpener::Redis> is handling pushing info to Redis
```

lib/ProductOpener/Redis.pm
```diff
@@ -14,39 +25,96 @@ BEGIN {
...
+		$redis_client = Redis->new(
+			server => $redis_url,
+			# we don't want to sacrifice too much performance for redis problems
+			cnx_timeout => 1,
+			write_timeout => 1,
+		);
+	};
 	if ($@) {
 		$log->warn("Error connecting to Redis", {error => $@}) if $log->is_warn();
+		$redis_client = -1;  # this ask for eventual reconnection
```
stephanegigandet, line 70: It works, but it took me a while to understand that the purpose of the variable $is_reconnect is only to avoid warning multiple times about $redis_url not being set.

I think it would much simpler to declare something like:

my $sent_warning_about_missing_redis_url = 0;

at the beginning after "my $redis_client;", and just test and set this variable once the warning has been sent.

That way we don't have to have a special -1 value for $redis_client: it's either a valid redis client object, or undef.

alexgarel: agree.

stephanegigandet (approved):
Looks good to me, thank you!

alexgarel:
As in dev people don't necessarily have openfoodfacts-search running

We also changed lib to Redis which is better maintained, and handle connection timeout.

PR #7541, opened 2022-10-13
fix: remove old carbon footprint fields

stephanegigandet:
This is to remove old carbon footprint fields (and associated computations) we implemented before the Eco-Score.


alexgarel (approved):
LGTM

Just to be sure: carbon_footprint.tt.json is unrelated right ?

stephanegigandet:
> Just to be sure: carbon_footprint.tt.json is unrelated right ?

It's unrelated, it's computed from Agribalyse.

PR #7551, opened 2022-10-14
fix: nginx config to serve fonts locally

stephanegigandet:
This is to be able to serve fonts hosted locally, from world.openfoodfacts.org (we request fonts from static.openfoodfacts.org, but as they are not on OVH3 yet, OVH3 redirects to world)


alexgarel (approved):
LGTM

PR #7553, opened 2022-10-14
docs: proposal for the new packagings API

stephanegigandet:
This PR:
- documents the existing "packagings" object (slightly extended to add the weight property), and put it in separate files for clarity
- proposes a new /api/v3/product endpoint that would be used with a JSON body instead of multipart/form-data. This new API would initially support only the update of the "packagings" structure.

@Roxie-32 , @alexgarel : sadly Spotlight studio seems to remove all TODOs that were added as comments in the yaml file... :-(

You can see the resulting documentation generated from the OpenAPI file using this url:

https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/openfoodfacts/openfoodfacts-server/packagings-api/docs/reference/api-v3.yml#tag/Write-Requests


stephanegigandet:
image

image

docs/reference/schemas/product_update_api_v3.yaml
```diff
@@ -0,0 +1,10 @@
+title: Product update API V3
+x-stoplight:
+  id: qnmmu2yuzc1xf
+type: object
+description: Model for creating or updating products using the v3 version of the product update API.
+properties:
+  packagings:
+    $ref: ./packagings/packagings.yaml
+  packagings_add:
+    $ref: ./packagings/packagings.yaml
```
alexgarel, line 10: If both fields are exclusive you can specify it with OpenAPI
I think it's:
```suggestion
  oneOf:
  - packagings:
      $ref: ./packagings/packagings.yaml
  - packagings_add:
      $ref: ./packagings/packagings.yaml
```

docs/reference/api.yml
```diff
@@ -209,20 +216,95 @@ paths:
...
+              type: object
+              properties:
+                userid:
+                  type: string
+                password:
+                  type: string
+                code:
+                  type: string
+                product:
+                  $ref: ./schemas/product_update_api_v3.yaml
```
alexgarel, line 246: Don't we need the language code ? Or do we consider it to be embeded in the url (but I don't think this is practical from a client point of view).

docs/reference/schemas/packagings/packaging_component.yaml
```diff
@@ -0,0 +1,41 @@
...
+  The shape, material and recycling properties are mapped to one entry in the packaging_shapes, packaging_materials and packaging_recycling taxonomies, and the value of the property is the canonical name of the taxonomy entry (e.g. en:bottle).
+
+  They may contain values that could not yet get matched to their respective taxonomy, in which case they will contain a free text value prefixed with the language code of this text value (e.g. "fr:Bouteille sphérique" might have been entered by a French user to indicate it is a spherical bottle).
+properties:
+  number:
+    type: integer
+    description: Number of units of this packaging component contained in the product (e.g. 6 for a pack of 6 bottles)
+  shape:
+    type: string
+    description: 'Shape, canonicalized with the packaging_shapes taxonomy.'
```
alexgarel, line 17: when used as the Post body, shouldn't we accept plain values that we will match against the taxonomy ?

Also I imagine sometime something is not yet part of the taxonomy, so it's just a normalized string, right ?
Should we precise if it's part of the taxonomy with a specific field ? (for display only)

stephanegigandet: Right, as input, we accept anything (either plain text in a specified language, or a taxonomy entry).
As output, we return the canonical taxonomy entry, or plain text prefixed by the language code if we couldn't match it to a taxonomy.

I was hoping we could avoid duplicating the packaging_component.yaml (one for input and one for output)

docs/reference/api.yml
stephanegigandet: Good point, I'll add it.

docs/reference/schemas/product_update_api_v3.yaml
stephanegigandet: @alexgarel it cannot be done exactly like that, oneOf expects object schemas, not properties linking to object schemas.

e.g.

```
              oneOf:
                - $ref: '#/components/schemas/Cat'
                - $ref: '#/components/schemas/Dog'
```

So if we wanted to do that, we would first have to define a model "packagings_packagings" that would just contain packagings: $ref: ./packagings/packagings.yaml
And then another "packagings_packagings_add" that would contain "packagings_add: $ref: ./packagings/packagings.yaml", and then we could do the oneOf between the 2.

I think it adds a lot of complexity for not much benefits. Especially since we would have to do that on all similar fields (e.g. for all taxonomies, we will have fields like "categories_add" to add a category to the existing list of categories.

stephanegigandet:
The proposed v3 API is now separated in a distinct api-v3.yaml file:

https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/openfoodfacts/openfoodfacts-server/packagings-api/docs/reference/api-v3.yml#tag/Write-Requests

docs/reference/api-v3.yml
```diff
@@ -0,0 +1,166 @@
...
+      responses: {}
+      description: Retrieve information for a product with a specific barcode.
+      operationId: get-product-by-barcode
+    parameters:
+      - schema:
+          type: string
+        name: barcode
+        in: path
+        required: true
+  /api/v3/product:
```
alexgarel, line 50: Didn't you forgot the barcode parameter ? In fact I think the POST entry is just a new entry under the above path !

alexgarel:
```suggestion
```

stephanegigandet: I was thinking about it, using the same path for both reading and writing, but that means we really have to use the method GET or POST to differentiate the two. It also means the complete "product id" must be in the URL, which is not only the barcode (e.g. on the producers platform, we also need an organization id).

So I'm not quite sure if it's such a good idea to have the same path for READ and WRITE.

docs/reference/schemas/product_update_api_v3.yaml
alexgarel: ok.

docs/reference/schemas/packagings/packaging_component.yaml
alexgarel: It's ok if you just precise it in the description.

docs/reference/api-v3.yml
alexgarel: > we really have to use the method GET or POST to differentiate the two
Yes  that's really what REST advocates (and for caching and many scenario, this is far better)

I really think if we want to have a REST API, that it should be the same path, this is surprising for most developper not to do so.

alexgarel: @stephanegigandet told he finally used the product id. So this should be reflected in this doc :-)

alexgarel:
@stephanegigandet is this PR meant to be closed, as https://github.com/openfoodfacts/openfoodfacts-server/pull/7614 already has the changes ?

PR #7555, opened 2022-10-14
refactor: perltidy (almost) everything

alexgarel:
### What
Perltidy all the code once for all !


alexgarel:
FTR

Perl critic failing checks:
- scripts/api_stats.pl: Code before warnings are enabled at line 41, column 1.  See page 431 of PBP.  (Severity: 4)
  --> add use warnings;
- tests/unit/templates.t: Subroutine "test_template" does not end with "return" at line 20, column 1.  See page 197 of PBP.  (Severity: 4)
  --> add a return;

CharlesNepote (approved):
Seems more readable in many cases, but some results seem also weird such as those semi-columns alone... but I guess all the formatting is automatic, so better homogenous than perfect.

lib/ProductOpener/Ingredients.pm
```diff
@@ -52,54 +52,53 @@ ingredients (is the product vegetarian, vegan, does it contain palm oil etc.)
...
+use Exporter qw< import >;
 
-BEGIN
-{
-	use vars       qw(@ISA @EXPORT_OK %EXPORT_TAGS);
+BEGIN {
+	use vars qw(@ISA @EXPORT_OK %EXPORT_TAGS);
 	@EXPORT_OK = qw(
-		&extract_ingredients_from_image
+	  &extract_ingredients_from_image
```
stephanegigandet, line 60: Is it normal that we have some indentations that are tabs, and some that are 2 spaces, sometime on the same line like here? That feels weird.

alexgarel: hum, this is not cool indeed.

alexgarel: Would it be ok according to you to remove all tabs ? And just use space everywhere ?

alexgarel: The algorithm of perltidy seems to be:
- add indentation as space
- convert any leading group of 4 spaces to tab
We can disable tabs if we want.

Here the "continuation indentation" (that is indentation for a split line) is 2 (the default).
I will change that to 4 to be consistent (I think it's better).

Thus we can also keep tabs transformation, as we want.

alexgarel: from the documentation :
> Using tab characters will almost certainly lead to future portability and maintenance problems, so the default and recommendation is not to use them. For those who prefer tabs, however, there are two different options.

alexgarel: We decided to keep with tab of length 4. Continuation line should be 4.

stephanegigandet (approved):
Looks good!

PR #7557, opened 2022-10-15
ci: More auto-labelling

teolemon:
### What
ci: More auto-labelling for Pull Requests, and also a good way to understand files correlated with a label


alexgarel (approved):
LGTM

PR #7566, opened 2022-10-17
docs: some improvments to knowledge panels api

alexgarel:
As I needed to check the API for a PR on facets knowledge panels, I edited the documentation.

I mostly uses an example and dart code, which already as nice comments to describe properties.
* https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/lib/model/KnowledgePanel.dart
* https://github.com/openfoodfacts/openfoodfacts-dart/blob/master/lib/model/KnowledgePanelElement.dart

There is more to be taken there :-)


docs/reference/schemas/knowledge_panels/elements/text_element.yaml
```diff
@@ -7,8 +7,14 @@ description: |-
 
   For some specific texts that correspond to a product field (e.g. a product name, the ingredients list of a product),the edit_field_* fields are used to indicate how to edit the field value.
 properties:
-  type:
+  text_type:
```
stephanegigandet, line 10: There is discrepancy with the implementation in Dart, the knowledge panels in the web site just use "type", not "text_type".

https://github.com/openfoodfacts/openfoodfacts-server/blob/main/templates/web/panels/panel.tt.html#L154

and the values implemented server side are those:

```
                [% IF text_element.type == "h1" %]
                    <h4 class="panel_title">
                [% ELSIF text_element.type == "quote" %]
                    <blockquote class="panel_text panel_text_quote">
                [% ELSIF text_element.type == "note" %]
                    <div class="panel_text panel_text_note">→
                [% ELSIF text_element.type == "warning" %]
                    <div class="panel_text panel_text_warning">⚠️
                [% ELSIF text_element.type == "callout_info" %]
                    <div class="panel_text alert-box info">
                [% ELSIF text_element.type == "callout_warning" %]
                        <div class="panel_text alert-box warning">⚠️
                [% ELSE %]
                    <div class="panel_text">
                [% END %]
```

docs/reference/schemas/knowledge_panels/elements/element.yaml
```diff
@@ -2,62 +2,35 @@ title: element
...
+      - action
+      - panel
+      - panel_group
+      - table
+    description: |
+      The type of the included element object. 
+      The type also indicates which field contains the included element object. 
+      e.g. if the type is "text", the included element object will be in the "text_element" field.
+
+      TODO: add Map type
```
stephanegigandet, line 22: One thing we should mention is that new element types may be introduced in the future, so we might get values that are not in the enum. In that case, clients should just ignore those elements and not display them. (and they should not crash)

stephanegigandet (approved):
Thanks a lot for completing the documentation! :)

Looks good, except text_type should be type.

PR #7567, opened 2022-10-17
docs: API tutorial (initial version)

Roxie-32:
This is the initial draft of the tutorial " Scan A Product to Get Nutriscore".  However, it still needs a lot of work. There are comments to include possible questions. I am not so sure if the style is appropriate to but I will improve it based on the reviews.


docs/tutorials/scan-a-product-tutorial.md
```diff
@@ -0,0 +1,79 @@
...
+```https://world.openfoodfacts.org/api/v2/product/{barcode}
+```
+
+The {barcode} is the barcode number of the product you are trying to get. The barcode for Nutella Ferrero is 3017624010701. Then the request will be :
+
+```https://world.openfoodfacts.org/api/v2/product/3017624010701
+```
+
+This request will returning every information about Nutella Ferrero on the database. In order to get the nutri-score, we need to limit the response by specify the nutriscore field which is the `nutriscore_data`.
+<!--Is it only nutriscore_data -->
```
alexgarel, line 28: Maybe you also want product_name

if we just want the nutriscore (and not the explanation of it's computation), it's enough to get nutrition_grades field.

Maybe add the nutriscore_data as a follow up to this scenario "Ana would like to be able to know how the score is computed…")

docs/tutorials/scan-a-product-tutorial.md
```diff
@@ -0,0 +1,79 @@
...
+
+**Query Parameters**
+To limit the response of Get A Product By Barcode response , use query parameters to specify the fields of the product to be returned. In this example you need one query parameter called field with the value product_name,nutriscore_data . The key of the parameter is fields and the value
```
alexgarel, line 31: put the new url that limit to some fields.

docs/tutorials/scan-a-product-tutorial.md
```diff
@@ -0,0 +1,79 @@
+# Scan A Product To Get Nutriscore
+
+This is a basic tutorial that shows you how to scan a product to get the nutriscore field. 
+
+Meet Dave. Dave is an active Open Food Facts contributor and a developer who wants to build HealthyFoodChoices, an Android app aimed at conscious consumers that buy healthy products. He has a consumer called Anna. Anna wants to know more on the nutritional facts of Nutella - Ferrero from the HealthyFoodChoices app. Dave needs his app to make an API call to provide her with this information.
+
+<!-- I am not sure If I should retain this scenario of Dave and Anna, Pierre suggested it -->
```
stephanegigandet, line 7: I would remove it, it does not bring any value, and sounds "fake".

Maybe we can replace it with something more factual:

This is a basic tutorial that shows you can get the Nutri-Score of a product, for instance to display it in a mobile app after scanning the product barcode.

(we don't really show how to scan a product, so I rephrased a bit)

docs/tutorials/scan-a-product-tutorial.md
stephanegigandet, line 31:
```suggestion
To limit the response of Get A Product By Barcode response , use query parameters to specify the fields of the product to be returned. In this example you need one query parameter called fields with the value product_name,nutriscore_data . The key of the parameter is fields and the value
```

docs/tutorials/scan-a-product-tutorial.md
stephanegigandet: > Maybe add the nutriscore_data as a follow up to this scenario "Ana would like to be able to know how the score is computed…")

This How-to is about the reader (i.e. the developer), so I would rephrase with "If you would like to be able to show how the score is computed...". Otherwise it means the reader needs to remember who Ana and Dave are (which one was the developer?), and that assumes the reader actually has read the introduction, otherwise it will make no sense at all.

alexgarel (commented):
LGTM

Before merging, if you want, you may add the link to reference.

docs/tutorials/using-the-OFF-API-tutorial.md
```diff
@@ -0,0 +1,117 @@
...
+        "product_name": "Nutella"
+    },
+    "status": 1,
+    "status_verbose": "product found"
+}
+```
+
+### Nutriscore Computation
+
+If you would like to be able to show how the score is computed, add some extra fields like `nutriscore_data` and `nutrition_data`.
```
alexgarel, line 60: nutrition_data is of poor interest here, it just say nutrition data is displayed on the package.
`nutriments` would be the field of interest, but it's a bit long, when you display it, use an ellipsis to just show part of it.

alexgarel (changes requested):
Sorry I requested changes because I saw a problem with the field you choose to display the Nutris-Score computation.

BTW, it would be better also to always write Nutris-Score when you are speaking of it (not when it's the name of a field, of course)

Issue #2325, opened 2019-09-13
CSV export revision (Adding new fields and removing some)


CharlesNepote:
CSV has not evolved since a long time. Some people need new data in the CSV. This issue allows to discuss which new fields could be exported.

**Fields to add:**
- `ingredients_analysis_tags`
  - description: this field is computed from ingredients' analysis to know if the product is or not vegan, vegetarian, and/or with palm-oil
  - Eg. `ingredients_analysis_tags: ["en:palm-oil","en:non-vegan","en:vegetarian-status-unknown"]`
  - rationale: it can be a helper to control data quality

- `nutrient_levels_tags`?
  - description: "It represents the traffic lights system made by the UK FSA (Food Safety Administration). It is used by some manufacturers on a voluntary basis in Great Britain, but was rejected by the European Commission in 2010. On Open Food Facts, when the nutritional values are known, the traffic lights are displayed on the product pages. The calculation formula defining the colors of the lights is described on a [dedicated page of Open Food Facts website](https://world.openfoodfacts.org/nutrition-traffic-lights). Nutrient levels can be found here on the website: https://world.openfoodfacts.org/nutrient-levels
  - Example: `nutrient_levels_tags: ["en:fat-in-high-quantity","en:saturated-fat-in-high-quantity","en:sugars-in-high-quantity","en:salt-in-low-quantity"]`
  - rationale: it can be a helper to control data quality

- `product_quantity`
  - description: "This is the normalized quantity of the product in grams (ISO system)."
  - Example: `product_quantity: "1500"`, computed from `quantity: "1,5 L"`; `product_quantity: "320"`, computed from `quantity: "2 x 160 g"`
  - rationale: it's easier to play with this data as Open Food Facts already do the computation. We already provide `serving_quantity`.

- `owner`
  - description: "This is the owner of the product, which have sent the product's data to Open Food Facts. The list of owners can be found at https://world.openfoodfacts.org/owners"
  - Example: `"owner: "org-carrefour"`
  - rationale: it can be a helper to control data quality

- `data_quality_errors_tags`
  - description: "Returns a list of all detected errors for the product."
  - Example: `data_quality_errors_tags: ["en:nutrition-saturated-fat-greater-than-fat"]` (speaks for itself).
  - rationale: 1. Open Food Facts reusers might want to remove the products with quality issues. 2. It can ease to build tools for data quality. It represents ~30K products as of 2022-10.

[...]

PR #7570, opened 2022-10-18
feat: more fields in CSV (issue 2325)

CharlesNepote:
### What

CSV refactoring

- documentation refactoring + .md file
- added:
  - ingredients_analysis_tags
  - additives_tags
  - nutrient_levels_tags
  - product_quantity
  - owner
  - data_quality_errors_tags
  - unique_scans_n
  - popularity_tags
  - completeness
  - last_image_t

- Enhance #2325

### DIFF
--- a/html/data/data-fields.md
+++ b/html/data/data-fields.md
@@ -92,7 +92,7 @@ The file encoding is Unicode UTF-8. The character that separates fields is <tab>
 
 ## misc. data
 
-- `serving_size` : serving size in g
+- `serving_size` : text describing the serving size
 - `serving_quantity`
 - `no_nutriments` : indicates if the nutrition facts are indicated on the food label
 - `additives_n` : number of food additives
--- a/html/data/data-fields.md
+++ b/html/data/data-fields.md
@@ -1,0 +1,250 @@
+This file describes the fields from the CSV export of the products in the Open Food Facts database.
+
+See https://world.openfoodfacts.org/data for more information.
+
+The file encoding is Unicode UTF-8. The character that separates fields is <tab> (tabulation).
+
+# Generalities
+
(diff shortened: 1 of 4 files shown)


html/data/data-fields.md
```diff
@@ -0,0 +1,252 @@
...
+- `ingredients_analysis_tags`
+- `allergens`
+- `allergens_en`
+- `traces`
+- `traces_tags`
+- `traces_en`
+
+## misc. data
+
+- `serving_size` : serving size in g
```
stephanegigandet, line 95:
```suggestion
- `serving_size` : text describing the serving size
```

html/data/data-fields.md
```diff
@@ -0,0 +1,252 @@
...
+- `serving_quantity`
```
stephanegigandet, line 96:
```suggestion
- `serving_quantity` : normalized serving size in grams or ml
```

stephanegigandet (approved):
I think we should fix the field names for nutrients that have a leading dash. It's a bug and it has no meaning.
What happened is that some fields had 1 or 2 leading dashes, and we just removed the first one. We should remove both.

html/data/data-fields.md
```diff
@@ -0,0 +1,252 @@
...
+
+Please see https://wiki.openfoodfacts.org/Nutrients_handling_in_Open_Food_Facts for more information on nutrients.
+
+- `energy-kj_100g`
+- `energy-kcal_100g`
+- `energy_100g`
+- `energy-from-fat_100g`
+- `fat_100g`
+- `saturated-fat_100g`
+- `-butyric-acid_100g`
```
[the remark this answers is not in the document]
CharlesNepote: This issue exists until 2017-09. I suggest we stay as we are not to break possible usages.

We can take more time later to build a completely new format with many more bug fixes and enhancements.

PR #7581, opened 2022-10-21
ci: add allergen taxonomy labelling

teolemon:
### What
ci: add allergen taxonomy labelling


stephanegigandet:
Already included, can be closed I think

PR #7586, opened 2022-10-21
fix: importing uploaded files on producers platform

stephanegigandet:
A recent change prevented uploaded files to be imported on the producers platform.


alexgarel (approved):
LGTM

PR #7603, opened 2022-10-24
docs: improvements to the image API

Roxie-32:
Moving on from knowledge panels, I moved on to get only images for a specific product.


alexgarel:
@Roxie-32 I'm a bit confused, I was expecting images schema to be in its own file !

Issue #7564, opened 2022-10-17
Request for feedback: new product WRITE API with structured JSON body and new packagings structure


stephanegigandet:
Hi Everyone! We are designing a new API so that clients can specify the different packaging components of a product. The packagings will be in a complex structure, so we are departing from the existing API that uses flat form fields to a new API that will accept directly a JSON structure as input. While we are focusing first on packagings, this new API can gradually be extended to support other product fields.

It would be great to get your feedback and suggestions on this new API, for packagings in particular, but also for updating products in general.

The API is described in OpenAPI in this PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/7553

You can see auto-generated documentation from this OpenAPI here: https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/openfoodfacts/openfoodfacts-server/packagings-api/docs/reference/api-v3.yml#operation/post-api-v3-product-barcode

To discuss, you can use this bug, and/or #packagings and #api on Slack.

monsieurtanuki:
@raphael0202 For the record in off-dart when we use `userId` / `password` we do that for WRITE operations (like here about saving the packaging) and we use POST syntax - all parameters will be in the body, not in the URL.
The `userId` is also in the header, as `From: `.
(Not the same when we call Robotoff).

@raphael0202 Good question about the `product` field: it looks like currently we don't use one and just put everything that belongs to a product directly to the root.

@stephanegigandet Generally speaking, anything that can improve the READ / WRITE is welcome! Especially with fields that works in both READ and WRITE.

I have no special interest in the `_add` syntax (off-dart, Smoothie).

Regarding JSON data, we could use
* one line for each component (e.g. `{material: "en:pet-polyethylene-terephthalate",recycling: "en:recycle", shape: "en:bottle"}`)
* one optional meta-line for items like "user says it's complete" and "we think something is missing"

I haven't understood how the data will be translated for the user, e.g. in Smoothie. What are we supposed to do with `en:recycle` - will there be a way to say "Please give me the translations too" in READ? Or are we supposed to download the taxonomy?

stephanegigandet:
Thanks for the feedback @raphael0202 and @monsieurtanuki 

1. Authentication in the body vs HTTP Basic Auth

We could use HTTP Basic Auth for authentication. It would conflict with some of the things we do (like having the extra "off:off" password protection on .net), but we could change that.

I propose this:
[...]

PR #7453, opened 2022-10-06
feat: add a validation process to new org creation

alexgarel:
### What
As a new user subscribe, telling he is from a company, if it's the first for that company, we currently create a new org and attach the user to it.

As there is a lot of potential fake data, we would like a "validation" round.

fixes: #7350

PR #7614, opened 2022-10-27
feat: Start of implementation of API v3 product read and write + integration tests

stephanegigandet:
Minimal implementation of API v3 to write packaging data.

Changes:
- New Routing.pm module to move code from Display.pm in charge of analyzing requests (urls, paths, parameters etc.)
- Refactoring to be able to read the JSON body of a request before it is consumed by Apache::Request or CGI.pm
- New API.pm module to put all code related to the API (in particular of all API v3 code)
- Separate APIReadProduct and APIWriteProduct submodules for each API type
- Created a new execute_tests() function in APITest.pm to test the API with various methods (GET / POST), paths, query strings, form params or JSON body
- Added tests for both API v2 and API v3 product READ and WRITE.
- Some restructuring of Packagings.pm so that it can be used in a more generic way

To test locally, you can send API requests using curl:

`curl -X POST http://world.openfoodfacts.localhost/api/v3/product/12345678 -H "Content-Type: application/json" -d '{"product": { "packagings_add": [{"shape": "bottle"}]}, "fields": "updated"}'`

The code is also deployed on the .dev server, so you can test API queries to it:

`curl -X POST https://off:off@world.openfoodfacts.dev/api/v3/product/12345678 -H "Content-Type: application/json" -d '{"product": { "packagings_add": [{"shape": "bottle"}]}, "fields": "updated"}'`

Related to #7564


lib/ProductOpener/APIProductWrite.pm
```diff
@@ -0,0 +1,210 @@
...
+
+			# Process the product data
+			analyze_and_enrich_product_data($product_ref);
+
+			# Save the product
+			my $comment = $request_body_ref->{comment} || "API v3";
+			if (store_product($User_id, $product_ref, $comment)) {
+				# Notify robotoff
+				send_notification_for_product_change($product_ref, "updated");
+			}
```
alexgarel, line 194: we should move that in store_product like for off-search indexation, it's silly to repeat this notification everywhere…

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -0,0 +1,210 @@
...
+		else {
+
+			# Update the product
+			update_product_fields($request_ref, $product_ref);
+
+			# Process the product data
+			analyze_and_enrich_product_data($product_ref);
+
+			# Save the product
+			my $comment = $request_body_ref->{comment} || "API v3";
```
alexgarel, line 190: don't we want to add updated_fields in the default message ?

lib/ProductOpener/APITest.pm
```diff
@@ -179,4 +183,112 @@ sub construct_test_url ($target, $prefix) {
...
+			elsif (defined $test_ref->{form}) {
+				$response = $ua->post($url, Content => $test_ref->{form});
+			}
+			else {
+				$response = $ua->post($url);
+			}
+		}
+
+		# Check if the request was successful
+		if (not $response->is_success) {
```
alexgarel, line 253: We could also test a  $response has a different return value, so I would instead handle a $test_ref->code parameter (and default to is_success)

lib/ProductOpener/APITest.pm
```diff
@@ -179,4 +183,112 @@ sub construct_test_url ($target, $prefix) {
...
+and the directories for the tests are derived from it.
+
+=head4 $tests_ref reference to list of tests
+
+The tests are in a structure like this:
+
+my $tests_ref = (
+    [
+		{
+			test_case => 'no-body',
```
alexgarel, line 204:
```suggestion
			test_case => 'no-body',  # a description of the test, should be unique to easily retrieve which test failed
```

lib/ProductOpener/APITest.pm
```diff
@@ -179,4 +183,112 @@ sub construct_test_url ($target, $prefix) {
 	return $url;
 }
 
+=head2 execute_api_tests($file, $tests_ref)
```
alexgarel, line 186: Great idea !

alexgarel (commented):
Here are my first comments, I didn't review it all yet ! (read up to Display.pm)

lib/ProductOpener/APIProductWrite.pm
stephanegigandet: for every change we already compute the diffs (added and changed fields) and display them in the history, so there's no need to put them in the comment

lib/ProductOpener/APIProductRead.pm
```diff
@@ -0,0 +1,183 @@
...
+
+=head3 Parameters
+
+=head4 $request_ref (input)
+
+Reference to the request object.
+
+=cut
+
+sub read_product_api ($request_ref) {
```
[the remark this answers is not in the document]
stephanegigandet: So today we still have the v0 to v2 requests handled in Display.pm display_product_api(), but I'm planning to remove all the code from it and to just make it a wrapper around this v3 function. I'll add a comment about that.

lib/ProductOpener/APIProductRead.pm
```diff
@@ -0,0 +1,183 @@
...
+
+		$request_ref->{code}
+			= $options{"sample_product_code_country_${cc}_language_${lc}"}
+			|| $options{"sample_product_code_country_${cc}"}
+			|| $options{"sample_product_code_language_${lc}"}
+			|| $options{"sample_product_code"}
+			|| "";
+	}
+
+	my $code = normalize_code($request_ref->{code});
```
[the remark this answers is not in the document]
stephanegigandet: Good idea

docs/reference/api-v3.yml
```diff
@@ -0,0 +1,203 @@
...
+                  - $ref: ./schemas/product.yaml
+      description: Retrieve information for a product with a specific barcode.
+      operationId: get-product-by-barcode
+    parameters:
+      - schema:
+          type: string
+        name: barcode
+        in: path
+        required: true
+  /api/v3/product:
```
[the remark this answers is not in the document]
stephanegigandet: Right, the barcode is now in the path for WRITE POST requests as well.

lib/ProductOpener/APIProductRead.pm
```diff
@@ -0,0 +1,183 @@
...
+	my $rev = single_param("rev");
+	local $log->context->{rev} = $rev;
+	if (defined $rev) {
+		$product_ref = retrieve_product_rev($product_id, $rev);
+	}
+	else {
+		$product_ref = retrieve_product($product_id);
+	}
+
+	if ($code !~ /^\d{4,24}$/) {
```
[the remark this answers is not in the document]
stephanegigandet: I'll move it.

lib/ProductOpener/APIProductRead.pm
```diff
@@ -0,0 +1,183 @@
...
+		$log->info("invalid code", {code => $code, original_code => $request_ref->{code}}) if $log->is_info();
+		add_error(
+			$request_ref->{api_response},
+			{
+				message => {id => "invalid_code"},
+				field => {id => $code, value => $code},
+				impact => {id => "request_failed"},
+			}
+		);
+		$request_ref->{api_response}{result} = {id => "product_not_found"};
```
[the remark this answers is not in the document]
stephanegigandet: I'll add it.

lib/ProductOpener/APIProductRead.pm
```diff
@@ -0,0 +1,183 @@
...
+	}
+	else {
+		$request_ref->{api_response}{result} = {id => "product_found"};
+
+		add_images_urls_to_product($product_ref, $lc);
+
+		# If the request specified a value for the fields parameter, return only the fields listed
+		my $fields = single_param('fields');
+		if (defined $fields) {
```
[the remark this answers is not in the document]
stephanegigandet: Right, we can call customize_response_for_product directly

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -0,0 +1,210 @@
...
+use ProductOpener::Display qw/:all/;
+use ProductOpener::Users qw/:all/;
+use ProductOpener::Lang qw/:all/;
+use ProductOpener::Products qw/:all/;
+use ProductOpener::API qw/:all/;
+use ProductOpener::Packaging qw/:all/;
+
+=head2 update_product_fields ($request_ref, $product_ref)
+
+Update product fields based on input product data.
```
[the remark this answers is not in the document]
stephanegigandet: I'm hoping that I can eventually remove most of the code from product_jqm_multilingual.pl (which handles v2) and make it wrapper to v3. I'll add a comment about that.

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -0,0 +1,210 @@
...
+
+	$request_ref->{updated_product_fields} = {};
+
+	my $input_product_ref = $request_body_ref->{product};
+
+	foreach my $field (sort keys %{$input_product_ref}) {
+
+		my $value = $input_product_ref->{$field};
+
+		if ($field =~ /^(packagings)(_add)?/) {
```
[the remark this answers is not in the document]
stephanegigandet: Right, we should definitely do that eventually, but I'm not sure yet what code can be shared / how to structure it for the different fields. When we have a few more fields implemented, it should be easier to determine the best way to do it.

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -0,0 +1,210 @@
...
+			if (ref($value) ne 'ARRAY') {
+				add_error(
+					$response_ref,
+					{
+						message => {id => "invalid_type_must_be_array"},
+						field => {id => $field},
+						impact => {id => "field_ignored"},
+					}
+				);
+			}
```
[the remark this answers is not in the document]
stephanegigandet: Good point

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -0,0 +1,210 @@
...
+
+	my $input_product_ref = $request_body_ref->{product};
+
+	foreach my $field (sort keys %{$input_product_ref}) {
+
+		my $value = $input_product_ref->{$field};
+
+		if ($field =~ /^(packagings)(_add)?/) {
+			$request_ref->{updated_product_fields}{$1} = 1;
+			my $add = $2;
```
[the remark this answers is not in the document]
stephanegigandet: ok

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -0,0 +1,210 @@
...
+				}
+			);
+		}
+
+		# Load the product
+		my $code = $request_ref->{code};
+		my $product_id = product_id_for_owner($Owner_id, $code);
+		my $product_ref = retrieve_product($product_id);
+
+		if (not defined $product_ref) {
```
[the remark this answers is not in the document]
stephanegigandet: I was looking into that, but PUT is intended to replace an existing document entirely, which is not what we do.

When we don't know the barcode, we could use POST on /product
When we know the barcode, ideally we should use PATCH on /product/[barcode] https://datatracker.ietf.org/doc/rfc5789/?include_text=1

but I didn't look how much support PATCH has in HTTP libraries

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -0,0 +1,210 @@
...
+				{
+					message => {id => "missing_field"},
+					field => {id => "tags_lc", default_value => $request_body_ref->{tags_lc}},
+					impact => {id => "warning"},
+				}
+			);
+		}
+
+		# Load the product
+		my $code = $request_ref->{code};
```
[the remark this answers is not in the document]
stephanegigandet: Good point, I'll do that.

lib/ProductOpener/APIProductWrite.pm
stephanegigandet: ok

lib/ProductOpener/APITest.pm
stephanegigandet: Good idea, will do that.

lib/ProductOpener/Packaging.pm
```diff
@@ -341,6 +345,283 @@ sub guess_language_of_packaging_text ($text, $potential_lcs_ref) {
...
+
+=head3 Parameters
+
+=head4 $product_ref product
+
+=head4 $packaging_ref packaging component
+
+=cut
+
+sub apply_rules_to_augment_packaging_component_data ($product_ref, $packaging_ref) {
```
alexgarel, line 484: :+1: for splitting in different functions with simpler scopes

lib/ProductOpener/Packaging.pm
```diff
@@ -341,6 +345,283 @@ sub guess_language_of_packaging_text ($text, $potential_lcs_ref) {
...
+				if (not defined $existing_packaging_ref->{$property}) {
+					next;
+				}
+
+				my $tagtype = $packaging_taxonomies{$property};
+
+				# $tagtype can be shape / material / recycling, or undef if the property is something else (e.g. a number of packagings)
+				if (not defined $tagtype) {
+					# If there is an existing value for the property,
+					# check if it is the same
```
alexgarel, line 563: why so ?
So if there was 1 plastic bag, and I fix for 2 plastic bag, this will indeed create a new entry ?

lib/ProductOpener/Packaging.pm
```diff
@@ -341,6 +345,283 @@ sub guess_language_of_packaging_text ($text, $potential_lcs_ref) {
...
+						# Type is a number
+						if ($existing_packaging_ref->{$property} != $packaging_ref->{$property}) {
+							$match = 0;
+							last;
+						}
+					}
+				}
+
+				# If there is an existing value for the property,
+				# check if it is either a child or a parent of the value extracted from the packaging text
```
alexgarel, line 574:
```suggestion
				# If there is an existing value for the taxonomized property,
				# check if it is either a child or a parent of the value extracted from the packaging text
```

lib/ProductOpener/Packaging.pm
```diff
@@ -360,12 +641,40 @@ sub analyze_and_combine_packaging_data ($product_ref) {
 	$log->debug("analyze_and_combine_packaging_data - start", {existing_packagings => $product_ref->{packagings}})
 		if $log->is_debug();
 
+	# Response structure to keep track of warnings and errors
+	my $response_ref = get_initialized_response();
```
alexgarel, line 645: I find it strange that it's not passed as an argument !
I imagine this function as a small part of a lot of processing like that, maybe combining and putting data in same `$response_ref`.

lib/ProductOpener/Packaging.pm
```diff
@@ -360,12 +641,40 @@ sub analyze_and_combine_packaging_data ($product_ref) {
...
+				$packaging_ref->{quantity_per_unit} = $packaging_ref->{quantity};
+				$packaging_ref->{quantity_per_unit_value}
+					= convert_string_to_number($packaging_ref->{quantity_per_unit_value});
+				$packaging_ref->{quantity_per_unit_unit} = $packaging_ref->{quantity_unit};
+			}
+			delete $packaging_ref->{quantity};
+			delete $packaging_ref->{quantity_value};
+			delete $packaging_ref->{quantity_unit};
+		}
+	}
```
alexgarel, line 676: I will put that in a small function `migrate_packanging_units_quantity` as it messes reading here.
We could even have a Migrations.pm module, so that it's easier to eventually dismiss some function, if at some point we decide to run them on whole database

tests/integration/api_v3_product_read.t
```diff
@@ -0,0 +1,67 @@
...
+
+use File::Basename "dirname";
+
+use Storable qw(dclone);
+
+remove_all_users();
+
+remove_all_products();
+
+wait_dynamic_front();
```
alexgarel, line 18: FYI, in [my pending PR](https://github.com/openfoodfacts/openfoodfacts-server/pull/7453), I added a `wait_application_ready` that also wait for server to be ready, for in some case, test where failing because apache server was not ready to serve request.

tests/integration/api_v3_product_write.t
```diff
@@ -0,0 +1,144 @@
...
+remove_all_products();
+
+wait_dynamic_front();
+
+my $ua = new_client();
+
+my %create_user_args = (%default_user_form, (email => 'bob@gmail.com'));
+create_user($ua, \%create_user_args);
+
+my $tests_ref = [
```
alexgarel, line 25: I think we should add a comment in all tests where we use this pattern, because that might be surprising for a new comer not to see where expected data is !

```suggestion
# Note: expected results are stored in json files, see execute_api_tests
my $tests_ref = [
```

tests/update_tests_results.sh
```diff
@@ -29,4 +29,6 @@ cd ..
 
 perl integration/import_csv_file.t --update-expected-results
 perl integration/export.t --update-expected-results
+perl integration/api_v3_product_read.t --update-expected-results
+perl integration/api_v3_product_write.t --update-expected-results
```
alexgarel, line 33: At some point we will write it as a `for ... in` loop :grimacing:

lib/ProductOpener/Packaging.pm
stephanegigandet: If you fix something, then you use packagings and send the whole structure, not packagings_add.

packagings_add is used when the caller does not know the whole structure, but parts of it. e.g. Robotoff sees a PET logo, so it sends something like packagings_add -> { material : "pet" }

lib/ProductOpener/Packaging.pm
stephanegigandet: So in fact, this instance of $response_ref is currently not used, we don't do anything with it. We just need one because some of the called functions in it expect one, as we reuse them in other places.

We could move up $response_ref in an argument, all the way to store_product() (which runs all the processing). But then currently we won't use the result.

stephanegigandet: so I'll move the $response_ref = get_initalized_response() in the 3 callers (Import.pm, product_multiliingual.pl and product_jqm_multilingual.pl)

stephanegigandet: and update_all_products.pl

lib/ProductOpener/Packaging.pm
stephanegigandet: ok, I'm creating a separate function migrate_old_number_and_quantity_fields_202211($product_ref)

lib/ProductOpener/API.pm
```diff
@@ -0,0 +1,596 @@
...
+
+=head4 $product_ref (input)
+
+Reference to the product object (retrieved from disk or from a MongoDB query)
+
+=head3 Return value
+
+Reference to the customized product object.
+
+=cut
```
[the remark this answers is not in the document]
stephanegigandet: Fixed.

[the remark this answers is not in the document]
stephanegigandet: In next push I mean.

lib/ProductOpener/API.pm
```diff
@@ -0,0 +1,596 @@
...
+				}
+			}
+		}
+		# Eco-Score
+		elsif ($field =~ /^ecoscore/) {
+
+			if (defined $product_ref->{$field}) {
+				$customized_product_ref->{$field} = $product_ref->{$field};
+			}
+		}
```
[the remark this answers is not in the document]
stephanegigandet: we can, I think it's because we used to initialize the Eco-Score there, but now we do it always.

lib/ProductOpener/API.pm
```diff
@@ -0,0 +1,596 @@
...
+					)
+				{
+					$customized_product_ref->{$field}{$imagetype_lc} = 0;
+				}
+			}
+		}
+
+		elsif ((not defined $customized_product_ref->{$field}) and (defined $product_ref->{$field})) {
+			$customized_product_ref->{$field} = $product_ref->{$field};
+		}
```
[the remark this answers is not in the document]
stephanegigandet: It's a good idea, but we do not currently maintain a list of possible fields. I will add a comment about this.

tests/integration/api_v3_product_read.t
stephanegigandet: Perfect.

stephanegigandet:
@alexgarel Thanks a lot for all the great suggestions, I think I made changes to address all of them.

The only remaining one is about which method to use for WRITE requests, I opened an issue to discuss it: https://github.com/openfoodfacts/openfoodfacts-server/issues/7667

PR #7645, opened 2022-11-02
docs: add open api documentation to index

alexgarel:
* Make Open API documentation appear on the right
* remove the TODO file from the generated doc


stephanegigandet (approved):

Roxie-32 (approved):
Now I understand, looking forward to how it will render on the view 😎

PR #7647, opened 2022-11-02
taxonomy: Meats and their products

moon-rabbitOFF:


taxonomies/categories.txt
```diff
@@ -81554,7 +81571,7 @@ protected_name_type:en: pdo
 origins:en: en:Finland
 #wikidata:en:
 
-<en:Meat-based products
+<en:Meats and their products
```
stephanegigandet, line 81574: That one is (and was) probably incorrect.

moon-rabbitOFF: It's corrected, thank you

stephanegigandet (approved):
Thanks!

Issue #7662, opened 2022-11-07
gulp-terser dependencies conflict in npm for tagify and gulp-terser-js / gulp-min


teolemon:
### What
- Country taxonomy is probably not built automatically on .net 
<img width="298" alt="image" src="https://user-images.githubusercontent.com/1689815/200352241-e836d197-b6f4-4833-b6b3-d045f91ef599.png">

stephanegigandet:
https://github.com/gulp-community/gulp-terser-js/issues/28 is still open

alexgarel:
I think it's solved.

Issue #7642, opened 2022-11-02
Tags input bars in product edit form are not full width


stephanegigandet:
### Describe the bug

May be due to the libraries updates to make npm work: https://github.com/openfoodfacts/openfoodfacts-server/pull/7619

![image](https://user-images.githubusercontent.com/8158668/199445376-d44220e6-c599-4ed7-a8e4-69622e646891.png)

### To Reproduce

Edit a product (on .net)

### Expected behavior

Full width inputs

### Type of device

Browser

stephanegigandet:
"fixed" by downgrading tagify

PR #7670, opened 2022-11-08
fix: downgrade tagify to solve npm dependencies conflict

stephanegigandet:
Solves #7662 

Also solves #7642


alexgarel (approved):
:shipit:

PR #7674, opened 2022-11-09
docs: improvement to the Images API

Roxie-32:
- Separate the image into 2 schemas (one for the normal image(with a number as the key) and one selected image(eg..packaging_fr))
- Create a request for Images API


Roxie-32:
@alexgarel I made another PR for the images API and closed the last one, the commits in the last PR were mixed up. I wasn't so sure if I got the pattern properties section but I made an attempt. I am also unsure of what the regex will be for selected images like front_en etc.

docs/reference/api.yml
```diff
@@ -249,20 +247,59 @@ paths:
...
+                  product:
+                    type: object
+                    properties:
+                      images:
+                        type: object
+                        # review the regex expression for pattern properties
+                        patternProperties:
+                          '/^\d+$/':
+                            $ref: ./schemas/images/image.yaml
+                          packaging_fr:
```
alexgarel, line 281: this also should be a regexp: 

```suggestion
                          '(front|ingredients|nutrition|packaging|other)_\w\w(-\w\w)?:
```

docs/reference/schemas/images/image.yaml
```diff
@@ -0,0 +1,38 @@
...
+    properties:
+      '100':
+        type: object
+        properties:
+          h:
+            type: integer
+            example: 100
+          w:
+            type: integer
+            example: 56
```
alexgarel, line 14: factor this out (also used in selected_image)

alexgarel (changes requested):
I would expect to have a description for every field (you may ask on slack for those who are too obscure)

Roxie-32:
@alexgarel I think I added all descriptions except for these `coordinates_image_size`, `rev`, `geometry` then `x1, x2....`. I didn't really get them.  Can you review it again and see if there are more corrections? I also factored out the image size(height and width to a separate file.

Roxie-32:
> @alexgarel I think I added all descriptions except for these `coordinates_image_size`, `rev`, `geometry` then `x1, x2....`. I didn't really get them. Can you review it again and see if there are more corrections? I also factored out the image size(height and width to a separate file.

@alexgarel @stephanegigandet were you able to see this please?

alexgarel (approved):
It's already a good improvement, thanks @Roxie-32 

Let's merge it.

PR #7676, opened 2022-11-09
chore: week 45 taxonomy updates

aleene:
### What
Many changes in various taxonomies.


taxonomies/categories.txt
```diff
@@ -7653,11 +7653,12 @@ nl:Oplosdranken, Warme oplosdranken, Warme oplosdrank
 ru:Быстрорастворимые горячие напитки
 
 <en:Plant-based foods and beverages
-en:Dairies substitutes
+en:Dairy substitutes
```
stephanegigandet, line 7656: When you rename the main entry, can you add it as a synonym? Otherwise the old products will have both the new entry (taxonomized) and old entry (not taxonomized)

stephanegigandet (commented):
Looks good, we just need to keep renamed canonical entries as synonyms.

taxonomies/categories.txt
aleene: OK

stephanegigandet: it's strange, I see you commit that added the synonym, but I don't see it in the resulting changes

taxonomies/categories.txt
```diff
@@ -96315,7 +96315,7 @@ fr:Saucisses végétariennes de Vienne, saucisses végétales de Vienne
 nl:Vegetarische weense worstjes
 
 <en:Plant-based foods
-<en:Dairy substitutes
+<en:Dairy substitutes, Dairies substitutes
```
stephanegigandet, line 96318: the synonym must be on the entry itself, not on the line that indicates that there is a parent

stephanegigandet (approved):
Thank you!

PR #7679, opened 2022-11-09
refactor: new load_data() function to be used in startup_apache2.pl and batch scripts

stephanegigandet:
- create a new LoadData.pm module, with a load_data() function that loads packaging codes data, Agribalyse, Eco-Score etc. 
This is to remove duplicate code in startup_apache2.pl, minion_producers.pl and import_csv_file.pm, and also to remove the INIT block in PackagersCodes.pm which printed a warning "too late for INIT block" when starting Apache
- Load all modules in startup_apache2.pl (we were missing a few new ones)
- use "use Module" instead of "require Module"
- remove warning "remove warning - sort (...) interpreted as function" in GS1.pm
- removed some tests scripts for minion (we only use minion_producers.pl):

minion-off-pro.service:ExecStart=/srv/off-pro/scripts/minion_producers.pl minion worker -m production -q pro.openfoodfacts.org
minion-off.service:ExecStart=/srv/off/scripts/minion_producers.pl minion worker -m production -q openfoodfacts.org


alexgarel (approved):
Cool move.

Only point:
* you pushed an empty tests/update_tests_results.sh file !
* you have to run `make lint_perltidy`

Issue #7505, opened 2022-10-11
NEW DESIGN - The Sort icon is too small in responsive mode


teolemon:
### What
- [ ] The Sort icon is too small in responsive mode
### Screenshot
![image](https://user-images.githubusercontent.com/1689815/195130464-45b30a9e-b3fe-49e4-8f54-0244d35a836e.png)
### Part of
- #7295

alexgarel:
Thank you @sultanowski for those contributions !

PR #7685, opened 2022-11-10
style: Fixed filter bar

sultanowski:
### What
* Tweaked filter bar to look better on both desktop and mobile.
* Deleted inline styles and created new classes in external file.

### Screenshots
https://imgur.com/a/6vSD9P3

- #7505 -> at least, partly.


stephanegigandet:
Thanks!

Would there be a way for the icons to not be reduced on very small screens?

image

stephanegigandet:
> I saw that every icon in button has this issue. Maybe we should refactor whole button component/class?

Is it every button? I'm trying to find others with the same issue, but the icons don't seem to be reduced:

image

sultanowski:
"Get an App" button in Navigation Bar also shrinks icon. I would say that there's few of them. Every of them should work the same.

sultanowski:
@stephanegigandet Now icons should not shrink on resize.

stephanegigandet:
@sultanowski Perfect, thank you!

stephanegigandet (approved):

PR #7689, opened 2022-11-12
chore: Packaging translations and descriptions

aleene:
### What
Added translations and descriptions. These have been taken from various wiktionary entries.
As there are many languages I tried to limit the coverage a bit.


stephanegigandet:
There's one translation associated with 2 entries:

ERROR - bg:пълнеж already is associated to en:backing - bg:пълнеж cannot be mapped to entry en:filling

stephanegigandet (approved):

Issue #7699, opened 2022-11-15
GitHub tests failing because of Perl modules install failing


stephanegigandet:
We seem to have more failed CI tests than usual lately, related to the install of cpan modules in Docker:
https://github.com/openfoodfacts/openfoodfacts-server/pull/7614

Pull Requests / Build backend dev image for tests (pull_request) Failing after 28m
failed test: https://github.com/openfoodfacts/openfoodfacts-server/pull/7614

grep fail :
```
#15 692.3 ! Failed to fetch distribution Task-Weaken-1.06
#15 782.5 ! Failed to fetch distribution PPI-1.276
#15 782.5 ! Installing the dependencies failed: Module 'PPI::Dumper' is not installed, Module 'PPI::Document' is not installed, Module 'Task::Weaken' is not installed
#15 782.5 ! Bailing out the installation for PPIx-Regexp-0.085.
#15 815.1 ! Installing the dependencies failed: Module 'PPI::Document' is not installed, Module 'PPI::Dumper' is not installed
#15 815.1 ! Bailing out the installation for PPIx-QuoteLike-0.023.
#15 848.7 ! Installing the dependencies failed: Module 'Task::Weaken' is not installed, Module 'PPI::Document' is not installed, Module 'PPI::Document::Fragment' is not installed, Module 'PPI::Dumper' is not installed, Module 'PPI' is not installed
#15 848.7 ! Bailing out the installation for PPIx-Utilities-1.001000.
#15 860.7 ! Installing the dependencies failed: Module 'PPIx::Regexp::Util' is not installed, Module 'PPIx::QuoteLike' is not installed, Module 'PPI::Document' is not installed, Module 'PPIx::Utilities::Node' is not installed, Module 'PPI::Token::Quote::Single' is not installed, Module 'Task::Weaken' is not installed, Module 'PPI::Document::File' is not installed, Module 'PPIx::Regexp' is not installed, Module 'PPI' is not installed, Module 'PPI::Node' is not installed, Module 'PPI::Token::Whitespace' is not installed, Module 'PPIx::Utilities::Statement' is not installed
#15 860.7 ! Bailing out the installation for Perl-Critic-1.140.
#15 861.9 ! Installing the dependencies failed: Module 'Perl::Critic::Utils' is not installed, Module 'Perl::Critic' is not installed, Module 'Perl::Critic::Violation' is not installed
#15 861.9 ! Bailing out the installation for Test-Perl-Critic-1.04.
[...]

PR #7701, opened 2022-11-15
fix: make command to install perl dev modules

stephanegigandet:
Trying to fix https://github.com/openfoodfacts/openfoodfacts-server/issues/7699

This removes some modules that are not needed by GitHub actions, and creates a "make install_perl_dev_modules" to install them.


alexgarel (approved):
@stephanegigandet I propose another way to deal with it, at build time.

stephanegigandet (commented):
Great, thanks!

PR #7710, opened 2022-11-17
ci: force css rebuild

alexgarel:
Trying to have a SCSS rebuild as it's broken on staging.


alexgarel:
This is not the cause of the bug, closing it.

PR #7713, opened 2022-11-17
build: remove some volumes in prod

alexgarel:
They are not useful, and may lead to errors


alexgarel:
Ok, those volumes are there for a reason :grimacing: we will have to deal with that !

PR #7715, opened 2022-11-17
fix: Routing.pm error

stephanegigandet:
fix error loading https://world.openfoodfacts.net/photographer/teolemon


teolemon (approved):

teolemon:
Do you have a suitable label for routing.pm ? API ?

stephanegigandet:
> Do you have a suitable label for routing.pm ? API ?

Not really, "Routing" could work if we really need to have a label for everything.

Issue #5720, opened 2021-09-21
Drive more donations to Open Food Facts


teolemon:
### Who for
- potential donators

teolemon:
Considering a public list of donors like OSM ? https://supporting.openstreetmap.org/donate/

PR #7719, opened 2022-11-17
fix: add UTM to the login page donation promo

teolemon:
### What
fix: add UTM to the login page donation promo
### Part of
- #5720


stephanegigandet (commented):
only en.po and not common.pot?

stephanegigandet (approved):

PR #7749, opened 2022-11-22
feat: translations in packagings READ API v3 

stephanegigandet:
This PR changes the structure of the returned "packagings" structure so that it can include translated name in the language specified by tags_lc (if sent), as described in #7564 and the OpenAPI specification.

Note that this new structure is only returned for v3 requests, in order not to break apps (not sure if there are some that use this field already) that expect the old structure.

Also added tests.

### DIFF
--- a/tests/integration/api_v2_product_read.t
+++ b/tests/integration/api_v2_product_read.t
@@ -101,9 +101,9 @@ my $tests_ref = [
 		query_string => '?fields=packagings',
 		expected_status_code => 200,
 	},
+	# test that on v2 tags_lc parameter have no effects
+	# packaging part should be exactly the same as get-packagings
 	{
-        # test that on v2 tags_lc parameter have no effects
-        # packaging part should be exactly the same as get-packagings
 		test_case => 'get-packagings-fr',
 		method => 'GET',
 		path => '/api/v2/product/200000000034',


tests/integration/expected_test_results/search_test/q1.json
```diff
@@ -1478,23 +1478,23 @@
          "entry_dates_tags" : "--ignore--",
          "food_groups_tags" : [],
          "forest_footprint_data" : {
-            "footprint_per_kg" : 0.0132222222222222,
+            "footprint_per_kg" : "0.0132222222222222",
```
olivier5741, line 1481: Integer to string (Perl issue apparently), can live with it :)

monsieurtanuki (approved):
Hi @stephanegigandet!
I'm not an expert in `.pm` (I don't even know what language it is) but I located your adding of the localized label and that looked appropriate.

Other remarks:
* it's really a pity to have some `int`s (not even all `int`s) becoming `String`s in json - I hope you can roll that back
* what happens if we put several comma-delimited languages in `tags_lc`?

olivier5741 (commented):
This gives a strange result : https://world.openfoodfacts.dev/api/v3/product/3335880004112?fields=packagings&tags_lc=fr,en

stephanegigandet:
> * it's really a pity to have some `int`s (not even all `int`s) becoming `String`s in json - I hope you can roll that back

I think that's an issue in the test framework, I still have numbers when I access the API.

stephanegigandet:
> * what happens if we put several comma-delimited languages in `tags_lc`?

It's not supported. Adding something like that could cause errors down the line. e.g. if you ask for "fr", you can get "Plastique", and then write it again, by specifying the same tags_lc=fr. If we try to support something like tags_lc=fr,nl,en , then it becomes very messy: we have to specify in the response which it is, and then in an edit interface, we can have a mix of languages, which the user could change. I don't think it's worth it.

tests/integration/api_v2_product_read.t
```diff
@@ -92,6 +94,20 @@ my $tests_ref = [
...
 	},
+	{
+		test_case => 'get-packagings',
+		method => 'GET',
+		path => '/api/v2/product/200000000034',
+		query_string => '?fields=packagings',
+		expected_status_code => 200,
+	},
+	{
+		test_case => 'get-packagings-fr',
```
alexgarel, line 105:
```suggestion
        # test that on v2 tags_lc parameter have no effects
        # packaging part should be exactly the same as get-packagings
		test_case => 'get-packagings-fr',
```

alexgarel (approved):
LGTM

PR #7753, opened 2022-11-22
taxonomy: Junkets

moon-rabbitOFF:
I didn't keep some of the previous name as a synonym because the categorization was completely wrong, and we have to change everything anyway.

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -47960,7 +47960,7 @@ zh:奶类甜品
 food_groups:en: en:Dairy desserts
 pnns_group_2:en:Dairy desserts
 
-<en:Milks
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Thanks!

taxonomies/categories.txt
```diff
@@ -47982,6 +47960,35 @@ zh:奶类甜品
 food_groups:en: en:Dairy desserts
 pnns_group_2:en:Dairy desserts
 
+<en:Milks
+en:Junkets
```
stephanegigandet, line 47964: I didn't know about junkets, what about more litteral translations like "curded milk" or "soured milk"?

PR #7755, opened 2022-11-24
feat: Editing packaging components on web site

stephanegigandet:
A simple table in the product edit form in order to enter packaging components information.

image

Mobile:

image

It's not very user friendly, but it's a first step in order to be able to enter some data.

Also added new tags to the misc facet so that we can track packaging data:

- packagings-complete 
- packagings-not-complete 
- packagings-empty 
- packagings-not-empty 
- packagings-not-empty-but-not-complete 

Will be done later in separate PRs:

- adding packagings_complete to the v3 product WRITE API
- better default ordering for the shape / material options (in particular moving all the numbered options for materials to the bottom of the list)
- ideally a more specific ordering of shape / material options that could depend on product category, previous selection for the component etc. (e.g. if it's beer, we put bottle at the top of shape, and if the shape is a bottle, we put glass on top of materials)
- suggestions to quickly add components based on product category and/or other packaging components


stephanegigandet:
Deployed on https://world.openfoodfacts.dev/

lib/ProductOpener/Packaging.pm
```diff
@@ -475,6 +473,11 @@ sub get_checked_and_taxonomized_packaging_component_data ($tags_lc, $input_packa
 		}
 	}
 
+	# If we don't have data at all, return undef
+	if (not $has_data) {
+		return;
+	}
+
 	return $packaging_ref;
 }
```
alexgarel, line 482: it's nice you fix bugs, but shan't we add some unit test to this function ?

scss/_off.scss
```diff
@@ -1247,4 +1247,49 @@ display: inline-block;
   flex-grow:1;
 }
 
+#packagings_components_edit input {
```
alexgarel, line 1250:
```suggestion
/* Packaging widget */
/* **************** */
#packagings_components_edit input {
```

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+        <div class="large-1 columns">[% lang("packaging_quantity_per_unit") %]</div>
+    </div>
+
+[% FOREACH packaging IN product.packagings %]
+    [% SET packaging_max = loop.index() %]
+    [% SET packaging_id = "packaging_" _ loop.index() %]
+    <div class="row packagings_components_edit_row" id="[% packaging_id %]_row">
+        <div class="small-12 large-1 columns">
+
+            <label class="hide-for-large-up" for="[% packaging_id %]_number_of_units">[% lang("packaging_number_of_units") %]</label>
```
alexgarel, line 31:
```suggestion
            <!-- label for small displays -->
            <label class="hide-for-large-up" for="[% packaging_id %]_number_of_units">[% lang("packaging_number_of_units") %]</label>
```

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+} 
+[% END %]
+
+\$("#[% packaging_id %]_delete").click(function(event) {
+    event.stopPropagation();
+    event.preventDefault();
+    \$("#[% packaging_id %]_row").remove();
+});
+</initjs>
+    [% END %]
```
alexgarel, line 95: It's hard to read without indentation… can't we indent ? (so that we see we are still in the for loop)

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+<initjs>
+\$("#[% select2_id %]").select2({
+    placeholder: "[% lang("packaging_$property") %]",
+    data: [% taxonomy %]_select2_options,
+    allowClear:true,
+    tags: true
+});
+
+[% IF packaging.$property.defined %]
+// Set the value, creating a new option if necessary
```
alexgarel, line 77:
```suggestion
// Set the value, creating a new option if necessary (for non taxonomy values)
```

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+        <div class="large-1 columns">[% lang("packaging_number_of_units") %]</div>
+        <div class="large-3 columns">[% lang("packaging_shape") %]</div>
+        <div class="large-3 columns">[% lang("packaging_material") %]</div>
+        <div class="large-3 columns">[% lang("packaging_recycling") %]</div>
+        <div class="large-1 columns">[% lang("packaging_weight") %] (g)</div>
+        <div class="large-1 columns">[% lang("packaging_quantity_per_unit") %]</div>
+    </div>
+
+[% FOREACH packaging IN product.packagings %]
+    [% SET packaging_max = loop.index() %]
```
alexgarel, line 26: why not `loop.max()` ?

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+            [% SET select2_id = packaging_id _ "_$property" %]
+
+        <div class="small-12 large-3 columns">
+            <label class="hide-for-large-up" for="[% select2_id %]">[% lang("packaging_$property") %]</label>
+            <select id="[% select2_id %]" name="[% select2_id %]" class="form-control [% taxonomy %]_select2">
+                <option></option>
+            </select>
+        </div>
+
+    [% IF packaging_max > 0 %]
```
alexgarel, line 67: This condition is obscure to me ! Is it to test if we have a value ?

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+            <input id="[% packaging_id %]_quantity_per_unit" name="[% packaging_id %]_quantity_per_unit" type="text" value="[% packaging.quantity_per_unit %]" %]">
+        </div>        
+
+    </div>
+[% END %]
+
+<input type="hidden" name="packaging_max" value="[% packaging_max %]">
+
+<initjs>
+var packaging_max = [% packaging_max %];
```
alexgarel, line 115: I prefer to track those kind of long range dependencies as they may be harmful in the future !

```suggestion
// defined by the FOREACH loop above
var packaging_max = [% packaging_max %];
```

But why not instead:

```suggestion
var packaging_max = [% product.packagings.length %];
```
(see [length virtual method](https://metacpan.org/dist/Template-Toolkit/view/lib/Template/Manual/VMethods.pod#length)

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+
+function add_packaging() {
```
alexgarel, line 117:
```suggestion
/**
 * Add a new row of packaging data
 */
function add_packaging() {
```

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+
+    \$(this).unbind("select2:select");
+    \$("#packaging_" + packaging_max + "_delete").show();
+
+    packaging_max = packaging_max + 1;
+    \$("#packaging_max").val(packaging_max);
+
+    var new_packaging_id = "packaging_" + packaging_max;
+    var new_packaging_html = \$("#packaging_0_row")[0].outerHTML;
```
alexgarel, line 126:
```suggestion
    // row 0 is our template row
    var new_packaging_html = \$("#packaging_0_row")[0].outerHTML;
```

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
+<!-- start templates/[% template.name %] -->
+
+<initjs>
+var packaging_shapes_select2_options = [% generate_select2_options_for_taxonomy_to_json("packaging_shapes") %];
+var packaging_materials_select2_options = [% generate_select2_options_for_taxonomy_to_json("packaging_materials") %];
+var packaging_recycling_select2_options = [% generate_select2_options_for_taxonomy_to_json("packaging_recycling") %];
+
+var component = 0;
```
alexgarel, line 8: what is this for ?

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
alexgarel: ok maybe it's more clear if you write:
```suggestion
    [% SET packaging_index = loop.index() %]
```

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
alexgarel:
```suggestion
    [%# row 0 is a template for new row, we do not populate it %]
    [% IF packaging_max > 0 %]
```

But in this case does it means you have an empty first row in product.packagings ?

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+
+    \$(this).unbind("select2:select");
+    \$("#packaging_" + packaging_max + "_delete").show();
+
+    packaging_max = packaging_max + 1;
+    \$("#packaging_max").val(packaging_max);
+
+    var new_packaging_id = "packaging_" + packaging_max;
+    var new_packaging_html = \$("#packaging_0_row")[0].outerHTML;
+    new_packaging_html = new_packaging_html.replaceAll("packaging_0", new_packaging_id);
```
alexgarel, line 127: smart :-)

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+    // Append it to the select
+    \$("#[% select2_id %]").append(newOption).trigger('change');
+} 
+[% END %]
+
+\$("#[% packaging_id %]_delete").click(function(event) {
+    event.stopPropagation();
+    event.preventDefault();
+    \$("#[% packaging_id %]_row").remove();
+});
```
alexgarel, line 93: we could also have this function only once by using a selector on a class, and getting the id of event target.

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+        event.preventDefault();
+        \$("#" + new_packaging_id + "_row").remove();
+    }).hide();    
+
+    \$("#" + new_packaging_id + "_shape").on('select2:select', add_packaging);
+}
+
+\$("#[% packaging_id %]_shape").on('select2:select', add_packaging);
+
+\$("#[% packaging_id %]_delete").hide()
```
alexgarel, line 161:
```suggestion
[%# packaging_id contains our last row %]
// last row triggers addition of a new row
\$("#[% packaging_id %]_shape").on('select2:select', add_packaging);
// disable removal of last row
\$("#[% packaging_id %]_delete").hide()
```

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+
+<input type="hidden" name="packaging_max" value="[% packaging_max %]">
+
+<initjs>
+var packaging_max = [% packaging_max %];
+
+function add_packaging() {
+
+    \$(this).unbind("select2:select");
+    \$("#packaging_" + packaging_max + "_delete").show();
```
alexgarel, line 120:
```suggestion
    // row that triggers new row addition is no more the last row
    // remove events and let it be removed
    \$(this).unbind("select2:select");
    \$("#packaging_" + packaging_max + "_delete").show();
```

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
```diff
@@ -0,0 +1,182 @@
...
+
+    \$("#" + new_packaging_id + "_shape").on('select2:select', add_packaging);
+}
+
+\$("#[% packaging_id %]_shape").on('select2:select', add_packaging);
+
+\$("#[% packaging_id %]_delete").hide()
+
+
+\$(document).on("focus", ".select2", function (e) {
```
alexgarel, line 164:
```suggestion
/**
 * select2 auto-open options
 */
\$(document).on("focus", ".select2", function (e) {
```

alexgarel (commented):
Here is my review.

IMO, we should add unit test to analyze_and_combine_packaging_data, it did not work as expected for me.

Eg, this one does seems to get merged correctly:
Capture d’écran du 2022-11-25 18-13-39

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
stephanegigandet: product_multilingual.pl adds an empty row at the beginning and at the end:

```
	# Add an empty packaging element to the form, that will be hidden and duplicated when the user adds new packaging items,
	# and another empty packaging element at the end
	if (not defined $product_ref->{packagings}) {
		$product_ref->{packagings} = [];
	}
	my $number_of_packaging_components = scalar @{$product_ref->{packagings}};

	unshift(@{$product_ref->{packagings}}, {});
	push(@{$product_ref->{packagings}}, {});
```

As it's the first empty row, it's hidden, so there's no point in creating select2 on it.

Then when product_multilingual.pl reads back the values, it ignores the empty rows:

```
	# And then we add each packaging component
	for (my $packaging_id = 1; $packaging_id <= single_param("packaging_max"); $packaging_id++) {
```

I will add a clearer comment.

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
stephanegigandet: I don't remember, it's not used, I guess I changed my mind and forgot it.

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
alexgarel: yes we need a good comment :-)

stephanegigandet:
> Eg, this one does seems to get merged correctly:

@alexgarel there are 2 issues for this one:

- your text is in French, but the field is packaging texts in English, so only "film" is recognized.
- I did not put "plastiques" as a synonym of "plastique", so "boites plastiques" does not work, I will add it to the taxonomy.
image

lib/ProductOpener/Packaging.pm
stephanegigandet: analyze_and_combine_packaging_data() has unit tests in [tests/t/packagings.t](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/tests/unit/packaging.t)

I will add a few more when nothing is recognized.

templates/web/pages/product_edit/product_edit_form_packagings.tt.html
stephanegigandet: sure, I'll indent

stephanegigandet:
Hi @alexgarel and @raphael0202 , thanks a lot for the tests, review and suggestions. I believe I addressed all the points you raised.

The new version is currently live on .dev : https://world-fr.openfoodfacts.dev/cgi/product.pl?type=edit&code=7771608000022

alexgarel (commented):
Cool, it works fine.

There is something we should however improve in the future:
The text field should be either erased, or regenerated after packaging update.

Currently if I remove a row in packaging (say tap) but do not pay attention and the element is present in text (1 tap), it won't be removed, and this really is counter intuitive to users.

Regenerating text comes with the advantage that it gives a clear template and is easy to modify. But the drawback is losing original input.

stephanegigandet:
> The text field should be either erased, or regenerated after packaging update.

The text often contains the text of the recycling instructions on the package. I would prefer to keep it.

PR #7759, opened 2022-11-26
taxonomy: some Spanish and Catalan translations

duhow:


taxonomies/ingredients.txt
```diff
@@ -76972,6 +76987,11 @@ pl:pasteryzowane żółtko jaja kurzego
 pt:gema de ovo pasteurizado
 # ingredient/fr:jaune-d’oeuf-pasteurisé has 81 products in 4 languages @2018-12-18
 
+<en:egg yolk
+en:pasteurised egg, pasteurized egg
```
stephanegigandet, line 76991: We should not create new entries with "pasteurised something". Instead, you can add "pasteuritzat" and "pasteurizado" to the ingredients_processing.txt taxonomy, and that will handle all "pasteurized something" in Catalan and Spanish.

stephanegigandet (changes requested):
translations for "pasteurized" something should go into ingredients_processing instead

stephanegigandet (approved):
Thank you!

PR #7767, opened 2022-11-28
docs: remove excess examples from the open api

Roxie-32:
I removed excess examples from everything except product schema. The examples in the product schema is quite bulky and confusing. I need help with it, please.

### DIFF
--- a/docs/reference/api.yml
+++ b/docs/reference/api.yml
@@ -108,7 +108,6 @@ paths:
           multipart/form-data:
             schema:
               $ref: ./requestBodies/add_photo_to_existing_product.yaml
-            examples: {}
         description: ''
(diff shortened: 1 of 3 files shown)


docs/reference/api.yml
```diff
@@ -192,20 +191,6 @@ paths:
           multipart/form-data:
             schema:
               $ref: ./requestBodies/add_or_edit_a_product.yaml
-            examples:
-              example-1:
-                value:
-                  code: '0074570036004'
-                  user_id: myusername
-                  password: mypassword
```
alexgarel, line 200: we could indeed write on or two simple example here (but you are rightt to remove this one which does not make sense)

alexgarel (approved):
Let's clean.

PR #7771, opened 2022-11-29
fix: fields=all returns customized field values

stephanegigandet:
fields=all had a difficult to understand behaviour: it returned all raw fields, without customization. Which meant that asking for packagings in v3 would return a different packagings field that asking for all in v3.

This PR:
- adds a new fields=raw option, to get whatever is stored in MongoDB
- makes values from fields=all customized
- add the possibility to request all + some fields that are computed on requests (e.g. fields=all,attribute_groups or fields=updated,knowledge_panels
- adds corresponding tests


docs/reference/api-v3.yml
```diff
@@ -38,6 +38,26 @@ paths:
...
+        - schema:
+            type: string
+          in: query
+          name: tags_lc
+          description: '2 letter language code to request names of tags in a specific language.  For READ requets: if passed, all taxonomized tags of the response will include a lc_name property with the translation in the requested language, if available. Otherwise, the property value will contain the name in the original language, prefixed by the 2 language code and a colon.'
+        - schema:
+            type: string
+          in: query
+          name: fields
+          description: 'Comma separated list of fields requested in the response. Special values: "none": returns no fields, "raw": returns all fields as stored internally in the database, "all": returns all fields except generated fields that need to be explicitly requested such as "knowledge_panels". Defaults to "all" for READ requests. The "all" value can also be combined with fields like "attribute_groups" and "knowledge_panels".'
```
alexgarel, line 60: You should not hesitate to use markdown for such entries.


```suggestion
          description: |-
            Comma separated list of fields requested in the response. 
            
            Special values: 
            * "none": returns no fields
            * "raw": returns all fields as stored internally in the database
            * "all": returns all fields except generated fields that need to be explicitly requested such as "knowledge_panels". 
            
            Defaults to "all" for READ requests. The "all" value can also be combined with fields like "attribute_groups" and "knowledge_panels".'
```

docs/reference/api-v3.yml
```diff
@@ -58,7 +81,7 @@ paths:
         description: Barcode of the product
     patch:
       summary: WRITE Product - Create or update product (API V3 - Implementation in progress)
-      operationId: post-api-v3-product-barcode
+      operationId: patch-api-v3-product-barcode
       responses:
         '200':
           description: |-
```
alexgarel, line 87: Why is the explanation below not in the content schema with an entry "fields" ?

docs/reference/api-v3.yml
alexgarel: we should also update the v2 api documentation ?

tests/integration/expected_test_results/search_v1/search-tags-categories-without-ingredients-from-palm-oil.json
```diff
@@ -74,7 +74,7 @@
             "2xxxxxxxxxxx"
          ],
          "complete" : 0,
-         "completeness" : "0.5",
+         "completeness" : 0.5,
```
alexgarel, line 77: Do you understand why you have this flip-flap between string and raw numbers ?

alexgarel (approved):
LGTM. It's a good change.

I only have small suggestions

tests/integration/expected_test_results/search_v1/search-tags-categories-without-ingredients-from-palm-oil.json
stephanegigandet: @alexgarel it's an issue with JSON:PP which select the type (string or number) based on what the last use of the scalar was. I tried to find out what exactly was using those scalars in a string context (like a debug dump function), but I failed. 

https://docs.mojolicious.org/JSON/PP#PERL---JSON

docs/reference/api-v3.yml
stephanegigandet: I didn't know we could use markdown for descriptions :)

docs/reference/api-v3.yml
stephanegigandet: There is a fields entry in the content schema for the request (which is above), below is the content schema of the response (which does not have a fields entry, but is affected by the fields parameter in the query)

PR #7775, opened 2022-11-30
feat: add packaging parts fields to import template

stephanegigandet:
This is to add fields related to packaging parts in the Excel template that is generated on the pro platform by https://fr.pro.openfoodfacts.org/cgi/import_file_upload.pl

I put a copy of the generated file for French here: https://world.openfoodfacts.org/files/openfoodfacts_import_packagings_fr.xlsx

image

Note that the pro platform is still running an old version of the code with the old design (without knowledge panels). So we may have to wait a bit to deploy this (but we can link to the generated file).


alexgarel (approved):
Great !

PR #7776, opened 2022-11-30
docs: Add more body parameters to Edit A Product

Roxie-32:
Add more options to editing a  product like brands , categories etc.


.github/labeler.yml
```diff
@@ -352,6 +352,15 @@ knowledge-panels:
...
 - tests/integration/expected_test_results/api_v3_product_read/get-knowledge-panels-fr.json
+- docs/reference/schemas/knowledge_panels/panels.yaml
+- docs/reference/schemas/knowledge_panels/panel.yaml
+- docs/reference/schemas/knowledge_panels/elements/element.yaml
+- docs/reference/schemas/knowledge_panels/elements/image_element.yaml
+- docs/reference/schemas/knowledge_panels/elements/panel_element.yaml
+- docs/reference/schemas/knowledge_panels/elements/panel_group_element.yaml
+- docs/reference/schemas/knowledge_panels/elements/table_element.yaml
+- docs/reference/schemas/knowledge_panels/elements/text_element.yaml
+- docs/reference/schemas/knowledge_panels/elements/title_element.yaml
```
alexgarel, line 363: It's better to write this [using path pattern](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)

```suggestion
- docs/reference/schemas/knowledge_panels/**
```

.github/labeler.yml
```diff
@@ -426,6 +435,16 @@ Packaging:
...
+- docs/reference/schemas/packagings/material-write.yaml
+- docs/reference/schemas/packagings/material.yaml
+- docs/reference/schemas/packagings/packaging_component-write.yaml
+- docs/reference/schemas/packagings/packaging_component.yaml
+- docs/reference/schemas/packagings/packagings-write.yaml
+- docs/reference/schemas/packagings/packagings.yaml
+- docs/reference/schemas/packagings/recycling-write.yaml
+- docs/reference/schemas/packagings/recycling.yaml
+- docs/reference/schemas/packagings/shape-write.yaml
+- docs/reference/schemas/packagings/shape.yaml
```
alexgarel, line 447: Same as above:


```suggestion
- docs/reference/schemas/packagings/**
```

docs/reference/requestBodies/add_or_edit_a_product.yaml
```diff
@@ -12,8 +12,24 @@ properties:
...
+    description: Product specific labels.
+    example: Kosher
+  categories:
+    type: string
+    description: The category of the product.
+    example: Desserts
+  packaging:
+    type: string
+    description: Packaging type, format, material.
+    example: Frozen
```
alexgarel, line 30: we should also cite that there is an API v3 which add much more structured data for this field and should be privileged. with a link to https://openfoodfacts.github.io/openfoodfacts-server/reference/api-v3/#operation/post-api-v3-product-barcode

alexgarel (changes requested):
Thanks for continuing the work.
Some minor changes requested.

docs/reference/requestBodies/add_or_edit_a_product.yaml
Roxie-32: fixed !

docs/reference/requestBodies/add_or_edit_a_product.yaml
```diff
@@ -12,8 +12,24 @@ properties:
     type: string
     description: A valid corresponding password.
     example: mypassword
+  brands:
+    type: string
+    description: The brand of the product.
+    example: Häagen-Dazs
```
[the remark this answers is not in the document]
Roxie-32: @alexgarel whats the way forward with this 😂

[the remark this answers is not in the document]
stephanegigandet: yes, fields like brands, categories, labels etc. are comma separated lists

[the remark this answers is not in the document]
alexgarel: @Roxie-32 to try to give you a handle on that. Sometimes an API want's its user to be able to pass array parameter like: `/my-api?label=organic&label=fair-trade&label=gluten-free` and get `["organic", "fair-trade", "gluten-free"]` as a value for label. (this is perfectly fine with http to repeat same parameter name).
This correspond to 
`{schema: {type: array, items: {type: string}}, style: form, explode: true}`

but if we turn explode to false, this correspond to an API call of type:

`/my-api?label=organic,fair-trade,gluten-free`

OpenAPI supports many way of specifying complex parameters (arrays, objects, etc.) using path, parameters, headers, etc. which are less expressive. Those are conventions that you already find in many projects (that's why OpenAPI supports them).

Roxie-32:
@alexgarel I have updated for label and categories, cc: @stephanegigandet ready to merge

alexgarel (approved):
LGTM, thanks.

PR #7780, opened 2022-12-01
fix:  make home first in the nav bar

Roxie-32:
Made Home first item in the navbar

### DIFF
--- a/docs/.pages
+++ b/docs/.pages
@@ -1,9 +1,10 @@
 # customize the navigation menu
 # see https://github.com/lukasgeiter/mkdocs-awesome-pages-plugin
 nav:
-  - ...
+  - index.md
   - introduction
   - tutorials
   - how-to-guides
   - explanations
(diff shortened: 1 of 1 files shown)


docs/.pages
```diff
@@ -1,9 +1,9 @@
 # customize the navigation menu
 # see https://github.com/lukasgeiter/mkdocs-awesome-pages-plugin
 nav:
+  - ...
```
alexgarel, line 4: Nope, we don't want to do it that way as `...` is for items not cataloged here.

Home correspond to the README.md that we move to index.md at generation time. So we should instead just cite it:

```yaml
nav:
  - index.md
  - introduction
  - tutorials
  - how-to-guides
  - explanations
  - reference
  - ...

Roxie-32:
@alexgarel its ready for review

alexgarel (approved):

PR #7791, opened 2022-12-03
Pl ingredients

moon-rabbitOFF:


taxonomies/ingredients.txt
```diff
@@ -40782,6 +40788,14 @@ de:Ananassirup
 fr:sirop d'ananas
 nl:ananassiroop
 
+<en:pineapple
```
stephanegigandet, line 40791: We don't need to add entries for "dried something" or "freeze-dried something" as we have "dried" and "freeze-dried" in the ingredients_processing taxonomy.

So we already recognize "ananas séché" and "ananas lyophilisé" : https://fr.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=ananas+s%C3%A9ch%C3%A9%2C+ananas+lyophilis%C3%A9&type=add&action=process&submit=Envoyer

taxonomies/ingredients.txt
```diff
@@ -45383,6 +45397,11 @@ fi:juurikaskuitu
 sv:betfiber
 #  fr:fibres-de-betterave has 43 products in 1 language @2018-10-22
 
+<en:beetroot
+en:dried beetroot
+fr:betterave déshydratée
+pl:suszony burak
```
stephanegigandet, line 45403: Can you remove the "dried beetroot" entry, and instead add "suszony" in the ingredients_processing taxonomy?

stephanegigandet (changes requested):
There is a new mechanism to handle things like "dried something", we should not add them to the ingredients taxonomy, and instead we can enrich the ingredients_processing.txt taxonomy to add translations of "dried" etc. in other languages.

taxonomies/ingredients.txt
moon-rabbitOFF: Ah cool! But it isn't working yet? I added those ingredients because they were not recognized

stephanegigandet: It should be working, you can test with this tool: https://fr.openfoodfacts.org/cgi/test_ingredients_analysis.pl
We may not have reprocessed yet the ingredients of the products you saw.

benbenben2:
@stephanegigandet

@moon-rabbitOFF is right, it does not work for all languages. Polish, for example: https://pl.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=liofilizowane+ananasy&type=add&action=process&submit=Envoyer

@moon-rabbitOFF, you have to edit the file lib/ProductOpener/ingredients.pm.
Under the comment
`# English, French etc. match before or after the ingredient, require a space`
you add a new line:
`											or ($product_lc eq 'pl')`

That can be tackled in a different pull request (you will have merge conflict if you do it from this actual branch).

For this current pull request, make sure you keep only ingredients in this ingredients.txt file and processing method in ingredients_processing.txt and that should be fine.

benbenben2:
@stephanegigandet, please review the last changes

Thank you @Teiron for your pull request related to this one

taxonomies/ingredients.txt
```diff
@@ -56199,12 +56202,12 @@ wikidata:en:Q1076906
 nutriscore_fruits_vegetables_nuts:en:no
 
 <en:sunflower seed
-en:shelled sunflower seeds
-de:Sonnenblumenkerne geschält
-fi:kuoreton auringonkukansiemen, kuorettomat auringonkukansiemenet
-fr:graines de tournesol décortiquées
-it:semi di girasole sbucciati
-pl:nasiona słonecznika łuskane, łuskane nasiona słonecznika
+# en:shelled sunflower seeds
```
stephanegigandet, line 56205: If everything is commented out, we need to remove the <en:sunflower seed on top

stephanegigandet (approved):
Looks good, thank you!

taxonomies/ingredients.txt
benbenben2: <en:sunflower seed
is now commented as well (not removed, if future contributor see it commented that will be a hint to understand that they are handled using processing taxonomy. Whereas if we remove the whole block there is risk that a future contributor add it in the taxonomy (although it should be noticed during code review))

stephanegigandet: ah ok my bad, I didn't look correctly

PR #7802, opened 2022-12-06
taxonomy: Lithuanian categories, allergens, ingredients

DanieliusAsm:
### What
 - Translated Buillons, lasagne, Pizzas
 - Added, translated allergens
 - Translated a little bit of ingredients


stephanegigandet (approved):
Thank you!

PR #7809, opened 2022-12-06
fix: chocolate eggs

moon-rabbitOFF:


taxonomies/categories.txt
```diff
@@ -34259,8 +34259,8 @@ wikidata:en:Q2164820
 agribalyse_proxy_food_code:en:31004
 #all agribalyse milk chocolate categories have the same CO2 score
 
-
-<en:Chocolates
+#Do not put this category under chocolates since "filled chocolate eggs" are bonbons, not chocolate
```
stephanegigandet, line 34262: So shouldn't it be under https://world.openfoodfacts.org/category/chocolate-candies or bonbons?

stephanegigandet: Ah ok, that's tricky.

stephanegigandet (approved):
Thanks!

PR #7813, opened 2022-12-08
fix: add authentification through JSON body for API v3 WRITE requests

stephanegigandet:
This PR adds authentification for API v3 WRITE requests, by providing user_id and password fields in the JSON body of the request (it was documented in OpenAPI, but not implemented).

Also  added tests for authentification for API READ + WRITE for v2 + v3


hangy:
I kinda missed the discussion: Why do we want the password in the JSON body instead of HTTP Basic Auth or a token (ORY Hydra+Kratos)? HTTP bodies might be logged somewhere, and using HTTP Headers for auth is a more standard behaviour.

OpenAPI has built-in support for basic auth, which is often also supported by design tools (ie. Swagger): https://swagger.io/docs/specification/authentication/basic-authentication/

stephanegigandet:
> I kinda missed the discussion: Why do we want the password in the JSON body instead of HTTP Basic Auth or a token (ORY Hydra+Kratos)? HTTP bodies might be logged somewhere, and using HTTP Headers for auth is a more standard behaviour.
> 
> OpenAPI has built-in support for basic auth, which is often also supported by design tools (ie. Swagger): https://swagger.io/docs/specification/authentication/basic-authentication/

Hi @hangy. Thanks for the feedback. We will use HTTP basic auth soon, this is a temporary solution because we need something working as soon as possible as we need the mobile app to support packagings before an event that happens on January 12th.

There is a discussion about it here: https://github.com/openfoodfacts/openfoodfacts-server/issues/7564#issuecomment-1281966739

lib/ProductOpener/Display.pm
```diff
@@ -531,24 +531,34 @@ $lc : language code
...
+
+
 =head3 Return value
 
 Reference to request object.
 
 =cut
 
-sub init_request() {
+sub init_request ($request_ref = {}) {
```
alexgarel, line 550: In python if you use a dict as a default parameter, it will be shared between every call.
(so you prefer to initialize with "undefined" and then if request_ref is undefined, you create a new dict).

But this seems to work in perl (I've done a test !)

alexgarel (approved):
All good !

PR #7822, opened 2022-12-09
feat: Ventilate issue to more GitHub projects

teolemon:
### What
feat: Ventilate issue to more GitHub projects


VaiTon:
@teolemon maybe performance-wise it would be better to run multiple steps in the same job, so that we don't initialize a vm each time

stephanegigandet (approved):

PR #7824, opened 2022-12-09
taxonomy: add Polish translations for caramel and safflower

Teiron:
small additions in polish ingredients taxonomy


stephanegigandet (approved):
Thanks!

PR #7828, opened 2022-12-10
feat: Hunger Games brand link on the product page when a brand is defined

teolemon:
### What

- Hunger Games brand link on the product page when a brand is defined
- Emulate a nice Power User Script feature

### Screenshot of the Power User Script version being emulated
image


templates/web/common/includes/display_field.tt.html
```diff
@@ -1,8 +1,20 @@
...
+    <sup>[<a href="https://hunger.openfoodfacts.org/questions?value_tag=[% field %]&amp;type=brand" title="Help with this brand on Hunger Games">Help with this brand on Hunger Games</a>]</sup>
+</p>
+[% ELSE %]
 
 <p id="field_[% field %]">
     <span class="field">[% name %][% sep %]: </span>
     <span class="field_value" id="field_[% field %]_value">[% value %]</span>
 </p>
 
+[% ENDIF %]
```
teolemon, line 16:
```suggestion
[% END %]
```

stephanegigandet:
I think something like this needs to be discussed more broadly, if we add power user features to the website, then the website becomes more complex and difficult to understand for newcomers. There are certainly some solutions to consider, like having a contributor mode that can be toggled. I suggest we open an issue for it.

templates/web/common/includes/display_field.tt.html
```diff
@@ -1,8 +1,20 @@
 <!-- start templates/[% template.name %] -->
+<!-- Special handling to show a Hunger Game promo -->
+[% IF FIELD_TYPE == 'BRAND' %]
```
teolemon, line 3:
```suggestion
[% IF FIELD == 'BRAND' %]
```

teolemon:
Based on feedback by @stephanegigandet , I will restrict that to moderators

templates/web/common/includes/display_field.tt.html
```diff
@@ -1,8 +1,20 @@
 <!-- start templates/[% template.name %] -->
+<!-- Special handling to show a Hunger Game promo -->
+[% IF FIELD == 'BRAND' %]
```
stephanegigandet, line 3:
```suggestion
[% IF field == 'brands' %]
```

templates/web/common/includes/display_field.tt.html
```diff
@@ -1,8 +1,20 @@
...
+<p id="field_[% field %]">
+    <span class="field">[% name %][% sep %]: </span>
+    <span class="field_value" id="field_[% field %]_value">[% value %]</span>
+    <sup>[<a href="https://hunger.openfoodfacts.org/questions?value_tag=[% field %]&amp;type=brand" title="Help with this brand on Hunger Games">Help with this brand on Hunger Games</a>]</sup>
```
stephanegigandet, line 7: It's not as simple, [% value %] contains a comma separated list of links, so we can't construct an hunger game URL like that.

e.g. if we have 2 brands, <span class="field_value" id="field_[% field %]_value">[% value %]</span> will generate this:

<span class="field_value" id="field_brands_value"><a itemprop="brand" href="/marque/activia">Activia</a>, <a itemprop="brand" href="/marque/danone">Danone</a></span>

hangy:
Could be a candidate for [`openfoodfacts-webcomponents`](https://github.com/openfoodfacts/openfoodfacts-webcomponents/), like we already do for the products page now

PR #7833, opened 2022-12-11
taxonomy: reviewed some beers and wines in Croatia

benbenben2:
added few varietals for wines
ingredients for beers
categories


stephanegigandet (approved):
Thanks!

Issue #6526, opened 2022-03-30
Protect data from producer should be the default setting when creating an organization


alexgarel:
### Describe the bug

We have a "Protect the data that is provided by the organization." settings which is not set by default, while we insure producers that their data is protected for one year. 

### To Reproduce

1. Create a new producer account for a new org
2. Go to pro platform as admin
3. Select the new org
4. Edit the profile
5. The setting is not set

### Expected behavior

It should be set by default, at org creation

Having it as a setting might be useful in the case where data published by a org reveals to be fake or some other scenario (moral contract is broken). But this is a rare case, so it should be checked by default and unchecked when needed.

- [ ] change the org creation code to have it set
- [ ] write a migrate script for existing org and run it

### Type of device

Browser

alexgarel:
We still have to run the script in staging (to test it) and in production.

manon-corneille:
Am I understanding that this works now ? :) 
If yes, should we close this issue too ? →  https://github.com/openfoodfacts/openfoodfacts-server/issues/2086

PR #7845, opened 2022-12-15
fix: org should have data protected by default

alexgarel:
* it is now the default on org creation
* added a migration script

fixes: #6526


stephanegigandet (approved):
Looks good

PR #7851, opened 2022-12-15
docs: some comments in Import.pm

alexgarel:
that I added as I try to understand the code

### DIFF
--- a/lib/ProductOpener/Import.pm
+++ b/lib/ProductOpener/Import.pm
@@ -1130,10 +1130,11 @@ sub import_csv_file ($args_ref) {
 					. $imported_product_ref->{$field . "_if_not_existing"} . "\n";
 				$imported_product_ref->{$field} = $imported_product_ref->{$field . "_if_not_existing"};
 			}
-
+			# if it is a field with a tag behaviour (taxonomized or not)
+			# (see %tags_fields in Tags.pm)
 			if (defined $tags_fields{$field}) {
 				foreach my $subfield (sort keys %{$imported_product_ref}) {
-
+					# empty values are skipped
 					next
 						if ((not defined $imported_product_ref->{$subfield})
 						or ($imported_product_ref->{$subfield} eq ""));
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Import.pm
```diff
@@ -1172,12 +1176,13 @@ sub import_csv_file ($args_ref) {
...
 							if (exists_taxonomy_tag($field, $tagid)) {
 								my $opposite_tagid = get_property($field, $tagid, "opposite:en");
+								# we have an opposite, add it
 								if (defined $opposite_tagid) {
 									$tag_to_add = $opposite_tagid;
 								}
 							}
 						}
-
+						# add the value to eventual existing values
```
stephanegigandet, line 1185: ``
```suggestion
						# If we have a tag to add (because we had a "yes" or a "no" value for a specific tag field), concatenate it to possible pre-existing tags
```

stephanegigandet: éventuel -> possible, not eventual

lib/ProductOpener/Import.pm
```diff
@@ -1419,17 +1436,19 @@ sub import_csv_file ($args_ref) {
...
 				}
+				# Processing non tag field
 				else {
-					# non-tag field
+					# new value replaces the old value
 					my $new_field_value = $imported_product_ref->{$field};
 
 					next if not defined $new_field_value;
-
+					# remove spaces
```
stephanegigandet, line 1445:
```suggestion
					# remove leading and trailing spaces
```

lib/ProductOpener/Import.pm
```diff
@@ -1443,7 +1462,7 @@ sub import_csv_file ($args_ref) {
 						$new_field_value =~ s/litre|litres|liter|liters/l/i;
 						$new_field_value =~ s/kilogramme|kilogrammes|kgs/kg/i;
 					}
-
+					# normalize spaces
```
stephanegigandet, line 1465:
```suggestion
					# remove leading and trailing spaces
```

PR #7853, opened 2022-12-15
Solve issue #7786: Barcode is not visible in the edit mode.

### LINKED ISSUE #7786: NEW DESIGN: The barcode is not visible in edit mode
### What
- NEW DESIGN: The barcode is not visible in edit mode
- example: https://world-fr.openfoodfacts.org/cgi/product.pl?type=edit&code=07370793

### Part of
- #7295

Sakshamgupta90:
### What

Added the barcode value in the product_edit_form_display file, and also updated the code input type from hidden to text.

### Screenshot
Screenshot 2022-12-16 at 12 19 20 AM

- Fixes #7786

### DIFF
--- a/templates/web/pages/product_edit/product_edit_form_display.tt.html
+++ b/templates/web/pages/product_edit/product_edit_form_display.tt.html
@@ -129,6 +129,9 @@
 
                 [% display_tab_product_characteristics %]
 
+                [% lang("barcode") %]  
+                <input type="text"  name="code" value="[% code %]">
+
                 [% FOREACH field IN display_fields_arr %]
                     [% field %]
                 [% END %]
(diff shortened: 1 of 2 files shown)


stephanegigandet (changes requested):
Hi @Sakshamgupta90 
Thanks for the PR.
The issue is not very descriptive, the idea is not to make the barcode an editable field, but just to display it, so that users can easily check that they are editing the right product.

So you just have to display it in this section, before the if moderator :


```
        <section id="misc" class="card fieldset">
            <div class="card-section">

        [% IF moderator %]
            <label for="new_code" id="label_new_code">[% label_new_code %]</label>
            <input type="text" name="new_code" id="new_code" class="text" value="" /><br/>
        [% END %]
```

Sakshamgupta90:
Hi @stephanegigandet, okay I got it. Just one thing to clarify did you mean to display barcode like this?
Screenshot 2022-12-16 at 4 02 40 PM

stephanegigandet:
Hi @Sakshamgupta90 , yes, maybe with a line feed after. Put it in <p> tag maybe.

PR #7856, opened 2022-12-16
feat: allow setting the packagings_complete field through API v3

stephanegigandet:
We have a field called packagings_complete that takes value 0 or 1 to indicate if the packagings field contains all element of the product.
This PR adds it to the API v3 WRITE product.

### DIFF
--- a/lib/ProductOpener/APIProductWrite.pm
+++ b/lib/ProductOpener/APIProductWrite.pm
@@ -51,6 +51,102 @@ use ProductOpener::Products qw/:all/;
 use ProductOpener::API qw/:all/;
 use ProductOpener::Packaging qw/:all/;
 
+=head2 update_field_with_0_or_1_value($request_ref, $product_ref, $field, $value)
+
+Update a field that takes only 0 or 1 as a value (e.g. packagings_complete).
+
+=cut
+
+sub update_field_with_0_or_1_value($request_ref, $product_ref, $field, $value) {
+
+	my $response_ref = $request_ref->{api_response};
+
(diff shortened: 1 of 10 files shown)


alexgarel (approved):
Great !

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -51,6 +51,101 @@ use ProductOpener::Products qw/:all/;
...
+			{
+				message => {id => "invalid_value_must_be_0_or_1"},
+				field => {id => $field},
+				impact => {id => "field_ignored"},
+			}
+		);
+	}
+	else {
+		$product_ref->{$field} = $value + 0;    # add 0 to make sure the value is stored as a number
+	}
```
alexgarel, line 79:
```suggestion
	}
	return 1;
```

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -51,6 +51,101 @@ use ProductOpener::Products qw/:all/;
...
+			# Taxonomize the input packaging component data
+			my $packaging_ref = get_checked_and_taxonomized_packaging_component_data($request_body_ref->{tags_lc},
+				$input_packaging_ref, $response_ref);
+
+			if (defined $packaging_ref) {
+				# Add or combine with the existing packagings components array
+				add_or_combine_packaging_component_data($product_ref, $packaging_ref, $response_ref);
+			}
+		}
+	}
```
alexgarel, line 146:
```suggestion
	}
	return;
```

PR #7904, opened 2022-12-24
taxonomy: add common polish categories and translations

ArturLange:
### What

- Added categories and translations for categories for products commonly sold in Poland


stephanegigandet (approved):
Thank you!

PR #7912, opened 2022-12-28
taxonomy: few more ingredients for HR products

benbenben2:
some more ingredients based on en:Ingredients unknown score above 10 quality issue

mostly in Croatian (few in Norwegian too)


stephanegigandet (approved):
Thank you!

PR #7913, opened 2022-12-29
feat: Show EcoScore attribute panel for world. Fixes #7378

### LINKED ISSUE #7378: Show the Eco-Score on world.off.org on non-logged in users
### Problem description

Most of our traffic comes from world.off.org where we don't show the Eco-Score because we don't know the core country
This is a massive lost opportunity to educate about Eco-Score and show the value of Open Food Facts

### Desired solution

Show the Eco-Score on world.off.org on non-logged in users, possibly with a warning or a country refiner

100%

### Mockups
image

- https://www.figma.com/file/iyFnUGkJTizApXGnQCdh5I/Eco-Score-on-world.off.org?node-id=3%3A11

john-gom:
### What

Show the EcoScore card even when there is no localised EcoScore. Displays a warning to indicate that transportation impact won't be as accurate as it could be

### Screenshot
See Issue #7378 

- Fixes #7378


po/common/common.pot
```diff
@@ -5747,6 +5747,14 @@ msgctxt "ecoscore_warning_international"
 msgid "The Eco-Score was initially developped for France and it is being extended to other European countries. The Eco-Score formula is subject to change as it is regularly improved to make it more precise and better suited to each country."
 msgstr "The Eco-Score was initially developped for France and it is being extended to other European countries. The Eco-Score formula is subject to change as it is regularly improved to make it more precise and better suited to each country."
 
+msgctxt "ecoscore_warning_transportation"
```
stephanegigandet, line 5750: I'm thinking that it would be better to have different messages for the 2 cases: not having a country selected, and having a country selected that we don't support yet.

The messages could be something like:

short: The impact of transportation is not included.

long on world: Select a country in order to include the impact of transportation.
long on a not supported country: The impact of transportation to your country is currently unknown.

john-gom: Do we need to say "...the **full** impact of transportation..." as some impacts are still shown even for world?

stephanegigandet: Sounds good, we can say "the full impact"

stephanegigandet (approved):
Looks good, thank you!

lib/ProductOpener/Attributes.pm
```diff
@@ -632,7 +632,9 @@ sub compute_attribute_ecoscore ($product_ref, $target_lc, $target_cc) {
 			$score = $product_ref->{ecoscore_data}{"scores"}{$target_cc} // 0;
 			$grade = $product_ref->{ecoscore_data}{"grades"}{$target_cc};
 		}
-
+		if (($target_cc eq "world") or (!defined $product_ref->{ecoscore_data}{"scores"}{$target_cc})) {
+			$attribute_ref->{missing} = lang_in_other_lc($target_lc, "ecoscore_warning_transportation_short");
```
stephanegigandet, line 636: In fact we use the "missing" field of attributes only when the attribute cannot be computed (e.g. we don't have the ingredients list, so we don't know if there's an allergen).

When we display attributes in the summary, we don't display warnings (like for the Nutri-Score, we don't say that fibers are missing and were not taken into account). We display the warning in the knowledge panel instead.

Otherwise it makes a very big summary:

image

stephanegigandet: That's where we should add "The full impact of transportion is not included." I think:

image

scss/_product-page.scss
```diff
@@ -75,6 +75,9 @@ $decalTop:48px;
         }
         .attr_text {
             padding-left:0.5rem;
+            .attribute_missing::before {
```
stephanegigandet, line 78: The attribute missing is not really a warning, it's more an explanation of why an attribute is not computed (and thus shown in grey), so we should not display a warning sign there.

john-gom: So hall I remove all the changes to the attribute panel?

lib/ProductOpener/Attributes.pm
john-gom: I'm unclear on what you are saying here - we are already displaying the "Select a country..." warning here. Are you proposing something in addition to that?

stephanegigandet: @john-gom We can leave the "Select a country" warning as-is.

scss/_product-page.scss
stephanegigandet: Yes, I think the attribute panel should stay as it was.

alexgarel:
Kudos @john-gom !

PR #7920, opened 2022-12-30
taxonomy: Add Animal Welfair / Bienestar Animal label

duhow:
### What

Add label for **Animal Welfair** certificate https://www.animalwelfair.com/ as `Bienestar animal`.


stephanegigandet (approved):
Thank you!

PR #7916, opened 2022-12-29
fix: scattered product name in responsive mode on product lists Fixes #7499

Aman-Jamshed:
### What

- Added css to make the product name responsive on product lists

### Related issue(s) and discussion

- Fixes #7499 ]

Issue #7499, opened 2022-10-11
NEW DESIGN - The product name is scattered in responsive mode on product lists


teolemon:
### What

- [ ] The product name is scattered in responsive mode on product lists
### Screenshot
![image](https://user-images.githubusercontent.com/1689815/195129776-c96bbe30-5335-4ec6-b7c0-bcb106448240.png)
### Part of
- #7295

Aman-Jamshed:
Hey @teolemon . 
I would like to work on this issue. Please assign this to me.

Aman-Jamshed:
@alexgarel @teolemon 
Please review this [here](https://drive.google.com/file/d/1xfddcwCij2w7Upo9s5g6OfnaY054iPaA/view?usp=sharing) and let me know if it looks good to you now.

alexgarel:
> @alexgarel @teolemon Please review this [here](https://drive.google.com/file/d/1xfddcwCij2w7Upo9s5g6OfnaY054iPaA/view?usp=sharing) and let me know if it looks good to you now.

Yes seems really good :-)

PR #7928, opened 2023-01-02
fix: responsive product list css

stephanegigandet:
Related to https://github.com/openfoodfacts/openfoodfacts-server/issues/7499
Continuation of https://github.com/openfoodfacts/openfoodfacts-server/pull/7916

This removes CSS code that was duplicated between _off.scss and _product-list.scss

It removes some special behaviour from the old design: on small screen, we had only 1 product card per line, but now we have 2.

When we have multiple cards per line (which is now always the case), we fix some heights so that the attribute icons are vertically aligned on the different product cards.

image

### DIFF
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -813,54 +813,6 @@ html[dir="rtl"] {
   background-color: #aaf;
 }
 
-// Horizontally and vertically center images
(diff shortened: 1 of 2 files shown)


alexgarel (approved):
I trust you on this !

PR #7933, opened 2023-01-03
taxonomy: changes for week 01

aleene:
### What
Some changes for week 1


taxonomies/nutrients.txt
```diff
@@ -3713,7 +3713,7 @@ dv_value:en: 1.7
 unit:en: mg
 
 zz:vitamin-pp
-en:Vitamin B3, Vitamin PP (Niacin), Vitamin B3, Vitamin PP, Niacin
+en:Vitamin B3 (Niacin), Vitamin PP (Niacin), Vitamin B3, Vitamin PP, Niacin
```
moon-rabbitOFF, line 3716: I thought we don't had things between parenthesis in the taxonomy?

aleene: I use it as a solution to fill the pop-up list. Sometimes a package says Niacin and sometimes B3. And I never remember which one goes with what. So now I get always the correct entry proposed. I also started using this approach for the categories, as the suggestions are based on the first letters only. This makes targeting very difficult in dutch (adjectives before the noun).

alexgarel: I confirm, it's ok to have a descriptive label as first entry if you repeat each entry separately.

alexgarel (approved):
LGTM

PR #7937, opened 2023-01-04
chore: column names for packagings

stephanegigandet:
change wording of columns for packagings (suggestions from ADEME)

### DIFF
--- a/po/common/en.po
+++ b/po/common/en.po
@@ -6522,3 +6522,30 @@ msgstr "Invalid user id and password"
 msgctxt "api_message_invalid_value_must_be_0_or_1"
 msgid "Invalid value: must be 0 or 1"
 msgstr "Invalid value: must be 0 or 1"
+
+# Unit = element, not unit of measure
+msgctxt "packaging_number_of_units_description"
+msgid "Enter the number of packaging units of the same shape and material contained in the product."
+msgstr "Enter the number of packaging units of the same shape and material contained in the product."
+
+msgctxt "packaging_shape_description"
+msgid "Enter the shape name listed in the recycling instructions if they are available, or select a shape."
+msgstr "Enter the shape name listed in the recycling instructions if they are available, or select a shape."
+
+msgctxt "packaging_material_description"
+msgid "Enter the specific material if it can be determined (a material code inside a triangle can often be found on packaging parts), or a generic material (for instance plastic or metal) if you are unsure."
+msgstr "Enter the specific material if it can be determined (a material code inside a triangle can often be found on packaging parts), or a generic material (for instance plastic or metal) if you are unsure."
+
+msgctxt "packaging_recycling_description"
+msgid "Enter recycling instructions only if they are listed on the product."
+msgstr "Enter recycling instructions only if they are listed on the product."
+
+# Unit = element, not unit of measure
+msgctxt "packaging_weight_description"
+msgid "Remove any remaining food and wash and dry the packaging part before weighting. If possible, use a scale with 0.1g or 0.01g precision."
+msgstr "Remove any remaining food and wash and dry the packaging part before weighting. If possible, use a scale with 0.1g or 0.01g precision."
+
+# Unit = element, not unit of measure
+msgctxt "packaging_quantity_per_unit_description"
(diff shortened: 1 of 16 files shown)


raphael0202 (approved):

po/common/en.po
```diff
@@ -6503,8 +6503,8 @@ msgstr "Weight of one empty unit"
 
 # Unit = element, not unit of measure
 msgctxt "packaging_quantity_per_unit"
-msgid "Quantity of product contained per unit"
-msgstr "Quantity of product contained per unit"
+msgid "Net quantity of product contained per unit"
```
CharlesNepote, line 6506:
```suggestion
msgid "Net quantity of product contained per unit (g)"
```
Doesn't it make sense to add the unit?

po/common/fr.po
```diff
@@ -6360,17 +6360,17 @@ msgstr ""
...
 msgid "Weight of one empty unit"
-msgstr ""
+msgstr "Poids d'un élément vide"
 
 # Unit = element, not unit of measure
 msgctxt "packaging_quantity_per_unit"
-msgid "Quantity of product contained per unit"
-msgstr ""
+msgid "Net quantity of product contained per unit"
+msgstr "Quantité nette de produit contenue par élément"
```
CharlesNepote, line 6373:
```suggestion
msgstr "Quantité nette de produit contenue par élément (g)"
```

CharlesNepote (changes requested):

po/common/en.po
alexgarel: it's not necessarily in g, it can be ml, it's like for portion, user should add the unit.

po/common/fr.po
alexgarel: See above, it can be ml, pieces, etc.

alexgarel (approved):
LGTM

po/common/fr.po
```diff
@@ -6360,17 +6360,17 @@ msgstr ""
...
 msgctxt "packaging_number_of_units"
 msgid "Number of units"
-msgstr ""
+msgstr "Nombre d'éléments"
 
 # Unit = element, not unit of measure
 msgctxt "packaging_weight"
 msgid "Weight of one empty unit"
-msgstr ""
+msgstr "Poids d'un élément vide"
```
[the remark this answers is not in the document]
stephanegigandet: We already display (g), but it's not in the translated string, it's added directly in the template for the edit form:

`<label class="hide-for-large-up" for="[% packaging_id %]_weight_measured">[% lang("packaging_weight") %] (g)</label>`

po/common/fr.po
stephanegigandet: Right, it's also why we use "Net quantity" instead of "Net weight": it can be a volume, a number of pieces (e.g. eggs) etc.

Also keep in mind this is a column name, it can't be too long, otherwise the display looks ugly.

Display with the changes from this PR:

image

CharlesNepote: Ok fine, but you had to explain it to me. I find "net quantity" a bit fuzzy... I think we should use help bubbles as much as possible. I generally use the `title` attribute, which is good old HTML: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title
But I have just read that `title` isn't very suited for mobile, tablet, keyboard, etc. There is a good resource here:
https://inclusive-components.design/tooltips-toggletips/

po/common/en.po
CharlesNepote: Fine! Where's the explanation for all the users?

po/common/fr.po
stephanegigandet: @CharlesNepote Good point, I filed https://github.com/openfoodfacts/openfoodfacts-server/issues/7939

stephanegigandet:
I think the best solution to address @CharlesNepote 's suggestions is to add text descriptions for each of the field, above the table:

image

On mobile, the table is turned into fields with field names, so it's better to have text explanations above, instead of tooltips or extra text in the column / field names  themselves.

image

CharlesNepote (approved):

PR #7941, opened 2023-01-04
feat: add erythritol as a nutrient

### LINKED ISSUE #7837: Add Erythritol (and others?) in nutrition facts features
Some products do display erythritol amount:
* https://world.openfoodfacts.org/product/0850005716358/brown-sugar
* https://world.openfoodfacts.org/product/0008346874637/slim-fast
* https://world.openfoodfacts.org/product/0008346874804/fat-bomb

We can find many more by listing products containing erythritol as one of their ingredients.

It should be added in the following cases/features:

- [ ] Displayed in the nutrition facts table, in the carbohydrates section
- [ ] complete-as-you-type property in the nutrition facts form
- [ ] exclude from energy computation; In the Wikipedia article about [Food Energy](https://en.wikipedia.org/wiki/Food_energy), we can read: "Some polyols, like [erythritol](https://en.wikipedia.org/wiki/Erythritol), are not digested and should be excluded from the count". See: #7789 

"Erythritol is a polyol, and according to the current rules as provided for in Article 5(1) of Directive 90/496/EEC, its energy would be calculated using the conversion factor for polyols, namely 10 kJ/g (2,4 kcal/g). Using this energy conversion factor would not fully inform the consumer about the reduced energy value of a product achieved by the use of erythritol in its manufacture. The Scientific Committee on Food in its opinion on erythritol, expressed on 5 March 2003, noted that the energy provided by erythritol was less than 0,9 kJ/g (less than 0,2 kcal/g). Therefore it is appropriate to adopt a suitable energy conversion factor for erythritol." Source: [amending Council Directive 90/496/EEC on nutrition labelling for foodstuffs as regards recommended daily allowances, energy conversion factors and definitions](https://eur-lex.europa.eu/LexUriServ/LexUriServ.do?uri=OJ:L:2008:285:0009:0012:EN:PDF).

Erythritol is **displayed at least on the nutrition facts table** of this product: https://world.openfoodfacts.org/product/0850005716358/brown-sugar
But other nutrients might be added that should be excluded from polyols when computing energy.

More than 1900 products has erythritol as an ingredient, and [more than 800](https://mirabelle.openfoodfacts.org/products?sql=select+%0D%0A%28select+count%28*%29+as+products_w_erythritol+from+%5Ball%5D%0D%0Awhere+ingredients_text+like+%27%25erythritol%25%27%29%2C%0D%0A%28select+count%28*%29+as+w_data_quality_errors+from+%5Ball%5D%0D%0Awhere+ingredients_text+like+%27%25erythritol%25%27+and+data_quality_errors_tags+%21%3D+%27%27%29) of them are facing at least a data quality error.

stephanegigandet:
This PR adds erythritol as a nutrient, and adds tests for the nutrients to energy coherency test.

Fixes #7837


alexgarel (approved):
Perfect !

lib/ProductOpener/Food.pm
```diff
@@ -726,37 +726,38 @@ sub mmoll_to_unit ($value, $unit) {
 			'sodium', '!carbohydrates',
 			'-fiber', '--soluble-fiber-',
 			'--insoluble-fiber-', '-sugars',
-			'-added-sugars', '--sucrose-',
+			'--added-sugars', '--sucrose-',
```
CharlesNepote, line 729: `--added-sugars`: as I understand it will be displayed as a sub-category of sugars, isn't it? (Which is good I think.)

Does it have an impact on its name in the CSV export?

CharlesNepote (approved):
Great!

You modified the test but not the algorithm to compute energy: don't you have to subtract erythritol from polyols? Do you intend to do it after?

lib/ProductOpener/Food.pm
stephanegigandet: yes, that makes added sugars a sub category of sugars (as displayed in USDA labels)

stephanegigandet:
> You modified the test but not the algorithm to compute energy: don't you have to subtract erythritol from polyols? Do you intend to do it after?

I had done it before in fact, but it didn't work because erythritol was not in the nutrients taxonomy:

https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/DataQualityFood.pm#L527

```
my %energy_from_nutrients = (
	europe => {
		carbohydrates_minus_polyols => {kj => 17, kcal => 4},
		polyols_minus_erythritol => {kj => 10, kcal => 2.4},
		proteins => {kj => 17, kcal => 4},
		fat => {kj => 37, kcal => 9},
		salatrim => {kj => 25, kcal => 6},    # no corresponding nutrients in nutrient tables?
		alcohol => {kj => 29, kcal => 7},
		organic_acids => {kj => 13, kcal => 3},    # no corresponding nutrients in nutrient tables?
		fiber => {kj => 8, kcal => 2},
		erythritol => {kj => 0, kcal => 0},    # no corresponding nutrients in nutrient tables?
	},
);
```

I'll remove the comment about no corresponding nutrients

CharlesNepote:
> I had done it before in fact, but it didn't work because erythritol was not in the nutrients taxonomy:

Yes I remember now, thanks a lot!

PR #7945, opened 2023-01-04
feat: make all glass bottles recyclable by default

### LINKED ISSUE #4698: All glass bottles should be recyclable by default
### What
All glass bottles should probably be recyclable by default.

### Screenshot
image

### Part of 
- #6374

diivi:
### What
All glass bottles should be recyclable by default.

### Screenshot
image
image

- Fixes #4698

### DIFF
--- a/lib/ProductOpener/Packaging.pm
+++ b/lib/ProductOpener/Packaging.pm
@@ -505,6 +505,15 @@ sub apply_rules_to_augment_packaging_component_data ($product_ref, $packaging_re
 		$packaging_ref->{"shape"} = "en:coffee-capsule";
 	}
 
+	# If the shape is bottle and the material is glass, mark recycling as recycle
+	if (    (defined $packaging_ref->{"shape"})
+		and ($packaging_ref->{"shape"} eq "en:bottle")
+		and (defined $packaging_ref->{"material"})
+		and ($packaging_ref->{"material"} eq "en:glass"))
+	{
+		$packaging_ref->{"recycling"} = "en:recycle";
+	}
+
 	# If we have a shape without a material, check if there is a default material for the shape
 	# e.g. "en:Bubble wrap" has the property packaging_materials:en: en:plastic
 	if ((defined $packaging_ref->{"shape"}) and (not defined $packaging_ref->{"material"})) {
--- a/tests/unit/expected_test_results/packaging/packaging_text_en_glass_bottle.json
+++ b/tests/unit/expected_test_results/packaging/packaging_text_en_glass_bottle.json
@@ -9,7 +9,8 @@
    "packagings" : [
       {
          "material" : "en:glass",
-         "shape" : "en:bottle"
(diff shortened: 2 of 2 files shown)


lib/ProductOpener/Packaging.pm
```diff
@@ -505,6 +505,15 @@ sub apply_rules_to_augment_packaging_component_data ($product_ref, $packaging_re
 		$packaging_ref->{"shape"} = "en:coffee-capsule";
 	}
 
+	# If the shape is bottle and the material is glass, mark recycling as recycle
+	if (    (defined $packaging_ref->{"shape"})
+		and ($packaging_ref->{"shape"} eq "en:bottle")
+		and (defined $packaging_ref->{"material"})
+		and ($packaging_ref->{"material"} eq "en:glass"))
```
stephanegigandet, line 512: We should add a condition to check that we don't already have a value for $packaging_ref->{"recycling"}

If we have a value, we should not override it. Some glass bottles are not to be recycled, but to be returned to the store for instance.

lib/ProductOpener/Packaging.pm
```diff
@@ -505,6 +505,15 @@ sub apply_rules_to_augment_packaging_component_data ($product_ref, $packaging_re
...
+	{
+		$packaging_ref->{"recycling"} = "en:recycle";
+	}
```
alexgarel, line 515: Only apply the rule if there was no value provided.

Also as the test showed there is a [en:recicle-in-glass-bin](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/taxonomies/packaging_recycling.txt#L41) which seems even more appropriate.


```suggestion
	if (    (! defined $packaging_ref->{"recycling"})
	    and (defined $packaging_ref->{"shape"})
		and ($packaging_ref->{"shape"} eq "en:bottle")
		and (defined $packaging_ref->{"material"})
		and ($packaging_ref->{"material"} eq "en:glass"))
	{
		$packaging_ref->{"recycling"} = "en:recycle-in-glass-bin";
	}
```

lib/ProductOpener/Packaging.pm
stephanegigandet, line 512: `		and ($packaging_ref->{"material"} eq "en:glass"))`

It would be better to check if the material is en:glass or a child of en:glass.
You can use Tags.pm is_a() function for that:

is_a ($tagtype, $child, $parent)

lib/ProductOpener/Packaging.pm
stephanegigandet: I would prefer that we stick with "en:recycle". I'm not sure if we will keep the "recycle in XYZ bin" in the long run.

stephanegigandet (approved):
Looks good, thank you!

stephanegigandet:
> LGTM. @stephanegigandet do you want to take a look ?

I did, let's merge :)

PR #7946, opened 2023-01-05
fix: quality check when we have erythritol without polyols

stephanegigandet:
This fixes an issue when erythritol is set, but polyols is not set.


CharlesNepote (approved):
LGTM, thanks!

Issue #4561, opened 2020-11-27
Automatically set sugars or saturated-fat to 0 when carbohydrates or fat is 0


stephanegigandet:
### What
- It would help to compute the nutriscore for some products.
e.g. https://world.openfoodfacts.org/product/8888090781045/sunflower-oil-golden-circle

### Part of
- #5538

CharlesNepote:
As of 2022-10-25, [more than 44,000 products are in this case](http://mirabelle.openfoodfacts.org/products?sql=select+*+FROM%0D%0A%28select+count%28*%29+as+all_products+from+%5Ball%5D%29%2C%0D%0A%28select+count%28*%29+as+sat_or_sugar_empty_while_0_fat_or_carb+from+%5Ball%5D+where+%28fat_100g+%3D%3D+0+and+%5Bsaturated-fat_100g%5D+is+null%29+or+%28carbohydrates_100g+%3D%3D+0+and+%5Bsugars_100g%5D+is+null%29%29%2C%0D%0A%28select%0D%0A++%28round%28%0D%0A++100.0*%28select+count%28*%29+from+%5Ball%5D+where%0D%0A+++%28fat_100g+%3D%3D+0+and+%5Bsaturated-fat_100g%5D+is+null%29+or+%28carbohydrates_100g+%3D%3D+0+and+%5Bsugars_100g%5D+is+null%29%0D%0A++%29%0D%0A++%2F%0D%0A++%28select+count%28*%29+from+%5Ball%5D%29%0D%0A++%2C2%29%29+as+percent%0D%0A%29%0D%0A%0D%0A), representing **1.7% of all products**.

CharlesNepote:
A question related to this issue is the interest to know if the 0 is written on the product or computed.

I find interesting to know:
* what is actually printed on the product or not (we can use the hyphen "-" to tell that the value is not on the packaging)
* what is computed or not
But that makes thing more complicated.

Some other nutrient values could be computed based on some rules. For example, if a product is virgin olive oil and he does not have some nutrient values, the missing values could be computed/deduced from either category means or other databases such as CIQUAL, etc.

alexgarel:
@stephanegigandet: 

* I imagine I can do this akin to fix_salt_equivalent (Food.pm) which is called in analyze_and_enrich_product_data (Products.pm) ? 

  I would write a fix_zero_carbohydrates_sugar and fix_zero_fat_saturated

[...]

PR #7947, opened 2023-01-05
feat: deduce 0 sugar/sat fat to compute nutriscore

alexgarel:
where fat is 0 we can deduce saturated fat is 0, same for sugar from carbohydrates we use it to compute Nutri-Score

Fixes:
- #4561

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -1204,6 +1204,199 @@ my @fruits_vegetables_nuts_by_category_sorted
 	= sort {$fruits_vegetables_nuts_by_category{$b} <=> $fruits_vegetables_nuts_by_category{$a}}
 	keys %fruits_vegetables_nuts_by_category;
 
+=head2 compute_fruit_ratio($product_ref, $prepared)
+
+Compute the fruit % according to the Nutri-Score rules
+
+<b>Warning</b> Also modifies product_ref->{misc_tags}
+
+=head3 return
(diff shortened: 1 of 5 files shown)


lib/ProductOpener/Food.pm
```diff
@@ -1345,6 +1538,10 @@ sub compute_nutrition_score ($product_ref) {
 
 		foreach my $nid ("energy", "fat", "saturated-fat", "sugars", "sodium", "proteins") {
 			if (not defined $product_ref->{nutriments}{$nid . $prepared . "_100g"}) {
+				# we have two special case where we can deduce data
+				next
+					if (saturated_fat_0_because_of_fat_0($product_ref, $prepared)
+					|| sugar_0_because_of_carbohydrates_0($product_ref, $prepared));
```
stephanegigandet, line 1544: I think this will fail if you don't have both saturated_fat and sugars, but you have 1 of carbohydrates == 0 or fat == 0. You will get a Nutri-Score even though you can deduce only one of them.

CharlesNepote (approved):
I'm not able to review in details, but looks good to me.
I completely agree with this first step.

lib/ProductOpener/Food.pm
alexgarel: Is that ok @stephanegigandet (I added a test, so that should do)

stephanegigandet (approved):

PR #7948, opened 2023-01-05
fix: parsing of dots in packagins / recycling instructions

stephanegigandet:
Small fix for an issue reported on Slack by Mathieu Morey: " the text extraction got it really wrong: "Film plastique à jeter. Étui carton à recycler." adds a plastic sleeve (ok) that is marked recyclable (nope).
If I fix it and change the sleeve as "discard", another plastic sleeve that is recyclable is added..."

Also added more tests.


teolemon:
Another one is "Barquette et film plastique à recycler" which from memory is not parsed properly

alexgarel (approved):
LGTM

Issue #7929, opened 2023-01-02
Compute packaging stats for categories


stephanegigandet:
Now that users can enter detailed information for each packaging component (shape, material, weight etc.), we can try to compute packaging statistics for categories. e.g. to determine the most common packaging components of yogurts, the distribution of their weights etc.

Those stats can be used to suggest packaging components when editing products, to show how one product compares to similar products etc.

Which exact stats to compute is yet to be defined. We will probably start with some basic stats, see what we get and how we could extend them.

zanovis:
Awesome. Will users be able to enter this data manually or does the product package need to contain the information?

stephanegigandet:
We now compute basic stats by country / category / shape, that we use for suggestions of shapes and materials.

PR #7949, opened 2023-01-05
feat: compute packagings stats

stephanegigandet:
This is a script to compute some packaging stats for categories of products. #7929 

Sample output: https://world.openfoodfacts.org/data/categories_stats/categories_packagings_stats.packagings-with-weights.json
(for products that have some packaging weights)

Currently the stats are only on country / category / shape / material.

We will also add stats for packaging weight.


scripts/gen_packaging_stats.pl
```diff
@@ -0,0 +1,220 @@
...
+
+
+=head4 query reference $query_ref
+
+=cut
+
+sub generate_packaging_stats_for_query ($name, $query_ref) {
+
+	$query_ref->{'empty'} = {"\$ne" => 1};
+	$query_ref->{'obsolete'} = {"\$ne" => 1};
```
alexgarel, line 93:
```suggestion
	# we will filter out empty and obsolet products
	$query_ref->{'empty'} = {"\$ne" => 1};
	$query_ref->{'obsolete'} = {"\$ne" => 1};
```

scripts/gen_packaging_stats.pl
```diff
@@ -0,0 +1,220 @@
...
+		categories_tags => 1,
+		packagings => 1,
+	};
+
+	my $socket_timeout_ms = 3 * 60 * 60 * 60000;    # 3 hours
+	my $products_collection = get_products_collection($socket_timeout_ms);
+
+	my $products_count = $products_collection->count_documents($query_ref);
+
+	print STDERR "$name: $products_count products\n";
```
alexgarel, line 106: If we are going to put that in a cron, it would be better to have an option to disable "informative" prints and only keeps real errors, to have mail only on real failures.

scripts/gen_packaging_stats.pl
```diff
@@ -0,0 +1,220 @@
...
+
+		if ($total % 1000 == 0) {
+			print STDERR "$name: $total / $products_count processed\n";
+		}
+
+		# Generate stats for all countries + en:world (products from all countries)
+		if (not defined $product_ref->{countries_tags}) {
+			$product_ref->{countries_tags} = [];
+		}
+		push @{$product_ref->{countries_tags}}, "en:world";
```
alexgarel, line 128:
```suggestion
		# add a virtual en:world country to every products
		if (not defined $product_ref->{countries_tags}) {
			$product_ref->{countries_tags} = [];
		}
		push @{$product_ref->{countries_tags}}, "en:world";
```

scripts/gen_packaging_stats.pl
```diff
@@ -0,0 +1,220 @@
...
+			push @{$product_ref->{categories_tags}}, "all";
+
+			foreach my $category (@{$product_ref->{categories_tags}}) {
+
+				# Go through all packaging components
+				if (not defined $product_ref->{packagings}) {
+					$product_ref->{packagings} = [];
+				}
+
+				foreach my $packaging_ref (@{$product_ref->{packagings}}) {
```
alexgarel, line 145: I would have change the order of those two loops for performance reason (avoid calling gen_tags_hierarchy_taxonomy for every country).

But it's not a big deal.

scripts/gen_packaging_stats.pl
```diff
@@ -0,0 +1,220 @@
...
+						) += 1;
+					}
+				}
+			}
+
+		}
+
+	}
+
+	# Create directories for the output if they do not exist yet
```
alexgarel, line 189: There again I feel like separating in a function would help reading, but it's ok.

scripts/gen_packaging_stats.pl
```diff
@@ -0,0 +1,220 @@
...
+
+	print STDERR "$name: $products_count products\n";
+
+	my $cursor = $products_collection->query($query_ref)->sort({created_t => 1})->fields($fields_ref);
+
+	$cursor->immortal(1);
+
+	my $total = 0;
+
+	my $packagings_stats_ref = {};
```
alexgarel, line 114: I think it's better to document this kind of structure where we declare them, to avoid having to read the full algorithm to understand the structure.

Or even better we could document it as a json schema (in yml) in the docs/reference, that would be cool (it could also be considered an API), and put here the path to the OpenAPI schema.

alexgarel (approved):
I approve, but I would better like to have the OpenAPI documentation shipped with it (see my comment).

scripts/gen_packaging_stats.pl
stephanegigandet: There's a description of the structure at the top of the file:

```
Aggregation counts are stored in a structure of the form:
{
    countries => {
        "en:world" => ..
        "en:france" => {
            categories => {
                "all" => .. # stats for all categories
                "en:yogourts" => {
                    shapes => {
                        "en:unknown" => ..
                        "all" => .. # stats for all shapes
                        "en:bottle" => {
                            materials_parents => .. # stats for parents materials (e.g. PET will also count for plastic)
                            materials => {
                                "all" => ..
                                "en:plastic" => 12, # number of products sold in France that are yogurts and that have a plastic bottle packaging component
                            }
                        },
                        ..
                    }
                },
                ..
            }
        },
        ..
    }
}
```

Regarding the doc in OpenAPI, why not, but at this point this is completely experimental, I don't know if we will keep that structure or not. We can document in OpenAPI once it's a bit more stabilized.

scripts/gen_packaging_stats.pl
stephanegigandet: Agreed, I will add a --quiet option

scripts/gen_packaging_stats.pl
stephanegigandet: Good point. I did it in this order because it was more logical in my head (I want packagings for all categories and for all countries), but indeed it multiplies the calls needlessly.

stephanegigandet:
Thanks for all the suggestions @alexgarel , I think I implemented them all in the last commit.

Please don't merge yet, there are a few details I want to change (like filtering out bogus entries in countries).

scripts/gen_packaging_stats.pl
alexgarel: ok.

alexgarel:
```suggestion
	# this will contains the final result
	# see structure on top of this file
	my $packagings_stats_ref = {};
```

teolemon:
I'm not quite sure what to make of this (I looked at the JSON beautified). 
@CharlesNepote Is the packaging structured field exported ? It would be easier to pivot, probably.

stephanegigandet:
I added a special export just for French yogurts, to make the data easier to explore in just a browser:

https://world.openfoodfacts.org/data/categories_stats/categories_packagings_stats.fr.fermented-dairy-desserts.packagings-with-weights.json

So for instance we can easily see the materials used for pots of yogurts:

image

Note that there are fields for "shape" / "shape_parents", and "material", "material_parents". This is so that we can see stats for all "pots" even if we have some components listed as "pots" and others as "individual pots". Same thing for materials: in "materials" you see the exact values entered by users, and in "materials_parents" you have all the parents values as well.

stephanegigandet:
I added weights (values + mean):

image

PR #7956, opened 2023-01-06
feat: producers platform with new design

stephanegigandet:
This PR is a work in progress, it migrates producers platform features that were in different places (like the side bar that we removed) to the new design.


stephanegigandet:
closing, will make fresh PR

PR #7957, opened 2023-01-06
taxonomy: added a-market swedish label

benbenben2:
Requested by Manon on Slack (taxonomy channel, 6/1/2023)


taxonomies/labels.txt
```diff
@@ -2741,6 +2741,14 @@ nl:Zonder kunstmatige toevoegingen
 pt:Sem aditivos artificiais
 label_categories:en: en:Health, en:Additives
 
+en:Ä-märket
+xx:Ä-märket
```
alexgarel, line 2745: I think you need to add se:Ä-märket in order for the image to be retrieved. (not sure, but better secure than sorry).
You could also put se:Ä-märket instead of en:Ä-märket it makes sense as it is a Swedish label.

PR #7961, opened 2023-01-07
taxonomy: added allergens and categories

benbenben2:
added few more ingredients


alexgarel (approved):
:+1: thanks @benbenben2

PR #7967, opened 2023-01-10
taxonomy: Added 'Refined palmolein oil'

hakuamesan:
### What
Added 'Refined palmolein oil' as a synonym of "refined palmolein"

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -14598,7 +14598,7 @@ sk:palma
 sv:Palm
 
 <en:palm oil
-en:palm olein, palmolein
+en:palm olein, palmolein, palm olein oil
 de:Palmolein
 es:oleína de palma
 fi:palmuoleiini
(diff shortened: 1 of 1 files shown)


taxonomies/ingredients.txt
```diff
@@ -14611,6 +14611,18 @@ wikidata:en:Q57312740
 description:en:The olein (liquid) fraction of palm oil.
 # 328 in 9 @2021-09-09
 
+<en:palm oil
+en:refined palmolein oil
```
stephanegigandet, line 14615: I think the main name should be "en:refined palm olein", as we have an entry "en:palm olein" already.

So "en:refined palm olein" should be under "<en:palm olein".

I'm guessing also that "palm olein oil" is the exact same thing as "palm olein" (I could be wrong).

In which case we should also add "palm olein oil" as a synonym of "palm olein".

stephanegigandet:
Hi @hakuamesan 

I'm gathering the feedback from Slack, here is what we should do:

- "en:palm olein oil" should be a new entry (under "<en:palm oil" and "<en:palm olein".
- en:refined should be added to ingredients_processing.txt (that way we don't have to add "refined palm olein oil")

alexgarel:
@stephanegigandet should we close it ? As https://github.com/openfoodfacts/openfoodfacts-server/pull/8253 is merged.

stephanegigandet:
This change has been made in another PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/8253

Thank you @hakuamesan for bringing this up!

PR #7980, opened 2023-01-13
taxonomy: remove useless item

raphael0202:
Remove useless entry in ingredient taxonomy


stephanegigandet (changes requested):
We should remove this entry completely, as the last part "issus de l'agriculture biologique" is matched automatically to the en:organic label.

e.g. https://fr.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=jus+de+fruits+issus+de+l%27agriculture+biologique&type=add&action=process&submit=Envoyer

We don't match the "originaires et non originaires de l'union européenne" yet, so we should keep "fr:mélange de miels originaires et non originaires de l'ue" for now.

raphael0202:
@stephanegigandet I've removed the entry.

stephanegigandet:
> @stephanegigandet I've removed the entry.

Perfect, thanks.

stephanegigandet (approved):

PR #7982, opened 2023-01-13
fix: added sugars for non US countries 

### LINKED ISSUE #7981: "Added sugars" is not suggested in the nutrition table product edit form
Reported by @aleene. On https://us.openfoodfacts.org , we show "Added sugars" below "Sugars" by default, but we don't display it for other countries, and it's not suggested when starting to type "Added sugars" in the other nutriments box.

stephanegigandet:
fixes #7981


aleene (changes requested):
Can you add "added-salt" as well?

stephanegigandet:
> Can you add "added-salt" as well?

Are there nutrition table on products with "added salt"? I'm guessing that in most products with a significant amount of salt, almost of it could be considered "added".

stephanegigandet:
image

stephanegigandet:
@aleene sounds good, I added "added salt", can you approve the PR?

alexgarel (approved):

PR #7986, opened 2023-01-14
taxonomy: added Croatian entries

benbenben2:
added some Croatian entries (categories, ingredients, etc.)


stephanegigandet (approved):
Thank you!

PR #7987, opened 2023-01-14
taxonomy: added and removed stopwords for ingredients in HR

benbenben2:
Updates are made after the following observations:

može sadržavati tragove
Sadrži
"Popis sastojaka"
"naziv proizvoda"
!!!! RM pasterizirano because ingredient can be "Nehomogenizirano pasterizirano mlijeko"!!!
!!!! RM HR because if the address is after ingredient "sastojci blablabla HR blablabla2" it starts after hr leading to: "blablabla2"
"Čuvati zatvoreno na"
"Cuvati" (accent missed)
"Čuvati pri sobnoj temperaturi."

"Izvor dijetalnih vlakana"
Proizvod je termički obrađen-pasteriziran
"Prosječne hranjive vrijednosti"
"Prosječne nutritivne vrijednosti"
"Protresti prije otvaranja"
"Uvoznik za"
"zaštićena oznaka zemljopisnog podrijetla"
"Zemlja podrijetla"
"Zemlja porekla kakao mase EU"


stephanegigandet (approved):
Thanks

benbenben2:
> Thanks

having trouble with perl tidy :-(

alexgarel:
@benbenben2 just run `make lint_perltidy` and commit changes.

alexgarel:
@stephanegigandet this is the error with perl tidy:

```
Use of strings with code points over 0xFF as arguments to bitwise xor (^) operator is not allowed at /opt/perl/local/lib/perl5/Perl/Tidy.pm line 2778.
```
Is this linked to a character used by @benbenben2 

@benbenben2, UTF-8 is a bit strange some times, as for example you might have the character e + ` as a modifier or è directly (you see same think but it's not the same). The problem here is with one of the character you input.

On think you can try is to use [re.pl](https://metacpan.org/pod/Devel::REPL) : 

- add `export CPANMOPTS="--with-develop --with-feature=off_server_dev_tools"` in your .envrc (if you use direnv or the Makefile, or directly in .env, but do not commit)
- rebuild
- the run `docker-compose run --rm backend re.pl`

It gives you a sandbox to test your expressions.

alexgarel:
@benbenben2 or maybe try to copy the first "ž" already present instead of retyping it in the rest of the sentence, because your ž might be different and cause the problem (wild guess)

PR #7992, opened 2023-01-16
feat: adding adhesive tapes as a packaging material

fuddl:
adding adhesive tapes such as on this product:

adhesive tape on a wrapper of a chocolate bar

There should probably more types of tape.

### DIFF
--- a/taxonomies/packaging_materials.txt
+++ b/taxonomies/packaging_materials.txt
@@ -1542,3 +1542,17 @@ en:fabric, fabrics
 de:Stoff, Stoffe, Gewebe
 fr:tissu, tissus
 nl:Stof, stoffen
+
+en:adhesive tape
+de:Klebeband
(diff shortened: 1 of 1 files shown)


stephanegigandet (commented):
For each packaging component, we have a shape, and a material.

What would be the corresponding shape? Maybe we need to add "en:tape" entry in packaging_shapes.txt

Then for materials, en:adhesive plastic tape would be under en:plastic, and en:washi tape would be under en:paper.

Another option may be to put "adhesive tape" as a packaging shape, and then we just use the existing "plastic" and "paper" materials.

fuddl:
i'd say the shape should be 'strip'

stephanegigandet (approved):

stephanegigandet:
> i'd say the shape should be 'strip'

Sounds good.

Issue #8002, opened 2023-01-17
New API v3 for dropdown and autocomplete suggestions for packaging shapes, materials and recycling taxonomy


stephanegigandet:
On the web, the product edit form has dropdown/autocomplete for packaging shapes and material with suggestions sorted alphabetically, which provides poor results (especially since people can type a material short name such as PET, or its number such as 1 or 01).

![image](https://user-images.githubusercontent.com/8158668/212877558-6ad8920f-8aa9-4273-8f9f-a1521d5dbf34.png)

We have a generic taxonomy suggestion API that is a bit better (priority given to strings at the start of taxonomy entries, match with xx: entries): https://world.openfoodfacts.org/cgi/suggest.pl?tagtype=packaging_materials&term=pe

We can do better by suggesting the most common shapes given the country and categories of the product. e.g. for wine we can suggest a bottle. For a wine bottle, we can suggest glass.

Requirements:
- the match must happen on the main name of the taxonomy entry + all its synonyms (e.g. in French the main name of "Recycle in the glass bin" is "A recycler dans le conteneur à verre", but we also want to match "bac verre")
- the match must also work with the xx: language independent entries that contain material codes that may not be in the language specific names
- the most popular entries for similar products (based on categories, and possibly other fields like brand) must come first

stephanegigandet:
A first version has been deployed, but we are still missing some features: matching synonyms, and matching xx: synonyms.

thomas-algo:
Hi ! For now, the most common shapes and materials given the country and categories are suggested first, I believe it would be interesting to do the same with recycling instruction (for now it is given in alphabetical order).

This would be useful for the autocomplete of recycling instructions in the packaging components.

stephanegigandet:
@thomas-algo Indeed, we could compute stats for recycling instructions by country + material. Filed https://github.com/openfoodfacts/openfoodfacts-server/issues/8189

PR #8004, opened 2023-01-17
fix: /cgi/suggest.pl + tests

stephanegigandet:
I added tests for /cgi/suggest.pl as I'm working on taxonomy suggestions: https://github.com/openfoodfacts/openfoodfacts-server/issues/8002 and I want to move most of the suggest.pl code to a separate module.

There was a "use HTTP" missing that we need for the CORS refactor.


alexgarel (approved):
Perfect

alexgarel:
As I saw your test, I was thinking : we even could generate documentation from those tests !

PR #8009, opened 2023-01-19
style: Add styles to greeter for not-logged-in users according to Figma mockup.

sultanowski:
### What
- [x] Added styling according to Figma mockup.
- [x] Added partially LTR support.
- [x] Removed not used conditions in `.../templates/site_layout.tt.html`
- [x] Removed spotted empty style block in `~/scss/_off.scss`
- [x] CSS LSP undescored that `font-ligature-setting` is provided only for WebKit browsers, so I added a line to remove this warning.
- [x] Commented out "Pernament link to results..." since this feature is not working anyway and it does not fit in this place anymore.


stephanegigandet:
The first page when not logged in looks very good!

image

stephanegigandet:
On medium screens, the text over the illustrations is not completely legible (even if the illustrations are slightly lightened).

image

templates/web/common/site_layout.tt.html
```diff
@@ -249,21 +249,9 @@ <h1 itemprop="name">[% title %]
...
 								</div>
 							[% ELSE %]
-								<div class="row">
-									<div class="small-12 column v-space-short">
-										<h1>[% title %]</h1>
-									</div>
-								</div>
 								[% content %]
 							[% END %]
-						[% ELSE %]
```
stephanegigandet, line 259: This part should not be removed, otherwise it makes all content from pages that are not product pages, list of products and product edit form have not content.

e.g. the "Discover" page: 
image

templates/web/common/site_layout.tt.html
```diff
@@ -249,21 +249,9 @@ <h1 itemprop="name">[% title %]
 									[% content %]
 								</div>
 							[% ELSE %]
-								<div class="row">
```
stephanegigandet, line 252: This should be kept, if there is no schema_org_itemtype defined. (today there's always the "Thing" schema by default, but that may change).

stephanegigandet:
When signed-in, there's an issue with the background colors:

image

stephanegigandet (changes requested):
The greeter looks very good, but it broke a few things.

sultanowski:
I am going to resolve those issues ASAP. Thansks for review!

sultanowski:
- [x] Reverted changes to `site-layout.tt.html` file - it renders content on other pages.
- [x] Added global class to not display HTML tags without content.
- [x] Changed opacity of images to lower on smaller devices to make texts more visible.
- [x] Fixed issues with full-height background color.

stephanegigandet (approved):
I tested it and it looks very good, thanks for making the changes @sultanowski!

Issue #7883, opened 2022-12-21
Have all API help in OpenAPI menu, or have OpenAPI rendering in mkdocs


alexgarel:
Currently, if you access https://openfoodfacts.github.io/openfoodfacts-server, you can see the tutorial, API intro, etc. but as soon as you dive into the API spec https://openfoodfacts.github.io/openfoodfacts-server/reference/api, the menu on the left only contains API points.

We should either include a link to intro + tutorials + cheasheet, etc in the redoc menu, or render the doc inside mkdocs.

alexgarel:
@teolemon @Roxie-32 including external markdown in redoc is really easy indeed. 
See https://redocly.com/docs/api-reference-docs/guides/embedded-markdown/

alexgarel:
There also seems to be a [sidebars.yaml to configure the left menu](https://redocly.com/docs/developer-portal/guides/reference-docs-integration/) (not sure if it works in OSS version).

alexgarel:
I tried and sidebars in the leftmenu is a feature which is NOT available in community edition !

I have two choices now:
* either we use a specific template with javascript to inject the menu entries we want (that seems doable, but I'm not sure)
* either we try https://rapidocweb.com/ which is fully Open Source, but I yet have to see how to use it.

PR #8011, opened 2023-01-19
docs: try rapidoc

alexgarel:
Trying [rapidoc](https://rapidocweb.com/api.html#slots)

to solve:
- #7883

### DIFF
--- a/docs/assets/api-rapidoc.html
+++ b/docs/assets/api-rapidoc.html
@@ -38,6 +38,7 @@
     show-components="true"
     show-header="false"
     info-description-headings-in-navbar="true"
+    default-api-server="https://off:off@world.openfoodfacts.net"
   >
   <img slot="logo" src="https://static.openfoodfacts.org/images/logos/off-logo-horizontal-light.svg" style="width:200px;"/>
   <nav  slot="nav-logo" class="off-nav">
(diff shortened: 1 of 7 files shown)


docs/assets/api-rapidoc.html
```diff
@@ -0,0 +1,55 @@
+<!doctype html> <!-- Important: must specify -->
```
alexgarel, line 1: Note to reviewer: download this file to try it locally it would work (appart the added menu urls)

stephanegigandet:
Looks good. So this page would replace https://openfoodfacts.github.io/openfoodfacts-server/reference/api/ ?

image

stephanegigandet:
The "parameters" list at the end is interesting, but it doesn't tell which path uses which parameter, so it's a bit weird.

stephanegigandet:
Could you generate the api-v3 doc as well? I would like to see if it has the same issues as https://openfoodfacts.github.io/openfoodfacts-server/reference/api-v3/#tag/Write-Requests which does not show the product->packagings input field.

alexgarel:
@stephanegigandet 

* at the moment it does not replace the old doc (but I will do it, as reverting is very easy as @teolemon made me notice)

* for the parameters, it's because I added the option to add the schema because I wanted to display "products" as an independent page (it could be handy for example to refer to that for csv export fields)

* I will add api-v3

alexgarel:
@stephanegigandet do you want to validate (it's easy to roll-back if needed).

stephanegigandet (approved):
Looks good!

PR #8016, opened 2023-01-20
feat: integrating openfoodfacts-server with codecov for unit test coverage

yuktea:
This PR has,
- Used CPAN::`Devel::Cover` for generating a coverage report for `ProductOpener`.
- Used `Devel::Cover::Report::Codecovbash` instead of the recommended `Devel::Cover::Report::Codecov` because instead of directly uploading to codecov, `Devel::Cover::Report::Codecovbash` generates a `json` file first.


alexgarel (commented):
Tests are failing, maybe because you need to build taxonomies and be sure to rebuild languages.

I think if we add coverage it has to replace test (because really it's running tests) in pull_request.yml

alexgarel:
Also if there is a way to avoid the cover progression, because it makes logs really hard to read…

.github/workflows/codecov.yml
```diff
@@ -0,0 +1,26 @@
...
+on: [push]
+jobs:
+  coverage:
+    name: unit test coverage
+    runs-on: ubuntu-latest
+    steps:
+       - uses: actions/checkout@master
+       - name: rebuild taxonomies
+         run:  |
+          git ls-files taxonomies/ | xargs -I{} git log -1 --date=format:%Y%m%d%H%M.%S --format='touch -t %ad "{}"' "{}" | bash
```
alexgarel, line 12: cool, can you add a comment on this, for our future us ;-)

alexgarel (approved):
Perfect ! Good work :-)

Let's start with this in parallel of normal tests, but I hope we will move fast to also have integration tests so that we can replace the main test run by this.

.github/workflows/codecov.yml
alexgarel: ignore it if you just copy/pasted it from the other file, in this case I might be the culprit :grimacing:

yuktea:
we can merge this now @alexgarel

alexgarel:
@yuktea if a PR is approved and checks are ok, feel free to merge it yourself :slightly_smiling_face:

Issue #7231, opened 2022-08-22
Add the logo for the Label Bee Friendly 


manon-corneille:
The Label [Bee Friendly](https://fr.openfoodfacts.org/label/bee-friendly) is available on Open Food Facts, but the logo is missing. 

You can find the logo below in two different resolutions : 
- ![bee-friendly 90x90](https://user-images.githubusercontent.com/104015489/185884742-b8def68c-f30e-49ff-8bed-b9bdd4f58b3f.png)
- ![bee-friendly 1772x1772](https://user-images.githubusercontent.com/104015489/185884744-54e1c06b-be46-41b0-913c-2de4c54e8b48.png)
 
Thank you :)

alexgarel:
if needed to do ourselves the reduction https://www.certifiedbeefriendly.org/wp-content/uploads/2022/01/cropped-cropped-BEEFRIENDLY_logoHD_RVB_020318-300x300.png is a better starting point.

pratyaksh1610:
@alexgarel I have added the above URL image in path - https://github.com/openfoodfacts/openfoodfacts-server/tree/main/html/images/lang/fr/labels . Is it preferred , then I can drop a PR for the same.

alexgarel:
good idea @pratyaksh1610, I can't see your PR though.

pratyaksh1610:
@alexgarel I was asking , is it the correct path ?

alexgarel:
Sorry @pratyaksh1610 for the latency, yes you are correct on the path.

pratyaksh1610:
@alexgarel Okay, cool 👍

manon-corneille:
Hi guys, 
Bee Friendly got back to me with the SVG logo (not sure if it's still relevant, but here it is) :) 

[BEE FRIENDLY logo png_adobe_express](https://user-images.githubusercontent.com/104015489/190621811-37caa057-b0ae-4b86-9ae8-46d09fc10598.svg)

alexgarel:
@pratyaksh1610 do you want to do it ?

pratyaksh1610:
@alexgarel Oh, I missed track of this.
Dropping a PR for the same.

alexgarel:
@pratyaksh1610 I think for this to work, we also have to taxonomize this entry, that is, add it to labels.txt in english and french, at least.

alexgarel:
[...]

PR #8018, opened 2023-01-20
taxonomy: fix bee friendly label name

alexgarel:
Bee friendly was added, see #7231 but under the wrong name (the tagline instead of the real name)

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -19396,9 +19396,9 @@ pt:Maçãs da França
 origins:en: en:france
 ingredients:en: en:apple
 
-en:bee friendly, I protect the bees
-fr:bee friendly, Je protège les abeilles
-xx:bee friendly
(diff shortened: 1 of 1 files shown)


taxonomies/labels.txt
```diff
@@ -19396,8 +19396,9 @@ pt:Maçãs da França
 origins:en: en:france
 ingredients:en: en:apple
 
-en:I protect the bees
-fr:Je protège les abeilles
+en:bee friendly, I protect the bees
```
stephanegigandet, line 19399: They capitalize their logo: BEE FRIENDLY. It's a bit harsh, I think we could use "Bee Friendly" as the main capitalization (which will be displayed in the lists of labels).

alexgarel: Fixed, thanks.

stephanegigandet (approved):
Thanks!

Issue #7918, opened 2022-12-30
Estimate Ingredient percentages from nutritional data


john-gom:
Use CIQUAL to get the the nutrient composition of each ingredient and use this to estimate the original recipe by matching with the nutritional information about the product.

Discussion on potential algorithms here: https://docs.google.com/document/d/1cPEEhl9Ty0feKEel2NxUmscRR6oWZTLJMpTSbLbzB5U/edit?usp=sharing

john-gom:
Issues observed during development:

- Have no CIQUAL codes for ingredients like:
  - [ ] Water
  - [ ] Vitamins (e.g. added B12)
  - [ ] Minerals (e.g. Calcium Carbonate)
  - [ ] Processed raw ingredients, e.g. Pea Protein (Isolate)
- [ ] Nutrient parsing seems to miss data sometimes (e.g. 7394376616709). Can't see source text from OCR

Improvements needed to the algorithm
- [ ] Some ingredients are not specific enough to ever be known (e.g. "Flavouring")
- [ ] Dealing with advertised Ingredient quantities
- [ ] Qualified nutrient levels e.g. less than, "trace" (in CIQUAL or product)

john-gom:
Hi @karthiknadar1204 , the code is now in the recipe-estimator repo on a working branch. There are instructions on running in docker if you want to play. Contact @alexgarel if you get any permission issues. There is a project board in the repo on areas I was hoping to look at...

alexgarel:
See https://github.com/openfoodfacts/recipe-estimator/

alexgarel:
@john-gom can we make the project board public in recipe-estimator ?

PR #8025, opened 2023-01-23
feat: Estimate Ingredient percentages from nutritional data

john-gom:
### What

Store the ciqual_food_code on each ingredient and calculate the number of ingredients that don't have Ciqual codes (ingredients_without_ciqual_codes_n) and list these (ingredients_without_ciqual_codes) to help prioritise updating taxonomies to have the biggest impact

New option --assign-ciqual-codes added to update_all_products.pl

- Contributes towards #7918


lib/ProductOpener/Ingredients.pm
```diff
@@ -2392,6 +2396,34 @@ sub extract_ingredients_from_text ($product_ref) {
...
+		if (defined $ingredient_ref->{ingredients}) {
+			push(@ingredients_without_ciqual_codes, get_missing_ciqual_codes($ingredient_ref->{ingredients}));
+		}
+		else {
+			my $ciqual_food_code = get_inherited_property("ingredients", $ingredient_ref->{id}, "ciqual_food_code:en");
+			if (defined $ciqual_food_code) {
+				$ingredient_ref->{ciqual_food_code} = $ciqual_food_code;
+			}
+			else {
+				$ingredient_ref->{ciqual_food_code} = undef;
```
stephanegigandet, line 2418: I would remove the undef, and just have a non existing field. It's what we do for all other fields, when it is undefined, we delete the field.

stephanegigandet (approved):
Looks good to me, thank you!
Just one minor request to remove the ciqual_food_code field when its value is "null"

john-gom:
> I would remove the undef, and just have a non existing field. It's what we do for all other fields, when it is undefined, we delete the field. 

I was anticipating a scenario where an ingredient was incorreclty assigned to a CIQAL code in the past, so we would need to remove any existing link. Should I delete instead in this case?

PR #8027, opened 2023-01-23
feat: rebuild taxonomies as they change fixes #6895

### LINKED ISSUE #6895: Rebuild taxonomies as they change
### What
- When a taxonomy change is merged in main, create an action that rebuild taxonomies and propose it as a PR.
- We do this because it will make validations on next pull requests faster (we have to rebuild changed taxonomies on every PR before launching tests).
- We could also rebuild taxonomies directly on pull requests and commit but I find it messy:
  * it will add a lot of commits for every commit made by author
  * author as to constantly remind to pull again before committing and some author will pull with a merge, making history on the PR quite difficult to read

john-gom:
### What

Creates a hash of source files and looks up from local or GitHub cache repo.

Updated build_lang so that it builds nutrient_levels to reduce amount of circular dependency between languages and taxonomies. Language taxonomy does still have to be built first but this is done on-the-fly when languages are built.

Build cache folder is created as a symlink in the docker image so volumes may need to be re-built.

- Fixes #6895


stephanegigandet:
The cache directory should probably be in .gitignore

lib/ProductOpener/Tags.pm
```diff
@@ -1601,10 +1737,33 @@ sub build_tags_taxonomy ($tagtype, $file, $publish) {
...
+
+=head3 Parameters
+
+=head4 Publish STO file $publish
+
+=cut
+
+sub build_all_taxonomies ($publish) {
+	foreach my $taxonomy (@taxonomy_fields) {
+		if ($taxonomy ne "traces" and rindex($taxonomy, 'data_quality_', 0) != 0) {
```
alexgarel, line 1764:
```suggestion
		# traces and data_quality_xxx are not real taxonomy per se 
		# (but built from allergens and data_quality)
		if ($taxonomy ne "traces" and rindex($taxonomy, 'data_quality_', 0) != 0) {
```

scripts/build_tags_taxonomy.pl
```diff
@@ -26,83 +26,17 @@
...
-binmode STDIN, ":encoding(UTF-8)";
-binmode STDOUT, ":encoding(UTF-8)";
-
-my $file = $tagtype . ".txt";
-
-# The nutrients_taxonomy.txt source file is created from values in the .po files
-if ($tagtype eq "nutrient_levels") {
-	create_nutrients_level_taxonomy();
+if ($tagtype eq '*') {
+	build_all_taxonomies($publish);
```
alexgarel, line 36: :+1:

Makefile
```diff
@@ -325,12 +325,14 @@ check_critic:
...
+	${DOCKER_COMPOSE} run --no-deps --rm -e GITHUB_TOKEN=${GITHUB_TOKEN} backend /opt/product-opener/scripts/build_tags_taxonomy.pl ${name}
 
-rebuild_taxonomies:
-	@echo "🥫 re-build all taxonomies on ${CPU_COUNT} procs"
-	${DOCKER_COMPOSE} run --rm backend make -C taxonomies all_taxonomies -j ${CPU_COUNT}
+rebuild_taxonomies: build_taxonomies
+
+build_taxonomies_test:
+	@echo "🥫 build taxonomies"
+	${DOCKER_COMPOSE_TEST} run --no-deps --rm -e GITHUB_TOKEN=${GITHUB_TOKEN} backend /opt/product-opener/scripts/build_tags_taxonomy.pl ${name}
```
alexgarel, line 335: I don't think we need a specific target for tests because we bind mount the taxonomies folder (so it's shared between tests and dev)

alexgarel (changes requested):
Really a great work !

I have some minor comments.

You will modify the pull_request.yml file to add the GITHUB_TOKEN variable.

lib/ProductOpener/Tags.pm
```diff
@@ -650,6 +657,57 @@ sub get_lc_tagid ($synonyms_ref, $lc, $tagtype, $tag, $warning) {
...
+	my $cache_root = "$data_root/build-cache/taxonomies";
+	my $local_cache_source = "$cache_root/$source";
+	if (-e $local_cache_source) {
+		copy($local_cache_source, $target);
+		return 1;
+	}
+
+	$File::Fetch::WARN = 0;
+	my $ff = File::Fetch->new(
+		uri => "https://raw.githubusercontent.com/openfoodfacts/openfoodfacts-build-cache/main/taxonomies/$source");
```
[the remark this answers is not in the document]
john-gom: Might need some help understanding how this will apply to obf and opff

lib/ProductOpener/Tags.pm
```diff
@@ -650,6 +657,57 @@ sub get_lc_tagid ($synonyms_ref, $lc, $tagtype, $tag, $warning) {
...
+		my $ua = LWP::UserAgent->new(timeout => 300);
+		my $url = "https://api.github.com/repos/openfoodfacts/openfoodfacts-build-cache/contents/taxonomies/$target";
+		$ua->put(
+			$url,
+			Accept => 'application/vnd.github+json',
+			Authorization => "Bearer $token",
+			'X-GitHub-Api-Version' => '2022-11-28',
+			Content => $content
+		);
+	}
```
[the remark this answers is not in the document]
john-gom: Adding a warning if the response is not success

lib/ProductOpener/Tags.pm
```diff
@@ -668,10 +726,88 @@ Like "categories", "ingredients"
...
+	if ($got_from_cache) {
+		$got_from_cache = get_from_cache("$cache_prefix.result.txt", "$tag_data_root.result.txt");
+	}
+	if ($got_from_cache) {
+		$got_from_cache = get_from_cache("$cache_prefix.json", "$tag_www_root.json");
+	}
+	if ($got_from_cache) {
+		print "obtained taxonomy for $tagtype from " . ('', 'local', 'GitHub')[$got_from_cache] . " cache.\n";
+		return;
+	}
```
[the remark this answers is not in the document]
john-gom: I used to be the same, but modern conventions seem to favour early returns to minimise nesting - in this case it would be a very big else block...

Makefile
```diff
@@ -225,14 +225,14 @@ lint: lint_perltidy
 
 tests: build_lang_test unit_test integration_test
 
-unit_test:
+unit_test: build_taxonomies_test
```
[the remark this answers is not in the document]
john-gom: Can we discuss?

lib/ProductOpener/Tags.pm
```diff
@@ -650,6 +657,106 @@ sub get_lc_tagid ($synonyms_ref, $lc, $tagtype, $tag, $warning) {
...
+		my $url = "https://api.github.com/repos/$build_cache_repo/contents/taxonomies/$target";
+		my $response = $ua->put(
+			$url,
+			Accept => 'application/vnd.github+json',
+			Authorization => "Bearer $token",
+			'X-GitHub-Api-Version' => '2022-11-28',
+			Content => $content
+		);
+		if (!$response->is_success()) {
+			print "Error uploading to GitHub cache for $target: ${\$response->message()}\n";
```
alexgarel, line 742: why not use $log->error here ?

alexgarel (approved):
Cool ! Thanks for the great work.

lib/ProductOpener/Tags.pm
john-gom: Copying other code. I think because it is used in build process the log wouldn't appear

PR #8041, opened 2023-01-26
build: When detecting MacOS then use MacOS-specific 'sysctl' command for setting CPU_COUNT

sgtSeme4ki:
### What

Added a Check which sets the CPU_COUNT variable via sysctl instead of nproc if run on MacOS. Otherwise building via the Makefile would fail because on MacOS because nproc doesn't exist on MacOS. 

- Fixes #[8040]


stephanegigandet (approved):
Thank you!

PR #8042, opened 2023-01-26
doc: Make api.yml a valid OpenAPI-3.1.0-Spec (checked via openapi-gene…

sgtSeme4ki:
…rator validate)

### What
The api.yaml (including tags_parameters.yaml and product.yaml) where invalid while validation via openapi-generator validate. The validation issues were fixed to make api.yaml a valid 3.1.0 openAPI-Spec so it can be used by clients for client-code generation via the openAPI-Generator plugin etc.

- Fixes #[8039]


stephanegigandet (approved):
Thank you!

stephanegigandet:
Thank you very much @sgtSeme4ki. We should add a test to make sure the files pass validation.

sgtSeme4ki:
I'm tried to add a github-action which validates the openApiSpec. Unfortunately the specific gitHubAction can't validate openApiSpecs with the 3.1.0-Version. I will try to use the OpenApiValidator (which I used locally) as a gitHub-Action to validate the openApiSpec.

docs/reference/schemas/tags_parameters.yaml
```diff
@@ -125,14 +125,15 @@ properties:
     in: query
     name: traces_tags
     description: |
-      The traces_tags of product(s) you are searching for. 
+      The traces_tags of product(s) you are searching for.
       The [OFF App](https://world.openfoodfacts.org/traces) shows a list of possible values for `traces`.
   tag_name_with_language_code:
     in: query
+    name: tag_name_with_language_code
```
alexgarel, line 132: I would use the one I used below:
```suggestion
    name: '<tag_name>_tags_<language_code>'
```

docs/reference/schemas/tags_parameters.yaml
```diff
@@ -125,14 +125,15 @@ properties:
...
     description: |
       You can add a language code to a specific tag to query it in a specific language
     schema:
-      type: Object
+      type: object
       name: '<tag_name>_tags_<language_code>'
```
alexgarel, line 137: So maybe this one is not useful (if it's in the part above) ?

alexgarel:
@sgtSeme4ki thank you so much for this :-)

What about using https://pypi.org/project/openapi-spec-validator/ for validation, it says it supports OpenAPI 3.1

sgtSeme4ki:
> @sgtSeme4ki thank you so much for this :-)
> 
> What about using https://pypi.org/project/openapi-spec-validator/ for validation, it says it supports OpenAPI 3.1

I tried the validator you mentioned, but while validating it gave me a python error "string indices must be integers". There is probably an Error while validation which isn't covered by this validator. I used the validator which is included in the openApiGenerator and included it as a docker run (https://openapi-generator.tech/docs/installation/). I also made some small fixes in the OpenApiSpec so that the validator validates without errors.

stephanegigandet:
Thank you very much @sgtSeme4ki !

@alexgarel : can we merge?

chris-hatton:
Since this merge, the OpenAPI schemas have changed path and neither `api.yml` or `api-v3.yml` are considered valid by:

- VSCode OpenAPI editor
- Gradle OpenAPI generation plugin

This may be a regression?

alexgarel:
@chris-hatton 
1. yes we renamed files for we reorganize documentation
2. I think there is a problem with the validation tool that was setup, for me it reports a false positive, saying "is not an object" when I use #Product (json reference).
3. feel free to fix it if you can, it would be great. But just take care that the documentation rendrering with rapidoc must continue to work.

shinjigi:
Hi everyone,

I wanted to bring to your attention a related issue that might provide additional context or insights into this bug. You can view the discussion and details here: [Issue #9419 - Comment](https://github.com/openfoodfacts/openfoodfacts-server/issues/9419#issuecomment-2410366007).

Feel free to review it and see if it helps in resolving the current issue.

Issue #5441, opened 2021-08-05
Incorrect "serving" value for serving size field, Unilever import through the producers platform


stephanegigandet:
### What

- [ ] https://fr.openfoodfacts.org/produit/8722700472575/amora-sauce-caesar-salade-sandwich-bouteille-1l?rev=23
  - [ ] "Informations nutritionnelles
  - [ ] Taille d'une portion : serving"

### Part of
- #5521

stephanegigandet:
Related to #5163

Issue #5163, opened 2021-04-19
Nouvelle facette serving_size="serving" pour data-quality


TaciteOFF:
Bonjour,

Il semblerait que quelques utilisateurs (ou applications) indiquent parfois par erreur "serving" comme taille de portion. 
Une facette pour détecter et corriger ça pourrait être utile.

Merci

teolemon:
See also #1008

stephanegigandet:
corresponding code is in DataQualityFood.pm

CharlesNepote:
* There are [more than 11,200+ products](https://mirabelle.openfoodfacts.org/products/simplified?_facet=serving_size&serving_size=serving&_facet=creator) with "serving" as `serving_size`. It is not brought by a particular app, it's just users.
* [1900 of 11,200 do not have fat value](https://mirabelle.openfoodfacts.org/products/all?_sort=rowid&_facet=serving_size&_facet=creator&_facet=data_quality_errors_tags&fat_100g__isnull=1&serving_size__exact=serving) (and probably no nutritional values at all), so `serving_size` containing "serving" should be deleted in this case
* Probably many products with "serving" have nutrition facts for 100g and not for a serving size.
* [990+ of them have a data quality error already detected](https://mirabelle.openfoodfacts.org/products?sql=--+count+number+of+products+with+a+data+quality+errors+which+serving_size+is+%22serving%22%0D%0Aselect+--+rowid%2C+code%2C+url%2C+creator%2C+created_datetime%2C+last_modified_datetime%2C+product_name%2C+serving_size%0D%0Acount%28data_quality_errors_tags%29+as+data_quality_errors_related_to_serving_size_equal_serving%0D%0Afrom+%5Ball%5D%0D%0Awhere%0D%0Aserving_size+%3D%3D+%22serving%22%0D%0Aand+data_quality_errors_tags+%21%3D+%22%22%0D%0Alimit+210), `en:nutrition-data-per-serving-serving-quantity-is-0` for nearly half of them

This would lead to increase data quality errors from 5.78% to ~6.16% for an issue that is having no impact on Nutri-Score, Nova, etc. for at least more than 10%.

So I would be in favour to:

- [ ] 1. Delete all the value "serving" in the `serving_size` as:
  - when users are often seeing this value they can think it's a good value
  - it's easier to complete an empty field than a completed one
  - it's not worse to have an empty field than a false one
- [ ] 2. Create the data quality error facet to manage the new ones

[...]

PR #8057, opened 2023-01-31
feat: add data quality warning for serving size without digit

benbenben2:
Tested on Gitpod

Related to:
https://github.com/openfoodfacts/openfoodfacts-server/issues/5163
https://github.com/openfoodfacts/openfoodfacts-server/issues/5441
https://github.com/openfoodfacts/openfoodfacts-server/issues/6473

**Questions (and answers):**
1. should it involve nutrition_data_prepared as well?
    -> **decision:** yes
2. raise quality error tag - only when nutrition_data_per or nutrition_data_prepared_per are "on" (option 1) - vs - if string without digit value is in serving_size (even if nutrition_data_per or nutrition_data_prepared_per are missing) (option 2)
    -> **decision:** all (option 2) because as stated by @CharlesNepote  "when users are often seeing this value they can think it's a good value" (https://github.com/openfoodfacts/openfoodfacts-server/issues/5163)
3. how come serving_quantity=0 when it is "serving" or any string?
    -> in `lib/ProductOpener/Food.pm`, the subroutine `normalize_serving_size` extracts any digit and return it or 0 if nothing is found. 
3.1. can this be used to setup the error?
   -> it cannot be serving_quantity (using normalize_serving_size) because serving_quantity is undefined if the value is something like "serving 1 and serving 2 and serving 3", which contain digit
4. what name to give to the error?
    -> existing one is "en:serving-size-in-mg"
    -> plural was used for "EcoScore - threatened species - Ingredients missing"
    hence, "_**en:serving-size-is-missing-digits**_" seems to be a good choice


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -659,6 +659,9 @@ sub check_nutrition_data ($product_ref) {
 					"en:nutrition-data-prepared-without-category-dried-products-to-be-rehydrated";
 			}
 		}
+		if ((defined $product_ref->{serving_size}) && (($product_ref->{serving_size} =~ /(\d+)/) eq "")) {
```
stephanegigandet, line 662: The test is better written this way I think:
- use !~ //   -> indicates it does not match
- check that the field is not an empty string
- check if there's one digit /\d/

```suggestion
		if ((defined $product_ref->{serving_size}) and ($product_ref->{serving_size} ne "") and ($product_ref->{serving_size} !~ /\d/) ) {
```

benbenben2:
Thank you @stephanegigandet
Your suggested code is working fine (tested with gitpod).
But for perltidy, no idea why it is failing (tried changing formatting, it is working with only 2 conditions but not with 3) :-(

stephanegigandet:
> But for perltidy, no idea why it is failing (tried changing formatting, it is working with only 2 conditions but not with 3) :-(

Hi @benbenben2, you need to run "make lint", and perltidy will change the indentation, split long lines etc.

stephanegigandet (approved):
Thank you!

CharlesNepote:
Didn't have time to thank you, it's great!

PR #8064, opened 2023-02-03
feat: unsaturated fat support

aleene:
### What
Added unsaturated fat to support dutch products

Maybe it is useful to create a section for products sold in the Netherlands?

However there should be a world version that contains everything.

Could be move this data to nutrients.txt?


alexgarel (approved):
Thanks.

alexgarel:
@aleene: don't forget to run "make lint_perltidy" to avoid failing on checks because of perltidy.

alexgarel:
I pushed a commit, it should be ok.

stephanegigandet (approved):

Issue #8038, opened 2023-01-26
Sessions expire everytime a new IP adress is used


raphael0202:
### Describe the bug

Every time the IP address of the device change, users on Product Opener get disconnected and have to reconnect again.
This is an important pain point for contributors, especially those that use several wifi networks or mobile devices on a 3G/4G connexion (where the IP address is not fixed).

### To Reproduce

See description

### Expected behavior

The session cookie should expire only after a specified time amount

### Type of device

Browser

PR #8066, opened 2023-02-03
fix: make user sessions work across different ip addresses 

stephanegigandet:
This fixes @raphael0202 's most wanted issue #8038

In practice, it means user sessions are now not restricted by ip address. So if the session cookie gets stolen somehow, it would be possible to use it from any ip.


raphael0202:
Thank you @stephanegigandet! How long is the cookie alive?

alexgarel (approved):
LGTM.

I don't see major threat to cookie being stolen as we are full https.

PR #8070, opened 2023-02-05
taxonomy: adding Cocamidopropyl betaine in allergens

fuddl:
I need to watch out for Cocamidopropyl betaine because I might be allergic to it. It wasn't yet listed as a allergent. [According to Wikipedia](https://en.wikipedia.org/wiki/Cocamidopropyl_betaine#Safety) the allergent status is  controversial. 🤷

### DIFF
--- a/taxonomies.obf/allergens.txt
+++ b/taxonomies.obf/allergens.txt
@@ -113,6 +113,8 @@ cas_number:en:106-22-9 / 1117-61-9 / 7540-51-4
 human_evidence:en:++
 eu_lists:en:Established contact allergens in humans
 
+en:COCAMIDOPROPYL BETAINE
+cas_number:en: 61789-40-0
 
 en:COUMARIN
 cas_number:en: 91-64-5


alexgarel:
@stephanegigandet and @teolemon I think we have a precise policy for this allergens taxonomy. How is it ?

benbenben2:
@fuddl, you have to prefix the title of your pull request by "taxonomy: " (all in lower case, it seems that the capital letter does not work)

stephanegigandet:
> @stephanegigandet and @teolemon I think we have a precise policy for this allergens taxonomy. How is it ?

For food, we limit the allergens to the 14 or so allergens that must be marked on the package.

For cosmetics, we don't have an official list to rely on. The issue is that almost all ingredients can be allergens to some people, so instead of using the allergens taxonomy, it's probably better to rely on an ingredients taxonomy, and build features so that anyone can state which ingredients they are allergic to.

stephanegigandet (approved):

stephanegigandet:
We can add it. Note that Open Beauty Facts deployments are currently frozen, as we have not deployed the new design on OBF yet.

PR #8080, opened 2023-02-08
fix: remove underlined links in tables/dropdowns

stephanegigandet:
This removes the underline for links in some dropdowns (e.g. the "Explore by" button) and tables (e.g. list of tags in /categories etc.)


alexgarel (approved):
:sweat_smile:

PR #8081, opened 2023-02-08
feat: download and convert Matomo events data for scanbot

stephanegigandet:
API calls for products scans from the new Smoothie app cannot be reliably identified from the backend nginx logs (user agent is dart:io and there's no indication if it's from our app, and from a scan action). We do have some event data in Matomo for users who opted in to send metrics, so we can use this data and feed it to scanbot.pl to populate popularity of products.


alexgarel (approved):
Great :-)

PR #8089, opened 2023-02-10
chore: taxonomy repairs

aleene:
### What
Trying to get the number of reported errors down.


stephanegigandet (approved):
Thank you @aleene !

PR #8094, opened 2023-02-11
feat: add quality facet ingredient percent over hundred

### LINKED ISSUE #2168: Quality facet - ingredient % over 100
### What
`Purée 404% (pomme de terre 20,5%, carotte 141%, eau, crème fratche (lait), sel, fibres de lin), eau, boulettes de viande bovine cutes 228% (viande bovine (Oriqine France), eau, fécule de tapioca, sel), tomate 116%, échalotes, fibres de lin, huile de colza, concentré de tomate, sel.`

Anything over 100% should be flagged as it is an OCR error.

### Code pointer
 lib/ProductOpener/DataQuality*.pm

### Part of
- #5538

benbenben2:
Resolve #2168 

 - option 1 - only in extracted ingredients / main language
there is ingredients>id
if percentage is given: "ingredients>percent"
if it is not given: "ingredients>" (the key "percent" is missing)
 - option 2 - in all languages (text)

option 1 is preferred because it is the main language that is used to analyze ingredients + option 2 would have to re-extract percentage from entire list of ingredients, which is not without risk (maybe something can be above 100% in the text but not related to ingredients)

name?
    en:ingredients- + description (similar to existing quality facets)
    en:ingredients-extracted-ingredient-from-picture-with-more-than-100-percent


stephanegigandet (approved):
Looks good, thank you. I'm not sure if it should be a warning or an error, I'm curious to see the affected products.

benbenben2:
You want it to be a warning instead of an error? I will change it later during the week.

CharlesNepote:
Looks good to me, thanks a lot Benoît. I feel that this is clearly an error and not a warning but we can wait and see the first products.

PR #8099, opened 2023-02-14
fix: ignore packaging_text if packagings already exists

### LINKED ISSUE #8097: Use packaging_text only for initialization of the packagings data structure if it does not already exist
We currently use data extracted from the packaging_text field to populate and extend the packagings data structure. This is causing issues for packaging components that may have different shape names in the packaging text and in the packagings data structure (e.g. pot vs jar).

To make sure we don't duplicate packagings component, we should only use the packagings text if we don't have existing packagings data.

stephanegigandet:
This is to avoid duplicate packaging components that are introduce if the packaging text specifies slightly different shapes than the packagings structured created by users (e.g. jar vs pot).

Fixes #8097

### DIFF
--- a/lib/ProductOpener/Packaging.pm
+++ b/lib/ProductOpener/Packaging.pm
@@ -776,54 +776,28 @@ sub set_packaging_misc_tags ($product_ref) {
 	return;
 }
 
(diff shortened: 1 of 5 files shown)


alexgarel (approved):
Perfect !

PR #8105, opened 2023-02-14
refactor: split huge function in Imports.pm in smaller huge functions

stephanegigandet:
The resulting functions are huge, but they are less huge than before. We can improve gradually. No functional change was made, just a split into functions.

Also renamed import_csv_file.t to convert_and_import_excel_file.t, in order to add another test that just import a "standard" OFF CSV file.


lib/ProductOpener/Test.pm
```diff
@@ -140,7 +140,7 @@ sub check_not_production() {
 			return get_products_collection()->count_documents({});
 		}
 	);
-	unless ((0 <= $products_count) && ($products_count < 1000)) {
+	unless ((0 <= $products_count) && ($products_count < 10000)) {
 		confess("Refusing to run destructive test on a DB of more than 1000 items\n");
```
alexgarel, line 144:
```suggestion
		confess("Refusing to run destructive test on a DB of more than 10000 items\n");
```

Consistency :wink:

alexgarel (approved):
Cool !

This is where you are happy to have tests, it ease this kind of refactoring.

lib/ProductOpener/Test.pm
alexgarel: arf, merged without it ! Not a big deal…

PR #8107, opened 2023-02-15
docs: API changelog

teolemon:
### What
docs: API changelog


stephanegigandet:
@teolemon Thanks for the PR. It's a good idea, but it will be easier if I write it myself. I filed https://github.com/openfoodfacts/openfoodfacts-server/issues/8115 for it.

PR #8108, opened 2023-02-15
taxonomy: oat milks

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -45867,6 +45867,22 @@ ciqual_food_name:en:Oat-based drink, plain
 ciqual_food_name:fr:Boisson à base d'avoine, nature
 wikidata:en:Q20972637
 
+<en:Oat milks
+en:Unsweetened natural oat milks
+fr:Boissons végétales d'avoine nature non sucrées
+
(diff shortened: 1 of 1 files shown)


taxonomies/categories.txt
```diff
@@ -72073,13 +72089,6 @@ lt:Salotos su daržovėmis
 nl:Groentesalade
 pl:Sałatki z warzywami
 
-<en:Plant-based foods
```
stephanegigandet, line 72076: Why is this removed?

moon-rabbitOFF: Because this category corresponds to nothing that exists, there are no "vegetable pasta". It could correspond to two things:
* Durum wheat pasta colored with vegetable. 
A category already exist for those products [Flavored and colored dry pastas](https://world.openfoodfacts.org/category/flavored-and-colored-dry-pastas). We could eventually create one for fresh pasta which contain higher amount of vegetable (up to 30%).
* Alternatives to wheat pasta made from chickpeas, split peas, lentils. They all have a category already too.

alexgarel (approved):

PR #8109, opened 2023-02-15
feat: data quality warning for 3 identical values in nutrition tables

### LINKED ISSUE #3554: Quality facet - Nutrition - Flag products where 3 or more nutritional values are identical
### What

- [ ] Flag products where 3 or more nutritional values are identical. This is a clear sign that someone entered bogus data.

https://dk-en.openfoodfacts.org/cgi/product.pl?type=edit&code=3760265420094

### Part of 
- #5538

benbenben2:
Resolve #3554

Data quality warning for 3 identical values and values above 1 in nutrition tables for major nutrients.

```
make lint
make checks
make cover 
```
are done and passed

Ping @CharlesNepote to review taxonomy for data quality in this PR as well.


adilwahla (approved):
looks good

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -758,6 +761,37 @@ sub check_nutrition_data ($product_ref) {
...
+				)
+			{
+				push(@major_nutriments_values, $product_ref->{nutriments}{$nid});
+			}
+
+		}
+
+		# create a hash key: nutriment value, value: number of occurence
+		foreach my $nutriment_value (@major_nutriments_values) {
+			$nutriments_values_occurences{$nutriment_value}++;
```
stephanegigandet, line 787: This works but Perl will emit a warning that we are incrementing an undefined value the first time, so it's better to check if it exists and set it to 0 otherwise first.

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -758,6 +761,37 @@ sub check_nutrition_data ($product_ref) {
...
+			if (
+				(
+					   ($nid eq 'fat_100g')
+					or ($nid eq 'saturated-fat_100g')
+					or ($nid eq 'carbohydrates_100g')
+					or ($nid eq 'sugars_100g')
+					or ($nid eq 'fiber_100g')
+					or ($nid eq 'proteins_100g')
+					or ($nid eq 'salt_100g')
+					or ($nid eq 'sodium_100g')
```
stephanegigandet, line 775: I would remove sodium, which is computed from salt. Otherwise we may get a lot of false positives when salt is 0 (thus sodium will be 0 too) and anything else is 0 (fat, sugar etc.)

stephanegigandet: ah I spoke too fast, there's a check > 1 after, so salt and sodium will be different. That's fine then.

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -758,6 +761,37 @@ sub check_nutrition_data ($product_ref) {
...
+				)
+				and ($product_ref->{nutriments}{$nid} > 1)
```
stephanegigandet, line 777: >= 1 maybe? e.g. people entering 1 everywhere

stephanegigandet (approved):
Thank you!

PR #8116, opened 2023-02-16
docs: small consistency fix in Test.pm

alexgarel:


stephanegigandet (approved):
Oops, I modified 1000 to 10000 without changing the comment. Thanks!

PR #8127, opened 2023-02-21
feat: added Croatian language taxonomy for processing

benbenben2:
WHAT?
Added Croatian language for food processing
Added 2 unit tests *
(added 1 more entry in Croatian taxonomy for processing)
(there were a couple of tests at the end in different language, there are now at the end of their respective languages => and after "make lint" it looks like a lot of things have changed in the file, sorry for that)

\* **Remark**: the case where [ingredient + processing] is a sub-ingredient in parenthesis was not yet included in the tests (like "antioksidans (ekstrakt ružmarina)"). That was a bit tricky to handle, but now it is included in the tests as well.


stephanegigandet (approved):
Thank you!

PR #8136, opened 2023-02-23
chore: a small script to export agribalyse

alexgarel:
A simple script to get a CSV of agribalyse categories.

### DIFF
--- a/scripts/snippets/agribalyse_categories_csv.py
+++ b/scripts/snippets/agribalyse_categories_csv.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+"""A small script to export agribalyse categories in a CSV using categories.full.json
+"""
+import json
+import csv
+
+datas = json.load(open("categories.full.json", "r"))
+
+
+# get agribalyse
+all_agri_cats = set(tagid for tagid, d in datas.items() for k in d.keys() if "agribalyse" in k)
+agri_cats = all_agri_cats
+
+# and children (until we have no more children)
+while agri_cats:
+    children = set(
+        tagid
+        for tagid, d in datas.items()
+        if set(d.get("parents",[])) & agri_cats
+    )
+    all_agri_cats |= children
+    agri_cats = children
+
+# get agribalyse exploring parents
+def agribalyse(tagid, depth=0):
+    data = datas[tagid]
+    code = data.get("agribalyse_food_code", data.get("agribalyse_proxy_food_code", {})).get("en")
+    if code:
+        return code, depth
+    # explore all parents and take the lowest depth
+    candidates = []
+    for p in data.get("parents", []):
+        candidate = agribalyse(p, depth + 1)
+        if candidate:
+            candidates.append(candidate)
+    candidates.sort(key=lambda c: c[1])
+    return candidates[0] if candidates else None
+
+
+# build data
+rows = []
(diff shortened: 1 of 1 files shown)


scripts/snippets/agribalyse_categories_csv.py
```diff
@@ -0,0 +1,62 @@
...
+while agri_cats:
+    children = set(
+        tagid
+        for tagid, d in datas.items()
+        if set(d.get("parents",[])) & agri_cats
+    )
+    all_agri_cats |= children
+    agri_cats = children
+
+# get agribalyse exploring parents
```
benbenben2, line 24: alternatively, you can put comment as docstring inside the function

scripts/snippets/agribalyse_categories_csv.py
```diff
@@ -0,0 +1,62 @@
...
+        continue
+    agri, depth = agribalyse(tagid)
+    rows.append({
+        "tagid": tagid,
+        "cat_en": d["name"].get("en", ""),
+        "cat_fr": d["name"].get("fr", ""),
+        "agribalyse_food_code": agri,
+        })
+
+len(rows)
```
benbenben2, line 55: Is it needed?

scripts/snippets/agribalyse_categories_csv.py
```diff
@@ -0,0 +1,62 @@
...
+    children = set(
+        tagid
+        for tagid, d in datas.items()
+        if set(d.get("parents",[])) & agri_cats
+    )
+    all_agri_cats |= children
+    agri_cats = children
+
+# get agribalyse exploring parents
+def agribalyse(tagid, depth=0):
```
benbenben2, line 25: I would use function annotation there: https://peps.python.org/pep-0008/#function-annotations
`def agribalyse(tafid: srtr, depth: int = 0)  -> (str, int): `
and then, return a str and a int all the time

scripts/snippets/agribalyse_categories_csv.py
```diff
@@ -0,0 +1,62 @@
...
+            candidates.append(candidate)
+    candidates.sort(key=lambda c: c[1])
+    return candidates[0] if candidates else None
+
+
+# build data
+rows = []
+for tagid in sorted(all_agri_cats):
+    d = datas[tagid]
+    if not ({"en", "fr"} & set(d["name"].keys())):
```
benbenben2, line 44: Isn't it "in" instead of "&"?

scripts/snippets/agribalyse_categories_csv.py
alexgarel: Non it's just a check :-)  I was running this in a console, really…

scripts/snippets/agribalyse_categories_csv.py
alexgarel: I using the intersection of two sets.

alexgarel:
@benbenben2  thanks a lot for reviewing. That said, it's a "snippet" (but I prefer to keep it in scripts/snippet rather than in github). So if you don't mind I wont fix "style". It's just to keep it for a time where we might need it again.

alexgarel:
@benbenben2 that said your approval would enable merging this :-D

benbenben2 (approved):

PR #8140, opened 2023-02-25
taxonomy: improve taxonomy for products available in Croatia

benbenben2:
updated taxonomies for products found in Croatia (include Serbian, Bosnian, Slovenian, German, etc. languages)

### DIFF
--- a/docs/reference/schemas/product.yaml
+++ b/docs/reference/schemas/product.yaml
@@ -1959,10 +1959,16 @@ Product:
                   type: string
     serving_quantity:
       type: string
+      description: |
+        Serving quantity entered 
+        by contributors. It must 
+        contain digits and known unit.
     serving_size:
       type: string
       description: |
-        Serving size in g (or ml).
+        Serving size in g (or ml). 
+        It is extracted from 
+        serving_quantity.
     serving_size_imported:
       type: string
     sortkey:
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -150,7 +150,8 @@ my %may_contain_regexps = (
 	en =>
 		"it may contain traces of|possible traces|traces|may also contain|also may contain|may contain|may be present",
 	bg => "продуктът може да съдържа следи от|може да съдържа следи от|може да съдържа",
-	cs => "může obsahovat",
+	bs => "može da sadrži"
+	cs => "může obsahovat", "může obsahovat stopy",
 	da => "produktet kan indeholde|kan indeholde spor af|kan indeholde spor|eventuelle spor|kan indeholde|mulige spor",
 	de => "Kann enthalten|Kann Spuren|Spuren",
 	es => "puede contener huellas de|puede contener trazas de|puede contener|trazas|traza",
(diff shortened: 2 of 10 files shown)


tests/unit/additives.t
```diff
@@ -668,7 +668,7 @@ is_deeply(
 
 $product_ref = {
 	lc => "fr",
-	ingredients_text => "huile de colza, orthophosphates de calcium, carbonate de calcium, citrates de potassium"
+	ingredients_text => "huile de colza, phosphates de calcium, carbonate de calcium, citrates de potassium"
```
stephanegigandet, line 671: Why did you change this test?

benbenben2:
> Why did you change this test? @stephanegigandet 

- **Long answer:**

After adding so many entries in the taxonomies I encountered couple of errors regarding duplicated entries (for example, an entry in additives that was already in minerals).
Then, I worked on solving duplicated entries that were listed in perl unit tests (GitHub workflow). 

In the logs there were also duplicates for entries not related to the changes I did (some fr and fi entries, mainly. I remember for vitamin c for these two languages that are in both additives and in vitamins)
And that was a challenge because some of those were also used in unit tests.
I tried to update unit tests according to the changes that I did to remove duplicates. 
At some point there were failures on both **unit test** and **perl unit test** github workflows, or when it was not on one workflow it was failing on the other one. Since I could not find out a proper way to fix those,(1)(2) I rolled back the changes for entries not related to the changes I did. 

Phosphate de calcium was in minerals and ingredients and that is good that you noticed that I forgot these two.

(1) could it be that one of the workflow (unit test) is related to the taxonomies files whereas the other workflow (perl unit test) is related to something else like ingredients.all ?
(2) why the duplicates that I created resulted into workflow failure whereas other duplicates (which appears in the logs at each pull request: checks, pull requests, perl unit tests, raw logs) are not resulting into failure? 

**- Short answer:** 

I messed up. I just fixed that.

**- Rolled back changes**

In any case, I rolled back changes as follow:
-  removed comment In minerals: 
replaced:
```
 # commented because of duplicate with en:phosphate in ingredients.txt
# synonyms:fr:orthophosphate, phosphate
```
by:
```
synonyms:fr:orthophosphate, phosphate
- put back the test as it was before.
```
-  removed comment In additives.t: 
replaced:
`	ingredients_text => "huile de colza, phosphates de calcium, carbonate de calcium, citrates de potassium"`
by:
`	ingredients_text => "huile de colza, orthophosphates de calcium, carbonate de calcium, citrates de potassium"`


I guess other duplicates entries not related to the changes I did could be addressed in different pull request.
Same goes for schema description.

stephanegigandet:
@benbenben2 So for ingredients, it is expected to have duplicates, as some ingredients can be both minerals and additives for instance. When we concatenate them into ingredients.all, there will be duplicates. But we ignore duplicates for ingredients, in the sense that they won't make the tests fail.

benbenben2:
I don't know, I misunderstood the logs after the failed tests. If you have time you can have a look at it (you can see the logs by scrolling here above, between "_Merge branch 'main' into taxonomy_HR_ingredients_20230221_" and "a69971f", if you click on the red cross, then "_Details_" next to "_Pull Requests / Perl unit tests (pull_request)_") and tell me what was wrong because I am a bit curious and would like to understand. If you don't have time let's forget about it.

Anyway, I applied changes since Yesterday. Let me know if something else need to be changed.

stephanegigandet (approved):
Looks good, thank you!

PR #8144, opened 2023-02-27
taxonomy: add :de: alias for Creme Fraiche without the accents

Strubbl:
### What

add :de: alias for Creme Fraiche without the accents e.g. as seen in ingredient list of https://de.openfoodfacts.org/produkt/4008366009763/schlemmerfilet-brokkoli-mandel-frosta

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -5037,7 +5037,7 @@ af:Crème fraîche
 ca:crema de llet
 cs:Crème fraîche
 da:Cremefraiche, creme fraiche
-de:Crème fraîche
+de:Crème fraîche, creme fraiche
 eu:Esne-gain gordina, esne-gain freskoa
 fi:ranskankerma, crème fraîche
 fr:crème fraîche épaisse, crème fraîche fermentée


benbenben2:
The ingredient in the product that you linked is "Créme fraîche" (with some accents, and it is also an unknown ingredient). Hence, I would suggest to add also "Créme fraîche" among the synonyms. What do you think?

Strubbl:
The problem of "Créme fraîche" vs. "Creme Fraiche" is that it is hard to type the french version on a German keyboard. So yes, adding a synonym sounds good to me. How can i add it to the synonyms?

stephanegigandet:
"Créme fraîche" (as written on the package) is actually a mispelling (should be "Crème fraîche). It's not a good idea to add mispellings in the taxonomy (even if some of them can be found on products). Instead we can improve the parser to match some mispellings.

For languages like German that use foreign words that are frequently listed / typed without accents, I think it's fine to add the unaccented version as a synonym as done in this PR.

Strubbl:
Thanks for explaining the accent spelling mistake. I just checked the packaging from the same product which i have at home at the moment. It is fixed in the meantime. I need to update the picture soon.

hangy (approved):

stephanegigandet (approved):

PR #8149, opened 2023-02-28
fix: changing orgs for pro moderators 

### LINKED ISSUE #8125: Produceur platform does not work for moderators (failing on org selection)
### Describe the bug

If you are a moderator on the pro plateform (and not an admin),
as you type a org name you want to reach on home page,
this fails (with different error messages).

### To Reproduce

1. create a new user on pro platform (new edition, aka .dev for now), logout
2. login as an admin, and make this user moderator, logout
3. log back in with new user
4. try to reach an org, see it fail with a strange error message

### Expected behavior

This should display products from that org to the moderator.

My guess: we have a wrong condition on a "if" in user.pl.

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Browser

### Browser version

_No response_

_No response_

### Time per product

_No response_

stephanegigandet:
fixes #8125

### DIFF
--- a/cgi/user.pl
+++ b/cgi/user.pl
@@ -116,7 +116,7 @@ if ($action eq 'process') {
 	}
 
 	# change organization
-	if ($type eq 'edit_owner' && $admin) {
+	if ($type eq 'edit_owner') {
 		# only admin and pro moderators can change organization freely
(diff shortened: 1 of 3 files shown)


alexgarel (approved):
LGTM

PR #8153, opened 2023-03-01
fix: removed underline of links in attribute cards

stephanegigandet:
small fix to remove underline on links in attributes

image


yuktea (approved):
looks good!

PR #8154, opened 2023-03-01
taxonomy: add packaging material code in canonical name

stephanegigandet:
I'm proposing to add the packaging material code in the main names of packaging material entries, so that it is easier to identify them.

This is useful in particular when entering materials for a product. The user sees a 1 or 01 in a triangle, he or she can type it, get suggestions for "PET - 1 - Polyethylene terephthalate" and validate it. We could make suggestions work differently (by matching on the xx: entries and synonyms, which we should also do), but if we just suggest "PET - Polyethylene terephthalate", the user cannot get a sense of whether it is a correct entry or not.

Note that this change will change the canonical entry for packagings. It also makes us repeat the local name 3 times in each language:

[abbrevation] - [number] - [name], [abbrevation] - [name], [name]

### DIFF
--- a/taxonomies/packaging_materials.txt
+++ b/taxonomies/packaging_materials.txt
@@ -109,16 +109,16 @@ en:Mixed plastics
 fr:Plastiques mixtes
 
 <en:Plastic
-en:PET - Polyethylene terephthalate, Polyethylene terephthalate
+en:PET - 1 - Polyethylene terephthalate, PET - Polyethylene terephthalate, Polyethylene terephthalate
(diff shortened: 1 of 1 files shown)


alexgarel:
LGTM. 

For PEHD in Frenche it would be cool to also have HDPE. Maybe English abbreviations might be added in all languages because they are commonly used in many countries.

stephanegigandet:
This change was incorporated in another PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/8190

PR #8155, opened 2023-03-01
taxonomy: added povratna naknada label and its image

benbenben2:
1-
New label with image.

2-
Fixed horrible typo for "hr:EU Organic".


stephanegigandet (approved):
Thank you!

PR #8157, opened 2023-03-02
chore: Week 9 additions

aleene:
### What
Some requested additions and some apple juice work


aleene:
@stephanegigandet I was able to remove a lot of test errors, but I do not see where this one originates.

stephanegigandet:
@aleene I'm looking into the failing tests

stephanegigandet:
@aleene "Orange marmelade" is marked "vegan" because it is a child of "Orange".

<en:orange
en:orange marmalade

I think we should remove the parent (same for other kinds of marmalade).  "Orange marmelade" is not a type of "Orange".

stephanegigandet (approved):
Looks good

PR #8161, opened 2023-03-05
taxonomy: Update ingredients in Catalan and Spanish

duhow:
### What

Update taxonomy languages Catalan and Spanish.


stephanegigandet (approved):
Thank you!

Issue #8168, opened 2023-03-07
Displaying products sorted by creation date timeout in Mongo


stephanegigandet:
Users have reported that home page does not load for them, it was caused by a timeout in MongoDB.

Adding an index seems to have fixed the problem:

```
> db.products.dropIndex("ciqual_food_name_tags_1_last_modified_t_-1")
{ "nIndexesWas" : 64, "ok" : 1 }
> db.products.createIndex({countries_tags: 1, created_t: -1}, { background: true})
{
	"createdCollectionAutomatically" : false,
	"numIndexesBefore" : 63,
	"numIndexesAfter" : 64,
	"ok" : 1
}
```

Note that there is already an index on created_t alone, but for some reason it probably was not used. Maybe because it is in ascending order... but it should not matter. :-| 

```
	{
		"v" : 2,
		"key" : {
			"created_t" : 1
		},
		"name" : "created_t_1",
		"background" : true
	}
```

PR #8169, opened 2023-03-07
fix: mongodb index for countries_tags + created_t 

stephanegigandet:
Added a new index to solve #8168

As we already have 64 indexes, I removed one for amino_acids which hopefully is not much useful


alexgarel (approved):
LGTM

PR #8171, opened 2023-03-07
taxonomy: add some labels commonly found in Croatia

benbenben2:
added labels commonly found in Croatia, most of them Croatian.
not sure about zoi, zozp and zts
removed some labels (Croatian-related) that were not assigned to any product.

Fixed path for EG-Öko-Verordnung that was incorrect (en instead of de)
Removed strange entry for en:German Agricultural Society in Portuguese (pt:NL-BIO-01)


benbenben2:
Hello @alexgarel 
```
obtained taxonomy for categories from GitHub cache.
[...] Could not load taxonomy: /mnt/podata/taxonomies/categories.result.sto at lib/ProductOpener/Tags.pm line 2057.
```
Is the error related to the cache for the taxonomy that need to be rebuild because the changes were done in a specific taxonomy (categories)? (https://openfoodfacts.github.io/openfoodfacts-server/explanations/taxonomy_build_cache/)

If yes, I am not sure to know how to troubleshoot it. If we need to force to rebuild the taxonomy, how do we do that?
And if no, then, I have no idea...

stephanegigandet (approved):
Thank you!

PR #8175, opened 2023-03-08
fix: remove File::Fetch

### LINKED ISSUE #8172: In production, some accents are broken with UTF8 sequences shown
This happened on off1 after deploying a new version:

image

stephanegigandet:
File::Fetch is causing issues in the prod and dev environment (but not in docker, maybe because of different versions).

I'm removing it in favor of LWP::UserAgent (which we already use).

fixes #8172


alexgarel (approved):
Good idea !

PR #8177, opened 2023-03-09
docs: fix open api documentation

alexgarel:
And remove the API check because it gives a false positive

Also slighten the product example a bit.

Right now OpenAPI docs is badly broken (`Product: {missing-type-info}` in schema on https://openfoodfacts.github.io/openfoodfacts-server/reference/api/#get-/api/v2/product/-barcode-), this should fix it.


alexgarel:
@sgtSeme4ki, sadly I had to disable the validation of OpenAPI because it creates a false positive and documentation was indeed broken.

alexgarel:
As soon as we merge this branch I should report the bug to openapitools.

stephanegigandet (approved):

PR #8178, opened 2023-03-09
chore: weekly taxonomy updates

aleene:
### What
Some updates to various taxonomies. Mainly inspired by the apricot nectars category.


aleene:
@stephanegigandet how is this supposed to work? I can find the expected test results and the integration test, but where is this defined (taxonomy)?

stephanegigandet:
> @stephanegigandet how is this supposed to work? I can find the expected test results and the integration test, but where is this defined (taxonomy)?

The suggestions are sorted by popularity for the category of product, it's not hardcoded in the taxonomy. The test result changed because the product category is mango juice, which we used to recognize. So we should change the test to use a category that exists.

One alternative would be to add "mango juices" as a synonym of "mango nectars".

stephanegigandet (approved):
Thank you!

PR #8190, opened 2023-03-13
feat: Match synonyms and xx: entries when computing taxonomy suggestions

stephanegigandet:
Previously, taxonomy suggestions were only returned when the canonical taxonomy entry matched the input string from the user. Now we also try to match the input string to synonyms of the entry, including in the xx: wildcard language.

Also added a unit test to ease debugging.


tests/integration/expected_test_results/api_v3_taxonomy_suggestions/categories-term-strawberry.json
```diff
@@ -2,6 +2,7 @@
    "errors" : [],
    "status" : "success",
    "suggestions" : [
+      "Mint-flavoured syrup with sugar diluted in water",
```
stephanegigandet, line 5: This entry is because we have a weird CIQUAL category that applies to both strawberry and mint flavoured syrups... To be fixed in the taxonomy.

tests/integration/expected_test_results/api_v3_taxonomy_suggestions/categories-string-strawberry.json
```diff
@@ -2,6 +2,7 @@
    "errors" : [],
    "status" : "success",
    "suggestions" : [
+      "Mint-flavoured syrup with sugar diluted in water",
```
stephanegigandet, line 5: This entry is because we have a weird CIQUAL category that applies to both strawberry and mint flavoured syrups... To be fixed in the taxonomy.

tests/integration/expected_test_results/api_cgi_suggest/categories-string-strawberry.json
```diff
@@ -1,4 +1,5 @@
 [
+   "Mint-flavoured syrup with sugar diluted in water",
```
stephanegigandet, line 2: This entry is because we have a weird CIQUAL category that applies to both strawberry and mint flavoured syrups... To be fixed in the taxonomy.

alexgarel (approved):
Cool ! That will help a lot.

I added some remarks.

lib/ProductOpener/TaxonomySuggestions.pm
```diff
@@ -333,45 +377,45 @@ sub filter_suggestions_matching_string ($tags_ref, $tagtype, $search_lc, $string
...
+			# matching inside
+			elsif ($best_match eq "inside") {
+				push @suggestions_c, $tag;
+			}
+			# fuzzy match
+			elsif ($best_match eq "fuzzy") {
+				push @suggestions_f, $tag;
+			}
 		}
 	}
```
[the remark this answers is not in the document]
stephanegigandet: Why not. In the worse case scenario (huge taxonomy like categories, and a text query that does not match anything so we have to go through everything, it takes about 0.5 seconds in prod). I'm not sure we will gain much on smaller taxonomies, but we only have a bit of space to lose.

Another option is to use memcached (so that the cache is shared between processes), but then it might be tricky when we update things and get old cache results even after restarting Apache.

lib/ProductOpener/Ecoscore.pm
```diff
@@ -472,24 +472,79 @@ sub load_ecoscore_data_packaging() {
...
+			if (not exists_taxonomy_tag("packaging_materials", $target_material)) {
+				die(      "target_material "
+						. $assignment_ref->{target_material}
+						. " does not exist in the packaging_materials taxonomy");
+			}
+			if (not exists_taxonomy_tag("packaging_materials", $source_material)) {
+				die(      "source_material "
+						. $assignment_ref->{source_material}
+						. " does not exist in the packaging_materials taxonomy");
+			}
```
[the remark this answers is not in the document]
alexgarel: cool :-)

lib/ProductOpener/TaxonomySuggestions.pm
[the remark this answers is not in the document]
alexgarel: Oh yes, I didn't remember we where using multiple processes (and quite a lot) so memcached would really be needed. Only thing is that I didn't see ready to use memoize using memcached with a RLU strategy (MRU would be even better).

lib/ProductOpener/Ecoscore.pm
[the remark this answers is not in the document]
stephanegigandet: yes, I'm still addressing the other suggestions

lib/ProductOpener/TaxonomySuggestions.pm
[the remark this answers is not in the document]
stephanegigandet: Right, I chose to use memcached, I'm refactoring some stuff around it.

stephanegigandet:
@alexgarel the PR is ready for review again, I added some caching, did a bit of refactoring and added more tests

lib/ProductOpener/Tags.pm
```diff
@@ -1797,7 +1797,7 @@ Build all taxonomies
 =cut
 
 sub build_all_taxonomies ($publish) {
-	foreach my $taxonomy (@taxonomy_fields) {
+	foreach my $taxonomy (@taxonomy_fields, "test") {
```
alexgarel, line 1800: :+1:

lib/ProductOpener/TaxonomySuggestions.pm
```diff
@@ -97,17 +98,56 @@ Hash of fields that can be taken into account to generate relevant suggestions
 - categories: comma separated list of categories (tags ids or strings in the $search_lc language)
 - shape: packaging shape (tag id or string in the $search_lc language)
 
+=head3 Note
+
+The results of this function are cached using memcached. Restart memcached if you want fresh results
```
alexgarel, line 103: Shouldn't we also add a time to live to our cached entries ? 1 day seems enough.

So right now as we use it for mongodb we might have old entries served ?

Memcached has built-in support for that in set you can add an expiration time.

lib/ProductOpener/Cache.pm
```diff
@@ -46,7 +48,37 @@ $memd = Cache::Memcached::Fast->new(
...
+
+sub generate_cache_key ($context_ref) {
+
+	# We generate a sorted JSON so that we always have the same key for the context object
+	# even if it contains hashes (Storable::freeze may not have the same order of keys)
+	my $key = $server_domain . "/" . $json->encode($context_ref);
+	my $md5_key = md5_hex($key);
+	$log->debug("generate_cache_key", {context_ref => $context_ref, key => $key, md5_key => $md5_key})
+		if $log->is_debug();
+	return $md5_key;
```
alexgarel, line 81: I find it a bit dangerous. Collision may happens.
They are not maybe a big issue if we at least prefix the key by - a domain + a function.

So I would associate each domain with a single character, same for each function, and then encode the rest of $context_ref with md5.

I'm not sure for the domain prefix for there might be too much case to handler, don't we have a variable saying the "product-family" ?  

```perl
SERVER_DOMAINS_PREFIX = {"openfoodfacts.org" => 'a', …}

FN_PREFIX = {"suggestion_cache" => 'a', "mongo_query" => 'b'}

sub generace_cache_key($fn, $context_ref) {
   $key = SERVER_DOMAINS_PREFIX->{$server_domain} . FN_PREFIX->{$fn};
   ...
   $key .= $md5_key
}
```

alexgarel: I see below that you did it indeed, by adding a prefix… but in this case, let's add the prefix as a parameter to this function, it will be more clear.

And ok then for the domain in the hash (the most important thing is to have a "result type" prefix.

lib/ProductOpener/Ecoscore.pm
```diff
@@ -472,24 +476,105 @@ sub load_ecoscore_data_packaging() {
...
+			if (not $target_material_id_exists_in_taxonomy) {
+				die(      "target_material "
+						. $assignment_ref->{target_material}
+						. " does not exist in the packaging_materials taxonomy");
+			}
+			if (not $source_material_id_exists_in_taxonomy) {
+				die(      "source_material "
+						. $assignment_ref->{source_material}
+						. " does not exist in the packaging_materials taxonomy");
+			}
```
alexgarel, line 534: Although the exists in taxonomy parameter make sense, I don't think you get my point with the refactoring. For me it was more that you are doing 4 times the same thing…

I would have reduced that to:

```perl
  my $target_material = get_canonical_or_die($assignement_ref, "target_material", "packaging_material");
  my $source_material  = get_canonical_or_die($assignement_ref, "source_material", "packaging_material");
  my $target_shape = get_canonical_or_die($assignement_ref, "target_shape", "packaging_shapes");
  my $source_shape = get_canonical_or_die($assignement_ref, "source_shape", "packaging_shapes");
```

stephanegigandet: Sounds good, I added a function like that. I did the other because there are many other places in the code where we canonicalize a tag and then check its existence in the taxonomy (e.g. for all the parsing of ingredients lists etc.)

lib/ProductOpener/Cache.pm
stephanegigandet: ok, I made the key server domain + result type + hash of context, directly in the generate_cache_key function

alexgarel (approved):
Perfect !

Issue #8188, opened 2023-03-13
<p> tags are display none in the CSS for Static pages


teolemon:
### What
<p> tags are display none in the CSS for Static pages
I was beginning to doubt my sanity during deployment, when I realized ```< p >``` tags are display none in the CSS for Static pages
![image](https://user-images.githubusercontent.com/1689815/224719645-f5c656ca-0d4b-433d-bcb1-588a39b42d50.png)

#### Happens on the NOVA page
https://world.openfoodfacts.org/nova

#### Does not happen on the Nutri-Score page
https://world.openfoodfacts.org/nutriscore

```
#main_column>h1, #main_column>p, .top-bar-section .has-dropdown.moved>a:after {
    display: none;
}
```

alexgarel:
It was introduced by https://github.com/openfoodfacts/openfoodfacts-server/pull/7703

alexgarel:
It was introduced by new design commit : 7e52e7a8f97

I'm not sure why this rule exists …

alexgarel:
Fixed by https://github.com/openfoodfacts/openfoodfacts-server/pull/8192

PR #8192, opened 2023-03-13
fix: remove css rule hiding legit content on pages

alexgarel:
Remove CSS rule that has no obvious intent, and that hides legit content on pages like "NOVA" explanation.

fixes:
- #8188


stephanegigandet (approved):

teolemon (approved):
Thanks @alexgarel . Hopefully, no horrendous side effect to the rule removal

alexgarel:
@teolemon you can test on .net

PR #8205, opened 2023-03-14
fix: Nutri Score Nutriments with an asterisk

### LINKED ISSUE #7610: In Edit Mode, indicate the required nutriments to compute the Nutri-Score by an asterisk
### What
- In Edit Mode, indicate the required nutriments to compute the Nutri-Score by an asterisk
### Part of
- #Nutri-Score tracker
- #road to score

MonalikaPatnaik:
### What
So, to calculate nutriscore-- fats,saturated-fats, sugar and salt are the required nutriments to be considered, so I have made necessary changes in the code to show these nutriments with an asterisk in the nutrition table:)

### Screenshot
Screenshot 2023-03-14 205358

- it fixes #7610


templates/web/pages/product_edit/product_edit_form_display.tt.html
```diff
@@ -207,7 +207,11 @@ <h1>[% title %]</h1>
                                     <td>
                                         <!--label starts-->
                                         [% IF nutriment.name.defined %]
-                                            <label class="nutriment_label" for="nutriment_[% nutriment.enid %]">[% nutriment.prefix %][% nutriment.name %]</label>
+                                        [% IF nutriment.nid == 'fat' || nutriment.nid == 'saturated-fat' || nutriment.nid == 'salt'|| nutriment.nid =='sugars' %]
+                                            <label class="nutriment_label" for="nutriment_[% nutriment.enid %]">[% nutriment.prefix %][% nutriment.name %]*</label>
+                                        [% ELSE %]
+                                        <label class="nutriment_label" for="nutriment_[% nutriment.enid %]">[% nutriment.prefix %][% nutriment.name %]</label>
+                                        [% END %]
```
jnsereko, line 214:
```suggestion
                                        [% ELSE %]
                                            <label class="nutriment_label" for="nutriment_[% nutriment.enid %]">[% nutriment.prefix %][% nutriment.name %]</label>
                                        [% END %] 
```

cc @yuktea @stephanegigandet

jnsereko (changes requested):
Thanks for working on this @MonalikaPatnaik 
fix some formatting issue though

stephanegigandet:
Hi @MonalikaPatnaik , thanks for the PR!

If we put a *, we need to explicitly tell users what this * means, otherwise there is no way to guess.
Could you add a string after the table that states "*: used to compute the Nutri-Score" or similar?

That string would need to be translatable, so you would need to use the lang() function, and add it in the translation files: po/common/common.pot + po/common/en.po

templates/web/pages/product_edit/product_edit_form_display.tt.html
```diff
@@ -207,7 +207,11 @@ <h1>[% title %]</h1>
                                     <td>
                                         <!--label starts-->
                                         [% IF nutriment.name.defined %]
-                                            <label class="nutriment_label" for="nutriment_[% nutriment.enid %]">[% nutriment.prefix %][% nutriment.name %]</label>
+                                        [% IF nutriment.nid == 'fat' || nutriment.nid == 'saturated-fat' || nutriment.nid == 'salt'|| nutriment.nid =='sugars' %]
```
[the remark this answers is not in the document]
MonalikaPatnaik: > Hi @MonalikaPatnaik , thanks for the PR!
> 
> If we put a `*`, we need to explicitly tell users what this `*` means, otherwise there is no way to guess. Could you add a string after the table that states "`*`: used to compute the Nutri-Score" or similar?
> 
> That string would need to be translatable, so you would need to use the lang() function, and add it in the translation files: po/common/common.pot + po/common/en.po

Thanks for the review..will make necessary changes :)

po/common/en.po
```diff
@@ -1709,6 +1709,10 @@ msgid "The table lists by default nutriments that are often specified. Leave the
 msgstr "The table lists by default nutriments that are often specified. Leave the field blank if it's not on the label.<br/>You can add extra nutriments (vitamins, minerals, cholesterol etc.)\n"
 "by typing the first letters of their name in the last row of the table."
 
+msgctxt "nutrition_data_table_asterisk"
+msgid "* : essential nutiments to calculate nutri-score."
```
stephanegigandet, line 1713:
```suggestion
msgid "essential nutrients to calculate the Nutri-Score."
```
It's better to keep the * out of the string. You can put it in the template instead, like this:

A few comments:

`*[% sep %]:[% lang('nutrition_data_table_asterisk') %]`

The [% sep %] is used to put a space in some languages that need a space before the : character (e.g. French but not English).

We decided to use the word "nutrient" instead of the lesser used "nutriment".
"Nutri-Score" is a trademark, it must be capitalized that way.

MonalikaPatnaik: Okay...fixing it!


> It's better to keep the * out of the string. You can put it in the template instead, like this:
> 
> A few comments:
> 
> `*[% sep %]:[% lang('nutrition_data_table_asterisk') %]`
> 
> The [% sep %] is used to put a space in some languages that need a space before the : character (e.g. French but not English).
> 
> We decided to use the word "nutrient" instead of the lesser used "nutriment". "Nutri-Score" is a trademark, it must be capitalized that way.

Okay...fixing it 👍

yuktea (approved):
tested it out 👍 
you should probably remove the extra changes that got staged

MonalikaPatnaik:
> tested it out 👍 you should probably remove the extra changes that got staged

Okay..Thanks<3

MonalikaPatnaik:
Heyy..done the req changes..check once!! Do I have to squash the commits into one ?

templates/web/pages/product_edit/product_edit_form_display.tt.html
```diff
@@ -59,6 +60,9 @@ <h1>[% title %]</h1>
                 <p id="barcode_paragraph"> [% lang("barcode") %][% sep %]:
                     <span id="barcode" property="food:code" itemprop="gtin13" style="speak-as:digits;">[% code %]</span>
                 </p>
+                <p id="barcode_paragraph"> [% lang("barcode") %][% sep %]:
```
stephanegigandet, line 63: same here, strings have been added twice

taxonomies/nutrient_levels.txt
```diff
@@ -117,7 +117,7 @@ nb:Fett i liten kvantitet
 nd:Fat in low quantity
 ne:Fat in low quantity
 ng:Fat in low quantity
-nl:Vetten in kleine hoeveelheid
+nl:Vetten in low quantity
```
stephanegigandet, line 120: can you remove the changes in this file? they are unrelated...

you can do "git checkout taxonomies/nutrient_levels.txt" to get back the original file

po/common/common.pot
```diff
@@ -1624,6 +1624,11 @@ msgid ""
 "by typing the first letters of their name in the last row of the table."
 msgstr ""
 
+msgctxt "nutrition_data_table_asterisk"
+msgid ""
```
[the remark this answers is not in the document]
yuktea: I might be missing something, can you throw some light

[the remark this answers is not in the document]
MonalikaPatnaik: Screenshot 2023-03-17 120325
 
ig this answers your question..and if you say I'll remove this extra quotes as it is not going to change anything overall :) (I just added for the safer side to keep the alignment neater..idk if this is req or not )

MonalikaPatnaik:
> Thank you for making the changes to it Monalika. Like I said in the previous comment, you should remove the unrelated changed taxonomy file.

Yup..I have discarded the changes and commited the previous state of the file only

yuktea:
> Yup..I have discarded the changes and commited the previous state of the file only
I don't think it should be a problem. but it's still showing up in your changed file diff even when there is no apparent change. 

I'd say as a general rule, only stage those files that you need to commit. 🙂

po/common/common.pot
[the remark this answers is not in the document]
yuktea: Thank you!

yuktea:
Above is just a nit-pick. This should be good to merge I believe. Thank you so much! 💯

MonalikaPatnaik:
> 



> > Yup..I have discarded the changes and commited the previous state of the file only
> 
> I don't think it should be a problem. but it's still showing up in your changed file diff even when there is no apparent change. I'd say as a general rule, only stage those files that you need to commit. 🙂
> 
> Edit: you might have to do a force-push to get that file out from changed files. but I really don't think it matters here since there seems to be no apparent change.

Okay..I'll keep this in mind..if u say I could squash all my commits into one having the necessary changed files only!

yuktea (approved):

stephanegigandet (approved):
Looks good, thank you for addressing all the comments @MonalikaPatnaik!

PR #8216, opened 2023-03-16
fix: increase conversion to the Pro Platform by making everything a link

teolemon:
### What
fix: increase conversion to the Pro Platform by making everything a link


stephanegigandet:
Probably not needed as we now underline the link and having a big line link is usually not good idea.

image

PR #8219, opened 2023-03-16
fix: underline links in panel text

stephanegigandet:
Links were almost invisible in the text of panels, this PR adds an underline. (related: #8216)

I believe links in general should be more obvious (e.g. with a blue color), but that's for another PR.

image

### DIFF
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -1425,6 +1425,12 @@ a.panel_title {
 .accordion-navigation a {
   text-decoration:none !important;
 }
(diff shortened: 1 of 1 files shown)


alexgarel (approved):
Nice. 

But I'm against blue color if it breaks the color palette :-D

stephanegigandet:
> But I'm against blue color if it breaks the color palette :-D

Well we'll need some color that is much more different than dark grey vs black ;)

PR #8222, opened 2023-03-17
fix: delete old logo of a-market

manon-corneille:

- Fixes #[ISSUE NUMBER]


stephanegigandet:
It turns out there are different characters: ä-märket <=>  ä-märket (which may or may not look the same on your system!)

image

stephanegigandet (approved):
We will need a better solution when we tackle https://github.com/openfoodfacts/openfoodfacts-server/issues/8225

PR #8223, opened 2023-03-17
fix: remove extra barcode in product edit form

stephanegigandet:
Removes the barcode at the top (which was previously hidden with a display:none css rule that we removed)

image


alexgarel (approved):
:sunglasses:

PR #8224, opened 2023-03-17
fix: open in Gitpod and mongo shell docs links should point to right resources

jnsereko:
### What
just simply made sure that docs reference to correct resources 

- Fixes #[ISSUE NUMBER]


jnsereko:
cc @yuktea @alexgarel @stephanegigandet

stephanegigandet (approved):
Thank you!

Issue #8209, opened 2023-03-15
incron does not exists in Debian Bullseye


alexgarel:
We currently rely on incron to send images to OCR. See `scripts/process_new_image_off.sh`.

But it is not available in Bullseye.From [this thread](https://github.com/piku/piku/issues/220), we can understand it was removed because of a [core dump issue](https://github.com/ar-/incron/issues/36). 

The piku thread also propose to move to [systemd path](https://www.freedesktop.org/software/systemd/man/systemd.path.html), but we could move to a classical [inotify listener script](https://metacpan.org/pod/Linux::Inotify2).

I think first option is good to explore first, as it might be more robust and easy to setup in prod.

alexgarel:
I did some search and did not find any good alternative: fswatch is cumbersome, inotify-tools does not provide a tool for that use case, inotify-hookable, entr, iwatch or systemd-path do not provide changed file as arguments.

On the other side, writing a script using the [Linux::Inotify2](https://metacpan.org/pod/Linux::Inotify2) is quite straight forward and would have the advantage of not reloading perl interpreter, and all taxonomy, etc. each time. So much more efficient.

But first, I would like to verify with @stephanegigandet if this thing is really needed !

stephanegigandet:
@alexgarel Instead of watching the file system to see if a new image is available, an alternative would be to actively add the image to a queue (e.g. in a minion queue, or a redis list or stream), and then have a separate service call the Google Cloud Vision API.

alexgarel:
Ok, I think for now we can inotify2, because we will have to do that also for sibling projects" so it's easier to just keep same approach.

Later we can go for redis streams.

stephanegigandet:
Sounds good.

PR #8234, opened 2023-03-20
feat: replacing incron with a specific script

alexgarel:
Replaces incron with a specific script.

Fixes:
- #8209


lib/ProductOpener/Test.pm
```diff
@@ -266,7 +266,7 @@ sub ensure_expected_results_dir ($expected_results_dir, $update_expected_results
 	return 1;
 }
 
-=head2 compare_to_expected_results($object_ref, $expected_results_file, $update_expected_results) {
+=head2 compare_to_expected_results($object_ref, $expected_results_file, $update_expected_results, $test_ref = undef) {
```
stephanegigandet, line 269: $test_ref is currently not used

lib/ProductOpener/Images.pm
```diff
@@ -1996,93 +1996,146 @@ sub extract_text_from_image ($product_ref, $id, $field, $ocr_engine, $results_re
...
 
-			open(my $OUT, ">>:encoding(UTF-8)", "$data_root/logs/cloud_vision.log");
-			print $OUT "success\t" . $image_url . "\t" . $res->code . "\n";
-			close $OUT;
+	my $api_request_ref = {
+		requests => [
+			{
+				features => [
+					{type => 'TEXT_DETECTION'},
+					{type => 'LOGO_DETECTION'},
```
stephanegigandet, line 2048: for cropped images for ingredients and nutrition, we just need the text detection, maybe we could add an extra parameter to send_image_to_cloud_vision() so that we can state which detection we want.

stephanegigandet (approved):
Looks good to me

lib/ProductOpener/Test.pm
alexgarel: it will be used when I merge data quality :-D

lib/ProductOpener/Images.pm
alexgarel: Oh, I see, I though it was not intended (as code was duplicated).

tests/unit/expected_test_results/ingredients/fr-marmelade.json
```diff
@@ -326,71 +326,73 @@
...
-	  "en:soya-lecithin",
-	  "en:e322",
-	  "en:e322i",
-	  "en:e440a",
-	  "en:e330","en:e333",
-	  "en:sodium-citrate",
-	  "en:minerals",
-	  "en:sodium",
-	  "en:e415"
+      "en:chocolate",
```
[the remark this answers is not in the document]
alexgarel: No it's just a formating problem (two lines merged in previous version… no idea why !) `"en:e330","en:e333",`

lib/ProductOpener/Images.pm
stephanegigandet: We pay for each type of detection, so if we just need the text, then we should remove the others (which have been done on the full uncropped image already)

cpanfile
```diff
@@ -84,6 +84,10 @@ requires 'Log::Any::Adapter::Log4perl', '>= 0.09'; # liblog-any-adapter-log4perl
 requires 'Action::CircuitBreaker';
 requires 'Action::Retry'; # deps: libmath-fibonacci-perl
 
+# AnyEvent
```
stephanegigandet, line 87: what about the EV module?

lib/ProductOpener/APITest.pm
```diff
@@ -634,4 +640,74 @@ sub normalize_mail_for_comparison ($mail) {
 	return \@lines;
 }
 
+=head2 fake_http_server($port, $dump_path, $responses_ref) {
```
stephanegigandet, line 643: Awesome!

yuktea: I second Stèphane, Awesome!

cpanfile
alexgarel: AnyEvent brings the INotify module. There is no direct integration with EV.

AnyEvent is an abstraction, EV is the event implementation layer. So we need both.

alexgarel:
@stephanegigandet I merged this, but it means you should change the setup in production ! (remove incron task, and add a systemd entry for this new script)

PR #8236, opened 2023-03-20
fix: Do not show the lilo prompt in countries where it's unsupported

jnsereko:
### What
Do not show the lilo prompt in countries where it's unsupported
https://www.lilo.org/fr/open-food-facts/

### Screenshot

https://user-images.githubusercontent.com/58003327/226471635-4f615e76-1a19-40cf-88e8-1d5c64baf575.mov

cc @yuktea @stephanegigandet @alexgarel 

- Fixes #[7905]

### DIFF
--- a/templates/web/pages/session/signed_in.tt.html
+++ b/templates/web/pages/session/signed_in.tt.html
@@ -4,7 +4,9 @@
 
 [% IF !server_options_producers_platform %]
     <h3>[% lang('you_can_also_help_us') %]</h3>
-    <p>[% lang('bottom_content') %]</p>
+    [% IF lc == "en" or lc == "fr" %]
+        <p>[% lang('bottom_content') %]</p>
+    [% END %]
 [% ELSE %]
 
     <h3>[% lang("resources") %]</h3>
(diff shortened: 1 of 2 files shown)


templates/web/pages/session/signed_in.tt.html
```diff
@@ -4,7 +4,9 @@
 
 [% IF !server_options_producers_platform %]
     <h3>[% lang('you_can_also_help_us') %]</h3>
```
stephanegigandet, line 6: Thanks for the PR @jnsereko!

Could you include this <h3> title inside the IF below? Otherwise we show "Help us to fund the OFF project" but with nothing under it.

jnsereko: Hey @stephanegigandet 
i have added the `<h3>` into the if, but i didn't like below but i didn't understand pretty well what you meant with "Otherwise we show "Help us to fund the OFF project" but with nothing under it."
```
    [% IF lc == "en" or lc == "fr" %]
        <h3>[% lang('you_can_also_help_us') %]</h3>
        <p>[% lang('bottom_content') %]</p>
    [% END %]
```

stephanegigandet: I meant that the h3 tag goes with the p tag, if we don't display the p tag, then we don't want to display just a h3 title with nothing under it.

stephanegigandet (approved):
Thank you!

PR #8239, opened 2023-03-21
taxonomy: fix naming hr + taxonomy for mairoluin

benbenben2:
fix hrtavska instead of hrvatska 

taxonomy from mairoluin in Slack


taxonomies/ingredients.txt
```diff
@@ -24175,6 +24191,10 @@ sv:fullkornsspeltvetemjöl
 <en:spelt flour
 fr:farine de petit épeautre
 
+<fr:farine de petit épeautre
+fr:farine de petit épeautre de Haute Provence
```
stephanegigandet, line 24195: For this one, it's possible that adding "Haute Provence" to the origins taxonomy would be sufficient.

stephanegigandet: Right, I just checked, it works. Could you remove this entry, and add Haute Provence in the origins.txt taxonomy instead?

That way all ingredients of the form "quelque chose de Haute Provence" will be recognized.

stephanegigandet (approved):
Thank you!

PR #8246, opened 2023-03-24
fix: re-canonicalize packaging component properties

stephanegigandet:
This is needed because we have changed the canonical names for packaging entries like PET.


alexgarel (approved):
LGTM

PR #8255, opened 2023-03-25
fix: Invalid facets in the url return 404

### LINKED ISSUE #6802: Facets: non existing facets url should return a 404
### Describe the bug

If you mispell a facet : 
* https://world.openfoodfacts.org/data-quality/non-nutrition-data (*non* instead of *no*) show you a page (eventually with no products)
* https://world.openfoodfacts.org/category/breads/no-nutrition-data (forgot data-quality in uri) the application show you breads

this would be better to have a 404, "the facet does not exists".

It bites me once or twice, and bites again today ! (on the producer platform)

MonalikaPatnaik:
### What
Basically, if we forget any facet in the url for eg: [https://world.openfoodfacts.org/category/breads/no-nutrition-data](https://world.openfoodfacts.org/category/breads/no-nutrition-data) data-quality in this case, it is showing the breads page instead of an error as it is not validating whether no-nutrition-data is a valid facet or not thus generating a canonical URL and return a page. Thus we need to keep a check on this. 

### Screenshot
**for the link [https://world.openfoodfacts.org/category/breads/no-nutrition-data ](https://world.openfoodfacts.org/category/breads/no-nutrition-data )(forgot data-quality in uri) it is showing error instead of breads.**
Screenshot 2023-03-25 114157

**for the link [https://world.openfoodfacts.org/data-quality/non-nutrition-data](https://world.openfoodfacts.org/data-quality/non-nutrition-data ) (non instead of no)if we misspell any facet, then the behaviour is not changing as facets can contain non taxonomized values too, thus returning a page.**

Screenshot 2023-03-25 115555

- It fixes #6802


alexgarel (approved):
LGTM ! Thanks @MonalikaPatnaik :tada:

Issue #6816, opened 2022-05-23
Verify coverage of producer data protection


alexgarel:
Only moderators should be able to modify data protected by producers.

Verify it's the case via API and via webapp.

Verify it's advertised directly in the form so that user get right feedback for why he can't edit.

stephanegigandet:
Current issues with protection of data from producers:
- There's a bug in the web form processing that disables the protection.
- In the API, there is a very broad protection of most fields and nutrients (if some data was sent by the producer, then all data is protected)
- In the web, nutrition is not protected.

manon-corneille:
I tested this, and found that I could edit: 
* Quantity 
* Packaging data
* Ingredient and Nutrition photos

I couldn't edit: 
* Nutrition table
* Front photos

CharlesNepote:
From a Slack user today ([2026-04-10](https://openfoodfacts.slack.com/archives/C03H290LF/p1775678964200969)): "_Today I tried once again [to fix something] but also fail. After that I found the little "factory" icon on top showing that this is manufacturer data.
Why do we have such badly formatted manufacturer data?
And second thing going to UI designer: this little "factory" icon needs to be more visible. I just never saw it in nearly two years of using the app. And it would be nice to see a message denying the edit instead of "edit is processed" and silently dismissed!_

Issue #6721, opened 2022-04-29
We have no warning mechanism in Smoothie in case a product is protected (and the edit possibly refused)


teolemon:
### What
- We have no warning mechanism in Smoothie in case a product is protected (and the edit possibly refused)
### Part of
-

teolemon:
This was fixed in Smoothie where we now show producer contributed fields

Issue #2012, opened 2019-06-26
Better API support for protected products


aleene:
**Is your feature request related to a problem? Please describe.**
An API user can not determine and react on protected products.

**Describe the solution you'd like**
Add a protected state to the product json. And if needed specify which fields are protected. If the info is in the product json, we can prohibit editing those fields.

If a protected product is written return with an error json, so the api user sees what happened and can react on it.

VaiTon:
@aleene Could you please post an example URL?

teolemon:
- Related to #6721

PR #8274, opened 2023-03-30
fix: refactor protection of data imported from producers

stephanegigandet:
This PR solves some issues from https://github.com/openfoodfacts/openfoodfacts-server/issues/6816

- Protection on the web now works (bug fixed)
- Protection in the API is more precise (only fields sent by the producer are protected)
- Started to refactor so that the protection works the same way in the API and on the web
- Added tests to verify that the protection works as intended


lib/ProductOpener/APIProductWrite.pm
```diff
@@ -50,6 +51,38 @@ use ProductOpener::Lang qw/:all/;
 use ProductOpener::Products qw/:all/;
 use ProductOpener::API qw/:all/;
 use ProductOpener::Packaging qw/:all/;
+use ProductOpener::Text qw/:all/;
+
+use Encode;
+
+=head2 skip_protected_field($product_ref, $field, $moderator = 0)
+
+Return 1 if we should ignore a field value sent by an user because we already have a value sent by the producer.
```
alexgarel, line 60:
```suggestion
Return 1 if we should ignore a field value sent by a user because we already have a value sent by the producer.
```

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -50,6 +51,38 @@ use ProductOpener::Lang qw/:all/;
...
+
+=cut
+
+sub skip_protected_field ($product_ref, $field, $moderator = 0) {
```
alexgarel, line 64: It's really cool to have the logic isolated here.

Normally we give protection for only 1 year, couldn't we add the logic here ?

tests/integration/protected_product.t
```diff
@@ -0,0 +1,228 @@
...
+		(
+			code => '0200000000135',
+			product_name => "A protected product",
+		)
+	},
+);
+
+foreach my $product_form_override (@products) {
+	edit_product($ua, $product_form_override);
+}
```
alexgarel, line 80:
```suggestion
# create the products in the database
foreach my $product_form_override (@products) {
	edit_product($ua, $product_form_override);
}
```

tests/integration/protected_product.t
```diff
@@ -0,0 +1,228 @@
...
+		"energy-kj" => 1680183938,
+		"fat" => 1680183938,
+		"proteins" => 1680183938,
+		"quantity" => 1680183938,
+	};
+	store_product("organization-owner", $product_ref, "protecting the product");
+}
+
+# Note: expected results are stored in json files, see execute_api_tests
+my $tests_ref = [
```
alexgarel, line 101: If I understand it well, you "get-edited-protected-product-web-form" test_case depends upon "edit-protected-product-web-form" (it's a verification)

I think it would be better to pair them in the tests_ref structure and then have, but it would not enable using execute_api_tests.

So we shall at least comment on that !

```suggestion
# Note: expected results are stored in json files, see execute_api_tests
# Each test is composed of two test case: 
# one that edits a product, 
# and one that get the product to verify if it was edited or proctected
my $tests_ref = [
```

tests/integration/protected_product.t
```diff
@@ -0,0 +1,228 @@
...
+			nutriment_sugars => '7',
+		},
+		expected_type => 'html',
+	},
+	{
+		test_case => 'get-edited-protected-product-web-form',
+		method => 'GET',
+		path => '/api/v2/product/0200000000135',
+	},
+];
```
alexgarel, line 224: You did not test,if it works for moderators on protected products.

alexgarel (commented):
Really cool implementation.

I have some comments. The most important being about the 1 year rule.

Also it would be great if we could act on the web form (and the mobile form) to avoid users to try to modify those fields if they can't ! I see a lot of complaints coming.

lib/ProductOpener/APIProductWrite.pm
stephanegigandet: It's a bit tricky, if the product has not changed in a year, it does not mean the data is out of date. My preference is to say that we give protection for as long as the information is up to date. If we have evidence the data is not up to date, even after 6 months, then the protection could be removed.

tests/integration/protected_product.t
stephanegigandet: Sounds good.

stephanegigandet:
> Also it would be great if we could act on the web form (and the mobile form) to avoid users to try to modify those fields if they can't ! I see a lot of complaints coming.

Related:

- https://github.com/openfoodfacts/openfoodfacts-server/issues/6721
- https://github.com/openfoodfacts/openfoodfacts-server/issues/2012

- And I filed https://github.com/openfoodfacts/openfoodfacts-server/issues/8282

tests/integration/protected_product.t
stephanegigandet: @alexgarel I added some tests for moderators.

lib/ProductOpener/APIProductWrite.pm
alexgarel: Ok, but right now we have no way of removing that protecting isn't it ? (we should at least open an issue…)

alexgarel: ok I see you did open the issue.

tests/integration/expected_test_results/search_v1/search-no-filter.json
```diff
@@ -2145,24 +2149,24 @@
          "ingredients" : [
             {
                "id" : "es:apple",
-               "percent_estimate" : 66.6666666666667,
+               "percent_estimate" : "66.6666666666667",
```
alexgarel, line 2152: Strange that strings strikes back :frowning_face:

alexgarel (approved):
Cool.

PR #8284, opened 2023-04-04
taxonomy: Taxonomy hr 20230401

benbenben2:
used chatgpt to pluralize categories in HR
+ added few ingredients/labels


stephanegigandet (approved):
We can merge as-is, but it would be a good idea to also keep the singular as a synonym, so that we can match the category if someone uses the singular (or if we get it from a producer import)

benbenben2:
Yes, yes, yes. I will apply those changes

benbenben2:
@stephanegigandet, singular are kept in the last commit

stephanegigandet (approved):
Perfect, thank you!

PR #8290, opened 2023-04-05
taxonomy: ro translations

moon-rabbitOFF:
### Part of
-


taxonomies/ingredients.txt
```diff
@@ -14353,6 +14353,7 @@ nn:ghi, ghee
 pa:ਘਿਉ
 pl:ghi, ghee
 pt:ghi, ghee
+ro:ulei de unt
```
benbenben2, line 14356: Do you think that we could add "ghee" as a synonym as well?
```suggestion
ro:ulei de unt, ghee
```

aleene: I rather keep them separate, just to keep the indian variant as its own entity. You could add it to butter fat.

alexgarel:
@benbenben2 and @aleene would you approve ?

taxonomies/ingredients.txt
benbenben2: @alexgarel, 
I understood that @aleene suggests to keep only ghee in this block (indian variant) and more ulei de unt in other entry (butter fat)

benbenben2: suggestion from @aleene has been applied (you can ignore the first comment of this thread)

benbenben2:
@alexgarel , @moon-rabbitOFF , @aleene 

As suggested by @aleene, "ulei de unt" was move under "en:butterfat" (which already contains "butter oil" ("ulei de unt" in English) as synonym).

And the other block where "ulei de unt" was located, now contains only "ghee" for all languages.

moon-rabbitOFF:
Thank yu for your changes @benbenben2

benbenben2 (approved):

Issue #8295, opened 2023-04-07
Issue on mobile app due to missing panel


alexgarel:
### Describe the bug

In knowledge panels we mention the contribution panel in root panel but do not provide it. It makes a strange behaviour on mobile app.

### To Reproduce

Update a product on mobile app, and scroll down.
There is a long grey area at the end.

### Expected behavior

No panel at all. Server should avoid doing that also.

### Screenshots

see https://openfoodfacts.slack.com/archives/CSV2748EB/p1680877048529079

### Type of device

Browser

PR #8297, opened 2023-04-07
fix: avoid reference to missing panel in root card

alexgarel:
Do not reference contribution_card in root card if not present.

Fixes: #8295


stephanegigandet (approved):

lib/ProductOpener/KnowledgePanelsContribution.pm
```diff
@@ -91,7 +91,7 @@ sub create_contribution_card_panel ($product_ref, $target_lc, $target_cc, $optio
 			"api/knowledge-panels/contribution/contribution_card.tt.json",
 			$panel_data_ref, $product_ref, $target_lc, $target_cc, $options_ref);
 	}
-	return;
+	return !!@panels;
```
stephanegigandet, line 94: I'd suggest return scalar @panels

PR #8303, opened 2023-04-09
taxonomy: added translation and labels found in hr

benbenben2:
added translation and labels found in hr


stephanegigandet (approved):
Thank you!

PR #8307, opened 2023-04-10
fix: loosing the second facet at the end of the url on redirection

### LINKED ISSUE #7979: Redirect for synonym on facets loose the end of the url
### Describe the bug

If you try to reach https://world.openfoodfacts.org/category/en:lemonades/data-quality
you are redirected to https://world.openfoodfacts.org/category/en:lemonade

Redirection is ok because it's a synonym, so we redirect to id. What is wrong is that we loose the second facet in the redirect.

### To Reproduce

1. go to https://world.openfoodfacts.org/category/en:lemonades/data-quality
2. The page you land on is not a data-quality page

### Expected behavior

https://world.openfoodfacts.org/category/en:lemonades/data-quality should redirect to https://world.openfoodfacts.org/category/en:lemonade/data-quality

### Additional context

Causes:
- https://github.com/openfoodfacts/facets-knowledge-panels/issues/85

jnsereko:
### What
I have appended `$request_ref->{groupby_tagtype}` to  the `$request_ref->{current_link}` so that there is no truncation in redirection.
I have also added tests to verify that, 
1. request is processed 
2. the second facet exists in the redirection link

cc @yuktea @alexgarel @stephanegigandet 

### Screenshot

https://user-images.githubusercontent.com/58003327/230877303-f4caf75a-4257-4580-9fdb-93911b339b79.mov

- Fixes #7979

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -3060,6 +3060,7 @@ sub display_tag ($request_ref) {
 	}
 
 	if (defined $request_ref->{groupby_tagtype}) {
+		$request_ref->{current_link} .= "/" . $tag_type_plural{$request_ref->{groupby_tagtype}}{$lc};
 		$request_ref->{world_current_link} .= "/" . $tag_type_plural{$request_ref->{groupby_tagtype}}{$lc};
 	}
 
--- a/tests/unit/display.t
+++ b/tests/unit/display.t
@@ -153,9 +153,9 @@ like(display_field($product_ref, 'states'), qr/$expected/);
 
 # should not loose the second facet at the end of the url on redirection
 my $facets_ref = {
-	'current_link' => 'category/en:lemonades/data-quality',
+	'tagtype' => 'categories',
 	'groupby_tagtype' => 'data_quality',
-	'tagid' => 'en:lemonade'
+	'tagid' => 'en:bread'
 };
 
 my $apache_util_module = Test::MockModule->new('Apache2::RequestUtil');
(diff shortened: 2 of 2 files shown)


tests/unit/display.t
```diff
@@ -150,4 +151,59 @@ $product_ref = {
 $expected = lang('to_do_status') . separator_before_colon($lc) . q{:};
 like(display_field($product_ref, 'states'), qr/$expected/);
 
+# should not loose the second facet at the end of the url on redirection
+my $facets_ref = {
+	'current_link' => 'category/en:lemonades/data-quality',
```
stephanegigandet, line 156: current_link will be set by display_tag(), so the value you set here will be overwritten, we can remove it.

we should add a tagtype => "categories" though, otherwise you get a //lemonade/data-quality link, instead of /category/lemonade/data-quality

Can you use "bread" instead of "lemonades"? There's an issue in the categories taxonomy, the canonical name should be in plural (lemonades), but currently it's in singular.

tests/unit/display.t
```diff
@@ -150,4 +151,59 @@ $product_ref = {
...
+$display_module->mock(
+	'redirect_to_url',
+	sub {
+		# Do nothing, am just mocking the method
+	}
+);
+
+display_tag($facets_ref);
+
+unlike($facets_ref->{'current_link'}, qr/en:lemonades\/data\-quality/);    # verify that link has been processed
```
stephanegigandet, line 206: instead of like() , unlike(), I think it's best to use is() in order to check the exact value and verify it's the one we expect

e.g. is($facets_ref->{'current_link'}, '/category/bread/data-quality')

lib/ProductOpener/Display.pm
```diff
@@ -3060,6 +3060,7 @@ sub display_tag ($request_ref) {
 	}
 
 	if (defined $request_ref->{groupby_tagtype}) {
+		$request_ref->{current_link} .= "/" . $tag_type_plural{$request_ref->{groupby_tagtype}}{$lc};
```
stephanegigandet, line 3063: Looks good

stephanegigandet:
Hi @jnsereko , thanks a lot for adding a test with the mock functions. Your fix looks good, and for the test I made some suggestions. Thanks!

stephanegigandet (approved):
Perfect, thank you very much @jnsereko !

PR #8320, opened 2023-04-14
fix: gzip all JSON OCRs when saving OCR file on disk

raphael0202:
To save space, all OCR files were gzipped, but new generated files are still saved as plain text JSON files.
Also add a `created_at` field in the OCR JSON file containing the timestamp of generation of the OCR file. This is useful to know if we should generate again old OCR files.


raphael0202:
@alexgarel I haven't set up Product Opener locally, will give it a try ;)

raphael0202:
@alexgarel it should be good now!

raphael0202:
I checked locally, the generated gzipped JSON file saved correctly.

alexgarel (approved):
Great @raphael0202 !

PR #8324, opened 2023-04-15
chore: cleaning up the potatoes category

aleene:
### What
The potatoes category will be cleanup and reorganised such that it contains only products with a single ingredient (potato) and reflects the NOVA 1 status of this category.


moon-rabbitOFF:
I think it is a good idea to remove the products with added oil from the category potatoes, but I am less sure about the canned potatoes which generally contain only added salt. If we remove them from potatoes then we should, to stay coherant, remove all the canned vegetables from the category vegetables.

stephanegigandet (approved):
Looks good to me

PR #8326, opened 2023-04-17
feat: /api/v3/product/test endpoint

### LINKED ISSUE #8321: Add way to get product analysis results through the API without saving an actual product
This is needed for the Rate my recipe project. We can use the existing API, with a product code like "none" to indicate that we just want to get the computed scores, knowledge panels etc. for the product we send, but that it is not an actual product that needs to be saved in the database.

stephanegigandet:
This PR adds a /api/v3/product/test endpoint that allows to get the analysis (e.g. ingredients analysis, and later Nutri-Score, Eco-Score etc.) of a "test" product that won't be saved in the database. That feature will be used in particular for the Rate my recipe project: https://github.com/openfoodfacts/rate-my-recipe

Fixes #8321


teolemon (commented):
- We should have abuse control mechanisms on this route (given the value provided, it might be used a lot)
- We should quickly leverage it for our own producer platform to allow producers to experiment, and potentially share data

stephanegigandet:
Done through this PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/8351

Issue #8335, opened 2023-04-19
Polyols are called "Sugar alcohol", at least in the USA


CharlesNepote:
### Describe the bug

I have searched for [10 random products sold in the US with polyols + a nutrition table image](https://mirabelle.openfoodfacts.org/products?sql=--+10+random+products+sold+in+the+US+with+polyols+%2B+a+nutrition+table+image%0D%0Aselect+code%2C+url%2C+creator%2C+created_datetime%2C+last_modified_datetime%2C+last_modified_by%2C+product_name%2C%0D%0A++countries_en%2C+image_nutrition_url%2C%0D%0A++polyols_100g%0D%0Afrom+%5Ball%5D%0D%0Awhere%0D%0A++%22polyols_100g%22+is+not+null%0D%0A++and+%22image_nutrition_url%22+%21%3D+%22%22%0D%0A++and+countries_en+like+%22%25states%25%22%0D%0Aorder+by+random%28%29%0D%0Alimit+10): for all these products, polyols were called "Sugar alcohol" on the nutrition table. So I guess "Sugar alcohol" is the main name used in the USA.

People entering data will try to type "sugar alcohol" in the "Add nutrient" field but won't find anything relevant.

In the UK, the word seems to be "polyols". Eg. https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=0634158794032#nutrition

### To Reproduce

Examples:
* https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=0034000141418#nutrition
* https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=0602652271380#nutrition
* https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=0842379149146#nutrition
* https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=0843076000051#nutrition

### Expected behavior

People entering "sugar alcohol" in the "Add a nutrient" field should see "Polyols (sugar alcohol)" or something like this.

In fact, all sugar alcohols are polyols: "Since they contain multiple –OH groups, they are classified as [polyols](https://en.wikipedia.org/wiki/Polyol)." (source: [Sugar alcohol](https://en.wikipedia.org/wiki/Sugar_alcohol
) wikipedia page).
Not all polyols seem to be sugar alcohols, but the ones used in the food industry probably are.

In Open Food Facts we might use either:
* "Sugar alcohols (polyols)"
* or "Polyols (Sugar alcohols"

[...]

PR #8336, opened 2023-04-19
chore: polyols

aleene:
### What
added (sugar alcohols) for easier findability


aleene:
#8335 solved

stephanegigandet (changes requested):
We need to keep "polyols" as a synonym too, otherwise a lot of parsing things / recognizing column names in files from manufacturers etc. won't work

CharlesNepote:
@stephanegigandet @aleene thanks a lot!
1. Will the name change in the CSV export???
2. Is it possible to put "Sugar alcohol" first for us.openfoodfacts.org (and only for this one)?
3. At least Germany does not seem to use "polyols" either (can't remember what).

stephanegigandet:
@CharlesNepote :
> 
> 1. Will the name change in the CSV export???

No, the column name is taken from the canonical entry: zz:polyols

> 2. Is it possible to put "Sugar alcohol" first for us.openfoodfacts.org (and only for this one)?

Not currently, we don't support language variants.

Issue #6374, opened 2022-02-01
 📦 Packaging (tracker)


teolemon:
### Packaging data completion and quality
- [x] #6373 

### Displaying packaging information visually + Knowledge panels
- [ ] #5648 
- [ ] Change the wikidata entries from PVC to PVC recycling to make it more useful

### Leveraging the community
- [ ] #5023

### Automating
- [x] #4698 

### Packaging in Eco-Score computations
- [ ] #4691
- [ ] #1655
- [x] #219

### Facilitating packaging input
- [ ] #394
- [ ] Add autosuggest for packaging shapes and materials, potentially biased on the category
- [ ] Add inline extraction of packaging, duplicating what's available for ingredients

### Getting food producers on board 
- [ ] #7021 

### Packaging elsewhere
- [ ] Robotoff: https://github.com/openfoodfacts/robotoff/issues/121
- [ ] Smoothie: https://github.com/openfoodfacts/smooth-app/issues/515
- [x] Dart package: https://github.com/openfoodfacts/openfoodfacts-dart/issues/379
- [ ] Hunger Games: https://github.com/openfoodfacts/hunger-games/issues/249
- [x] API doc: https://github.com/openfoodfacts/api-documentation/issues/60

### Deprecated
- [ ] Classic apps: https://github.com/openfoodfacts/openfoodfacts-androidapp/issues/4193 - https://github.com/openfoodfacts/openfoodfacts-ios/issues/959

### Figma files

- https://www.figma.com/file/2SUpDoZi6nRyBF0SmcIQbd/Untitled
- https://www.figma.com/file/iVoEL5LaH9tIh2efsBqktF/Packaging-self-tracking
- https://www.figma.com/file/uEtDC3hqIZZeOs6AQAed3D/OpenLists
- https://www.figma.com/file/ne2vKoKODeq01geg2Cj3PT/Untitled
- https://www.figma.com/file/SOsp5de78KfJWsKkGsAPqx/Untitled
- https://www.figma.com/file/Gme79AHqYkupXL5UZuvF0D/Packaging-shapes-from-the-French-CITEO-system

teolemon:
Potentially, we could have a check on that kind of weights
<img width="418" alt="image" src="https://user-images.githubusercontent.com/1689815/204132165-d5583fd4-78bd-4aed-9f4f-f9a5c7ebdf4f.png">

PR #8337, opened 2023-04-20
chore: Added the statiegeldblik entry in recyling and logo

aleene:
### What
Added the statiegeldblik entry in recyling and logo

### Part of
- #6374


aleene:
It fails, but I did not see the cause

alexgarel (approved):
Thanks @aleene
LGTM, but you have commited a .DS_Store by error I think

stephanegigandet:
tried to fix conflicts

stephanegigandet (approved):
I'm not sure if this is correct, merging in order to get the logos and we can fix later if needed

PR #8329, opened 2023-04-17
feat: load CIQUAL database 

stephanegigandet:
part of #2861

will be needed for the rate my recipe project

Issue #2861, opened 2020-01-17
Try to recompute nutrition facts from the ingredient list


teolemon:
Now that we're able to parse % of ingredients, and to bound max and min values, it would be amusing to try and compute min and max values for nutrition facts, based on average values for each ingredients in databases such as CIQUAL

PR #8349, opened 2023-04-24
feat: use ciqual to estimate nutrients and nutriscore

stephanegigandet:
GitHub does not show changes for the PR https://github.com/openfoodfacts/openfoodfacts-server/pull/8329 , this is a new PR with a copy of all the changes, in only one commit. I hope it will work better.

part of #2861

will be needed for the 'rate my recipe project'

This PR includes:
- Loading CIQUAL data
- Estimating nutrients from ingredients (currently very basic, and only looking at top level ingredients)
- Storing estimated nutrients if we have a high proportion of ingredients for which we have nutrient data (95% or more of the ingredients)
- Computing an estimated Nutri-Score
- Displaying the estimated Nutri-Score, with some warnings.

e.g. for apples for which we don't have nutrition facts listed:

image


stephanegigandet:
github fails to show diffs, closing

image

PR #8350, opened 2023-04-24
refactor: move ecoscore and forest-footprint to external-data

stephanegigandet:
This is to avoid having lots of paths at the root level (makes GitHub repo difficult to read, and we need changes in docker configuration files whenever we want to add new external data, such as the ciqual database (coming next)).


stephanegigandet:
> Didn't we have emb codes to move there also ?

Yes, we can do that gradually.

stephanegigandet:
Done through this PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/8351

PR #8351, opened 2023-04-24
feat: use ciqual to estimate nutrients and nutriscore

stephanegigandet:
part of #2861

will be needed for the [Rate my recipe project](https://docs.google.com/document/d/1FZe_d8FRIEMt6afHvx80HMg5zdWURh-hFXCbd3Ohuu0/edit#)

This PR includes:
- Loading CIQUAL data
- Estimating nutrients from ingredients (currently very basic, and only looking at top level ingredients)
- Storing estimated nutrients if we have a high proportion of ingredients for which we have nutrient data (95% or more of the ingredients)
- Computing an estimated Nutri-Score
- Displaying the estimated Nutri-Score, with some warnings.
- Using API v3 to upload data (including ingredients_text and category) for a test product, and get back estimated nutrients and Nutri-Score
- 
e.g. for apples for which we don't have nutrition facts listed:

image


docs/api/ref/schemas/product_update_api_v3.yaml
```diff
@@ -11,3 +11,18 @@ properties:
...
+    description: 2 letter language code of the main language of the product (the most prominent on the packaging)
+  quantity:
+    type: string
+  serving_size:
+    type: string
+  categories_tags:
+    type: array
+    description: An array of categories
+    items:
+      $ref: ./packagings/input_taxonomy_tag.yaml
```
alexgarel, line 28: So we are making those new docs in parallel to the v2 fields ?

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -217,6 +281,66 @@ sub update_product_fields ($request_ref, $product_ref) {
 
 			update_field_with_0_or_1_value($request_ref, $product_ref, $field, $value);
 		}
+		# language fields
+		elsif ( ($field =~ /^(.*)_(\w\w)$/)
```
alexgarel, line 285: So we stick to using _xx in API v3 ?

This kind of pattern is not really OpenAPI friendly as it force to have a lot of patternProperties

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -217,6 +281,66 @@ sub update_product_fields ($request_ref, $product_ref) {
...
+
+			update_tags_fields($request_ref, $product_ref, $tagtype, $is_addition, $tags_lc, $value);
+		}
+		# Simple product fields
+		elsif (defined $product_simple_fields{$field}) {
+			$product_ref->{$field} = remove_tags_and_quote($value);
+			$request_ref->{updated_product_fields}{$field} = 1;
+		}
+		# Main language
+		elsif ($field eq "lang") {
```
alexgarel, line 315: Is it a change in strategy from previous API ? I mean, in previous API I could give a lc and it would interpret fields without language using this lc.

What about here ? And what if I change the "lang" of an existing product ? (or is language value aware only on tags fields ?)

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -186,13 +187,76 @@ sub update_packagings ($request_ref, $product_ref, $field, $is_addition, $value)
...
 }
 
-=head2 update_product_fields ($request_ref, $product_ref)
+=head2 update_tags_fields ($request_ref, $product_ref, $field, $is_addition, $value)
+
+Update packagings.
+
+=cut
+
+sub update_tags_fields ($request_ref, $product_ref, $field, $is_addition, $tags_lc, $value) {
```
alexgarel, line 196: I'm not sure $is_addition is a good name. It can mean, that the tags are new or that it's adding to existing tags…
I would better have a $do_replace (which would be the contrary of $is_addition ?).

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -186,13 +187,76 @@ sub update_packagings ($request_ref, $product_ref, $field, $is_addition, $value)
...
+			{
+				message => {id => "invalid_type_must_be_array"},
+				field => {id => $field},
+				impact => {id => "field_ignored"},
+			}
+		);
+	}
+	else {
+		# Generate a comma separated list of tags, so that we can use existing functions to add tags
+		my $tags_list = join(',', @$value);
```
alexgarel, line 213: Today I saw a products with many times the same brand, it disapeared once I edited. But I think a fool-proof mechanism (and a warning) to remove duplicates would be good.

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -186,13 +187,76 @@ sub update_packagings ($request_ref, $product_ref, $field, $is_addition, $value)
...
+
+		if ($is_addition) {
+			add_tags_to_field($product_ref, $tags_lc, $field, $tags_list);
+		}
+		else {
+			$product_ref->{$field} = $tags_list;
+		}
+
+		compute_field_tags($product_ref, $tags_lc, $field);
```
alexgarel, line 222: I like the fact that we have add_tags_to_field and compute_field_tags methods, so the business logic is better separated !

alexgarel (commented):
First part of my review.

lib/ProductOpener/APIProductWrite.pm
alexgarel: see https://github.com/openfoodfacts/openfoodfacts-server/issues/8355

docs/api/ref/schemas/product_update_api_v3.yaml
stephanegigandet: I'm not too sure about it. It's likely that v3 fields will evolve and be a bit different, and we don't support the full product fields from the v2 READ API, so that's why I added them here.

lib/ProductOpener/APIProductWrite.pm
stephanegigandet: $add_to_existing_tags would be better I think, it's more in line with the API logic: we replace by default, or we add if the field ends with _add

stephanegigandet: It mimicked what we had for packagings (adding a new packaging component):

update_packagings($request_ref, $product_ref, $field, $is_addition, $value)

alexgarel: :+1: for `$add_to_existing_tags`

lib/ProductOpener/Food.pm
```diff
@@ -1011,16 +1013,24 @@ sub compute_fruit_ratio ($product_ref, $prepared) {
...
+=head4 $nutriments_ref - ref to the nutriments of a product
+
+Reference to either the "nutriments" or "nutriments_estimated" structure.
+
+=head4 $prepared - string contains either "" or "prepared"
+
 =cut
 
-sub saturated_fat_ratio ($product_ref, $prepared) {
+sub saturated_fat_ratio ($nutriments_ref, $prepared) {
```
alexgarel, line 1030: cool to have changed this :-)

lib/ProductOpener/Tags.pm
```diff
@@ -234,6 +235,20 @@ To this initial list, taxonomized fields will be added by retrieve_tags_taxonomy
...
+	labels => 1,
+	origins => 1,
+	manufacturing_places => 1,
+	emb_codes => 1,
+	allergens => 1,
+	traces => 1,
+	purchase_places => 1,
+	stores => 1,
+	countries => 1,
+);
```
alexgarel, line 250: cool to have it documented this way.

tests/integration/api_v3_product_write.t
```diff
@@ -57,33 +57,26 @@ my $tests_ref = [
...
 		path => '/api/v3/product/1234567890007',
-		body => '{"product": {"fields": "updated", "packagings_add": [{"shape": {"lc_name": "bottle"}}]}}'
-	},
-	# Get updated fields + attributes and knowledge panels
-	{
-		test_case => 'patch-packagings-add-one-component',
-		method => 'PATCH',
-		path => '/api/v3/product/1234567890007',
-		body => '{"product": {"fields": "updated", "packagings_add": [{"shape": {"lc_name": "bottle"}}]}}'
+		body => '{"product": { "packagings_add": [{"shape": {"lc_name": "bottle"}}]}}'
```
alexgarel, line 60: OK, you removed a duplicate test :+1:

tests/unit/ingredients_percent.t
```diff
@@ -941,6 +962,108 @@ my @tests = (
...
+				'id' => 'en:lemon-flavouring',
+				'percent_estimate' => 0,
+				'percent_max' => 5,
+				'percent_min' => 0,
+				'text' => 'lemon flavouring'
+			}
+		]
+
+	],
+
```
alexgarel, line 1066: Great testing :+1:

tests/unit/nutrition_ciqual.t
```diff
@@ -0,0 +1,18 @@
...
+
+use Test::More;
+use Log::Any::Adapter 'TAP';
+
+use ProductOpener::NutritionCiqual qw/:all/;
+use ProductOpener::Test qw/:all/;
+
+my ($test_id, $test_dir, $expected_result_dir, $update_expected_results) = (init_expected_results(__FILE__));
+
+load_ciqual_data();
```
alexgarel, line 11:
```suggestion
# this test will just load ciqual data and verify we get expected structure for some values
load_ciqual_data();
```

tests/unit/producers.t
```diff
@@ -15,13 +15,16 @@ use ProductOpener::Test qw/:all/;
...
-		$update_expected_results
-	);
+	init_fields_columns_names_for_lang($l)
+		# 2023/04/24: the files are growing too much (currently 100Mb), which is too much for GitHub
+		# commenting out this test
+		#compare_to_expected_results(
+		#	init_fields_columns_names_for_lang($l),
+		#	$expected_result_dir . "/column_names_$l.json",
+		#	$update_expected_results
+		#);
```
alexgarel, line 27: maybe we can just check the list length some values are there…

alexgarel (approved):
Cool. 

Thanks for adding tests.

I made some suggestions.

lib/ProductOpener/APIProductWrite.pm
alexgarel: Discussed on weekly meeting. API v3 is experimental, in v3.1 we will break this to make it better.

lib/ProductOpener/Food.pm
```diff
@@ -1286,6 +1327,27 @@ sub compute_nutrition_score ($product_ref) {
...
+	# we can use use the nutriments_estimated hash if it exists.
+	# We only do so if there are no key nutrients specified, so that we don't compute an estimated Nutri-Score
+	# if we are only missing a few nutrients.
+	my $nutriments_field = "nutriments";
+
+	if (    (defined $product_ref->{"nutrition_grades_tags"})
+		and (($product_ref->{"nutrition_grades_tags"}[0] eq "unknown"))
+		and ($key_nutrients == 0)
+		and ($prepared eq '')
+		and (defined $product_ref->{nutriments_estimated}))
```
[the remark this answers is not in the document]
stephanegigandet: I added a check for it.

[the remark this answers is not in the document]
stephanegigandet: I updated the comment.

lib/ProductOpener/NutritionEstimation.pm
```diff
@@ -0,0 +1,114 @@
...
+				$total_with_nutrients += $ingredient_ref->{percent_estimate};
+
+				# Add nutrient value of the ingredient for all nutrients
+				while (my ($nid, $value) = each(%{$ciqual_data{$ciqual_id}{nutrients}})) {
+					deep_val($nutrients{$nid}) += $value * $ingredient_ref->{percent_estimate} / 100;
+				}
+			}
+			else {
+				# Keep track of ingredients for which we don't have nutrient values
+				deep_val($unknown_ingredients{$ingredient_ref->{id}}) += $ingredient_ref->{percent_estimate} / 100;
```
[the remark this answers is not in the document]
stephanegigandet: It's not needed indeed, $some_hash{"some_inexisting_key"}+=  does set the value without a warning.

lib/ProductOpener/NutritionCiqual.pm
```diff
@@ -0,0 +1,182 @@
...
+
+=head2 load_ciqual_data()
+
+Loads the Ciqual CALNUT database.
+
+Documentation of Ciqual CALNUT: https://ciqual.anses.fr/cms/sites/default/files/inline-files/Table%20CALNUT%202020_doc_FR_2020%2007%2007.pdf
+
+=cut
+
+sub load_ciqual_data() {
```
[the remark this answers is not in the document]
stephanegigandet: We could, I'm filing a bug for it: https://github.com/openfoodfacts/openfoodfacts-server/issues/8390

lib/ProductOpener/NutritionCiqual.pm
```diff
@@ -0,0 +1,182 @@
...
+use ProductOpener::Tags qw/:all/;
+use ProductOpener::Numbers qw/:all/;
+
+use Storable qw(dclone freeze);
+use Text::CSV();
+use Data::DeepAccess qw(deep_get deep_exists);
+
+=head1 VARIABLES
+
+=head2 %ciqual_data
```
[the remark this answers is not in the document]
stephanegigandet: If we don't plan to expose it externally, that's a bit over the top I think.

PR #8359, opened 2023-04-25
feat: export packaging data

stephanegigandet:
This PR is needed to export packagings data from the producers platform to the public platform.

### DIFF
--- a/lib/ProductOpener/ImportConvert.pm
+++ b/lib/ProductOpener/ImportConvert.pm
@@ -975,8 +975,7 @@ sub clean_fields ($product_ref) {
 			my $canon_tagid = canonicalize_taxonomy_tag($product_ref->{lc}, "countries", $product_ref->{$field});
 			if (not exists_taxonomy_tag("countries", $canon_tagid)) {
 				assign_value($product_ref, "origin_" . $product_ref->{lc}, $product_ref->{$field});
-				delete $product_ref->{$field};
-				next;
+				$product_ref->{$field} = "";
 			}
 		}
 


lib/ProductOpener/Export.pm
```diff
@@ -280,6 +280,26 @@ sub export_csv ($args_ref) {
...
+						foreach my $packaging_ref (@{$product_ref->{packagings}}) {
+							$i++;
+							my $j = 0;    # number of the field
+							foreach my $field (
+								qw(number_of_units shape material recycling quantity_per_unit weight_specified weight_measured)
+								)
+							{
+								$j++;
+								if (defined $packaging_ref->{$field}) {
+									my $field_sort_key = sprintf("%08d", $group_number * 1000 + $i * 10 + $j);
```
alexgarel, line 296: can you add a comment to explain this $field_sort_key ?
(also above).

also why don't you use $field_sort_key . "_$j" like on line 272 ?

alexgarel (approved):
LGTM, but a comment to add, if possible.

lib/ProductOpener/Export.pm
stephanegigandet: There's a comment a bit above:

```
		# %populated_fields will contain the field name as the key,
		# and a sort key as the value so that the CSV columns are in the order of $options{import_export_fields_groups}
		my %populated_fields = ();
```

stephanegigandet: I'll add another comment for packagings:
```

									# Generate a sort key so that the packaging fields in the CSV file are in this order:
									# - all fields for packaging component 1, then all fields for packaging component 2
									# - number_of_units, shape, material etc. (same a for loop above)
```

PR #8362, opened 2023-04-25
feat: export packaging components data 

### LINKED ISSUE #8361: Export packaging components data in dedicated CSV export
Packagings data is now structured: different fields like shape and material for a varying number of components for products.
To ease the analysis of packaging data, we can generate a new CSV for packaging data, with 1 line per packaging component.

Note that only the child shape / material / recycling will be exported, so a PET bottle will have 1 line for a component with a PET material, and no line with the parent "plastics".

An alternative way to explore the data is to use the aggregated packaging stats per country and categories, which includes data on parents: https://world.openfoodfacts.org/data/categories_stats/categories_packagings_stats.packagings-with-weights.json

stephanegigandet:
Fixes #8361

Generate a new CSV file for packaging data, 1 line per packaging component

### DIFF
--- a/scripts/gen_packaging_stats.pl
+++ b/scripts/gen_packaging_stats.pl
@@ -70,6 +70,7 @@ use File::Path qw(mkpath);
 use JSON::PP;
 use Data::DeepAccess qw(deep_exists deep_get deep_set deep_val);
 use Getopt::Long;
+use Text::CSV;
 
 my $quiet;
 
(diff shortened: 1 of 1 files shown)


scripts/gen_packaging_stats.pl
```diff
@@ -365,13 +406,38 @@ ($name, $query_ref)
 
 	my $packagings_stats_ref = {};
 
+	# Export packaging components of all products to a CSV file
+	my $filehandle;
+	my $filename = "$www_root/data/$name.csv";
```
alexgarel, line 411: Couldn't we export to a csv.gz ?

alexgarel (approved):
LGTM

scripts/gen_packaging_stats.pl
stephanegigandet: We could.. but it's just a one off at this point, so that we have data for the report of the packaging operation.

PR #8378, opened 2023-04-28
taxonomy: Danish ingredients

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]


taxonomies/ingredients.txt
```diff
@@ -29135,7 +29135,7 @@ fr:sirop d'agave concentré biologique
 es:jarabe de ágave concentrado ecológico
 
 <en:vegetable fiber
-en:agave fibre
+en:agave fiber
```
benbenben2, line 29138: american vs british english. You can keep fibre as synonym maybe as it may occurs. 
Or even better, add it as synonym on top of the file (that will handle all entries containing fibre/fiber):
`synonyms:en:fiber, fibre`

moon-rabbitOFF: Didn't know that, thx!

PR #8384, opened 2023-05-02
fix: explore files in sorted order to make tests non-random

stephanegigandet:
This fixes a failing test in recent PRs.


teolemon (approved):
Interesting. There could also be alleys for optimizing the way we run tests (eg time consuming / boring / unlikely to fail ones at the end and/or beginning)

PR #8397, opened 2023-05-04
fix: export obsolete products - draft

stephanegigandet:
Show the number of products that are not exported in both products and products_obsolete collections, and export both.

### DIFF
--- a/lib/ProductOpener/Export.pm
+++ b/lib/ProductOpener/Export.pm
@@ -186,196 +186,221 @@ sub export_csv ($args_ref) {
 
 	$log->debug("export_csv - start", {args_ref => $args_ref}) if $log->is_debug();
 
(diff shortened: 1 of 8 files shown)


lib/ProductOpener/Export.pm
```diff
@@ -186,196 +190,221 @@ sub export_csv ($args_ref) {
...
+		push @collections, {"products_obsolete"};
+	}
 
+	# Create a list of the fields that we will export
 	my @sorted_populated_fields;
 
 	my %other_images = ();
 
-	if (not defined $fields_ref) {
+	# We will have one cursor for each collection
```
alexgarel, line 207:
```suggestion
	# We will have one cursor for each collection
	# We store cursors because we will iterate them twice
```

alexgarel (approved):
Good work !

PR #8398, opened 2023-05-04
feat: add config files

alexgarel:
after off2 reinstall


stephanegigandet (commented):
@alexgarel It looks like there are tons of files in /lang that should not be there?
the /po and /html/images/misc changes are weird too.

alexgarel:
> @alexgarel It looks like there are tons of files in /lang that should not be there? the /po and /html/images/misc changes are weird too.

Can we look at it together ?

stephanegigandet (approved):

PR #8409, opened 2023-05-11
fix: new misc tags fort packaging imports, + set packagings_complete

stephanegigandet:
This PR creates new fields for the number of packaging components (in particular to ease debugging of packagings imports from producers).
When we get packagings imported from producers, it also sets the packagings_complete field to 1.


alexgarel (approved):
LGTM

PR #8258, opened 2023-03-27
feat: Prevention - On the fly quality checks in the product edit form

jnsereko:
### What
Created an invisible span having a warning on saturated fats and sugars values.  inputing an element above the range displays the span.

### Screenshot

https://user-images.githubusercontent.com/58003327/234127304-bb7f3a63-b37d-43d3-a8ca-8a758112b6ac.mov

### Related issue(s) and discussion

- Fixes #1905

PR #8411, opened 2023-05-12
fix: nutrition edit warnings

stephanegigandet:
This PR builds upon the work of @jnsereko : https://github.com/openfoodfacts/openfoodfacts-server/pull/8258
It fixes some issues with the original PR, that were found when doing more testing before pushing to production.

- We need to use parseFloat() on input values, otherwise "47" < "8"

image

- Commas are replaced by dots, so that French numbers like 4,7 are converted to 4.7
- Re-ordered the logic so that some warnings like the sugars > 100g warning is not cleared if carbohydrates are then modified (and are above sugars). In practice the "Please enter a value between 0 and 100" now takes priority over the other.
- Added translations in .po files.
- Allow values > 100g if nutrients are per serving and not per 100g

### DIFF
--- a/templates/web/pages/product_edit/product_edit_form_display.tt.js
+++ b/templates/web/pages/product_edit/product_edit_form_display.tt.js
@@ -40,61 +40,44 @@ function show_warning(should_show, nutrient_id, warning_message){
 		\$('#nutriment_'+nutrient_id).css("background-color", "rgb(255 237 235)");
 		\$('#nutriment_question_mark_'+nutrient_id).css("display", "inline-table");
 		\$('#nutriment_sugars_warning_'+nutrient_id).text(warning_message);
-	}else {
-		\$('#nutriment_'+nutrient_id).css("background-color", "white");
-		\$('#nutriment_question_mark_'+nutrient_id).css("display", "none");
+	} else {
+		// clear the warning only if the warning message we don't show is the same as the existing warning
+		// so that we don't remove a warning on sugars > 100g if we change carbohydrates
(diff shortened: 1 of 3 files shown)


templates/web/pages/product_edit/product_edit_form_display.tt.js
```diff
@@ -40,61 +40,44 @@ function show_warning(should_show, nutrient_id, warning_message){
...
-		var nutrient_value = \$(this).val();
-		var is_above_or_below_100 = isNaN(nutrient_value) || nutrient_value < 0 || nutrient_value > 100;
-		show_warning(is_above_or_below_100, nutrient_id, "Please enter a value between 0 and 100");
 
-		var crutial_nutrients = ['fat', 'saturated-fat', 'sugars', 'carbohydrates'];
+		// check the changed nutrient value
+		var nutrient_value = \$(this).val().replace(',','.');
+		var is_above_or_below_100 = (isNaN(nutrient_value) && nutrient_value != '-') || nutrient_value < 0 || nutrient_value > 100;
+		// if the nutrition facts are indicated per serving, the value can be above 100
+		if ((nutrient_value > 100) && (\$('#nutrition_data_per_serving').is(':checked'))) {
```
alexgarel, line 62: can't we check that it's below serving size ?

alexgarel (approved):
LGTM ! Thanks.

templates/web/pages/product_edit/product_edit_form_display.tt.js
stephanegigandet: It's tricky, because serving size is parsed on the server (it can be specified with different units etc.), but the user can change it in the form while editing nutrients.

PR #8422, opened 2023-05-16
build: Deploy stagging nfs mount

alexgarel:
Use NFS mounted volumes directly instead of bind mounting system mounted NFS shares.

This will ease updating clones.


stephanegigandet (approved):

Makefile
```diff
@@ -14,6 +14,10 @@ MOUNT_POINT ?= /mnt
 DOCKER_LOCAL_DATA ?= /srv/off/docker_data
 OS := $(shell uname)
 
+# monut point for shared data (default to the one on staging)
```
stephanegigandet, line 17:
```suggestion
# mount point for shared data (default to the one on staging)
```

Issue #7309, opened 2022-09-06
NEW DESIGN: products not aligned on first row and last row of lists of products


stephanegigandet:

eeshaanSA:
Would like to take this up. @stephanegigandet

stephanegigandet:
@eeshaanSA : thank you!

Corresponding code: https://github.com/openfoodfacts/openfoodfacts-server/blob/redesign2/templates/web/pages/product/product_page.tt.html

+ the SCSS code in /scss/ 

+ JS code in /html/js/product-search.js

jainshubham766:
I would like to resolve this issue, under hacktoberfeast.
Please assign

stephanegigandet:
> I would like to resolve this issue, under hacktoberfeast.
> Please assign

@jainshubham766 Thanks, I assigned the issue to you.

jainshubham766:
@stephanegigandet Can u please specify the file to target? I will help me to resolve fast.

stephanegigandet:
@jainshubham766 you can set up the project locally in docker: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/introduction/dev-environment-quick-start-guide.md

The relevant files:

/scss/*.scss
/html/js/product-search.js

PR #8423, opened 2023-05-16
fix: make products aligned again in list #7309

jusfla:
### What

Fix the alignments of products at the beginning and the end of the list of products so as to make it more aesthetically coherent with the whole list and new app design.

### Screenshot

E43CAE7F-480C-460C-9DF2-62EED5949E6E

Screenshot 2023-05-16 at 14 36 50

- Fixes #[7309]

### DIFF
--- a/scss/_product-list.scss
+++ b/scss/_product-list.scss
@@ -25,6 +25,30 @@ $productsGutter:12;
     }
 }
 
+.search_results::before {
+  content: none;
+}
+
+.search_results::after {
+  content: "";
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Thank you very much @jusfla! I tested it, it is working very well with all the different screen sizes.

There's one issue when the last line has more than 1 entry but is missing 2 or more entries to get a full line, but it's probably a different bug that we can fix in another PR.

PR #8436, opened 2023-05-20
fix: Update nova-group-unknown.svg

g123k:
Hi everyone!

On the mobile app, [we have some issues](https://github.com/openfoodfacts/smooth-app/issues/3990) with the current version of the Unknown group's SVG file.

After regenerating it with Illustrator, it seems to be OK on our side.

Could it be possible to use this one instead?
But please check, there is no visual regression (I don't think, but always better).


alexgarel (approved):
LGTM

g123k:
Thanks @alexgarel!

PR #8360, opened 2023-04-25
feat: quality test for nutriscore on olive oils

stephanegigandet:
Added a test to verify that we estimate the fruits/vegetables content of olive oils without ingredients to 100% for nutriscore computation.

PR #8444, opened 2023-05-22
feat: bypass data quality error for citrus

### LINKED ISSUE #8425: Integrate acides in energy computation or bypass data quality error for citrus
### Describe the bug

A producer sent us 8 lemon juices which were found to have an error: "Energy value in kJ does not correspond to the value calculated from the other nutrients".

The producer tells me "The reason is that our on-pack energy value includes energy from organic acids (mainly citric acid) at 3 kcal/g.
Our labels and Equadis data are therefore correct, knowing that we cannot display these organic acids on-pack due to the INCO regulation."
 
A solution could be to integrate acides in energy computation or bypass data quality error for citrus. 

List of products concerned: 
3478820088879
3456300014280
3456300013658
3456300009224
3456300002492
3456300002430
3456300002508

benbenben2:
- [x] PR title is prefixed by one of the following: feat, fix, docs, style, refactor, test, build, ci, chore, revert, l10n, taxonomy
- [x] Code is well documented
- [x] Include unit tests for new functionality
- [x] Code passes GitHub workflow checks in your branch
- [x] If you have multiple commits please combine them into one commit by squashing them.
- [x] Read and understood the [contribution guidelines](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/CONTRIBUTING.md)

### What
Added feature to bypass "Energy value in kJ does not correspond to the value calculated from the other nutrients"

### Screenshot
Not sure what to screenshot because there is no error anymore :-D

- Fixes #8425

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -594,39 +594,10 @@ sub check_nutrition_data_energy_computation ($product_ref) {
 			# following error/warning should be ignored for some categories
 			# for example, lemon juices containing organic acid, it is forbidden to display organic acid in nutrition tables but
 			# organic acid contributes to the total energy calculation
-			my $skip_energy_error = 0;
-			if (defined $product_ref->{categories_tags}) {
-				# start from end, categories_tags has the most specific categories at the end
-				my $i = @{$product_ref->{categories_tags}} - 1;
-				while (
-					($i >= 0)
-					and (
-						not(
-							# category - or a parent - with expected tag
-							defined get_inherited_property(
-								"categories", $product_ref->{categories_tags}[$i],
-								"ignore_energy_calculated_error:en"
-							)
-							and (
-								# we expect single letter a, b, c, d, e for nutriscore grade in the taxonomy. Case insensitive (/i).
-								get_inherited_property(
-									"categories", $product_ref->{categories_tags}[$i],
-									"ignore_energy_calculated_error:en"
-								) =~ /yes$/i
-							)
-						)
-					)
-					)
-				{
-					$i--;
-				}
+			my $ignore_energy_calculated_error
+				= get_inherited_property_from_categories_tags($product_ref, "ignore_energy_calculated_error:en");
 
-				if ($i >= 0) {
-					$skip_energy_error = 1;
-				}
-			}
-
-			if (not($skip_energy_error)) {
+			if (not($ignore_energy_calculated_error)) {
 				# Compare to specified energy value with a tolerance of 30% + an additiontal tolerance of 5
 				if (   ($computed_energy < ($specified_energy * 0.7 - 5))
 					or ($computed_energy > ($specified_energy * 1.3 + 5)))
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -591,22 +591,59 @@ sub check_nutrition_data_energy_computation ($product_ref) {
 				$computed_energy += $grams * $energy_per_gram;
 			}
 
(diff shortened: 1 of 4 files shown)


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -591,22 +591,59 @@ sub check_nutrition_data_energy_computation ($product_ref) {
...
+					)
+					)
+				{
+					$i--;
+				}
+
+				if ($i >= 0) {
+					$skip_energy_error = 1;
+				}
 			}
```
alexgarel, line 627: Can we make it a separate method ? (easier to read code)

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -591,22 +591,59 @@ sub check_nutrition_data_energy_computation ($product_ref) {
...
-					"en:energy-value-in-$unit-does-not-match-value-computed-from-other-nutrients";
+			# following error/warning should be ignored for some categories
+			# for example, lemon juices containing organic acid, it is forbidden to display organic acid in nutrition tables but
+			# organic acid contributes to the total energy calculation
+			my $skip_energy_error = 0;
+			if (defined $product_ref->{categories_tags}) {
+				# start from end, categories_tags has the most specific categories at the end
+				my $i = @{$product_ref->{categories_tags}} - 1;
+				while (
+					($i >= 0)
```
alexgarel, line 602: It seems to me you could rewrite this with a foreach and the reverse function + a break if you match condition to skip energy error. It would be more consistent with the rest of the codebase.

alexgarel: Also, you make two call to get_inherited_property while it's a costly method, you'd better store the result in a local variable.

lib/ProductOpener/DataQualityFood.pm
benbenben2: There is a new function/subroutine in **Tags.pm** (because it is based on **get_inherited_property** subroutine).
That new subroutine is called from **DataQualityFood.pm**

The idea would be to reuse this function in the other PR (https://github.com/openfoodfacts/openfoodfacts-server/pull/8360)
(eventually) It could be improved, because we expect a "yes" and we could maybe add what is expected as input parameter in the function. Here we expect a "yes", in the other PR we expect a nutriscore. What do you think?

lib/ProductOpener/DataQualityFood.pm
benbenben2: yes, thank you for the review and comments. Sorry, I am still learning. I will try to improve it. 

I was inspired by the following snippet in **ProducersFood.pm**
```

	my $i = @{$product_ref->{categories_tags}} - 1;

	while (
		($i >= 0)
		and not((defined $categories_nutriments_ref->{$product_ref->{categories_tags}[$i]})
			and (defined $categories_nutriments_ref->{$product_ref->{categories_tags}[$i]}{nutriments}))
		)
	{
		$i--;
	}
	# categories_tags has the most specific categories at the end

	if ($i >= 0) {
```

Now I reworked it based on **compute_ecoscore_agribalyse** found in **Ecoscore.pm**

Can you review this PR as well: https://github.com/openfoodfacts/openfoodfacts-server/pull/8360? 
I did code something similar there too.

lib/ProductOpener/Tags.pm
```diff
@@ -381,6 +382,44 @@ sub get_inherited_property ($tagtype, $canon_tagid, $property) {
...
+=head4 $product_ref - the product reference
+=head4 $property - the property - string
+
+=head3 Return
+
+The property if found.
+
+=cut
+
+sub get_inherited_property_from_categories_tags ($product_ref, $property) {
```
stephanegigandet, line 400: Good idea to create this function!

PR #8450, opened 2023-05-23
docs: fix OpenAPI `Product` schema file, extract agribalyze.yaml

### LINKED ISSUE #8449: The `Product` schema in OpenAPI does not work with code gen tools
I'm trying to update the nodejs SDK, and I'm trying to use code generation for the OpenAPI schema. The problem is that this file does not produce a valid schema:

https://github.com/openfoodfacts/openfoodfacts-server/blob/7953dbc8f668378239d1bac4bb495e27d7e2a369/docs/api/ref/schemas/product.yaml#L1

### Solution
Remove the initial `Product:` and de-indent everything.

It works with that edit.

VaiTon:
Closes #8449


alexgarel (approved):
Thank you so much @VaiTon ! It's really appreciated.

PR #8452, opened 2023-05-24
fix: Preventing Email Addresses as Entries in Fields

### LINKED ISSUE #6539: Prevent people from entering email addresses as values

MonalikaPatnaik:
### What
Written a separate subroutine preprocess_product_field that iterates over the values of tag fields and removes email entries in it by calling the remove_email () func unless the tag doesn't contains email addresses.

### Relevant Screenshot
image
### Expected Behaviour
Removes the mail entries and substitutes an empty string in place.

- Fixes #6539


lib/ProductOpener/Products.pm
```diff
@@ -2954,6 +2955,32 @@ or "slack_CHANNEL_NAME" (B<warning> currently channel name is ignored, we post t
 
 =cut
 
+sub removeEmailValues ($product_ref) {
+
+	# Iterate over the product fields
+	foreach my $field (keys %{$product_ref->{product}}) {
```
stephanegigandet, line 2961: There's no "product" field in the product object, the fields are at the root of the hash.

```suggestion
	foreach my $field (keys %$product_ref) {
```

lib/ProductOpener/Products.pm
```diff
@@ -2954,6 +2955,32 @@ or "slack_CHANNEL_NAME" (B<warning> currently channel name is ignored, we post t
 
 =cut
 
+sub removeEmailValues ($product_ref) {
```
stephanegigandet, line 2958: In Product Opener (and most Perl programs I think), we use snake_case instead of camelCase for variable and function names.

lib/ProductOpener/Products.pm
```diff
@@ -2954,6 +2955,32 @@ or "slack_CHANNEL_NAME" (B<warning> currently channel name is ignored, we post t
...
+
+		if (defined $product_ref->{product}->{$field}) {
+			my $value = $product_ref->{product}->{$field};
+			my $emailRegex = /^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$/;
+
+			# Check if the field is multi-valued
+			if (ref($value) eq 'ARRAY') {
+				my @new_values;
+				foreach my $entry (@$value) {
+					push @new_values, $entry unless $entry && $entry =~ $emailRegex;
```
stephanegigandet, line 2971: A product can have nested arrays / hashes. For instance the "packagings" field is an array of "packaging components" hashes. So this could apply the regexp match to a reference to an array or hash, which would not work.

lib/ProductOpener/Products.pm
```diff
@@ -2954,6 +2955,32 @@ or "slack_CHANNEL_NAME" (B<warning> currently channel name is ignored, we post t
...
+
+	# Iterate over the product fields
+	foreach my $field (keys %{$product_ref->{product}}) {
+
+		if (defined $product_ref->{product}->{$field}) {
+			my $value = $product_ref->{product}->{$field};
+			my $emailRegex = /^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$/;
+
+			# Check if the field is multi-valued
+			if (ref($value) eq 'ARRAY') {
```
stephanegigandet, line 2968: I think it would be best to check only scalar values, and not values that are inside arrays. The _tags fields (that are arrays) may contain emails, but those are derived from other fields (e.g. brands_tags is derived from brands).

stephanegigandet:
There are 2 different things:
1. preventing entry of emails
2. removing emails that may already be there

I'm thinking we could start with 1. 2 is more dangerous, especially if we run the remove email function on all fields of all products: if we make a mistake, we could remove valuable data. e.g. it's possible that some fields could contain valid e-mails that we want to keep (e.g. texts that we get from producers, like consumer information).

So in that case, the files to change would be /cgi/product_multilingual.pm and /cgi/product_jqm_multilingual.pl (and of course any shared function could be in Products.pm as you did)

lib/ProductOpener/Products.pm
MonalikaPatnaik: You are right! It needs to be modified to handle references to arrays or hashes correctly.
Thanks for pointing out!

lib/ProductOpener/Products.pm
MonalikaPatnaik: Okay..
Experimenting with it!

tests/unit/text.t
```diff
@@ -61,4 +62,25 @@ is(normalize_percentages('2500%', 'fr'), "2\N{U+00A0}500\N{U+00A0}%");
 is(normalize_percentages('2,50%', 'en'), "2.5%");
 is(normalize_percentages('2.50%', 'en'), "2.5%");
 
+# Test preprocess_product_field
```
stephanegigandet, line 65: in text.t, we should make a unit test only for the remove_email() function of Text.pm, you can test it with a few different values that have (or don't have) email addresses

MonalikaPatnaik: Okok...doing rn

lib/ProductOpener/Products.pm
```diff
@@ -2946,6 +2947,21 @@ or "slack_CHANNEL_NAME" (B<warning> currently channel name is ignored, we post t
 
 =cut
 
+sub preprocess_product_field ($product_ref) {
+
+	my @tag_fields = qw(brands categories origins labels packaging stores manufacturing_places);
+	foreach my $field (@tag_fields) {
+		if (defined $product_ref->{$field}) {
+			my @tags = split(/,/, $product_ref->{$field});
```
stephanegigandet, line 2955: This works, but you could call directly remove_email on $product_ref->{$field} and it should give exactly the same result, without having to do a split, a foreach and a join.

cgi/product_jqm_multilingual.pl
```diff
@@ -144,6 +144,9 @@ =head1 DESCRIPTION
 	print $out encode_json(Vars());
 	close $out;
 
+	# Processes fields to remove email values as entries
+	preprocess_product_field($product_ref);
```
stephanegigandet, line 148: So as we discussed yesterday, I think it would be better to process each field when we input it: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/cgi/product_multilingual.pl#L586

We have lines like this one:
$product_ref->{$field} = remove_tags_and_quote(decode utf8 => single_param($field));

This is the line that reads the value entered by the user, and removes HTML tags from it.

Instead we could do this:

$product_ref->{$field} = preprocess_product_field($field, decode utf8 => single_param($field));

The preprocess_product_field() would then call remove_tags_and_quotes() for all fields, and unless the field can contain emails (I think only "customer_service" and "other_information"), it also calls remove_email()

Note that the field name $field would be passed as the first argument: preprocess_product_field($field, $value)
and the function would return the preprocessed value.

lib/ProductOpener/Products.pm
MonalikaPatnaik: Fixing it!

cgi/product_jqm_multilingual.pl
MonalikaPatnaik: Done the required changes!

stephanegigandet (commented):
Perfect, thank you @MonalikaPatnaik !

To be complete, could you also add the call to preprocess_product_field() in cgi/product_jqm_multilingual.pl (that's V2 of the API) and in lib/ProductOpener/APIProductWrite.pm (that's V3 of the API) ?

MonalikaPatnaik:
> Perfect, thank you @MonalikaPatnaik !
> 
> To be complete, could you also add the call to preprocess_product_field() in cgi/product_jqm_multilingual.pl (that's V2 of the API) and in lib/ProductOpener/APIProductWrite.pm (that's V3 of the API) ?

Hii @stephanegigandet, let me know if I called the subroutine at incorrect location in these files!
Thankyouu<3

stephanegigandet (approved):
Looks good, thank you @MonalikaPatnaik !

PR #8453, opened 2023-05-24
taxonomy: e-243 additive exposure

stephanegigandet:
EFSA conclusion in https://efsa.onlinelibrary.wiley.com/doi/10.2903/j.efsa.2019.5621

"Based on the above, the Panel concluded that the current ADI of 0.5 mg/kg bw would be reached in toddlers and children at the 95th percentile already for exposure estimates calculated using the currently permitted ML for ethyl lauroyl arginate (E 243)."


benbenben2 (approved):
Looks good to me

PR #8455, opened 2023-05-24
chore: migrate gulpfile to TypeScript

VaiTon:
### What
- Use TypeScript for gulpfile.js (gulpfile.ts).
- Use typescript-eslint for gulpfile.ts linting


VaiTon:
> @VaiTon is this PR changing the generated files that we serve to users, or are the generated files exactly the same?

The generated files should be exactly the same. Gulp should do the exact same operations, but with type safety :)

VaiTon:
@stephanegigandet the only thing that I changed about file generation is replacing `gulp-terser-js` with `gulp-terser`, because the former has no types (d.ts files).

- https://www.npmjs.com/package/gulp-terser-js
- https://www.npmjs.com/package/gulp-terser

stephanegigandet (approved):
Thank you!

PR #8459, opened 2023-05-25
feat: add all logo variants for all projects in a normalized way

teolemon:
### What
- add all logo variants for all projects in a normalized way


stephanegigandet (approved):
Thank you!

Issue #8420, opened 2023-05-15
On the fly quality checks issues


CharlesNepote:
### What

- There are at issues with online quality checks.

1. The quality check is wrongly triggered in the Kcal and KJ fields

![image](https://github.com/openfoodfacts/openfoodfacts-server/assets/562352/3ffe3080-c755-4996-83a7-bb900709173a)

* KJ should always be < 3800 (**per 100 g**).
* Kcal should always be < 900 (**per 100 g**).

2. The quality check is triggered for value > 100 even if unit is mg

If the unit is mg, then it should be < 100,000. Same issue for µg, etc.

![image](https://github.com/openfoodfacts/openfoodfacts-server/assets/562352/50e6589b-2c93-4661-a7d9-5a71a3eafeee)

3. Helper in black is under the image (see previous screenshot).

4. The quality check should be triggered at page load, even if the user does not enter anything.

### To Reproduce

See any page in edition. Eg. https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=8410764010474

### Context
* https://github.com/openfoodfacts/openfoodfacts-server/issues/1905
* https://github.com/openfoodfacts/openfoodfacts-server/pull/8258
* https://github.com/openfoodfacts/openfoodfacts-server/pull/8411

### Part of
- #5538

benbenben2:
Related to this, if it could be possible to use the "<" symbol without having the field in red, that would be great as well:
![image](https://github.com/openfoodfacts/openfoodfacts-server/assets/110821832/2411fbae-4b94-4f6e-a027-c1a1fdc9e830)

benbenben2:
solved by Stephane

PR #8464, opened 2023-05-26
fix: on the fly nutrition edit checks #8420

stephanegigandet:
Solves most of the issues reported by @CharlesNepote and @benbenben2 in https://github.com/openfoodfacts/openfoodfacts-server/issues/8420

image

Does not address the tooltip below the image issue.


CharlesNepote (approved):
Thanks a lot! Looks good but didn't check on dev. Could you merge to let us verify on openfoodfacts.net?

PR #8471, opened 2023-05-29
Automatically add issues with the Nutri-Score label to the Nutri-Score project

teolemon:
### What
- Automatically add issues with the Nutri-Score label to the Nutri-Score project


alexgarel (approved):
LGTM

PR #8472, opened 2023-05-29
feat: Update delete_user to append unique hash

### LINKED ISSUE #8164: Allow users to be removed while keeping moderation capabilities
We want users to be able to remove their accounts without any moderator intervention. This is more inline with GDPR and this avoid loosing time on this.
But right now, we do this manually because we want to insure user did not vandalise products, as with account deletion, we loose the changeset.

To change it, we want to keep a unique userid for the changeset (so that we can visualize changes). This enables us to let people delete their account themselves.

- [ ] change account deletion procedure, so that instead of attributing changes to "openfoodfact-contributors" we attribute them to "anonymous-HASH" where HASH is a random hash (with reasonable collision probability)
- [x] implement an account deletion page accessible to all users (should be accessible from profile page, but a distinct page, with a proper explanation on consequences and a confirmation, beware to only enable a user to remove his own account)

### Part of
- #5809 

Better do this in conjunction with:
- #8270

john-gom:
### What

When deleting a user renames them to anonymous with a unique random id. Increased MongoDB timeouts

- Fixes #8164


lib/ProductOpener/Products.pm
```diff
@@ -1925,7 +1926,7 @@ sub replace_user_id_in_product ($product_id, $user_id, $new_user_id) {
 	}
 
 	if ((defined $most_recent_product_ref) and (not $most_recent_product_ref->{deleted})) {
-		my $products_collection = get_products_collection();
```
stephanegigandet, line 1928: Could we make the field mandatory, and not have a default?

Then we would need a change in scripts/find_and_replace_user_id_in_products.pl as well

john-gom: Actually no need to change the script as the script uses the parent, plural sub

stephanegigandet (approved):
Thank you!

PR #8473, opened 2023-05-29
taxonomy: Added German variant of inositol

c-schuhmann:
### What

Added the German variant "Inosit" to inositol, which is often contained in beverages like energy drinks. In German, the word "Inosit" is more common than "Inositol".

### DIFF
--- a/taxonomies/nutrients.txt
+++ b/taxonomies/nutrients.txt
@@ -6415,7 +6415,7 @@ bg:Инозитол
 ca:inositol
 cs:inositol
 da:inositol
-de:Inositol, Inosit
+de:Inosit, Inositol
 el:Ινοσιτόλη
 es:inositol
 et:Inositool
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Thank you!

stephanegigandet:
@c-schuhmann Would you have some pictures of nutrition facts table in German with Inosit or Inositol? If Inosit is much more frequent, then we could make it the main name.

c-schuhmann:
@stephanegigandet Yes, please see this example: https://world.openfoodfacts.org/product/5060639126897/monster-energy-ultra-paradise (bottom left of the ingredients list)

I would say we could set "Inosit" as the main German name of this ingredient, should we swap the two names (like "de:Inosit, Inositol") in this case?

stephanegigandet:
> I would say we could set "Inosit" as the main German name of this ingredient, should we swap the two names (like "de:Inosit, Inositol") in this case?

Yes, we should do that, it's indeed much more frequent in German: https://de.openfoodfacts.org/ingredients?filter=inosit

c-schuhmann:
Sorry for the late reply, thank you for your info & merging!

PR #8491, opened 2023-05-31
taxonomy: Add Spanish/Catalan sugars for creatine products

duhow:
From [product 8424644002497](https://es.openfoodfacts.org/producto/8424644002497) , there is no nutrition table, but instead a list of components weight.
Adding few translations for sugars in Spanish and Catalan, and a bit of some additional info.


alexgarel:
@CharlesNepote could you review ?

CharlesNepote (approved):
Looks good to me.

PR #8511, opened 2023-06-03
feat: Able to view the list of users associated with an Organization

### LINKED ISSUE #6978: Being able to view the list of users attached to an organisation on the pro platform
### What
- It would be helpful to know which users are behind an organisation so that I can contact them easily.

### Solution

- ~~I can imagine the solution as a button on the left panel : "View the organisation users".~~
- it could be on the org profile page. 

It would be great to have : 
- userid: kurzeja12
- name: kurzeja12
- email: kur****2@blabla.com
- langage (lc)
- country (cc)
- date of last activity
- In the future: CRM status? 

+ Any other field you might find relevant :)

### Part of
- #5521

MonalikaPatnaik:
### What

Iterated over the keys of the `members ` hash reference in `$user_or_org_ref ` and the reference to @org_members is assigned to the `org_members` key of the `$user_template_data_ref` hash reference in order to access the list within the template.

### Screenshot
Currently the table is looking like this without CSS :') will try adding a generic CSS to the tables in a separate PR..for smaller screens the table contents are hidden from right.

image

- Fixes #6978


lib/ProductOpener/Display.pm
```diff
@@ -3898,6 +3898,11 @@ HTML
 						$user_template_data_ref->{edit_profile} = 1;
 						$user_template_data_ref->{orgid} = $orgid;
 					}
+					my @org_members;
+					foreach my $member_id (keys %{$user_or_org_ref->{members}}) {
```
stephanegigandet, line 3902: In this loop, you will need to load the user file that contains details such as the e-mail, language, country etc.

In Users.pm and Display.pm, we have code like this to do that:

`my $user_file = "$data_root/users/" . get_string_id_for_lang("no_language", $user_id) . ".sto";

			if (-e $user_file) {
				$user_ref = retrieve($user_file);`

It would be a good thing to create a retrieve_user($user_id) function in Users.pm to do that.

lib/ProductOpener/Display.pm
```diff
@@ -3898,6 +3898,11 @@ HTML
 						$user_template_data_ref->{edit_profile} = 1;
 						$user_template_data_ref->{orgid} = $orgid;
 					}
+					my @org_members;
```
stephanegigandet, line 3901: This feature should be available only to moderator of the pro platform, so we should have a condition on the value of $User{pro_moderator}, if the user is not a moderator, we don't create the structure with org members details.

MonalikaPatnaik:
I think it should work fine..only issues with the translatable string( make build_lang is not working in pro shell maybe)..Also, pls lmk when I have to add CSS to the table..will do that<3

lib/ProductOpener/Display.pm
```diff
@@ -3898,6 +3898,14 @@ HTML
 						$user_template_data_ref->{edit_profile} = 1;
 						$user_template_data_ref->{orgid} = $orgid;
 					}
+					if (defined $User{pro_moderator}) {
+						my @org_members;
+						foreach my $member_id (keys %{$user_or_org_ref->{members}}) {
```
stephanegigandet, line 3903: Let's add a sort so that we always get users in the same order. In Perl keys %some_hash can return keys in different order every time it's executed.

```suggestion
						foreach my $member_id (sort keys %{$user_or_org_ref->{members}}) {
```

po/common/common.pot
```diff
@@ -5039,6 +5039,38 @@ msgctxt "official_site"
...
+
+msgctxt "number_of_members"
+msgid "Number of Members"
+msgstr "Number of Members"
+
+msgctxt "serial_no"
+msgid "S.No"
+msgstr "S.No"
+
+msgctxt "member_userid"
```
stephanegigandet, line 5054: We already have strings for "Name", "Email", "Country", "Language" etc. in the .po file, so we can reuse them without adding new translation strings.

templates/web/pages/org_profile/org_profile.tt.html
```diff
@@ -14,6 +14,37 @@
 	<p>&rarr; <a href="[% link %]">[% lang("official_site") %]</a></p>
 [% END %]
 
+[% IF org_members.size %]
+<h2>[% lang("organization_members") %]</h2>
+	<p>[% lang("number_of_members") %]: [% org_members.size %]</p>
```
stephanegigandet, line 19:
```suggestion
	<p>[% lang("number_of_members") %][% sep %]: [% org_members.size %]</p>
```

This will add a space in some languages (like French) where we need a space before the : character.

MonalikaPatnaik:
> For the table styling, we can use the same styles than we have for the packagings table when you edit a product:
> 
> image
> 
> Currently we have specific styles for the packaging edition, but maybe we can make those styles more generic, and apply them to both the packaging edition and your table.
> 
> `#packagings_components_edit_header { background-color:$macchiato; border-radius:20px; color:white; font-weight:600; margin:0; padding:1rem; }
> 
> .packagings_components_edit_row { border-radius:20px; margin:0 !important; padding:1rem; } .packagings_components_edit_row:nth-of-type(even) { background-color:$latte; } `

Thankyou so much...but the thing is I am not able to see the CSS changes in my localhost even after running make restart( tho not req but still)..not able to verify the CSS changes as it is not showing up :')

stephanegigandet:
> Thankyou so much...but the thing is I am not able to see the CSS changes in my localhost even after running make restart( tho not req but still)..not able to verify the CSS changes as it is not showing up :')

I believe this is because the po_pro-dynamicfront that watches for scss changes is not started on the pro platform:

`(pro) docker-compose start
[+] Running 6/7
 ✔ Container po_pro-mongodb-1       Started                                                        0.6s 
 ⠧ Container po_pro-dynamicfront-1  Starting                                                       1.8s 
 ✔ Container po_pro-memcached-1     Started                                                        0.3s 
 ✔ Container po_pro-minion-1        Started                                                        1.0s 
 ✔ Container po_pro-incron-1        Started                                                        0.5s 
 ✔ Container po_pro-backend-1       Started                                                        1.1s 
 ✔ Container po_pro-frontend-1      Started                                                        0.4s 
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/home/stephane/openfoodfacts-server/gulpfile.js" to rootfs at "/opt/product-opener/gulpfile.js": mount /home/stephane/openfoodfacts-server/gulpfile.js:/opt/product-opener/gulpfile.js (via /proc/self/fd/6), flags: 0x5000: not a directory: unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
`

Short term: as this feature should also be accessible on the public platform (not only the pro platform), you could test it in a non-pro environment.

Longer term: we need to fix the po_pro-dynamicfront bug.

lib/ProductOpener/Display.pm
```diff
@@ -3898,6 +3898,14 @@ HTML
 						$user_template_data_ref->{edit_profile} = 1;
 						$user_template_data_ref->{orgid} = $orgid;
 					}
+					if (defined $User{pro_moderator}) {
+						my @org_members;
+						foreach my $member_id (sort keys %{$user_or_org_ref->{members}}) {
+							my $member_details = retrieve_user($member_id);
```
stephanegigandet, line 3904: As a convention, when a variable is an object reference, we suffix it with _ref. Coul you rename $member_details to $member_user_ref or similar?

po/common/common.pot
stephanegigandet: Instead of "member_userid" can you use the existing "username" string?

templates/web/pages/org_profile/org_profile.tt.html
```diff
@@ -14,6 +14,41 @@
...
+			<th>[% lang("name") %]</th>
+			<th>[% lang("email") %]</th>
+			<th>[% lang("select_lang") %]</th>
+			<th>[% lang("select_country") %]</th>
+
+		</tr>
+	</thead>
+		[% SET count = 1 %]
+		[% FOREACH users IN org_members %]
+		<tbody id="org_table_body">
```
stephanegigandet, line 35: the tbody element should be outside the FOREACH loop, otherwise you will have a table with more than 1 tbody.

templates/web/pages/org_profile/org_profile.tt.html
```diff
@@ -14,6 +14,41 @@
...
+		<tr>
+			<td data-cell =[% lang("serial_no") %]>[% count %].</td>
```
stephanegigandet, line 37: What is "data-cell =[% lang("serial_no") %]" for?

scss/_off.scss
```diff
@@ -1451,3 +1451,50 @@ a.panel_title {
     background-color: scale-color($button-bg-color, $lightness: -$button-function-factor / 2);
     color: #fff;
 }
+// organization members table in producers platform
```
stephanegigandet, line 1454: We can do it in a separate PR, but it would be a good thing to see if we can make the CSS code that we have for packagings generic, so that it can be applied to any table, instead of having similar separate CSS code for each table.

MonalikaPatnaik: Yes..It would be good to make the CSS generic for these tables👍...would open a separate PR for CSS things..will do it asap!

templates/web/pages/org_profile/org_profile.tt.html
MonalikaPatnaik: The data-cell is a custom attribute used to store additional data or metadata related to the HTML element. This attribute is being used by CSS to manipulate the element for smaller screens--
```
td::before {
      content:attr(data-cell) ": ";
       float:left;
      text-transform: capitalize;
      font-weight: 600;
}
```
It would add : after the data cell value i.e serial no in this case when the screen width is small

stephanegigandet: I understand, thanks for the explanation.

stephanegigandet (approved):
Looks good, thank you very much for the PR and incorporating all the feedback and suggestions!

PR #8516, opened 2023-06-05
feat: Run user deletion as a background task

### LINKED ISSUE #8270: Make account data removal asynchronous and batched
### Describe the bug

When you remove a user, there are a lot of chances that your request ends up with an error because we got a operation exceeded time limite.

This is because on user removal we must process a lot of mongodb data to change the username.

### To Reproduce

Remove a user in production.

### Expected behavior

We should have a nice return. The user removal should proceed safely in the background.

We can use minions to do it. 
We should also do operation in batch to avoid flooding the database. 

### Screenshots

image

### Additional context

_No response_

### Type of device

Browser

### Browser version

_No response_

_No response_

### Time per product

_No response_

john-gom:
### What

User deletion is run as a minion job rather than synchronously

- Fixes #8270

### DIFF
--- a/lib/ProductOpener/Producers.pm
+++ b/lib/ProductOpener/Producers.pm
@@ -2008,4 +2008,8 @@ sub update_export_status_for_csv_file_task ($job, $args_ref) {
 	return;
 }
 
+sub queue_job { ## no critic (Subroutines::RequireArgUnpacking)
+	return $minion->enqueue(@_);
+}
+
 1;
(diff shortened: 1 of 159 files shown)


stephanegigandet (approved):
Wonderful, thank you.

The only thing is for the .po files: in practice we need only to change en.po and common.pot, as the other files will be populated by PRs made through Crowdin. In practice Product Opener uses the English translation if there's no translation in a local language. Could you remove the changes to the other .po files? Otherwise it will create lots of diffs (most of them will be removed unless we already have a translation) in the next Crowdin update.

stephanegigandet (approved):
Thank you!

PR #8517, opened 2023-06-05
taxonomy: Add CCPAE logo

duhow:
## What

Add CCPAE translations and logo from their website: https://www.ccpae.org/index.php?option=com_content&task=view&id=1161&Itemid=254&lang=en

- Resized original image to 56x90
- Rendered AI file to SVG (larger resolution)


alexgarel (changes requested):
Hi @duhow thank a lot for your contribution, and sorry for the delay.

You have to resize image so that's it's consistent with other labels (90px height). The image name must, of course, reflect new dimensions.

Would you mind changing that ?

alexgarel (approved):
I moved the logo to "en" otherwise it would only have been visible in catalan.

Also it was not using the right name.

alexgarel:
Thanks @duhow !

PR #8532, opened 2023-06-07
fix: Restore Search Bar in Producers Platform

### LINKED ISSUE #8413: Restore the search bar in the Pro platform
For design purpose, the search bar was removed from the "Welcome page" in the Pro platform (the page you can see before creating an account).
The consequence is that now, the search bar is not available at all whereas I think it would be use for producers to search products. 
Could it be restored on the page, but only when a producer is logged in? :)

cc @stephanegigandet 
### Part of
- #5521 
### Screenshot

image

MonalikaPatnaik:
### What
Search bar is shown only when the producer is logged in or it's user id is defined. 

### Screenshot
image

- Fixes #8413

### DIFF
--- a/templates/web/common/site_layout.tt.html
+++ b/templates/web/common/site_layout.tt.html
@@ -190,8 +190,8 @@
 						</li>
 					</ul>
 
-					[%# Don't show the search bar on the producers platform %]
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Looks good, thank you!

PR #8533, opened 2023-06-08
fix: Added Template Comments instead of html

MonalikaPatnaik:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/templates/web/common/producers_resources_list.tt.html
+++ b/templates/web/common/producers_resources_list.tt.html
@@ -1,5 +1,5 @@
-<!-- Please include the following code within a list -->
-<!-- Example: <ul> ... [% INCLUDE 'web/common/producers_resources_list.tt.html' %] ... </ul> -->
+[%# Please include the following code within a list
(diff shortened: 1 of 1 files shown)


templates/web/common/producers_resources_list.tt.html
```diff
@@ -1,5 +1,5 @@
-<!-- Please include the following code within a list -->
-<!-- Example: <ul> ... [% INCLUDE 'web/common/producers_resources_list.tt.html' %] ... </ul> -->
+[%# Please include the following code within a list
+Example: <ul> ... [% INCLUDE 'web/common/producers_resources_list.tt.html' %] ... </ul> %]
```
hangy, line 2: Does the template engine understand that the first `%]` doesn't close the comment?

MonalikaPatnaik: Not sure..tho it is not giving any kind of error🥲looking at it for clarification👀

MonalikaPatnaik: I think it won't close the comment by first %]

stephanegigandet (approved):
Looks good

PR #8538, opened 2023-06-08
fix: Replace existing values with Producer Supplied Allergens and Traces values

### LINKED ISSUE #8522: Make allergens and traces values supplied by producers completely replace existing values
The usual behaviour for "tags" fields such as categories, labels, countries, allergens etc. is that values supplied by producers are added to any existing value entered by users (e.g. producers might sent us only the "Organic" label while we may have more precise labels and mentions such as "AB Organic" or "Rich in fiber").

For allergens and traces though, if the producer sends a value, we can assume that the value is complete, and that it can replace entirely any pre-existing value.

We need to do 2 things:

1. On the pro platform, when importing allergens and traces, if the producers supplies a value (non empty), then we replace existing values.
2. When exporting data from the pro platform and importing it to the public platform, if there is a non empty value for allergens, we replace existing values.

Related to https://github.com/openfoodfacts/openfoodfacts-server/issues/7850

### Part of
- #5521

MonalikaPatnaik:
### What

Added relevant code in import.pm to check whether the producer enters a non-empty value for the field defined in `$options{replace_existing_values_when_importing_those_tags_fields}`, if yes then it removes the existing values by setting the field to an empty string and deleting the corresponding tags.
Updated `import_csv_file.t` in `tests/integration/ import_csv_file.t` so that it can import and run more than one CSV file..also added the results in `tests/integration/expected_test_results/import_csv_file` ( I think the expected results looks good..lmk if I'm mistaken)

- Fixes #8522


stephanegigandet:
> Couldn't configure what to change in `Config_off.pm`

Config_off.pm contains options and list of fields that have specific features. We use this file to configure Product Opener: on Open Beauty Facts for instance, products have different fields than Open Food Facts (e.g. there's no nutrition facts for cosmetics).

So you could add something like this:

```
$options{replace_existing_values_when_importing_those_tags_fields} = {
	"allergens" => 1,
	"traces" => 1,
};
```

lib/ProductOpener/Import.pm
```diff
@@ -519,6 +519,14 @@ sub set_field_value (
 			$product_ref->{$field} = "";
 			delete $product_ref->{$field . "_tags"};
 		}
+
+		if (($server_options{producers_platform}) and (($field eq "allergens") or ($field eq "traces"))) {
```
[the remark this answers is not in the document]
MonalikaPatnaik: This means I have to write a separate condition for the public platform right?

stephanegigandet:
> how to write test in `import_csv_file.t`

import_csv_file.t currently imports one CSV file, and then compare the resulting products with the expected output.

to test this feature, you will need to import 2 CSV files: one where you set initial values for a few products, with or without allergens, labels etc. And then another one where you update those products (with the same barcode) with possibly different allergens, traces, labels etc.

What I suggest is to create those 2 new CSV files in tests/integration/inputs/import_csv_file (with names like replace_existing_values_1.csv and replace_existing_values_2.csv for instance), and that you first try to test your feature, by calling $stats_ref = ProductOpener::Import::import_csv_file($args); twice, for each of the files.

Then once that works, you could then change import_csv_file.t so that it can run more than 1 test: the one we had before, plus the one you add. You can look at how we did it in tests/integration/convert_and_import_excel_file.t :+1: 

We created a @tests structure:

`my @tests = (
	{
		test_case => "test",
		excel_file => "test.xlsx",
		columns_fields_json => "test.columns_fields.json",
		default_values => {lc => "en", countries => "en", brands => "Default brand"},
	},
	{
		test_case => "packagings-mousquetaires",
		excel_file => "packagings-mousquetaires.xlsx",
		columns_fields_json => "packagings-mousquetaires.columns_fields.json",
		default_values => {lc => "fr", countries => "fr"},
	}
);`

and then we iterate on it.

lib/ProductOpener/Import.pm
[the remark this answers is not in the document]
stephanegigandet: You can just remove the condition for the pro platform:

```suggestion
		if (($field eq "allergens") or ($field eq "traces")) {
```

lib/ProductOpener/Import.pm
```diff
@@ -519,6 +519,14 @@ sub set_field_value (
 			$product_ref->{$field} = "";
 			delete $product_ref->{$field . "_tags"};
 		}
+
+		if (($field eq "allergens") or ($field eq "traces")) {
```
stephanegigandet, line 523: You can use $options{replace_existing_values_when_importing_those_tags_fields} in this condition.

tests/integration/import_csv_file.t
```diff
@@ -42,20 +42,40 @@ sub fake_download_image ($) {
 	return $response;
 }
 
+my @tests = (
+	{
+		test_case => "test",
+		csv_file => "test.csv",
+	},
+	{
+		test_case => "replace_existing_values_1",
```
stephanegigandet, line 51: For this test case, you want to import 2 CSV files, so instead of having a csv_file argument, you could have a csv_files argument:

```
my @tests = (
	{
		test_case => "test",
		csv_files => ["test.csv"],
	},
	{
		test_case => "replace_existing_values",
        csv_files => ["replace_existing_values_1.csv", "replace_existing_values_1.csv"],
},
```

You cannot have the replace_existing_values 1 and 2 in 2 different tests cases, because when we iterate on the test cases, one of the first things we do is remove_all_products()

lib/ProductOpener/Test.pm
```diff
@@ -259,6 +260,19 @@ sub remove_all_orgs () {
...
 
+=head2 allergens_traces_replace_existing_values ()
+
+For integration tests, we need to start from an empty organizations base
+
+This function should only be called by tests, and never on production environments.
+
+=cut
+
+sub allergens_traces_replace_existing_values(){
```
stephanegigandet, line 271: What is this function for?

lib/ProductOpener/Import.pm
```diff
@@ -519,6 +519,15 @@ sub set_field_value (
 			$product_ref->{$field} = "";
 			delete $product_ref->{$field . "_tags"};
 		}
+
+		# If we are on the producers platform, replace existing values by producer supplied values for allergens and traces
+		if (defined $options{replace_existing_values_when_importing_those_tags_fields}->{$field}) {
```
stephanegigandet, line 524: I think you don't need the ->

One thing is that this code will fail of $options{replace_existing_values_when_importing_those_tags_fields] does not exist (for instance in Config_obf.pm).

So instead you can add this import at the top of the file:

use Data::DeepAccess qw(deep_exists);

and then use this condition

if (deep_exists(\%options, "replace_existing_values_when_importing_those_tags_fields", $field))

this function checks that each of the fields exist in a nested hash like %options

stephanegigandet (approved):
Looks good, thank you!

PR #8548, opened 2023-06-12
feat: Allow users to delete their own accounts

### LINKED ISSUE #6435: Allow to initiate account deletion from the website/API
### What
- [ ] Allow to initiate account deletion from the website/API
  - [ ] Minimal version: Link pre-filling an email to contact@openfoodfacts.org
  - [ ] Maximal version: self-service system, with an API for the apps
### Why
`
Account deletion within apps
App Store Review [Guideline 5.1.1](https://developer.apple.com/app-store/review/guidelines/#5.1.1) provides people with greater control over their personal data by stating that all apps that allow for account creation must also allow users to initiate deletion of their account from within the app. This requirement was set to apply to all app submissions by January 31, 2022. Due to the complexity of implementing this requirement, we’ve extended the deadline to June 30, 2022 to give you more time.
`

### Part of
- #5809
- #6436

john-gom:
### What

Enable account deletion for regular users with suitable warnings

### Screenshot
image

image

- Fixes #6435


stephanegigandet:
Having a separate Delete button should be relatively easy to add through the template I think.

stephanegigandet:
> I've just pushed a change to do this

Thanks @john-gom 

Maybe we could put the "Danger zone" after the regular "Save" button:

image

(maybe with an extra <br> or 2 or a css style to separate well the danger zone from the rest of the form).

What do you think?

john-gom:
> > I've just pushed a change to do this
> 
> Thanks @john-gom
> 
> Maybe we could put the "Danger zone" after the regular "Save" button:
> 
> image
> 
> (maybe with an extra or 2 or a css style to separate well the danger zone from the rest of the form).
> 
> What do you think?

Very easy to do but when I look at it I'm a bit concerned as the "call to action" for a form is typically at the bottom so I think having delete there could result in it being more likely to be clicked.

stephanegigandet (approved):
Thank you very much!

PR #8557, opened 2023-06-13
fix: Build Taxonomies does not cache the full json files

### LINKED ISSUE #8556: Build Taxonomies does not cache the full json files
### Describe the bug

The build cache mechanism used for taxonomies only checks if the result.sto, result.txt and json files are in the cache, hence if a taxonomy is retrieved from the cache the full.json file will not be fetched

### To Reproduce

Checkout the main branch and run the following:

```make build_taxonomies```

Check the /opt/product-opener/html/data/taxonomies folder. Only the json files will be there and no .full.json files

### Expected behavior

The full.json files should also be cached and fetched from the cache

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Other

### Browser version

_No response_

_No response_

### Time per product

_No response_

john-gom:
### What

Also cache full.json files

- Fixes #8556

### DIFF
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -965,6 +965,9 @@ sub get_from_cache ($tagtype, @files) {
 		$got_from_cache = get_file_from_cache("$cache_prefix.json", "$tag_www_root.json");
 	}
 	if ($got_from_cache) {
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Thank you!

PR #8561, opened 2023-06-14
feat: allow admins to specify a timeout for MongoDB queries

stephanegigandet:
This PR adds a &timeout=[number of ms] parameter that can be used by admins in order to override the default timeout (currently 50s in production), in order to be able to run queries that run longer. This could be useful for instance for monitoring queries (e.g. getting stats on recognized ingredients).

Note that the proxy timeout in nginx is currently not set (except on the pro platform where we can have big uploads), and the default is 60 sec. So in order for the timeout to have a real effect, we would also need to change the nginx config to include something like:

```
        proxy_connect_timeout       1200;
        proxy_send_timeout          1200;
        proxy_read_timeout          1200;
        send_timeout                1200;
```

### DIFF
--- a/lib/ProductOpener/Data.pm
+++ b/lib/ProductOpener/Data.pm
@@ -139,7 +139,7 @@ A reference to a parameters object that can be passed to get_products_collection
 
 sub get_products_collection_request_parameters ($request_ref, $additional_parameters_ref = {} ) {
 
-	my $parameters_ref = {}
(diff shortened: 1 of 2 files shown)


alexgarel:
@stephanegigandet if you have to change nginx, you can do it in the PR (see conf folder), even if at release time you have to copy it (on off1).

alexgarel (approved):
Cool.

PR #8574, opened 2023-06-16
taxonomy: Finnish translations for doughnut categories

Ban3:
### What

* Added English translation as the main entry for `fr:Beignets fourrés à la framboise`
* Added 'Munkit' as the main Finnish translation for Doughnuts. It was 'donitsit', which most of the time refers to 'Ring Doughnuts'.
Source: https://fi.wikipedia.org/wiki/Munkki_(leivonnainen)
* Added Finnish translations for doughnut subcategories.


stephanegigandet (approved):
Thank you!

PR #8577, opened 2023-06-18
feat: Adding Tutorial about Account Creation in Pro platform

### LINKED ISSUE #8528: Add tuto about account creating in the Pro platform
**Video link:** https://www.youtube.com/watch?v=RXKxDuuKYV0&list=PLjAH-USadsF3xgSVfUUBS4we3XBr-1r55&index=3

**Where to add it:** https://world.pro.openfoodfacts.org/cgi/user.pl
Youtube provides auto-translation so we could publish it on all platforms (fr.pro / it.pro / de.pro / es.pro / etc)

**Mockup:**
image

### Part of
- #5521

MonalikaPatnaik:
### What
Made relevant changes in `templates\web\pages\user_form\user_form_page.tt.html` for adding video beside the form.

### Screenshot
image

### for medium and smaller screens
image
image

- Fixes #8528


templates/web/pages/user_form/user_form_page.tt.html
```diff
@@ -45,7 +45,9 @@
 
     <!-- Start form -->
 
-    <form method="post" action="/cgi/user.pl" enctype="multipart/form-data" id="user_form">
+<div class="donation-banner__content">
```
stephanegigandet, line 48: Why do we have a class for the donation-banner ?

templates/web/pages/user_form/user_form_page.tt.html
```diff
@@ -97,6 +99,17 @@
                 [% END %]
 
             [% END %]
+        </form>
```
stephanegigandet, line 102: That's strange, if you put `</form>` here, don't you get inputs, buttons etc. that are outside the `<form>..</form>` ?

templates/web/pages/user_form/user_form_page.tt.html
```diff
@@ -97,6 +99,17 @@
...
+    </div>
+    <!-- show video about account creation in Pro Platform -->
```
stephanegigandet, line 104: We don't want the video displayed when we are not on the pro platform, is that the case?

templates/web/pages/user_form/user_form_page.tt.html
MonalikaPatnaik: Because the css is same i.e. for smaller screens we want the flex-direction to be column instead of row, so instead of writing separate css for this div container, i reused the css for donation-banner class..
If you want separate css for this div that we can even do it in that way too!

templates/web/pages/user_form/user_form_page.tt.html
```diff
@@ -60,44 +60,57 @@
...
-                            [% field.label %]
-                        </label>
-                    [% END %]
-                [% ELSIF field.type == 'hidden' %]
-                    <label for="pro">
-                        <input type="hidden" name="[% field.field %]" value="[% field.value %]">
-                    </label>
+                </div>
+                <!-- show video about account creation beside user form section-->
+                [% IF section.id != "professional" AND section.id != "contributor_settings" %]
```
stephanegigandet, line 103: the condition should probably be reversed: instead of showing the video when the section is not something and not something else, it should be shown only for the section we want it to be shown. Otherwise it will be shown in other sections (e.g. the administrator section, or any new section we add later)

templates/web/pages/user_form/user_form_page.tt.html
alexgarel: Maybe it's better to duplicate the class @MonalikaPatnaik so that it's more meaning full (you can add the class to same css rules as donation-banner__content.

MonalikaPatnaik: Done @alexgarel ✅

stephanegigandet (approved):
Looks good, thank you!

PR #8585, opened 2023-06-19
fix: add LCA icon in Eco-Score panel

### LINKED ISSUE #8563: The LCA icon seems to have disappeared
The LCA icon seems to have disappeared
image

https://fr.openfoodfacts.org/produit/20511296/wholemeal-seeded-farmhouse-deluxe#panel_ecoscore

stephanegigandet:
Add an icon in the Eco-Score subpanel for LCA.

Note that the LCA icon is always grey, as we currently have a mechanism only to change the color according to a good/average/bad evaluation, but not to a A to E grade color.

Fixes #8563

image

### DIFF
--- a/templates/api/knowledge-panels/environment/ecoscore/agribalyse.tt.json
+++ b/templates/api/knowledge-panels/environment/ecoscore/agribalyse.tt.json
@@ -4,6 +4,9 @@
         "environment"
     ],
     "title_element": {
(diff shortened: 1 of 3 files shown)


stephanegigandet:
@teolemon can we merge this? it's likely to generate more merge conflicts if we wait

teolemon (approved):

PR #8594, opened 2023-06-20
feat: packaging stats per parent materials

stephanegigandet:
This PR:
- computes a new packagings_materials structure for each product, that agregates weights by parent material (platics, glass, paper/cardboard, metal) for each packaging component
- displays the data in a knowledge panel

image


teolemon:
@stephanegigandet could you add a comparison to category ?

teolemon:
@stephanegigandet could you swap column A and column B ?

stephanegigandet:
> @stephanegigandet could you add a comparison to category ?

Well it will be the next step, we first need to aggregate by product, then we can aggregate by category, and we can show a comparison to the category on each product. It will be much more work though.

templates/api/knowledge-panels/environment/packaging_components.tt.json
```diff
@@ -0,0 +1,64 @@
...
+[% SET recycling_types = {} %]
+[% SET unknown = "en:unknown" %]
+[% IF product.packagings %]
+    [% FOREACH packaging IN product.packagings %]
+        [% IF packaging.recycling.defined AND ((packaging.recycling == "en:discard") OR (packaging.recycling == "en:recycle")) %]
+            [% SET recycling_types.${packaging.recycling} = 1 %]
+        [% ELSE %]
+            [% SET recycling_types.$unknown = 1 %]
+        [% END %]
+    [% END %]
```
alexgarel, line 11:
```suggestion
    [%# Check which of recycling / discard / unknown packaging types we have to build each panel (one for each) %]
    [% FOREACH packaging IN product.packagings %]
        [% IF packaging.recycling.defined AND ((packaging.recycling == "en:discard") OR (packaging.recycling == "en:recycle")) %]
            [% SET recycling_types.${packaging.recycling} = 1 %]
        [% ELSE %]
            [% SET recycling_types.$unknown = 1 %]
        [% END %]
    [% END %]
```

templates/api/knowledge-panels/environment/packaging_components.tt.json
```diff
@@ -0,0 +1,64 @@
...
+                    "icon_url": "[% static_subdomain %]/images/icons/dist/delete.svg",
+                    "icon_alt": "[% display_taxonomy_tag_name("packaging_recycling",recycling_type) %]",
+                    [% ELSE %]
+                    "evaluation": "neutral",
+                    "icon_url": "[% static_subdomain %]/images/icons/dist/help.svg",
+                    "icon_alt": "[% lang('unknown') %]",
+                    [% END %]                
+                    "html": `
+                        [% FOREACH packaging IN product.packagings %]
+                            [% IF packaging.recycling == recycling_type OR (recycling_type == "en:unknown" AND ((NOT packaging.recycling.defined) OR ((packaging.recycling != "en:discard") AND (packaging.recycling != "en:recycle")))) %]
```
alexgarel, line 46: A better way to do that, would have been to dispatch packagings in arrays in  recycling_types instead of putting `=> 1`. So that you just have to iterate over the hash, and elements in array.

Because this condition is not trivial to read !

tests/unit/packaging.t
```diff
@@ -655,6 +655,56 @@ is_deeply(
...
+		},
+		'en:paper-or-cardboard' => {
+			'weight' => 10,
+			'weight_100g' => '5',
+			'weight_percent' => '1.81818181818182'
+		},
+		'en:plastic' => {}
+	}
+
+) or diag explain $product_ref->{packagings_materials};
```
alexgarel, line 708: Would be good to add at least one test with an unknown material.

I always like to also test with an empty hash.

alexgarel (approved):
Thank you.

Approving, but I point room for improvement. I don't know if you have the time to pick some.

lib/ProductOpener/Packaging.pm
```diff
@@ -846,6 +851,71 @@ sub set_packaging_misc_tags ($product_ref) {
...
+			my $material = $packaging_ref->{material};
+			my $parent_material = "en:unknown";
+			if (defined $material) {
+				foreach my $parent ("en:paper-or-cardboard", "en:plastic", "en:glass", "en:metal") {
+					if (is_a("packaging_materials", $material, $parent)) {
+						$parent_material = $parent;
+						last;
+					}
+				}
+			}
```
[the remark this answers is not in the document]
garu: I don't know enough of the source code to suggest an entirely different approach to the logic itself, but a common pattern for getting the first item in a list that matches a certain criteria is using [List::Util's `first`](https://metacpan.org/pod/List::Util#first):

```perl
use List::Util qw(first);

$parent_material = first {
      is_a("packaging_materials", $material, $_)
} ("en:paper-or-cardboard", "en:plastic", "en:glass", "en:metal");
``` 

but if this code is hot (i.e. running several times per call) maybe it could be worth turning `is_a()` into a hash lookup.

Either way, the current implementation is okay and I wouldn't change it unless it was a bottleneck.

templates/api/knowledge-panels/environment/packaging_components.tt.json
stephanegigandet: Agreed, but it's likely that we will change how we display packaging components in the future. This code did not change, I just renamed the template for clarity.

lib/ProductOpener/Packaging.pm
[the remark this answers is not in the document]
stephanegigandet: Thanks for the suggestions. I changed to @garu 's suggestion as I find it more readable with first indeed.

It certainly could be done more efficiently, but this code is executed only when writing products, and given the shallowness of the packaging materials taxonomy, it's just a dozen operations, so I'd prefer not to spend time on it. If we want to optimize things, there are targets that are much more likely to make a difference (e.g. for read operations of products, where we compute attributes and knowledge panels for 100 products...)

PR #8596, opened 2023-06-21
fix: remove source files in agena3000 sftp

stephanegigandet:
There is a quota on the sftp account Agena3000 created for us. This is to remove the files that we have already received.


alexgarel (approved):
LGTM

PR #8610, opened 2023-06-27
fix: Added Protection On Images for normal Users

### LINKED ISSUE #2086: Identify and protect images and data sent by producers
This is to be done in combination with https://github.com/openfoodfacts/openfoodfacts-server/issues/2085

We currently have a rather crude approach that blocks edits from the API (apps, robotoff etc.) when product data has been imported. The new approach will specifically mark which fields have been sent by the producer.

1. use the product edit history to create a new producer_provided_fields array with the list of fields added/changed by the manufacturer.
- when the fields were changed by import_csv_file.pl
- or when the fields were changed by a user that is marked as a producer who sends us complete data

2. allow only specific expert users to change those fields

Tag fields like labels etc. can still be added to, and images can still be added (not not selected if an existing image is present).

### Part of
- #5521

MonalikaPatnaik:

- Fixes #2086


tests/integration/protected_images.t
```diff
@@ -0,0 +1,364 @@
...
+	labels => "organic",
+	origin => "france",
+	brands => "Test brand",
+	packaging_text_en =>
+		"1 wooden box to recycle, 6 25cl glass bottles to reuse, 3 steel lids to recycle, 1 plastic film to discard",
+	'nutriment_energy-kj' => 10,
+	nutriment_fat => 5,
+	nutriment_proteins => 2,
+	quantity => "90g",
+	nutrition_facts_image => "image_nutrition",
```
stephanegigandet, line 61: Images can't be sent or selected with product_jqm_multilingual.pl or product_multilingual.pl, you will need to use the image upload API: https://openfoodfacts.github.io/openfoodfacts-server/api/ref-v2/#post-/cgi/product_image_upload.pl

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -85,6 +85,34 @@ sub skip_protected_field ($product_ref, $field, $moderator = 0) {
...
 }
 
+=head2 skip_protected_image($product_ref, $imagefield, $moderator = 0)
+
+=cut
+
+sub skip_protected_image ($product_ref, $imagefield, $moderator = 0) {
+
+	# Add image-specific checks here
+	if (defined $product_ref->{imgupload_ $imagefield}) {
```
stephanegigandet, line 95: I don't think this will work: `$product_ref->{imgupload_ $imagefield})`

For fields, we have the owner_fields structure that tells us if a field value was set by its owner.

For images, it's more complex, as currently at least, we don't have images in owner_fields

If you look at this product for instance, you can look at the "images" field:

https://world.openfoodfacts.org/api/v3/product/3596710384396/lentilles-corail-l-oiseau?fields=images

then if we check what we have for the "front_fr" image:

```
front_fr: {
angle: 0,
coordinates_image_size: "400",
geometry: "0x0--7--7",
imgid: "16",
normalize: "false",
rev: "57",
sizes: {
100: {
h: 100,
w: 66
},
200: {
h: 200,
w: 132
},
400: {
h: 400,
w: 263
},
full: {
h: 2248,
w: 1480
}
},
white_magic: "false",
x1: "-1",
x2: "-1",
y1: "-1",
y2: "-1"
},
```

we have "imgid: 16", so we can check who uploaded that image:

```
{
sizes: {
100: {
h: 100,
w: 86
},
400: {
h: 400,
w: 344
},
full: {
h: 2831,
w: 2438
}
},
uploaded_t: 1688498812,
uploader: "org-auchan-apaw"
},
```

and here we see it's the producer who uploaded the image:  (org-auchan-apaw)

So in skip_protected_image, we should do that: identify the imgid of the selected image, and then identify the uploader and check if the uploader is the owner of the product.

To simplify testing, I think it would be useful to first create a function get_selected_image_uploader("front_fr") that would return the uploader of a selected image, and have unit tests for it.

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -85,6 +86,20 @@ sub skip_protected_field ($product_ref, $field, $moderator = 0) {
...
 }
 
+=head2 skip_protected_image($product_ref, $imagefield, $moderator = 0)
+
+=cut
+
+sub is_protected_image ($product_ref, $imagefield, $moderator = 0) {
+
+	my $selected_uploader = get_selected_image_uploader($product_ref, $imagefield);
+	if ((not $server_options{producers_platform}) and (not $moderator) and (defined $selected_uploader)) {
```
stephanegigandet, line 96: There should always be an uploader defined for each selected image. In order to know if the image is protected, we need to check if this uploader is the same as the owner of the product, with the owner field.

e.g. try this query: https://world.openfoodfacts.org/api/v3/product/8076800195057?fields=images,owner

the "front_fr" image is image imgid 69, and imgid 69's uploader is "org-barilla-france-sa" which is the same as the owner.

So this function should be changed to compare the uploader to the owner (if it exists).

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -85,6 +86,20 @@ sub skip_protected_field ($product_ref, $field, $moderator = 0) {
 	return 0;
 }
 
+=head2 skip_protected_image($product_ref, $imagefield, $moderator = 0)
+
+=cut
+
+sub is_protected_image ($product_ref, $imagefield, $moderator = 0) {
```
stephanegigandet, line 93: I think we should remove the $moderator from the parameter: the function should indicate if the image is protected or not. Then if an user is a moderator, we can decide not to check if the image is protected.

lib/ProductOpener/Images.pm
```diff
@@ -357,6 +358,23 @@ sub scan_code ($file) {
 	return $code;
 }
 
+sub get_selected_image_uploader ($product_ref, $imagefield) {
+
+	# Retrieve the product's image data
+	my $image_ref = $product_ref->{images}{$imagefield};
+
+	if ($image_ref && $image_ref->{imgid}) {
```
stephanegigandet, line 366: This will fail if $image_ref is not defined (if we try to get the uploader of a selected image that does not exist).

Instead we can do something like $imgid = deep_get($product_ref, "images", $imagefield, "imgid");

Then we can use deep_get as well to get the uploader

lib/ProductOpener/APIProductWrite.pm
stephanegigandet: You can get the owner from $product_ref->{owner}

tests/integration/protected_images.t
```diff
@@ -0,0 +1,103 @@
...
+	},
+);
+# create the products in the database
+foreach my $product_form_override (@products) {
+	edit_product($ua, $product_form_override);
+}
+
+# Setting the owner of the product
+my $product_ref = retrieve_product('0900000000139');
+$product_ref->{owner} = "sample", store_product("sample", $product_ref, "protecting image");
```
stephanegigandet, line 65: It's a good thing to set the owner to "sample" that way. Then for the following tests, you need 2 users: one user "sample" which is the owner, who should upload an image: the image should be protected as it's uploaded and selected by the owner.

And then you need to use another user (like the "tests" user) to try to upload and select a different image: it should not work as the image is protected.

cgi/product_image_upload.pl
```diff
@@ -313,7 +313,9 @@
 			# Changed 2020-04-20: don't overwrite selected images if the source is the product edit form
 			and (  (not defined single_param('source'))
 				or (single_param('source') ne "product_edit_form")
-				or (not defined $product_ref->{images}{$imagefield}))
+				or (not defined $product_ref->{images}{$imagefield}) and ($User{moderator})
```
stephanegigandet, line 316: There's an issue with the conditions here. We should keep all the previous conditions, and in addition to that, we need to check that the image is not protected (unless the user is the moderator)

MonalikaPatnaik:
Thankyou @stephanegigandet..will wrap this PR by today only❤️

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -85,6 +86,22 @@ sub skip_protected_field ($product_ref, $field, $moderator = 0) {
 	return 0;
 }
 
+=head2 is_protected_image($product_ref, $imagefield)
```
stephanegigandet, line 89: Could you move this function to Images.pm instead? It's not specific to the API (it's useful to know if an image is protected on the website as well, for instance we could display it to users when they edit the product page)

cgi/product_image_upload.pl
```diff
@@ -314,6 +315,8 @@
 			and (  (not defined single_param('source'))
 				or (single_param('source') ne "product_edit_form")
 				or (not defined $product_ref->{images}{$imagefield}))
+			and (not is_protected_image($product_ref, $imagefield) or $User{moderator})
```
stephanegigandet, line 318: We need to add a similar check on 2 other files:

product_image_crop.pl : it is called when you edit a product page on the web, and try to crop / select another image.

product_image_unselect.pl : used to unselect a selected image

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -53,12 +54,12 @@ use ProductOpener::API qw/:all/;
 use ProductOpener::Packaging qw/:all/;
 use ProductOpener::Text qw/:all/;
 use ProductOpener::Tags qw/:all/;
+use ProductOpener::Images qw/:all/;
 
 use Encode;
 
-=head2 skip_protected_field($product_ref, $field, $moderator = 0)
+=head2 skip_protected_field ($product_ref, $field, $moderator = 0)
```
stephanegigandet, line 61: the header was copy/pasted and does not match the function below I think

lib/ProductOpener/Images.pm
```diff
@@ -357,6 +359,17 @@ sub scan_code ($file) {
 	return $code;
 }
 
+sub get_selected_image_uploader ($product_ref, $imagefield) {
+
+	# Retrieve the product's image data
+	my $imgid = deep_get($product_ref, "images", $imagefield, "imgid");
+
+	# Retrieve the uploader of the image
```
stephanegigandet, line 367: Could you add a check that $imgid is defined? Otherwise it could trigger warnings if we try to retrieve a hash value for an undef key I think.

tests/integration/protected_images.t
```diff
@@ -0,0 +1,106 @@
...
+		test_case => 'post-select-protected-image',
+		method => 'POST',
+		path => '/cgi/product_image_upload.pl',
+		form => {
+			code => '0900000000139',
+			imagefield => "front_en",
+			imgupload_front_en => ["$sample_products_images_path/1.jpg", '1.jpg'],
+		},
+		ua => $ua,
+		expected_status => 403,
```
stephanegigandet, line 94: Do we get a 403 status?

MonalikaPatnaik: No we don't

lib/ProductOpener/APIProductWrite.pm
MonalikaPatnaik: Where can we display it to users at the product edit page, maybe we could add a message below the image when user tries to crop or unselect an image?

stephanegigandet (approved):
Looks good, thank you!

lib/ProductOpener/APIProductWrite.pm
stephanegigandet: Very good idea. On the product edit page, we should completely hide the part where you can select or upload a new image, and instead display a message like "An image has been sent by the manufacturer. If you think it is incorrect or not up-to-date, please contact us."
image

PR #8617, opened 2023-06-27
taxonomy: Japanese translations for ingredients processing

Naruyoko:
**IMPORTANT CHECKLIST**
Make sure you've done all the following (You can delete the checklist before submitting)
- [x] PR title is prefixed by one of the following: feat, fix, docs, style, refactor, test, build, ci, chore, revert, l10n, taxonomy
- [ ] Code is well documented
- [ ] Include unit tests for new functionality
- [ ] Code passes GitHub workflow checks in your branch
- [ ] If you have multiple commits please combine them into one commit by squashing them.
- [ ] Read and understood the [contribution guidelines](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/CONTRIBUTING.md)

### What

Added Japanese translations for some things in `ingredients_processing.txt`.

PR #8621, opened 2023-06-28
feat: activate Japanese ingredients processing

stephanegigandet:
This PR builds upon the ingredients processing Japanese translations by @Naruyoko in https://github.com/openfoodfacts/openfoodfacts-server/pull/8617

The PR
- enables ingredients processing parsing for Japanese: the processing is matched both before and after the ingredient, without requiring a space
- adds some tests


benbenben2 (approved):
Well done!

Good idea to have added the English translations in ingredients_text.

I will approve it now, but don't forget to apply your comment in Slack:
> Right, for that one, I think the right solution is to add クリーミングパウダー in the ingredients taxonomy, as "creaming" is not an ingredient.
> クリーミングパウダー is listed as an ingredient in several Japanese products we have.

stephanegigandet:
I'll make a separate PR for the creaming powder.

PR #8626, opened 2023-06-28
feat: Test for Image Upload API

### LINKED ISSUE #8619: Create an integration test for the image upload API
The test should use the https://openfoodfacts.github.io/openfoodfacts-server/api/ref-v2/#post-/cgi/product_image_upload.pl API to add an image to a product, and verify that the image has been uploaded.

We should also try different cases, like:
- uploading an image on a product that does not exist yet
- on a product that exists
- uploading an image that is too small
- uploading twice the same image

To create those tests, we can extend the APITest.pm execute_api_tests() function to add support for uploading files

e.g. in api_v2_product_write.t we have something like:

```
my $tests_ref = [
	{
		test_case => 'post-product',
		method => 'POST',
		path => '/cgi/product_jqm_multilingual.pl',
		form => {
			cc => "be",
			lc => "fr",
			code => "1234567890001",
			product_name => "Product name",
			categories => "Cookies",
			quantity => "250 g",
			serving_size => '20 g',
			ingredients_text_fr => "Farine de blé, eau, sel, sucre",
			labels => "Bio, Max Havelaar",
			nutriment_salt => '50.2',
			nutriment_salt_unit => 'mg',
			nutriment_sugars => '12.5',
		}
	},
```

for image upload, we could have form parameters like imgupload_front_en => ["/some/path/image.jpg", "image.jpg"]

This should be relatively easy as https://metacpan.org/pod/HTTP::Request::Common takes care of encoding files for POST uploads.
(see the part that starts with "The POST method also supports the multipart/form-data content used for Form-based File Upload as specified in RFC 1867. You trigger this content format by specifying a content type of 'form-data' as one of the request headers. If one of the values in the $form_ref is an array reference, then it is treated as a file part specification with the following interpretation:")

MonalikaPatnaik:
### What
Made changes in exexute_api_tests() subroutine in APITest.pm to add support for uploading files ( not sure tho if it is the right way) Created a test upload_images.t, not sure what to mention in image path.

- Fixes #8619


tests/integration/upload_images.t
```diff
@@ -0,0 +1,67 @@
...
+        path => '/cgi/product_image_upload.pl',
+        form => {
+            code => 'nonexistent_product_code',
+            imgupload_front_en => ['/path/to/image.jpg', 'image.jpg']
+        }
+    },
+    {
+        test_case => 'upload_image_existing_product',
+        method => 'POST',
+        path => '/cgi/product_image_upload.pl',
```
MonalikaPatnaik, line 39: Do i need to update this path to the API like this :')
path =>' https://openfoodfacts.github.io/openfoodfacts-server/api/ref-v2/#post-/cgi/product_image_upload.pl'

tests/integration/upload_images.t
```diff
@@ -0,0 +1,67 @@
...
+
+
+my $tests_ref = [
+    {
+        test_case => 'upload_image_nonexistent_product',
+        method => 'POST',
+        path => '/cgi/product_image_upload.pl',
+        form => {
+            code => 'nonexistent_product_code',
+            imgupload_front_en => ['/path/to/image.jpg', 'image.jpg']
```
stephanegigandet, line 33: Here we need to reference the path of an existing image.
We can reuse an image from https://github.com/openfoodfacts/openfoodfacts-server/tree/main/tests/integration/inputs/sample-products-images/300/000/000/0001 like 1.jpg

To find the path: my $sample_products_images_path = dirname(__FILE__) . "/inputs/sample-products-images/";

tests/integration/upload_images.t
stephanegigandet: path => '/cgi/product_image_upload.pl',  is the right path

tests/integration/upload_images.t
```diff
@@ -0,0 +1,67 @@
...
+    {
+        test_case => 'upload_image_existing_product',
+        method => 'POST',
+        path => '/cgi/product_image_upload.pl',
+        form => {
+            code => 'existing_product_code',
+            imgupload_front_en => ['/path/to/image.jpg', 'image.jpg']
+        }
+    },
+    {
```
stephanegigandet, line 45: here, add a GET request with the same product barcode as the first POST request, and ask for the "images" field, to verify that the image was uploaded.

lib/ProductOpener/APITest.pm
```diff
@@ -442,7 +442,15 @@ sub execute_api_tests ($file, $tests_ref, $ua = undef) {
 				);
 			}
 			elsif (defined $test_ref->{form}) {
-				$response = $test_ua->post($url, Content => $test_ref->{form}, %$headers_in);
```
stephanegigandet, line 445: For POST requests, we can add Content_Type => 'multipart/form-data' to the post() call, and keep the same call for both forms that contain files (indicated by passing an array reference as the value of a parameter, like imgupload_front_en => ["/path/file.jpg"] ), and forms that don't contain files

alexgarel:
I close this, as @stephanegigandet built upon this in #8658

Issue #5555, opened 2021-08-26
📸 Open Products Facts (Tracker)


teolemon:
### What
* https://wiki.openfoodfacts.org/Open_Products_Facts

### Taxonomies
* https://wiki.openfoodfacts.org/Open_Products_Facts_taxonomies

### Add Scores and value-added to Open Products Facts
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/10753
- [ ] #4710
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/10751
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/10752
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/2818
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/1320
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/10756
* https://wiki.openfoodfacts.org/Open_Products_Facts_use_cases
* https://wiki.openfoodfacts.org/Open_Products_Facts_scores

#### Bugs
- [ ] #2664
- [ ] #3942
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/6169
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/2517

##### Editing
- [ ] #2517

#### Bring Additional Fields / Folksonomy to Open Products Facts (to model any product)
- [ ] #1320
- [x] #5556

### Blend Open Food, Products, Beauty and Pet Food Facts together
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/797
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/497

### Mobile app
- https://github.com/openfoodfacts/smooth-app/issues/3923
- https://github.com/openfoodfacts/smooth-app/discussions/1378

### DevOps - Reduce the marginal cost of maintenance of Open Products Facts to 0
- [x] #1536
- [ ] Automate the release of the dev version of Open Products Facts
- [ ] #6408
- [x] Update Open Products Facts to the latest codebase from Open Food Facts
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/8319

### Part of
- #5529

Issue #5505, opened 2021-08-21
Open Beauty Facts (tracker)


teolemon:
### What
Help the world make more informed cosmetic choices
#### Make Open Food Facts and Open Beauty Facts work better together

- [ ] #401 
- [ ] #3372 

#### Make Open Beauty Facts work better
- [ ] #3962 
- [ ] #629

teolemon:
### Code pointers
You can find some [code pointers here](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/.github/labeler.yml). If some are missing, feel free to add them yourself in this file, or to ask them.

open beauty facts:
- taxonomies.obf/**/*
- po/openbeautyfacts/**/*
- scripts/gen_obf_inci_functions_taxonomy_from_cosing.pl
- scripts/gen_obf_ingredients_taxonomy_from_cosing.pl
- html/files/tagline-obf.json
- lib/ProductOpener/Config_obf.pm
- scripts/generate_polyquaternium.pl
- conf/nginx/sites-available/obf
- conf/apache-2.4/sites-available/obf.conf
- scripts/gen_feeds_obf.sh
- html/files/tagline-obf.json
- scripts/gen_feeds_daily_obf.sh
- lib/ProductOpener/Config_obf.pm
- taxonomies.obf/ingredients-old-obf.txt
- taxonomies.obf/ingredients-cosing-obf.txt
- taxonomies.obf/ingredients-cosing-obf.result.sto
- scripts/gen_obf_ingredients_taxonomy_from_cosing.pl
- scripts/gen_obf_inci_functions_taxonomy_from_cosing.pl
- html/images/svg/crowdsourcing-obf-round-smartphones.svg
- taxonomies.obf/brands.txt
- taxonomies.obf/labels.txt
- taxonomies.obf/safety.txt
- taxonomies.obf/allergens.txt
- taxonomies.obf/body_parts.txt
- taxonomies.obf/categories.txt
- taxonomies.obf/abbreviations.txt
- taxonomies.obf/labels.result.sto
- taxonomies.obf/labels.result.txt
- taxonomies.obf/inci_functions.txt
- taxonomies.obf/allergens.result.sto
- taxonomies.obf/allergens.result.txt
- taxonomies.obf/categories.result.sto
- taxonomies.obf/categories.result.txt
- tests/unit/expected_test_results/import_gs1/equadis_unilever_obf_signal.json
- taxonomies.obf/special_ingredients.txt
[...]

PR #8627, opened 2023-06-28
Opff reinstall fixes

alexgarel:
As I install obf (see #5505) and opf (see #5555), I see something I can do better.


alexgarel:
I closed, I will reopen (because I didn't start from opff-main)

Issue #5536, opened 2021-08-23
Open Pet Food Facts (Tracker)


teolemon:
### Who for
🐍 , 🐈 , 🐶 
### Bugs
- [ ] #1190 
- [ ] #5386
- [ ] #3942

### Taxonomies
- [ ] #3513

### Maintenability
- [ ] #1536

### Static content
- [ ] #1302 

### Work better with other OFF projects
- [ ] #1174 

### Ingredients
- [ ] #831
### Uncommitted assets

- [ ] #824
- [ ] #5345

### Nutrition
- [ ] #818
- [ ] #3550
### Part of
- #5529

teolemon:
### Static content for Open Pet Food Facts
- [ ] #1302 

### Ensure Open Pet Food Facts works better with other OxF projects
- [ ] #1174 

### Ingredients on Open Pet Food Facts
- [ ] #831

teolemon:
### Fixed
- [ ] #3942
- [x] #1536
- [ ] #3513

teolemon:
### Handle Pet Food Nutrition on Open Pet Food Facts
- [ ] #818
- [ ] #3550

PR #8628, opened 2023-06-28
fix: Open Pet Food Facts reinstall fixes

alexgarel:
### What
- as I re-install obf #5505 and opf #5555, I do something better.

### Part of
- #5536 
- #5512


stephanegigandet (approved):

html/data/data-fields.txt
teolemon, line 1: Why are we deleting this ? Elsewhere now ?

PR #8631, opened 2023-06-29
taxonomy: creaming powder, creaming agent

stephanegigandet:
adds creaming powder entry (used in Japanese ingredients: ja:クリーミングパウダー )


alexgarel (approved):
LGTM

PR #8632, opened 2023-06-29
fix: do not display empty packagings materials panel

stephanegigandet:
Small fix to avoid an empty panel if we have not reprocessed packagings yet.

image


alexgarel (approved):
Good.

PR #8657, opened 2023-07-05
docs: fix formating and broken link

alexgarel:
fix in quick start guide:
* table formatting 
* broken link


hangy (approved):

alexgarel:
Thanks for the validation @hangy :-) always nice to see you around !

alexgarel:
@teolemon FYI, this is no more mergable due to the change we introduced in #8650 … because we skip check that are required. This is discussed on [stackoverflow](https://stackoverflow.com/questions/66751567/return-passing-status-on-github-workflow-when-using-paths-ignore).
Github does not offers a nice alternative so for now, I will revert that :-(

PR #8661, opened 2023-07-05
chore: remove google analytics

alexgarel:
# What

Google analytics is not GDPR compatible (because of laws in the USA).
Remove it.

Note: On opf / obf / opff a matomo is in place but the code is in the respective branches: opf-main, obf-main and opff-main.


alexgarel:
@stephanegigandet FYI, it's already de-activated in production.

stephanegigandet (approved):

PR #8663, opened 2023-07-05
taxonomy: Taxonomy update (mostly Polish)

ArturLange:
### What

- updated some values in taxonomies based on most frequent unknown and missing values from website
- added some frequently used regexes to be ignored
- added some cases for ingredient enumeration in Polish - flours, meats, juices


lib/ProductOpener/Ingredients.pm
```diff
@@ -439,7 +439,7 @@ my %the = (
 my @labels = (
 	"en:fair-trade-organic", "en:organic", "en:fair-trade", "en:pgi", "en:pdo", "fr:label-rouge",
 	"en:sustainable-seafood-msc", "en:responsible-aquaculture-asc",
-	"fr:aoc", "en:vegan", "en:vegetarian"
+	"fr:aoc", "en:vegan", "en:vegetarian", "pl:bio",
```
stephanegigandet, line 442: It looks like we don't have a Polish translation for the en:organic label in the labels.txt taxonomy. (I'm guessing bio means organic in Polish?).  So instead of adding pl:bio here, it should be added to the labels.txt taxonomy instead.

stephanegigandet:
Awesome, thank you very much for all the changes @ArturLange ! The only suggestion I have is to move the pl:bio to labels.txt

lib/ProductOpener/Ingredients.pm
ArturLange: Yes, that's a translated "en:organic", it's added to `labels.txt` in this PR
My understanding was that I should put it here if I want it to be parsed from ingredients list, e.g. with ingredients like "papaja bio". Is that wrong?

stephanegigandet: Here we need only the canonical name of the label you want to be parsed from the ingredients list (en:organic), and all synonyms in all languages will be used.

ArturLange: Awesome, I didn't know that. 
Can I add this as a comment here to clarify it for the future? :slightly_smiling_face:

alexgarel:
Done in #8771

PR #8674, opened 2023-07-10
test: adding a synonym test for the suggestion feature

albatrousse:
adding a synonym test for the suggestion feature, in my case i used the product "Dairy Drink" in French. The input "jus de fruits au lait" should give "Boissons au fruit et au lait" as a result which is the first word in the taxonomy matching "Dairy Drink" in French.


alexgarel (approved):
Great :tada:

PR #8684, opened 2023-07-11
taxonomy: some fixes to ingredients taxonomy

stephanegigandet:
Solved some of the errors reported by the taxonomy build of ingredients.txt


alexgarel (approved):
Cool, thanks !

How far away are we from being able to enforce the check ?

PR #8685, opened 2023-07-11
feat: updated test for serving_size for nutrition table

### LINKED ISSUE #3029: Quality facet - Nutrition - list products where serving size can't be parsed
### What
Serving size has a specific goal: to let Open Food Facts app make a proportional calculation of each nutrient per serving size -- see: https://en.wiki.openfoodfacts.org/Data_fields

But dozens of "serving_size" values can't be parsed. Examples: `30 gr`, `2 liters`, `60`, `9 candies and 2 biscuits` and son on.

We should have a new quality facet to list each product where serving_size (and quantity?) has a value that can't be parsed.

### Part of 
- #5538

benbenben2:
- Fixes #3029 

**Reminder:** serving_size is the input from the contributor, serving_quantity is the extracted quantity from the serving_size

1) Nutrition data per serving - Serving quantity is unknown
2) Nutrition data per serving - Serving quantity is not recognized
3) Nutrition data per serving - Serving quantity is 0 --> only when we actually get "0 g" or similar

1) would be exact same as: en:nutrition-data-per-serving-missing-serving-size
previously it was a warning
updated to be an error

**Remark about 2 & 3)**  
>    input "une crepe 25 grammes" -> output serving_quantity=25 (special case, supposedly due to unit_equivalences_list in Units.pm)
>     input "100 milliliter" -> output serving_quantity not in result
>     input "3 pcs" -> output serving_quantity not in result
>     2 comprimés -> missing
>     1 container -> missing 
>     serving -> 0
>     input "1/4 cup" -> output serving_quantity=0

**Reason** is that normalize_serving_size subroutine in Units.pm returns 0 if it cannot extract the serving otherwise it returns the result of the subroutine unit_to_g, and this latter return either something or nothing (i.e., not 0)

2) (new) en:nutrition-data-per-serving-serving-quantity-is-not-recognized
This covers both cases:
2)a) Serving quantity is not recognized - but regex did work) and 
2)b) Serving quantity is not recognized - and regex to extract "quantity" and "unit" does not work
This should never be 0

3) en:nutrition-data-per-serving-serving-quantity-is-0
For "0 g" or "0 mL"

4) (from #8057) en:Nutrition data per serving - Serving size is missing digits

tested locally:
input - result
10 ml - ok
empty - 1
serving - 2 and 4
0g - 3 
0 mL - 3

### DIFF
--- a/tests/unit/dataqualityfood.t
+++ b/tests/unit/dataqualityfood.t
@@ -521,6 +521,92 @@ check_quality_and_test_product_has_quality_tag(
 	'serving size should contains digits', 0
 );
 
+# serving size is missing
+$product_ref = {
+	nutrition_data => "on",
+	nutrition_data_per => "serving"
+};
+check_quality_and_test_product_has_quality_tag(
+	$product_ref,
+	'en:nutrition-data-per-serving-missing-serving-size',
+	'serving size should be provided if "per serving" is selected', 1
+);
+check_quality_and_test_product_has_quality_tag(
+	$product_ref,
+	'en:nutrition-data-per-serving-serving-quantity-is-0',
+	'serving size equal to 0 is unexpected', 0
+);
+check_quality_and_test_product_has_quality_tag(
+	$product_ref,
+	'en:nutrition-data-per-serving-serving-quantity-is-not-recognized',
+	'serving size cannot be parsed', 0
+);
+# serving size equal to 0
+$product_ref = {
+	nutrition_data => "on",
+	nutrition_data_per => "serving",
+	serving_quantity => "0",
+	serving_size => "0g"
+};
+check_quality_and_test_product_has_quality_tag(
+	$product_ref,
+	'en:nutrition-data-per-serving-missing-serving-size',
+	'serving size should be provided if "per serving" is selected', 0
+);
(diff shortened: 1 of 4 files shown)


tests/unit/dataqualityfood.t
```diff
@@ -521,6 +521,92 @@ check_quality_and_test_product_has_quality_tag(
...
+check_quality_and_test_product_has_quality_tag(
+	$product_ref,
+	'en:nutrition-data-per-serving-serving-quantity-is-0',
+	'serving size equal to 0 is unexpected', 0
+);
+check_quality_and_test_product_has_quality_tag(
+	$product_ref,
+	'en:nutrition-data-per-serving-serving-quantity-is-not-recognized',
+	'serving size cannot be parsed', 1
+);
```
alexgarel, line 586: When I see we reapeat the test 3 times, shan't we change check_quality_and_test_product_has_quality_tag to have a list of tags to check for presence, and a list of tags to check for absence ?

Of course in another PR.

alexgarel (approved):
Thanks a lot @benbenben2

CharlesNepote:
Very good! Thanks a lot @benbenben2 this is far more clear and useful as before.

PR #8693, opened 2023-07-13
fix: fix  madenearme by using jsonl instead of mongodb

alexgarel:
madenearme is stalled because of MongoDB issues.

Use jsonl file instead of MongoDB to generate them.

Also added an integration test.


lib/ProductOpener/Display.pm
```diff
@@ -6573,70 +6578,35 @@ sub get_packager_code_coordinates ($emb_code) {
 
 }
 
-sub search_and_map_products ($request_ref, $query_ref, $graph_ref) {
+# an iterator over a cursor to unify cases between mongodb and external data (like filtered jsonl)
```
stephanegigandet, line 6581: Nice

stephanegigandet (approved):
Perfect, thank you, and thanks for all the refactoring and doc!

Issue #8005, opened 2023-01-17
Reviewed email template for Open Food Facts subscribers 


galnaf:
Story: Purpose: when a person signs up to Open Food Facts for the first time with their email, they receive an email with their newly created account login details. It needs an update. 

Here's the proposition: 

_(horizontal logo - centered - medium sized)_
![EXE_LOGO_OFF_RVB_Plan de travail 1 copie 0](https://user-images.githubusercontent.com/105732921/212959120-82c8e6ae-c5c5-4d70-829f-688ff0d8d6bd.png)

### **Hi (name) and welcome to [Open Food Facts](https://openfoodfacts.org/) !** 🔎🍊

Thank you for joining our international community of people eager to see more transparency in the food industry by contributing to the world’s largest open food database! 

**Here is your user name**: xxxxxx

Sign in on the [website](https://openfoodfacts.org/) to add and edit products. 

📲We also have a free, ad-free & private [mobile scanning app](https://world.openfoodfacts.org/open-food-facts-mobile-app) you may like.

(_this next part would be inside a frame/box_see screenshots attached below)
### **Open Food Facts is a collaborative project to which you can bring much more than new products: your energy, enthusiasm & ideas!**

🌎 To discuss projects and connect as a community, join us on [our Slack](https://slack.openfoodfacts.org/).

🙋🙋‍♀️To become a volunteer for the project, sign up  to the [Contributor Skill Pool ](https://docs.google.com/forms/d/e/1FAIpQLSfGHAn5KxW7ko3_GlDfQpVGKpPAMHMbDvY2IjtxfJSXxKJQ2A/viewform)(4 min form)! 

💫 To follow our latest news: 
Instagram ([English](https://www.instagram.com/open.food.facts/), [French](https://www.instagram.com/openfoodfacts.fr/))
Facebook ([English](https://www.facebook.com/OpenFoodFacts), [French](https://www.facebook.com/OpenFoodFacts.fr/))
Twitter ([English](https://twitter.com/OpenFoodFacts), [French](https://twitter.com/OpenFoodFactsFr))
[LinkedIn](https://www.linkedin.com/company/open-food-facts)
[Our Blog ](https://blog.openfoodfacts.org/en/)(available in several languages)
[Youtube](https://www.youtube.com/@openfoodfacts1170) 
[Subscribe to our newsletter](https://blog.openfoodfacts.org/en/newsletter-2) (no more than one newsletter a month!)

[...]

PR #8703, opened 2023-07-18
fix: Update welcome email

raphael0202:
Use new welcome email (when user register). We switched from a plain text template to an HTML one, which required some codebase modification. The email template is stored in openfoodfacts-web.

I tested the changes locally with the content added by [this PR](https://github.com/openfoodfacts/openfoodfacts-web/pull/422).

Solves #8005


lang-default/en/emails/user_welcome.html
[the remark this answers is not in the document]
alexgarel: @hangy we ask ourselves the question, but it seems better to have a whole html file to translate than separate sentences. That's why we go for this.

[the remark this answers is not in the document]
alexgarel: Given that there were really few variables.

lib/ProductOpener/Mail.pm
```diff
@@ -160,6 +162,75 @@ sub send_email ($user_ref, $subject, $text) {
...
+
+
+=head3 Arguments
+
+=head4 FILENAME
+
+The HTML file name (ex: user_welcome.html)
+
+=head4 LANG
+
```
alexgarel, line 211: You must keep constistent with real definition

```suggestion
=head2 get_html_email_content ($filename, $lang )

Fetch the HTML email content in $DATA_ROOT/lang/emails. If a translation is available
for the requested language, we provide the translated version otherwise English is the default.


=head3 Arguments

=head4 $filename

The HTML file name (ex: user_welcome.html)

=head4 $lang

```

lib/ProductOpener/Mail.pm
```diff
@@ -160,6 +162,75 @@ sub send_email ($user_ref, $subject, $text) {
...
+The 2-letter language code
+
+=head3 Return Values
+
+The HTML string
```
alexgarel, line 216:
```suggestion
=head3 Return Values

The HTML string or undef if the file does not exists or is not readable.
```

lib/ProductOpener/Mail.pm
```diff
@@ -160,6 +162,75 @@ sub send_email ($user_ref, $subject, $text) {
...
+
+sub get_html_email_content ($filename, $lang) {
+	# if an email does not exist in the local language, use the English version
+	my $file = "$data_root/lang/$lang/emails/$filename";
+
+	if (!-e $file) {
+		$file = "$data_root/lang/en/emails/$filename";
+	}
+
+	open(my $IN, "<:encoding(UTF-8)", $file);
```
alexgarel, line 228: Better ensure is does not break badly
```suggestion
	open(my $IN, "<:encoding(UTF-8)", $file) or log->error("Can't open $file for reading");
	return unless $IN;
```

alexgarel (approved):
LGTM but some small suggestions. I let you apply them and merge.

Thanks a lot for this PR !

lib/ProductOpener/Mail.pm
raphael0202: Alright, I took example from another docstring :sweat_smile:

Issue #8707, opened 2023-07-19
Attach frontend dist assets in a tar gziped file to releases 


alexgarel:
We want to avoid recompiling frontend assets when we deploy in production.
Those assets are available in the docker image for frontend.
Put them in a tar.gz file and add them to the release using the CI.

PR #8708, opened 2023-07-19
Feat add dist to releases

alexgarel:
fixes: #8707


alexgarel:
I did a test (see the second to last commit) and it seems to works like a charm… (see file on [v2.14.0 release](https://github.com/openfoodfacts/openfoodfacts-server/releases/tag/v2.14.0)) but we might need to wait for a real release to be sure.

raphael0202 (approved):
LGTM

PR #8712, opened 2023-07-20
test: [WIP] testing the user deletion feature and completing create_user.t …

albatrousse:
completing create_user.t by looking for keywords, which the ua entered when filling the sign up form, in the edit preferences page.
 
adding a test for the user deletion feature in which :
- a common user is created
- the user adds a product then deletes his account
- an admin user is created and checks if the user account is well deleted (checking the edit preferences page and contributor page)
- the admin also checks if the product added by the user is anonymized (the product should display "anonymous" as creator)
- checking that no one can log in with the user logs

### DIFF
--- a/tests/integration/delete_user.t
+++ b/tests/integration/delete_user.t
@@ -37,10 +37,32 @@ my %product_fields = (
 	nutriment_salt_unit => "g",
 );
 edit_product($ua, \%product_fields);
-delete_user($ua);
+
+my @words = ("Delete the user", "User is being deleted. This may take a few minutes.", "Invalid user.", "Unknown user.");
+my $url_userid = construct_test_url("/cgi/user.pl?type=edit&userid=tests", "world");
+my $url_delete = construct_test_url("/cgi/user.pl", "world");
+my $response_edit = $ua->get($url_userid);
+
+my %delete_form = (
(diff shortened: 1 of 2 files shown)


tests/integration/delete_user.t
albatrousse, line 1: I'm trying to fix something because the first two like are successful but the last three not. And I'm quite confused, because it kind of means that the account is deleted but when the admin still can check the user account. Does someone know what the problem could be ?

albatrousse: I checked and the problem is that the user account was not deleted, in fact I can still log in with his ids. Does anyone have a clue on how to delete the user account ?

john-gom: The delete user process creates a minion job to do the work, so the delete won't happen immediately. Are you running minion in your test environment and allowing time for the job to run?

albatrousse: Alex explained me the same thing, I have to assign a minion job to do the work and then wait until the job is done in order to make sure the user account is deleted. I'm working on it, thank you for your help !

albatrousse:
I will create two different PR : one for completing the create_user.t test and another one for the user deletion test

PR #8713, opened 2023-07-20
feat: Added Hide/Show toggle eye icon beside password field 

MonalikaPatnaik:
#### in user sign-in form
image

#### in user-registration-form
image

- Fixes #[ISSUE NUMBER]


templates/web/pages/session/sign_in_form.tt.html
```diff
@@ -2,6 +2,38 @@
 
 [% INCLUDE 'web/common/includes/error_list.tt.html' %]
 
+<!-- Add Font Awesome CSS -->
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.15.3/css/all.min.css">
```
alexgarel, line 6: We want to avoid using CDN as they track people, better add it to the assets (in html/fonts). I think you can add the link to load the font in the base template.

Also we already have material font (`material-icons-v139`) [which provides "Visibility" and "Visibility off" symbols](https://fonts.google.com/icons?icon.query=visibility), it would be better to stick to those.

Eg:
```html
<span class="material-icons material-symbols-visibility">
```

templates/web/pages/session/sign_in_form.tt.html
```diff
@@ -2,6 +2,38 @@
...
+        cursor: pointer;
+        position: absolute;
+        right: 10px;
+        top: 50%;
+        transform: translateY(-50%);
+    }
+    .password-container {
+        position: relative;
+    }
+</style>
```
alexgarel, line 35: It's better to add it to main _off.scss file.

templates/web/pages/session/sign_in_form.tt.html
```diff
@@ -2,6 +2,38 @@
...
+        if (passwordInput.type === "password") {
+            passwordInput.type = "text";
+            toggleIcon.classList.replace("fa-eye-slash", "fa-eye");
+           
+        } else {
+            passwordInput.type = "password";
+            toggleIcon.classList.replace("fa-eye", "fa-eye-slash");
+        }
+    }
+    </script>
```
alexgarel, line 22: as you include it to both form I would put it in a specific js file.

alexgarel (changes requested):
Thanks for the good idea, but there are some details to take care of !

stephanegigandet (approved):
Looks good, thank you!

scss/_off.scss
```diff
@@ -1399,6 +1399,17 @@ $section-cyan: #008c8c;
...
+.toggle-password {
+        cursor: pointer;
+        position: absolute;
+        right: 10px;
+        top: 60%;
+        transform: translateY(-50%);
+    }
+    .password-container {
+        position: relative;
+    }
```
alexgarel, line 1412: Good ! 

Can you please fix indentation ?

alexgarel (approved):
Very good @MonalikaPatnaik !

Only an indentation fix and you can merge :-)

scss/_off.scss
alexgarel: I did a commit for that :-)

MonalikaPatnaik:
Thankyou so much..I think we could merge <3

stephanegigandet (approved):
Thank you!

PR #8717, opened 2023-07-22
taxonomy: German ingredient

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -46141,7 +46141,7 @@ pt:chicória
 
 <en:chicory
 en:chicory root
-de:Chikoreewurzel
+de:Chikoreewurzel, Zichorienwurzel
 fr:racine de chicorée
 hr:cikorija korijen, korijen cikorije
 it:radice di cicoria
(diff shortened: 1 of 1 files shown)


taxonomies/ingredients.txt
```diff
@@ -86318,7 +86318,7 @@ wikipedia:en:https://en.wikipedia.org/wiki/Xylose
 # Tonka beans are prohibited for use in food in the US: https://www.govinfo.gov/content/pkg/CFR-2006-title21-vol3/xml/CFR-2006-title21-vol3-sec189-130.xml
 en:tonka bean
 es:haba tonka
-de:Tonkabohn, Tonkabohnen
+de:Tonkabohn, Tonkabohnen, Tonkabohne
```
hangy, line 86321: I think "Tonkabohn" is probably a typo. If you search it on the internet, it only results in a handful of entries, which seems to confirm this. Let's just remove that alias, IMHO.
```suggestion
de:Tonkabohne, Tonkabohnen
```

moon-rabbitOFF: You are right, I modified it, thanks

taxonomies/ingredients.txt
```diff
@@ -13698,6 +13698,11 @@ sl:modificiran koruzni škrob
 sr:modifikovani kukuruzni skrob
 sv:modifierad majsstärkelse
 
+<en:modified starch
```
stephanegigandet, line 13701: I think we could add "modified" and "partially modified" in ingredients_processing.txt, that way we could (eventually) remove all the related entries in ingredients.txt.

But let's merge this PR first, we can do it later.

stephanegigandet (approved):
Thank you!

PR #8719, opened 2023-07-23
taxonomy: added Croatian ingredients in taxonomies

benbenben2:
### What

added Croatian ingredients into the taxonomies.

(remark: big block "en:chokeberry" has been deleted because it was duplicated in the taxonomy)


taxonomies/ingredients.txt
```diff
@@ -3191,6 +3218,14 @@ de:pasteurisierte entrahmte Kuhmilch
 es:leche de vaca desnatada pasterizada
 th:น้ำนมโคขาดมันเนย
 
+<en:cow's milk
+en:homogenized milk with 2 8 milk fats
```
stephanegigandet, line 3222: should be 2.8% right? (same for entries below)

stephanegigandet (approved):

PR #8723, opened 2023-07-24
test: adding an integration test for the account deletion feature

albatrousse:
adding a test for the user deletion feature in which :

- a common user is created
- the user adds a product then deletes his account
- an admin user is created and checks if the user account is well deleted (checking the edit preferences page and contributor page)
- the admin also checks if the product added by the user is anonymized (the product should display "anonymous" as creator)
- checking that no one can log in with the user logs

currently working on assigning the deletion account part to a minion job and waiting it to be done before having an admin to check whether the account is well deleted or not


tests/integration/delete_user.t
albatrousse, line 1: I can't really understand why locally all my tests passed but Git tests tell me that one of the sevens tests failed which is weird considering that all it does is a `make tests`

albatrousse: Never mind, I just had to rerun the tests (which is questionnable)

albatrousse, line 1: I added a sequence where the user agent sign out after its account is deleted. This was done in order to avoid a troublesome interaction where the admin user would get his own edit page instead of the now deleted user edit page

tests/integration/delete_user.t
```diff
@@ -0,0 +1,123 @@
...
+			$waited++;
+		}
+	}
+}
+#checking if there is still delete_user_task jobs
+$jobs = $minion->jobs({tasks => ["delete_user_task"]});
+my $jobs_count = 0;
+while (my $job = $jobs->next) {
+	$jobs_count++;
+}
```
alexgarel, line 79: I would like you to put the logic of returning jobs corresponding to a task and started after a given ts in TestAPI.pm (because we might reuse it).

so this code would read:
```perl
my $jobs_ref = get_minion_jobs("delete_user_task", $before_delete_ts, 400);
is(scalar @{$jobs_ref}, 1, "We have one job launched to remove user"); 
is($jobs_ref->[0]->state, "finished", "And job is completed");
```
get_minion_jobs would return all job of type delete_user_task (first argument, eg $job_name)  started after $before_delete_ts (second argument, eg $after_ts) waiting for each job at most 400 seconds ($max_wait).

This way it will test that the state is "finished" and not "failed".
Because otherwise it's not easy to understand why following tests will fail, while here we got a good indicator.

tests/integration/delete_user.t
```diff
@@ -0,0 +1,123 @@
...
+create_user($ua, \%create_client_args);
+
+#new admin user agent
+my $admin = new_client();
+create_user($admin, \%admin_user_form);
+
+#common ua add a new product then delete the account while being still logged in
+edit_product($ua, \%default_product);
+
+my @words = (
```
alexgarel, line 27: It's interesting pattern (yeah I can see how it was interesting to copy paste the same line all over) but it hinders readability of your test, and it should be favored.

So I would replace @words[1] etc... by the string instead.

tests/integration/delete_user.t
```diff
@@ -0,0 +1,123 @@
...
+my $response_userid = $admin->get($url_userid);
+
+my $url_email = construct_test_url('/cgi/user.pl?type=edit&userid=bob@test.com', "world");
+my $response_email = $admin->get($url_email);
+
+my $url_contributor = construct_test_url("/contributor/tests", "world");
+my $response_contributor = $admin->get($url_contributor);
+
+#checking if the edit page of the common ua is well deleted
+like($response_userid->content, qr/\Q$words[3]\E/i, "the userid edit page is well deleted");
```
alexgarel, line 102: why not do the checks immediately after accessing the url ? Here again it would be easier to read.

tests/integration/delete_user.t
```diff
@@ -0,0 +1,123 @@
...
+	password => "testtest",
+	submit => "Sign in"
+);
+my $response_login = $ua->post($url_login, \%login_form);
+like($response_login->content, qr/\Q$words[5]\E/i, "an user can't login with the deleted account ids");
+
+#checking if the added product has been anonymized
+my $url_product = construct_test_url("/cgi/product.pl?type=edit&code=2000000000001", "world");
+my $response_product = $admin->get($url_product);
+like($response_product->content, qr/\Q$words[6]\E/i, "the product has been anonymized");
```
alexgarel, line 121: You would rather go to "/api/v2/product/2000000000001", decode the json response you get and check the creator and editors_tags fields.

alexgarel (changes requested):
Really good work, now let's improve a bit :-)

albatrousse:
Thank you for your review @alexgarel, I'll work on the changes that have to be made

tests/integration/delete_user.t
albatrousse: Okay, yes it's more readable, I'll change this

tests/integration/delete_user.t
albatrousse: I just wrote it in the moment, I guess I should have reviewed a little bit this part ! I changed it

tests/integration/delete_user.t
albatrousse: Is it a better way ? I tried it and it works well

tests/integration/delete_user.t
albatrousse: I'm currently working on it, but I have a "big" trouble : when importing ProductOpener::Producers into APITest.pm, whatever test I run it encounters "Error 137", I don't know what to do (I think it may be a compilation error but why ?)

albatrousse:
I'm going to push my code with the troublesome issue I've been dealing with for the past few days. 
If you write `use ProductOpener::Producers qw/:all/` in APITest.pm in order to import the module that contains the variable $minion, you will end up getting Error 137 whatever integration test you run.
Does someone have an idea on why ?

stephanegigandet:
The Apache + mod_perl process is crashing a few seconds after startup.

docker ps -a

f630a7c7d5d1   openfoodfacts-server/backend:dev                                                      "/docker-entrypoint.…"   6 weeks ago     Exited (139) 34 seconds ago                                po-backend-1

docker logs f630a7c7d5d1
...
Action '-D FOREGROUND' failed.
The Apache error log may have more information.
Action '-D FOREGROUND' failed.
The Apache error log may have more information.
Segmentation fault (core dumped)
Segmentation fault (core dumped)
Action '-D FOREGROUND' failed.
The Apache error log may have more information.
Action '-D FOREGROUND' failed.
The Apache error log may have more information.
Segmentation fault (core dumped)

albatrousse:
> The Apache + mod_perl process is crashing a few seconds after startup.
> 
> docker ps -a
> 
> f630a7c7d5d1 openfoodfacts-server/backend:dev "/docker-entrypoint.…" 6 weeks ago Exited (139) 34 seconds ago po-backend-1
> 
> docker logs f630a7c7d5d1 ... Action '-D FOREGROUND' failed. The Apache error log may have more information. Action '-D FOREGROUND' failed. The Apache error log may have more information. Segmentation fault (core dumped) Segmentation fault (core dumped) Action '-D FOREGROUND' failed. The Apache error log may have more information. Action '-D FOREGROUND' failed. The Apache error log may have more information. Segmentation fault (core dumped)

Thank you very much @stephanegigandet !!

lib/ProductOpener/APITest.pm
```diff
@@ -768,4 +770,44 @@ sub fake_http_server ($port, $dump_path, $responses_ref) {
...
+	while (my $job = $jobs->next) {
+		#only those who were created after the timestamp
+		if ($job->created > $created_after_ts) {
+			#waiting the job to be done
+			while ($job->state eq "inactive" or $job->state eq "active" or $waited < $max_waiting_time) {
+				sleep(2);
+				$waited++;
+			}
+			#checking if it did not fail
+			is($job->state, "finished");
```
alexgarel, line 806: You want to have this one outside of this method, in your test.

1. this is easier to read (in your test)
2. at some point you may want to test that a job fail indeed
```suggestion
```

lib/ProductOpener/APITest.pm
```diff
@@ -768,4 +770,44 @@ sub fake_http_server ($port, $dump_path, $responses_ref) {
...
+		}
+	}
+	$jobs = get_minion()->jobs({tasks => [$task_name]});
+	return $jobs;
```
alexgarel, line 810: You want to only return jobs after created, so instead in your for loop you have to build the list of such jobs.

```suggestion
    my @run_jobs = ();
	#iterate on job
	while (my $job = $jobs->next) {
		#only those who were created after the timestamp
		if ($job->created > $created_after_ts) {
			#waiting the job to be done
			while ($job->state eq "inactive" or $job->state eq "active" or $waited < $max_waiting_time) {
				sleep(2);
				$waited++;
			}
			push @run_jobs, $job;
		}
	}
	return \@run_jobs;
```

tests/integration/delete_user.t
```diff
@@ -0,0 +1,114 @@
...
+edit_product($ua, \%default_product);
+
+my @words = (
+	"User is being deleted. This may take a few minutes.",
+	"See you soon!",
+	"Invalid user.",
+	"Unknown user.",
+	"Incorrect user name or password.",
+	"/editor/anonymous"
+);
```
alexgarel, line 34: We can remove it now I think.
```suggestion
```

tests/integration/delete_user.t
```diff
@@ -0,0 +1,114 @@
...
+#waiting the deletion task to be done
+my $max_time = 400;
+my $jobs_ref = get_minion_jobs("delete_user_task", $before_delete_ts, $max_time);
+
+#checking if there are remaining jobs
+my $count = 0;
+while (my $job = $jobs_ref->next) {
+	$count++;
+}
+is($count, 0, "waiting job is completed");
```
alexgarel, line 68: This is a strange pattern and I don't understand how it works…

Taking into account my suggestion above, I would instead check that there was only one job, and it's finished.

```suggestion
my $jobs_ref = get_minion_jobs("delete_user_task", $before_delete_ts, $max_time);

is(scalar @{$jobs_ref}, 1, "One delete_user_task was triggered");
is(@{$jobs_ref}[0]->state, "finished", "delete_user_task finished without errors");
```

alexgarel (commented):
Awesome, I'm just pointing some minor refactoring still to do.

tests/integration/delete_user.t
albatrousse: Oups, after all these commits I even forgot to remove this

lib/ProductOpener/APITest.pm
albatrousse: Oh I see, it's a clever way to inspect all the job the minion run

tests/integration/delete_user.t
albatrousse: I did it this way so that I could check if there was any remaining task to be done because I could get the "length" of the jobs because it was not a list. But if get_minion_jobs returns a list of job doing it this way is cleaner 😮

lib/ProductOpener/APITest.pm
albatrousse: Oooh, okay I see thank you !

lib/ProductOpener/APITest.pm
```diff
@@ -768,4 +771,60 @@ sub fake_http_server ($port, $dump_path, $responses_ref) {
...
+				my $job_id = $job->{id};
+				# retrieving the job state
+				my $job_state = $job->{state};
+				# waiting the job to be done
+				while (($job_state eq "active") or ($job_state eq "inactive")) {
+					sleep(2);
+					$waited += 2;
+					# reload to get updated state
+					$job = get_minion()->job($job_id);
+					$job_state = $job->info->{state};
```
alexgarel, line 814: Ok we may have a problem because $job when returned by `jobs` method returns a hash, while 'job' method return an object.

It's dangerous because, here it work for your test, but it might break in the future if you have multiple jobs for exemple (more chance the job is not waited for).

So I propose that we systematically get the object through job method. cf proposal.

lib/ProductOpener/APITest.pm
```diff
@@ -768,4 +771,60 @@ sub fake_http_server ($port, $dump_path, $responses_ref) {
...
+		my $jobs = get_minion()->jobs({tasks => [$task_name]});
+		# iterate on job
+		while (my $job = $jobs->next) {
+			next if (defined $run_jobs{$job->{id}});
+			# only those who were created after the timestamp
+			if ($job->{created} > $created_after_ts) {
+				# retrieving the job id
+				my $job_id = $job->{id};
+				# retrieving the job state
+				my $job_state = $job->{state};
```
alexgarel, line 807:
```suggestion
				# retrieving the job object (to be consistent with below)
				$job = get_minion()->job($job_id);
				# job id
				my $job_id = $job->id;
				# retrieving the job state
				my $job_state = $job->state;
```

lib/ProductOpener/APITest.pm
alexgarel, line 814: We have an object. Use accessor:

```suggestion
					$job_state = $job->state;
```

lib/ProductOpener/APITest.pm
```diff
@@ -768,4 +771,60 @@ sub fake_http_server ($port, $dump_path, $responses_ref) {
...
+			}
+		}
+		if (!(scalar %run_jobs)) {
+			# try to wait for jobs
+			sleep(2);
+			$waited += 2;
+		}
+	}
+	# sort by creation date to have jobs in predictable order
+	my @all_jobs = sort {$_->{created}} (values %run_jobs);
```
alexgarel, line 826: If we use objects
```suggestion
	my @all_jobs = sort {$_->created} (values %run_jobs);
```

alexgarel (approved):
I made a small commit just to ensure consistency in get_minion_jobs.

Thank you @albatrousse !

lib/ProductOpener/APITest.pm
alexgarel: wrong, your code was correct.

PR #8728, opened 2023-07-24
fix: use states_tags index for remove_empty_products.pl

stephanegigandet:
the script was timing out because it used the "empty" field for which we don't have an index
also removed useless cruft


alexgarel (approved):
LGTM

PR #8744, opened 2023-07-27
fix: return empty noindex webpage when crawlers hit specific pages

raphael0202:
An analysis of nginx logs made us realize that 6% of our traffic was due to Bing bot, half of the queries were "facet" queries that involve aggregate MongoDB queries and consume a lot of resources.
See https://openfoodfacts.slack.com/archives/C1FPYCWM7/p1690454042958259 discussion for more context.

As a result, we decided to prevent known crawlers from crawling nested facet pages (2 facets). This should limit drastically the number of crawlable pages and reduce server load (especially the DB server).

I checked locally with custom User Agent, nested facet pages are blocked as expected.

### DIFF
--- a/cgi/display.pl
+++ b/cgi/display.pl
@@ -110,6 +110,13 @@ if (defined $request_ref->{error_message}) {
 	return Apache2::Const::OK;
 }
 
+if ($request_ref->{no_index} eq 1) {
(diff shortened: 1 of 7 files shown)


alexgarel (approved):
Perfect but… could you add a test ?

PR #8752, opened 2023-07-29
taxonomy: added Japanese ingredients in taxonomies

benbenben2:
### What

added Japanese ingredients in the taxonomies.


stephanegigandet (approved):
Arigatou gozaimasu

PR #8753, opened 2023-07-29
taxonomy: added ingredients in taxonomies

benbenben2:
### What

added ingredients in the taxonomies.


stephanegigandet (approved):
Thanks!

PR #8763, opened 2023-07-31
feat: Allowing Org Admins to Add/Invite Users

MonalikaPatnaik:
### What

Allowing Org admins to add/invite users to their organization..
for users that does not have off account, as of now we tell the org admin that we don’t have an user matching some of the emails

### Screenshot

image

image

- Fixes #[ISSUE NUMBER]


lib/ProductOpener/Users.pm
```diff
@@ -949,6 +969,27 @@ sub remove_user_by_org_admin ($orgid, $user_id) {
...
+	foreach my $email (@emails) {
+		# Check if the email is associated with an OpenFoodFacts account
+		my $user_id = is_email_has_off_account($email);
+		if (defined $user_id) {
+			# Add the user to the organization
+			add_user_to_org($org_id, $user_id, ["members"]);
+		}
+		else {
+			# we can send an invitation email to the user
+			return 0;
```
stephanegigandet, line 986: If you use "return", the function will exit immediately, and the other e-mail addresses after that one won't be processed.

lib/ProductOpener/Users.pm
```diff
@@ -949,6 +969,27 @@ sub remove_user_by_org_admin ($orgid, $user_id) {
 	return;
 }
 
+sub add_users_to_org_by_admin ($org_id, $email_list) {
```
stephanegigandet, line 972: Maybe the function could return an object with 2 lists, 1 for e-mails that have been added, and another for e-mails that have been invited.

stephanegigandet (commented):
Looks good to me :)

lib/ProductOpener/Users.pm
MonalikaPatnaik: Yes, right as of now if one of the mails in the list is not having off acc, then other mails after that won't be processed.

cgi/org.pl
```diff
@@ -358,18 +358,41 @@
...
 		}
 
 	}
-	$template_data_ref->{result} = lang("edit_org_result");
+	elsif ($type eq 'add_users') {
+		if (is_user_in_org_group($org_ref, $User_id, "admins")) {
+			my $email_list = remove_tags_and_quote(single_param('email_list'));
+			my $email_ref = add_users_to_org_by_admin($orgid, $email_list);
+
+			my @messages;
```
stephanegigandet, line 379: Looks good. But all the logic to create the messages with the lists of users invited and added should be moved to the HTML template file, so that we don't have HTML tags in the Perl code, and that designers can easily change the format of the messages (for instance using a <ul> list to list the users invited / added).

templates/web/pages/org_form/org_form.tt.html
```diff
@@ -54,6 +54,15 @@ <h2>[% lang("organization_members") %]</h2>
 	</div>
 	[% END %]
 
+		<form method="post" action="/cgi/org.pl" enctype="multipart/form-data" style="margin-bottom: 20px;">
+		<p>Enter email addresses of users to invite (comma-separated):</p>
+		<textarea id="email_list" name="email_list" style="height:100px;width:50vw"></textarea>
+		<input type="hidden" name="action" value="process" />
+		<input type="hidden" name="type" value="add_users" />
+		<input type="hidden" name="orgid" value="[% orgid %]" />
+		<input type="submit" name=".submit" class="button" value="Invite Users" />
```
stephanegigandet, line 63: "Invite users" should be added to the .po files

stephanegigandet (approved):
Looks good, thank you!

alexgarel:
@manon-corneille we don't support multi-org profiles, and I think it's a bit dangerous to invite someone already in another org (it will leave an org with 0 users).

manon-corneille:
@alexgarel ok thanks, I understand your point.  
As for my first point, let's see how it goes! :-) Maybe it's not so painful from a user's perspective

PR #8766, opened 2023-07-31
test: adding a test that check if preferences edits made by an user are saved

albatrousse:
quick test to check if preferences are saved when an user made some edits


alexgarel (approved):
:clap: good idea @albatrousse !

PR #8768, opened 2023-07-31
taxonomy: Add ingredients in Catalan and Spanish

duhow:


benbenben2 (approved):
Looks good.

Suggestions, for future pull requests:
- when you add new entry, check if it exists in wikidata. If yes, you can add the wikidata id and you can copy paste all available translations.
- add the English translation whenever it is missing

PR #8772, opened 2023-08-01
fix: disable facets knowledge panels for crawlers

stephanegigandet:
Related to https://github.com/openfoodfacts/facets-knowledge-panels/pull/123 : we don't want to load facets knowledge panels for crawlers.

In the future, we might want to disable other things for crawlers, this PR makes is_crawl_bot available to the template engine (only for tags right now, it should be added for all templates later)


raphael0202 (approved):
LGTM

PR #8777, opened 2023-08-01
taxonomy: spices and sauces updates

OpenChris:
• add Aniseed alias to Anise, proxy to Fennel same plant as green anise
• add Glutamate as a _Flavour enhancers_
• add "Roux" butter + flour sauces
• link Yakitori sauces to Soy sauces (and get proper Ciqual)
• add Sukiyaki sauces
• add Fish broth to dehydrated broths
• use cooked beetroot as proxy for preserved beetroot
• add dried Chives
• add dried dill
• add dried tarragon
• add mustard seeds as a spice
• add lemon and lime juice concentrates - Proposal : proxy lime juice to lemon juice instead of lime pulp (slightly lower score, more coherent with the pulp-->juice score decrease observed on lemon)
• add English translation to filets de canard
• fix typo on English name for frozen cooked snails : (the snail is cooked, not just the sauce)
• add frozen snails with parsley butter
• add Frozen baguettes under Baguettes
• proxy white asparagus to peeled white asparagus
• proxy all "pâte pressée non cuite" cheeses to existing "Saint Paulin" Agrib.
• remove duplicate entry "stuffings"
• add "frozen" under "vegetable gratins"


OpenChris:
Why would this be different from say "Apples" that are both an ingredient and a category of packaged products?
We would need a category for [these](https://fr.openfoodfacts.org/cgi/search.pl?search_terms=glutamate+monosodique&search_simple=1&action=process) products, wouldn't we?

aleene:
Indeed an ingredient and category can be the same for single ingredient products (this fact is not yet leveraged by the way).

The ingredients are split over multiple files, as some require special processing (legal and regulation). Ingredients.txt is the main one, specials are additives, vitamins, minerals and some others. If all is well there are no duplicates ingredients between those files.

OpenChris:
Thanks, so how would you name the category for these packaged products: [6901104158113](https://fr.openfoodfacts.org/produit/6901104158113/gourmet-powder-ve-tsin), [8850464000311](https://fr.openfoodfacts.org/produit/8850464000311/monosodium-glutamate-young-co) or [8850464000083](https://fr.openfoodfacts.org/produit/8850464000083/glutamate-monosodique)?
I suggest either:
1) Glutamate
2) Monosodium glutamate

taxonomies/categories.txt
```diff
@@ -65094,6 +65097,10 @@ en:Garam masala
 ja:ガラムマサラ
 xx:Garam masala
 
+<en:Spices
```
moon-rabbitOFF, line 65100:
```suggestion
<en:Flavour enhancers
```

moon-rabbitOFF (changes requested):

taxonomies/categories.txt
stephanegigandet: @OpenChris additives.txt is automatically included in ingredients.txt, so glutamate should already be covered, did you find products where it doesn't work?

stephanegigandet: ah my mistake I thought it was ingredients.txt and not categories.txt

Issue #8764, opened 2023-07-31
Performance (Tracker)


teolemon:
### What
- Ensure that Open Food Facts has a stellar App performance
### Why
- People will leave negative reviews on the app
- Search engine will downrank Open Food Facts

```[tasklist]
### Tasks
```

### Part of
- #525

PR #8790, opened 2023-08-02
feat: Timeout before calling autocomplete

duhow:
### What

As users are writing into Tagified content that performs autocompletion, each letter typed will result in an immediate call to API.

image

This amounts of requests and cancellations may overload the API service, which in turn cause that lookup takes long (+5 seconds to get response).

In order to improve this situation:
- Whenever the user types any letter, a `timeout` starts.
- If the user continues to write, the `timeout` will restart from beginning.
- If the `timeout` completes, the REST API call will be executed.

---

**NOTE:** This code is not tested. Feel free to make any additional changes required.

### Part of
- #8764

### DIFF
--- a/html/js/product-multilingual.js
+++ b/html/js/product-multilingual.js
@@ -545,7 +545,7 @@ function initializeTagifyInput(el) {
 
     let abortController;
     let debounceTimer;
-    let timeoutWait = 300;
+    const timeoutWait = 300;
 
     input.on("input", function(event) {
         const value = event.detail.value;
--- a/html/js/product-multilingual.js
+++ b/html/js/product-multilingual.js
@@ -544,26 +544,33 @@ function initializeTagifyInput(el) {
     });
 
     let abortController;
+    let debounceTimer;
+    let timeoutWait = 300;
+
     input.on("input", function(event) {
         const value = event.detail.value;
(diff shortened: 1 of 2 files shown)


html/js/product-multilingual.js
```diff
@@ -544,26 +544,33 @@ function initializeTagifyInput(el) {
...
         if (el.dataset.autocomplete && el.dataset.autocomplete !== "") {
-            // https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort
-            if (abortController) {
-                abortController.abort();
-            }
+            clearTimeout(debounceTimer);
+
+            debounceTimer = setTimeout(function(){
+                // https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort
+                if (abortController) {
```
stephanegigandet, line 559: Shouldn't we do the abort before the setTimeout? i.e. when the user types something, we cancel the previous request immediately, and then we issue the request after 300ms? Otherwise we could get back suggestions that don't match all the letters we typed.

stephanegigandet:
Good idea I think, I will test it.

stephanegigandet:
Seems to work quite well. :)

stephanegigandet (approved):
Thank you!

Issue #8779, opened 2023-08-01
Disable completely indexation on subdomains with unusal cc-lc pairs (ex: fr-it, uk-es,...)


raphael0202:
## What

Web crawlers can currently index every country code/language code pairs possible:
For example, `fr-es.openfoodfacts.org` currently has 120k webpages indexed: [URL](https://www.google.com/search?q=site%3Afr-es.openfoodfacts.org&source=hp&ei=3AHJZIfaCrjskdUP77m06AQ&iflsig=AD69kcEAAAAAZMkP7EQ5ZVdRcuyiFccsKDgoWH9BBOMJ&ved=0ahUKEwiHuIeFwruAAxU4dqQEHe8cDU0Q4dUDCAg&uact=5&oq=site%3Afr-es.openfoodfacts.org&gs_lp=Egdnd3Mtd2l6IhxzaXRlOmZyLWVzLm9wZW5mb29kZmFjdHMub3JnSJkxUABYgTBwAXgAkAEAmAFKoAHyC6oBAjI5uAEDyAEA-AEBwgILEAAYgAQYsQMYgwHCAhEQLhiDARjHARixAxjRAxiABMICERAuGIAEGLEDGIMBGMcBGNEDwgILEC4YigUYsQMYgwHCAggQABiABBixA8ICCBAuGIAEGLEDwgILEC4YgAQYsQMYgwHCAgUQABiABMICCxAAGIoFGLEDGIMBwgIIEAAYgAQYyQPCAggQABiKBRiSAw&sclient=gws-wi)

There are [242 country code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) (+ world subdomain) and [183 language codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) = 34 749 subdomains to crawl.
This is obviously not the right approach to index everything.

## Proposed solution

1. Only allow access to crawlers to subdomains with (country code, lang code) pairs that makes sense (official language for this country). This can be done by (we can implement both approach):
    1. having a dynamic robots.txt that is generated on the fly by product opener (Disable /)
    2. returning `noindex` page for web crawlers
2. Deny access to bots to all world-{lc} where {lc} != 'en', to avoid allowing the indexation of 530 millions webpages (2,9M products * 183 language codes).

[...]

PR #8791, opened 2023-08-02
fix: Improve web crawlers indexation

raphael0202:
This PR introduces all changes mentioned in #8779.

## Prevent web crawlers from indexing unwanted subdomains

Through `noindex` page and `robots.txt`. `robots.txt` is now dynamic, it depends on the subdomain:
- use standard `robots.txt` for `{cc}` subdomain and `world`
- use a deny-all `robots.txt` for `world-{lc}` and ssl-api

## Restrict most facet pages for web crawlers
in standard `robots.txt` (+ `noindex` page for crawlers that don't respect robots.txt).
We now also return noindex page on list of tags pages.

Normalization of of all facet tagtype in URLs has to be discussed first and will be implemented (if we decide to do so) in another PR.


lib/ProductOpener/Display.pm
```diff
@@ -900,12 +925,12 @@ Set two attributes to `request_ref`:
...
-	my $user_agent = user_agent();
-	$request_ref->{user_agent} = $user_agent;
+	my $user_agent_str = user_agent();
+	$request_ref->{user_agent} = $user_agent_str;
 
 	my $is_crawl_bot = 0;
-	if ($user_agent
-		=~ /Googlebot|Googlebot-Image|bingbot|Applebot|YandexBot|YandexRenderResourcesBot|DuckDuckBot|DotBot|SeekportBot|AhrefsBot|DataForSeoBot|SeznamBot|ZoomBot|MojeekBot|QRbot|www\.qwant\.com|facebookexternalhit/
+	if ($user_agent_str
+		=~ /Googlebot|Googlebot-Image|Google-InspectionTool|bingbot|Applebot|YandexBot|YandexRenderResourcesBot|DuckDuckBot|DotBot|SeekportBot|AhrefsBot|DataForSeoBot|SeznamBot|ZoomBot|MojeekBot|QRbot|www\.qwant\.com|facebookexternalhit/
```
stephanegigandet, line 933:
```suggestion
		=~ /Googlebot|Googlebot-Image|Google-InspectionTool|bingbot|Applebot|YandexBot|YandexRenderResourcesBot|DuckDuckBot|DotBot|SeekportBot|AhrefsBot|DataForSeoBot|SeznamBot|ZoomBot|MojeekBot|QRbot|www\.qwant\.com|facebookexternalhit/i
```

Just in case bingbot is renamed to BingBot...

lib/ProductOpener/Display.pm
```diff
@@ -1069,6 +1094,60 @@ sub display_no_index_page_and_exit () {
...
+- otherwise: the standard robots.txt. We disallow indexing of most facet pages, the
+  exceptions can be found in ProductOpener::Config::index_tag_types
+
+=cut
+
+sub display_robots_txt_and_exit ($request_ref) {
+	my $template_data_ref = {facets => []};
+	my $vars = {deny_access => $request_ref->{deny_all_robots_txt}, disallow_paths_localized => []};
+
+	foreach my $type (keys %tag_type_singular) {
```
stephanegigandet, line 1113: Addiing a sort so that we always get the same order in the robots.txt file (useful for testing)

```suggestion
	foreach my $type (sort keys %tag_type_singular) {
```

lib/ProductOpener/Routing.pm
```diff
@@ -574,20 +577,26 @@ sub analyze_request ($request_ref) {
...
 	# Return noindex empty HTML page for web crawlers that crawl specific facet pages
-	if (($request_ref->{is_crawl_bot} eq 1) and (defined $request_ref->{tagtype})) {
-		if ($request_ref->{tagtype} !~ /^brands|categories|labels|additives|nova-groups|ecoscore|nutrition-grades$/) {
-			# Only allow indexation of a selected number of facets
-			# Ingredients were left out because of the number of possible ingredients (1.2M)
-			$request_ref->{no_index} = 1;
-		}
-		elsif ($request_ref->{page} >= 2) {
-			# Don't index facet pages with page number > 1 (we want only 1 index page per facet value)
+	if ($request_ref->{is_crawl_bot} eq 1) {
```
stephanegigandet, line 580:
```suggestion
	if ($request_ref->{is_crawl_bot} == 1) {
```

or just if ($request_ref->{is_crawl_bot})

lib/ProductOpener/Routing.pm
```diff
@@ -574,20 +577,26 @@ sub analyze_request ($request_ref) {
...
+		if (defined $request_ref->{groupby_tagtype}) {
+			# $request_ref->{no_index} is set to 0 by default in init_request()
 			$request_ref->{no_index} = 1;
```
stephanegigandet, line 583: Can we remove all the elsif and have only one if, so that we don't repeat $request_ref->{no_index} = 1; 4 times? We can have the comments inside the conditions

tests/integration/page_crawler.t
```diff
@@ -136,6 +156,101 @@ my $tests_ref = [
...
+	{
+		test_case => 'crawler-get-non-official-cc-lc',
+		method => 'GET',
+		path => '/?cc=world&lc=es',
+		headers_in => {'User-Agent' => $CRAWLING_BOT_USER_AGENT},
+		expected_status_code => 200,
+		expected_type => 'html',
+		response_content_must_match => '<h1>NOINDEX</h1>'
+	},
+	# Indexing should be disabled on all world-{lc} subdomains
```
stephanegigandet, line 190: For the robots.txt tests, I think it would be worth it to save the whole robots.txt we expect (just like we do for JSON responses). That way we see everything that's added / changed / removed, now and in the future when we add facets.

lib/ProductOpener/Display.pm
```diff
@@ -1069,6 +1094,60 @@ sub display_no_index_page_and_exit () {
...
+			my $tag_value_singular = $tag_type_singular{$type}{$lang};
+			my $tag_value_plural = $tag_type_plural{$type}{$lang};
+			if (
+					defined $tag_value_singular
+				and length($tag_value_singular) != 0
+				and not($tag_value_singular ~~ $vars->{disallow_paths_localized})
+				# check that it's not one of the exception
+				# we don't perform this check below for list of tags pages as all list of
+				# tags pages are not indexable
+				and not($type ~~ @ProductOpener::Config::index_tag_types)
```
stephanegigandet, line 1125: in the initialization of Display.pm, we could create a %index_tag_types hash with keys for all entries in @index_tag_types so that we can directly check if the key exists, instead of iterating on each array entry.

On a side note, the ~~ smartmatch operator is deprecated in the newest Perl and will be removed in future versions.

lib/ProductOpener/Display.pm
```diff
@@ -733,6 +746,19 @@ sub init_request ($request_ref = {}) {
 		}
 	}
 
+	# If lc is not one of the official languages of the country and if the request comes from
+	# a bot crawler, don't index the webpage (return an empty noindex HTML page)
+	# We also disable indexing for all subdomains that don't have the format world, cc or cc-lc
+	my $hostname_components_count = $hostname =~ tr/\.//;
```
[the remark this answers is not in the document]
raphael0202: I found some weird domains indexed in Google search console (such as *preprod.openfoodfacts.org or *m.openfoodfacts.org), but I remember it was mostly linked with `ssl-api` subdomain, so it should be fine to remove it. We can still add exceptions if needed.

tests/integration/page_crawler.t
raphael0202: Good idea indeed, I'm adding this.

lib/ProductOpener/Routing.pm
```diff
@@ -574,21 +577,29 @@ sub analyze_request ($request_ref) {
...
+						or (defined $request_ref->{tagtype2})
+					)
+				)
+			)
+		)
+		)
+	{
+		# $request_ref->{no_index} is set to 0 by default in init_request()
+		$request_ref->{no_index} = 1;
 	}
```
alexgarel, line 603: would have been worth making it a sub…

lib/ProductOpener/Display.pm
```diff
@@ -1069,6 +1099,60 @@ sub display_no_index_page_and_exit () {
...
+				and ($type !~ /^ecoscore|products$/) and not($tag_value_plural ~~ $vars->{disallow_paths_localized})
+				)
+			{
+				push(@{$vars->{disallow_paths_localized}}, $tag_value_plural);
+			}
+		}
+	}
+
+	my $text;
+	$tt->process("web/pages/robots/robots.tt.txt", $vars, \$text);
```
alexgarel, line 1149: Would it be worth putting the result in memcached ?

lib/ProductOpener/Display.pm
```diff
@@ -1069,6 +1099,60 @@ sub display_no_index_page_and_exit () {
...
+
+	foreach my $type (sort keys %tag_type_singular) {
+		# Get facet name for both english and the request language
+		foreach my $lang ('en', $request_ref->{lc}) {
+			my $tag_value_singular = $tag_type_singular{$type}{$lang};
+			my $tag_value_plural = $tag_type_plural{$type}{$lang};
+			if (
+					defined $tag_value_singular
+				and length($tag_value_singular) != 0
+				and not($tag_value_singular ~~ $vars->{disallow_paths_localized})
```
alexgarel, line 1126: Usually if you gonna check an element is not in a list, it's better to construct a hashmap and store values in keys, it makes existence checking O(1)).

I'm not sure it's significant here though.

lib/ProductOpener/Display.pm
```diff
@@ -1069,6 +1099,60 @@ sub display_no_index_page_and_exit () {
...
+				# we don't perform this check below for list of tags pages as all list of
+				# tags pages are not indexable
+				and not(exists($index_tag_types_set{$type}))
+				)
+			{
+				push(@{$vars->{disallow_paths_localized}}, $tag_value_singular);
+			}
+			if (
+				defined $tag_value_plural
+				and length($tag_value_plural)
```
alexgarel, line 1137: Don't you check it's not already in $vars->{disallow_paths_localized} ?

templates/web/pages/robots/robots.tt.txt
```diff
@@ -1,9 +1,13 @@
-User-agent: *
+[% IF deny_access %]User-agent: *
+Disallow: /
+[% ELSE %]User-agent: *
```
alexgarel, line 3: you can use [chomping](https://metacpan.org/dist/Template-Toolkit/view/lib/Template/Manual/Syntax.pod#Chomping-Whitespace) `-%` to make it more readable

```suggestion
[% IF deny_access -%]
User-agent: *
Disallow: /
[% ELSE -%]
User-agent: *
```

…and so one also below.

raphael0202:
@stephanegigandet I've made the requested modifications :)

alexgarel (approved):
Great work !

Some minor comments.

lib/ProductOpener/Display.pm
raphael0202: I don't think so, it's super fast to load, and robots.txt is rarely fetched by crawlers (it takes up to a month to be updated by Bingbot for example)

lib/ProductOpener/Routing.pm
raphael0202: yes indeed, I'm changing it

lib/ProductOpener/Display.pm
raphael0202: It's indeed not a concern here performance-wise.

Issue #8792, opened 2023-08-02
Memcache for the Homepage


teolemon:
Loading the webpage takes two seconds. That means, but we probably don't use memcache to serve it. Amirite ?
@stephanegigandet @raphael0202

raphael0202:
caching in memcache is enabled, but it looks like there is an error when saving MongoDB results in memcached:

```
[267909] /opt/product-opener/lib/ProductOpener/Display.pm 1375 mongodb {} get_cache_results - miss - key: openfoodfacts.net:search_products/32d75fc9c3e8b63dd2f640a563e199ee
[268135] /opt/product-opener/lib/ProductOpener/Display.pm 1390 mongodb {} set_cache_results - setting value - key: openfoodfacts.net:search_products/32d75fc9c3e8b63dd2f640a563e199ee - total_size: 26584379
[269054] /opt/product-opener/lib/ProductOpener/Display.pm 1398 mongodb {} set_cache_results - error - key: openfoodfacts.net:search_products/32d75fc9c3e8b63dd2f640a563e199ee
```

the test was performed on staging. On prod when I filtered `mongodb_log4perl.log` by 'error', errors were logged everytime I hit the home page (non logged-in user), so I suspect the issue can be reproduced in staging.  

it works correctly locally.

teolemon:
Nice catch @raphael0202 👍

raphael0202:
After further investigation, here is what I found.
Due to personalized search, we always request all fields to MongoDB when searching products on OFF: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Display.pm#L4978
Most popular products (as displayed on the main page of OFF) are the most complete ones (and the ones with the most data), hence the giant 30MB response from MongoDB.
We should definitely restrict the fields to only the required ones. We can expect from this:
- faster queries to MongoDB
- use of memcached (faster loading time)

stephanegigandet:
> We should definitely restrict the fields to only the required ones.

Definitely. We'll have to be very careful to list all the fields we need for the attributes.

Also there are some fields we may need to organize differently (e.g. images, so that we don't pull the list of the 542324 images of nutella just to know which one is the front image...)

raphael0202:
> Definitely. We'll have to be very careful to list all the fields we need for the attributes.

Is it listed somewhere? I had trouble understanding where this is computed in Product Opener code.

[...]

PR #8797, opened 2023-08-04
fix: don't save object in memcached above allowed size 

raphael0202:
First PR to fix memcached caching error investigated in #8792. It doesn't solve the caching issue, but at least we're not trying to save a 30MB blob in memcached anymore.
It turns out this saving attempt was very costly, with this PR the main webpage loads in 600-1000ms (instead of 1500-2000s), it's deployed and can be tried on https://openfoodfacts.net.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -1372,9 +1372,17 @@ sub get_cache_results ($key, $request_ref) {
 sub set_cache_results ($key, $results) {
 
 	$log->debug("Setting value for MongoDB query key", {key => $key}) if $log->is_debug();
+	my $result_size = total_size($results);
+
+	# memcached max object size is 1 048 576 bytes
+	if ($result_size >= 1048576) {
+		$mongodb_log->info(
+			"set_cache_results - skipping - setting value - key: $key (total_size: $result_size > max size)");
+		return;
+	}
 
 	if ($mongodb_log->is_debug()) {
-		$mongodb_log->debug("set_cache_results - setting value - key: $key - total_size: " . total_size($results));
+		$mongodb_log->debug("set_cache_results - setting value - key: $key - total_size: $result_size");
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -1385,9 +1372,17 @@ sub get_cache_results ($key, $request_ref) {
 sub set_cache_results ($key, $results) {
 
 	$log->debug("Setting value for MongoDB query key", {key => $key}) if $log->is_debug();
+	my $result_size = total_size($results);
+
+	# memcached max object size is 1 048 576 bytes
+	if ($result_size >= 1048576) {
```
stephanegigandet, line 1378: Can we put this number as a constant in Cache.pm ? 1M is the default memcached max object size, but it could be changed.

lib/ProductOpener/Display.pm
```diff
@@ -1201,8 +1190,6 @@ sub display_text ($request_ref) {
 		$html .= search_and_display_products($request_ref, {}, "last_modified_t_complete_first", undef, undef);
 	}
 
-	$html =~ s/\[\[query:(.*?)\]\]/$replace_query->($1)/eg;
```
stephanegigandet, line 1204: This feature (including the result of a query inside a text) is not used anymore, so I think it's ok to remove it indeed.

```
lang/fr/texts/index.old.html:[[query:{"lc":"fr","sort_by":"last_modified_t_complete_first"}]]
lang/fr/texts/index-test.html:[[query:{"lc":"fr","sort_by":"last_modified_t"}]]

```

stephanegigandet (approved):
Awesome! I just have a minor request to make the 1M number a variable in Cache.pm, it may be that we need a bit more than 1M for some uses.

alexgarel:
Awesome @raphael0202 !

PR #8811, opened 2023-08-08
fix: obsolete support in update_all_products.pl and export status fix on producers platform

stephanegigandet:
### What
- When exporting obsolete products from the pro platform to the public platform, we were not updating their export status on the obsolete collection of the pro platform (and instead created them on the non obsolete collection).
- This PR fixes the export status, and also add support for selecting the obsolete collection in update_all_products.pl, + an option to remove obsolete / non-obsolete products that are not in the appropriate collection.
```
16019 products updated (pretend: ) - 0 new versions created
9227 removed from wrong collection (obsolete or current)
off@off1:/srv/off-pro/scripts$ ./update_all_products.pl --query owners_tags=org-auchan-apaw --fix-obsolete 
```
### Part of
- #8814

### DIFF
--- a/lib/ProductOpener/Import.pm
+++ b/lib/ProductOpener/Import.pm
@@ -2824,14 +2824,12 @@ sub update_export_status_for_csv_file ($args_ref) {
 	my $columns_ref = $csv->getline($io);
 	$csv->column_names(@{deduped_colnames($columns_ref)});
 
-	my $products_collection = get_products_collection();
-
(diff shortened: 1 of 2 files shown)


teolemon (approved):
I have created a tracker issue for obsolete product issues, and a call for documentation of our policies/behaviours on the wiki

alexgarel (approved):
LGTM

Issue #8750, opened 2023-07-28
The classic Open Products Facts gets a 413 HTTP Error


teolemon:
### What
- The classic Open Products Facts gets a 413 HTTP Error
- [413 Payload Too Large - HTTP - MDN Web Docs](https://developer.mozilla.org/fr/docs/Web/HTTP/Status/413)

### Steps to reproduce the behavior

1. Install Open Products Facts classic
2. Add a new product
3. Add a front photo and then an ingredients photo
4. See 413 error as a tooltip

alexgarel:
Might be linked to the move on off2 server. I will check for nginx configuration.

alexgarel:
Here are nginx difference with previous config:

```
$ diff /tmp/nginx-off1.conf /etc/nginx/nginx.conf 
19,20d18
< 	tcp_nodelay on;
< 	keepalive_timeout 65;
34c32
< 	ssl_protocols TLSv1 TLSv1.1 TLSv1.2 ; # Dropping SSLv3, ref: POODLE
---
> 	ssl_protocols TLSv1 TLSv1.1 TLSv1.2 TLSv1.3; # Dropping SSLv3, ref: POODLE
49d46
< 	gzip_disable "msie6";
56c53
< 	gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript text/csv;
---
> 	# gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
64,65d60
< 
< 	client_max_body_size 20M;
```

I will add it to the single configuration files, I think it's more explicit.

alexgarel:
This should be fixed now, @teolemon can you test it ?

teolemon:
It is! Thank you @alexgarel

PR #8823, opened 2023-08-09
fix: gzip more files in nginx for Open Product Facts

alexgarel:
part of: #8750


teolemon:
why isn't labeler.yml running and labelling this Open Products Facts ? 🤔

alexgarel:
@teolemon 
> why isn't labeler.yml running and labelling this Open Products Facts ? thinking

because we are not on main branch but on opf-main… github action files are not the same.

alexgarel:
Can you validate @teolemon ?

teolemon (approved):

Issue #8749, opened 2023-07-28
Add a field in user profile to tell which country they live in


alexgarel:
Currently we get language field from the "initial_cc" value which is determined by the domain name of the subscription form.

Add a user_cc field (country where user live).
while initial_cc is mandatory and hidden, user_cc is not mandatory (it can be empty)

Let user modify it at subscription time or in their user profile.

On edition:
* if we have a initial_cc, use it to populate user_cc
* else fallback to actual $cc

On subscription:
* prefill with $cc

### Part of
- #5730

PR #8826, opened 2023-08-09
feat: enable user to edit their preferred language and the country they live in 

### LINKED ISSUE #8748: Enable user to edit the preferred communication language field
Currently we get language field from the "inital_lc" value which is determined by the domain name of the subscription form.

Add a user_lc field (preferred communication language).

Let user modify it at subscription time or in their user profile.

On edition:
* if we have a initial_lc, use it to populate user_lc
* else fallback to actual $lc

On subscription:
* prefill with $lc

### Part of
- #5730

albatrousse:
### What

The goal is to add - when signing up and in the user profile - a field for the user to inform their preferred language and the country they live. When signing up, the fields will be prefill with the country associated with the website (looking at the country code) the user uses and the same goes for the language.
For instance, this could lead to better cohesion between user's preferred language and its newsletter 

- Fixes #8748 and #8749


albatrousse:
I have created two new sub functions get_countries and get_languages which respectively returns a list of the languages (or the countries) in the language stored in $lc

lib/ProductOpener/Display.pm
```diff
@@ -11430,4 +11430,51 @@ sub generate_select2_options_for_taxonomy_to_json ($target_lc, $tagtype) {
...
+sub get_languages{
+	my @languages_list = ();
+	my @tags_list = get_all_taxonomy_entries("languages");
+	my $language;
+	foreach my $tag (@tags_list){
+		$language = display_taxonomy_tag($lc, "languages", $tag);
+		push @languages_list, $language;
+	}
+	return @languages_list;
+}
```
alexgarel, line 11454: This would be could to cache it… let's do it at the end :-)

alexgarel: at the end you should sort the languages_list.

albatrousse: I sorted it before returning the list, but I did not quite understand what should I do.
Why caching it is a bad idea ?

albatrousse:
I'm currently trying to correct the functions get_languages() and get_countries() in Display.pm, they are causing errors I don't know why but I will try to resolve it

alexgarel:
@stephanegigandet I added a commit to try to finish this PR. Can you review ?

lib/ProductOpener/Web.pm
```diff
@@ -291,14 +298,21 @@ The data structured can be passed to HTML templates to construction a list of op
 
 sub get_languages_options_list ($target_lc) {
 
+	return $lang_options_cache{$target_lc} if (defined $lang_options_cache{$target_lc});
+
 	my @lang_options = ();
 
 	my %lang_labels = ();
 	foreach my $l (@Langs) {
```
stephanegigandet, line 306: It wasn't changed in this PR, but can we rename $l to $language_code or something? In the GitHub font I can't really tell $l from $1 : `$l / $1`

lib/ProductOpener/Web.pm
```diff
@@ -291,14 +298,21 @@ The data structured can be passed to HTML templates to construction a list of op
...
-		$lang_labels{$l} = display_taxonomy_tag($target_lc, 'languages', $language_codes{$l});
+		my $label = display_taxonomy_tag($target_lc, 'languages', $language_codes{$l});
+		# remove eventual language prefix
+		if ($label =~ /^..:/) {
```
stephanegigandet, line 309: I would replace the whole condition by just `$label =~ s/^\w\w://;`

lib/ProductOpener/Web.pm
```diff
@@ -311,7 +325,53 @@ sub get_languages_options_list ($target_lc) {
...
+
 	return \@lang_options;
 }
 
+# cache for get_countries
+my %countries_options_lists = ();
+
+=head2 get_countries_options_list( $target_lc )
+
+Generates all the countries name in the $lc language suitable for an select option list
```
stephanegigandet, line 339:
```suggestion
Generates all the countries name in the $target_lc language suitable for an select option list
```

lib/ProductOpener/Web.pm
```diff
@@ -311,7 +325,53 @@ sub get_languages_options_list ($target_lc) {
...
+=head4 $target_lc - language code for labels
+
+=head3 Return value
+
+A reference to a list of hashes with every country code and their label in the $lc language
+[{value => "fr", label => "France"},…]
+
+=cut
+
+sub get_countries_options_list ($target_lc) {
```
stephanegigandet, line 352: There's similar code in /cgi/country.pl which is used to display the drop down of countries in the header of each page. In fact we could reuse the dropdown.

Maybe we could merge the two. The one in country.pl is missing caching, that's a good idea.

lib/ProductOpener/Web.pm
```diff
@@ -311,7 +325,53 @@ sub get_languages_options_list ($target_lc) {
...
+		return $countries_options_lists{$target_lc};
+	}
+	# compute countries list
+	my @countries_list = ();
+	my @tags_list = get_all_taxonomy_entries("countries");
+	foreach my $tag (@tags_list) {
+		next if $tag eq 'en:world';
+		my $country = display_taxonomy_tag($target_lc, "countries", $tag);
+		# remove eventual language prefix
+		if ($country =~ /^..:/) {
```
stephanegigandet, line 364: $country =~ s/^\w\w://;

stephanegigandet (approved):
Looks good, some minor suggestions. Thank you!

PR #8827, opened 2023-08-09
chore: Migration of Open Food Facts from off1 (bare metal) to off2 (proxmox)

alexgarel:
### What
- Migration of Open Food Facts (site and pro platform) from off1 (bare metal) to off2 (proxmox)
- Modifications linked to moving off and off-pro to off2 server. 
- This includes a lot of path modifications:
  - some directories where removed from the repository and are created dynamically
  - we try to differentiate folders containing code (only this is in git), data, cache data and dist files
  - we control this by using a fixed set of base path (see Paths.pm)
- Most of server configurations are now in the git repository
- Import scripts where reworked 

See also: https://github.com/openfoodfacts/openfoodfacts-infrastructure/pull/243

### Part of
- https://github.com/openfoodfacts/openfoodfacts-infrastructure/issues/182


Makefile
```diff
@@ -299,13 +299,15 @@ bash:
...
 # check perl compiles, (pattern rule) / but only for newer files
 %.pm %.pl: _FORCE
-	if [ -f $@ ]; then perl -c -CS -Ilib $@; else true; fi
+	@if [[ -f $@ ]]; then perl -c -CS -Ilib $@; else true; fi
 
 
 # TO_CHECK look at changed files (compared to main) with extensions .pl, .pm, .t
+# filter out obsolete scripts
 # the ls at the end is to avoid removed files.
+# the first commad is to check we have git (to avoid trying to run this line inside the container on check_perl*)
```
teolemon, line 308:
```suggestion
# the first command is to check we have git (to avoid trying to run this line inside the container on check_perl*)
```

cgi/product_image_upload.pl
```diff
@@ -172,13 +173,8 @@
...
 my $interface_version = '20120622';
 
-# Create image directory if needed
-if (!-e "$www_root/images") {
-	mkdir("$www_root/images", 0755);
-}
-if (!-e "$www_root/images/products") {
-	mkdir("$www_root/images/products", 0755);
-}
+# check image directory exists
```
teolemon, line 176:
```suggestion
# Check that the image directory exists
```

lib/ProductOpener/Paths.pm
```diff
@@ -0,0 +1,310 @@
...
+
+=head1 VARIABLES
+
+=cut
+
+=head2 %BASE_DIRS
+A hashmap containing references to base directories
+=cut
+
+%BASE_DIRS = ();
```
stephanegigandet, line 63: I like the idea of having extra documentation about what each directory is, but on the other hand, it's less obvious when reading the code that a directory is for data (input or output), or that it will be publicly accessible on the web. It was more clear to me with $data_root and $www_root. The extra indirection kind of obfuscate things, especially as the mapping from DIR_NAME to the actual name of the directory on the disk varies.

stephanegigandet: Maybe the extra documentation could be put in README.md files instead.
And we could better arrange what we have under $data_root, to clearly separate input and output data.

alexgarel: The main idea was not that but to forbid the dynamic creation of the base folders (creating folders inside base folders is ok, but not creating those base folders). It's to avoid creating data at the wrong place (eg. on the VM disk instead of a ZFS dataset).

You can see that apache start up script verifies all these folders exists or refuse to start (with an explicit message).

I am also boldly in favor of a $data_root and $cache_root, but I didn't want to change everything right now.

Maybe we could prefix folders name, or have more explicit name (we don't use them so many times in the code, so longer names are affordable). At least, add PUBLIC to every public folders.

teolemon:
small merge conflict to solve @alexgarel

stephanegigandet (approved):

PR #8831, opened 2023-08-10
chore: replace packagings banner with OFF days banner

stephanegigandet:
large:

image

mobile:

image

French:

image


alexgarel (approved):
:tada:

PR #8833, opened 2023-08-10
fix: remove legacy robotoff-proxy nginx config

raphael0202:
This config is not used anymore (since january 2023).


alexgarel (approved):
Thank you :broom: !

PR #8838, opened 2023-08-11
fix: add me,ry,sh,wa languages to taxonomy

stephanegigandet:
New .po files were added by Crowdin but we didn't have entries for them in the languages taxonomy, which resulted in those languages not working, and dozens of warnings at startup:

[Fri Aug 11 09:29:39 2023] Food.pm: Redundant argument in sprintf at /opt/product-opener/lib/ProductOpener/Food.pm line 736.


teolemon (approved):
Great, thanks

PR #8840, opened 2023-08-12
feat: Enable Org Admins to View and Modify Member Admin Status

### LINKED ISSUE #8780: Allow org admins to see and change the admin role of other org members on the producers platform
Related to https://github.com/openfoodfacts/openfoodfacts-server/pull/8763

In the table that displays org members, we should display if members are admins of the org, and there should be a way to add or remove admin status. Maybe we could just have an "admin" column with a checkbox.

MonalikaPatnaik:
### Video Demo

https://www.loom.com/share/bf155cf22ffe4fd3a6d3a75dd0b4429a?sid=3961c777-2553-4f23-bfad-e3e23d7913e7

- Fixes #8780


stephanegigandet:
Looks good :)

templates/web/pages/org_form/org_form.tt.html
```diff
@@ -32,6 +33,10 @@ <h2>[% lang("organization_members") %]</h2>
 			[% FOREACH users IN org_members %]
 			<tr>
 				<td>[% count %].</td>
+				<td>
+					<input type="checkbox" name="admin_status_[% users.userid %]" value= "1" [% IF admin_status.users.userid %]checked[% END %]>
```
MonalikaPatnaik, line 37: idk why but the condition inside if block( "admin_status.users.userid")  is not working i.e the checkbox is not already checked for admins.

templates/web/pages/org_form/org_form.tt.html
```diff
@@ -51,6 +68,24 @@ <h2>[% lang("organization_members") %]</h2>
 			[% END %]
 		</tbody>
 		</table>
+		<div>
+			<form id='hiddenInputForm' method="post" action="/cgi/org.pl">
```
alexgarel, line 72: Why not put this around the table above where we display status, this is not a problem as those are the only values that can be modified ?
This method of duplicating the form is smart, but might be more difficult to debug.

cgi/org.pl
```diff
@@ -384,6 +386,30 @@
...
+						push @admin_status, $user_id;
+
+					}
+				}
+				else {
+					remove_user_from_org($org_ref, $user_id, ["admins"]);
+					@admin_status = grep {$_ != $user_id} @admin_status;
+				}
+				store_org($org_ref);
+			}
```
alexgarel, line 408: again it's quite ok, but I would rather build a list of admins according to the form sent, and then compare it to the current list of the org and only act on differences.
(if you sort them before it's more easy to compare lists).

BTW @stephanegigandet could be useful to have libs like [Array::Diff](https://metacpan.org/pod/Array::Diff) for this kind of operations.

cgi/org.pl
```diff
@@ -384,6 +386,30 @@
...
+					unless (grep {$_ == $user_id} @admin_status) {
+						push @admin_status, $user_id;
+
+					}
+				}
+				else {
+					remove_user_from_org($org_ref, $user_id, ["admins"]);
+					@admin_status = grep {$_ != $user_id} @admin_status;
+				}
+				store_org($org_ref);
```
stephanegigandet, line 407: store_org() should be outside the for loop, otherwise we will save the org as many times as there are users

templates/web/pages/org_form/org_form.tt.html
```diff
@@ -32,6 +37,18 @@ <h2>[% lang("organization_members") %]</h2>
 			[% FOREACH users IN org_members %]
 			<tr>
 				<td>[% count %].</td>
+				<td>
+					[% SET user_in_admin_status = 0 %]
+					[% FOREACH admin_id IN admin_status %]
+						[% IF admin_id == users.userid %]
+						  [% SET user_in_admin_status = 1 %]
+ 				        [% END %]
+					[% END %]
```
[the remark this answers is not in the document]
MonalikaPatnaik: But this line `[% SET user_in_admin_status = user_is_admin[userid] %]` gives template error having unexpected token [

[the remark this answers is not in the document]
alexgarel: yes sorry it should be written `[% SET user_in_admin_status = user_is_admin.$userid %]`

templates/web/pages/org_form/org_form.tt.html
alexgarel: you didn't do it ? You would not need any JS if you do it.

alexgarel: it seems you indeed have more than one form, but you can associate your input above to this form by using the "form" attribute on the input. see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#form
I would avoid javascript if possible.

alexgarel:
@MonalikaPatnaik I did a commit to propose changes to simplify form (avoid JS).

Now I let you see if it's working as desired and fix lint and critic.

alexgarel:
sorry @MonalikaPatnaik I didn't have time, yet to look into this but I will !

alexgarel (approved):
Let's merge :-)

PR #8843, opened 2023-08-14
taxonomy: Extend en taxonomy of subcategories of `en:Biscuit with a filled chocolate bar covering`

Naruyoko:
### What

Add en taxonomy for the 4 subcategories of `en:Biscuit with a filled chocolate bar covering` by obvious derivative.


stephanegigandet (approved):
Thank you!

PR #8845, opened 2023-08-14
feat: show image uploader and upload time to all users

stephanegigandet:
In the product edit form, we now display the image uploader and upload time for all users when the mouse hovers the thumbnails.

This feature used to be for moderators only.


alexgarel (approved):
:+1:

PR #8848, opened 2023-08-14
fix: Remove category

stephanegigandet:
Added an option to remove a category (might be useful for mispellings)


alexgarel (approved):
LGTM, thanks.

Issue #8844, opened 2023-08-14
Vegan analysis on ingredients fails even if all sub ingredients are vegan


alexgarel:
Bug described by Alizarine on slack.

Why is this "chocolate" ingredient labelled as "maybe" vegan while all the ingredients it is composed of are considered "vegan" on [5690576575023](https://world.openfoodfacts.org/product/5690576575023/dark-chocolate-with-caramelized-almonds-and-icelandic-sea-salt-sirius) ?

![image](https://github.com/openfoodfacts/openfoodfacts-server/assets/144788/03b71f24-15fa-41fe-aae1-40fa67549334)

Ingredients:
```
Chocolate (Cocoa mass, sugar, cocoa butter, emulsifier (soya lecithin), bourbon vanilla. Cocoa dry solids: min. 56%), caramelized almonds (14%) (almonds, sugar), sea salt (1%).
```

aleene:
Probably the chocolate hierarchy contains a maybe vegan ingredient.

ArturLange:
Actually the `en:chocolate` itself has an explicitly defined `vegan:en:maybe`.
It should be `vegan:en:ignore` instead, right?

aleene:
I think we need another approach. If chocolate is a standalone ingredient it should be maybe, if it is a compound ingredient, it can be ignore. 
This can be generalised for all ingredients.

ArturLange:
I thought that's the way `ignore` works right now.
It sounds reasonable 👍

stephanegigandet:
In the current implementation:
- all sub ingredients of an ingredient are analyzed, regardless of the status (yes, no, maybe, ignore) status of their parent
- the ingredient itself is completely ignored if its status is "ignore" or if the property is palm oil and its status is "maybe" and it has sub ingredients.

I don't remember why I made that distinction only for palm oil, and not vegan and vegetarian.

That means that if a product has a list like "vegetables, sauce", sauce is marked as "ignore", so the product will be marked vegetarian.

I think we should change all "ignore" properties in the ingredients.txt taxonomy to "maybe", and extend the rule that if we have sub-ingredients, then we ignore the "maybe" parent and instead look at its sub-ingredients.

stephanegigandet:
The related PR is here: https://github.com/openfoodfacts/openfoodfacts-server/pull/8849
Note that it does change a bit the ouput of some tests, but it looks fine to me.

PR #8849, opened 2023-08-14
fix: simplify vegan/vegetarian ingredients analysis and merge 'ignore' and 'maybe'

stephanegigandet:
fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/8844


alexgarel (approved):
Great !

PR #8850, opened 2023-08-14
chore: remove unused code/variable full_width

stephanegigandet:
Removing unused coed. In the new website design, there's no side column, so no need for a full width mode either.


alexgarel (approved):
Thanks for cleaning ! :broom:

PR #8851, opened 2023-08-14
fix: Logic error in user account form

### LINKED ISSUE #8824: Delete sentences in the professional account creation form
When I create an account in the Pro platform, even if I don't enter any value in the field "organisation", the 2 following sentences appear : 
image

I assume these sentences would be the result of a logical check after the user has entered the name of his/her organisation (?) 

Could someone help fix this, please?

MonalikaPatnaik:
### What

We have to display the sentences( for existing org warning) in the user account form only when type is edit i.e when an user edits their account parameters, not when an account is created 

- Fixes #8824


stephanegigandet:
Hi @MonalikaPatnaik , your change would solve the symptom (wrong strings that are displayed), but here the bigger question is why is requested_org_ref defined when we display the form to add an user.

There's a much bigger bug in org.pl in fact.

Because of it, we load the org for the "" (empty orgid) org, which was in fact created (because of another bug), so we load the /orgs/.sto file, which contains org data with the "Cacolac" name that we see in Manon's screenshot.

off1:/srv/off/scripts# ls -lrt ../orgs/.sto 
-rw-r--r-- 1 off off 337 Sep 26  2022 ../orgs/.sto

So instead of changing the template, the bug in org.pl should be fixed instead.

stephanegigandet:
@MonalikaPatnaik I made a mistake, the issue is not in org.pl, it's in user.pl, here:

```
	elsif ((defined $options{product_type}) and ($options{product_type} eq "food")) {
		my $requested_org_ref = retrieve_org($user_ref->{requested_org});
		$template_data_ref->{requested_org_ref} = $requested_org_ref;
		$template_data_ref->{org_name} = sprintf(lang("add_user_existing_org"), org_name($requested_org_ref));
		$template_data_ref->{teams_flag}
			= not((defined $server_options{private_products}) and ($server_options{private_products}));
	}
```

we should only retrieve the requested org if $user_ref->{requested_org is defined and not empty

and in retrieve_org(), we should check that the argument is defined and not empty too

because otherwise, if there's no requested org, then we try to load the file ".sto", which in fact exists on the production server

manon-corneille:
Any news on this bug by any chance? :)

MonalikaPatnaik:
> Any news on this bug by any chance? :)

Doing it..will do it by today :)

stephanegigandet (approved):
Perfect, thank you

PR #8854, opened 2023-08-15
feat: Added display message for normal users for protected images

MonalikaPatnaik:
### What

-  showing the “upload” button
when the image is uploaded, a message “Image received” is displayed only, no need to display cropping tool for normal users.
-  the already selected image (from the producer) is shown.
- not showing the unselect button (only shown for admins/moderators)
- if the user is a moderator / an admin, he/she should still be able to select / unselect / crop images

### Screenshot

Screenshot 2023-09-28 205250

- Fixes #[ISSUE NUMBER]


html/js/product-multilingual.js
```diff
@@ -704,24 +714,26 @@ function get_recents(tagfield) {
 
                 var $this = $(this);
                 var id = $this.attr('id');
+                var clas= $this.attr("data-info");
```
stephanegigandet, line 717: the variable is named "clas" because "class" is a protected word? As the attribute we read is not the class anyway, could we rename the variable to data_info maybe?

stephanegigandet (approved):
I tested it, it is working well, thank you!

html/js/product-multilingual.js
```diff
@@ -427,16 +427,26 @@ function update_nutrition_image_copy() {
 
 function update_display(imagefield, first_display) {
 
-    var display_url = imagefield_url[imagefield];
+    var flag=0;
```
stephanegigandet, line 430: It's a bit strange to use the imagefield to specify that the image is protected, and then removing "protect_" etc.

Could you just add a parameter to update_display() ?   update_display(imagefield, protected, first_display). It would be clearer that way I think, and we would not need to have the flag, modifiedimagefield etc.

stephanegigandet (approved):
Perfect, thank you!

Issue #5546, opened 2021-08-25
Ingredients empty if ingredient text empty for main language


zond:
### What
I refer to our conversation in Slack (https://openfoodfacts.slack.com/archives/C0289KSGR3K/p1629800910053500).

TL;DR

If the ingredient text for the main language is empty, the ingredient list will be empty too - even if other languages have ingredient texts. Me an @CharlesNepote at least agreed that it was reasonable to take some other language in that case.

### Part of
- #5706

raphael0202:
@alexgarel we discussed this a few weeks ago, but there is actually already an issue filled on Github for this ;)
I think as well that the current approach is confusing to users, as they have to change the main language for ingredient analysis to be performed.

I would suggest:
- to parse ingredient list in product main language first if available, otherwise check other available languages.
- add a new `ingredients_lc` field in MongoDB (and sto files) that store the language code that was used to parse ingredient list.
- display on product page (in ingredient analysis panel) and product edit page the language that was used to parse the ingredient list.

This would also make implementation of automatic ingredient extraction by Robotoff easier, as we wouldn't have to switch product main language for ingredient parsing to be done (and @stephanegigandet was anyway not in favor of changing the main language, as it is supposed to be the most visible language on the packaging).

raphael0202:
Done by #8855.

PR #8855, opened 2023-08-15
fix: use ingredient list if lang != lc to parse ingredients

raphael0202:
Solves #5546.

As we add a new field, many expected results files are updated.
Many changes are added to search expected results files, but most of them don't seem to be due to the changes added in this PR.


stephanegigandet:
> Many changes are added to search expected results files, but most of them don't seem to be due to the changes added in this PR.

Those changes are due to the way we construct the products to be searched on, by calling the API with values like:

```
	{
		%{dclone(\%default_product_form)},
		(
			code => '200000000039',
			lang => "es",
			product_name => "Vegan Test Snack",
			generic_name => "Tester",
			ingredients_text => "apple, water, palm oil",
			origin => "spain",
			categories => "snacks"
		)
	},
```

There's a bug in the API, and this results in "ingredients_text" + "ingredients_text_en" to be set with "apple, water, palm oil".

With the existing code, we assume "ingredients_text" is in Spanish (because of the lang set to "es"). With your code, we don't look at ingredients_text and instead see the "ingredients_text_en" and assume it is in English.

stephanegigandet:
I'll fix the API bug.

stephanegigandet:
Could you add a couple of tests cases specifically for this new feature?

stephanegigandet:
@raphael0202 This should fix the test differences in search: https://github.com/openfoodfacts/openfoodfacts-server/pull/8877

stephanegigandet (approved):
Looks good to me

PR #8856, opened 2023-08-15
feat: Allowing Users to subscribe to the pro-newsletter

### LINKED ISSUE #8505: Make it possible to subscribe to the pro newsletter
Story: #
Today, When a user creates an account on the Pro platform and ticks “Subscribe to the newsletter”, we do keep the info, but it's not used. 
As we are now publishing a newsletter every couple of months from SendinBlue, it would be good to allow users to subscribe to it. 

### Part of
- #5521 

image

MonalikaPatnaik:
### What

tried writing functions in Brevo.pm module for adding users to contact list and getting their contact info.

- Fixes #8505


lib/ProductOpener/Brevo.pm
```diff
@@ -0,0 +1,121 @@
...
+        username => $username,
+        cc => $cc,
+        lc => $lc,
+        # Add more contact data fields as needed
+    };
+
+    my $json_data = encode_json($contact_data);
+
+    my $request = POST("$api_base_url$api_endpoint", %headers, Content => $json_data);
+
```
MonalikaPatnaik, line 82: I don't know whether it should be done in this way?

tests/unit/brevo.t
```diff
@@ -0,0 +1,46 @@
...
+use ProductOpener::Brevo qw/:all/;
+use ProductOpener::APITest qw/:all/;
+use ProductOpener::Test qw/:all/;
+use File::Temp ();
+
+# Mock the LWP::UserAgent module
+my $mocked_ua = Test::MockModule->new('LWP::UserAgent');
+$mocked_ua->mock(
+    'request' => sub {
+        my ($request) = @_;
```
alexgarel, line 17: you could instead define $request above, so that after the test you can verify you get what you expected:

```suggestion
my $request;
mock(...
	($request) = @_;
	...
);
...
# test
...
is_deeply($request, {....});
```

tests/unit/brevo.t
```diff
@@ -0,0 +1,46 @@
...
+
+# my $dump_path = File::Temp->newdir();
+# my @responses = (
+# 	HTTP::Response->new("200", "OK", HTTP::Headers->new(), '{"status": "success"}'),
+# );
+# # start fake server
+# my $httpd = fake_http_server(8881, $dump_path, \@responses);
+
+# Call the function
+my $result = add_contact_to_list($email, $username, $cc, $lc);
```
alexgarel, line 38: Due to my proposal above, you will also need to mock $ProductOpener::Brevo::brevo_api_key; to contain something.

tests/unit/brevo.t
```diff
@@ -0,0 +1,46 @@
...
+# 	HTTP::Response->new("200", "OK", HTTP::Headers->new(), '{"status": "success"}'),
+# );
+# # start fake server
+# my $httpd = fake_http_server(8881, $dump_path, \@responses);
+
+# Call the function
+my $result = add_contact_to_list($email, $username, $cc, $lc); 
+diag("Result: $result");
+
+ok($result, 'Contact added successfully');
```
alexgarel, line 41: if we add the feature toggle on brevo_api_key, we must verify result more precisely
```suggestion
is($result, 1, 'Contact added successfully');
```

alexgarel (changes requested):
Thanks, It's almost good, some changes proposed though to avoid slowing down dev environment and tests.

tests/unit/brevo.t
```diff
@@ -0,0 +1,91 @@
...
+# Test the add_contact_to_list function
+{
+	my $request;
+
+	# Mock $ProductOpener::Brevo::brevo_api_key
+	my $mocked_brevo_api_key = Test::MockModule->new('ProductOpener::Brevo');
+	$mocked_brevo_api_key->mock(
+		'brevo_api_key' => sub {
+			return $brevo_api_key;
+		}
```
alexgarel, line 24: why not directly a constant here ?
```suggestion
		'brevo_api_key' => $brevo_api_key;
```

tests/unit/brevo.t
```diff
@@ -0,0 +1,91 @@
...
+	my $mocked_ua = Test::MockModule->new('LWP::UserAgent');
+	$mocked_ua->mock(
+		'request' => sub {
+			($request) = @_;
+			# diag($request);
+			return HTTP::Response->new("200", "OK", HTTP::Headers->new(), '{"status": "success"}');
+		}
+	);
+
+	$brevo_api_key = 'abcdef1234';
```
alexgarel, line 37: I don't think this one is useful

tests/unit/brevo.t
```diff
@@ -0,0 +1,91 @@
...
+	$username = 'elly';
+	$country = 'world';
+	$language = 'english';
+
+	# Call the function
+	my $result = add_contact_to_list($email, $username, $country, $language);
+	diag("Result: $result");
+
+	is($result, 1, 'Contact added successfully');
+	is($brevo_api_key, 'abcdef1234', 'Verify brevo_api_key');
```
alexgarel, line 48: why this ?
You should instead verify $request content

tests/unit/brevo.t
```diff
@@ -0,0 +1,91 @@
...
+
+	my $ua = LWP::UserAgent->new;
+	my $response = $ua->request($request);
+	print to_json(decode_json($response->decoded_content), {pretty => 1});
+	is_deeply(
+		decode_json($response->decoded_content),
+		{status => 'success'},
+		'Verify response body is {"status": "success"}'
+	);
```
alexgarel, line 57: we don't want to submit the $request again, we just want to see if it contains parameters we expected !

tests/unit/brevo.t
```diff
@@ -0,0 +1,91 @@
...
+
+	my $ua = LWP::UserAgent->new;
+	my $response = $ua->request($request);
+	print to_json(decode_json($response->decoded_content), {pretty => 1});
+	is_deeply(
+		decode_json($response->decoded_content),
+		{status => 'error'},
+		'Verify response body is {"status": "error"}'
+	);
+	$mocked_ua->unmock_all();
```
alexgarel, line 88: same remark as above, we don't need to submit $request, it was done by add_contact_to_list, we just want to verify it's parameters.

alexgarel (changes requested):
Nearly good, I added comments, I hope they are not too cryptic. Otherwise I can try to fix it for you.

MonalikaPatnaik:
@alexgarel could you pls tell how would I mock $ProductOpener::Brevo::brevo_api_key in my test, I have tried but it is using the actual value defined in the Brevo.pm module only :"

alexgarel:
@MonalikaPatnaik this was not possible to change the brevo_api_key in the module because it was a private variable. So I changed a bit the pattern to use a function so that we can mock it.

At the end I refactored the test a bit to make it less repetitive and easier to read.

Thanks for this PR !

alexgarel (approved):

MonalikaPatnaik:
> @MonalikaPatnaik this was not possible to change the brevo_api_key in the module because it was a private variable. So I changed a bit the pattern to use a function so that we can mock it.
> 
> At the end I refactored the test a bit to make it less repetitive and easier to read.
> 
> Thanks for this PR !

Thankyou so much @alexgarel🙂

teolemon:
Just to be clear @MonalikaPatnaik , there's no UI yet, right ?

MonalikaPatnaik:
> Just to be clear @MonalikaPatnaik , there's no UI yet, right ?

No @teolemon, it's a functionality only

PR #8871, opened 2023-08-16
feat: allow to return JSON response in `/cgi/test_ingredients_analysis.pl`

raphael0202:
with json=1 parameter
This is needed by Robotoff to know which ingredient list is the best among candidates.

### DIFF
--- a/cgi/test_ingredients_analysis.pl
+++ b/cgi/test_ingredients_analysis.pl
@@ -28,6 +28,7 @@ use ProductOpener::Config qw/:all/;
 use ProductOpener::Store qw/:all/;
 use ProductOpener::Index qw/:all/;
 use ProductOpener::Display qw/:all/;
+use ProductOpener::HTTP qw/:all/;
 use ProductOpener::Users qw/:all/;
 use ProductOpener::Lang qw/:all/;
 use ProductOpener::Tags qw/:all/;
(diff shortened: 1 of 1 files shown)


stephanegigandet:
I think it would be better not to create a new API, and instead to use the API v3 to write a "test" product: https://openfoodfacts.github.io/openfoodfacts-server/api/ref-v3/#patch-/api/v3/product/-barcode- (just use "test" for the barcode: it will run the ingredients analysis, but not write the product to the database).

That's what we do for the rate my recipe project: https://github.com/openfoodfacts/rate-my-recipe/blob/main/redux/reducers/recipes.ts#L47

raphael0202:
> I think it would be better not to create a new API, and instead to use the API v3 to write a "test" product: https://openfoodfacts.github.io/openfoodfacts-server/api/ref-v3/#patch-/api/v3/product/-barcode- (just use "test" for the barcode: it will run the ingredients analysis, but not write the product to the database).

I don't really agree on this, writing a test product to run an analysis is a bit hacky and does not follow API best practices. I think it would be much cleaner to use it this way.

raphael0202:
After discussion with the team at the last Product Opener call, we agreed it would be a good idea to implement a standard API for different aspects of Product Opener such as ingredient parsing, but we could use the fake "test" product approach in the meantime. Closing this PR.

PR #8897, opened 2023-08-22
feat: parsing of ingredients with quantity in various units

stephanegigandet:
This PR makes it possible to analyze ingredients lists like "lemon  juice 10cl, sugar 5g, mint 1kg, rum 5L". It is needed in particular for the ratemyrecipe project, but is also useful for some rare product ingredients lists.

In addition we now clearly differentiate percent and quantity in the ingredients structure.

Also includes some refactoring of ingredients_percent.t tests to make it easier to compare diffs and update results.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -78,6 +78,7 @@ BEGIN {
 		&flatten_sub_ingredients
 		&compute_ingredients_tags
 
+		&get_percent_or_quantity_and_normalized_quantity
 		&compute_ingredients_percent_values
 		&init_percent_values
 		&set_percent_min_values
@@ -114,6 +115,7 @@ use ProductOpener::Products qw/:all/;
 use ProductOpener::URL qw/:all/;
 use ProductOpener::Images qw/:all/;
 use ProductOpener::Lang qw/:all/;
+use ProductOpener::Units qw/:all/;
 
 use Encode;
 use Clone qw(clone);
@@ -885,7 +887,7 @@ sub add_specific_ingredients_from_labels ($product_ref) {
 	return;
 }
 
-=head2 parse_specific_ingredients_from_text ( product_ref, $text, $percent_regexp )
+=head2 parse_specific_ingredients_from_text ( product_ref, $text, $percent_or_quantity_regexp )
 
 Lists of ingredients sometime include extra mentions for specific ingredients
 at the end of the ingredients list. e.g. "Prepared with 50g of fruits for 100g of finished product".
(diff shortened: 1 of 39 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -1221,6 +1223,27 @@ sub parse_origins_from_text ($product_ref, $text) {
 	return $text;
 }
 
+=head2 get_percent_or_quantity_and_normalized_quantity($percent_or_quantity_value, $percent_or_quantity_unit)
+
+Used to assign percent or quantity for strings parsed with $percent_or_quantity_regexp.
+
```
alexgarel, line 1229: can you add a small example ?

lib/ProductOpener/Ingredients.pm
```diff
@@ -1290,16 +1313,14 @@ sub parse_ingredients_text ($product_ref) {
...
-		. $min_regexp
-		. '|\s|\.|:)*(\d+(?:(?:\,|\.)\d+)?)\s*(?:\%|g)\s*(?:'
-		. $min_regexp . '|'
-		. $ignore_strings_after_percent
-		. '|\s|\)|\]|\}|\*)*';
+	# Regular expression to find percent or quantities
+	# $percent_or_quantity_regexp has 2 capturing group: one for the number, and one for the % sign or the unit
+	my $percent_or_quantity_regexp = '(?:<|' . $min_regexp . '|\s|\.|:)*' . '(\d+(?:(?:\,|\.)\d+)?)\s*'    # number
+		. '(\%|g|gr|mg|kg|ml|cl|dl|l)\s*'    # % or unit
+		. '(?:' . $min_regexp . '|' . $ignore_strings_after_percent . '|\s|\)|\]|\}|\*)*';
```
alexgarel, line 1320: what about using a ['x' modifier](https://perldoc.perl.org/perlre#/x-and-/xx) to make this kind of expression easier to understand ?

tests/unit/ingredients_percent.t
```diff
@@ -419,701 +72,132 @@ my @tests = (
...
-
-	]
+	],
 
+	# Ingredients in other quantities than grams
+	[
+		'ingredients-in-different-quantities',
+		{
+			lc => 'en',
+			ingredients_text => "lemon 1 KG, orange juice 10cl, sugar 5g, salt 0.5mg, apple juice 1L, ice cream 100ml",
```
alexgarel, line 192: crazy cocktail !

alexgarel (approved):
LGTM !

lib/ProductOpener/Ingredients.pm
stephanegigandet: sure, I added more doc and an example, also I found that I wasn't using this function everywhere I wanted to, so that's fixed.

PR #8898, opened 2023-08-22
taxonomy: Update polish taxonomy

ArturLange:
### What
I added and fixed some missing/wrong Polish translations in taxonomies


benbenben2 (approved):
Well done.
Looks good to me.

PR #8903, opened 2023-08-23
taxonomy: new beer categories and reorganization

DanieliusAsm:
### What

- Reorganized some of the categories under Ale
- Added a LOT of IPA subcategories
- Added: Cocoa capsules
- Added: Scottish, Norwegian, Swedish beers
- new: Non-alcoholic IPA


taxonomies/categories.txt
```diff
@@ -1417,41 +1417,304 @@ agribalyse_proxy_food_name:en:Beer, dark
...
+en:Pale ales, Pale ale
+
+############ IPA ############
+
+<en:Pale ales
 ca:Cervesa IPA, IPA
+en:India Pale Ale (IPA), IPA ales, India pale ale, India pale ales
+es:Cerveza IPA, IPA
+fr:Bières IPA, Bieres IPA, Bière ipa
 xx:India Pale Ale (IPA)
```
stephanegigandet, line 1440: I would suggest to put some synonyms in the xx: definition so that we don't have to repeat them in all languages

```suggestion
xx:India Pale Ale (IPA), India Pale Ale, IPA
```

taxonomies/categories.txt
```diff
@@ -1417,41 +1417,304 @@ agribalyse_proxy_food_name:en:Beer, dark
...
 wikidata:en:Q1543214
 
 <en:India Pale Ale (IPA)
-en:Milkshake IPA, Milkshake India Pale Ale
-wikidata:en:Q103351397
+xx:Amber IPA
```
stephanegigandet, line 1444: For categories, we always put an en: entry as the canonical entry, and then optionally we can add a xx: category that matches all languages

So all xx: entries below should be converted to en:, and we can add a corresponding xx: entry below.

taxonomies/categories.txt
```diff
@@ -1417,41 +1417,304 @@ agribalyse_proxy_food_name:en:Beer, dark
...
+xx:Double IPA, Imperial IPA, Extra IPA, DIPA
+
+<xx:American IPA
+xx:East Coast IPA, East IPA
+
+<xx:American IPA
+xx:Hurricane IPA, Tornado IPA
+
+<xx:American IPA
+xx:New England IPA, NEIPA, Hazy IPA, Juicy IPA, Unfiltered IPA
```
stephanegigandet, line 1465: the hazy / juicy / unfiltered IPA should not be added as synonyms I think, there could be "unfiltered IPAs" from lots of other places

stephanegigandet: On a related note, New England IPA should probably not be under American IPA, it's a description of a style, not an origin.

e.g. this page lists lots of NEIPA brewed in France: https://unepetitemousse.fr/blog/neipa-new-england-ipa/

Issue #8208, opened 2023-03-15
Move to debian bulleye


alexgarel:
We want to do new production install on latest debian.

We should first test it in the Docker images (and on staging). We can also move to official debian image (instead of minideb).

```[tasklist]
### Tasks
- [x] use Debian Bullseye image in Dockerfile
- [x] test stagging with Debian Bullseye
- [ ] https://github.com/openfoodfacts/openfoodfacts-server/issues/8209
```

alexgarel:
It's done. See https://github.com/openfoodfacts/openfoodfacts-server/pull/8234

PR #8911, opened 2023-08-26
feat: Allow scanning more GS1 barcodes

### LINKED ISSUE #8925: Update barcode recognition to understand more barcode types
### What
- Barcode parsing should understand more barcode formats, like DataMatrix and DataBar, and code normalization should be able to extract the GTIN from a GS1-128, GS1 DataMatrix, a GS1 QR Code, a GS1 Data Link, or a GS1 AI data string. 
- This will enable users to use those barcodes, which will become the norm with the GS1 "Sunrise 2027" initiative, in search.

### Part of
- #8333

hangy:
### What

Use the official GS1 parser library to extract more types of barcodes from images, as to support the GS1 "Sunrise 2027" initiative.

- Closes #8925


hangy:
Currently missing: `Barcode::ZBar` doesn't understand DataMatrix codes, and there doesn't seem to be a Perl package that does. Might be worth to create a wrapper around zxing-cpp, or so.

tests/unit/products.t
```diff
@@ -20,6 +20,16 @@ is(normalize_code(' just a simple test 036000291452 here we go '),
...
 	'036000291455', 'should not add leading 0 to cleaned invalid UPC12');
+is(normalize_code('0104044782317112'), '4044782317112', 'should reduce GS1 AI unbracketed string to GTIN');
+is(normalize_code('(01)04044782317112(17)270101'), '4044782317112', 'should reduce GS1 AI bracketed string to GTIN');
+is(normalize_code('^010404478231711217270101'),
+	'4044782317112', 'should reduce GS1 AI unbracketed string with ^ as FNC1 to GTIN');
+is(normalize_code("\x{001d}010404478231711217270101"),
+	'4044782317112', 'should reduce GS1 AI unbracketed string with original FNC1 to GTIN');
+is(normalize_code("\x{241d}010404478231711217270101"),
+	'4044782317112', 'should reduce GS1 AI unbracketed string with GS as FNC1 to GTIN');
+is(normalize_code('https://id.gs1.org/01/04044782317112/22/2A'),
```
stephanegigandet, line 31: Awesome! Could you add a few more tests?
e.g. from GS1 implementation guide:
https://dalgiardino.com/01/09506000134376/10/ABC/21/123456?17=211200
https://dalgiardino.com/01/09506000134369

Maybe some QR code urls that are not GS1 Digital Links, like https://world.openfoodfacts.org/

hangy: Yup, I was adding more tests from the GS1 guide when I noticed that DataMatrix doesn't work with ZBar. This feature probably doesn't have that much use before that, because GS1 DataMatrix seems to be *the* major format, as it's already being used.

teolemon:
@hangy We're having the Product Opener call in 20 minutes if you'd like to present the PR >>> https://meet.google.com/nnw-qswu-hza

tests/unit/products.t
hangy: Added a first test for GS1 DataMatrix. I'll have a look at some additional test cases in the upcoming days.

Dockerfile
```diff
@@ -8,10 +8,11 @@ ARG CPANMOPTS=
 ######################
 # Base modperl image stage
 ######################
-FROM debian:bullseye AS modperl
+FROM debian:bookworm AS modperl
```
hangy, line 11: @alexgarel I noticed that you moved to `bullseye` in #8208 just a few weeks ago. Unfortunately, the `libzxing-dev` version [required by `Imager::zxing`](https://github.com/tonycoz/imager-zxing/blob/fc9ff87e2edf35df3317cf8467f110b17e124910/README#L5-L6) is only available in `bookworm`. If the intention is to stay on `bullseye`, especially if there's still some bare-metal install remaining (is it?), we may have to see how we can achieve that.

tests/unit/products.t
hangy: I added a handful of more tests. There's one case I'm not quite sure how you want to handle it: Even if the new GS1 parser isn't used `normalize_code('http://spam.zip/420')` would return `420`, because of `$code =~ s/\D//g;`. Strictly speaking, keeping that case is probably better for backwards compatibility, but it's probably not a valid GS1 Data Link. 🤷

hangy: @stephanegigandet I tried adding that restriction in baf76d5037b0e18e181486584cfe9e4a3b016ca4, but this change causes some integration tests to fail: https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/6150811346/job/16689935236?pr=8911#step:4:16384 and https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/6150811346/job/16689935236?pr=8911#step:4:19025

By rejecting shorter codes during normalization, would you rather have the limit be removed or the integration tests adjusted?

hangy: BTW: The API checks for `^\d{4,24}$`, so maybe that's a better indicator (or at least more consistent with the current behaviour)?

Dockerfile
tonycoz: After some recent work on Imager::zxing I'd recommend installing zxing-cpp 2.1.0 from source rather than using the bookworm packaged version of the library.

I've had some images (see tonycoz/imager-zxing#2) supplied by another user that causes an assertion failure when trying to parse them, both with git Imager::zxing and with the Debian built ZXingReader tool that's part of the zxing-cpp-tools package.

This assertion was previously reported upstream as zxing-cpp/issues/399 and is apparently fixed in zxing-cpp 2.0.0, though I've only tested 2.1.0.

tests/unit/products.t
stephanegigandet: @hangy It would be good to be consistent, I guess we can keep 4 to 24, and adjust the tests to put a few more digits in the test barcode :)

hangy: Turns out this is a bit more complicated, because `normalize_code`, in the original version, removes everything but digits, so that a theoretically invalid `http://spam.example/id/01/123` would be "normalized" to `0123`. I think I'll postpone this until #8926.

Dockerfile
hangy: Thanks so much for the heads-up, @tonycoz! I prepared using zxing-cpp 2.1.0, but since `debian:bullseye` uses an old version of GCC/G++, C++17 needs to be enabled explicitly. I'll just wait until https://github.com/tonycoz/imager-zxing/commit/829a1293ce8836f46bc93b55feac909abf07737e hits CPAN. 😄

hangy: Done 😄

PR #8912, opened 2023-08-27
taxonomy: adding some german translation for ingredients of More Protein Brownie

FemmeNoire:
added some german translation for ingredients of More Protein Brownie (some still missing, because I could not find an english equivalent) and a second translation for pumpin seed oil, because it was still missing


benbenben2 (approved):
Looks good to me.
Thank you for your contribution.

PR #8931, opened 2023-09-01
taxonomy: add agribalyse property for skyrs and yoghurts

stephanegigandet:
This is so that we can compute the Eco-Score for skyrs.

Also added an agribalyse proxy for the general en:yogurts category.


alexgarel (approved):
LGTM

PR #8921, opened 2023-08-28
feat: add HR packager codes

benbenben2:
### What
Add addresses for Croatian packager codes

PR #8932, opened 2023-09-01
refactor: replace Data::Table by Text::CSV in hr-packagers-refresh.pl

stephanegigandet:
write the CSV file with Text::CSV so that we don't need to install Data::Table


alexgarel:
@benbenben2 and @stephanegigandet is this PR to be closed ?

stephanegigandet:
done in https://github.com/openfoodfacts/openfoodfacts-server/pull/8921

PR #8941, opened 2023-09-02
taxonomy:Romanian categories

DanieliusAsm:
Changes:

I went over to https://ro-en.openfoodfacts.org/categories and added some of the most popular categories (used google translate) that ro users used to categorize products.


stephanegigandet (approved):
Thank you!

PR #8942, opened 2023-09-03
fix: improve ingredients extraction

### LINKED ISSUE #3030: Ignore explanations at end of list of ingredients like ". *Organic Ingredients"
### What
OFF seems not to handle explanations at the end of lists of ingredients like * Organic Ingredients very well. I've seen examples of these quite often here in the US

### Steps to Reproduce
https://world.openfoodfacts.org/product/0850003875088/jalapeno-vegan-cheddar-hippeas
https://world.openfoodfacts.org/product/0099482485375/fried-rice-style-riced-cauliflower-whole-foods-market

### Expected behavior
OFF should ignore a string like `. *` and whatever follows after it. It's mostly just explanations

### Discussion
https://openfoodfacts.slack.com/archives/C06A7LENM/p1584094524048900

### Part of
- #5706

benbenben2:
### What
0850003875088
 - Lngredietns, correct typo,     ok
 - *Organic ingredients, upd taxonomy,     ok

0099482485375
 - text considered ingredients, add to phrases_after_ingredients_list (this same text appears in 272 products), 

0856500004013, 0856500004037, 0029737700144
 - wrong extraction, reeaxtracted, ok

0029737700144
 - non-gmo soybean oil is unknown, added in taxonomy, ok

0029737210070
0073872746109
8006013990644
0008005958876
0008005985179
0008005958043
0008005959101
- contributors question, folksonomy engine. key: ingredient_list:multiple, value: 'nb of ingredients list'

2252559300126
 - *SPICES AND OR VEGETABLE POWDER as last ingredient when that flavor, added to phrases_after_ingredients_list (this text appears in this single product only), ok 

### Question
Is there a way to test the image extraction locally?
Screenshot_20230903_110511

- Fixes #3030

### Comment
Thanks @bredowmax to have collected so many examples. That is of great help.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -3955,7 +3955,9 @@ my %phrases_after_ingredients_list = (
 	],
 
 	en => [
+		'adds a trivial amount',    # e.g. adds a trivial amount of added sugars per serving
 		'after opening',
+		#'Best before',
 		'nutrition(al)? (as sold|facts|information|typical|value[s]?)',
 		# "nutrition advice" seems to appear before ingredients rather than after.
 		# "nutritional" on its own would match the ingredient "nutritional yeast" etc.
@@ -3966,8 +3968,8 @@ my %phrases_after_ingredients_list = (
 		'once opened[,]? (consume|keep|refrigerate|store|use)',
 		'(Storage( instructions)?[: ]+)?Store in a cool[,]? dry place',
 		'(dist(\.)?|distributed|sold)(\&|and|sold| )* (by|exclusively)',
-		#'Best before',
 		#'See bottom of tin',
+		'spices and or vegetable powder as last ingredient',
 	],
 
 	es => [
@@ -4216,7 +4218,7 @@ my %ignore_phrases = (
 		'\d\d?\s?%\sFett\si(\.|,)\s?Tr(\.|,)?',    # 45 % Fett i.Tr.
 		"inklusive",
 	],
-	en => ["na|n/a|not applicable",],
+	en => ["na|n/a|not applicable", "Contains 2% and less of",],    # Contains 2% and less of (Xanthan Gum)
 	fr => ["non applicable|non concerné",],
 
 );
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -16566,6 +16566,9 @@ ciqual_food_name:fr:Huile de soja
 # ingredient/soya-oil has 41836 products 1in 28 languages @2021-08-16
 
 <en:soya oil
+en:non-gmo soybean oil
+
+<en:soya oil
 en:refined soya oil
 pl:rafinowany olej sojowy
 
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -4376,7 +4376,7 @@ nl:Niet geschikt voor kinderen onder 1 jaar
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 #en:description:Labels used for identifying products (and ingredients) that have been grown organically.
 
-en:Organic, organically grown, organically produced, ingredient produced organically, from organic farming, From Organic Agriculture
+en:Organic, organically grown, organically produced, ingredient produced organically, from organic farming, From Organic Agriculture, organic ingredients
 bg:Био, биологично земеделие, биологично
 ca:Orgànic,de cultiu ecologic
 cs:Bio


stephanegigandet (approved):
Thank you @benbenben2 , it looks great, but could you add a test for those in unit/ingredients.t so that we don't break it in the future?

lib/ProductOpener/Ingredients.pm
```diff
@@ -3966,8 +3968,8 @@ my %phrases_after_ingredients_list = (
 		'once opened[,]? (consume|keep|refrigerate|store|use)',
 		'(Storage( instructions)?[: ]+)?Store in a cool[,]? dry place',
 		'(dist(\.)?|distributed|sold)(\&|and|sold| )* (by|exclusively)',
-		#'Best before',
 		#'See bottom of tin',
+		'spices and or vegetable powder as last ingredient',
```
stephanegigandet, line 3972: What's the product with this string?

benbenben2: 2252559300126
See my comment in the main message of the PR. There is a single product with that text after the ingredients list. 
Is it an overkill to have it in the code? Do you prefer that I remove it?

stephanegigandet: I think we should remove it, in fact in this product there's also an "and SPINACH as second ingredient when that flavor -" that we won't recognize either.

benbenben2:
# Tests

> Thank you @benbenben2 , it looks great, but could you add a test for those in unit/ingredients.t so that we don't break it in the future?

@stephanegigandet, yes, that was my question (is there a way to test image ingredients list extraction from image). So, I created tests for the **cut_ingredients_text_for_lang** subroutine (related to picture extraction). They are in a new file called **ingredients_extract.t**.


# %ignore_phrases bug fix
%ignore_phrases were NOT removed from the text. I assume that it was a bug. It is fixed now.
See for example German "inklusive": https://world.openfoodfacts.org/product/5000396000467, also [Mirabelle](https://mirabelle.openfoodfacts.org/products?sql=select+url%2C+code%2C+countries_tags%2C+ingredients_text+from+%5Ball%5D%0D%0AWHERE+ingredients_text+LIKE+%22%25inklusive%25%22+--AND+countries_tags+IN+%28%22en%3Aunited-states%22%29%0D%0AGROUP+BY+random%28%29%0D%0ALIMIT+10%0D%0A)

Additionally, I reviewed and removed 'na|n/a|not applicable' from the list:
-> too much false positives for **na**
-> some false positives for **n/a** (41 values in total)
-> 0 occurrence for **"not applicable"**

I kept those for FR but text of the image is often unrelated to ingredients:
**non applicable** -> 1 occurence
**non concerné** -> 15 occurences

I kept DE phrases as they seem pretty coherent:
**inklusive** -> 29 occurences

lib/ProductOpener/Ingredients.pm
```diff
@@ -4214,10 +4215,10 @@ my %prefixes_before_dash = (fr => ['demi', 'saint',],);
 my %ignore_phrases = (
 	de => [
 		'\d\d?\s?%\sFett\si(\.|,)\s?Tr(\.|,)?',    # 45 % Fett i.Tr.
-		"inklusive",
+		'inklusive',
 	],
-	en => ["na|n/a|not applicable",],
```
stephanegigandet, line 4219: The "not applicable" is needed for some manufacturers that import ingredients list with "not applicable" for some products that don't have ingredients listed (e.g. wine, products that are made of one single ingredient like sugar).
I think we should add it back, with a comment to indicate that it's needed and why.

lib/ProductOpener/Ingredients.pm
```diff
@@ -4214,10 +4215,10 @@ my %prefixes_before_dash = (fr => ['demi', 'saint',],);
...
-	fr => ["non applicable|non concerné",],
+	en => ['contains 2% or less of:',],    # Contains 2% or less of (Xanthan Gum)
```
stephanegigandet, line 4220: We should keep the "contains 2% or less of:", it's an important part of the ingredients list. We don't parse it correctly yet, but we do not want to remove it from ingredients lists. Instead we can ignore it during parsing (or better, add support for it to set the percent_min property), and keep it in the list of ingredients.

tests/unit/ingredients_extract.t
```diff
@@ -0,0 +1,42 @@
...
+use ProductOpener::Products qw/:all/;
+use ProductOpener::Tags qw/:all/;
+use ProductOpener::TagsEntries qw/:all/;
+use ProductOpener::Ingredients qw/:all/;
+
+my @lists = (
+	# en phrases_after_ingredients_list
+	[
+		"en",
+		"carrots, green peas, corn, scallion. *adds a trivial amount of added sugars per serving.",
```
stephanegigandet, line 20: Thanks a lot for adding tests!

tests/unit/ingredients_extract.t
```diff
@@ -0,0 +1,42 @@
...
+		"carrots, green peas, corn, scallion.",
+	],
+	# en ignore_phrases,
+	[
+		"en",
+		"Egg White, Contains 2% or less of: Xanthan Gum, Salt, Glucono-delta-lactone.",
```
stephanegigandet, line 26: I think we should keep this test, but that we should keep the "Contains 2%..." part.

stephanegigandet (approved):
Thank you!

alexgarel:
Amazing @benbenben2 ! :muscle:

PR #8946, opened 2023-09-04
fix: delete_user_test should not hang indefinitely 

### LINKED ISSUE #8944: delete_user.t integration test hangs indefinitely
the delete_user.t test never finishes when running it locally in docker

ok 1 - the delete button does exist
ok 2 - the account was deleted
Initializing Minion backend configured in lib/ProductOpener/Config2.pm
(test hangs here)

stephanegigandet:
fixes #8944

### DIFF
--- a/lib/ProductOpener/APITest.pm
+++ b/lib/ProductOpener/APITest.pm
@@ -812,8 +812,6 @@ sub get_minion_jobs ($task_name, $created_after_ts, $max_waiting_time) {
 				my $job_state = $job->{state};
 				# check if the job is done
 				if (($job_state eq "active") or ($job_state eq "inactive")) {
-					print STDERR
-						"task $task_name - job $job_id - state $job_state - waiting 2 - (waited: $waited, max_waiting_time: $max_waiting_time)\n";
 					$jobs_complete = 0;
 					sleep(2);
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/APITest.pm
```diff
@@ -787,42 +787,42 @@ The timestamp of the creation of the task
...
-				my $job = get_minion()->job($job_id);
 				# retrieving the job state
-				my $job_state = $job->info->{state};
-				# waiting the job to be done
-				while (($job_state eq "active") or ($job_state eq "inactive")) {
+				my $job_state = $job->{state};
+				# check if the job is done
+				if (($job_state eq "active") or ($job_state eq "inactive")) {
+					print STDERR
+						"task $task_name - job $job_id - state $job_state - waiting 2 - (waited: $waited, max_waiting_time: $max_waiting_time)\n";
```
alexgarel, line 816: to be removed ?
```suggestion
```

alexgarel (approved):
LGTM, I think we did the loop the way we did it before because there was the fear, that the first call to `get_minion()->jobs` might not return all created job (because of race condition), but I think it's might not be justified.
I think we can try it this way and see if it works consistently.

PR #8951, opened 2023-09-05
ci: wait docker images to avoid failing in container deploy

alexgarel:
It happens we fail because the container image is not immediately accessible in ghcr.io. ([like this morning](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/6081527840/job/16497359812).
Add a step to wait for images availability.

### DIFF
--- a/.github/workflows/container-deploy.yml
+++ b/.github/workflows/container-deploy.yml
@@ -175,6 +175,36 @@ jobs:
           make create_external_volumes && \
           make create_external_networks
 
+    - name: wait docker images
+      # we sometimes have the images not ready immediately after build, so we prefer to wait for it
+      uses: appleboy/ssh-action@master
+      with:
+        host: ${{ env.SSH_HOST }}
+        username: ${{ env.SSH_USERNAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/container-deploy.yml
```diff
@@ -175,6 +175,36 @@ jobs:
           make create_external_volumes && \
           make create_external_networks
 
+    - name: wait docker images
```
teolemon, line 178:
```suggestion
    - name: wait for docker images to be available
```

.github/workflows/container-deploy.yml
```diff
@@ -175,6 +175,36 @@ jobs:
...
+              # timeout after 6 minutes (gloablly)
+              if [[ $waited -gt 360 ]]; then
+                echo "Timeout waiting for image $image_name"
+                # Abort the job
+                exit 1
+              fi
+            done
+          done
+
     - name: init backend
```
teolemon, line 208:
```suggestion
    - name: initialize the backend
```

PR #8959, opened 2023-09-05
Taxonomy: correction

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -42896,7 +42896,7 @@ ciqual_food_name:fr:Yaourt, lait fermenté ou spécialité laitière, nature, su
 
 <fr:Spécialités laitières
 <en:Plain fermented dairy desserts
-fr:Spécialités laitières nature non sucrées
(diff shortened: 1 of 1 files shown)


taxonomies/categories.txt
```diff
@@ -42896,7 +42896,7 @@ ciqual_food_name:fr:Yaourt, lait fermenté ou spécialité laitière, nature, su
 
 <fr:Spécialités laitières
 <en:Plain fermented dairy desserts
-fr:Spécialités laitières nature non sucrées
+fr:Spécialités laitières nature non sucrées, Spécialité laitière nature
```
stephanegigandet, line 42899: shouldn't "Spécialité laitière nature" be the main name in fr?

moon-rabbitOFF: Yes it could, I just want it to be explicite that is it without sugar. In Spain you can have "natural" yogurts that are sweetened, but in France I don't think it is the case. I changed it

stephanegigandet (approved):

PR #8965, opened 2023-09-06
fix: use "Plus Jakarta Sans" font for title

raphael0202:
I'm integrating the mobile landing page, and I noticed we don't use Plus Jakarta Sans for titles.
According to brand guidelines, Plus Jakarta Sans should be used for all titles

Before/after:
Capture d’écran du 2023-09-06 16-07-08
Capture d’écran du 2023-09-06 16-06-19
Capture d’écran du 2023-09-06 16-07-53
Capture d’écran du 2023-09-06 16-08-38


stephanegigandet:
Also there are some things in Jakarta Sans that are just ugly, the "fi" ligature in particular is terrible IMHO.

image

I would be in favor of changing the brand guidelines to pick another more universal font.

stephanegigandet:
on desktop, the browser picks another font for Greek and Ukrainian (better than on Flutter were apparently glyphs were not displayed at all if missing from the font..)

image

raphael0202:
I noticed the fallback for cyrillic languages, but indeed the fi ligatures are not pretty...
As we need these fonts for the landing pages, we can just add the font loading code.

raphael0202:
I'm closing this PR, I will open a new one with the necessary changes for the landing page :)

PR #8976, opened 2023-09-08
refactor: GS1 XML to JSON conversion + tests

stephanegigandet:
We have 2 scripts to convert GS1 XML to JSON, that create slightly different JSON files. The GS1.pm module should be able to handle the differences.

See this commit to see sample differences: https://github.com/openfoodfacts/openfoodfacts-server/pull/8976/commits/adb696d4d5ed6098f254f8a6a9537f727aa4311a#diff-e203f2af6fbc612b40c57a0439aa4145468d4c9028b462a084c8b77dd67d27a8

Added some tests to verify that.


tests/unit/convert_gs1_xml_to_json.t
```diff
@@ -0,0 +1,74 @@
...
+	next if $file !~ /\.xml$/;
+	my $file = $`;    # remove xml extension
+	my $json_file = "$input_dir/$file.json";
+
+	convert_gs1_xml_file_to_json("$input_dir/$file.xml", $json_file);
+
+	if (open(my $json_fh, "<:encoding(UTF-8)", $json_file)) {
+		local $/;    #Enable 'slurp' mode
+		my $json = JSON->new->allow_nonref->canonical;
+		my $json_ref = $json->decode(<$json_fh>);
```
alexgarel, line 34: Personally instead of using slurp mode, I prefer to use join:
```suggestion
		my $json = JSON->new->allow_nonref->canonical;
		my $json_ref = $json->decode(join("", <$json_fh>));
```

tests/unit/convert_gs1_xml_to_json.t
```diff
@@ -0,0 +1,74 @@
...
+
+my $input_dir = "$test_dir/inputs/$test_id";
+
+opendir(my $dh, $input_dir) or die("Could not open the $input_dir directory: $!\n");
+
+foreach my $file (sort(readdir($dh))) {
+
+	next if $file !~ /\.xml$/;
+	my $file = $`;    # remove xml extension
+	my $json_file = "$input_dir/$file.json";
```
alexgarel, line 27: You should instead use a File::Temp->newdir(); and put you files there (see usage in other tests)

scripts/prettyfy_json_files_in_dir.pl
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/perl -w
```
alexgarel, line 1: Can we start to put this kind of file in a "utils" subfolder ?

alexgarel (approved):
Great !

tests/unit/convert_gs1_xml_to_json.t
stephanegigandet: ok.

scripts/prettyfy_json_files_in_dir.pl
stephanegigandet: Good idea

PR #9003, opened 2023-09-14
fix: Make contries map responsive

### LINKED ISSUE #8997: The map on facets should be responsive
### What
- The map on facets should be responsive

hangy:
### What

Replaced `jvectormap` with `jsvectormap`, made the map have a min and max size, and restyled the map. Also moved the map code to a separate JS file, so that it can be linted, and edited more easily, and added separate RTL and LTR formats for tooltips.

### Screenshot

- Map at min width:
  image
- Map at max width:
  image

- Fixes #8997


stephanegigandet (approved):
Perfect, thank you!

PR #9006, opened 2023-09-15
taxonomy: add Frozen and Canned to Meals with meat

OpenChris:
add 'Frozen' and 'Canned' subcategories to 'Meals with meat'


stephanegigandet (approved):
Thanks!

PR #9009, opened 2023-09-16
fix: Use images subdomain for images

hangy:
### What

Use the image subdomain in more places where the current subdomain is currently used.

- Related to #8994


stephanegigandet (approved):
Thank you!

PR #9010, opened 2023-09-16
taxonomy: Bulgarian additions

Roto15:


stephanegigandet (approved):
Thank you!

taxonomies/categories.txt
```diff
@@ -33586,6 +33586,7 @@ es:Regañás, Regañá
 wikidata:en:Q6103703
 
 en:Breakfasts
+bg:Закуски
```
stephanegigandet, line 33589: @Roto15: this word is already used for en:snacks, we can't have the same translation for 2 different entries, could you try to find another translation for one of them?

PR #9014, opened 2023-09-17
feat: Allow GS1 barcode types to be used in search and read API

### LINKED ISSUE #9013: Allow GS1 barcode types to be used in search
## Proposal

GS1 barcode types like AI Strings, or GS1 Data URI should be usable in search and API.

### Part of

- #8333

hangy:
### What

Allows using more barcode types in search and READ API

- Fixes #9013
- Depends on #8911


tests/unit/routing.t
```diff
@@ -145,7 +145,59 @@ my @tests = (
...
+			'api_action' => 'product',
+			'api_method' => undef,
+			'api_version' => '3',
+			'cc' => 'world',
+			'lc' => 'en',
+			'original_query_string' =>
+				'api/v3/product/https%3A%2F%2Fid.gs1.org%2F01%2F03564703999971%2F10%2FABC%2F21%2F123456%3F17%3D211200',
+			'query_string' =>
+				'api/v3/product/https%3A%2F%2Fid.gs1.org%2F01%2F03564703999971%2F10%2FABC%2F21%2F123456%3F17%3D211200',
+			'code' => 'https://id.gs1.org/01/03564703999971/10/ABC/21/123456?17=211200',
```
hangy, line 195: This case is bugging me a bit right now, because GS1 Data URI and `?rev`-parameters (or other API parameters) don't seem to mix too well. Hence the draft status of the PR.

stephanegigandet:
This looks great, thanks for adding all the test cases etc. :)

I'm trying to build zxing on stretch (what we have in prod today), no success yet, I'll try harder.

stephanegigandet:
I tested it in docker, it works great. :)

In production though, as we are still on stretch, I'm unable to build zxing. It needs cmake 3.15 and gcc 7, while we have cmake 3.7.2 and gcc 6.3... 

So we would have to wait for the migration of off1 to merge and deploy this unfortunately.. :(

hangy:
> I tested it in docker, it works great. :)

Nice! 😄 

> In production though, as we are still on stretch, I'm unable to build zxing. It needs cmake 3.15 and gcc 7, while we have cmake 3.7.2 and gcc 6.3...

We could try building zxing-cpp as a static lib, bundle that as a `.deb`, and try using that? I don't have too much experience with building Debian packages, but `cpack -G DEB` seems to handle a lot of stuff for you.

PR #9020, opened 2023-09-18
fix: avoid spam user creation

alexgarel:
A simple spam user creation protection


lib/ProductOpener/Users.pm
```diff
@@ -329,6 +330,13 @@ sub check_user_org ($user_ref, $new_org_id) {
 	return;
 }
 
+sub is_supsicious_name ($value) {
```
raphael0202, line 333:
```suggestion
sub is_suspicious_name ($value) {
```

tests/unit/spam_user.t
```diff
@@ -0,0 +1,19 @@
...
+use Test::More;
+use Log::Any::Adapter 'TAP';
+
+use ProductOpener::Users qw/:all/;
+
+foreach my $name ("click here", "go forms.yandex.ru", "https://test.me") {
+	ok(is_supsicious_name($name), "$name is a suspicious name");
+}
+
+foreach my $name ("danilo.go", 'roberto@test.me', "Lalilou", "Dan stanley") {
```
raphael0202, line 14: Currently email addresses such as roberto@test.me should trigger the suspicious detection, no?

lib/ProductOpener/Users.pm
```diff
@@ -329,6 +330,13 @@ sub check_user_org ($user_ref, $new_org_id) {
...
+	# email or xxx.nunsrt
+	my $email_re = qr/^[\w_.+]+(?:@[\w._+]+)?$/;
```
raphael0202, line 335: Shouldn't we have an URL detector too? (cf the unit test)

raphael0202 (approved):
LGTM about the honeypot, I added one remark about URL detection

tests/unit/spam_user.t
alexgarel: For me it's another problem. I mean for an email we should return a real error to the user (not a cryptic one)

lib/ProductOpener/Users.pm
```diff
@@ -329,6 +330,13 @@ sub check_user_org ($user_ref, $new_org_id) {
 	return;
 }
 
+sub is_suspicious_name ($value) {
+	# email or xxx.nunsrt
+	my $email_re = qr/^[\w_.+]+(?:@[\w._+]+)?$/;
+	my $invite_re = qr/(?:click here|wants to meet you|:\/\/|\.\w{2,3}\b)/i;
```
stephanegigandet, line 336: I'm not sure about the dot + 2 alphanumeric characters, that seems very broad. It would be good to check all existing accounts that match it, to see if there are false positives.

Maybe change \w to [a-z] to limit a bit the possibilities.

alexgarel: this is because "dot + 2 alphanumeric characters" is a bit wide that I added email_re which removes more than emails, but also things where there is only one word.

alexgarel: ok for [a-z]

PR #9021, opened 2023-09-18
fix: limit the number of fields fetched from MongoDB

raphael0202:
It does not solve issue #8792 as we're still over Memcached limit for main page, but it should still be faster (payload went from 30MB to 3MB). Without images, the payload is 2.3MB, so even while rearranging images we potentially need to increase memcached limit.
Furthermore, knowing which fields to fetched to compute attributes will be necessary once search-a-licious project is deployed, as data will be fetched from Elasticsearch (and not MongoDB).

It was tested locally, all attributes work as expected (a double check should be done once merged in staging).

### DIFF
--- a/lib/ProductOpener/Attributes.pm
+++ b/lib/ProductOpener/Attributes.pm
@@ -31,6 +31,10 @@ the same structured format for all attributes.
 
 See https://wiki.openfoodfacts.org/Product_Attributes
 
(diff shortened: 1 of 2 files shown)


alexgarel (approved):
LGTM
good move.

stephanegigandet (approved):
It's working great, thank you!

PR #9022, opened 2023-09-18
fix: Mark script/css as base layout

### LINKED ISSUE #8799: All JS scripts get executed twice due to robotoff-asker
### What
When analyzing XHR queries made on Open Food Facts, I noticed that some XHR queries were performed twice (`/preferences`, `/attribute_groups`). The culprit is robotoff-asker that clones every script and add it inside its HTML node:
```javascript
    const scripts = document.querySelectorAll('script');
    for (let i = 0; i < scripts.length; ++i) {
      const script = scripts[i];
      content.appendChild(script.cloneNode(true));
    }
```

In the #1705 PR, it seems that we cannot use JS and CSS from the main page and need to copy it, but we definitely don't need custom scripts.

I understand why we need CSS, but do we even need these scripts?
poke @hangy

hangy:
### What

Add a `data-`-attribute to some `script` and `link` tags, so that `robotoff-asker` can reduce the amount of JS/CSS run.

- Fixes #8799

### DIFF
--- a/html/js/display-product.js
+++ b/html/js/display-product.js
@@ -109,7 +109,7 @@ class RobotoffAsker extends HTMLElement {
     this.question = this.questions ? this.questions.pop() : null;
     if (!this.question) {
       const serverDomain = getServerDomain();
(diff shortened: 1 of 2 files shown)


raphael0202 (approved):
Thank you! :+1:

PR #9028, opened 2023-09-19
fix: Add "kan" to the NL ingredients taxonomy

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]


lib/ProductOpener/Ingredients.pm
```diff
@@ -196,7 +196,7 @@ my %may_contain_regexps = (
 	lt => "sudėtyje gali būti",
 	lv => "var saturēt",
 	nl =>
-		"Dit product kan sporen van|bevat mogelijk sporen van|Kan sporen bevatten van|Kan sporen van|bevat mogelijk|sporen van",
+		"Dit product kan sporen van|bevat mogelijk sporen van|Kan sporen bevatten van|Kan sporen van|bevat mogelijk|sporen van|kan",
```
stephanegigandet, line 199: Sorry for my late reply, do you remember which product had that? It would be useful to add a test for it.
cc our Dutch expert @aleene

aleene: Using just one word "kan" is very open. The word might be used for other things. Would it be possible to check that first?

benbenben2: The word "kan " (followed by a space) is not found in the taxonomies file. At least from this side, not false positives.

aleene:
Kan is like can in english

benbenben2:
Kan gluten, melk, ei, selderij, soja, mosterd bevatten,
Kan ei, soja, sel erd bevatten
Kan noten en pinda's bevatten. 
Kan noten bevatten.

It is already handled.


This query list ~100 products (not all are "kan .... bevatten", there is also "kan bevatten: ..." which is also already handled)
[mirabelle](https://mirabelle.openfoodfacts.org/products?sql=select+url%2C+ingredients_text++from+%5Ball%5D+%0D%0AWHERE+ingredients_text+LIKE+%22%25kan+%25%22+AND+ingredients_text+NOT+LIKE+%22%25kan+sporen%25%22+%0D%0AAND+countries_tags+LIKE+%22%25en%3Anetherlands%25%22+%0D%0AORDER+BY+random%28%29+LIMIT+10)


Screenshot_20231221_214007


We don't need to add "kan" @moon-rabbitOFF. If you find a counter-example, let us know. I will make suggestion on your PR.

teolemon:
Should we close then ? @moon-rabbitOFF @benbenben2 @aleene ?

stephanegigandet (approved):

stephanegigandet:
Closing as it seems it is not needed.

PR #9030, opened 2023-09-19
taxonomy: German synonym for antioxydant

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/taxonomies/additives_classes.txt
+++ b/taxonomies/additives_classes.txt
@@ -211,7 +211,7 @@ bg: Антиоксидант, антиоксиданти
 ca: Antioxidant, antioxidants, agent antienfosquiment
 cs: Antioxidant, antioxidanty
 da: Antioxidant, Antioxidanter
-de: Antioxidant, Antioxidations-mittel, Antioxidationsmittel, Antioxidantien, Antioxidanzien
+de: Antioxidant, Antioxidations-mittel, Antioxidationsmittel, Antioxidantsmittel, Antioxidantien, Antioxidanzien
 el: Αντιοξειδωτικό
 es: Antioxidante, antipardeamiento, sinérgicos de antioxidantes, antioxidantes
 et: Antioksüdant


taxonomies/additives_classes.txt
```diff
@@ -211,7 +211,7 @@ bg: Антиоксидант, антиоксиданти
 ca: Antioxidant, antioxidants, agent antienfosquiment
 cs: Antioxidant, antioxidanty
 da: Antioxidant, Antioxidanter
-de: Antioxidant, Antioxidations-mittel, Antioxidationsmittel, Antioxidantien, Antioxidanzien
+de: Antioxidant, Antioxidations-mittel, Antioxidationsmittel, Antioxidantsmittel, Antioxidantien, Antioxidanzien
```
hangy, line 214: I haven't heard this before, and Duden doesn't know it. I can't find too many Google matches, either. Do we have a lot of these typos on OFF?

stephanegigandet: If it's a typo, we should not add it as a synonym.

We do have 2 products with it, and it is indeed written on the package though: https://de.openfoodfacts.org/ingredient/Antioxidantsmittel

Is it a mispelling, or an actual synonym?

hangy: For the Rotweinessig, the text didn't match whats written on the label.

https://de.openfoodfacts.org/images/products/541/118/813/7955/3.jpg says "Antioxidationsmittel", https://de.openfoodfacts.org/images/products/541/118/813/7955/5.jpg says "Antioxidantsmittel", but I can't find that in a lot of sources or dictionaries. I think the producer might have created an incorrect label here. 😅

moon-rabbitOFF: The word seems to exist (I can find it on some German websites), but is not common. Alpro is not a German brand, so I guess they didn't choose the best translation

stephanegigandet: Indeed in Google there seems to be a number of product ingredients lists that have Antioxidantsmittel. In practice it does not hurt to add it, even if it's very uncommon.

stephanegigandet (approved):

PR #9046, opened 2023-09-21
taxonomy: add animal welfare claims on days of grazing for cows and beef

OpenChris:
Proposal to recognize 2 animal welfare claims:
 120 days grass-fed (pâturage)
 150 days grass-fed (pâturage)
NB : these can apply to cows (milk) but also to beef (meat)


taxonomies/labels.txt
```diff
@@ -19356,6 +19356,12 @@ nl:Kippenwelzijn
 fr:Lait de paturage
 nl:Weidemelk
 
+<en:Animal welfare
+fr:120 jours de pâturage
```
stephanegigandet, line 19360: Could you also add the English translations? it's easier if we always have an English name for all (or most) entries

stephanegigandet (approved):
Thank you!

PR #9057, opened 2023-09-22
taxonomy: add Norwegian synonyms for tarragon

tkbremnes:
### What

Added Norwegian synonym for tarragon.


stephanegigandet (approved):
Thanks

PR #9062, opened 2023-09-24
FIX #9033 Changed Username to UserId using replace

### LINKED ISSUE #9033: Change "user name" for user id
In the signin form we use "Name" and "User name" while the second really is a user id (and is referred as such in the code).

It makes hard to describe field in privacy policy and also for users to understand.

Change forms to say "user id"

Codeioholic:

- Fixes #[ISSUE NUMBER]


stephanegigandet:
@Codeioholic we should change only the handful of strings we actually show to users, not all the completely unrelated usernames for github config files etc. It's a good thing to grep to see where we might need to change some strings, but then we should only change the ones that are actually shown (e.g. in translation files in /po/common/ )

PR #9063, opened 2023-09-24
fix: error and warning for vegan label with non-vegan ingredients

### LINKED ISSUE #5249: Non-vegan product marked as vegan
### What

https://world-fr.openfoodfacts.org/product/8718265015494/groente-loempia-albert-heijn

### Screenshot
image

### User report
Johan Vandingenen Hi, i noticed there’s a product marked as vegan (by analysis) but it contains egg. Is it a bug? Or is it a human error? https://world.openfoodfacts.org/product/8718265015494/groente-loempia-albert-heijn

### Part of
- #5706

benbenben2:
### What
Added data quality errors:
- vegan label but non-vegan ingredients
- vegetarian label but non-vegetarian ingredients

Added warning:
- vegan label but could not confirm for some ingredients
- vegetarian label but could not confirm for some ingredients

### Discussion
Question mark about the warning, because of these 2 cases:
- some ingredients (like thickener, for example) do not have the key-value vegan and vegetarian) -> but it can be additives classes that could contain only vegan/vegetarian additives. false-positive in this case (example: https://world.openfoodfacts.org/api/v2/product/3263859506216.json)
- key-value vegan and vegetarian can be yes, no or maybe. For this latter, we could also generate a warning.

These two would need further investigations. Maybe, we can start with this error and the warning as it is now, and come back to it in the future.

### Screenshot
Screenshot_20230923_230032

- Fixes #5249

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -1384,36 +1384,6 @@ Vegetarian label: check that there is no non-vegetarian ingredient.
 =cut
 
 sub check_labels ($product_ref) {
-	# if (defined $product_ref->{labels_tags} && has_tag($product_ref, "labels", "en:vegan")) {
-	# 	if (defined $product_ref->{ingredients}) {
-	# 		my @ingredients = @{$product_ref->{ingredients}};
-
-	# 		while (@ingredients) {
-
-	# 			# Remove and process the first ingredient
-	# 			my $ingredient_ref = shift @ingredients;
-	# 			my $ingredientid = $ingredient_ref->{id};
-
-	# 			# Add sub-ingredients at the beginning of the ingredients array
-	# 			if (defined $ingredient_ref->{ingredients}) {
-
-	# 				unshift @ingredients, @{$ingredient_ref->{ingredients}};
-	# 			}
-
-	# 			if (defined $ingredient_ref->{"vegan"}) {
-	# 				if ($ingredient_ref->{"vegan"} eq 'no') {
-	# 					push @{$product_ref->{data_quality_errors_tags}}, "en:vegan-label-but-non-vegan-ingredient";
-	# 				}
-	# 				# else 'yes', 'maybe'
-	# 			}
-	# 			else {
-	# 				push @{$product_ref->{data_quality_warnings_tags}},
-	# 					"en:vegan-label-but-could-not-confirm-for-all-ingredients";
-	# 			}
-	# 		}
-	# 	}
-	# }
-
 	# this also include en:vegan that is a child of en:vegetarian
 	if (defined $product_ref->{labels_tags} && has_tag($product_ref, "labels", "en:vegetarian")) {
 		if (defined $product_ref->{ingredients}) {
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -1373,6 +1373,69 @@ sub check_categories ($product_ref) {
...
+Checks related to specific product labels.
+
+Vegan label: check that there is no non-vegan ingredient.
+
+Vegetarian label: check that there is no non-vegetarian ingredient.
+
+=cut
+
+sub check_labels ($product_ref) {
+	if (defined $product_ref->{labels_tags} && has_tag($product_ref, "labels", "en:vegan")) {
```
stephanegigandet, line 1387: Here you check for the vegan label, but then below at line 1420 you have data quality warnings for the vegetarian label

stephanegigandet:
> * some ingredients (like thickener, for example) do not have the key-value vegan and vegetarian) -> but it can be additives classes that could contain only vegan/vegetarian additives. false-positive in this case (example: https://world.openfoodfacts.org/api/v2/product/3263859506216.json)

We have code like this in Ingredients.pm for the additives classes:

```
						# additives classes in ingredients are functions of a more specific ingredient
						# if we don't have a property value for the ingredient class
						# then ignore the additive class instead of considering the property undef
						elsif (exists_taxonomy_tag("additives_classes", $ingredientid)) {
							$value = "ignore";
							#$ingredient_ref->{$property} = $value;
						}
```

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -1373,6 +1373,100 @@ sub check_categories ($product_ref) {
...
+				}
+
+				# some additives_classes (like thickener, for example) do not have the key-value vegan and vegetarian
+				# it it can be additives_classes that contain only vegan/vegetarian additives.
+				# to avoid false-positive - instead of raising a warning (else below) we ignore additives_classes
+				if (!exists_taxonomy_tag("additives_classes", $ingredientid)) {
+					# vegan
+					if (defined $ingredient_ref->{"vegan"}) {
+						if ($ingredient_ref->{"vegan"} eq 'no') {
+							push @{$product_ref->{data_quality_errors_tags}}, "en:vegan-label-but-non-vegan-ingredient";
```
stephanegigandet, line 1441: The product could have th en:vegetarian label, not the en:vegan label

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -1373,6 +1373,103 @@ sub check_categories ($product_ref) {
...
+Checks related to specific product labels.
+
+Vegan label: check that there is no non-vegan ingredient.
+
+Vegetarian label: check that there is no non-vegetarian ingredient.
+
+=cut
+
+sub check_labels ($product_ref) {
+	# if (defined $product_ref->{labels_tags} && has_tag($product_ref, "labels", "en:vegan")) {
```
stephanegigandet, line 1387: Can you remove the commented out code?

stephanegigandet (approved):
Looks good. I tested it on ~10k products and got those results:

image

It's definitely useful, I reviewed a few cases and found some real issues we have:
- in the additives taxonomy: E1105 lyzozyme can be vegetarian (we have vegetarian:en:no for it)
- https://fr.openfoodfacts.org/produit/3760020500658/pate-a-tartiner-chocolat-noisette-jardin-bio should not have the vegan label

PR #9074, opened 2023-09-26
fix: Nova 4 marker for mechanicaly separated meat and fruit juice concentrates + better Halal / Kosher parsing

stephanegigandet:
This PR adds entries for mechanically separated meat, and makes it a NOVA 4 marker, along with fruit juice concentrates, as they are listed on https://www.fao.org/3/ca5644en/ca5644en.pdf (NOVA paper from Carlos Monteiro 2019).

We have ingredients processing for "mechanically separated" and "concentrated", and we currently do not support NOVA 4 markers for processings, so this is currently not perfect. We can continue to improve it and add support for ingredients processing markers (and keeping an eye on potential false positives that could come from "concentrated" ingredients that are not fruit juices).

This PR also adds support for Halal and Kosher labels, so that we can parse things like "viande séparée mécaniquement de poulet halal"


raphael0202 (approved):

raphael0202:
There is a merge conflict with `ingredients.t`, but otherwise looks good!

PR #9082, opened 2023-09-27
taxonomy: add proxy for Tête de Moine as it's similar to Abondance

OpenChris:
taxonomy updates for categories:
- add proxy for Tête de Moine as it's similar to Abondance type of cheese
- proxy whole duck foie gras to duck raw liver
- english translation


stephanegigandet (approved):
Thank you

PR #9089, opened 2023-09-29
refactor: make ingredients percent estimation available as a product service / microservice

stephanegigandet:
This PR is a proposal to gradually introduce a microservice architecture in Product Opener. The idea is to define a generic API for all product operations (e.g. ingredients list parsing, ingredients percentage estimation, Nutri-Score computation) that can operate on a product object.

Each operation takes as input specific fields on the product object (e.g. "ingredients_text_*" for ingredients list parsing, and creates or updates specific output fields (e.g. the "ingredients" structure for ingredients list parsing). I think it's simple to keep the product object as the generic input and output structure. We lose a bit of separation of concerns (e.g. for ingredients list parsing we could force passing only a list of ingredient, and getting back only a structured list of ingredients), but it simplifies everything, and it makes chaining services much easier (e.g. take that ingredient list, parse it, estimate the ingredients percent, estimate the nutrition facts, and then compute the Nutri-Score).

The approach is iterative: first we just make available some of those operations, so that they can be run with an API. But this opens the way to then replace or extend the current product operations coded in Product Opener by services developed and executed outside of Product Opener.

I have a very concrete first use case in mind: the estimation of the percentages of each ingredients.

The [first task of the project](https://docs.google.com/document/d/1cPEEhl9Ty0feKEel2NxUmscRR6oWZTLJMpTSbLbzB5U/edit#heading=h.k8512q7c8p7) is to create a test framework. Instead of strongly tying it to the Product Opener code, I'm thinking we could make it completely independent and put it in another repo (e.g. we can create a metrics framework in this repo: https://github.com/openfoodfacts/recipe-estimator ). I'm thinking to code it in Python. My plan is to have golden sets of products with some or all percentages specified, and have a framework that removes those specified percentages, that then calls multiple APIs to estimate the percentages. After that we can measure the accuracy of those methods.

With the proposed microservice architecture, we have a /api/v3/product_services/estimate_ingredients_percent endpoint that would allow us to test the current very basic estimation we have inside Product Opener.

Then hopefully we will be able to get much better results with another approach (e.g. the linear solver), that is fast enough to be run synchronously in realtime for each product update. It's very likely that this other approach won't be coded in Perl (e.g. better linear solving libraries in Python). We could then just have Product Opener call an HTTP service for the the new API instead of the internal function.


lib/ProductOpener/APIProductServices.pm
```diff
@@ -0,0 +1,180 @@
...
+and get back resulting product data (possibly filtered to get only specific fields back).
+
+=head2 INTERFACE
+
+=head3 Request
+
+The Routing.pm and API.pm module offer an HTTP interface of this form:
+POST /api/v3/product_services/[comma separated list of services]
+
+The POST body is a JSON hash with those fields:
```
hangy, line 41: Probably a mix-up of Perl and JavaScript terms; In JSON it's probably not a hash, but an object.

lib/ProductOpener/APIProductServices.pm
```diff
@@ -0,0 +1,180 @@
...
+
+=head4 product
+
+A product object
+
+=head4 field
+
+A comma separated list of fields to return. If empty, all available fields
```
hangy, line 49: Could this also be an array instead of a comma separated list?

```json
{
    "fields": ["code", "categories"]
}
```

will probably feel more natural to people working with JSON than

```json
{
    "fields": "code,categories"
}
```

lib/ProductOpener/APIProductServices.pm
```diff
@@ -0,0 +1,180 @@
...
+ask for one or more services to be executed on the input product data
+(for instance computing the min, max and estimated percentages of each ingredient),
+and get back resulting product data (possibly filtered to get only specific fields back).
+
+=head2 INTERFACE
+
+=head3 Request
+
+The Routing.pm and API.pm module offer an HTTP interface of this form:
+POST /api/v3/product_services/[comma separated list of services]
```
hangy, line 39: Just a personal observation: Commas in the path to designate separate services doesn't feel that natural *to me*. A lot of frameworks allow multiple values for a single parameter, so that something like `/api/v3/product_services?service=echo&service=estimate_ingredients_percent` could work.

lib/ProductOpener/APIProductServices.pm
```diff
@@ -0,0 +1,180 @@
...
+A product object
+
+=head4 field
+
+A comma separated list of fields to return. If empty, all available fields
+(input fields + any fields added or changed by the service) are returned.
+
+=head3 Response
+
+The response is in the JSON API v3 response format, with a resulting product object.
```
hangy, line 54: It could also be an error, possibly with a specific HTTP status code? Or is the possibility of something being an error defined globally? (In OpenAPI, it's usually per operation, IIRC)

lib/ProductOpener/APIProductServices.pm
stephanegigandet: Good idea.

lib/ProductOpener/APIProductServices.pm
stephanegigandet: Yes, in API v3, the response always get a status, errors and warnings in the same format (for all operations): https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/api/ref/responses/response-status/response_status.yaml

lib/ProductOpener/APIProductServices.pm
stephanegigandet: I'd prefer not to have both query parameters + parameters in the JSON body.
Maybe we could just make the endpoint /api/v3/product_services and have a services array inside the JSON body instead.

lib/ProductOpener/APIProductServices.pm
stephanegigandet: Will fix, thanks.

lib/ProductOpener/APIProductServices.pm
john-gom: I'm personally not sure about having this very generic "product_services" name. It feels like each operation should be its own service, e.g. /api/v3/estimate_ingredients, /api/v3/calculate_nutriscore. If the caller wants to chain multiple processes then they can simply make multiple API calls. I think this will make it easier in the futures to switch the routes for specific services to go somewhere other than product opener.

Each service could take a product object or product id as input and then just return the specific fields that it computes.  I'm not sure whether the service itself should then update the product with the modified fields or whether that should be the caller's responsibility. In a "pure" microservice model each service would store just the information that it owns in its own logically separate database and then emit an event to let other services know that this data is now available. In situations where you want to be able to provide the entire set of information from a single API you could either replicate this data to a "consolidated product" service or have some form of API federation to call out to the different services and bring all of the data together (I think the former would be more resilient)

lib/ProductOpener/APIProductServices.pm
john-gom: I'd prefer that each API return just the fields that it modifies / generates

lib/ProductOpener/APIProductServices.pm
john-gom: As discussed above, my preference is that each API service would just return the fields that it "owns"

lib/ProductOpener/APIProductServices.pm
```diff
@@ -0,0 +1,180 @@
...
+
+	# If we did not get a fatal error, we can execute the services on the input product object
+	if (not $error) {
+
+		my $product_ref = $request_body_ref->{product};
+
+		foreach my $service (split(/,/, $request_ref->{services})) {
+			my $service_function = $service_functions{$service};
+			if (defined $service_function) {
+				&$service_function($product_ref);
```
john-gom, line 155: This feels like it could be open to some form of injection attack

lib/ProductOpener/Ingredients.pm
```diff
@@ -2726,11 +2718,34 @@ sub get_missing_ciqual_codes ($ingredients_ref) {
...
+		delete_ingredients_percent_values($product_ref->{ingredients});
+		$product_ref->{ingredients_percent_analysis} = -1;
+	}
+	else {
+		$product_ref->{ingredients_percent_analysis} = 1;
+	}
+
+	compute_ingredients_percent_estimates(100, $product_ref->{ingredients});
+
+	return;
```
john-gom, line 2741: I think it would be good to return the specific properties that were modified, rather than modifying them in place, as that will make it easier to separate functions in the future

lib/ProductOpener/APIProductServices.pm
raphael0202: I agree with hangy that naming the services in the query URL this way feels weird, I would much rather see them in the post body.
I also agree with john about the "product_services" name, although I would like to have all product-computation related operations under a single API path.
Why not using `/api/v3/product/*`? We could have:
- `/api/v3/product/estimate_ingredients` and other for each individual processing items
- `/api/v3/product/pipeline` if you want to chain processing items as you did here

lib/ProductOpener/APIProductServices.pm
raphael0202: I agree with john, returning by default only the modified/generated fields with an option to return all fields would be better imo

stephanegigandet: In API v3 product WRITE, we already have a special "updated" value for fields: only updated fields are returned. We can do the same here, and make "updated" the default instead of "all".

https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/api/ref/requestBodies/fields_tags_lc.yaml

lib/ProductOpener/APIProductServices.pm
stephanegigandet: In my mind, the /api/v3/product_services is the federation API. You can use it to execute 1 or more services, either by specifying them individually, or by using aliases that bundle a bunch of operations together (like a Makefile). This way we can cut services in smaller nested services easily. e.g. we could start with one big "ingredients_analysis" service, that we could further cut in smaller services like ingredients_parsing, ingredients_percent_estimation, ingredients_vegan_vegetarian_analysis etc. (if we want to access only a specific part externally, or if internally we want to do something outside of product opener), without having to change all the callers: they can still call the "parent" service and it will be transparent to them.

It's important to have this federated pipeline, because that's how we will use those services, at least initially. e.g. for the Rate my recipe project, I need at least 5 services: ingredient parsing, ingredient percent estimation, nutrient estimation, nutriscore, ecoscore. It doesn't make a lot of sense to have the client make 5 back and forth requests, one after the other. And in practice the federated product_services API will know that all those services are provided internally by Product Opener, so it will use a fast path and not actually issue HTTP queries, those will be only needed for services provided by something else than Product Opener.

john-gom: Hi @stephanegigandet , I agree with your points. I think there are potentially two different threads here that have different requirements:

1. Provide services that other applications / services can consume to perform specific tasks. In these cases the calling application should not need to know what steps are involved in getting the data they need, they just want the result (synchronously). They potentially aren't updating a product in the database
2. In our own pipeline where we are saving a product and then chaining all of the various computations that are possible with the data that has been provided. In this case we could potentially adopt an asynchronous model to give better scalability, so each individual computation would be listening for the event that would enable them to complete their part in the chain and then emit an event when their computation is complete

You are mainly looking to address item 1 and so the approach you are taking makes sense for this. In which case I think my main comment would be that rather than the calling application have to know which steps need to be run it should just indicate the various states that it wants the results to be in, e.g. "nutriscore_computed", "ingredients_estimated", "ecoscore_computed". It would then be up to the federation API to work out which internal (or external) services need to be called, and in what order, to satisfy those states. That way the calling application does not need to be updated if we enhance our algorithms to require further inputs.

For example, if an application just needs the ecoscore_computed, then at the moment we just need a product category. However, in the future, with better ingredient estimation, we might enhance the ecoscore accuracy by looking at the specific ingredients, hence the ingredients estimation service would also need to be called. The calling application should not need to care about this change, it just wants us to return the ecoscore.

By the way, in this model it makes sense to me now that we potentially return a whole product object, or just the fields that the caller wants, although I still feel that internally our services should only return /modify the fields that they "own".

stephanegigandet:
Thanks for all the feedback, I changed services to an array, and now by default only updated fields are returned.

lib/ProductOpener/APIProductServices.pm
stephanegigandet: By default the API now only returns the created or updated fields. For structures that are updated (like when we add percent fields to the ingredients structure), I return the whole structure.

lib/ProductOpener/APIProductServices.pm
stephanegigandet: > * `/api/v3/product/estimate_ingredients` and other for each individual processing items
> * `/api/v3/product/pipeline` if you want to chain processing items as you did here

So for now, I put the specification of services in the JSON of the request body, with a "services" array. So it's like the "pipeline" option above. If you just want one service, you just put only one service in the array.

lib/ProductOpener/APIProductServices.pm
```diff
@@ -0,0 +1,218 @@
...
+					impact => {id => "failure"},
+				}
+			);
+			$error = 1;
+		}
+		else {
+			# Echo back the services that were requested
+			$response_ref->{services} = $request_body_ref->{services};
+		}
+	}
```
alexgarel, line 177: My preferences would be that when we have meaningful blocks like that (check inputs), we put them in a separate private sub. This helps reading the code.

tests/integration/expected_test_results/api_v3_product_read/get-fields-all-knowledge-panels.json
```diff
@@ -2203,7 +2203,7 @@
             "material" : {
                "id" : "en:plastic"
             },
-            "number_of_units" : 1,
```
alexgarel, line 2206: This kind of diff is really annoying…

PR #9091, opened 2023-09-30
taxonomy: Remove spaces found in some Japanese translations

Naruyoko:
### What

Removes spaces found in some `ja` taxonomy that should not be there.


benbenben2:
Thank you @Naruyoko

benbenben2 (approved):
LGTM

PR #9094, opened 2023-10-01
More german ingredient

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/taxonomies/ingredients.txt
+++ b/taxonomies/ingredients.txt
@@ -24340,6 +24340,7 @@ ciqual_food_name:fr:Avoine, crue
 <en:oat
 en:oat grain
 es:granos de avena
+de:Haferkerne
 fi:kauranjyvä, kauranjyviä, kauranjyvät
 fr:grains d'avoine
 hr:zobena prekrup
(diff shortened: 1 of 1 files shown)


taxonomies/ingredients.txt
```diff
@@ -43538,6 +43543,11 @@ fr:Framboises entières
 nl:hele frambozen
 # 10@2019-06-07
 
+<en:raspberry
+en:raspberry pieces
+de:Himbeerstücke
```
stephanegigandet, line 43548: We should try not to add entries like "something pieces", and use the ingredients_processing.txt instead.
Here it doesn't work because in ingredients.txt we have
de:Himbeeren, Himbeere

If we add "Himbeer" as a synonym in ingredients.txt, then we will recognize Himbeerstücke automatically because we have a processing for it:

en:pieces
bg:парчета, парченца, парченца от
ca:peça, peces, trossos de, trossos d', guarnició
af:stukkies
de:stücke, stücken, stückchen, gestückelt, stück

I don't know German unfortunately, so I'm hoping it's fine to put "Himbeer" without an "e" in the synonyms?

stephanegigandet (approved):

PR #9102, opened 2023-10-02
fix: better count of fruits/vegetables/legumes for Nutri-Score

stephanegigandet:
Several improvements to how we count fruits/vegetables/legumes for the Nutri-Score (old ad new formula):
- en:legume (e.g. beans) is now under en:vegetable
- entries like en:chickpea-flour that are under en:flour are not counted anymore
- entries that are vegetables but that have a processing like en:powder, en:freeze-dried or en:flour are not counted anymore
- in the categories taxonomy we can specify categories for which vegetables should not be counted because they are too much processed (e.g. crips, pop corn).

Related doc from the Nutri-Score FAQ (for the old 2021 formula, but the new 2023 formula state that this was not revisited "Of note, to date the full classification (currently based on the Eurocode 2) and the transformation
processes allowed for the inclusion in the component have not been reviewed by the ScC. The revision of
the ingredients list and the types of processes qualifying under the component will be undertaken by the
ScC in the next year (see Next steps – agenda of the ScC page 91).": 

"2. CALCULATING THE QUANTITY OF FRUITS, VEGETABLES AND PULSES IN PROCESSED PRODUCTS

Acceptable levels of processing for inclusion in the calculation

The health benefits of fruits and vegetables are associated with the whole product, including the vitamins
they provide.

Processing can result in loss of fibre and vitamins. Therefore, it would not be appropriate for ingredients
such as concentrated fruit juice sugars that are added to foods to increase sweetness to count for the
purpose of calculating a score in the same way as intact fruit and vegetables.

Intact fruits and vegetables (including those that are cooked and dried) and minimally processed fruit,
vegetables and pulses (peeled, sliced, tinned, frozen, purees, pulp, grilled, roasted or marinated) count
for the purpose of calculating a score. Roasted nuts can be also counted. Remember that fruits,
vegetables, pulses and nuts only count when their content exceeds 40%.

However, fruits, vegetables and pulses that are subject to further processing (e.g. concentrated fruit juice
sugars, powders, freeze-drying, candied fruits, fruits in stick form, flours leading to loss of water) do not
count. As an example, corn in the form of popcorn or soy proteins cannot be considered as vegetables.
Regarding the frying process, fried vegetables which are thick and only partially dehydrated by the process
can be taken into account, whereas crisps which are thin and completely dehydrated are excluded.

Any addition of external ingredients that are not classified as Fruits, Vegetables, Pulses and Nuts cannot
be counted (with exception of reconstitution with water to 100% concentrated fruit juices and dehydrated
vegetable). For example, in the case of addition of sugars/ honey on fruits or nuts, only the fruits or nuts
proportion is included in the fruit and vegetable component. Similarly, as detailed previously, additional
water added to dilute coconut in coconut milk is not included in the fruit and vegetable component."


alexgarel (approved):
LGTM except cosmetic comments.

PR #9103, opened 2023-10-02
feat: add knowledge content to additive knowledge panels

raphael0202:
Allow to easily change the knowledge content for additive knowledge panels. Content is stored in openfoodfacts-web.

Before:
Capture d’écran du 2023-10-02 17-21-25

After:
Capture d’écran du 2023-10-02 17-22-28

### DIFF
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -110,7 +110,7 @@ BEGIN {
 		&add_tags_to_field
 
 		&init_tags_texts
-
+		&get_knowledge_content
 		&get_city_code
 		%emb_codes_cities
 		%emb_codes_geo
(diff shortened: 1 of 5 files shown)


lib/ProductOpener/Tags.pm
```diff
@@ -4937,6 +4937,66 @@ sub cmp_taxonomy_tags_alphabetically ($tagtype, $target_lc, $a, $b) {
 		cmp($translations_to{$tagtype}{$b}{$target_lc} || $translations_to{$tagtype}{$b}{"xx"} || $b);
 }
 
+
+=head2 get_knowledge_content ($tag_type, $tag, $lang_id, $country_code)
```
stephanegigandet, line 4941: Could you rename $lang_id and $country_code to $target_lc and $target_cc? That's what we use elsewhere in the code for attributes, knowledge panels etc. to specify which language and country is desired.

lib/ProductOpener/Tags.pm
```diff
@@ -4937,6 +4937,66 @@ sub cmp_taxonomy_tags_alphabetically ($tagtype, $target_lc, $a, $b) {
...
+sub get_knowledge_content ($tag_type, $tag, $lang_id, $country_code) {
+	# tag value is normalized:
+	# en:250 -> en_250
+	$tag =~ s/:/_/g;
+
+	my $base_dir = "$lang_dir/$lang_id/knowledge_panels/$tag_type";
+	my $file_path = undef;
+
+	foreach my $target_country_code ($country_code, "world") {
+		$file_path = "$base_dir/$tag" . "_" . "$target_country_code.html";
```
stephanegigandet, line 4986: You can just declare my $file_path here and remove the my $file_path = undef; outside the loop.

```suggestion
		my $file_path = "$base_dir/$tag" . "_" . "$target_country_code.html";
```

lib/ProductOpener/Tags.pm
stephanegigandet: and $tag_type -> $tagtype, $tag -> $tagid (just to be consistent with the rest of the code)

raphael0202: Sure, I'm making the changes

lib/ProductOpener/Tags.pm
raphael0202: Ok, I thought it would be better to explicitely return `undef`, so that readers know it's expected, but I'm making the changes.

stephanegigandet (approved):
Looks good, thank you!

PR #9122, opened 2023-10-06
fix: processing method in one of the two ingredients around 'and'

benbenben2:
### What
Moved out the processing method analysis from parse_ingredients_text routine to a new routine called parse_processing_from_text. This function can be called 1/ where was the code initially and 2/ under _# 2 known ingredients separated by "and" ?_ 

See for example of non-parsed (začini i ekstrakti začina): https://hr.openfoodfacts.org/product/3850139240061
### Screenshot
Screenshot_20231006_203316

==============================================
The same ingredient list as in the url given before:
Screenshot_20231007_000043

- Fixes #-none-


lib/ProductOpener/Ingredients.pm
```diff
@@ -1277,6 +1278,226 @@ sub match_origin_of_the_ingredient_origin ($ingredients_lc, $text_ref, $matched_
...
+						ingredient => $ingredient,
+						new_ingredient => $new_ingredient,
+						processing => $ingredient_processing_regexp_ref->[0],
+						regexp => $regexp
+					}
+				) if $log->is_debug();
+
+				$matching = 1;
+				$matches++;
+				if ($new_processing ne "") {
```
stephanegigandet, line 1421: I think you also need to add a comma if the input $processing is non empty.

lib/ProductOpener/Ingredients.pm
```diff
@@ -1831,33 +2055,81 @@ sub parse_ingredients_text ($product_ref) {
...
-						canon_ingredient1 => $canon_ingredient1,
-						canon_ingredient2 => $canon_ingredient2
-					}
-				) if $log->is_debug();
-
+				# both ingredients are in the taxonomy as is
 				if (    (exists_taxonomy_tag("ingredients", $canon_ingredient1))
 					and (exists_taxonomy_tag("ingredients", $canon_ingredient2)))
 				{
+					$debug_ingredients
```
stephanegigandet, line 2073: This works very well, but I think we could further simplify it:

We change the parse_processing_from_text in a function that first check if the ingredient exists, and otherwise tries to parse the processing.

Then we call it on both ingredients right away, and we can just check if both ingredients were recognized (either as-is, or by removing processing) in only one condition.

And that way we would also recognize "fried steak and grilled mushrooms" with that same condition.

stephanegigandet:
> Moved out the processing method analysis from parse_ingredients_text routine to a new routine called parse_processing_from_text. This function can be called 1/ where was the code initially and 2/ under _# 2 known ingredients separated by "and" ?_

That's a great idea! I think we could improve it even further (see comments).

lib/ProductOpener/Ingredients.pm
benbenben2: I tried something in that direction

lib/ProductOpener/Ingredients.pm
```diff
@@ -1277,6 +1278,231 @@ sub match_origin_of_the_ingredient_origin ($ingredients_lc, $text_ref, $matched_
...
+
+English first element for that ingredient (en:pear, for example)
+
+=head4 ingredient_recognized
+
+0 or 1
+
+=cut
+
+sub parse_processing_from_ingredient ($ingredients_processing_regexps,
```
stephanegigandet, line 1341: It's a bit confusing to have so many input parameters. When the function is called, it's difficult to remember which parameter is what.

I think we can keep only $ingredients_lc and $ingredient. It forces us to canonicalize the ingredient at the beginning of the function, but it simplifies things a lot I think.

I made a PR here with a proposed refactor: https://github.com/openfoodfacts/openfoodfacts-server/pull/9146

stephanegigandet (approved):
Thank you very much @benbenben2 !

PR #9127, opened 2023-10-09
fix: add ingredients-original facet, start to split tagtype and taxonomy

stephanegigandet:
In addition to ingredients_tags, we also have ingredients_original_tags which contains the ingredients actually listed in the ingredients text, and not their parents.

This PR adds a facet /ingredients-original that allows us to list those. It will be very useful in particular to see which ingredients are the most popular.

As ingredients_original uses the ingredients taxonomy, this PR starts to differentiate the tag type (ingredients or ingredients_original from the taxonomy : ingredients). We could continue down this path to have traces use the allergens taxonomy without copying it as we do today (same for the data quality facets)

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -1514,6 +1514,7 @@ sub query_list_of_tags ($request_ref, $query_ref) {
 	add_country_and_owner_filters_to_query($request_ref, $query_ref);
 
 	my $groupby_tagtype = $request_ref->{groupby_tagtype};
+
 	my $page = $request_ref->{page};
 	# Flag that indicates whether we cache MongoDB results in Memcached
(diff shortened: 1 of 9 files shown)


alexgarel (approved):
Good !

FTR: robot crawling is also disabled thanks to @index_tag_types in Config_off (white list of facets)

alexgarel:
@stephanegigandet won't this break @john-gom current PR for off-request ?

stephanegigandet:
> @stephanegigandet won't this break @john-gom current PR for off-request ?

Right, we will need to add it, otherwise we'll get 0 results, except if we add ?no_cache=1 in which case we'll go to the real mongodb.

It's a new internal facet for now, so we have time to add it to the query service and to document it.

PR #9133, opened 2023-10-10
fix: change paths in frontend-dist.tgz to match expected paths in production

stephanegigandet:
This is so that we can copy all 3 directories in one command, instead of having to specify where each of the js, icons and css dirs should go.


alexgarel:
@stephanegigandet in the target installation on off2, this will be this structure (remember, we want to be able to git pull, so we keep the repository part clean, and we add symlinks).
That's why we should not change it… (your situation was transient).

PR #9141, opened 2023-10-11
fix: milk, dairy drinks and plant-based beverages Nutri-Score 2023

stephanegigandet:
In the new Nutri-Score formula, milk, dairy drinks and dairy drinks substitutes are considered beverages.


alexgarel (approved):
Good

PR #9142, opened 2023-10-11
fix: exists_taxonomy_tag for ingredients_original tag type

stephanegigandet:
continuation of https://github.com/openfoodfacts/openfoodfacts-server/pull/9127 so that https://world.openfoodfacts.org/popularity/top-1000-scans-2022/ingredients-original?stats=1 can work correctly.


alexgarel (approved):
:+1:

PR #9075, opened 2023-09-26
taxonomy: Lithuanian ingredients

DanieliusAsm:
### What
added some ingredients for products

PR #9148, opened 2023-10-13
taxonomy: Lithuanian ingredients + processing

DanieliusAsm:
### What
Retrying because large conflicts in #9075

Plus adding ingredients_processing


taxonomies/ingredients.txt
```diff
@@ -13774,7 +13780,7 @@ kk:Крахмал
 ko:녹말
 la:Amylum
 li:Zètmael
-lt:Krakmolas
+lt:Krakmolas, tirštiklis krakmolas
```
benbenben2, line 13783: is it synonym of Krakmolas or should synonym of "Modifikuotas krakmolas" in additives.txt?

benbenben2:
For "lt:regeneruotas nugriebtas pienas", you can 
1/ add "lt:regeneruotas" under "en:reconstituted" in ingredients_processing.txt 
2/ make sure that "nugriebtas pienas" is in ingredients.txt (this is the case, under "en:skimmed milk") and 
3/ add "lt" in the processing parsing in Ingredients.pm (this is under "# Try to remove ingredients processing "cooked rice" -> "rice", you have 3 possibilities: 
-   a) # match before or after the ingredient, require a space, 
-   b) #  match before or after the ingredient, does not require a space, 
-   c) # match after the ingredient, does not require a space + # match before the ingredient, require a space, 

I think "lt" would be a)


"en:watermelon juice" , "lt:arbūzų sulčių koncentratas" would not be needed after these feat


It will be a benefit for you because not only "arbūzų sulčių koncentratas", for example will be recognized but all ingredients like "something sulčių koncentratas" as long as "something" is in the taxonomy already.

taxonomies/ingredients.txt
DanieliusAsm: tirštiklis is thickener. In one of the product's ingredients it was listed "tirštiklis krakmolas". "**tirštiklis**" should probably be somewhere else not sure where.

DanieliusAsm:
@benbenben2 
1. Done
2. Already is
3. I think I added it (First time in Ingredients.pm) not sure if thats it.

Shouldn't then "arbūzų" be also in ingredients_processing just like with reconstituted example ? and then "sulčių koncentratas" in ingredients.txt ? But then I need to add all kinds of possibilities for juice concentrates: Banana, orange, apple etc.

Or does it work if I have in ingredients.txt: "en:watermelon" and "en:concentrated juice" defined ?

taxonomies/ingredients.txt
benbenben2: me neither. Let's leave it in ingredients.txt for now.

taxonomies/ingredients.txt
```diff
@@ -4034,6 +4035,7 @@ fr:lait écrémé réhydraté
 # hr:rekonstituirano obrano mlijeko # see ingredients_processing.txt
 hu:visszaállított sovány tej, feloldott sovány tej
 it:latte scremato reidratato
+lt:regeneruotas nugriebtas pienas
```
benbenben2, line 4038:
```suggestion
# lt:regeneruotas nugriebtas pienas # see ingredients_processing.txt
```

benbenben2:
> Shouldn't then "arbūzų" be also in ingredients_processing just like with reconstituted example ? and then "sulčių koncentratas" in ingredients.txt ? But then I need to add all kinds of possibilities for juice concentrates: Banana, orange, apple etc.
> 
> Or does it work if I have in ingredients.txt: "en:watermelon" and "en:concentrated juice" defined ?

It works if you have in ingredients.txt: "en:watermelon" and in ingredients_processing.txt "en:concentrated juice" defined.

benbenben2:
For the failing tests, you have to run "make lint" locally.

alexgarel:
I added the lint commit and resolved conflict.

@benbenben2 is this ready to merge (if tests pass ?)

benbenben2 (approved):
Good job @DanieliusAsm!

benbenben2:
Thanks @alexgarel

PR #9154, opened 2023-10-15
fix: avoid duplicated for data quality tags

benbenben2:
### What

avoid duplicated data quality tags for vegan (vegetarian) labels having non-vegan (non-vegetarian) ingredients.

### Screenshot
Before the fix:
Screenshot_20231015_162014

- Fixes #-none-


stephanegigandet (changes requested):
There is an easier way in fact, with the add_tag() function from Tags.pm:

add_tag($product_ref, "data_quality_errors_tags", "en:vegan-label-but-non-vegan-ingredient");

It won't add the tag if it already exists, so we don't need the check on has_tag().

benbenben2:
seems that it should be:
`add_tag($product_ref, "data_quality_errors", "en:vegan-label-but-non-vegan-ingredient");`
without _tags

stephanegigandet (approved):
Thank you!

PR #9162, opened 2023-10-17
fix: fix madenearme scripts to load geo codes

alexgarel:
attempt to correct bug on madenearme (currently only showing french locations)


stephanegigandet:
> attempt to correct bug on madenearme (currently only showing french locations)

so does it work?

alexgarel:
@stephanegigandet we will see it when we deploy !

PR #9181, opened 2023-10-21
taxonomy: Even more beer categories

DanieliusAsm:
### What
- Added some more beer categories
- Added a lot of descriptions to better understand categories.
- Most of the work was done towards Lagers - one of the primary types of beers.

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -1756,14 +1756,15 @@ wikipedia:en:https://en.wikipedia.org/wiki/Sour_beer
 description:en:Sour is a type of beer which is intentionally brewed to have sour taste
 
 <en:Sour beers
-en:Catarina Sour beers, Catarina Sour
-xx:Catarina Sour beers, Catarina Sour
-description:en:Catarina sour is a Brazilian origin sour fruit beer.
+en:Catharina Sour beers, Catharina Sour
+xx:Catharina Sour beers, Catharina Sour
+description:en:Catharina sour is a Brazilian origin sour fruit beer.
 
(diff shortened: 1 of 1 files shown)


aleene:
Great work. I wonder how you decide when a new category is needed. A category should have the function to help a user decide between products within a category based on any product attribute. If there are only very few products within the category, a category has no sense. This however might change for the better when new products are added.

DanieliusAsm:
@aleene I scanned and added [a lot of beers](https://lt.openfoodfacts.org/category/alus). Before me there were only 2 lithuanian beers and now theres 293. Some of them had these categories that I am adding, but there are other categories that I just got from a beer chart online and they might not have any products currently. Like Grodziskie doesnt have any but is considered a historical beer that nearly went extinct.

The PR where I added lots of IPA's was where I went through all IPA's uploaded on Openfoodfacts and found out about them.

I Guess I also thought about the future. Maybe it would be possible to get a beer community to join OFF and scan and add lots of beers.

aleene:
Great that you reviewed the already available beers already. You might have noticed that I added a few in the years. For a long time beers was my top category.
It would be great to have beer enthusiasts sign up and add beers. Especially as I have a big worry on beer categorisation. Who understands beers enough to be able to categorise them. Any help you can enter in the taxonomy is very helpful. It can take the form of "look for these words then it belongs there". In the future OFF will provide more help to the user and these sentences should form the basis for that.

stephanegigandet (approved):

alexgarel:
really cool to see that kind of contribution @DanieliusAsm !

PR #9188, opened 2023-10-24
taxonomy: add "abbey beers" proxy to Trappist

OpenChris:
add "abbey beers" proxy to Trappist, since abbey beers are by definition in trappist style

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -1604,7 +1604,7 @@ de:Abteibier
 fi:luostariolut
 fr:Bières d'abbayes, Bières d'abbaye, Bières d'abbayes régionales
 hu:Apátsági sör
-it:Birre d'abbazia, Birre trappiste
+it:Birre d'abbazia
 nl:Abdijbieren
 agribalyse_food_code:en:5011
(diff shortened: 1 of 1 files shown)


stephanegigandet (changes requested):
Hi @OpenChris , I think the best solution would be to put en:trappist beers under:
<en:Ales
en:Abbey ales, Abbey beer, Abbey ale, Abbey beers

What do you think? That way it would inherit the properties. cc @aleene @DanieliusAsm

DanieliusAsm:
@stephanegigandet the description under each explains those 2 categories.

- Trappist:  A type of beer that is brewed by trappist monks. There is a strict criteria managed by the International Trappist Association that grants Authentic Trappist Product label to breweries that have monks.
- Abbey: an ale brewed in the style of Trappist beers but not by actual Trappist monks. It is an ale that originated from Belgium. 
Basically Abbey is the same style of beer like Trappist, but abbey is not brewed by monks. 

I dont think it should be merged into a single category since only one of them is produced by monks. But maybe thats not too important for OFF and can be merged under one category.

PR #9190, opened 2023-10-25
fix: Nutri-Score A to B threshold and olive oils

stephanegigandet:
- we had a wrong A to B threshold for food products that are not fats
- olive oil / avocado oil should be counted as fruits/vegetables/legumes when a product is in the fats category
- more tests


alexgarel (approved):
LGTM

PR #9192, opened 2023-10-25
fix: avoid division by 0 in image crop

### LINKED ISSUE #9191: Illegal division by zero at /srv/off/lib/ProductOpener/Images.pm line 1318
Sentry Issue: [SMOOTHIE-1SF](https://openfoodfacts.sentry.io/issues/3901370484/?referrer=github_integration)

```
_Exception: Exception: Bad response (500): <h1>Software error:</h1>
<pre>Illegal division by zero at /srv/off/lib/ProductOpener/Images.pm line 1318.
</pre>
<p>
For help, please send mail to the webmaster (<a href="mailto:contact@openfoodfacts.org">contact@openfoodfacts.org</a>), giving this error message 
and the time and date of the error.

</p>
```

alexgarel:
relates to #9191


lib/ProductOpener/Images.pm
```diff
@@ -1309,28 +1309,35 @@ sub process_image_crop ($user_id, $product_id, $id, $imgid, $angle, $normalize,
...
-	my $geometry = "${nw}x${nh}\+${ox1}\+${oy1}";
-	$log->debug("geometry calculated",
-		{geometry => $geometry, ox1 => $ox1, oy1 => $oy1, ox2 => $ox2, oy2 => $oy2, w => $w, h => $h})
-		if $log->is_debug();
-	if ($nw > 0) {    # image not cropped
-		my $imagemagick_error = $source->Crop(geometry => $geometry);
-		($imagemagick_error)
-			and $log->error("could not crop to geometry", {geometry => $geometry, error => $imagemagick_error});
+	# avoid a potential divide by zero error
+	if (!$w or !$h) {
```
stephanegigandet, line 1313: That will avoid the division by zero, but I'm not sure what happens later on.
It seems the problem was that $new_product_ref->{images}{$imgid} is not defined.
So in this case, we would not crop the image, but select it.

lib/ProductOpener/Images.pm
```diff
@@ -1309,28 +1309,35 @@ sub process_image_crop ($user_id, $product_id, $id, $imgid, $angle, $normalize,
...
+
+		my $ox1 = int($x1 * $ow / $w);
+		my $oy1 = int($y1 * $oh / $h);
+		my $ox2 = int($x2 * $ow / $w);
+		my $oy2 = int($y2 * $oh / $h);
+
+		my $nw = $ox2 - $ox1;    # new width
+		my $nh = $oy2 - $oy1;
+
+		my $geometry = "${nw}x${nh}\+${ox1}\+${oy1}";
```
stephanegigandet, line 1332: There's a $geometry on line 1343 below that won't exist if you use "my $geometry" here I think (I may be reading the diff wrong)

alexgarel: Oh yes, you're right I misread the `=~` for an `=` so I though it was redefined.

lib/ProductOpener/Images.pm
alexgarel: And so what would you advise to do.

stephanegigandet: It's fine as is I guess. Or we could add a check earlier: if the imgid is not in images, then we don't select and crop it. I don't really understand the conditions of this bug happening, if that's an issue on a few products, or if there's a specific image that consistently results in this.

teolemon:
Run make check_critic
🥫 Checking with perlcritic
test -z "lib/ProductOpener/Images.pm" || docker-compose --env-file=.env  run --rm --no-deps backend perlcritic lib/ProductOpener/Images.pm
Creating po_backend_run ... 
Creating po_backend_run ... done
Mixed high and low-precedence booleans at line 1313, column 6.  See page 70 of PBP.  (Severity: [4](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/6665169777/job/18114664222?pr=9192#step:7:5))
2
make: *** [Makefile:3[5](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/6665169777/job/18114664222?pr=9192#step:7:6)4: check_critic] Error 2
Error: Process completed with exit code 2.

PR #9197, opened 2023-10-25
fix: speed up Lang.pm initialization, save 2 seconds x 100 tests

stephanegigandet:
This is an attempt to speed up tests. Almost all tests load Lang.pm.
- removed useless sleep()
- made parsing of the po/tags/ file saved in a file (like we do for po/common)

In profiling done with nytprof, this saves about 2 seconds per initialization. As we have 100 unit + integration tests, I'm hoping this should save 3 minutes when running all tests.

### DIFF
--- a/lib/ProductOpener/Lang.pm
+++ b/lib/ProductOpener/Lang.pm
@@ -254,7 +254,7 @@ sub lang_in_other_lc ($target_lc, $stringid) {
 
 $log->info("initialize", {data_root => $data_root}) if $log->is_info();
 
-# Load stored %Lang from Lang.sto
+# Load stored %Lang from Lang.sto and Lang_tags.sto
 
 my $path = "$data_root/data/Lang.${server_domain}.sto";
 if (-e $path) {
@@ -275,7 +275,6 @@ if (-e $path) {
 	}
 
 	$log->info("Loaded languages", {langs => (scalar @Langs)}) if $log->is_info();
-	sleep(1) if $log->is_info();
 }
 else {
 	$log->warn("Language translation file does not exist, \%Lang will be empty. Run scripts/build_lang.pm to fix this.",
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Lang.pm
```diff
@@ -275,27 +275,49 @@ if (-e $path) {
...
-# Tags types to path components in URLS: in ascii, lowercase, unaccented,
-# transliterated (in Roman characters)
-#
-# Note: a lot of plurals are currently missing below, commented-out are
-# the singulars that need to be changed to plurals
-my ($tag_type_singular_ref, $tag_type_plural_ref)
-	= ProductOpener::I18N::split_tags(ProductOpener::I18N::read_po_files("$data_root/po/tags/"));
-%tag_type_singular = %{$tag_type_singular_ref};
-%tag_type_plural = %{$tag_type_plural_ref};
+$path = "$data_root/data/Lang_tags.${server_domain}.sto";
```
alexgarel, line 285: This is more cache data, could you move it to the cache folder.

scripts/build_lang.pl
```diff
@@ -48,6 +49,7 @@
 	mkdir("$data_root/data", 0755) or die("Could not create target directory $data_root/data : $!\n");
 }
 store("$data_root/data/Lang.${server_domain}.sto", \%Lang);
```
alexgarel, line 51: here also it would be better to have it in a cache folder.

alexgarel (approved):
Good idea, I am just a bit picky about using a cache directory instead of a data directory (help to better identify what to backups, be redundant, etc.)

stephanegigandet:
> Good idea, I am just a bit picky about using a cache directory instead of a data directory (help to better identify what to backups, be redundant, etc.)

Let's do it once the migration is done. In fact /data was intended to be a "cache" folder in the sense that it contains things that can be recomputed. We should revisit those things globally. The built lang files are more like the built taxonomies, we could just use the same system we have for taxonomies.

PR #9203, opened 2023-10-26
fix: frontend fixed

raphael0202:
### What

A few CSS issues that needed fixing:

- the path for playstore badges were not commited (only in prod server), I switched to the file paths that really exists on the repo
- the mobile banner margin was incorrect + remove radius
- fix spacing issues in footer links

#### Mobile banner (before/after)

Capture d’écran 2023-10-26 à 10 33 46
Capture d’écran 2023-10-26 à 10 33 48

#### Footer (before/after)

Capture d’écran 2023-10-26 à 10 34 07
Capture d’écran 2023-10-26 à 10 34 02

### DIFF
--- a/po/common/aa.po
+++ b/po/common/aa.po
@@ -193,9 +193,9 @@ msgctxt "android_apk_app_link"
 msgid "https://world.openfoodfacts.org/files/off.apk"
 msgstr ""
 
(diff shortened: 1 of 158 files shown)


stephanegigandet (approved):
Thank you!

PR #9210, opened 2023-10-26
fix: compute percent of vegetables for Nutri-Score after removing water for canned vegetables

stephanegigandet:
Some categories of products typically have nutrition facts for the "drained product" (produit égoutté in French). We need to take that into account when computing the % of fruits/vegetables.

We don't record if the nutrition facts is for the drained product, so we take a guess that products in the "canned vegetables" category have nutrition facts for the drained product. There are counter examples, but usually those counter examples have little water anyway.

"IS THE NUTRI-SCORE CALCULATED FOR THE PRODUCT WITH THE COVERING LIQUID?
The nutritional score is calculated using the nutritional data listed on the package for 100 g of the product,
whose nutrients form part of the mandatory nutritional declaration or are included as supplemental
information, in accordance with Article 30 of the INCO regulation no. 1169/2011.
If the covering liquid is taken into account in the mandatory nutritional declaration, the Nutri-Score should
also be calculated for the product + the covering liquid and vice versa. To calculate the percent of “fruits,
vegetables, pulses, nuts and oils », you have to use the same basis as the table of nutritional values. If the
covering liquid is taken into account, the calculation should include it. If the nutritional values do not take
the covering liquid into account, then it should be recalculated without the covering liquid."


lib/ProductOpener/Ingredients.pm
```diff
@@ -6790,33 +6790,55 @@ The match function takes 2 arguments:
...
+		# in that case, try to get the possible percent values in nested sub ingredients
 		elsif (defined $ingredient_ref->{ingredients}) {
-			$count += add_ingredients_matching_function($ingredient_ref->{ingredients}, $match_function_ref);
+			my ($sub_ingredients_percent, $sub_ingredients_water_percent)
+				= add_ingredients_matching_function($ingredient_ref->{ingredients}, $match_function_ref);
+			$percent += $sub_ingredients_percent;
+			$water_percent += $sub_ingredients_water_percent;
+		}
+		# Keep track of water
+		elsif (is_a("ingredients", $ingredient_ref->{id}, 'en:water')) {
```
alexgarel, line 6836: I'm not sure why it's an elsif here ?

Are we sure that water never match ? Why not a simple if ? Or, maybe, put it as first condition ?

alexgarel (approved):
I have a comment that needs either a change or a comment in the code.

All good for the rest :+1:

alexgarel:
Also you need a make tidy !

lib/ProductOpener/Ingredients.pm
stephanegigandet: It was avoid double counting water if we have something like "flavoured water (water, flavour)", but I'll do it differently.

alexgarel: A comment maybe enough then

PR #9215, opened 2023-10-27
fix: % of fruits/vegetables for canned foods and Nutri-Score

stephanegigandet:
small fixes to compute the % of fruits/vegetables for canned foods (exclude water from all canned plant based foods, except canned fruits)


alexgarel (approved):
LGTM

Issue #5706, opened 2021-09-20
Add more ingredient analysis (tracker)


teolemon:
### Who for
- Users with a food intolerance
- Users with a special diet

### What

- [ ]

### Why
- Because we can
- Because users will be grateful
- Because it's a good way to increase number of thematic reuses
- Because it's a good way to increase incentives to complete ingredients

### Part of
- #5529

PR #9224, opened 2023-10-28
feat: support percentage range

benbenben2:
### What
allow range in percentage.
**Example:** kakao prah smanjene masti (10-12%) in https://hr.openfoodfacts.org/product/3858886803621

**proposed solution:** ignore first part of the range, _i.e._ keep upper value only.
kakao prah smanjene masti (10-12%) -> kakao prah smanjene masti (12%)

- #5706


alexgarel (approved):
Great !

stephanegigandet (approved):
Thanks!

PR #9227, opened 2023-10-29
taxonomy: Added unknown Croatian ingredients to the taxonomy (part 8)

benbenben2:
### What
 - added HR in taxonomy
- "Pasterizirano mlijeko (s 1.0% mliječne masti)" part in parenthesis is not ingredient but apply to previous ingredient. Also, not processing, also not sub ingredient... hence, tries category-type
- added %max_regexp. But that may cause issue for percentage calculation:
issue for max: **1st ingredient is 70% min, 2nd 30% max**, and there are more ingredients but estimated at 0...
https://hr.openfoodfacts.org/product/3850355006779/salakis-feta-president

- Fixes #-none-


lib/ProductOpener/Ingredients.pm
```diff
@@ -3882,7 +3884,16 @@ sub normalize_a_of_b ($lc, $a, $b, $of_bool) {
 	$b =~ s/^\s+//;
 
 	if (($lc eq "en") or ($lc eq "hr")) {
-		return $b . " " . $a;
+		# start by "with" (example: "mlijeko (s 1.0% mliječne masti)"), in which case it $b should be added after $a
+		# start by "with etc." should be added at the end of the previous ingredient
+		my %with = (hr => '(s | sa )',);
+		my $with = $with{$lc} || " will not match ";
+		if ($b =~ /^(s |sa )/i) {
```
stephanegigandet, line 3891:
```suggestion
		if ($b =~ /^$with/i) {
```

stephanegigandet (approved):
Thank you!

At some point we'll need to handle better the parsing of min and max so that the value go to percent_min and percent_max, but we can do that later.

PR #9228, opened 2023-10-30
feat: Use query service for obsolete queries

### LINKED ISSUE #9134: Querying obsolete products through openfoodfacts-query
It's possible to get lists of obsolete products by adding ?obsolete=1 to URLs. In that case we issue query on the products_obsolete collection.

e.g. to see brands for which we have products marked obsolete: https://world.openfoodfacts.org/brands?obsolete=1

This behavior does not seem to be implemented with the openfoodfacts-query service.

john-gom:
### What

The query parameters used in the PO query are passed through to the query service so that the obsolete flag can be set.

Also fixed an issue where the obsolete flag was not being taken into account for results caches.

Removed all references to the MongoDB products_tags collection

- Fixes #9134


lib/ProductOpener/Display.pm
```diff
@@ -1515,6 +1515,14 @@ sub can_use_query_cache() {
 			and (not $server_options{producers_platform}));
 }
 
+sub generate_query_cache_key ($name, $context_ref) {
+	# Generates a cache key taking the obsolete parameter into account
+	if (scalar param("obsolete")) {
```
stephanegigandet, line 1520:
```suggestion
	if (scalar request_param("obsolete")) {
```

Just in case we have API search queries with the obsolete parameter in the JSON body in the future (we don't currently).

stephanegigandet (approved):
Thank you!

PR #9236, opened 2023-11-02
taxonomy: Added unknown Croatian ingredients to the taxonomy (part 9)

benbenben2:
### What

- added HR in taxonomy

- seems to me that we could make difference between: 
  - filling (garniture in FR) that is inside a preparation (like in doughnut)
  - coating (nappage in FR) that is around a preparation
 
  I tried to apply changes accordingly

- Also wondering if we could merge preparation and filling? Can we say that preparations are fillings? As it is not clear to me (there is term en:compound as well), I made all preparations as children of en:preparation (although preparation alone does not exists). 

- Fixes #-none-


taxonomies/ingredients_processing.txt
```diff
@@ -1851,6 +1855,32 @@ hr:rafinirano, djelomično rafinirano
 #fr:Dessucré, partiellement dessucré
 #hr:bez šećera, sa smanjenom količinom šećera
 
+en:Homogenization, homogenisation
```
stephanegigandet, line 1858: I think this entry should be "Homogenized", not "Homogenization"

stephanegigandet:
> * seems to me that we could make difference between:
>   
>   * filling (garniture in FR) that is inside a preparation (like in doughnut)
>   * coating (nappage in FR) that is around a preparation
>   
>   I tried to apply changes accordingly
> * Also wondering if we could merge preparation and filling? Can we say that preparations are fillings? As it is not clear to me (there is term en:compound as well), I made all preparations as children of en:preparation (although preparation alone does not exists).

Another option woul be to make "preparation", "filling" and "coating" ingredients processing instead. That way we don't have to add entries for basically all fruits, vegetables etc. ("strawberry preparation" and so on) and their translations in all languages.

We will have to be careful when counting % of fruits and vegetables (e.g. "strawberry preparation 50% (strawberries, water, sugar)" should not count as 50% of fruits), but it can be done easily.

aleene:
In principle I am not in favour to put it in processing, as the idea was that a processing step would not influence the characteristics of an ingredient. For filling this is not the case (strawberry and strawberry filling are only related on one ingredient.
However we are talking here about compounds: an X filling can have multiple ingredients. So the question is more what to do with compounds? As an ingredient this is unimportant. For percentage determination maybe. For analysis purpose we do not care where it is a filling or whatever. It messes up the ingredients taxonomy and could just be replaced by the word compound. It belongs more to a list of "ingredients" to be ignored.

benbenben2:
Good point from @aleene 
Would that be an option to create a new .txt file for compounds?

"strawberry filling" may be related to more than one ingredient. For example, this "apple filling" (Punjenje od jabuke) contains more than apple: https://hr.openfoodfacts.org/product/3856021206184/pita-of-jabuka-s-budget

stephanegigandet:
Some queries that are useful to see what kind of fillings etc. we have in ingredients lists:

https://uk.openfoodfacts.org/ingredients?filter=filling

https://fr.openfoodfacts.org/ingredients?filter=preparation

benbenben2:
This is a good idea.

stephanegigandet:
@benbenben2 I think we can merge as-is (maybe change "homogenization" to "homogenized" in the ingredients processing first).

I'm filing a bug for the proposal to deal with compounds.

stephanegigandet (approved):

stephanegigandet:
New issue for the parsing of "xyz preparation": https://github.com/openfoodfacts/openfoodfacts-server/issues/9345

stephanegigandet (approved):
Thank you!

PR #9237, opened 2023-11-02
taxonomy: correction fruit juice concentrate

moon-rabbitOFF:
Example of a product for which the amount of fruits, vegetable and nuts was wrongly calculated https://world.openfoodfacts.org/product/6413300017861/luonto-vanilja-arla


stephanegigandet (approved):
Thanks

PR #9247, opened 2023-11-06
fix: Nutri-Score for olive oils with unrecognized ingredients

stephanegigandet:
This PR:
- fixes an issue in Tags.pm for some functions that return a property value. When the property value was 0, it was not returned.
- simplifies and makes it more clear that for some categories we force the % of fruits/vegetables while for some other we provide default values in case we do not have ingredients
- this fixes products like olive oils that have unrecognized ingredients (e.g. an expected way to write that it contains 100% olive oil): we force them to have 100% fruits/vegetables


taxonomies/categories.txt
```diff
@@ -35242,7 +35242,7 @@ wikidata:en:Q1163138
 ciqual_food_code:en:31027
 ciqual_food_name:en:Candied fruits
 ciqual_food_name:fr:Fruit confit
-nutriscore_without_unprocessed_fruits_vegetables_legumes:en:yes
+nutriscore_category_override_for_fruits_vegetables_legumes:en: 0
```
alexgarel, line 35245: Have we any documentation on how this property works ?

alexgarel (approved):
LGTM !

taxonomies/categories.txt
stephanegigandet: @alexgarel I'm adding a comment for it at the top of the taxonomy, where there is some doc for some properties already.

aleene:
so the category defines the nutriscore. You can also let it define the ingredient (so you get the right nutriscore?).
This can be extended to all single ingredient categories.

stephanegigandet:
> so the category defines the nutriscore. You can also let it define the ingredient (so you get the right nutriscore?). This can be extended to all single ingredient categories. Could help as a category quality check as well: if there are more than one ingredient detected, then either the category is wrong or the ingredient.

@aleene We already have a mechanism for that in fact (added by @benbenben2 ),
e.g. for the extra-virgin olive oils category:

```
<en:Olive oils
en:Extra-virgin olive oils, Extra virgin olive oil
[..]
expected_nutriscore_grade:en:c
expected_ingredients:en: en:olive-oil
```

aleene:
Can't you the expected_ingredient field to calculate the nutriscore?

PR #9249, opened 2023-11-06
fix: update donation banner text

stephanegigandet:
image

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -3984,6 +3984,10 @@ msgctxt "thank_you"
 msgid "Thank you!"
 msgstr ""
 
+msgctxt "thank_you_very_much"
+msgid "Thank you very much!"
+msgstr ""
+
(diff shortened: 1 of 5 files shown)


teolemon (approved):

teolemon:
Can we possibly add bold, and possibly "independant from the food industry" @stephanegigandet

stephanegigandet:
> Can we possibly add bold, and possibly "independant from the food industry" @stephanegigandet

We can anything, please discuss it with Gala who sent me the new text.

teolemon:
@g123k do you think we could have a simple non looping animation for the heart

g123k:
> @g123k do you think we could have a simple non looping animation for the heart

You, it's totally possible with Rive
For the web? For the app?

teolemon:
both would be nice @g123k 
the usecase here would be web, but we could put it in the app

PR #9251, opened 2023-11-06
feat: Implement OIDC sign-in

### LINKED ISSUE #1596: Consider a central User Service for auth
Somewhat related to #1204 

Right now, the option to login to different Product Opener sites (OpenFoodFacts, OpenBeautyFacts, OpenPetFoodFacts, and OpenProductFacts) is implemented by sharing the `$data_root/users` between all sites. In production, this is (AFAIK) done by a symlink. A more could be to have a central authentication/authorization site, that provides a M2M API (ie. OIDC), so that the connected sites can use that API instead of having direct access to the `$data_root/users` directory.

hangy:
### What

The main aim is to use Keycloak as an IdP, and connect Product Opener as a RP. We want to move the basics of user management out of Product Opener.

In the future, this can be used as a base to enable #1204 to be implemented, even though it will probably not be part of the initial change as to not blow up the PR too much. One important requirement is to keep compatibility with the current Basic Auth mechanism.

### Screenshot

TBD

- closes #1596

Adding a task list to track progress:

- [x] @hangy: Move all interaction with user_emails.sto to call Keycloak APIs instead
- [x] @john-gom: Move all read and write operations to user.sto files to one place
- [x] Update user.sto read/write operations to get user "signup" fields from Keycloak (not required, as we want Keycloak as the canonical source)
- [x] @hangy: Pass country / language to login page
- [x] @hangy Add delete user event listener to Keycloak, trigger Redis event openfoodfacts/openfoodfacts-auth#3
- [x] @hangy Store deleted users in a table openfoodfacts/openfoodfacts-auth#14
- [x] @hangy Redis event subscriber on Product Opener to remove user name from products (TBA extend to other applications)  (WIP https://github.com/hangy/openfoodfacts-server/tree/keycloak-redis)
- [x] @john-gom: Add all "signup" fields to Keycloak and import language and country lists from taxonomy json
- [x] @john-gom: Fix locale pick list to be searchable
- [x] Upgrade to 24.0.0 when available and check localization is working for account page
- [x] Set up Crowdin translation project openfoodfacts/openfoodfacts-auth#2
- [x] @hangy Test with Mobile and other API usage scenarios
  - Flutter app seems to work fine
  - Hunger games shows user logged in (with modified hunger games branch, as `main` currently has a lot of hard coded URLs)
  - Write test with robotoff not that easy in separate setup 😞 
- [x] @hangy Benchmark import of the number of users we have in production
  `time ./scripts/migrate_users_to_keycloak.pl` with around 400.000 synthetic accounts took about 45 minutes on my desktop PC in WSL2.

Save for a later phase:

- [ ] HOLD on this pending unification. Live with inconsistent passwords for now. Update OPF, OBF and OPFF to login via Keycloak and redirect to OFF for user account info
- [ ] Extend user attribute table to store more text
- [ ] Select correct language / country for user after login redirect back to OFF
  Example: If the users was on `fr.openfoodfacts.org` before signing up, and they change their locale to Spanish during registration in Keycloak, do we want them to be redirected to a Spanish page instead?
  Problem: Differentiating between first login (because of registration) and subsequent login would have to be done based on whether or not `${userid}.sto` exists.
- [ ] Get Product Opener client to self register on initialization and store secret somewhere safe
  Advantage: The client secret does not have to be exposed to the admin. However, they'll have access to the secret anyways, unless we were to store the secret in some kind of HSM.
  Disadvantage: Some client configuration can be done during self-registration, but some necessary permissions like realm-management need to be configured manually by an admin, anyways.
- [ ] Get rid of user.sto files completely
  To consider: This file currently contains information about the users' active sessions. Do we want to replace the opaque session identifier by an encrypted cookie? Also, we need some kind of directory to identify users that have used ProductOpener know if we have to do something when a deletion event comes in through Redis. (Obviously, this could be done in MongoDB or PostgreSQL instead, but is there a huge difference?)
- [ ] Move API consumers to use client_credentials rather than login
- [ ] Update mobile app to use Keycloak web forms for login (eg. OIDC Authorization Code Flow)
- [x] ~~Figure out how/if org management should be done in Keycloak~~  (Done in MongoDB)

Moved to `openfoodfacts-auth`:
- [x] Create message files for all fields we will need translating openfoodfacts/openfoodfacts-auth#33
- [x] Finalize theming for all Keycloak screens that users will see openfoodfacts/openfoodfacts-auth#31
- [x] Configure signup / forgot password email in Keycloak openfoodfacts/openfoodfacts-auth#32
- [x] figure out security configuration for production deployment openfoodfacts/openfoodfacts-auth#34
- [x] work out what to do about staging, e.g. share keycloak with production or have separate instances and maybe copy database across periodically openfoodfacts/openfoodfacts-auth#35
- [x] Support social login providers openfoodfacts/openfoodfacts-auth#49
- [x] Support passkeys openfoodfacts/openfoodfacts-auth#50


cgi/oidc-callback.pl
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/perl -w
```
hangy, line 1: This is still a really early PoC. My intention is to move this to `Routing.pm` and `Display.pm` before publishing the PR draft, but for the PoC stage, using `cgi` was a bit easier than figuring out the "best" `Routing.pm`. 🤷

hangy: For now, I've decided not to move the files to ` Routing.pm` . That can still be done, if we decide to move all URLs to that structure. Right now, the required URL parameters seem to clash with the routing handling too much.

hangy:
This definitely isn't done, yet, but I'd love to know your feedback on the general approach, ie. field names, etc. @stephanegigandet @alexgarel @john-gom 

Missing, from the top of my head

- Manage more attributes in Keycloak (no handling for orgs, or teams, yet)
- User migration script

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+	my $create_user_request = HTTP::Request->new(POST => $keycloak_users_endpoint);
+	$create_user_request->header('Content-Type' => 'application/json');
+	$create_user_request->header('Authorization' => $token->{token_type} . ' ' . $token->{access_token});
+	$create_user_request->content($json);
+	my $new_user_response = LWP::UserAgent::Plugin->new->request($create_user_request);
+	unless ($new_user_response->is_success) {
+		display_error_and_exit($new_user_response->content, 500);
+	}
+
+	my $get_user_request = HTTP::Request->new(GET => $new_user_response->header('location'));
```
alexgarel, line 428:
```suggestion
	# continue the process by fetching user data,
	# which profile location is given in previous response
	my $get_user_request = HTTP::Request->new(GET => $new_user_response->header('location'));
```

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+	unless ($get_user_response->is_success) {
+		display_error_and_exit($get_user_response->content, 500);
+	}
+
+	my $json_response = $get_user_response->decoded_content(charset => 'UTF-8');
+	my @created_users = decode_json($json_response);
+	return $created_users[0];
+}
+
+sub create_user_in_product_opener ($id_token) {
```
alexgarel, line 441: Is this a fallback to having users created in keycloak ? What is the intent here ?

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+	if ((defined $email) and ($email =~ /\@/)) {
+		$emails_ref->{$email} = [$user_id];
+	}
+
+	if (defined $user_ref->{old_email}) {
+		delete $emails_ref->{$user_ref->{old_email}};
+		delete $user_ref->{old_email};
+	}
+
+	store("$BASE_DIRS{USERS}/users_emails.sto", $emails_ref);
```
alexgarel, line 468: It's a pity to have it repeated in User.pm, can we instead create a sub for that ?

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+
+Open ID Access token, or undefined if sign-in wasn't successful.
+
+=cut
+
+sub get_token_using_password_credentials ($username, $password) {
+	_ensure_oidc_is_discovered();
+
+	my $token_request = HTTP::Request->new(POST => $oidc_discover_document->{token_endpoint});
+	$token_request->header('Content-Type' => 'application/x-www-form-urlencoded');
```
alexgarel, line 496: Couldn't we use json instead for clarity ?

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+	}
+
+	my $access_token = decode_json($token_response->content);
+	$log->info('got access token from password credentials', {access_token => $access_token}) if $log->is_info();
+	return $access_token;
+}
+
+=head2 get_token_using_client_credentials()
+
+Gets a token for the user.
```
alexgarel, line 521: Is it a token for the user, or a token for the application (acting as a user)

Because in the request I don't see the user mentioned !

If it's so, I would add a note to explain that this token must remain private and only be used server side for specific operations (like creating a new user).

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+		'-domain' => $cookie_domain,
+		'-samesite' => 'Lax',
+	};
+
+	return cookie(%$cookie_ref);
+}
+
+=head2 verify_access_token($access_token_string)
+
+Verifies the access token by decoding and validating it using the JSON Web Key Set (JWKS).
```
alexgarel, line 592:
```suggestion
Verifies the access token by decoding and validating it using the JSON Web Key Set (JWKS).
(see https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets)
```

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+	unless ($id_token_verified) {
+		return;
+	}
+
+	return $id_token_verified;
+}
+
+=head2 get_azp($access_token)
+
+Retrieves the authorized party (client ID) from the access token.
```
alexgarel, line 638:
```suggestion
Retrieves the authorized party (client ID) from the access token.

It is different for example between the website and the mobile app.

This is useful for example for products change log.
```

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+				sub => $access_token->{sub}
+			}
+		) if $log->is_warn();
+		return;
+	}
+
+	return $access_token->{azp};
+}
+
+sub _get_client () {
```
alexgarel, line 672:
```suggestion
# return the OIDC client profile. 
# See https://metacpan.org/pod/OIDC::Lite::Client::WebServer
sub _get_client () {
```

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+	$client = OIDC::Lite::Client::WebServer->new(
+		id => $oidc_options{client_id},
+		secret => $oidc_options{client_secret},
+		authorize_uri => $oidc_discover_document->{authorization_endpoint},
+		access_token_uri => $oidc_discover_document->{token_endpoint},
+	);
+	return $client;
+}
+
+sub _ensure_oidc_is_discovered () {
```
alexgarel, line 687: Can you comment to explain a bit the role of this ?

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+
+	$oidc_discover_document = decode_json($discovery_response->content);
+	$log->info('got discovery document', {discovery => $oidc_discover_document}) if $log->is_info();
+
+	_load_jwks_configuration_to_oidc_options($oidc_discover_document->{jwks_uri});
+
+	return;
+}
+
+sub _load_jwks_configuration_to_oidc_options ($jwks_uri) {
```
alexgarel, line 710:
```suggestion
# JWKS aka JSON Web Key Sets are essential to validate access tokens
# https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets
sub _load_jwks_configuration_to_oidc_options ($jwks_uri) {
```

alexgarel (commented):
@hangy you've done an incredible job !

I didn't have the time to review it all, but here are some comments.

There is one point that we need to consider: we have old Product Opener Instances for the other projects.
Currently for user creation, we already are redirecting to openfoodfacts, and this could also be done for password reset.

Is there a strategy to bring minimal changes and support those old instances.

Of course we want to upgrade those instance this year, but it won't happen in a finger snap !

Dockerfile.keycloak
```diff
@@ -0,0 +1,20 @@
+# syntax = docker/dockerfile:1.2
+
+ARG DEPENDENCY_KEYCLOAK_VERSION=23.0.4
+
+FROM maven:3-eclipse-temurin-17 AS builder
+RUN set -x && \
+    git clone --single-branch --branch feat/support-v23 --depth=1 https://github.com/hangy/scrypt-password-hash-provider.git && \
```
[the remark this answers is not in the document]
hangy: 👍🏻 Yes, I wanted to suggest that *if* I manage to get this to work with existing hashes

lib/ProductOpener/API.pm
```diff
@@ -833,4 +837,103 @@ sub customize_response_for_product ($request_ref, $product_ref, $fields_comma_se
...
+	param('user_session', $user_session);
+	init_user($request_ref);
+
+	return;
+}
+
+=head2 _read_auth_header ( $request_ref, $r )
+
+Using the Authorization HTTP header, check if it looks like a 
+Bearer token, and if it odes, copy it to the request_ref
```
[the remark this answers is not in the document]
hangy: oops 😅

Dockerfile.keycloak
```diff
@@ -0,0 +1,20 @@
...
+    cd scrypt-password-hash-provider && \
+    mvn -B package
+
+ARG DEPENDENCY_KEYCLOAK_VERSION
```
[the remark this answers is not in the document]
hangy: I didn't know that, either. Didn't have to use dynamic image tags before, but I found this way somewhere online, and thought it was a good match for this use-case. :)

conf/keycloak/open-products-facts-realm.json
```diff
@@ -0,0 +1,2407 @@
...
+  "otpPolicyInitialCounter": 0,
+  "otpPolicyDigits": 8,
+  "otpPolicyLookAheadWindow": 1,
+  "otpPolicyPeriod": 30,
+  "otpPolicyCodeReusable": false,
+  "otpSupportedApplications": [
+    "totpAppFreeOTPName"
+  ],
+  "localizationTexts": {},
+  "webAuthnPolicyRpEntityName": "Open Products Facts",
```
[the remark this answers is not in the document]
hangy: To be honest, I'm not sure if we need to use this exact config to set Keycloak in prod. Might have to change some parameters, anyways, because of differences when running in developer mode and integrated H2 database, and probably PostgreSQL in production.

conf/keycloak/open-products-facts-realm.json
```diff
@@ -0,0 +1,2407 @@
+{
```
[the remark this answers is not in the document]
hangy: I totally missed that. 😀 I'm not super sure about editing this config (yet), because I just created it by exporting the realm that I configured in Keycloak.

conf/keycloak/open-products-facts-realm.json
```diff
@@ -0,0 +1,2407 @@
...
+      "directAccessGrantsEnabled": true,
+      "serviceAccountsEnabled": true,
+      "publicClient": false,
+      "frontchannelLogout": true,
+      "protocol": "openid-connect",
+      "attributes": {
+        "oidc.ciba.grant.enabled": "false",
+        "client.secret.creation.time": "1698609487",
+        "backchannel.logout.session.required": "true",
+        "post.logout.redirect.uris": "http://world.openfoodfacts.localhost/cgi/oidc-signout-callback.pl",
```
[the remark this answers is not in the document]
hangy: Fair point. I think I'll at least have a look, and see how well that works

lib/ProductOpener/API.pm
```diff
@@ -833,4 +837,103 @@ sub customize_response_for_product ($request_ref, $product_ref, $fields_comma_se
...
+
+	my $access_token;
+	eval {$access_token = verify_access_token($token);};
+	my $error = $@;
+	if ($error) {
+		$log->info('Access token invalid', {token => $token}) if $log->is_info();
+	}
+
+	unless ($access_token) {
+		add_error(
```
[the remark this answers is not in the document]
hangy: I only added `process_auth_header` to v3 for now, anyways. Do you think we should look into adding it to older API versions, too? I was thinking that it's the API version most likely to be used by applications that will easily integrate OAuth/OIDC, anyways.

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+	}
+
+	my $token = get_token_using_client_credentials();
+	unless ($token) {
+		display_error_and_exit('Could not get token to manage users with keycloak_users_endpoint', 500);
+	}
+
+	my $api_request_ref = {
+		email => $user_ref->{email},
+		emailVerified => $JSON::PP::true,    # TODO: Keep this for compat with current register endpoint?
```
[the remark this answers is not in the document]
hangy: We briefly touched this topic on Slack. One issue with email verification is that this breaks the "create user" API, which is currently used by some apps/SDKs. [Test case here](https://github.com/openfoodfacts/openfoodfacts-server/blob/9db5cd4c616b457681cea8b9567fa24a9459fa98/tests/integration/create_user.t). Currently, the assumption after successfully posting to `cgi/user.pl` is to get an active user that can be used to sign in and edit products directly. Enabling email verification would break with that idea. As far as I know, we don't have a way to communicate that changed behaviour to end-users right now.

cgi/oidc-signin-callback.pl
```diff
@@ -0,0 +1,51 @@
...
+# License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
```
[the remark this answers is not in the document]
hangy: Will do!

scripts/migrate_users_to_keycloak.pl
```diff
@@ -0,0 +1,136 @@
...
+use URI::Escape::XS qw/uri_escape/;
+
+sub create_user_in_keycloak_with_scrypt_credential ($user_ref, $credential) {
+	my $keycloak_users_endpoint = $oidc_options{keycloak_users_endpoint};
+	unless ($keycloak_users_endpoint) {
+		die 'keycloak_users_endpoint not configured';
+	}
+
+	my $token = get_token_using_client_credentials();
+	unless ($token) {
```
hangy, line 43: tbd: Reuse this token until it's close to expiring. Getting a new token for each interaction works in dev, but won't if we want to migrate thousands of users.

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+
+# Initialize some constants
+
+my $cookie_name = 'oidc';
+my $cookie_domain = '.' . $server_domain;    # e.g. fr.openfoodfacts.org sets the domain to .openfoodfacts.org
+$cookie_domain =~ s/\.pro\./\./;    # e.g. .pro.openfoodfacts.org -> .openfoodfacts.org
+if (defined $server_options{cookie_domain}) {
+	$cookie_domain
+		= '.' . $server_options{cookie_domain};    # e.g. fr.import.openfoodfacts.org sets domain to .openfoodfacts.org
+}
```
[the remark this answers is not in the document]
hangy: Agreed. I'll try to find a good place for that function, and replace the one in `Users.pm`, too

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+
+=head4 A reference to a hash containing request information. $request_ref
+
+=head3 Return Values
+
+None
+=cut
+
+sub start_authorize ($request_ref) {
+	my $nonce = generate_token(64);
```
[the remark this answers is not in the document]
hangy: It's basically used like a CSRF token. The relying party only accepts the callback if the same value as this nonce (stored in a cookie) so it doesn't get tricked into accepting a callback from some random source

lib/ProductOpener/Auth.pm
hangy: I think form-encoded data is usually required by OAuth2 token endpoints

lib/ProductOpener/Auth.pm
hangy: On the contrary. `create_user_in_product_opener` is called by `get_user_id_using_token` if the user that's signing in from Keycloak doesn't exist in Product Opener, yet: https://github.com/openfoodfacts/openfoodfacts-server/blob/b5fde3dc88ffcf41c698bc4eda84da11b9d626ae/lib/ProductOpener/Auth.pm#L282-L291

The idea is that we probably still need the `sto` files for users for some basic functionality (ie. user facets?), unless we want to rewrite _everything_ to implement this change. So if the user is new, create the `sto` file for the new user, based on data from Keycloak. Am I making any sense at all? 😅🤣

lib/ProductOpener/Auth.pm
hangy: Generally, there's probably no reason to expose any tokens that exist in the server's RAM. Even the tokens that I'm currently logging for development purposes should be redacted later, so that they're not exposed in logs.

This specific token uses the server's client secret, and should definitely be kept private, because it will have admin-like permissions to create/edit users in keycloak, because of the `create_user_in_keycloak` sub.

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+=head4 A reference to a hash containing request information. $request_ref
+
+=head3 Return values
+
+The return URL after successful authentication.
+=cut
+
+sub signin_callback ($request_ref) {
+	if (not(defined cookie($cookie_name))) {
+		start_authorize($request_ref);
```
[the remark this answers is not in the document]
hangy: Yup, there might be an issue if a users disables cookies entirely. I'll look into displaying something useful.

Dockerfile.keycloak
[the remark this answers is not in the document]
hangy: [Fixed](https://github.com/hangy/scrypt-password-hash-provider/commit/a39e21d56e0acf4b1717bde7d17f80a498169801). I don't have permission to move this to transfer this to the [org](https://github.com/openfoodfacts/) though

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,743 @@
...
+		return;
+	}
+
+	my $code = single_param('code');
+	my $state = single_param('state');
```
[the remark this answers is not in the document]
hangy: What's the current best practice? Add the values to `$request_ref`, or use normal method parameters?

lib/ProductOpener/Auth.pm
hangy: Added a comment to the sub.

hangy:
> There is one point that we need to consider: we have old Product Opener Instances for the other projects. Currently for user creation, we already are redirecting to openfoodfacts, and this could also be done for password reset.
> 
> Is there a strategy to bring minimal changes and support those old instances.
> 
> Of course we want to upgrade those instance this year, but it won't happen in a finger snap !

Great question. To be honest, my idea was to update all sites at once, but maybe we can figure something out to enable a slower rollout.

As long as the `users` directory is there, existing users will still be able to sign in on old sites, that aren't connected to Keycloak, yet. However, the passwords of new users won't be exposed from Keycloak, and changed passwords would also likely use an even more secure hashing algorithm than scrypt, and that would not be reflected by to OFF to the `users` directory. (At least not right now.)

So, what could we do? If we know the commits of the other sites, and we can update the older sites at all, the `Auth::password_signin` method could be adapted to that, and could be used to provide password based user login?

john-gom:
Hi @hangy . Had you given any thought to how the change password flow would work? I don't think we can keep the existing form as changing the password via the API is not allowed. It looks like [Application Initiated Actions](https://github.com/keycloak/keycloak-community/blob/main/design/application-initiated-actions.md) might be an option.

In my other work we have used Auth0 and the only way that allow you to change your password is by generating a password reset email.

john-gom:
Are we thinking that we would move all user attributes into Keycloak (i.e. be able to completely delete the user.sto files?)

I think this might be cleaner to avoid things getting out of sync. Looking at the code I can see the following additional attributes being referenced in code but not sure if all of them are stored:

- preferred_language
- country
- newsletter
- org
- org_id
- requested_org
- display_barcode
- edit_link
- missions
- old_email
- pro
- moderator

I know it might seem worth putting this off until later, but I have found that storing data in two places can cause real confusion and ends up taking up a lot more time to fix things.

It looks like we can store relatively unlimited amounts by increasing the [default field size](https://stackoverflow.com/questions/44851052/max-size-of-custom-user-attribute-in-keycloak) although we would want to limit to some extent as the data is stored in the token.

hangy:
> Had you given any thought to how the change password flow would work? I don't think we can keep the existing form as changing the password via the API is not allowed. It looks like [Application Initiated Actions](https://github.com/keycloak/keycloak-community/blob/main/design/application-initiated-actions.md) might be an option.

In my mind, we'd try skinning the Keycloak UI with something like [keycloakify](https://www.keycloakify.dev/), and remove most user management UIs/code from ProductOpener. Then, we can just use everything built-in to Keycloak, and don't have to worry about hardening those forms in ProductOpener.

If we don't want to do that, the [Keycloak Admin REST API](https://www.keycloak.org/docs-api/23.0.4/rest-api/index.html) allows sending a password reset mail with `PUT /admin/realms/{realm}/users/{id}/execute-actions-email`, but would also allow forcefully setting a password with `PUT /admin/realms/{realm}/users/{id}/reset-password`.

hangy:
> Are we thinking that we would move all user attributes into Keycloak (i.e. be able to completely delete the user.sto files?)

Yes, that's probably a good idea. I'd like deleting the `sto` files, but some functionality (ie. Missions, `display_points`) still depends on those files, and could be out of scope for _this_ PR.

> I know it might seem worth putting this off until later, but I have found that storing data in two places can cause real confusion and ends up taking up a lot more time to fix things.

I agree. 

> It looks like we can store relatively unlimited amounts by increasing the [default field size](https://stackoverflow.com/questions/44851052/max-size-of-custom-user-attribute-in-keycloak) although we would want to limit to some extent as the data is stored in the token.

I'm not sure how to handle all the attributes, yet. A `moderator` flag sounds like it could be a user/realm role, rather than a generic user profile field, for example. I think there was also some discussion about possibly discontinuing the `org` feature in general.

Dockerfile.keycloak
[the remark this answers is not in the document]
alexgarel: I made you owner of our org so you can now transfer it !

lib/ProductOpener/API.pm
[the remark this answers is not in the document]
alexgarel: I think we should add it to v2 because we want the situation to improve for many and v3 is considered a not stable version of the API (we may change things without warning).
Some already developed app using v2 API might want to change the authentication mode. The open food facts mobile app for example use v2 and v3 at the same time.

lib/ProductOpener/Auth.pm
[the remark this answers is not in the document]
alexgarel: ok, thanks for adding the comment.

lib/ProductOpener/Auth.pm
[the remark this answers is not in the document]
alexgarel: It depends on where you need the values. If you use them immediately, add them as method parameters. If you tend to use them more deeply in the code, use $request_ref. In short, use method parameters if it's not cumbersome.
I think using $request_ref here make sense.

lib/ProductOpener/Auth.pm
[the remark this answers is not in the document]
alexgarel: You are right. 

Could we, none the less, add a parameter to ask for email validation ? (so default would be no validation, but we could use validation, eg. on the web).
Do it only if it's easy to do.

lib/ProductOpener/Auth.pm
alexgarel: Yes we want to keep sto file for now you are right.

So we just had to add documentation here to make it clear as you did :-)

lib/ProductOpener/Config2_docker.pm
```diff
@@ -118,6 +119,20 @@ $events_url = $ENV{EVENTS_URL};
 $events_username = $ENV{EVENTS_USERNAME};
 $events_password = $ENV{EVENTS_PASSWORD};
 
+%oidc_options = (
```
alexgarel, line 122: We have a lot of options to move to environment here :-)

alexgarel (commented):
All reviewed :-)

That's a fantastic move ! It's does not disturb the code that much, that's a good point.

I will now test it locally.

lib/ProductOpener/Users.pm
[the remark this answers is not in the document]
hangy: Yep good point!

scripts/migrate_users_to_keycloak.pl
```diff
@@ -0,0 +1,154 @@
...
+	my $request_token = get_token_if_we_dont_have_one_yet_or_it_is_expired();
+	my $create_user_request = HTTP::Request->new(POST => $keycloak_users_endpoint);
+	$create_user_request->header('Content-Type' => 'application/json');
+	$create_user_request->header(
+		'Authorization' => $request_token->{token_type} . ' ' . $request_token->{access_token});
+	$create_user_request->content($json);
+	my $new_user_response = LWP::UserAgent::Plugin->new->request($create_user_request);
+	unless ($new_user_response->is_success) {
+		# TODO: Log?
+		die $new_user_response->content;
```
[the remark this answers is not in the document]
hangy: We might want to analyze the result and decide based on the status - no need to continue with other users if Keycloak went down, for example

lib/ProductOpener/Users.pm
```diff
@@ -1051,6 +1072,38 @@ sub init_user ($request_ref) {
...
+		if (defined $user_id) {
+
+			my $user_file = "$BASE_DIRS{USERS}/" . get_string_id_for_lang("no_language", $user_id) . ".sto";
+
+			if (-e $user_file) {
+				$user_ref = retrieve($user_file);
+				$user_id = $user_ref->{'userid'};
+				$log->context->{user_id} = $user_id;
+
+				if (not defined request_param($request_ref, 'no_log')) # no need to store sessions for internal requests
```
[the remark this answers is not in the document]
hangy: I stole it from here 😅 https://github.com/openfoodfacts/openfoodfacts-server/blob/acf96bb5625f02ea1bc3bc0bcbdb82153a896517/lib/ProductOpener/Users.pm#L1111

Unless I'm mistaken, the idea is to not store the session in the `sto` file if they present the user+password, or an OIDC token (for the API request), because they won't use the session cookie in that way, anyways.

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,915 @@
...
+
+=head4 The username for password-based authentication. $username
+=head4 The password for password-based authentication. $password
+
+=head3 Return Values
+
+A list containing the user's ID, refresh token, refresh token expiration time, access token, and access token expiration time.
+=cut
+
+sub password_signin ($username, $password) {
```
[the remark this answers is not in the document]
hangy: According to https://lists.jboss.org/pipermail/keycloak-user/2017-January/008954.html, we just need to set the default/policy hashing algorithm in Keycloak to another one, then it should do so automatically.

I'd see removing the password from the  `sto` files as a separate step that we can do after the migration.

lib/ProductOpener/Users.pm
```diff
@@ -966,13 +987,13 @@ sub retrieve_user ($user_id) {
 sub is_email_has_off_account ($email) {
 
 	# First, check if the email exists in the users_emails.sto file
-	my $emails_ref = retrieve("$data_root/users/users_emails.sto");
+	my $emails_ref = retrieve("$BASE_DIRS{USERS}/users_emails.sto");
```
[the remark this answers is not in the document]
hangy: No problem :)

alexgarel:
> Looking at the code I can see the following additional attributes being referenced in code but not sure if all of them are stored:


I think there are different types of attributes:
* user attributes (like preferred_language, country, requested_org, org, org_id, pro, moderator)
* application specific attributes (preferences): display_barcode, edit_link
* preferences attributes we don't want to keep for now: old_email, missions

I would try to store user attributes in keycloak (and the list might be discussed) but preferences on website side.

alexgarel:
Some tests:

* user creation
  * did use the keycloak register form and it's not working because email is not verified ! 
    * The problem is that I'm blocked with "signin" on the website which send to signin-callback and I can't sign out ! (I had to go to keycloak admin and logout all sessions)
  * using the signin form of product opener, works fine
    * but I had to enter the url by hand (no link on the website)
* login / logout ok (in normal situation)

alexgarel:
Test on migrate_users_to_keycloak.pl:

Runs ok and created user fine.

Ended with a `{"errorMessage":"User exists with same email"}` but probably because I recreated an existing user before migration (my bad)

I can login with an imported users.

Tested looking at a user account using an admin account, it works.

alexgarel:
A point that needs handling also is email change.

Also username change, could be a plus to rename some users that put their email address on it !

hangy:
> A point that needs handling also is email change.

Oh. Yes that could be an issue, because I use the mail address to look up users right now. Maybe I should just use the username directly.

> Also username change, could be a plus to rename some users that put their email address on it !

Usernames aren't editable in OFF, or Keycloak right now. Since it's basically the primary key for the user, and is referenced in change logs, and so on, should it be editable?

Dockerfile.keycloak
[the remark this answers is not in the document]
hangy: Done. Thank you!

lib/ProductOpener/Users.pm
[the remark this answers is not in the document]
hangy: Is it correct that the anonymization doesn't work across sites right now? If I delete an account in OBF, but the account isn't modified in OFF?

conf/keycloak/open-products-facts-realm.json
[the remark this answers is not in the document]
hangy: Moved some settings to ENV vars.

lib/ProductOpener/Auth.pm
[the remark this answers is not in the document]
hangy: Parameter added. Defaulted to false for now.

lib/ProductOpener/Auth.pm
hangy: Auth.pm now uses a bit different approach now, because I changed the mail handling for users from OIDC.

lib/ProductOpener/Config2_docker.pm
hangy: Fixed

lib/ProductOpener/Config_off.pm
```diff
@@ -378,6 +379,7 @@ $facets_kp_url = $ProductOpener::Config2::facets_kp_url;
 # server options
 
 %server_options = %ProductOpener::Config2::server_options;
+%oidc_options = %ProductOpener::Config2::oidc_options;
```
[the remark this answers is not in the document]
hangy: Done

lib/ProductOpener/Auth.pm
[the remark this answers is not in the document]
hangy: While I generally agree, the `signin_callback` sub also needs to use `param` to set stuff for `init_user`, I think I'm going leave it as-is for now. Changing the method signature of `init_user` affects too many other places for now, in my opinion.

lib/ProductOpener/API.pm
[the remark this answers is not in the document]
hangy: Should now work for API v2, including image upload.

lib/ProductOpener/Users.pm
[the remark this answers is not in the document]
alexgarel: Yes, you are right… we should implement this (on a daily bases for example, anonymize removed accounts)

docker/dev.yml
```diff
@@ -99,6 +98,29 @@ services:
...
+      - KEYCLOAK_ADMIN_PASSWORD
+      - KEYCLOAK_BASE_URL
+      - KEYCLOAK_REALM_NAME
+      - PRODUCT_OPENER_DOMAIN
+      - PRODUCT_OPENER_OIDC_CLIENT_ID
+      - PRODUCT_OPENER_OIDC_CLIENT_SECRET
+    volumes:
+      - keycloak_data:/opt/keycloak/data
+      - ./conf/keycloak:/opt/keycloak/data/import
+    command: start-dev --import-realm --features="account3,declarative-user-profile"
```
john-gom, line 115: Could we import the realm in the base image so that we keep the keycloak functionality more self-contained?

hangy: I think the realm needs to be different for local dev, staging, and production setup. This one's useful for local development, hence why I map it in the dev branch, but we shouldn't import one with static secrets for security sensitive environments. The base setup for staging/production might have a place in https://github.com/openfoodfacts/openfoodfacts-infrastructure/tree/ddf4070c9ae9188af4e263c12cf1aab5dfe7b18b/confs, though?

scripts/migrate_users_to_keycloak.pl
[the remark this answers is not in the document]
hangy: Main idea is to just use the batch import now.

scripts/migrate_users_to_keycloak.pl
```diff
@@ -0,0 +1,154 @@
...
+
+	return;
+}
+
+if (opendir(my $dh, "$BASE_DIRS{USERS}/")) {
+	foreach my $file (readdir($dh)) {
+		if (($file =~ /.+\.sto$/) and ($file ne 'users_emails.sto')) {
+			migrate_user("$BASE_DIRS{USERS}/$file");
+		}
+	}
```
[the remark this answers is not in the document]
hangy: I've mostly moved the script to use a batch upload. That will fail, and won't allow checking all users but if we have batches of a few thousand users, that should be doable IMHO.

lib/ProductOpener/Users.pm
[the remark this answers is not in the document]
hangy: I'm not 100% sure about the meaning of  your comment. Should that be part of this PR, or elsewhere?

There's no way to actually sync that action to all ProductOpener instances right now, correct?

john-gom:
Hi All. I'm not happy about the way that keycloak is handling the locale and other pick lists, like country. It doesn't seem to do any kind of logical sorting and also doesn't support typeahead search, which is pretty important given the size of our country and language lists. I can think of three ways around this (so far):

1. Fork Keycloak itself, make the changes and raise a PR in the hope that this is accepted. Concern would be keeping up to date if our PR is not accepted, as we need to 
2. Create our own themes (currently I've just been tweaking CSS). This could require us to start from scratch on the main screens that users see, e.g. using keycloakify.
3. Add a JavaScript hack to our simple theme that dynamically modifies the page content to sort properly, etc. Possibly less work but could be quite brittle

All have their pros and cons. Does anyone have any preference based on past experience? cc. @hangy , @alexgarel , @stephanegigandet , @raphael0202

hangy:
> Does anyone have any preference based on past experience?

I don't have any experience with Keycloak prior to this PR, so I can't tell how complicated the templates are, or how willing they're when it comes to PRs, sorry.

john-gom:
> As discussed in PO meeting today, I'll create a fork (https://github.com/john-gom/keycloak) and see how willing the Keycloak team are to accept a PR

I've added some discussion to https://github.com/keycloak/keycloak/issues/17776 and https://github.com/keycloak/keycloak/discussions/19191 on these subjects...

john-gom:
> I've added some discussion to [keycloak/keycloak#17776](https://github.com/keycloak/keycloak/issues/17776) and [keycloak/keycloak#19191](https://github.com/keycloak/keycloak/discussions/19191) on these subjects...

Have raised the following PRs in keycloak: https://github.com/keycloak/keycloak/pull/27737 and https://github.com/keycloak/keycloak/pull/27726

Some tips when building keycloak...

To do an incremental build with -pl you need to include the and quarkus/deployment and quarkus/dist projects, e.g.

```./mvnw -DskipTests -DskipTestsuite -DskipExamples -pl themes,quarkus/deployment,quarkus/dist -am install```

**Comment by john-gom** [comment]
Adding feedback from @teolemon here so I can track it:

List of issues and suggestions for improvement [Screenshots](https://photos.google.com/share/AF1QipOr2jgPQBrbZa8ZXfLjZRSxo_aBtYYMQX_4XlSlrGUlBqRcUJrsvYs4qxjS5z8sag?key=NGFzbTZ4b0JHX3dKcTZUczk2dGhDS1FRRnZQYUR3):
Login page:
- [ ] The “Back to login” button should be placed at the top of the view.
- [ ] The country and the Open Food Facts logo overlap.
- [ ] The “Register” button should be “Create an Open Food Facts account”.
- [ ] “name” does not have a capital letter.
- [ ] The password reset instructions should be just below the title and above the form.
- [ ] Page titles should be wider and in bold.
- [ ] There should be a shortcut to common email clients (Gmail, Hotmail).
- [ ] Use Matomo to track all steps of the login, signup, and reset process (needed to spot issues quickly and early)
- [ ] Fix the bug that sets the country to “Perensis” by default on Autofill (not sure what value provoked that, probably a value starting with P ?)
- [ ] The eye to hide the password goes beyond the red error line of the field.
- [ ] If the email is entered as a username, the cause of the error should be indicated more clearly (currently: “invalid format”).
- [ ] Display the validity of the username in real time (while mitigating abuse) to save a lot of time to the user (validating several times, loosing input)
- [ ] The password field is emptied without warning when an invalid form is submitted (submission errors loop for the unattentive user)
 Center:
- [ ] Display the user’s first name in the website Account Parameters
- [ ] List all possible actions in the User Center in the account settings
- [ ] Center the 1st spinner when loading the “User Center”.
- [ ] Change the spinner color to brown/beige according to the graphic charter.
- [ ] Improve the readability of the text in the “User Center”.
- [ ] _Keycloak PR https://github.com/keycloak/keycloak/pull/28512._ Allow filtering the country by the first characters (time saving, esp on mobile)
- [ ] Mention that the username cannot be changed and explain why (signup and user center)
Account deletion
- [ ] When deleting the account, mention that the contributed products to the OFF common will be reassigned to the anonymous user.
- [ ] Account Deletion: It is very disturbing to see your username still visible when I thought I had deleted it. And I’m wondering if I should log in. There is no explanation that it is necessary to log in to finalize the deletion. I really did it by chance.
- [ ] The delete confirmation buttons overlap between “confirm deletion” and “cancel”.
Improve the account deletion experience:
- [ ] Wider text.
- [ ] Button vs link to return to the OFF homepage.
- [ ] Illustration of the orange saying goodbye.
- [ ] Optional survey on the reasons for the deletion.
- [ ] Redirect to the Open Food Facts home page after deleting the account.

File: docker/dev.yml [hunk 34]
```diff
@@ -27,6 +27,9 @@ x-backend-conf: &backend-conf
   # e.g. when developing with Robotoff or openfoodfacts-query
   extra_hosts:
     - "host.docker.internal:host-gateway"
+    - "auth.openfoodfacts.localhost:host-gateway"
```
hangy, line 30: Note that there's already an alias for `auth.${PRODUCT_OPENER_DOMAIN}` assigned to the nginx `frontend` container @john-gom

john-gom: Hi @hangy , yes I had seen that but I didn't want the Keycloak deployment to depend on Product Opener, so I wanted the Keycloak URLs (and ports) to be the same whether running via Product Opener or just stand-alone.

scripts/migrate_users_to_keycloak.pl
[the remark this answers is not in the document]
hangy: Batch import with ~400k users took ~45 minutes (see updated https://github.com/openfoodfacts/openfoodfacts-server/pull/9251#issue-1979948357)

lib/ProductOpener/Users.pm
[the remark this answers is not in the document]
hangy: Related to https://github.com/openfoodfacts/openfoodfacts-server/pull/9251#issuecomment-2042423273

> Account Deletion: It is very disturbing to see your username still visible when I thought I had deleted it. And I’m wondering if I should log in. There is no explanation that it is necessary to log in to finalize the deletion. I really did it by chance.

The Minion job had probably not finished, yet. (Difficult to test right now, because ProductOpener and Keycloak aren't connected to the same Redis instance right now.) There's one big issue with this in general: If we delete user info in Keycloak directly, when they request it, and the data in ProductOpener is only deleted in a scheduled background job, then a user with the same name might register in Keycloak while the user info and product attribution are still being worked on in Minion/ProductOpener. Maybe we have to rethink this and only disable users in Keycloak until the background job has finished deleting user info and products?

hangy:
Test `tests/integration/delete_user.t` fails intermittently. It works at least sometimes with more than one execution. I fear I need to debug the keycloak->redis->minion chain a bit more

john-gom:
> Test `tests/integration/delete_user.t` fails intermittently. It works at least sometimes with more than one execution. I fear I need to debug the keycloak->redis->minion chain a bit more

Maybe testing the whole flow is a bit much and we should test in smaller sections with mocking downstream stuff?

Makefile
```diff
@@ -50,28 +50,34 @@ endif
 HOSTS=127.0.0.1 world.productopener.localhost fr.productopener.localhost static.productopener.localhost ssl-api.productopener.localhost fr-en.productopener.localhost
 # commands aliases
 DOCKER_COMPOSE=docker compose --env-file=${ENV_FILE} ${LOAD_EXTRA_ENV_FILE}
-# docker command that do not need the shared network
-DOCKER_COMPOSE_BUILD=COMPOSE_FILE="${COMPOSE_FILE_BUILD}" ${DOCKER_COMPOSE}
+# Ensure shared_network is referenced when running locally
+DOCKER_COMPOSE_RUN=COMPOSE_FILE="${COMPOSE_FILE};docker/run.yml" ${DOCKER_COMPOSE}
```
alexgarel, line 54: @hangy I think there might have been a merge problem.
I think we did this on the main branch and revert it after.
So I don't think you need this change, as well as other changes that are underneath.

docker-compose.yml
```diff
@@ -103,8 +110,14 @@ services:
...
     <<: *backend-conf
-    # This service watch for new images to trigger ocr and robotoff processing
+    # This service watches for new images to trigger ocr and robotoff processing
     command: ["perl", "scripts/run_cloud_vision_ocr.pl", "/mnt/podata/new_images"]
+  redis-listener:
+    <<: *backend-conf
+    # we want the redis_listerner for off, but not for pro profile
+    profiles: ["off"]
+    # This service watches for new events on a redis stream to delete users
+    command: ["perl", "scripts/listen_to_redis_stream.pl"]
```
alexgarel, line 120: @john-gom this means we have to define a new systemd service in prod to do this ?

john-gom: Hi @alexgarel . Yes, I had created an issue in the project here: https://github.com/orgs/openfoodfacts/projects/119?pane=issue&itemId=89376373&issue=openfoodfacts%7Copenfoodfacts-server%7C11082

hangy:
I don't know what to do here, tbh. I can't put enough resources into this PBI right now to solve newer requirements like openfoodfacts/openfoodfacts-auth#157, and the branch is beginning to diverge from main, again. What's [issue](https://github.com/openfoodfacts/openfoodfacts-auth/issues) strictly necessary to actually put this into production, and what's "nice to have" and basically an improvment over the the current login/sign-up process?

john-gom:
Hi @hangy . I've been having similar concerns. I'm wondering whether we should try to feature flag this somehow so we can deploy without moving everyone over. So maybe add a URL parameter like `keycloak=1` that sets a cookie so that Keycloak is used going forward in that browser until it is removed with `keycloak=0`.  I'm also thinking it might be best to have both systems in production for a while so we can keep the users in sync as otherwise we could have downtime while we do the import.

hangy:
In hindsight, it would've been way better to have some kind of feature toggle for the OIDC integration to avoid such a long-lived feature branch. Truthfully, I didn't think it'd live this long, though. I'm not sure a short downtime would warrant the development effort to reenable the current password auth, and user page conditionally. However, if we can't come up with a list of actions (+ must haves, blockers, …), and it helps to get this branch merged, I'm all for it. 👍🏻 Regularly solving merge conflicts and updating integration tests without a clear path forward for this long isn't too motivating, tbh. 😅

lib/ProductOpener/APITest.pm
```diff
@@ -101,8 +101,13 @@ sub wait_auth() {
 	my $count = 0;
 	my $ua = new_client();
 	my $target_url = construct_test_url("");
+	my $discovery_endpoint
```
hangy, line 104: IIRC, at least in my first tests, we needed the public URL to be used for the discovery endpoint, so that the URL could be used for redirects. Otherwise, the backchannel URL (internal docker net) was used for public URLs, and thus not really working from the internet @john-gom

john-gom: I checked the response and it seemed to be returning the public URL for those kinds of thing, but will double check

lib/ProductOpener/Users.pm
```diff
@@ -997,27 +1177,98 @@ sub store_user_session ($user_ref) {
...
 sub store_user ($user_ref) {
-	my $userid = $user_ref->{userid};
+	my $oidc_implementation_level = get_oidc_implementation_level();
+	if ($oidc_implementation_level < 5) {
+		# We are still using legacy code for registration
+
+		if ($oidc_implementation_level > 0) {
+			# Sync the user with Keycloak
+			my $keycloak = ProductOpener::Keycloak->new();
+			$keycloak->create_or_update_user($user_ref);
```
hangy, line 1187: This could be intentional, but we should discuss it real quick.

`store_user` is called by `update_login_time`, which is called after the password was successfully validated in `init_user`. Especially for some apps and API clients, the user+password tuple is sent with every request, causing the login time to be updated on each request. `ProductOpener::Keycloak::create_or_update_user` is relatively heavy as it executes several REST requests. Of the Keycloak API endpoints, those that update user records are probably not as well optimized as something like issuing a token.

I wonder, if that might be a performance bottleneck?

cgi/auth.pl
```diff
@@ -31,6 +31,7 @@
 use ProductOpener::Users qw/$User_id %User is_admin_user/;
 use ProductOpener::Lang qw/:all/;
 use ProductOpener::Tags qw/country_to_cc/;
+use ProductOpener::Auth qw/write_auth_deprecated_headers/;
```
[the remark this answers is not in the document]
hangy: Nope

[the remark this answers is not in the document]
hangy: I think this can be removed

lib/ProductOpener/Users.pm
john-gom: I think it should probably call store_user_session instead

hangy: Don't get me wrong: There are advantages of calling it upon login. Namely, active users get synced to Keycloak immediately. However, I tend to agree because of the performance impact.

john-gom: I don't think we should be updating the user in Keycloak on each login, only when their details change

lib/ProductOpener/Auth.pm
```diff
@@ -0,0 +1,878 @@
...
+=head3 Return values
+
+None.
+
+=cut
+
+sub write_auth_deprecated_headers() {
+	if (get_oidc_implementation_level() >= 3) {
+		my $r = Apache2::RequestUtil->request();
+		$r->err_headers_out->set('Deprecation', 'Mon, 01 Apr 2024 00:00:00 GMT');
```
hangy, line 855: TODO: Actual values need to be discussed. Maybe only after the OIDC level is raised to 3

cgi/login.pl
```diff
@@ -31,69 +31,110 @@
...
+	if ($ENV{'REQUEST_METHOD'} eq 'POST') {
+		my $user_ref = retrieve_user($User_id);
+		if (not(defined $user_ref)) {
+			push @errors, 'undefined user';
+			$template_data_ref->{success} = 0;
+		}
+
+		my $hash_is_correct
+			= check_password_hash(encode_utf8(decode utf8 => single_param('password')),
+			$user_ref->{'encrypted_password'});
```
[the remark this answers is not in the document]
john-gom: Perl tidy does not allow this

lib/ProductOpener/Redis.pm
```diff
@@ -74,6 +117,196 @@ sub init_redis() {
...
+
+			if ($reply_ref) {
+				# Process any received messages
+				my $last_processed_message_id = process_xread_stream_reply($reply_ref);
+				if ($last_processed_message_id) {
+					$search_from = $last_processed_message_id;
+				}
+			}
+
+			# Start listening for the next batch of messages
```
john-gom, line 185: Think we might need a sleep here as I noticed on opff-new that thew redis_listener service was taking 100% CPU

lib/ProductOpener/Redis.pm
```diff
@@ -74,6 +117,196 @@ sub init_redis() {
...
+		# Read Keycloak events to process actions following user creation / deletion
+		_read_user_streams('$');
+	}
+
+	return;
+}
+
+sub _read_user_streams($search_from) {
+	# Listen for user-deleted events so that we can redact product contributions for this flavor
+	my @streams = ('BLOCK', 0, 'STREAMS', 'user-deleted');
```
john-gom: We should maybe block for a while here and also add a count (same as off-query which uses COUNT 1000 BLOCK 5000)

.env
```diff
@@ -76,6 +76,16 @@ ODOO_CRM_PASSWORD=
 # you might also use the .net service
 NUTRIPATROL_URL=http://nutripatrol.localhost
 
+# Keycloak configuration
+# See Auth.pm for definition of levels
+OIDC_IMPLEMENTATION_LEVEL=1
+OIDC_DISCOVERY_URL=http://keycloak:8080/realms/open-products-facts/.well-known/openid-configuration
+OIDC_CLIENT_ID=OFF
+OIDC_CLIENT_SECRET=Cf4NdSAjZsNO9HLcuXeuvukzFu00roQa
```
alexgarel:
```suggestion
# IMPORTANT: change this on public deployments
OIDC_CLIENT_SECRET=Cf4NdSAjZsNO9HLcuXeuvukzFu00roQa
```

cgi/change_password.pl
```diff
@@ -40,64 +41,71 @@
...
 }
-
-$template_data_ref->{errors} = \@errors;
-
-my $html;
-process_template('web/pages/change_password/change_password.tt.html', $template_data_ref, \$html) or $html = '';
-if ($tt->error()) {
-	$html .= '<p>' . $tt->error() . '</p>';
+else {
+	my $redirect = ProductOpener::Keycloak->new()->get_account_link($request_ref->{canon_url});
```
alexgarel, line 110:
```suggestion
	# redirect to keycloak for authentication
	my $redirect = ProductOpener::Keycloak->new()->get_account_link($request_ref->{canon_url});
```

cgi/reset_password.pl
alexgarel, line 1: shan't we disable it if get_oidc_implementation_level() == 5 ?

docker/integration-test.yml
alexgarel, line 1: This is smart :-)

env/env.off-pro
```diff
@@ -17,3 +17,10 @@ APACHE_ENVVARS=/etc/apache2/off-envvars
 # app configurations
 PRODUCT_OPENER_FLAVOR=openfoodfacts
 PRODUCT_OPENER_FLAVOR_SHORT=off
+
+# We don't enable the redis listener for the pro-platform so this is not strictly needed
+PROCESS_GLOBAL_REDIS_EVENTS=0
+
+# Use a different OIDC client for the Pro platform
+OIDC_CLIENT_ID=OFF_PRO
+OIDC_CLIENT_SECRET=1ILwRhBtNJJDQrHp8lR7yrP9S0N2W7al
```
alexgarel: IMPORTANT: Those are real file deployed in production, we can't put the OIDC_CLIENT_SECRET here !

Maybe we should add a env.secrets.off-pro on servers and change setenv to also source those files if they exists.

lib/ProductOpener/Config_off.pm
```diff
@@ -456,6 +459,7 @@ $rate_limiter_blocking_enabled = $ProductOpener::Config2::rate_limiter_blocking_
...
```
alexgarel: nitpick: We can now do things directly in Config.pm using Env.

lib/ProductOpener/Redis.pm
```diff
@@ -74,6 +117,196 @@ sub init_redis() {
...
+	# If this process handles global events then also listen for the user-registered events from Keycloak
+	if ($process_global_redis_events) {
+		push(@streams, 'user-registered');
+		push(@streams, $search_from);
+	}
+	# Note the message index to search from is provided at the end of the list in the same order as the list of keys
+	push(@streams, $search_from);
```
alexgarel, line 165: aren't we pushing twice the $search_from value here ?

lib/ProductOpener/Users.pm
```diff
@@ -997,27 +1185,98 @@ sub store_user_session ($user_ref) {
...
+		my $email = $user_ref->{email};
+
+		if ((defined $email) and ($email =~ /\@/)) {
+			$emails_ref->{$email} = [$userid];
+		}
+		if (defined $user_ref->{old_email}) {
+			delete $emails_ref->{$user_ref->{old_email}};
+			delete $user_ref->{old_email};
+		}
+		store("$BASE_DIRS{USERS}/users_emails.sto", $emails_ref);
```
alexgarel, line 1212: nitpik: we could avoid this as soon as $oidc_implementation_level == 5 and remove it after.

alexgarel:
@john-gom I use auth url to sign in keykloak (getting an old url from previous attempt), it came back to https://world.new.openpetfoodfacts.org/cgi/oidc_signin_callback.pl but I got an error:

```
Undefined subroutine &ProductOpener::Auth::lang called at /srv/opff/lib/ProductOpener/Auth.pm line 157.
```

alexgarel:
I did a test on https://world.new.openpetfoodfacts.org/ which is configured as level 1, with verification of what happens in keycloak:
on the web:
* creating a user
* resetting password
* loging in logout
* modify email
* removing user
Using mobile app:
* creating a user
* (resetting password, no but it's because the mobile app tries to connect to world.new.openfoodfacts.org…)
* login and logout

It all worked fine :tada: 

The only problem I saw is that on the web I choose french for my preferred language but got the email in english (language of the website). I think this is a pre-existing problem and not a big deal.

lib/ProductOpener/Redis.pm
john-gom: We track both streams from the same last message index. So if the last message index for either stream was 123 then we search stream_a from 123 and stream_b from 123 but the API requires us to list all the streams and then all of the indexes, so "stream_a streamb 123 123"

env/env.off-pro
john-gom: I'm thinking the simplest option initally might be to prefix the env settings with the flavor plus producers platform flag, e.g. OFF_OIDC_CLIENT_SECRET, OPFF_OIDC_CLIENT_SECRET, OFF_PRO_OIDC_CLIENT_SECRET
We could potentially get rid of the client id and automatically set this based on the flavour too

lib/ProductOpener/Config_off.pm
john-gom: What is the process you go through to make sure that new environment settings are added to production?

john-gom:
> @john-gom I use auth url to sign in keykloak (getting an old url from previous attempt), it came back to https://world.new.openpetfoodfacts.org/cgi/oidc_signin_callback.pl but I got an error:
> 
> ```
> Undefined subroutine &ProductOpener::Auth::lang called at /srv/opff/lib/ProductOpener/Auth.pm line 157.
> ```

The specific error was caused by missing lang from the use ProductOpener::Lang declaration. But it was ultimately still just reporting an error "oidc_signin_no_cookie". The cookie is generated by access_to_protected_resource before the Keycloak sign-in so I don't think you can really test this at level 1

lib/ProductOpener/Users.pm
john-gom: There is already a `if ($oidc_implementation_level < 5) {` block surrounding that (starts at line 1189)

env/env.off-pro
john-gom: Maybe just name it OFF_CLIENT_SECRET, etc. as that is consistent with the naming in the Keycloak repo too and could correspond with GitHub secrets

cgi/reset_password.pl
john-gom: Done

env/env.off-pro
john-gom: @alexgarel , I've made some more changes here. Do they look OK?

lib/ProductOpener/Users.pm
john-gom: I've made these changes now

lib/ProductOpener/Auth.pm
john-gom: I've shifted the dates out a bit for now

alexgarel (approved):
Thank you so so so much @hangy and @john-gom !

PR #9272, opened 2023-11-08
fix: count ingredients with specified percent

stephanegigandet:
Adds new misc tags that indicate if a product has some or all percent of ingredients specified. Useful in particular to create test sets for ingredients percent estimation.


benbenben2 (approved):
Looks good to me

teolemon:
ok to merge @stephanegigandet ?

PR #9276, opened 2023-11-09
feat: set max percent of sugar and salt ingredients based on nutrition facts

stephanegigandet:
A small improvement to ingredients percent estimation, made to test the new https://github.com/openfoodfacts/recipe-estimator-metrics framework.

Corresponding changes: https://github.com/openfoodfacts/recipe-estimator-metrics/pull/1/commits/32cb0585ed3e8b104b50ba1924c8e5c849ed4b9b

e.g. test summary differences on a small test set of 62 products:

test-sets/results/product_opener/fr-62-products-10-specified-ingredients/results_summary.json

image


stephanegigandet:
I added a new test set with 1000 products with some specified ingredients percent.

Before this PR:

Test set world-1000-some-specified
number of products: 1000
total difference:27906.457323507402
average difference: 27.906457323507404

With this PR:

Test set world-1000-some-specified
number of products: 1000
total difference:26289.049043323128
average difference: 26.28904904332313

lib/ProductOpener/Ingredients.pm
```diff
@@ -886,6 +887,60 @@ sub add_properties_from_specific_ingredients ($product_ref) {
...
+	my $sugars_100g = deep_get($product_ref, qw(nutriments sugars_100g));
+	if (defined $sugars_100g) {
+		push @ingredient_max_values, {ingredientid => "en:sugar", value => $sugars_100g};
+	}
+	my $salt_100g = deep_get($product_ref, qw(nutriments salt_100g));
+	if (defined $salt_100g) {
+		push @ingredient_max_values, {ingredientid => "en:salt", value => $salt_100g};
+	}
+
+	if ((defined $sugars_100g) or (defined $salt_100g)) {
```
alexgarel, line 909: why not directly the foreach ? this if does not bring anything.

lib/ProductOpener/Ingredients.pm
```diff
@@ -886,6 +887,60 @@ sub add_properties_from_specific_ingredients ($product_ref) {
...
+
+		foreach my $ingredient_max_value_ref (@ingredient_max_values) {
+
+			# Traverse the ingredients tree, breadth first
```
alexgarel, line 913: You put sub ingredients at the beginning of the list, so it's depth first.
```suggestion
			# Traverse the ingredients tree, depth first
```

lib/ProductOpener/Ingredients.pm
alexgarel, line 909: I find it more logical to put the condition on the item you will use in you foreach. And why would you need this if as an empty list means no foreach run.

```suggestion
	if (scalar @ingredient_max_value) {
```

lib/ProductOpener/Ingredients.pm
```diff
@@ -886,6 +887,60 @@ sub add_properties_from_specific_ingredients ($product_ref) {
...
+				my $ingredient_ref = shift @ingredients;
+				my $ingredientid = $ingredient_ref->{id};
+
+				# Add sub-ingredients at the beginning of the ingredients array
+				if (defined $ingredient_ref->{ingredients}) {
+
+					unshift @ingredients, @{$ingredient_ref->{ingredients}};
+				}
+
+				foreach my $ingredient_max_value_ref (@ingredient_max_values) {
```
alexgarel, line 929: We already iterated above ! Why do we need to iterate twice ?
(I prefer if we keep this inner loop, that said).

lib/ProductOpener/Ingredients.pm
```diff
@@ -886,6 +887,60 @@ sub add_properties_from_specific_ingredients ($product_ref) {
...
+					my $value = $ingredient_max_value_ref->{value};
+					if (is_a("ingredients", $ingredient_ref->{id}, $ingredient_max_value_ref->{ingredientid})) {
+						if (not defined $ingredient_ref->{percent_max}) {
```
alexgarel, line 932: why don't we do it if percent_max is above current max ?
```suggestion
						if ((not defined $ingredient_ref->{percent_max}) or ($value < $ingredient_ref->{percent_max})) {
```

lib/ProductOpener/Ingredients.pm
```diff
@@ -886,6 +887,60 @@ sub add_properties_from_specific_ingredients ($product_ref) {
...
+							$ingredient_ref->{percent_max} = $value;
+						}
+					}
+
+				}
+			}
+		}
+	}
```
alexgarel, line 940: Shan't we do a pass after that to eventually re-order the list of ingredients (or sub-ingredients) ?

alexgarel (approved):
I approve for urgency, but I think there are things to fix !

lib/ProductOpener/Ingredients.pm
```diff
@@ -886,6 +887,60 @@ sub add_properties_from_specific_ingredients ($product_ref) {
 	return;
 }
 
+=head2 add_percent_max_for_salt_and_sugar_ingredients_from_nutrition_facts ( $product_ref )
+
+Add a percent_max value for salt and sugar ingredients, based on the nutrition facts.
+
+=cut
+
+sub add_percent_max_for_salt_and_sugar_ingredients_from_nutrition_facts ($product_ref) {
```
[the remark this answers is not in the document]
stephanegigandet: sure

lib/ProductOpener/Ingredients.pm
stephanegigandet: We could, it was just to be safe if the sugar or salt is not 100% sugar or salt.

lib/ProductOpener/Ingredients.pm
stephanegigandet: The ingredients are listed by quantity order, we keep them in this order.

lib/ProductOpener/Ingredients.pm
stephanegigandet: ah good point, initially I had it in the inner loop but I wanted to move it to the outer loop, and I forgot to remove the inner loop.. :(

we can keep the inner loop only

teolemon:
@stephanegigandet applied changes, @alexgarel can we merge ?

PR #9304, opened 2023-11-12
feat: adding German plurals to the oils in Ingredients.pm

moon-rabbitOFF:
### What

feat: adding German plurals to the oils in Ingredients.pm


lib/ProductOpener/Ingredients.pm
```diff
@@ -3877,36 +3885,66 @@ string, comma-joined category and type, example: 'palm vegetal oil' or 'sunflowe
...
+
+	if (defined $alternate_names_ref) {
+
+		my $name_exists;
+		canonicalize_taxonomy_tag($lc, "ingredients", $a_of_b, \$name_exists);
+
+		if (not $name_exists) {
+			foreach my $alternate_name (@{$alternate_names_ref}) {
+				my $alternate_name_copy
+					= $alternate_name;    # make a copy so that we can modifyt it without changing the array entry
```
benbenben2, line 3932:
```suggestion
					= $alternate_name;    # make a copy so that we can modify it without changing the array entry
```

lib/ProductOpener/Ingredients.pm
```diff
@@ -4944,23 +4983,60 @@ This function lists each individual ingredient:
...
+			types => ["Kokosnuss", "Palm", "Palmkern", "Raps", "Shea", "Sonnenblumen",],
+			# Kokosnussöl, Sonnenblumenfett
+			alternate_names => ["<type>fett", "<type>öl"],
+		},
+	],
+
+Simple plural (just an additional "s" at the end) will be added in the regexp.
+
+Note that a "<categories> ([list of types])" enumeration will be developed only if all the types can be matched
+to the specified types in ingredients_categories_and_types.
```
benbenben2, line 5011: It seems that even if all the types are not matched, it works:
https://fr.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=piment+%28vert%2C+rouge%2C+jaune%29&type=add&action=process&submit=Envoyer

Due to this:
`join(", ", map {normalize_a_of_b($lc, $category, $_, $of_bool, $alternate_names_ref)} @list) . $trailing_space;`

lib/ProductOpener/Ingredients.pm
```diff
@@ -5134,7 +5207,7 @@ sub develop_ingredients_categories_and_types ($ingredients_lc, $text) {
 
 		foreach my $categories_and_types_ref (@{$ingredients_categories_and_types{$ingredients_lc}}) {
```
benbenben2, line 5208: Do we need to define an **$alternate_names_regexp** there as well?

There is only German for now, maybe German accents (for oil, for example) should be handled there?

lib/ProductOpener/Ingredients.pm
stephanegigandet: It's a bug in fact, we don't want to create ingredients that do not make sense. I'm fixing it and adding tests for it.

lib/ProductOpener/Ingredients.pm
stephanegigandet: The alternate names are only for finding the right name associated with a type, for matching the categories we still use the categories field. Maybe I'm misunderstanding what you want to do, could you give me an example of something that you want to add parsing support for, and what it should be parsed to?

benbenben2: Ah, I understand now. If there is no need of regex after, then we don't need that. Okay okay.


But what if we have ingredient:
Pflanzliche ole (Kurbiskern)

(Instead of "Pflanzliche öle (Kürbiskern)" with accents)

Category: Pflanzliche ole (ok)
Type: Kurbiskern (ok)
alternate_names: Kurbiskernöl (not ok but not related to alternate_names)

So yes, you are totally right, we don't need $alternate_names_regexp

Kurbiskern vs kürbiskern would be handled in canonicalize_taxonomy_tag, is that correct? 
If yes, then, all good you can close that comment.

benbenben2 (approved):
Looks good to me
Great work 👍

PR #9307, opened 2023-11-13
fix: parsing of beef meat from animals fed without GMO

stephanegigandet:
Enables to parse strings like "beef meat from animals fed without GMO" / ""Viande de boeuf issue d'animaux nourris sans OGM""


alexgarel (approved):
LGTM

PR #9317, opened 2023-11-14
fix: fix_avoid_eiweiss_false_positive_for_allergens

### LINKED ISSUE #8386: Separate 'Eiweiß' from 'Ei', because 'Eiweiß' does not contain any 'Ei'
### Describe the bug

The indication of allergens and vegan ingredients is regularly incorrect in German ingredient lists.

The problem: The German word for _egg_ is _Ei_, the German word for _protein_ is _Eiweiß_. While the _word_ Eiweiß contains _Ei_, there's no egg in it. If the ingredient list says 'Weizen**ei**weiß' (wheat protein), OFDB will give this as an allergen (egg=Ei).

Example: https://de.openfoodfacts.org/produkt/4260389050812

This would have to be separated if possible.

### To Reproduce

Open https://de.openfoodfacts.org/produkt/4260389050812, see allergens.

### Expected behavior

No more 'Egg' in the list of allergens if it contains protein.

### Screenshots

grafik

### Additional context

_No response_

### Type of device

Browser

### Browser version

LibreWolf 112

Thousands.

### Time per product

0

benbenben2:
### What
This is due to underscores. Replacing: "_Weizen_eiweiß" by: "Weizen eiweiß" or "Weizeneiweiß" avoids having the eggs allergen.

For other allergens, the following regex seems to work:
```
			$text
				=~ s/(^| - |_|\(|\[|\)|\]|,|$the|$and|$of|;|\.|$)((\s*)\w.+?)(?=(\s*)(^| - |_|\(|\[|\)|\]|,|$and|;|\.|\b($traces_regexp)\b|$))/replace_allergen_between_separators($language,$product_ref,$1, $2, "",$`)/iesg;
```
However, in the case of _Weizen_eiweiß it leads to 2 allergens. Hence, suggested solution is to add (for German only to minimize probability of false positive) the case when underscore after the allergen is not at the end of the word (remove \b of the previous line. This is done only for \b after the allergen, not before, again to minimize probability of false positive).

### Screenshot
Screenshot_20231114_170812

- Fixes #8386

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -6689,6 +6689,10 @@ sub detect_allergens_from_text ($product_ref) {
 			$text =~ s/\b___([^,;_\(\)\[\]]+?)___\b/replace_allergen($language,$product_ref,$1,$`)/iesg;
 			$text =~ s/\b__([^,;_\(\)\[\]]+?)__\b/replace_allergen($language,$product_ref,$1,$`)/iesg;
 			$text =~ s/\b_([^,;_\(\)\[\]]+?)_\b/replace_allergen($language,$product_ref,$1,$`)/iesg;
+			# _Weizen_eiweiß is not caught in last regex because of \b (word boundary).
+			if ($language eq 'de') {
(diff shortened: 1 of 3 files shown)


stephanegigandet (approved):
Thank you!

PR #9320, opened 2023-11-14
feat: dq_all_val_in_nutrition_are_identical

### LINKED ISSUE #1564: Reject edits where all entered nutrition values are 0
Some users from 3rd party apps just enter 0s everywhere.

benbenben2:
### What
New error, when all values in the nutrition table are equal.
Note that it can happen only when they are equal to 0 because otherwise salt and sodium values are automatically calculated from each other's value.

Additionally, not coherent naming in the data quality taxonomy (Error vs errrors) + noticed that the first part of the file lists errors and second part lists warnings, but some errors were listed in the second part, so I moved them up to the first part.

### Screenshot
Screenshot_20231114_234657

- Fixes #1564

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -787,19 +787,19 @@ sub check_nutrition_data ($product_ref) {
 				$total += $product_ref->{nutriments}{$nid . "_100g"};
 			}
 
-			# variables to check if there are 3 or more duplicates in nutriments
+			# variables to check if there are many duplicates in nutriments
 			if (
-				(
-					   ($nid eq 'fat_100g')
-					or ($nid eq 'saturated-fat_100g')
-					or ($nid eq 'carbohydrates_100g')
-					or ($nid eq 'sugars_100g')
-					or ($nid eq 'fiber_100g')
-					or ($nid eq 'proteins_100g')
-					or ($nid eq 'salt_100g')
-					or ($nid eq 'sodium_100g')
-				)
-				and ($product_ref->{nutriments}{$nid} > 1)
+				# (
+				   ($nid eq 'energy-kj_100g')
+				or ($nid eq 'energy-kcal_100g')
+				or ($nid eq 'fat_100g')
+				or ($nid eq 'saturated-fat_100g')
+				or ($nid eq 'carbohydrates_100g')
+				or ($nid eq 'sugars_100g')
+				or ($nid eq 'fiber_100g')
+				or ($nid eq 'proteins_100g')
+				or ($nid eq 'salt_100g')
+				or ($nid eq 'sodium_100g')
 				)
 			{
 				push(@major_nutriments_values, $product_ref->{nutriments}{$nid});
@@ -807,7 +807,7 @@ sub check_nutrition_data ($product_ref) {
 
 		}
 
-		# create a hash key: nutriment value, value: number of occurence
+		# create a hash key: nutriment value, value: number of occurences
 		foreach my $nutriment_value (@major_nutriments_values) {
 			if (exists($nutriments_values_occurences{$nutriment_value})) {
 				$nutriments_values_occurences{$nutriment_value}++;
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -816,14 +816,27 @@ sub check_nutrition_data ($product_ref) {
...
 
+		# retrieve max number of occurences
+		my $nutriments_values_occurences_max_value = -1;
+		foreach my $key (keys %nutriments_values_occurences) {
+			if ($nutriments_values_occurences{$key} > $nutriments_values_occurences_max_value) {
+				$nutriments_values_occurences_max_value = $nutriments_values_occurences{$key};
+				# $nutriments_values_occurences_max_key = $key;
+			}
+		}
+		# raise error if all values are identical (this can only apply when they are 0 (because salt or sodium are automatically generated from each others))
```
stephanegigandet, line 835: Then what would be the difference with the already existing en:Nutrition all values zero ?

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -816,14 +816,27 @@ sub check_nutrition_data ($product_ref) {
...
+		# raise warning if there are 3 or more duplicates in nutriments and nutriment is above 1
 		foreach my $keys (keys %nutriments_values_occurences) {
-			if ($nutriments_values_occurences{$keys} > 2) {
+			if (($nutriments_values_occurences{$keys} > 2) and ($keys > 1)) {
 				push @{$product_ref->{data_quality_warnings_tags}}, "en:nutrition-3-or-more-values-are-identical";
 				last;
 			}
 		}
 
+		# retrieve max number of occurences
```
stephanegigandet, line 827: I think that you can do something simpler that would also work when people put a non zero value (e.g. 1) in all fields:

Just check if you have only 1 key, or you have 2 keys and 1 of them is "salt_100g" or "sodium_100g". (and also that you have at least 3 or 4 entries in @major_nutriments_values)

lib/ProductOpener/DataQualityFood.pm
benbenben2: I did not see that before, actually. That already existing dq error facet has nothing: https://world.openfoodfacts.org/data-quality-error/en:all-nutrition-values-are-set-to-0
Because:
1) 
```
if ($nid !~ /fruits-vegetables-nuts-estimate-from-ingredients/) {
	if ($product_ref->{nutriments}{$nid} == 0) {
		$nid_zero++;
	}
	else {
		$nid_non_zero++;
	}
}
```
does not exclude "fruits-vegetables-legumes-estimate-from-ingredients_100g"

**comment:** instead of adding this exclusion, that would be better to list which main nutrients we want to select. That way if tomorrow there is a new fruits-something-blabla_100g we will not have problem if we forget to update that condition.

2)
there is a typo: 
```($nid_zero == $nid_n)```
where
**$nid_zero**: number of $nid (nutrient) that contains "_100g" and equal to 0. That number is at most all "_100g" minus fruits-vegetables-nuts-estimate-from-ingredients_100g.
**$nid_n**: all nutrient including "_100g" as well as without "_100g". Example for proteins: "proteins", "proteins_100g", "proteins_unit", "proteins_value".

According to my comment, I am removing this dq facet to replace it by the one implemented in this PR.

benbenben2:
I do not understand why test needed to be updated for that part:
`	"compared_to_category" : "en:cereals-and-potatoes",`

lib/ProductOpener/DataQualityFood.pm
benbenben2: I did something like that. But, because it is an error, I would not add that part "and also that you have at least 3 or 4 entries" for now.

Note that there are not much more rows if we count salt or sodium instead of both (about 10 more): [mirabelle](https://mirabelle.openfoodfacts.org/products?sql=select+*+from+%5Ball%5D+where+%22energy-kj_100g%22+%3D+%22energy-kcal_100g%22+and+%22energy-kcal_100g%22+%3D+%22fat_100g%22+and+%22fat_100g%22+%3D+%22saturated-fat_100g%22+and+%22saturated-fat_100g%22+%3D+%22carbohydrates_100g%22+and+%22carbohydrates_100g%22+%3D+%22sugars_100g%22+and+%22sugars_100g%22+%3D+%22fiber_100g%22+and+%22fiber_100g%22+%3D+%22proteins_100g%22+and+%28%22proteins_100g%22+%3D+%22salt_100g%22+or+%22proteins_100g%22+%3D+%22sodium_100g%22%29%0D%0Aorder+by+random%28%29+limit+10)

benbenben2:
Note for later (maybe @CharlesNepote could be interested for Data Quality).

If we count energy kcal or kj instead of both that leads to 10000 rows: [mirabelle](https://mirabelle.openfoodfacts.org/products?sql=select+DISTINCT+code%2C+url+from+%5Ball%5D+where+%28%22energy-kj_100g%22+%3D+%22fat_100g%22+or+%22energy-kcal_100g%22+%3D+%22fat_100g%22%29+and+%22fat_100g%22+%3D+%22saturated-fat_100g%22+and+%22saturated-fat_100g%22+%3D+%22carbohydrates_100g%22+and+%22carbohydrates_100g%22+%3D+%22sugars_100g%22+and+%22sugars_100g%22+%3D+%22fiber_100g%22+and+%22fiber_100g%22+%3D+%22proteins_100g%22+and+%28%22proteins_100g%22+%3D+%22salt_100g%22+or+%22proteins_100g%22+%3D+%22sodium_100g%22%29+%0D%0A--+and+%22energy-kj_100g%22+%21%3D+%22energy-kcal_100g%22+%0D%0Aand+%22energy-kcal_100g%22+%21%3D+0%0D%0AGROUP+BY+code%2C+url%0D%0Aorder+by+random%28%29+)

With many false positives: 
https://world.openfoodfacts.org/product/0896161001235/alkaline-water
https://world.openfoodfacts.org/product/8001120715456/te-nero-zenzero-e-scorze-di-limone-biologico-coop
https://world.openfoodfacts.org/product/0086854052747/canola-oil-cooking-spray-laura-lynn
https://world.openfoodfacts.org/product/0011141107442/coffee-beans
https://world.openfoodfacts.org/product/8720354187838/eau-mineral-petillante-hema

 - Some of them are kj=1 kcal 0. ... In that case we could count only if kj is higher than 4 (because 4.2 kj = 1 kcal) which would imply kcal = 1 at least
 - We could add category whitelist... (waters, coffees, ).. although most of them do not even have category.
 - Most of these false positives are in the u.s. where only cal (kcal) is provided (because it is per portion). So we are good if we keep counting both kj and kcal because kj is missing for these products (field should be empty)

stephanegigandet (approved):

PR #9323, opened 2023-11-15
taxonomy: add Cachaça (the spirit in Caipirinha)

OpenChris:
add Cachaça (the spirit in Caipirinha)
Cachaça is similar to Rum but a distinct spirit --> parent would be `fr:Eaux-de-vie` like for Rum


stephanegigandet (approved):
Thank you!
Could you add xx:Cachaça at the bottom so that it matches in all languages?

taxonomies/categories.txt
```diff
@@ -4394,6 +4394,13 @@ fr:Rhums vieux, rhum vieux
 nl:Oude rums, Oude rum
 #wikidata:en:
 
+<fr:Eaux-de-vie
+en:Cachaça, pinga, caninha
+fr:Cachaça, pinga, cachasa, cachase
```
teolemon, line 4399:
```suggestion
fr:Cachaça, pinga, cachasa, cachase
xx:Cachaça
```

PR #9326, opened 2023-11-15
taxonomy: German and Finnish ingredients

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]


taxonomies/ingredients.txt
```diff
@@ -4479,6 +4479,16 @@ ciqual_food_name:fr:Yaourt à la grecque, nature
 #ciqual_food_name:en:Yogurt, Greek-style, ewe's milk
 #ciqual_food_name:fr:Yaourt à la grecque, au lait de brebis
 
+<en:dairy
+en:skyr
```
stephanegigandet, line 4483: we can just have this:

en:skyr (we want a en: entry as the main name of all ingredients)
xx:skyr (will match skyr in all languages, and will be the default translation for all languages)

moon-rabbitOFF: Ah true I forgot, I modified it

stephanegigandet (approved):
Thank you!

PR #9330, opened 2023-11-16
feat: improvements to parsing of ingredient with labels

stephanegigandet:
This PR:
- improves the way we parse labels inside ingredients lists
- add some entries to ingredients, ingredients_processing and labels taxonomies


aleene:
I find it a pity that we lose the link between ingredient and label. The label becomes often to generic. But I am in favour of any improvement here.

stephanegigandet:
> I find it a pity that we lose the link between ingredient and label. The label becomes often to generic. But I am in favour of any improvement here.

@aleene we do not lose the link between ingredient and label, when there is something like "farine de blé CRC" in the list of ingredients, it is parsed and attached to the ingredient:

```
   "ingredients" : [
      {
         "id" : "en:wheat-flour",
         "labels" : "fr:crc",
         "percent_estimate" : 62.5,
         "percent_max" : 100,
         "percent_min" : 25,
         "text" : "Farine de blé",
         "vegan" : "yes",
         "vegetarian" : "yes"
      },
```

benbenben2 (approved):
Looks good to me

PR #9336, opened 2023-11-17
fix: dq prevent split and when percent

### LINKED ISSUE #7816: en:nutrition-value-over-105-fruits-vegetables-nuts-estimate-from-ingredients: false positive
### Describe the bug

Sometimes fruit estimation is higher than 105 while it shouldn't be.
Eg. https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=3038354191904#ingredients
image

The json file displays 174.5 in this example:
image

### To Reproduce

...

### Expected behavior

...

### Screenshots

_No response_

### Additional context

_No response_

### Type of device

Browser

### Browser version

_No response_

_No response_

### Time per product

_No response_

benbenben2:
### What
do no split "Ingredient A and Ingredient B (81%)" to "Ingredient A (81%), Ingredient B (81%)"

### Screenshot
Screenshot_20231117_175722

- Fixes #7816

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -2101,7 +2101,7 @@ sub parse_ingredients_text_service ($product_ref, $updated_product_fields_ref) {
 		my @ingredients = ();
 
 		# 2 known ingredients separated by "and" ?
-		if ($before =~ /$and/i) {
+		if (($before =~ /$and/i) and (not defined $percent_or_quantity_value)) {
 
 			my $ingredient = $before;
 			my $ingredient1 = $`;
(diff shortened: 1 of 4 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -2101,7 +2101,7 @@ sub parse_ingredients_text_service ($product_ref, $updated_product_fields_ref) {
 		my @ingredients = ();
 
 		# 2 known ingredients separated by "and" ?
-		if ($before =~ /$and/i) {
+		if (($before =~ /$and/i) and (not defined $percent_or_quantity_value)) {
```
stephanegigandet, line 2104: If we do that, we fix the error with the percent being applied twice, but we introduce an issue: the ingredients won't be recognized.

An option could be to add a "compound" ingredient, as discussed in the comments of https://github.com/openfoodfacts/openfoodfacts-server/pull/9236

in practice it would then result in "[compound ingredients] 80% (pulpe de tomates, puree de tomates)"

benbenben2:
Alright, I changed the PR as draft.

Related to #9345

PR #9339, opened 2023-11-19
feat: dq_kcal_does_not_match_exclude_more

### LINKED ISSUE #7789: Energy-value-in-kcal-does-not-match-value-computed-from-other-nutrients: exclude more products
### Describe the bug

In the Wikipedia article about [Food Energy](https://en.wikipedia.org/wiki/Food_energy), we can read: "_Some polyols, like [erythritol](https://en.wikipedia.org/wiki/Erythritol), are not digested and should be excluded from the count_".

For example, stevia sugars generally contain 90+g/100g of polyols and those are erythritol.
1000+ sweeteners are concerned, mainly stevia or 0 calories sugars: https://world.openfoodfacts.org/data-quality-error/energy-value-in-kcal-does-not-match-value-computed-from-other-nutrients/category/sweeteners

Calorie-free ingredients:
* erythritol (E968) (0.2 Kcal/g): https://en.wikipedia.org/wiki/Erythritol
* stevia (natural): https://en.wikipedia.org/wiki/Stevia
* acesulfame K (E950): https://en.wikipedia.org/wiki/Acesulfame_potassium
* saccharin (E954): https://en.wikipedia.org/wiki/Saccharin
* psicose, aka allulose (0.2 Kcal/g): https://en.wikipedia.org/wiki/Psicose

benbenben2:
### What
Need to be reviewed carefully.
- See comments in the issue
- see how tests had to be updated (for polyols and erythritol)
- note that honeys category is under sweeteners and en:energy-value-in-kj-does-not-match-value-computed-from-other-nutrients will be shutdown for those as well

- Fixes #7789

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -597,7 +597,14 @@ sub check_nutrition_data_energy_computation ($product_ref) {
 			my ($ignore_energy_calculated_error, $category_id)
 				= get_inherited_property_from_categories_tags($product_ref, "ignore_energy_calculated_error:en");
 
-			if (not((defined $ignore_energy_calculated_error) and ($ignore_energy_calculated_error eq 'yes'))) {
+			if (
+				(
+					not((defined $ignore_energy_calculated_error) and ($ignore_energy_calculated_error eq 'yes'))
+					and (  ($product_ref->{nutriments}{"energy-kj_value"} > 55)
+						or ($product_ref->{nutriments}{"energy-kcal_value"} > 13))
+				)
+				)
+			{
 				# Compare to specified energy value with a tolerance of 30% + an additiontal tolerance of 5
 				if (   ($computed_energy < ($specified_energy * 0.7 - 5))
 					or ($computed_energy > ($specified_energy * 1.3 + 5)))
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -601,8 +601,12 @@ sub check_nutrition_data_energy_computation ($product_ref) {
 				(
 					not((defined $ignore_energy_calculated_error) and ($ignore_energy_calculated_error eq 'yes'))
 					# consider only when energy is high enough to minimize false positives (issue #7789)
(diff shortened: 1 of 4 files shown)


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -597,7 +597,14 @@ sub check_nutrition_data_energy_computation ($product_ref) {
 			my ($ignore_energy_calculated_error, $category_id)
 				= get_inherited_property_from_categories_tags($product_ref, "ignore_energy_calculated_error:en");
 
-			if (not((defined $ignore_energy_calculated_error) and ($ignore_energy_calculated_error eq 'yes'))) {
+			if (
+				(
+					not((defined $ignore_energy_calculated_error) and ($ignore_energy_calculated_error eq 'yes'))
+					and (  ($product_ref->{nutriments}{"energy-kj_value"} > 55)
```
stephanegigandet, line 603: Can you add a comment to explain why we have this condition on energy? (so that we know if we revisit it later...)

benbenben2:
> Instead of disabling the error, I think it would be better to change a bit the condition for the error, in case the energy is low.
> 
> (in the part # Compare to specified energy value with a tolerance of 30% + an additiontal tolerance of 5)
> 
> That way if the energy is 5, but it should be 1500, we want to have the error raised. We should add a test for that (low energy but lots of fat and sugar for instance)

Very good suggestion.

Now, we consider either $specified_energy > 55 or $computed_energy > 55

Now, we can also ignore my previous comment in first message:
`see how tests had to be updated (for polyols and erythritol)`

Existing test, already covers that case:
```
# energy does not match nutrients
$product_ref = {
	nutriments => {
		"energy-kj_value" => 5,
		"carbohydrates_value" => 10,
		"fat_value" => 20,
		"proteins_value" => 30,
		"fiber_value" => 2,
	}
};
ProductOpener::DataQuality::check_quality($product_ref);
is($product_ref->{nutriments}{"energy-kj_value_computed"}, 1436);
ok(has_tag($product_ref, 'data_quality', 'en:energy-value-in-kj-does-not-match-value-computed-from-other-nutrients'),
	'energy not matching nutrients')
	or diag explain $product_ref;
```

stephanegigandet (approved):
Looks good, thank you!

PR #9356, opened 2023-11-21
fix: og-url for donate page to get correct Facebook share preview, change meta title

stephanegigandet:
The broken og:url results in a big "Error" when sharing the donation page on Facebook.

### DIFF
--- a/html/donate/en.html
+++ b/html/donate/en.html
@@ -9,10 +9,10 @@
 
     <meta property="fb:app_id" content="219331381518041" />
     <meta property="og:type" content="article" />
-    <meta property="og:title" content="Help us fund the Open Food Facts 2024 budget!" />
+    <meta property="og:title" content="Help us lead the change in the food sector!" />
     <meta
       property="og:url"
-      content="https://world.openfoodfacts.org/donate-to-openfoodfacts"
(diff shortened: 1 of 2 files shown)


teolemon (approved):

html/donate/en.html
```diff
@@ -9,10 +9,10 @@
 
     <meta property="fb:app_id" content="219331381518041" />
     <meta property="og:type" content="article" />
-    <meta property="og:title" content="Help us fund the Open Food Facts 2024 budget!" />
+    <meta property="og:title" content="Help us lead the change in the food sector!" />
     <meta
       property="og:url"
-      content="https://world.openfoodfacts.org/donate-to-openfoodfacts"
+      content="https://world.openfoodfacts.org/donate-to-open-food-facts"
```
teolemon, line 15: @stephanegigandet we should put an UTM in there 
```suggestion
      content="https://world.openfoodfacts.org/donate-to-open-food-facts?utm_source=donate-page&utm_campaign=donation-2024&utm_content=donate-page-en"
```

stephanegigandet: OK but adding utm tags to an open graph urls is the kind of things that is likely to have unforeseen side effects.

PR #9358, opened 2023-11-21
feat: dq_extract_polyols_starch_from_sugar_in_edit_page

### LINKED ISSUE #8124: Starch and polyols seem to be included in sugar
### Describe the bug

When looking at how we display starch and polyols in the product form, they seem included in the sugar:
image

Sugar, starch and polyols should be at the same level, under carbohydrates, isn't it?

We already have a data quality facet: [Nutrition - Sugars plus starch greater than carbohydrates](https://world.openfoodfacts.org/data-quality-error/nutrition-sugars-plus-starch-greater-than-carbohydrates).

Shouldn't we have one: `Nutrition - Sugars plus starch plus polyols greater than carbohydrates`?

### To Reproduce

eg. https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=8410636000077

benbenben2:
### What
Set sugar, starch and polyols at the same level, under carbohydrates.

### Screenshot
Screenshot_20231121_174534

- Fixes #8124


lib/ProductOpener/Food.pm
```diff
@@ -425,7 +425,7 @@ It is a list of nutrients names with eventual prefixes and suffixes:
 		(
 			'!energy-kj', '!energy-kcal',
 			'!energy-', '-energy-from-fat-',
-			'!fat', '!-saturated-fat',
```
stephanegigandet, line 428: We want to keep the ! in !-saturated-fat, it indicates that when we display the nutrition facts table in the product page (not the product edit form), we want to display Saturated fat even if we don't have a value for the product.

stephanegigandet:
@benbenben2 Thanks for looking into this. I think the issue is only in the code for the product edit form in fact. We don't want to change how the nutrition facts table on the product page is displayed. I think the right fix would be something like this: https://github.com/openfoodfacts/openfoodfacts-server/pull/9420

benbenben2:
Closed by #9420

PR #9380, opened 2023-11-23
fix: facets for EU packager codes (EC) and for users

### LINKED ISSUE #9372: Contribution count over-reported after server migration
### What
- Contribution count over-reported after server migration
- I get 2,9M contributions :-) 
- Stephane seems to say it's not a classic facet, but he will comment probably tomorrow with more details.
- cc @stephanegigandet @john-gom 
image

stephanegigandet:
- Fixes the normalization of the EC packager codes
- Also fixes #9372

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -4511,7 +4511,7 @@ my %ignore_params = (
 # Parameters that can be query filters
 # It is safer to use a positive list, instead of just the %ignore_params list
 
-my %valid_params = (code => 1,);
+my %valid_params = (code => 1, creator => 1);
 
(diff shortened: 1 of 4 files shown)


teolemon (approved):

stephanegigandet:
> LGTM although I don't understand how it solves the attribution bug (and don't we have any test on this ?).

The attribution bug is fixed by 
my %valid_params = (code => 1, creator => 1);

This is the list of parameters that can be used as filters even though they are not tag fields.

I'll add a comment and a test.

stephanegigandet:
@alexgarel I added a test and a comment to show that it fixes the contributors facet

PR #9395, opened 2023-11-24
feat: Synonyms in taxonomized suggestions

### LINKED ISSUE #8578: Allergens autocompletion should use synonyms
### Problem
When typing "OE" in the allergen text box in french, you don't get "Œuf" entry although the ["oeuf" variant](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/taxonomies/allergens.txt#LL102C21-L102C21) is present in taxonomy

This is quite cumbersome for on windows œ is not really accessible.

### Proposed solution

All synonyms should be taken into account when you type in allergens.

### Additional context

This field does not use a suggest.pl request. If it was, [it would work](https://world.openfoodfacts.org/cgi/suggest.pl?tagtype=allergens&term=oe&lc=fr&limit=300).

Also this might affect other fields, which would benefit from synonyms.

Naruyoko:
- [X] PR title is prefixed by one of the following: feat, fix, docs, style, refactor, test, build, ci, chore, revert, l10n, taxonomy
- [ ] Code is well documented
- [ ] If you have multiple commits please combine them into one commit by squashing them.
- [ ] Read and understood the [contribution guidelines](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/CONTRIBUTING.md)

### What
When creating suggestions for taxonomized fields like categories, the API gives back results where one of the synonyms match. However, they are not displayed when they don't include the typed string. This tries to fix that by returning which synonym the input matched to build the suggestions list.

- Fixes #8578


Naruyoko:
This is still a prototype, and the values shown are still broken. The synonyms should be sent by the display format rather than the ID.

Naruyoko:
Here is a demo showing the current progress. I haven't yet figured out how to return the synonyms' names properly. That should be done on the server side which has the full taxonomy information.

https://github.com/openfoodfacts/openfoodfacts-server/assets/32781397/2c4ef0ed-b1d5-4f97-b0c8-db460395bb7f

Here is the current result of `http://fr.openfoodfacts.localhost/api/v3/taxonomy_suggestions?tagtype=allergens&string=o&get_synonyms=1`:

```json
{"errors":[],"matched_synonyms":["orge","oeufs","arachis-hypogaea","langoustine","fruits-a-coque-dure","beurre-concentre","petoncles","graines-de-moutarde","rouget","tofu"],"status":"success","suggestions":["Gluten","Œufs","Arachides","Crustacés","Fruits à coque","Lait","Mollusques","Moutarde","Poisson","Soja"],"warnings":[]}
```

Naruyoko:
Suggestions are now shown in proper names.
Proper names of synonyms

Naruyoko:
The response from `http://fr.openfoodfacts.localhost/api/v3/taxonomy_suggestions?tagtype=allergens&string=o&get_synonyms=1` now looks like this:
```json
{"errors":[],"matched_synonyms":["Orge","Œufs","Arachis hypogaea","Langoustine","Fruits à coque dure","Beurre concentré","Pétoncles","Graines de moutarde","Rouget","Tofu"],"status":"success","suggestions":["Gluten","Œufs","Arachides","Crustacés","Fruits à coque","Lait","Mollusques","Moutarde","Poisson","Soja"],"warnings":[]}
```

docs/api/ref/api-v3.yml
```diff
@@ -247,6 +247,11 @@ paths:
                         description: Array of sorted strings suggestions in the language requested in the "lc" field.
                         items:
                           type: string
+                      matched_synonyms:
```
alexgarel, line 250: from what I undertand seeing tests, you did not only change api-v3 but also api-v2 (in api.yml). shan't you change it there also ?

alexgarel (changes requested):
@Naruyoko: thanks a lot for this PR.

I have a blocker though. There are other people relying on this suggestion API. So we must not change the response format.

To better play with OpenAPI, I propose that we do a suggestion2.cgi API with your new format (@stephanegigandet do you have a better idea ?)

docs/api/ref/api-v3.yml
alexgarel: This comment might not be relevant if we go for a suggestion api v2.

Naruyoko: I am not sure where I should separate my changes. Do I copy [this part](https://github.com/openfoodfacts/openfoodfacts-server/blob/417a4c0de43b234b07f608400fdbf12aaf38c25d/lib/ProductOpener/API.pm#L405) with a new name with the new function exposed only there? If so, do I need to copy [this function](https://github.com/openfoodfacts/openfoodfacts-server/blob/086a0a0a926589aecf14a1f33e12785a1c1270c5/lib/ProductOpener/APITaxonomySuggestions.pm#L67) as well? Or do the changes need to be separated at the level of inner functions, like from [here](https://github.com/openfoodfacts/openfoodfacts-server/blob/086a0a0a926589aecf14a1f33e12785a1c1270c5/lib/ProductOpener/TaxonomySuggestions.pm#L317)?

I am also not sure how I should change `api.yml`, since [it appears mostly empty](https://github.com/openfoodfacts/openfoodfacts-server/blob/086a0a0a926589aecf14a1f33e12785a1c1270c5/docs/api/ref/api.yml#L277).

alexgarel: @Naruyoko in the code you show, condition should be enough to sort out the cases.

But we could add a parameter to `get_taxonomy_suggestions` and have a `cgi/suggest_v2.pl` alike [suggest.pl](https://github.com/openfoodfacts/openfoodfacts-server/blob/417a4c0de43b234b07f608400fdbf12aaf38c25d/cgi/suggest.pl)

alexgarel: @Naruyoko, do not hesitate to ping me on slack if it's not clear enough.

alexgarel: After discussion:

* do not change suggest.pl output
* we can change api/v3/suggest output conditioned by a specific parameter (with_synonyms?)
* we will ask the mobile app to use this new parameter so that we can remove the old output for api/v3/suggest (easier to match to OpenAPI)

Naruyoko: The output of suggest.pl should be reverted now. It wasn't clear to me that I should expose the feature through a new CGI.

I think we should discuss the usage and the returned structure for the new v3. Currently,

- The synonym information is controlled by the new `get_synonyms` flag being set to `1`.
- The flag adds the list of synonyms as `matched_synonyms` to the data (See [here](https://github.com/openfoodfacts/openfoodfacts-server/pull/9395#issuecomment-1839231773)). They are in the same order as the list of tags.

I think of two considerations:
- What would be an appropriate name for this flag (you put `with_synonyms`)? Would it even make sense to switch the behavior with a such flag?
- Is there a better format? Maybe they should be transposed so it returns an array of tag-synonym pairs (as it is internally)?

alexgarel: > This just occurred to me: is the "v3" API the in-dev/evolving version?

Exactly :-)

alexgarel: > Is there a better format? Maybe they should be transposed so it returns an array of tag-synonym pairs (as it is internally)?

I thinks it's better to use dict in an API, as you dead, it's cleaner and more "self documented".

Naruyoko: By using dict, do you mean like this (which was what I intended by "traspose"):

```json
{"errors":[],"suggestions":[{"matched_synonym":"Orge","tag":"Gluten"},{"matched_synonym":"Œufs","tag":"Œufs"},{"matched_synonym":"Arachis hypogaea","tag":"Arachides"},{"matched_synonym":"Langoustine","tag":"Crustacés"},{"matched_synonym":"Fruits à coque dure","tag":"Fruits à coque"},{"matched_synonym":"Beurre concentré","tag":"Lait"},{"matched_synonym":"Pétoncles","tag":"Mollusques"},{"matched_synonym":"Graines de moutarde","tag":"Moutarde"},{"matched_synonym":"Rouget","tag":"Poisson"},{"matched_synonym":"Tofu","tag":"Soja"}],"warnings":[]}
```

Or like this:

```json
{"errors":[],"suggestions":{"Gluten":"Orge","Œufs":"Œufs","Arachides":"Arachis hypogaea","Crustacés":"Langoustine","Fruits à coque":"Fruits à coque dure","Lait":"Beurre concentré","Mollusques":"Pétoncles","Moutarde":"Graines de moutarde","Poisson":"Rouget","Soja":"Tofu"},"warnings":[]}
```

Or like this:

```json
{"errors":[],"matched_synonyms":{"Gluten":"Orge","Œufs":"Œufs","Arachides":"Arachis hypogaea","Crustacés":"Langoustine","Fruits à coque":"Fruits à coque dure","Lait":"Beurre concentré","Mollusques":"Pétoncles","Moutarde":"Graines de moutarde","Poisson":"Rouget","Soja":"Tofu"},"status":"success","suggestions":["Gluten","Œufs","Arachides","Crustacés","Fruits à coque","Lait","Mollusques","Moutarde","Poisson","Soja"],"warnings":[]}
```

alexgarel: I like your third version, because it keeps things simple as you can just retrieve the simple suggestion list, and it's easy to find back matched synonym.

Just one point: shan't it be `"matched_synonyms":{"Gluten":["Orge"],…}` as there might be more than one matched synonym (or don't we care because we just return one ? Which is still really ok for me, as it covers the needs). 

Also is it useful to put entries like "Œufs": "Œufs" inside matched_synonyms or can the API user deduce from the fact that Œufs is not in the synonyms dict that it means it was directly matched ?

Naruyoko: I agree with using the third version.

The matched synonym is currently the first matched one of the best match (with "start">"inside">"fuzzy"). I don't see how much useful it is to return all possible matches.

I think it is better to keep `"a":"a"` entries for consistency.

Naruyoko: I pushed a change with the new format. I will have to update the API docs and create new tests.

alexgarel: Yes, I read your changes, seems perfect :muscle: 

So I wait for the other changes. Don't hesitate to ping me on slack when you are ready @Naruyoko :-)

docs/api/ref/api-v3.yml
```diff
@@ -239,6 +239,11 @@ paths:
...
+                        type: object
+                        description: Dictionary of strings associating canonical names (as seen in suggestions field) with the synonym that best matches the query. An entry is present for all suggestions, even when the synonym is the same with the canonical name.
```
alexgarel, line 244:
```suggestion
                        description: |                      
                            Dictionary of strings associating canonical names (as seen in suggestions field) with the synonym that best matches the query. An entry is present for all suggestions, even when the synonym is the same with the canonical name.
                            
                            This value is present only if get_synonyms parameter is present.
```

alexgarel (approved):
Very good @Naruyoko, thanks for this great contribution !

I add a very small change to documentation.

PR #9425, opened 2023-11-28
fix: new local Fairtrade International members

stephanegigandet:
Added all national members of Fairtrade International https://en.wikipedia.org/wiki/Fairtrade_International


odtvince (approved):
LGTM

teolemon (approved):

PR #9428, opened 2023-11-28
fix: use Redis streams instead of redis queue

raphael0202:
This PR switch from Redis queues to Redis stream to notify all services of product updates, including search-a-licious.

We cap the number of events to ~10000 for the moment.

See https://redis.io/docs/data-types/streams/ for more information about Redis streams.


john-gom:
Appreciate that storage could be an issue but how would a 10,000 event cap work when we do bulk updates to all products? Would it be better to have more of a time based cap instead?

raphael0202:
@john-gom good point, I didn't think of bulk updates, 10k is definitely not enough, we should support having X millions updates.
We can regularly trim it using [XTRIM](https://redis.io/commands/xtrim/) with a cron job.

raphael0202:
@john-gom I added you as a reviewer, as Alex is quite busy lately ;)

lib/ProductOpener/Products.pm
```diff
@@ -1452,11 +1452,11 @@ sub store_product ($user_id, $product_ref, $comment) {
 
 	$log->debug("store_product - done", {code => $code, product_id => $product_id}) if $log->is_debug();
 
-	# index for search service
-	push_to_search_service($product_ref);
+	my $update_type = $product_ref->{deleted} ? "deleted" : "updated";
+	# Publish information about update on Redis stream
+	push_to_redis_stream($user_id, $product_ref, $update_type, $comment, $diffs);
```
john-gom, line 1457: I'm concerned that if we have bulk updates that the diff could bloat the stream size. Do we need to send the diffs to redis? I couldn't see that search-a-licious is using them?

john-gom:
Not specifically to do with this PR, but should we pull the redis deployment out of search-a-licious so that it can be easily used in other projects too?

alexgarel:
> Not specifically to do with this PR, but should we pull the redis deployment out of search-a-licious so that it can be easily used in other projects too?

This is the target, but I think we can first try it like this.

lib/ProductOpener/Products.pm
raphael0202: Search-a-licious doesn't require the diff, but Robotoff does. And we plan to switch Robotoff to Redis streams as well. It only contains the names of the fields that were updated/deleted/created though, not the content.

john-gom: OK, I wasn't fully aware of what the diff contained

john-gom:
Hi @raphael0202  I've just pushed updates to upade_all_products.py so that it updates the last_modified_t and pushes to Redis, but only if something has actually changed (does a deep compare). This has a marginal impact on performance:

With 2340 products locally, with the deep clone, deep compare and streaming to Redis it takes 45 seconds, where it only took 28 s without the comparison. However, if there are no changes detected it takes 31 s with the deep compare (as it doesn't have to do any writes).

At the moment I'm setting the User for the updates as "update_all_products" and the diff is empty

john-gom:
> By the way, there is a potential issue here which is often resolved with a "transactional outbox" pattern (article here explains the problem in more detail: https://microservices.io/patterns/data/transactional-outbox.html)
> 
> Theoretically our message to Redis could fail but the Product Opener update could succeed, in which case search-a-licious, off-query, etc. would be out of date.
> 
> In our case we could send the message before committing to the database, as if the database update fails the user is likely to retry so we will get two refreshes downstream rather than one, which isn't really a problem. However, if we do this we need to be sure that the event subscriber doesn't process the event before we have actually committed the change to the database. I'm guessing this could theoretically possible, albeit unlikely.
> 
> It might be we can mitigate this by including the last_modified_t in the message and then the consumer can compare this with the product data they got back from the API. If the product value is earlier than the value in the message then the consumer could set the last message id to that of the previous message and then stop processing messages, so that the offending product is re-fetched on the next iteration. This does place a bit of a burden on the consumer though.
> 
> The transactional outbox approach would have us use the MongoDB change log to generate messages, rather than doing this in the Product Opener code. This might be a better option with this PR in place as now for jobs like update_all_products we don't update MongBD if nothing has changed.
> 
> Any thoughts on this?

Had a few more thoughts this morning. Given that this scenario would be quite unusual (can write to MongoDB but not to Redis) then we could simply have some kind of watchdog process to see if it is happening. I could adapt off-query to keep track of this - so normally it would receive updates via redis but then every evening it could do a manual sync using a last_modified_t based query on MongoDB. If a record is found with the latter that wasn't received via the former then we could log an issue and potentially re-generate the event.

scripts/update_all_products.pl
```diff
@@ -1347,7 +1349,18 @@
...
+			if (!$any_change) {
+				# Deep compare with original (if we don't already know that a change has been made)
+				$any_change = !Compare($product_ref, $original_product);
+			}
+			if (!$any_change) {
+				print STDERR ". Skipped";
+			}
+		}
+
+		if ($any_change) {
```
alexgarel, line 1363: is'nt it:
```suggestion
		if ($any_change and (!$pretend)) {
```

scripts/update_all_products.pl
```diff
@@ -1359,21 +1372,24 @@
 
 			# Otherwise, we silently update the .sto file of the last version
 			else {
-
 				# make sure nutrient values are numbers
 				ProductOpener::Products::make_sure_numbers_are_stored_as_numbers($product_ref);
 
+				# Make sure product _id and code are saved as string and not a number
```
alexgarel, line 1378: I really find it a pity that we have this kind of duplicate code instead of Products.pm handling all this (may it with special parameters).

alexgarel (approved):
I approve but there are two important points:

- :bomb: I think $pretend is not honored in your current code
- A test would be great

alexgarel:
@raphael0202 before we deploy that to production we would need to have a shared REDIS (on free imo, with stunnel for communications)

raphael0202:
I added an integration test that checks in the log that the Redis call was successful. I also added Redis to `docker-compose.yml` docker file so that we can have a Redis locally and in staging (it makes more sense for it to be part of Product Opener docker-compose file than search-a-licious).

scripts/update_all_products.pl
raphael0202: @alexgarel it's @john-gom who did this change, I'm not familiar with this part of the codebase

PR #9432, opened 2023-11-30
chore: Adding Intake24 category codes to the category taxonomy

aleene:
### What
There is an open source software called Intake24 that is used for cohort studies (similar to NutriNet-Santé) in several countries (it started in the UK but is used in several other countries now, and France will use it as well). The French team that uses it would like to link Intake24 to OFF, so that cohort participants can search food in Open Food  Facts. For that, we would need to map the categories from Intake24 to OFF categories (and possibly create some categories if we miss them). Could you help looking up the corresponding OFF categories by completing some of them in this table? https://docs.google.com/spreadsheets/d/1PE2HLKeNC44a8JkzRSgKg5KMW5IXZIxg/edit#gid=1916021440
Once the table is complete, we could also add the Intake24 categories as properties in our own categories taxonomy.


aleene:
I can not find the error in the files. I repaired those. It seems like the test  is based on the old taxonomy?

stephanegigandet (approved):
Thank you @aleene!

stephanegigandet:
> I can not find the error in the files. I repaired those. It seems like the test is based on the old taxonomy?

I reenabled the taxonomy suggestion tests and updated their results.

alexgarel:
As always incredible work @aleene ! Kudos !

PR #9439, opened 2023-11-30
fix: ingredient additions to norwegian bokmål

odin-h:
Added some new abbreviations, made a minor regex tweak for Nynorsk, and added an ignore_regexps for cocoa percentage that can sometimes be found at the bottom of the ingredients list for chocolates.


stephanegigandet (approved):
Thank you!

PR #9441, opened 2023-11-30
chore: temporarily remove X-Frame-Options DENY header

stephanegigandet:
this is to allow graphs to be embedded in an iframe


jayaddison:
@alexgarel @stephanegigandet are those graphs still used/embedded from somewhere?

NB: as far as I know, `X-Frame-Options` is still widely-supported; however there is also a modern alternative that can be embedded in `Content-Security-Policy` -- the [`frame-ancestors` directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors).

PR #9442, opened 2023-11-30
fix: Reduce size of image on donate banner

Valimp:
Fix :
minimize size of image on donate banner

@stephanegigandet


alexgarel (approved):
thanks !

Issue #9311, opened 2023-11-13
Wrong pagination link for facet page on world in another language


hangy:
### What

Pagination link is broken for non-default language.

### Steps to reproduce the behavior

1.Open a list of brands on OFF in non-default language, ie. World in German, ie. https://world-de.openfoodfacts.org/brands
2. Click on the page number 2
4. See error, because https://world-de.openfoodfacts.org/marken/brands/2 was opened

### Expected behavior

The href should have pointed to https://world-de.openfoodfacts.org/brands/2 or https://world-de.openfoodfacts.org/marken/2

### Why

It's currently not possible to browse a country's facets in another language, which could hinder some contributors.

hangy:
After some research, I think this could be a duplicate of #3850. It wasn't obvious from the get-go, because it doesn't actually seem to be a language issue, but a general issue with the face pagination.

alexgarel:
seems plausible that it's a duplicate of #3850 as the problem there is on facets that are not real tags.

alexgarel:
It's because we use `groupby_tagtype` in `display_pagination`. We should add a parameter in the request to give the right name.

alexgarel:
I will propose a PR

Issue #3850, opened 2020-07-15
Navigating to the 2nd page of contributors is broken


teolemon:
### What
- Navigating to the 2nd page of contributors is broken: the bottom links redirects to users/
- https://world.openfoodfacts.org/contributors/users/2

### Code pointer
```html
<ul id="pages" class="pagination"><li class="unavailable">Pages:</li><li class="current"><a href="">1</a></li><li><a href="/contributors/users/2">2</a></li><li><a href="/contributors/users/2" rel="next$nofollow">Next</a></li></ul>
```

### Screenshot
Still in production with the redesign (Nov 2022)

![Image](https://user-images.githubusercontent.com/1689815/201684105-cb6998e1-9173-42f5-a616-5263ac221199.png)

### Part of
- #5529

jolesh:
Yes the link doesn't work, where have you found it?

Ban3:
@jolesh https://world.openfoodfacts.org/contributors You can get there from homepage by selecting `Contributors` from `Explore products by...`

jolesh:
Ah thanks, found it

teolemon:
Given that we have fixed perf issue on the contributors page, this issue is now impactful again.

teolemon:
I have spotted a similar issue on packager codes: I have a similar issue on the packager codes: https://world.openfoodfacts.org/packager-codes/emb_codes/2

PR #9443, opened 2023-11-30
fix: fixing pagination on facets

alexgarel:
* fixing link in pagination for contributor and translated facets
* fixing pagination for contributors query

fixes:
- #9311 
- #3850


stephanegigandet (approved):
Looks good!

PR #9457, opened 2023-12-01
taxonomy: More Wikidata entries for amino acids

teolemon:
### What
More Wikidata entries for amino acids


benbenben2 (approved):
Looks good to me

PR #9459, opened 2023-12-01
chore: translation of labels

teolemon:
### What
chore: translation of labels


teolemon:
ERROR - fr:durable already is associated to en:durable (labels) - fr:durable cannot be mapped to entry en:sustainable

benbenben2 (approved):
Looks good to me

PR #9474, opened 2023-12-05
fix: move Intermarché / Les Mousquetaires import to producers platform

stephanegigandet:
A few things to note:
- Intermarché wants us to remove the files from the SFTP as we process them, thus we move them to a backup folder, so that we can reprocess them if needed
- The files are named in DD-MM-YYYY format and sometimes we get multiple months uploaded at the same second, we rename them to YYYY-MM-DD so that can import them in order
- The CSV files are in OFF format, but they have a byte order mark which breaks Text::CSV, so we remove the mark first


alexgarel (approved):
Thanks !

PR #9478, opened 2023-12-05
taxonomy: add translations for Label categories

teolemon:
### What
taxonomy: add translations for Label categories


benbenben2 (approved):
LGTM

PR #9479, opened 2023-12-05
taxonomy: add translations for shapes and materials

teolemon:
### What
taxonomy: add translations for shapes and materials


benbenben2 (approved):
LGTM

Issue #7768, opened 2022-11-28
Serving_quantity is wrongly computed based on serving_size


CharlesNepote:
### Describe the bug

In [this product](https://world.openfoodfacts.org/product/0851042007034/prebiotic-fiber-protein-whole-plant-smoothie) I filled serving size = 12 fl oz and the api answers "serving_quantity": null, "serving_size": "12 fl oz".

Other example: https://world.openfoodfacts.org/product/10662564/grand-arome-poulain
![image](https://user-images.githubusercontent.com/562352/204342941-84216efe-d952-42a0-8570-858a40df30fe.png)

### To Reproduce

See previous examples.

### Expected behavior

Either compute the good result.

Either the user that the serving_size is not recognized to compute the result.

### Type of device

Browser

benbenben2:
Additionally:
serving_size="0.5 cup (130 g)", leads to serving_quantity=130

Which is as expected, whereas the opposite:
serving_size="130 g (0.5 cup)", leads to serving_quantity=0.0 

The subroutine extracting the serving_quantity is in lib/ProductOpener/Units.pm, and more precisely line 297:
"$serving =~ /^(.*[ \(])?(?<quantity>(\d+)(\.|,)?(\d+)?)( )?(?<unit>\w+)\b/i"

benbenben2:
Example 1 is solved
Example 2 is solved
Example 3 is NOT solved (is still the same, 15). But again, what would you expect in this case?
Example 4 is solved

I think that @stephanegigandet managed to improve considerably the regex for serving_size. 

That may close this issue @CharlesNepote?

CharlesNepote:
Example3: I would expect `215`, don't you think so?

Maybe we can open another issue for the example 3?

jusdekiwi:
![image](https://github.com/user-attachments/assets/e164bc9c-7ff8-487b-8128-c357c4ee017c)
![image](https://github.com/user-attachments/assets/d18d799a-ee31-4553-843c-1a2aab9db3e6)

"serving size: 1 cup" is still interpreted as 0

PR #9484, opened 2023-12-05
taxonomy: dq_serving_size_parsing_improvements

benbenben2:
### What
Reviewed the Mirabelle results from the issue for serving size (see below)

> Here are 100 examples randomly: http://mirabelle.openfoodfacts.org/products?sql=select+code%2C+url%2C+creator%2C+created_datetime%2C+last_modified_datetime%2C+product_name%2C+serving_size%2C+serving_quantity%0D%0Afrom+%5Ball%5D+%0D%0Awhere+%28serving_quantity+%3D%3D+%22%22+or+serving_quantity+IS+NULL%29+and+%28serving_size+%21%3D+%22%22%29%0D%0Aorder+by+random%28%29%0D%0Alimit+100

Brazil (pt:) added "gm". 5,484 rows
xx:onz, 128 rows
en:metric teaspoon 2 rows
	note that beside metric teaspoon, teaspoon or tsp depends on the country (us, uk, canada)
	same for cup and tbsp which is also a factor of the density (i.e., 1 tbsp of water will be different than 1 tbsp of sugar, for example)
gg instead of g is typo, not added in taxonomy. It has to be fixed by contributors.

- Related to #7768

### DIFF
--- a/lib/ProductOpener/Units.pm
+++ b/lib/ProductOpener/Units.pm
@@ -283,8 +283,6 @@ sub normalize_serving_size ($serving) {
 
 		return unit_to_g($q, $u);
 	}
-
-	#$log->trace("serving size normalized", { serving => $serving, q => $q, u => $u }) if $log->is_trace();
 	return;
 }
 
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Thanks!

PR #9490, opened 2023-12-08
taxonomy: translation of the Open Beauty Facts body part taxonomy

teolemon:
### What
taxonomy: translation of the Open Beauty Facts body part taxonomy


benbenben2 (approved):
LGTM

PR #9498, opened 2023-12-08
fix: dq_sugar_plus_starch_greater_than_carbohydrates

### LINKED ISSUE #9267: Error Sugars plus starch greater than carbohydrates incorrectly shown
### What
The error `Nutrition - Sugars plus starch greater than carbohydrates` is shown for correctly entered data.

See https://world.openfoodfacts.org/product/01681321/cooked-ham-slices-sainsbury-s

### Expected behaviour
No error should be triggered

### Additional context
The error trigger procedure should ignore the less than fields

I assume this is an issue for all products which have a less than entry

### Time per product
a few minutes

benbenben2:
### What
See linked issue.

Suggested solution:
	trigger error if "<" symbol is missing
	do not trigger error if at least one nutriment has "<" symbol
		but trigger error if there is "<" symbol and 1 nutrient (sugar or starch) is above carbohydrates (that way we can still get error if the sugar or starch too high)

- Fixes #9267


stephanegigandet (approved):
Thank you!

PR #9502, opened 2023-12-08
taxonomy: jp_allergens_1

benbenben2:
### What
Added some allergens in JP. Added as synonym when available

according to this document: https://www.caa.go.jp/policies/policy/food_labeling/food_sanitation/allergy/assets/food_labeling_cms204_210514_01.pdf
see chapter 3 (page 28)
see also table of comparison with other countries regulations (page 7)

See also this document in English: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8691970/
See Table 1

- Fixes #-none-


benbenben2:
Reworked the order to make appear the one from the document in first position (those will appear on the website for allergens)

Different allergens from the Japanese document are found inside the same block in the taxonomy. Example: walnuts (くるみ) and almonds (アーモンド). Instead of splitting the block into two blocks (todo in the future?), I just added them as synonyms, and the first element of the line contains both with a slash: くるみ/アーモンド.

Tests containing "apple" failed after the changes (apple is an allergen in Japan). So, tests were updated.

benbenben2:
Replaced by #9502

PR #9512, opened 2023-12-11
feat: jp_allergens_2

benbenben2:
### What

Added feature to detect if sub-ingredients in parentheses contain text which imply allergens/traces (....を含む, or 一部に...を含む, for examples)
Added but commented out (see first comment below) the feature to detect if parentheses contain ONLY allergens (all subingredient should be allergens and known).

I would need you help/suggestion there for the devs/rework.

**FIRST COMMENT**
(A) When we have "香料 (ラッカセイ, 種実類, 魚).", the allergens are extracted from the text. It is already working like this (not related to this PR). 
But it is not well extracted. It turns out that allergens tag has only 1 (Peanuts, ラッカセイ) out of the 3 allergens (Fish, Nuts, Peanuts). Whereas traces tag has 3 allergens (Fish, Nuts, Peanuts)

However - similar to "pasta (gluten), water" (see Croatian example of the PR and this: https://openfoodfacts.slack.com/archives/CC1FPRJ8G/p1698956397179069) - we could wish to not only add allergens to allergens tags but also remove them from the ingredients list. To not impact nova score. It is tackled and works for a single subingredient BUT it does not work for more sub-ingredients. It seems to me that (A) takes precedence over the commented block in my PR that should handle it. 

SECOND COMMENT
(一部に卵・小麦・えび・豚肉・鶏肉・大豆・やまいも・もも・りんごを含む) -> seems to me that it is equivalent to "may contain traces of blabla". It is found at the end of ingredients list only (Naruyoko, correct me if I am wrong). BUT it is in parenthesis. And I have found more convenient to handle it similarly to allergens written parenthesis and ending (as well) with "を含む". And it is working fine. See example with "ミートボール(一部に卵・グルテン・乳成分・大豆を含む)". Another option - that I did not investigate - could be to rework existing code for "may contain traces of blabla" to work when there are parenthesis. 

### Screenshot

Screenshot_20231211_173857
---------
Screenshot_20231211_173917
--------
Screenshot_20231211_173934

- Fixes #-none-


lib/ProductOpener/Ingredients.pm
```diff
@@ -1960,6 +1960,76 @@ sub parse_ingredients_text_service ($product_ref, $updated_product_fields_ref) {
...
+					# allergens or traces
+					# single allergen in parenthesis, for example: Krupica od durum pšenice (gluten) -> durum wheat semolina (gluten)
+					# more than a single allergen is handle just after
+					# in Japanese, 豚肉を含む (contains (を含む) pork (豚肉))
+					# 一部に卵・小麦・乳成分・大豆を含む (contains (を含む) parts of (一部に), eggs, wheat, milk, soybeans (卵・小麦・乳成分・大豆)
+					if (
+						(
+							($between !~ /$separators|$and/)
+							and (
+								exists_taxonomy_tag(
```
stephanegigandet, line 1972: This is problematic because the allergens taxonomy contains tons of synonyms that are in fact not synonyms of the allergen, but ingredients that contain that allergen (e.g. "en:milk" has synonyms for all kinds of cheeses).

That means that if we have "Cheese (parmigiano reggiano)", "parmigiano reggiano" will be in the allergens taxonomy. We do not want to remove it.

One possibility could be to canonicalize the ingredient in parenthesis using the ingredients taxonomy, and check whether it matches the exact allergen:  en:milk, or en:gluten.

stephanegigandet:
Some tests where failing that showed unwanted removal of subingredients because they are in the allergens taxonomy (even though they are not really allergens, they are listed as synonyms because they contain the allergen...)

I added a few more tests in ingredients.t to make sure that we don't remove sub ingredients that we want to keep.

lib/ProductOpener/Ingredients.pm
benbenben2: ah yeah, you got a point there. It will be an issue:
- for milk vs lactose, both can appear in parentheses for allergens :(
	https://world.openfoodfacts.org/product/0074175597009/blue-frosted-sugar-cookies-cleo-leo
	https://world.openfoodfacts.org/product/03003491216/classic-condensed-soup-cream-of-chicken

- Is it an option to... split allergens taxonomy in two? oO (1. the allergens and 2. the substances containing allergens) 
- Is it an option to make a list of reference allergens? (maybe directly in Ingredients.pm)
- Otherwise, I will have to roll back these changes and keep only condition for Japanese keyword

Naruyoko:
(一部にX・Y・Xを含む) at the end is just an alternative format to list allergens, putting it all in a single list rather than placing it after each ingredient. They are **not** traces information (these are placed outside ingredients labels).

lib/ProductOpener/Ingredients.pm
```diff
@@ -1960,6 +1960,76 @@ sub parse_ingredients_text_service ($product_ref, $updated_product_fields_ref) {
...
+									"allergens", canonicalize_taxonomy_tag($ingredients_lc, "allergens", $between)
+								)
+							)
+						)
+						or ($between =~ /を含む/)
```
Naruyoko, line 1977: Would it make sense to also match ~由来 ("derived from ~") here as well? They are used for additives.

benbenben2: seems a good idea yes

香料(乳･小麦･大豆･りんご由来) 
https://world.openfoodfacts.org/product/4901335509656/corn-snack-bbq-flavor-koikeya

香料(乳由来)
https://world.openfoodfacts.org/product/4901111201668/%E3%82%B9%E3%83%86%E3%82%A3%E3%83%83%E3%82%AF-%E7%B4%85%E8%8C%B6%E3%82%AA%E3%83%AC-%E5%91%B3%E3%81%AE%E7%B4%A0-%E3%83%96%E3%83%AC%E3%83%B3%E3%83%87%E3%82%A3

乳化剤(大豆由来)
https://world.openfoodfacts.org/product/4953823097626/%E3%83%A2%E3%83%8A%E7%8E%8B-%E3%83%90%E3%83%8B%E3%83%A9-lotte

lib/ProductOpener/Ingredients.pm
benbenben2: I apply changes to do what you wrote. But again, that will not work...

lib/ProductOpener/Ingredients.pm
```diff
@@ -1971,6 +1971,85 @@ sub parse_ingredients_text_service ($product_ref, $updated_product_fields_ref) {
...
+					$log->debug("parse_ingredients_text - BEFORE2 $ingredients_lc:$between.")
+						if $log->is_debug();
+					if (
+						(
+							($between !~ /$separators|$and/)
+							and (
+								(
+									exists_taxonomy_tag("allergens",
+										canonicalize_taxonomy_tag($ingredients_lc, "allergens", $between))
+									and (canonicalize_taxonomy_tag($ingredients_lc, "allergens", $between) eq "en:"
```
stephanegigandet, line 1992: The comparison to "en:" . $between will only work if ingredients are in English, and if $between is simple, like "gluten".
It's better to canonicalize $between with the ingredients taxonomy (not the allergens taxonomy) and check if it is en:milk etc.

stephanegigandet:
@benbenben2 I made 2 changes:
1. canonicalize the single sub-ingredient with the ingredients taxonomy (not allergens taxonomy) to check if it is the actual allergen (e.g. en:milk), and not an ingredient that contains the allergen (e.g. en:cheese)
2. check that the parent ingredient is just not a label, so that something like "MSC (fish)" is considered to be the MSC label applied to fish, and not a MSC ingredient with a fish allergen.

stephanegigandet:
@benbenben2 should we merge this branch as-is?

benbenben2:
`1. canonicalize the single sub-ingredient with the ingredients taxonomy (not allergens taxonomy) to check if it is the actual allergen (e.g. en:milk), and not an ingredient that contains the allergen (e.g. en:cheese)`
Yes, it took me some time to understand, but now I get it. This is super!


`2. check that the parent ingredient is just not a label, so that something like "MSC (fish)" is considered to be the MSC label applied to fish, and not a MSC ingredient with a fish allergen.`
Perfect too

`benbenben2 should we merge this branch as-is?`
I think so :)

stephanegigandet (approved):
Thank you!

PR #9513, opened 2023-12-11
docs: OpenAPI nutrients enhancements

alexgarel:


benbenben2 (approved):
Looks good to me

docs/api/ref/schemas/product_nutrition.yaml
```diff
@@ -82,7 +106,7 @@ properties:
           imports.
         example: 12.5
     patternProperties:
-      '(?<nutrient>\w+)_unit':
+      '(?<nutrient>[\w-]+)_unit':
```
stephanegigandet, line 109: Good :)

docs/api/ref/schemas/product_nutrition.yaml
```diff
@@ -11,15 +11,38 @@ properties:
...
+  nutrition_data_per:
+    type: string
+    enum:
+    - serving
+    - 100g
+    description: |
+      The nutrition data on the package can be per serving or per 100g.
+
+      This is essential to understand if `<nutrient>_value` and `<nutrient>` 
+      values in `nutriments` applies for a serving or for 100g.
```
stephanegigandet, line 23: Can we add something like: "When writing products, this setting applies to all existing nutrients values for the product, not only the nutrient values sent in the write request. So it should not be changed unless all nutrients values are provided with values that match the nutrition_data_per field."

stephanegigandet (approved):
Thank you! I added one suggestion.

PR #9515, opened 2023-12-12
taxonomy: Remove redundant alias

hangy:
### What

Remove redundant `es:Calidad certificada` taxonomy label entry.


stephanegigandet (approved):
Thanks!

Issue #9412, opened 2023-11-27
Folksonomy Engine isn't displayed for some products


CharlesNepote:
At least: https://world.openfoodfacts.org/product/3033491485763/hipro-25g-proteines-vanille-cookie-danone

The template is simply not loaded by Product opener. I don't see any reason why.

alexgarel:
I find the problem:

Folksonomy script is searching for the div to add the properties to with `$("div[itemtype='https://schema.org/Product']")`

while this product has `itemtype='https://schema.org/DietarySupplement'` instead !

This property itemtype is set by Display.pm::display_product

I propose to use `$("div.main-product")` instead !

PR #9517, opened 2023-12-12
fix: allow folksonomy engine on dietetary products

alexgarel:
fixes: #9412

### DIFF
--- a/html/js/folksonomy.js
+++ b/html/js/folksonomy.js
@@ -146,7 +146,7 @@ function folskonomy_engine_init() {
  */
 function displayFolksonomyPropertyValues() {
     //$(".details").before(
-    $("div[itemtype='https://schema.org/Product']").append(
+    $("div.main-product").append(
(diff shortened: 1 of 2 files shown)


html/js/folksonomy.js
```diff
@@ -146,7 +146,7 @@ function folskonomy_engine_init() {
  */
 function displayFolksonomyPropertyValues() {
     //$(".details").before(
-    $("div[itemtype='https://schema.org/Product']").append(
+    $("div.main-product").append(
```
hangy, line 149: Not sure if it fits this PR, but if used in this way, it might be better to make it `div id="main-product"` instead of a `class`?

hangy (approved):

html/js/folksonomy.js
alexgarel: you are right, I will do both to avoid disturbing CSS :-)

PR #9518, opened 2023-12-12
taxonomy: Bulgarian additions

Roto15:


Roto15:
@stephanegigandet Can you just check if the added PDO products in Categories are on the right place? Thanks

benbenben2 (approved):
Very good

PR #9519, opened 2023-12-12
feat: eggs_to_g_in_eu

### LINKED ISSUE #1416: Compute caliber or # of eggs
S: 43-53g
M: 53-63g
L: 63-73g
XL: 73g+

according to Test Achats

- Use the weight + Label of caliber to compute probable number of eggs
- Use the weight + number of eggs in quantity or product name to compute probable caliber

benbenben2:
### What
add eggs weight range according to EU regulation
https://eur-lex.europa.eu/legal-content/en/TXT/?uri=CELEX%3A32023R2465

- Fixes #1416


teolemon:
- @benbenben2 you can have a look at: https://github.com/openfoodfacts/openfoodfacts-server/issues/1416

stephanegigandet:
In practice for eggs we don't have a weight, and the quantity field contains a number of eggs (e.g. "6", sometimes "6 eggs").

So I'm thinking that we are trying to resolve an issue that does not really exist, and I would be in favor of just closing https://github.com/openfoodfacts/openfoodfacts-server/issues/1416#issuecomment-1862648382 (I added a comment there).

Do you have a different view?

benbenben2:
> In practice for eggs we don't have a weight, and the quantity field contains a number of eggs (e.g. "6", sometimes "6 eggs").
> 
> So I'm thinking that we are trying to resolve an issue that does not really exist, and I would be in favor of just closing [#1416 (comment)](https://github.com/openfoodfacts/openfoodfacts-server/issues/1416#issuecomment-1862648382) (I added a comment there).
> 
> Do you have a different view?

I agree that computing probable number of eggs and computing probable egg category (caliber) would 1) almost never be possible and 2) not be of any use (or?)

I was not aware of the issue #1416, when I worked on this PR.
And actually my approach was the other way around. 

The quantity for eggs product is almost never recognized (I have seen, in HR and FR, some eggs products with different eggs categories (caliber), in which case, the maximum weight is written on the package, but this is quite rare).

How to transform "10 eggs category XL" into something quantifiable. That is the purpose of this PR.
I believe that it is still a good feature. And that it would allow solving "en:Quantity not recognized" for eggs products.

It is maybe still incomplete, because not many people are going to add the eggs' category in the quantity field, but this can be completed later with a different data quality info/warning or something else.

stephanegigandet:
I think it would be useful to look at all the cases where we have a number of "units" in the quantity field. Eggs is a very common case. But there are others, like cucumbers, coconuts etc. Some may specify calibers like eggs, e.g. 300/400g for a cucumber.

My initial idea was that for those kind of products, the unit is just "unit" ("pièce" in French). I don't think we should multiply the number of units by having things like "egg", "cucumber" and "coconut" in units. And if we put calibers as well in units, then we could have tons of them.

benbenben2:
I did not know that we want to limit number of units in the taxonomy file.

What will you do with eggs then is not clear to me. Let's say we have 10 eggs (10 pieces is written in the quantity field, or simply 10). I guess, you can fetch that these are eggs based on the category. So, we will know that we have 10 eggs. Is it all? If we could get to know the caliber/category of the eggs, we could have a rough estimation of the total weight of the product.

benbenben2:
I am closing this PR. Opened for too long.

PR #9526, opened 2023-12-13
chore: pates a tartiner

aleene:
### What
Resolve the issue with pate a tartiner. it does not have an english variant and introduces a french tree. It is specific to France.


taxonomies/categories.txt
```diff
@@ -92924,13 +92924,16 @@ fi:Tyrnihyytelöt
 fr:Gelées d'argousiers
 
 <en:Sweet spreads
+#en:Sweet spreads in France
```
benbenben2, line 92927:
```suggestion
en:Sweet spreads in France
```
I would keep en: as first line. Or eventually: "en:Pâtes à tartiner"

benbenben2 (approved):
Looks good to me. 

Check my comment before to merge

taxonomies/categories.txt
aleene: OK I'll do that

PR #9527, opened 2023-12-13
quick-start docs - "make dev" errors

rkiddy:
As I find the solution, I will document it here.


rkiddy:
My laptop's Docker installation is horked and I do not think it is worth trying to figure out the problem. I was able to create an ec2 instance and build the server there. Kind of lame, but there it is.

PR #9528, opened 2023-12-13
feat: dq_ignore_nutrition_if_single_value

### LINKED ISSUE #1759: Ignore nutrition values if salt is the only value passed and it is 0
Yuka has been sending almost empty nutrition facts, with just 0 for salt. (currently 3500 products): https://world.openfoodfacts.org/quality/nutrition-only-has-zero-salt-and-sodium

1. Create a rule to ignore salt if it's the only value passed and it is 0.

2. Use a script to update existing products.

benbenben2:
### What

Empty the nutrition table, when a **single** value has been input
Not only for 0. Generalized to any integer.

- Fixes #1759

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -2920,6 +2920,16 @@ sub assign_categories_properties_to_product ($product_ref) {
 This function reads the nutriment values passed to the product edit form, or the product edit API,
 and assigns them to the product.
 
+=head3 Arguments
+
+=head4 $product_ref
+
+=head4 $nutriment_table
+
+The nutriment table format according to the country as defined in Diplay.pm
+
+=head4 $can_edit_owner_fields
+
 =cut
 
 sub assign_nutriments_values_from_request_parameters ($product_ref, $nutriment_table, $can_edit_owner_fields = 0) {
@@ -3093,6 +3103,34 @@ sub assign_nutriments_values_from_request_parameters ($product_ref, $nutriment_t
 			delete $product_ref->{nutriments}{$nid . "_unit"};
 			delete $product_ref->{nutriments}{$nid . "_label"};
 		}
+	}
+
+	# Ignore changes if a single nutrient is zero but all others are undefined (see issue #1759)
+	my $ncount = 0;
+	# if salt (sodium) is entered and sodium (salt) is missing,
+	# values are automatically added, leading to 2 values instead of 1 in the nutriments
+	my $salt_or_sodium = 0;
+	foreach my $nid (keys %{$product_ref->{nutriments}}) {
+		# regardless if it is prepared or not
+		# all nutrients (not only majors one) to avoid false positives for food supplements or water
+		# exclude salt (sodium) if sodium (salt) has already been counted
+		if (($nid =~ "_value") && !((($nid eq "salt_value") or ($nid eq "sodium_value")) and ($salt_or_sodium == 1))) {
+			$ncount++;
+			if (($nid eq "salt_value") or ($nid eq "sodium_value")) {
+				$salt_or_sodium++;
+			}
+		}
+
+		# not need to continue if we have at list 2 values
+		if ($ncount > 1) {
+			last;
+		}
+	}
+	if ($ncount < 2) {
+		$log->debug("assign_nutriments_values_from_request_parameters - single nutriment. Empty table.")
+			if $log->is_debug();
+		$product_ref->{nutriments} = {};
+		return;
 	}
 
 	if ((defined $product_ref->{no_nutrition_data}) and ($product_ref->{no_nutrition_data} eq 'on')) {
--- a/stop_words.txt
+++ b/stop_words.txt
@@ -60,6 +60,7 @@ dans
 dataset
 datasets
 de
+Diplay
 dont
 dropdown
 ecoscore
--- a/tests/unit/food.t
+++ b/tests/unit/food.t
@@ -550,4 +550,73 @@ is_deeply(
 	}
 ) or diag explain $product_ref;
 
+# Empty nutrition table if a single nutrient has been input only (see #1759)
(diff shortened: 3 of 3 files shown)


stephanegigandet:
@benbenben2 I don't think we should generalize not saving single values. #1759 was to fix a clear bug from Yuka. Discarding nutrient data for just salt: 0 is I think safe enough (but probably not that useful anymore as Yuka seems to have fixed it). But we should not extend that to other nutrients. e.g. "energy kcal 0" could be a meaningful value, there are lots of products that have logs saying "0 calories", and we could use that to populate just energy-kcal.

We also can't generalize to values than are not 0. The "salt 0" was a bug, but if there's a non zero value, it's not a bug anymore, it's intended by the user.  There are also products like wine where you will have only one value for "alcohol".

benbenben2:
> @benbenben2 I don't think we should generalize not saving single values. #1759 was to fix a clear bug from Yuka. Discarding nutrient data for just salt: 0 is I think safe enough (but probably not that useful anymore as Yuka seems to have fixed it). But we should not extend that to other nutrients. e.g. "energy kcal 0" could be a meaningful value, there are lots of products that have logs saying "0 calories", and we could use that to populate just energy-kcal.
> 
> We also can't generalize to values than are not 0. The "salt 0" was a bug, but if there's a non zero value, it's not a bug anymore, it's intended by the user. There are also products like wine where you will have only one value for "alcohol".

Okay. Those are good inputs. I reviewed again on Mirabelle possibilities of false positives. 
-  (very often) table is not totally filled by contributor but there is a picture, and often this single value is correct 
- picture is truncated and we see only 1 value
- quality of picture is so bad that we cannot read all nutrient
- sometimes, however, it is really wrong data


This being said, since the code is there, and since many of the cases are due to the table not being totally filled, I am now thinking about converting this into a new data quality facet (if there is no similar, yet).
What do you think @stephanegigandet ? Also, @CharlesNepote what do you think?

Good point for the energy that can be only input value (I did not see only kcal or only kj yet (only both in the examples I could find), but I suppose that it can happen). There may be more tables having single nutrient (alcohol, acidity).

CharlesNepote:
I'm sorry to reply so late.

For the moment, since there is no possible feedback for data quality warnings/info, I find those useless. There is enough work on data quality errors, why should contributors care about warnings containing so much false positive. So my questions are: when can we consider some entries lead to an error? are the impacts high?

* alcohol value only: that's possible when a bottle of wine only has the mandatory alcohol percentage (this is quite frequent)
* energie value: that's possible to have the energy on the front of pack but no photos of the nutrients' table
* fat value: that's possible for waters; the nutrition might be incomplete but this is leading to an error. eg. https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=0044850406056
* sat fat: while it can be incomplete data, it won't produce any issue (the Nutri-Score isn't computed in this case)
* carbs: idem
* sugar: idem
* fibers: idem; sometimes it's written on the front of pack: xx% of fibers
* proteins: idem
* salt/sodium: idem

So we don't have any certainty that a unique value in one of those field is false.
That said, many products containing only one value **are** false data.

### What are the impacts?

DuckDB comes to the rescue: it allows use REGEXP in the where clauses.
```sql
-- see https://github.com/openfoodfacts/openfoodfacts-server/issues/7660#issuecomment-1861286612 to create tempo.db
-- then duckdb tempo.db
select 0
+ (select count(code) from products where "energy-kj_100g" is not null and columns('^[^energy\-kj].*_100g') is null) 
+ (select count(code) from products where "energy-kcal_100g" is not null and columns('^[^energy\-kcal].*_100g') is null) 
+ (select count(code) from products where fat_100g is not null and columns('^[^fat].*_100g') is null) 
+ (select count(code) from products where "saturated-fat_100g" is not null and columns('^[^saturated\-fat].*_100g') is null) 
+ (select count(code) from products where carbohydrates_100g is not null and columns('^[^carbohydrates].*_100g') is null) 
+ (select count(code) from products where sugars_100g is not null and columns('^[^sugars].*_100g') is null)
+ (select count(code) from products where fiber_100g is not null and columns('^[^fiber].*_100g') is null) 
+ (select count(code) from products where proteins_100g is not null and columns('^[^proteins].*_100g') is null)
+ (select count(code) from products where salt_100g is not null and columns('^[^salt|sodium].*_100g') is null)
;
73207
```
~2.5% of the products have one of those fields completed (energy-kj_100g, energy-kcal_100g, fat_100g, saturated-fat_100g, carbohydrates_100g, sugars_100g, fiber_100g, proteins_100g, salt_100g) but not other nutrients' value.

The impacts are not that high:
* some people might see a false value on less than 2.5% products
* the Nutri-Score is not computed in those cases
* it might have an impact on scientific studies based on OFF data but in general scientists are cleaning/filtering the data

### What are the options?

1. **Do nothing**. Some of these values are true so we can keep them as is. We already ask users to complete the data if they want the Nutri-Score.
2. We could **delete some nutrients** when **there is no picture at all** to prove there are true. It could be done but we should try to delete the nutrient that does make sense. For example, we could do this based on the mean of the main category (Mirabelle could help to produce such lists or some IA techniques).
3. We could **trigger an error** when there are pictures, but no picture is selected for the nutrients' table to prove the values are true. It would allow the data quality team to: either select a picture where these data can be controlled; either delete the data. It could be called something like: `No picture to check weird nutrition facts`.

I would be in favor of 2 and 3, but I won't make it a top priority.

benbenben2:
Thank you very much @CharlesNepote for these inputs.

Now, I don't think we should put more effort into this potential new facet. We have already enough tasks in the backlog. 

Note that there is already 1 error when "en:nutrition-values-are-all-identical" and above 1, as well as 1 warning when "en:nutrition-3-or-more-values-are-identical".

I will close this PR.

We should close the corresponding issue too if it is not needed anymore (#1759)

PR #9530, opened 2023-12-14
taxonomy: Add (gym) protein nutrients

duhow:
Multiple "sport suplements" contain differnt proteins that can be listed.
Most commonly there is `creatine`, but there are other aminoacids that are interesting.


stephanegigandet (approved):
Thanks

PR #9541, opened 2023-12-15
taxonomy: Add a color taxonomy to assist ingredient parsing

teolemon:
### Idea
- Add a color taxonomy to assist ingredient parsing
### Why
- Multiple Color Variants: Many vegetables and fruits exist in multiple color variants. For example, bell peppers can be red, green, yellow, or orange; carrots can be found in purple, orange, yellow, or white; and tomatoes can vary from red to yellow, green, or even purple. Recognizing these variations is crucial for accurate identification.

Feel free to close if it's a silly idea


stephanegigandet:
In most cases I think we want different taxonomy entries because the fruits / vegetables are different varieties (e.g. green onions, citrons verts etc.).

PR #9544, opened 2023-12-16
chore: Re-enable devcontainers (experimental) 

hangy:
Updated to the newer [devcontainer spec](https://containers.dev/), this allows using Perl in vscode via `Perl::LanguageServer`.

Note that this doesn't have feature parity with what's available in the Makefile, yet. Notably, tests don't work, yet, and it might be beneficial to run `make dev`  before opening the devcontainer.


stephanegigandet (approved):
Thanks!

I'd love to have a demo of what this allows to do someday.

PR #9547, opened 2023-12-17
chore: Resolve some issues brought up by sonarcloud

hangy:
- Replace `var` by `let` or `const`
- Use `for of` loop instead of `for` loop
- Use `RegExp.exec` instead of `String.match`
- Exclude some obsolete files from analysis
- Move some old scripts to obsolete directory
- Minor code formatting


alexgarel (commented):
Great work ! :tada:

alexgarel (approved):

PR #9561, opened 2023-12-20
taxonomy: add new cream ingredient

duhow:
For https://es.openfoodfacts.org/producto/8480000155979/caocream-hacendado ,

The listed ingredient `cream` is non-milk, but made of sugar and oils.
Created new ingredient as `cream filling`.


stephanegigandet (approved):
Thanks

PR #9565, opened 2023-12-21
fix: export_producers_platform_data_to_public_database.sh

stephanegigandet:
multiline arrays don't work in sh, needs bash


alexgarel (approved):
Thanks !

PR #9564, opened 2023-12-20
taxonomy: German vitamins

moon-rabbitOFF:
### Related issue(s) and discussion

- Fixes #[ISSUE NUMBER]

PR #9567, opened 2023-12-21
taxonomy: vit. abbreviation for vitamins

stephanegigandet:
In ingredients preparasing, replace Vit. by Vitamin in German.

+ added "vit." as a synonym of "vitamin" in ingredients.txt taxonomy so that it also applies if ingredients are not preparsed.

It should be a more generic solution than https://github.com/openfoodfacts/openfoodfacts-server/pull/9564


CharlesNepote (approved):
Fine.

Issue #9537, opened 2023-12-15
"Huile de palme, noisettes" wrongly converted to "Huile de palme, huile de noisettes"


stephanegigandet:
See famous product: https://fr.openfoodfacts.org/product/3017620422003/nutella-ferrero

PR #9568, opened 2023-12-21
fix: parsing of (fr) huile de palme, noisettes

stephanegigandet:
fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/9537


alexgarel (approved):
LGTM

PR #9570, opened 2023-12-21
feat: madenearme speedup

alexgarel:
Make madenearme faster.

Also add fast JSON implementations for other needs.


stephanegigandet (approved):
Looks good

PR #9573, opened 2023-12-21
fix: remove warnings for empty/undef quantities

stephanegigandet:
This is to remove warnings when the quantity is empty:

Use of uninitialized value $unit_id in hash element at /srv/off/lib/ProductOpener/Units.pm line 312.


alexgarel (approved):
Thanks !

PR #9581, opened 2023-12-21
feat: dq_correct_wrong_lang__for_tags

### LINKED ISSUE #7838: Create a script to correct all taxonomized values for categories that are stored in an incorrect language by storing them in the right language
### Why
- Some Eco-Score, including on important products are wiped out by apps which try saving back values to Open Food Facts without a proper logic
### What
- [ ] Create a script to correct all taxonomized values for categories that are stored in an incorrect language by storing them in the right language
  - [ ] It could either be generic, or focused towards the main culprit (Read #6043 for context)
- [ ] Those are created by apps which try saving back values to Open Food Facts without a proper logic
- [ ] An example: https://world.openfoodfacts.org/product/3258563180357/vinaigre-balsamique-de-modene-les-delices-de-belle-francebelle-france#

### Part of
- #5538

benbenben2:
### What
Python code to: 
 - fetch unknown tags for categories, allergens, etc.
 - check if each unknown tag exists in the taxonomy in different language
 - if yes, to all products having the tag in wrong language:
    - add existing tag in different language 
    - remove tag in wrong language 
   
Code is generalized to - not only categories, but also - additional tags fields.

Lots of comments in the code, which should be helpful for reviewer and user.

Instructions on how to run the code, test it, run it in dev vs prod, are provided on top of the file.

The code was run successfully - in **net** environment - for 4 different tags:
```
"en:teilentrahmte-milch" => "de:teilentrahmte-milch"
"en:trempettes" => "fr:trempettes"
"en:turrones-de-chocolate" => "es:turrones-de-chocolate"
"en:кисели-краставички" => "bg:кисели-краставички"
```

Corresponding to the following products that have been updated:
```
4316268628273,categories;en:teilentrahmte-milch;de:teilentrahmte-milch
4061462842764,categories;en:teilentrahmte-milch;de:teilentrahmte-milch
4316268153898,categories;en:teilentrahmte-milch;de:teilentrahmte-milch
4300175269025,categories;en:teilentrahmte-milch;de:teilentrahmte-milch
3596710467815,categories;en:trempettes;fr:trempettes
8410086391138,categories;en:trempettes;fr:trempettes
5410921014546,categories;en:trempettes;fr:trempettes
0234590005996,categories;en:trempettes;fr:trempettes
5295000700387,categories;en:trempettes;fr:trempettes
8480017232182,categories;en:turrones-de-chocolate;es:turrones-de-chocolate
8410223704418,categories;en:turrones-de-chocolate;es:turrones-de-chocolate
8410886113527,categories;en:turrones-de-chocolate;es:turrones-de-chocolate
8480017229793,categories;en:turrones-de-chocolate;es:turrones-de-chocolate
8480000362537,categories;en:turrones-de-chocolate;es:turrones-de-chocolate
3800203290881,categories;en:кисели-краставички;bg:кисели-краставички
3800229785491,categories;en:кисели-краставички;bg:кисели-краставички
3800500012469,categories;en:кисели-краставички;bg:кисели-краставички
3800214420604,categories;en:кисели-краставички;bg:кисели-краставички
3800128903026,categories;en:кисели-краставички;bg:кисели-краставички
5290113000089,categories;en:кисели-краставички;bg:кисели-краставички
3800203741154,categories;en:кисели-краставички;bg:кисели-краставички
3800234140018,categories;en:кисели-краставички;bg:кисели-краставички
```

- Fixes #7838

### DIFF
--- a/scripts/update_tags_per_languages.py
+++ b/scripts/update_tags_per_languages.py
@@ -1,0 +1,365 @@
+"""
+This Python code: 
+ - fetch unknown tags for categories, allergens, etc.
+ - check if each unknown tag exists in the taxonomy in different language
+ - if yes, to all products having the tag in wrong language:
+    - add existing tag in different language 
+    - remove tag in wrong language 
+
+- To use a virtual environment (depend on the OS:
+  https://python.land/virtual-environments/virtualenv)
+```python3.xx -m venv venv```
+```source venv/bin/activate```
+```pip install requests```
+
+Need to update user_id and password variables before to run the code.
+
+dev mode (dev = True) will:
+ - run in .net environment
+ - iterate over all tags until one unknown tag is found in the taxonomy (a)
+   for a different language.
+ - print possible_new_tag, already_referenced_tags, possible_wrong_language_tags
+ - iterate over a single product of all products having the tag corresponding to (a)
+ - will NOT update the products
+
+- to run tests
+```python3 -m unittest update_tags_per_languages_tests.py```
+
+
+The url used to update products should be updated to corresponding country. 
+Otherwise the language of the tags (example: categories_lc) will be updated to "en"
+Before to update all products, comment the following part
+    post_call_url_res = requests.post(
+        post_call_url.format(country=country),
+        data=data,
+        headers=headers,
+    )
+    if post_call_url_res.status_code != 200:
+        print(f"ERROR: when updating product {product['code']}. {post_call_url_res.status_code} status code")
+        sys.exit()
+and run the code, to be sure that each <tag_plural>_lc of all products 
+that should be updated are having a corresponding country in the dictionary 
+mapping_languages_countries. If not, it should print similar message:
+    "ERROR: when updating product 8850718818921. Unknown country for this language: th"
+"""
+
+import requests
+import os
+import re
+import sys
+
+# define parameters
+map_tags_field_url_parameter = {
+    # tag field in API, file name in taxonomies: url parameter
+    # "allergens": "allergen",
+    # "brands": "brand", # all unknown for now (no taxonomy for it)
+    "categories": "category",
+    # "countries": "country",
+    # "labels": "label",
+    # "origins": "origin", 
+}
+
+dev = True
+if dev:
+    env = "net"
+    user = "off:off@" 
+else:
+    env = "org"
+    user = "" 
+
+headers = {
+    'Accept': 'application/json', 
+    'User-Agent': 'UpdateTagsLanguages',
+}
+data = {
+    'user_id': '',
+    'password': '',
+}
+
+logs_file = "update_tags_per_languages_logs"
+if os.path.isfile(logs_file):
+    os.remove(logs_file)
+
+mapping_languages_countries = {
+    "aa": "dj",
+    "ar": "world",  # ar but categories are en:<french name> 
+    "be": "by",
+    "bg": "bg",
+    "br": "fr",
+    "bs": "ba",
+    "ca": "fr",
+    "cs": "cz",
+    "da": "dk",
+    "de": "de",
+    "el": "gr",
+    "en": "world",
+    "xx": "world", # xx but categories are en:<french name>
+    "es": "es",
+    "et": "ee",
+    "fa": "ir",
+    "fi": "fi",
+    "fr": "fr",
+    "hr": "hr",
+    "id": "id",
+    "is": "is",
+    "it": "it",
+    "ja": "jp",
+    "lt": "lt",
+    "ms": "my",
+    "nb": "no",
+    "nl": "nl",
+    "pl": "pl",
+    "pt": "pt",
+    "ro": "ro",
+    "ru": "ru",
+    "sk": "sk",
+    "sl": "si",
+    "sr": "rs",
+    "sv": "se",
+    "th": "th",
+    "zh": "cn",
+}
+
+def get_from_api(get_call_url: str) -> dict:
+    """Send a GET request to the given URL.
+    
+    :param get_call_url: the URL to send the request to
+
+    :return: the API response
+    stop the code if returned status_code is different than 200
+    """
+    get_call_url_res = requests.get(
+        get_call_url,
+        headers=headers,
+    )
+
+    if get_call_url_res.status_code != 200:
+        print(f"ERROR: when calling api. {get_call_url_res.status_code} status code. url: {get_call_url}")
+        sys.exit()
+    
+    return get_call_url_res.json()
+
+
+def unknown_tags_taxonomy_comparison(all_tags: dict, file: str, tag_type: str) -> dict:
+    """Iterate over all referenced tags, filter unknowns tags, and search in taxonomy
+
+    :param all_tags: all_tags as returned by API call
+    :param file: path of the corresponding taxonomy file
+    :param tag_type: the current tag type (category, allergens, etc.). Used to save output file
+
+    :return: possible_wrong_language_tags {"old_tag": "tag_found_in_different_lang"}
+    """
+    possible_new_tags = []
+    possible_wrong_language_tags = {}
+    already_referenced_tags = []
+
+    for tag in all_tags["tags"]:
+        if tag["known"] == 0:
+            # limit number of iterations
+            if dev and len(possible_wrong_language_tags) > 0:
+                break
+
+            tag_name = tag['name']
+            # should retrieve all "en:blablabla, tag_name" or "it:tag_name"
+            tag_regex = re.compile(f'\n([a-z][a-z]:(?:[\w\s\-\']*\,-)*{tag_name})[,|\n]')
+
+            with open(file, "r") as f:
+
+                tag_regex_res = tag_regex.findall(f.read().lower().replace(" ", "-", -1))
+                print(tag_regex_res)
+
+                # found more than a single occurence in the taxonomy
+                # if exists, take value that correspond to "en" (i.e., unknown but 
+                #   already referenced in the taxonomy)
+                # otherwise (i.e., only different languages than "en"), keep first occurence
+                if len(tag_regex_res) > 1:
+                    # in the case that "en" is not in the list
+                    tag_regex_res_first = tag_regex_res[0]
+                    
+                    tag_regex_res = [x for x in tag_regex_res if "en:" in x]
+
+                    # "en" was not in the last put back first value in the list
+                    if not tag_regex_res:
+                        tag_regex_res.append(tag_regex_res_first)
+
+                # got one occurence in the taxonomy
+                if len(tag_regex_res) == 1:
+                    # world is in "en", hence if the tag is found in the taxonomy for "en" line,
+                    # it means that the tag is already referenced in the taxonomy
+                    if tag_regex_res[0][:2] == "en":
+                        already_referenced_tags.append(tag['id'])
+                    else:
+                        possible_wrong_language_tags[tag['id']] = tag_regex_res[0].split(',')[0]
+
+
+                # 0 occurences
+                else:
+                    possible_new_tags.append(tag['id'])
+
+
+    # print (for dev only) and save results of possible new tags in a file
+    if dev:
+        print(f"> Possible new tags for {tag_type}: <")
+        for possible_new_tag in possible_new_tags:
+            print("  ", possible_new_tag)
+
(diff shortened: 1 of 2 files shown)


scripts/update_tags_per_languages.py
```diff
@@ -0,0 +1,365 @@
...
+        
+        # by default the query return 24 results. 
+        # increase to 1000 (so far chips in EN (should be crisps in EN)
+        #   had max number of products for categories with ~550)
+        products_list_for_tag_url = f"https://{user}world.openfoodfacts.{env}/{singular}/{{tag_id_placeholder}}.json?page_size=1000"
+        
+        file = os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', f'taxonomies/{plural}.txt')) 
+        
+        # country is needed otherwise <plural>_lc will be "en"
+        post_call_url = f"https://{user}{{country}}.openfoodfacts.{env}/cgi/product_jqm2.pl"
```
alexgarel, line 297: (not to change) Just in case you care typically this kind of globals would better fits being attributes of a class, where you then call a method to process one tag.  

class ProcessTag:

   def __init__(self, plural, singular):
       self.tags_list_url = …
       ...
   def process(self):
       all_tags = self.get_from_api()
       etc.

scripts/update_tags_per_languages.py
```diff
@@ -0,0 +1,365 @@
...
+        if tag["known"] == 0:
+            # limit number of iterations
+            if dev and len(possible_wrong_language_tags) > 0:
+                break
+
+            tag_name = tag['name']
+            # should retrieve all "en:blablabla, tag_name" or "it:tag_name"
+            tag_regex = re.compile(f'\n([a-z][a-z]:(?:[\w\s\-\']*\,-)*{tag_name})[,|\n]')
+
+            with open(file, "r") as f:
```
alexgarel, line 166: I would have used the json of taxonomy (like https://static.openfoodfacts.org/data/taxonomies/categories.json) and loaded it in memory. This way you don't have to parse anything but more iterate on a map.

But I would do it once and for all outside the for loop.

scripts/update_tags_per_languages.py
```diff
@@ -0,0 +1,365 @@
...
+
+    for tag in all_tags["tags"]:
+        if tag["known"] == 0:
+            # limit number of iterations
+            if dev and len(possible_wrong_language_tags) > 0:
+                break
+
+            tag_name = tag['name']
+            # should retrieve all "en:blablabla, tag_name" or "it:tag_name"
+            tag_regex = re.compile(f'\n([a-z][a-z]:(?:[\w\s\-\']*\,-)*{tag_name})[,|\n]')
```
alexgarel, line 164: I'm not sure why we include a comma in it ? Why don't we stop on commas ?
Is this to ensure, tag_name appear in a synonym list ?

scripts/update_tags_per_languages.py
```diff
@@ -0,0 +1,365 @@
...
+
+            with open(file, "r") as f:
+
+                tag_regex_res = tag_regex.findall(f.read().lower().replace(" ", "-", -1))
```
alexgarel, line 168: Performace: at this line you load all the file in memory, which is ok. But why not load it once and for all outside the for loop ? you could even use mmap for that: https://docs.python.org/3/library/mmap.html

scripts/update_tags_per_languages.py
```diff
@@ -0,0 +1,365 @@
...
+            #      ],
+            #  }
+
+            if dev:
+                i = 0
+            for product in all_products_for_tag["products"]:
+                if dev:
+                    if i > 0:
+                        break
+                    i += 1
```
alexgarel, line 332: It's ok to always count, and it's easier to read :-) 
```suggestion
            for i, product in enumerate(all_products_for_tag["products"]):
                if dev and i > 0:
                    break
```

scripts/update_tags_per_languages.py
```diff
@@ -0,0 +1,365 @@
...
+        print(f"> Tags to update for {tag_type} (current => new): <")
+        for current_tag, updated_tag in possible_wrong_language_tags.items():
+            print(f"  {current_tag} => {updated_tag}")
+
+    # only save possible new tags to be reviewed and added
+    with open("update_tags_per_languages_possible_new_tags_" + tag_type, "w") as f:
+        f.write("possible_new_tags")
+    with open("update_tags_per_languages_possible_new_tags_" + tag_type, "a") as f:
+        for possible_new_tag in possible_new_tags:
+            f.write("\n" + possible_new_tag)
```
alexgarel, line 219: The search can be long, so I would save all the data you found out in a json file (for each map).

This way you could add an option to only load those json file and process them (if they exists), avoiding rerunning the search phase in case the update phase crashes.

scripts/update_tags_per_languages.py
```diff
@@ -0,0 +1,365 @@
...
+                        post_call_url.format(country=country),
+                        data=data,
+                        headers=headers,
+                    )
+                    if post_call_url_res.status_code != 200:
+                        print(f"ERROR: when updating product {product['code']}. {post_call_url_res.status_code} status code")
+                        sys.exit()
+
+                with open(logs_file, "a") as f:
+                    f.write(f"{product['code']},{plural};{current_tag};{updated_tag}\n")
```
alexgarel, line 361: nit: you can also keep the file open, and use `f.flush()` to ensure you writes are immediately written on disk.

I find it great that you write what is processed. You could check for this file to avoid re-processing things already done in case the script failed at some point.

scripts/update_tags_per_languages_tests.py
```diff
@@ -0,0 +1,63 @@
...
+        all_tags_dict = {
+            "tags": [
+                {"id": "en:snacks", "known": 1, "name": "snacks"}, # known
+                {"id": "en:groceries", "known": 0, "name": "groceries"}, # possible_new_tags
+                {"id": "en:cured-hams", "known": 0, "name": "cured-hams"}, # already_referenced_tags
+                {"id": "en:chips", "known": 0, "name": "chips"}, # possible_wrong_language_tags
+            ],
+        }
+        file = os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', f'taxonomies/categories.txt')) 
+        possible_wrong_language_tags = unknown_tags_taxonomy_comparison(all_tags_dict, file, "categories")
```
alexgarel, line 45: great to have tested this function

scripts/update_tags_per_languages.py
```diff
@@ -0,0 +1,365 @@
...
+    'User-Agent': 'UpdateTagsLanguages',
+}
+data = {
+    'user_id': '',
+    'password': '',
+}
+
+logs_file = "update_tags_per_languages_logs"
+if os.path.isfile(logs_file):
+    os.remove(logs_file)
```
alexgarel, line 81: maybe not if you want to continue an old failed script.

alexgarel (approved):
@benbenben2 that's a really great job !

I have no blocker, so I approve.

But one think I would advise to do is to be able to be able to recover in case of incident:
That is:
- after the search phase, store results in json
- eventually skip search phase if I have json results and jump to update
- use the already updated products information you log to skip already done updates.

scripts/update_tags_per_languages.py
```diff
@@ -0,0 +1,365 @@
...
+
+mapping_languages_countries = {
```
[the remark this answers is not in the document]
benbenben2: thanks

left in the hashmap would be language_code in your url
right in the hashmap would be country_code_2 in your url

That would work for Denmark, ok
Not sure for Canada (language_code = "en,fr").
That would not work for "ar", "en", "xx"

I dont know if I could use your url. But it could be good if I could use it, otherwise if we find a language not listed in this hashmap while running the script my script would crash.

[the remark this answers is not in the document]
benbenben2: It is not ideal. 
"aa", "br" (breton, french regional lang) would be missing, "xx" too -> would need to add them on top of the extraction from the url...
"ar" is special case in current dict
there are some languages (en, fr, es, etc.) that are officials in more than one country. The mapping can results in strange mapping like fr -> mg, en -> ca, es -> ni, de -> ch, etc. Not sure that it would be problematic, but would need to test it.
"en" should be replaced from whatever value to "world"


I am afraid that it would result in more lines of code. The actual code raise an error in the log. I do believe that it would be enough for a first version of the script.

scripts/update_tags_per_languages.py
benbenben2: f'\n([a-z][a-z]:(?:[\w\s\-\']*\,-)*{tag_name})[,|\n]'

There are 2 commas in the regex.

if you have:

> en:Hamburger buns, Burger buns, **Rolls for hamburger**, Buns for hamburger

and you are searching for "Rolls for hamburger"
the first comma in the regex would ignore the comma at the end of

> en:Hamburger buns, Burger buns, 

the second comma in the regex would ignore the comma right after the occurence:

> , Buns for hamburger

scripts/update_tags_per_languages.py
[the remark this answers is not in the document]
alexgarel: ok, that was just a suggestion ! No problem.

scripts/update_tags_per_languages.py
alexgarel: Ok, I see, I'll try to comment, feel free to change:

```suggestion
            # should retrieve all "en:blablabla, tag_name" or "it:tag_name"
            # the prefix is either the language or a comma. 
            # Suffix is either an end of line or a comma
            tag_regex = re.compile(f'\n([a-z][a-z]:(?:[\w\s\-\']*\,-)*{tag_name})[,|\n]')
```

Do you know you can also use [positive lookahead or lookbehing](https://docs.python.org/3/library/re.html#) (`(?=...)` and `(?<=...)`)

scripts/update_tags_per_languages.py
benbenben2: Let's stick to simple open() and not include mmap(), since we don't want to apply changes to the file content, but just read and search in it.

scripts/update_tags_per_languages.py
benbenben2: It does not show the synonyms in the json.
I would keep the taxonomy file locally instead.

scripts/update_tags_per_languages.py
benbenben2: interesting

Actual regex captures the whole block: `"([a-z][a-z]:(?:[\w\s\-\']*\,-)*({tag_name})"`
It can be "en:blablabla, tag_name" or "it:tag_name", for example.

In the next lines, it checks if it starts with "en:" or another language, and if not it keeps the first element of the list of synonyms (for example, if we looked for "Alga spirulina" and it found "it:Spirulina, Alga spirulina", then it will keep "it:Spirulina")

Eventually, we could put "(?P<some name>[a-z][a-z]:(?:[\w\s\-\']*\,-)*(?P<tag_name>{tag_name})", but maybe it is good as it is already now.

benbenben2:
@alexgarel 

I applied your suggestions, script now:
 - can be run with input parameters
 - can be resumed if it crashes

The code was run for a single tag in .net environment. The file update_tags_per_languages_wrong_languages_detected_categories contained only:

> current tag, <language>:found tag
> en:ядки,bg:ядки

The script was interrupted before the end of the update and run again. It resumed the updates.

Before
Screenshot_20240217_145520

After
Screenshot_20240217_171819

Note that it did not add new tag, because the tag was already listed in English:

> <en:Nuts and their products
> en:Nuts, Fruit nuts
> bg:Ядки

scripts/update_tags_per_languages.py
```diff
@@ -282,54 +312,141 @@ def update_tags_field(tags_field_string: str, tags_field_lc: str, current_tag: s
...
+                pass
+
+        # retrieve result of the previous step
+        possible_wrong_language_tags = {}
+        with open(log_file_name_1.format(plural=plural), 'r') as log_file_1:
+            # skip header
+            log_file_1.readline()
+            for line in log_file_1:
+                key, value = line.strip().split(',')
+                possible_wrong_language_tags[key] = value
```
alexgarel, line 389: nit: fashioned pythonista ;-)

```suggestion
            possible_wrong_language_tags = dict(
                line.strip().split(',', 1) for line in log_file_1
            )
```

alexgarel (approved):
Great job @benbenben2  !

PR #9588, opened 2023-12-23
feat: allow_longer_user_name

### LINKED ISSUE #9558: Allow username longer than 20 characters
image

20 characters is not that long. Many brands, for example, are longer than 20 chars.

Sometimes it's important to have a clear username. For a bot, I can't use "off-delete-nutrients-bot" for example.

30 or 35 would be ok I would say.

benbenben2:
### What
allow users to have user_name length up to 40 characters
updated all files in po, but not sure how it is working, if it will be overwritten by crowdin or something like that...

### Screenshot
Screenshot_20231223_112757_2

- Fixes #9558

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -662,7 +662,7 @@ msgid "The user name must contain only unaccented letters, digits and dashes."
 msgstr ""
 
 msgctxt "error_username_too_long"
-msgid "The user name is too long (maximum 20 characters)."
+msgid "The user name is too long (maximum 40 characters)."
 msgstr ""
 
 msgctxt "error_name_too_long"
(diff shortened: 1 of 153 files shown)


po/common/common.pot
```diff
@@ -662,7 +662,7 @@ msgid "The user name must contain only unaccented letters, digits and dashes."
 msgstr ""
 
 msgctxt "error_username_too_long"
-msgid "The user name is too long (maximum 20 characters)."
+msgid "The user name is too long (maximum 40 characters)."
```
alexgarel, line 665: FYI @benbenben2, you don't have to change this in every file. Changing it in common.pot it enough (crowdin handles the rest).

alexgarel (approved):
LGTM thanks.

PR #9589, opened 2023-12-23
taxonomy: soy_sauce_vegan

benbenben2:
### What
added vegan and vegetarian tag for soy sauce.

### Screenshot
Screenshot_20231223_115932

- Fixes #-slack-


aleene:
I would not add vegan tags to compound ingredients, as it contains to uncertainties. So it is maybe. It all depends on ingredients. In fact it might contain fish extracts.

benbenben2:
Updated to be maybe as it is a compound ingredient.
+ added filter to skip compound ingredients (any ingredients having "ingredients = [etc.]"

See discussion on Slack

PR #9605, opened 2023-12-30
taxonomy: parsing plant proteins in French and German

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -5417,6 +5417,11 @@ my %ingredients_categories_and_types = (
 				"thym",
 			]
 		},
+  		# plant protein
+		{
+  			categories => ["protéines végétales",],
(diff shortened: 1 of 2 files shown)


benbenben2:
"Simple plural (just an additional "s" at the end) will be added in the regexp."
"$category_regexp .= '|' . $category . '|' . $category . 's';"
That would not word for "protéine végétale" because it will be "protéine végétales" and not "protéines végétales"
Hence, let's keep "protéines végétales" as it is.






If you have categories: "protéines végétales" and types "avoine" , then it will search for "protéines végétales d'avoine". This is not in the taxonomy.
But instead "protéine d'avoine, protéines d'avoine" is in the taxonomy. Hence, you can use the alternate_names features that Stephane implemented the other day.

Same for German, it will search for "pflanzliche Proteine ackerbohnen"

Screenshot_20231231_102800
Screenshot_20231231_102817

moon-rabbitOFF:
I forgot about it, thank you! I added Eiweiß, which is a synonym of Protein

stephanegigandet (approved):
Thank you!

Issue #1414, opened 2018-10-14
Quality facet - Nutrition - Check that quantity of nutrients/ingredients match legal product claims


teolemon:
### What
- Check that quantity of sugar match legal product claims
  - if label= "A faible teneur en sucre"= en:Low sugar <5g/100g
  - if label= "Sans sucres" = en:No sugar <0,5g/100g
  - if label = "Sans sucres ajoutés" = en:No added sugar ingredients should not contain: "saccharose, lactose, glucose, fructose…" (according to Test Achats)
- Jams
  - https://informaciongastronomica.com/la-mermelada-mas-sana-del-mercado/ "EU law stipulates that jams sold must contain a minimum percentage of fruit of 35%, as a general rule."
### Part of
- #5538

hangy:
- "A faible teneur en sucre" = en:Low sugar
- "Sans sucres" = en:No sugar
- "Sans sucres ajoutés" = en:No added sugar

benbenben2:
Under Regulation (EU) 1924/2006 (EC 2006) for health and nutrition claims made on foods, the following sugars-related nutrition claims are permitted:

Sugar-free claims, 'may only be made where the product contains no more than 0,5 g of sugar per 100 g or 100 ml'.

With no added sugars claims 'may only be made where the product does not contain any added mono- or disaccharides or any other food used for its sweetening properties. If sugars are naturally present in the food, the following indication should also appear on the label: CONTAINS NATURALLY OCCURRING SUGARS'.

Low sugars claims 'may only be made where the product contains no more than 5g of sugar per 100 g for solids or 2,5 g of sugar per 100 ml for liquid'.

https://knowledge4policy.ec.europa.eu/health-promotion-knowledge-gateway/sugars-sweeteners_en

https://eur-lex.europa.eu/legal-content/en/ALL/?uri=CELEX%3A32006R1924
See annexe for all others "high fibers", "low fat" etc

benbenben2:
For jams.

The quantity of pulp and/or purée used for the manufacture of 1 000 g of finished product must not be less than:                  
350 g | as a general rule,



https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A32001L0113

aleene:
I agree. Regulation kind of things can be added to the labels taxonomy.

benbenben2:
1. Idea is to start with warning, review after few weeks and switch to error if everything is fine.
[...]

PR #9606, opened 2023-12-30
feat: dq_category_specific_ingredient_percent_2

benbenben2:
### What
Add quality facets for jams having too small fruit quantity:
errors:
- en:specific-ingredient-fruit-quantity-is-below-the-minimum-value-of-35-for-category-jams
- en:specific-ingredient-fruit-quantity-is-below-the-minimum-value-of-25-for-category-redcurrants-jams
- ["en:specific-ingredient-<$specific_ingredient_id>-quantity-is-below-the-minimum-value-of-$quantity_threshold-for-category-<$category_id>]

info:
- en:missing-specific-ingredient-for-this-category

Following comment from @CharlesNepote and @aleene, I tried (maybe not perfectly done) to write the thresholds values directly in the taxonomy. 

This is only for specific ingredients, only for lower than the value provided. If it works well, in future PR, it could be generalized for ingredients and nutriments and for maximal values (for labels it can be minimum (sugar, for example) or maximum (fibers, for example)). Also, to do eventually in the future, detect if jam/jelly should be extra-jam/extra-jelly.

### Screenshot
Screenshot_20231231_002712

Part of https://github.com/openfoodfacts/openfoodfacts-server/issues/1414

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -2391,6 +2391,62 @@ sub check_labels ($product_ref) {
 				"en:high-in-omega-3-label-claim-but-ala-or-sum-of-epa-and-dha-below-limitation");
 		}
 	}
+
+	# In EU, compare categories and regulations
+	# https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A02001L0113-20131118
+	# my %spread_categories_regulation = (
+	# 		europe => {
+	# 			"en:jams" => "35",
+
+	# some categories have mininal amount content required by regulations
+	# $expected_minimal_amount_specific_ingredients = "en:fruit, 35, en:eu"
+	my ($expected_minimal_amount_specific_ingredients, $category_id)
+		= get_inherited_property_from_categories_tags($product_ref, "expected_minimal_amount_specific_ingredients:en");
+
+	# convert as a list, in case there are more than a countries having regulations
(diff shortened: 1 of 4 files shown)


CloCkWeRX:
My only suggestion would be seeing if there's an existing way to say "yes, this fails a data quality check for category, but no; it is not because the data is wrong".

This would cater for scenarios like https://www.npr.org/2020/10/01/919189045/for-subway-a-ruling-not-so-sweet-irish-court-says-its-bread-isnt-bread or https://www.delish.com/food-news/a49216/things-you-didnt-know-about-pringles/ (Potato Chips made from dried potato vs potato crisp vs potato chip)

* Average person and AI would call it... "bread" or "potato chip" when they are scanning it
* A regulator would call it something different

Otherwise, this somewhat implies that categories can only be derived from regulations

benbenben2:
> My only suggestion would be seeing if there's an existing way to say "yes, this fails a data quality check for category, but no; it is not because the data is wrong".
> 
> This would cater for scenarios like https://www.npr.org/2020/10/01/919189045/for-subway-a-ruling-not-so-sweet-irish-court-says-its-bread-isnt-bread or https://www.delish.com/food-news/a49216/things-you-didnt-know-about-pringles/ (Potato Chips made from dried potato vs potato crisp vs potato chip)
> 
>     * Average person and AI would call it... "bread" or "potato chip" when they are scanning it
> 
>     * A regulator would call it something different
> 
> 
> Otherwise, this somewhat implies that categories can only be derived from regulations

Yes, this is a good point. We really would like to have that. It is not only for this PR, but more generic. To be able to shutdown alerts that are false positives.

PR #9607, opened 2023-12-31
feat: Generate stores taxonomy from name-suggestion-index

### LINKED ISSUE #7632: Add autosuggest for stores
### What
- Add autosuggest for stores
### Why
- For the mobile app
### Code pointers
- Source code: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/cgi/suggest.pl
- Stores taxonomy: https://static.openfoodfacts.org/data/taxonomies/stores.json
- Suggest not working for a taxonomized entry: https://world.openfoodfacts.org/cgi/suggest.pl?tagtype=stores&term=franpr&lc=fr

CloCkWeRX:
Fixes #7632 ?

### What

This adds a snapshot of the BSD 3 clause licenced name-suggestion-index records for shop/supermarket, transformed into a stores.txt taxinomy format.

There may be some mapping inconsistencies with countries.txt, as I used ISO3601 2 letter codes and added 1 or 2 minor adjustments.

While I added this as a snapshot in git, it would be trivial to fetch the live data from upstream.

Running/generating:
```
clockwerx@LAPTOP-2K4VT916:~/openfoodfacts-server/taxonomies$ node external/name-suggestion-index/shop/supermarket.js >> unused/stores.txt
```

### What should be done to make this better

- [ ] If there is a central list of languages from an i18n package or similar, that has mappings from countries, that should be used
- [ ] Use of countries.txt or similar
- [ ] Evaluate how important it is to keep stores.txt as is, vs just say, adopting the name-suggestion-index format.
- [ ] Is the wikidata syntax right?
- [ ] eslint?


teolemon:
cc @raphodn

teolemon (commented):
It looks really nice to me, I just required a review from the 2 Raphael from Open Prices, since this will potentially be really helpful for them.
What was your personal intent behind this PR ?

CloCkWeRX:
Bigger picture, I want to link a particular food item to "where can I get it".

Use cases:
- https://github.com/danslimmon/oscar Throw out a package, add to shopping list, potentially you can do an online order/click and collect
- Walk into an area with a geofence - https://github.com/alltheplaces/alltheplaces / open streetmap - have home assistant push a notification to get the three items from that Wikidata:brand store that are on your list
- I need (uncommon ingredient or food); from the places it is sold which is closest to me?

raphodn:
Also : https://github.com/openfoodfacts/open-prices

alexgarel:
@raphodn can you validate the PR ? (If it fits openprice needs)

aleene:
We should be careful here creating a taxonomy independent of OFF Prices. I would prefer to uses Prices as starting point.

aleene:
I would move this to another taxonomy: store_brands in order to distinguish from actual individual stores.

aleene:
See also https://wiki.openfoodfacts.org/Global_stores_taxonomy

aleene:
We already have store data entered by the users. We should integrate that in someway.

aleene:
I do not get where the data comes from.

CloCkWeRX:
> I do not get where the data comes from.

So: OpenStreetMap + Wikipedia have found they have some intersecting interests.

One of those is _wikidata_, so there are stable identifiers for concepts like _brand of store/chain of store_.

OpenStreetMap has a large number of contributors who survey things on the ground, but of course everyone agreeing that a Carrefour is a Carrefour in the **exact** same way is difficult.

So; to help but not replace end user judgement, a number of the editing tools made the **Name Suggestion Index**.

Example of it in use by an editor:
image

Most of the time, it's right, or an end user says it's not.


So, what does it have to do with OpenFoodFacts?

- People in general have to eat
- The openstreetmap folks who go to their local chain supermarket and observe the lat/lon are recording one facet of it
- The wikipedia/wikidata folk who like to classify things are recording another facet of it
- OpenFoodFact contributers who fill out "store" field are looking at the store brands ("I bought my peanut butter at a Carrefour"), but _sometimes are physically standing in close proxmity to a specific store when they do it_

So, we are left with:
- A comunity maintaining if a chain of stores exists or has gone bankrupt (aka namesuggestionindex)
- If a user consents, the potential to ask them if they bought product X from chain Y via a mobile app interaction
- A bunch of real world stores/brands that exist because users have both surveyed them, and felt they were common enough to put in the name suggestion index.

Does that make sense?

aleene:
Complex way of explaining. Note I am a contributor to OSM and Wikidata and taxonomy maintainer/developer for OFF. So I get all the details. I did not know this tool though, interesting.

aleene:
I am all in favour of getting this taxonomy going. Checkout the wiki link above. Integration with prices.openfoodfacts.org is key however. We should introduce new taxonomies with small steps, one usecase at the time. Indeed using OSM and wikidata will be central in this. OSM for actual shop locations with OSM identifiers and Wikidata for shop brand information.

raphael0202:
@CloCkWeRX I'm sorry that I only give some feedback on your PR now.

The real question in my opinion is, how much information about the store we bought the product in should we store on Open Food Facts? Most of the time, on OFF, the `stores` field is far from being complete and can be obsolete, with no way to know whether the information is still valid (no image to back this up).
On the contrary, Open Prices provides a way to know that a specific store sold a specific product at a specific time.

We can transfer this data from Open Food Facts to Open Prices using OFF API, but for the moment I find it more interesting to have this store information in Open Prices instead of Open Food Facts.

It will be tremendously helpful to answer questions such as :
- Is store chain A more expensive than store chain B for this product?
- I want to display all prices of chain A for this product
- Chain A only sell organic products, I want the "organic" checkbox to be toggled by default when adding a raw product

CloCkWeRX:
I suspect what you have done with open prices supercedes this a bunch, as you model the specific OSM way or node ID, from the looks of it?

I am more than happy to give it a go against a different repo, basically point me at the brands data and I'll have at it. May be a few days until I have the bandwidth

Can bash this out as python, node, whatever

raphael0202:
> I suspect what you have done with open prices supercedes this a bunch, as you model the specific OSM way or node ID, from the looks of it?

It's not really a superseed of it, as we currently have no way of knowing whether a specific store belongs to a bigger chain.

Basically, on Open Prices, I would like to have a new table that could be named something like `store_brands`, where a specific store in the `stores` table could be linked to this item.
This would open up the new features on Open Prices described above (and also help you on your use cases).

I will open a new discussion on Open Prices repo to discuss the matter.

edit: Done here: https://github.com/openfoodfacts/open-prices/discussions/863

taxonomies/unused/stores.txt
raphodn: sorry, seeing this PR only now, this is really cool.
one thing we mentioned recently with @teolemon & @aleene is to transform the taxonomy to `xx:`
doable ?

taxonomies/external/name-suggestion-index/shop/supermarket.json
raphodn, line 1: I wouldn't restrict to just shop/supermarket.json
- OFF: `shop/convenience`, `shop/alcohol`, `shop/wine`, `shop/bakery`...
- OBF: `shop/beauty`
- OPF: `shop/doityourself`, `shop/toys`, `shop/sports`, `shop/books`...
- OPFF: `shop/pet`

raphodn: ==> take all the `shop/*`

taxonomies/external/name-suggestion-index/shop/supermarket.js
raphodn, line 1: I'm a big fan of the idea of leveraging NSI 👍 
fyi I'm regularly contributing to the project, see https://github.com/openfoodfacts/open-prices/issues/467#issuecomment-2437176771

raphodn: Maybe store a `nsi_id: carrefour-c622af` to know that it comes from NSI ? 
- to easily find it back
- and differentiate in the taxonomy shops coming from NSI from shops coming from OFF contributors ?
- thus knowing which shops could be added back to NSI

taxonomies/external/name-suggestion-index/shop/supermarket.json
CloCkWeRX: So, can do that, would the preference be 1x script per topic; or all in one?

taxonomies/external/name-suggestion-index/shop/supermarket.js
CloCkWeRX: Happy to split these out into an stores-nsi.txt or something of that nature to more clearly split it; I feel like at the time I did this manually.

Script likely needs to be re-run since it's been ~2 years.

aleene:
Let us go ahead with this and readjust later.

taxonomies/external/name-suggestion-index/shop/supermarket.js
raphodn: I can have a look later this week and try to re-run the script myself.
if we switch to `xx:` then the country mapping could be removed ?

taxonomies/external/name-suggestion-index/shop/supermarket.json
raphodn: all in one ? 😁 
with a `nsi_kv: shop/convenience` ?

Freso: Definitely all in one. It would lead to a lot of duplication (or complication) if store-brands had to be repeated for each OxF type they were applicable for (e.g., most supermarkets/grocery or general stores/etc. would fit in all of them).

taxonomies/external/name-suggestion-index/shop/supermarket.js
Freso: >  if we switch to `xx:` then the country mapping could be removed ?

I think country mapping could still be useful, esp. for cases where store-brands may have identical or near-identical names in different regions (see e.g. the 50 or so “COOP”s). (Mappings like `country:en: en:United States of America,en:United States of America,en:United States of America,en:United States of America,en:United States of America` are probably a bit excessive though…). Maybe this is too much “preemptive optimisation(/complication)” though, and could be easily solved with a human-added `comment:en:` key or similar where it’s actually needed.

Definitely +1 to using `xx:` similar to our `brands.txt` taxonomy though, with possible locale specific overrides for locales with scripts other than the first/primary `xx:` alias where relevant.

teolemon:
@raphodn @CloCkWeRX @aleene @Freso @raphael0202 Should I do as suggested by @aleene , approving and merging ? Otherwise, can someone else do so ?

Freso:
I’m fine with whatever!

teolemon (approved):
Approving this one since it's peripheral and useful

taxonomies/unused/stores.txt
```diff
@@ -787,4 +794,3501 @@ en: Sam’s Club
...
+
+cn:永辉超市
+wikidata:en:Q17028520
+country:en: en:China
+
+cn:沃尔玛
+wikidata:en:Q483551
+country:en: en:China
+
+117.1140042,36.6507007,50:济南华联超市
```
teolemon, line 4248: @CloCkWeRX issue here

teolemon (commented):
some issues to fix

taxonomies/external/name-suggestion-index/shop/supermarket.js
```diff
@@ -0,0 +1,298 @@
...
+  "yt": "Mayotte",
+  "yu": "Yugoslavia",
+  "za": "South Africa",
+  "zm": "Zambia",
+  "zw": "Zimbabwe",
+  "fx": "France (Metropolitan)"
+}
+
+// Overrides for the language codes/mappings inferred by country code.
+// TODO: This list is far, far from accurate or complete and should be replaced with any internal/authoritative mappings
```
Freso, line 250: Would it be possible to use the language property from the `countries` taxonomy?

taxonomies/external/name-suggestion-index/shop/supermarket.js
```diff
@@ -0,0 +1,298 @@
+// A nodejs script to parse NameSuggestionIndex formatted JSON data and attempt to output valid taxonomy data.
+// Do not use results without careful review
+const data = require('./supermarket.json');
+
+// A quick ISO 3601 to human labels (english) mapping.
+// TODO: Ensure this matches exactly with countries.txt
+// TODO: Replace with a better lookup
```
Freso, line 7: Could this utilise the `countries` taxonomy?

PR #9609, opened 2023-12-31
build: Migrate to Compose V2

hangy:
Docker Compose V1 (`docker-compose`)  is deprecated and should not be used any more. ([Source](https://docs.docker.com/compose/migrate/)) This replaces the old command with `docker compose`, which is Compose V2, and bundled with current Docker distributions.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -144,6 +144,6 @@ jobs:
         make dev
         make status
     - name: Test all is running
-      run: make livecheck || ( tail -n 300 logs/apache2/*error*log; docker-compose logs; false )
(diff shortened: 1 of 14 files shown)


alexgarel (approved):
LGTM ! Do you think we need to change version number ? is it big enough a change ?

Thanks @hangy :-)

hangy:
> Do you think we need to change version number ? is it big enough a change ?

I don't know. Right now, this is mostly used for development, and the dev server, right? In some way, it could be regarded as a breaking change, because it won't work with old versions of Docker any more.

alexgarel (changes requested):
@hangy your PR is fine, but I block it until I have install docker compose plugin on the staging environment !

alexgarel (approved):
staging is ready: `apt install docker-ce  docker-compose-plugin`

PR #9615, opened 2024-01-03
chore: adding beverage preparations

aleene:
### What
Introduction of a new super category, which helps to distinguish between beverages (ready to drink) and products that form the basis of beverages (after adding a fluid (water/milk/etc) to dissolve, dilute, infuse, etc).

Additional subcategories could be added if necessary.

After the categories are available some products need clean up (esp teas).


taxonomies/categories.txt
```diff
@@ -1031,6 +1031,14 @@ fr:Préparations à base de miel et gelée royale, mélanges de miel et gelée r
 it:Miscele a base di miele e pappa reale, Mix di miele e pappa reale
 nl:Mengsels van honing en koninginnegelei, Mengelingen van koninginnegelei
 
+en:Beverage preparations
+es:Preparaciones de bebidas
+de:Getränkezubereitungen
+fr:Préparations pour boissons
+it:Preparati per bevande
+nl:Drankbereidingen
+description:en:Any product which purpose it is to create a beverage which uses water (hor or cold) in some way (diluting, infusing, dissolving, etc).
```
alexgarel, line 1040: I see it's not in Beverage category, But can't it change Nutri-Score computation ?

See is_beverage_for_nutrition_score_2023 and is_beverage_for_nutrition_score_2021 in Food.pm, we rely on the category.

cc: @stephanegigandet

aleene: The Nutriscore was under the subcategories.

stephanegigandet: @aleene: I think it makes sense to put "beverage preparations" under "beverages". Tea bags etc. have nutrition facts for the prepared beverage, and are considered beverages for the Nutri-Score. It's also where users are likely to look for them.

aleene: Can do, but then we miss out on the category `ready to drink beverages`.

aleene: Could we add a category under beverages to indicate ready to drink beverages?

stephanegigandet: Well do we really need one? It's a bit strange to add a category that should be applied to 99% of beverages.

Or maybe another possibility could be to have a parent "Beverages and beverages preparations"?

aleene: Thats why I wanted to set the preparations apart, but we can create a super as you suggested.

stephanegigandet (approved):
Thank you @aleene !

PR #9618, opened 2024-01-04
fix: Product name truncated #7502

### LINKED ISSUE #7502: NEW DESIGN - Product name is truncated (JS disabled only ?)
### What

- [ ] Product name is truncated (JS disabled only ?)
### Screenshot
image
### Part of
- #7295

TCatinaud:
### What

The product name is truncated due to a fixed height defined in the product link. This allows you to have a list of products with a unified height.   

Several solutions are possible (such as using grids), I preferred:
- keep a fixed height by adapting it to avoid seeing part of the text truncated.   
- use the [`-webkit-line-clamp` property](https://developer.mozilla.org/fr/docs/Web/CSS/-webkit-line-clamp) to display 3 points if the text is longer. The latter is still in [WD](https://drafts.csswg.org/css-overflow-4/#line-clamp) but supported by [97% of browsers](https://caniuse.com/?search=-webkit-line-clamp) and offers gracefull degradation if not supported.  

The full product name remains accessible by screen readers and when hovering over the map (via the title attribute of the link).  
I also added the line-clamp when JS is activated with the alternative in the link title.

### Screenshot
- JS disabled, before fix: 
7502-js-disabled-ko
- JS disabled, after fix: 
7502-js-disabled-ok
- JS disabled, show title:
7502-js-disabled-ok-title
- JS enabled, before fix:  
7502-js-enabled-ko
- JS enabled, after fix:  
7502-js-enabled-ok
- JS enabled, show title: 
7502-js-enabled-ok-title
  

- Fixes #7502

### DIFF
--- a/scss/_product-list.scss
+++ b/scss/_product-list.scss
@@ -126,10 +126,12 @@ $productsGutter:12;
     padding-left: 8px;
     font-size: 14px;
     color: $black;
-    display: inline-flex;
-    // Force the height of the product name so that icons
-    // below are vertically aligned with other cards on the same line
-    height:82px;
+    // Force the height and use line-clamp of the product name so that
+    // icons below are vertically aligned with other cards on the same line
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
+    -webkit-line-clamp: 4;
+    height: 82px;
     overflow: hidden;
 }
 
@@ -164,11 +166,11 @@ $productsGutter:12;
   .f-dropdown {
     a { color: $primary-color; }
   }
-  &__results { 
+  &__results {
     display: flex;
     align-items: center;
     gap: 8px;
-    color: $white; 
+    color: $white;
     .material-icons { font-size: 18px; }
   }
   &__buttons {
@@ -287,7 +289,7 @@ $greeter-contribute-bgcolor: #FFF2DF;
       background: linear-gradient(90deg, $greeter-discover-bgcolor 50%, $greeter-contribute-bgcolor 50%);
     }
   }
-  &__discover, 
+  &__discover,
   &__contribute {
     display: flex;
     margin: 0;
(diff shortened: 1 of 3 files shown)


scss/_product-list.scss
```diff
@@ -126,10 +126,12 @@ $productsGutter:12;
...
     font-size: 14px;
     color: $black;
-    display: inline-flex;
-    // Force the height of the product name so that icons
-    // below are vertically aligned with other cards on the same line
-    height:82px;
+    // Force the height and use line-clamp of the product name so that
+    // icons below are vertically aligned with other cards on the same line
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
```
hangy, line 132: This seems to be related to the (old?) standard `display: box`, but `box-orient` is apparently deprecated according to https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient#browser_compatibility? Or is that related to the W3C draft?

alexgarel:
FYI, on firefox 121.0 on ubuntu 23.10:

* works well without js for me

* with js, the 4th line is already below the other tile:
  image

stephanegigandet:
> * with js, the 4th line is already below the other tile:

@alexgarel in fact it depends on the screen width, there is a .v-space-tiny class that gets from 3.5 pixels to 10.5 pixels as the screen grows, but it has no sense here as we put in more columns when the screen grows.. I'll try replacing it with a fixed size instead.

stephanegigandet:
I removed the v-space-tiny class: it made the padding-top of the text go from 3.5 pixels to 10.5 pixels as the screen grew, resulting on the 4th line being not shown completely when the screen was large.

The 4 lines now always fit:

image

stephanegigandet (approved):
Thank you so much for this change @TCatinaud , it's perfect. :)

PR #9135, opened 2023-10-11
fix: arrow is not centered in Manage image button Fixes #7489

TCatinaud:
### What

 The arrow is not centered in Manage image button

### Related issue(s) and discussion
- Fixes #7489

PR #9619, opened 2024-01-04
fix: Removed Extra Character in Delete Images Button

### LINKED ISSUE #9277: Invalid character in the Delete the images button
### What
image

MonalikaPatnaik:
Fixes #9277
Removed the extra character in the "Delete Images" button.
image

### DIFF
--- a/scss/_product-page.scss
+++ b/scss/_product-page.scss
@@ -135,7 +135,7 @@ $decalTop:48px;
       pointer-events:none;
     }
   }
-  &.active a::after {
+  &.active a {
     content:"\e5c7";
   }
 }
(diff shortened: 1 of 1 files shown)


MonalikaPatnaik:
> Thanks @MonalikaPatnaik ! One issue is that the "Manage image" dropdown now always has its arrow going down, even if it is open:
> 
> image
> 
> Could you take a look?

 This is a follow-up fix for a PR #9135 that addresses the "dropdown icon" alignment to center but there is no need to add a separate css for only aligning. I have incremented the value of the top property so that the icon is properly aligned with the "Manage Images" text.

image
image

stephanegigandet:
@MonalikaPatnaik Cpuld you change the top: property only for that specific dropdown? Right now it's affecting all the other dropdowns, like the ones we have on knowledge panels:

image

stephanegigandet (approved):
Thank you! :)

PR #9624, opened 2024-01-06
fix: bug_ingredients_wrongly_specific_ingredients

### LINKED ISSUE #9623: bug: ingredients wrongly parsed as specific ingredients
### What
Some ingredients are wrongly parsed to specific ingredients in PL language

Example product: https://world.openfoodfacts.org/product/4056489019091/fruit-to-go-freshona
Aslo: https://world.openfoodfacts.org/api/v2/product/4056489019091

### Expected behavior
Those ingredients are NOT specific ingredients but there are ingredients.

### Additional context
**jabłka and zagęszczony sok z aceroli** are parsed as ingredients.
whereas all others are parsed as specific ingredients.

I first thought that it is because those ingredients may not be recognized. But no, because for example "przecier z mango" is recognized as mango with processing "pureed".
It seems to be language related because if we google translate the list of ingredients in EN, it works fine: [https://world.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=84%25+[…]ted+acerola+juice.&type=add&action=process&submit=Envoyer](https://world.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=84%25+apples%2C+10%25+mango+puree%2C+3%25+concentrated+apple+juice%2C+1.5%25+concentrated+passion+fruit+juice%2C+concentrated+acerola+juice.&type=add&action=process&submit=Envoyer)

I suspect that it could be related to the function "parse_specific_ingredients_from_text" that we reworked recently with Stephane. But not sure.

benbenben2:
### What
Make sure that variables are 1) defined + 2) non equal to "", to detect specific ingredients

### Screenshot
Screenshot_20240106_124229

- Fixes #9623

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -197,7 +197,7 @@ my %may_contain_regexps = (
 	fi =>
 		"saattaa sisältää pienehköjä määriä muita|saattaa sisältää pieniä määriä muita|saattaa sisältää pienehköjä määriä|saattaa sisältää pieniä määriä|voi sisältää vähäisiä määriä|saattaa sisältää hivenen|saattaa sisältää pieniä|saattaa sisältää jäämiä|sisältää pienen määrän|jossa käsitellään myös|saattaa sisältää myös|joka käsittelee myös|jossa käsitellään|saattaa sisältää",
 	fr =>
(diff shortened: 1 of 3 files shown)


stephanegigandet (approved):
Good catch, thank you @benbenben2 !

PR #9640, opened 2024-01-09
fix: add retry to download_matomo_logs.pl

stephanegigandet:
needed to compute scan stats from the app matomo logs


alexgarel (approved):
Good :-)

PR #9644, opened 2024-01-11
chore: misc tags for ecoscore

stephanegigandet:
added some misc_tags for ecoscore

### DIFF
--- a/scripts/gen_feeds_daily_obf.sh
+++ b/scripts/gen_feeds_daily_obf.sh
@@ -7,7 +7,7 @@ export PERL5LIB="../lib:${PERL5LIB}"
 ./gen_top_tags_per_country.pl
 ./compute_missions.pl
 ./export_database.pl
-./mongodb_dump.sh /srv/obf/html openbeautyfacts 10.0.0.3 obf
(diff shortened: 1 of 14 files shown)


scripts/gen_feeds_daily_obf.sh
```diff
@@ -7,7 +7,7 @@ export PERL5LIB="../lib:${PERL5LIB}"
 ./gen_top_tags_per_country.pl
 ./compute_missions.pl
 ./export_database.pl
-./mongodb_dump.sh /srv/obf/html openbeautyfacts 10.0.0.3 obf
+./mongodb_dump.sh /srv/obf/html openbeautyfacts 10.1.0.102 obf
```
alexgarel, line 10: :-o thanks for fixing !
That said this won't fix it on obf, we have to fix on obf-main (same for others)

I will do it because I also have to fix refresh_products_tags things, so I will group that.

alexgarel (approved):
Thank you so much for gen_feeds_daily fixes.

I will report them where it is due (it would be far better  to get them from a config, that said…).

Issue #8860, opened 2023-08-16
Add last_updated_t for silent updates (in addition to existing last_modified_t for new product revisions)


stephanegigandet:
We currently have a last_modified_t field that is set when we create a new product revision.
This field is not updated when we have "silent" updates that do not create new revisions (e.g. when we re-process ingredients or packagings, recompute scores etc.).

We could create a new last_updated_t field for those silent updates.

This could in particular be very useful for updating copies of the "source" OFF data: alternate mongodb collections, postgres databases, CSV exports etc.

john-gom:
We should also push to the redis stream in this case too for subscribers like search-a-licious and off-query

CharlesNepote:
I reopen this, as #9428 did not solve the issue. @stephanegigandet was mentioning:
> This could in particular be very useful for updating copies of the "source" OFF data: alternate mongodb collections, postgres databases, CSV exports etc.

In data quality day-to-day tasks it can be important to know what have changed since some time.
This is also a need if we want to provide historized exports, such as currently tested in #7660.

`last_updated_t` should be also exported in the CSV as `last_updated_datetime`.

alexgarel:
Done in #9646

PR #9646, opened 2024-01-12
fix: create last_updated_t in addition to last_modified_t

stephanegigandet:
The last_modified_t field used to be updated only when the product input data changed (when we created a new version), but since https://github.com/openfoodfacts/openfoodfacts-server/pull/9428 it is also changed when derived data (e.g. ingredients analysis, scores etc.) are updated through update_all_products.pl

This PR changes back last_modified_t so that it is the data of the last input data change, and creates a new last_updated_t that corresponds to the last input or derived data change. See https://github.com/openfoodfacts/openfoodfacts-server/issues/8860

In addition, this PR also changes the mechanism to skip updates that do not change product data (input or derived), so that we can still record the update_key change. The update_key is very useful when we want to reprocess 3M products:  it can take days, and it's very likely that the script will be interrupted at some point, so we want to make sure that we don't have to reprocess products that have already been reprocessed even if their data did not change.


john-gom:
Hi @stephanegigandet , I'm good with this but could we initialise the new last_updated_t field to be equal to the last_modified_t for existing products so that the refresh mechanism in off-query can just look to this one field? Maybe get update_all_products to set the last_update_t to the last_modified_t if it is currently undefined and we aren't doing a genuine update (where it would be set to the current time instead)?

stephanegigandet:
> Maybe get update_all_products to set the last_update_t to the last_modified_t if it is currently undefined and we aren't doing a genuine update (where it would be set to the current time instead)?

@john-gom That sounds good. In addition we should read changes.sto to get the true last_modified_t : I ran update_all_products.pl last week without noticing that it changed last_modified_t, so the last_modified_t is incorrect for about 300k products I think.

john-gom (approved):

alexgarel:
:clap:

PR #9660, opened 2024-01-16
fix: small updates to scanbot for 2023

stephanegigandet:
Main change is that we update last_updated_t and push changed product to redis.


scripts/scanbot.pl
```diff
@@ -103,7 +103,7 @@
 
 my %codes = ();
 
-my $j = 0;    # API calls (or scans if logs have been filtered to keep only scans)
+my $scans = 0;    # API calls (or scans if logs have been filtered to keep only scans)
```
alexgarel, line 106: thanks for changing that kind of little things :-)

alexgarel (approved):
LGTM

PR #9672, opened 2024-01-18
taxonomy: fix test taxonomies

eric-nguyen-cs:
### What
Small parent id fix on the `test.txt` taxonomy


stephanegigandet (approved):
Thanks!

PR #9674, opened 2024-01-21
feat: Converted JS in Perl template into real JS

### LINKED ISSUE #9106: Convert JS in Perl template into real JS
=> https://github.com/openfoodfacts/openfoodfacts-server/blob/main/templates/web/pages/product_edit/product_edit_form_display.tt.js

As discussed with @stephanegigandet there's not much sense to have this code in a template, full of escaped chars. It should be converted into raw JS, to ease development.

MonalikaPatnaik:
Fixes #9106 
Converted the code in [product_edit_form.display.tt.js](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/templates/web/pages/product_edit/product_edit_form_display.tt.js) and [user_form.tt.js](https://github.com/openfoodfacts/openfoodfacts-server/blob/9db5cd4c616b457681cea8b9567fa24a9459fa98/templates/web/pages/user_form/user_form.tt.js)  into real javacsript.
Let me know if any further changes are required.

### DIFF
--- a/cgi/generate_sample_import_file.pl
+++ b/cgi/generate_sample_import_file.pl
@@ -48,7 +48,7 @@ my $request_ref = ProductOpener::Display::init_request();
 my $r = Apache2::RequestUtil->request();
 
 $r->headers_out->set("Content-type" => "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet");
-$r->headers_out->set("Content-disposition" => "attachment;filename=openfoodfacts_import.xlsx");
+$r->headers_out->set("Content-disposition" => "attachment;filename=openfoodfacts_import_template_$lc.xlsx");
 
 print "Content-Type: application/vnd.openxmlformats-officedocument.spreadsheetml.sheet\r\n\r\n";
 


stephanegigandet (changes requested):
Hi @MonalikaPatnaik , thank you!
I think the main goal of #9106 is that instead of having the template engine generate JS from the templates dynamically at each page view, we move the scripts to /html/js/ as static JS files.
For that we need to remove all the Template::Toolkit variables and logic (like you did for "[% FOREACH nutrition IN nutrition_products %]").
If there are not more of them, then we can make the files static JS files instead of templates.

MonalikaPatnaik:
Hi @stephanegigandet, could you please test my branch now? I have moved the template js files into html/js as static js files and modified the related code accordingly.

html/js/product_edit_form_display.js
```diff
@@ -0,0 +1,188 @@
+/*global lang */
+/* eslint-disable no-undef */
+const nutritionProducts = document.querySelectorAll(".nutrition_product");
```
alexgarel, line 3: This does not return anything for me.

Did you forgot to add a modification to `templates/web/pages/product_edit/product_edit_form_display.tt.html` ? (to add this element and related data ?).
It's a good idea to add an element with specific properties for each nutrition type, but you have to put it in product_edit_form_display.tt.html

alexgarel (changes requested):
@MonalikaPatnaik it's not working for me. 

See my comment.

html/js/product_edit_form_display.js
MonalikaPatnaik: Oh yes @alexgarel ..I somehow missed to add the script to `product_edit_form_display.tt.html`. Now it should work fine!

templates/web/pages/product_edit/product_edit_form_display.tt.html
```diff
@@ -390,4 +390,5 @@ <h1>[% title %]</h1>
     opacity: 0.7;
 }
 </style>
+<script type="text/javascript" src="/js/product_edit_form.js"></script>
```
stephanegigandet, line 393: Hi @MonalikaPatnaik , there's a name mismatch for the script, the name of the file checked in is product_edit_form_display.js and not product_edit_form.js

Also gulpfile.ts generates some "dist" version of the JS scripts in the /js/dist/ directory, could you add your new scripts to gulpfile.ts and then load them from /js/dist ?

gulpfile.ts currently contains this:

```
const jsSrc = [
  "./html/js/display*.js",
  "./html/js/product-multilingual.js",
  "./html/js/search.js",
  "./html/js/hc-sticky.js",
  "./html/js/stikelem.js",
  "./html/js/scrollNav.js",
];

```

PR #9683, opened 2024-01-22
feat: Add short name, value and evaluation to knowledge panels

### LINKED ISSUE #9368: New knowledge panels properties: half size panel + very short title and description
On mobile (app + web) we want to show some panels in a more condensed way, 2 by 2:

image

We need to define new properties for specifying that the panel should be shown half size on mobile, and add very short titles and descriptions.

### Part of
- #8919

john-gom:
### What

Create simple, short details for each panel to allow more compact visualisation, e.g. he new two-column approach in the mobile app.

- Fixes #9368


john-gom:
Hi @stephanegigandet . Creating a draft pull request so you can review the proposed changes to the knowledge panel schema

docs/api/ref/schemas/knowledge_panels/panel.yaml
```diff
@@ -13,6 +13,20 @@ properties:
   expand_for:
     type: string
     description: 'If set to "large", the content of the panel should be expanded on large screens, but it should still be possible to unexpand it.'
+  name:
```
stephanegigandet, line 16: for name and description, I think it would fit best in the title element where we already have a title / sub_title:
https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/api/ref/schemas/knowledge_panels/elements/title_element.yaml

Currently we have:

```
title_element: {
icon_size: "small",
title_element: {
icon_size: "small",
icon_url: "https://static.openfoodfacts.org/images/misc/high.svg",
title: "Saturated fat in high quantity (10.6%)"
},icon_url: "https://static.openfoodfacts.org/images/misc/high.svg",
title: "Saturated fat in high quantity (10.6%)"
},
```

We could add "name", "value" in there, as they would be displayed as the clickable knowledge panel title.

We could also add "value_label" or similar so that we can display the "Additives to avoid" in the mockup:

image

john-gom: Aren't the "Additives to avoid" just names too? e.g. we could equally well display it as:

Additives to avoid >
 (red circle) 0

teolemon:
@john-gom @stephanegigandet any blockers to merging? 

cc @g123k

teolemon:
@john-gom @stephanegigandet what's the command to update test results ? That might solve the conflict

teolemon:
@g123k FYI, the PR for the square knowledge panel is actually there

stephanegigandet:
I'm adding a property "half_width_on_mobile": true, for nutrient levels, and updating the tests.

docs/api/ref/schemas/knowledge_panels/panel.yaml
stephanegigandet: Indeed, in terms of data, both the Fat and Additive to avoid panels are the same. There's a "field" name (fat or additive to avoid) and a value, possibly a unit (%), and an evaluation. So no need for "value_label"

teolemon (approved):

docs/api/ref/schemas/knowledge_panels/elements/title_element.yaml
alexgarel, line 1: Cool to have updated the documentation here !

alexgarel (approved):
Great !

PR #9688, opened 2024-01-24
chore: add args to make build

alexgarel:
Enable using args="--progress plain" to get build log in console in case of error


stephanegigandet (approved):
Great

Issue #9659, opened 2024-01-16
Add a short explainer of the Nutri-Score in panel


teolemon:
### Add a short explainer
![image](https://github.com/openfoodfacts/openfoodfacts-server/assets/1689815/32667dd6-716b-441e-bac2-2a27edbbd439)

```
Le Nutri-Score est un logo sur la qualité nutritionnelle globale des produits. Le score de A à E est calculé en fonction des nutriments et aliments à favoriser (fibres, protéines, fruits et légumes, légumes secs…) et des nutriments à limiter (calories, acides gras saturés, sucres, sel). Le score est calculé à partir des données du tableau de la déclaration nutritionnelle qui figure au dos des emballages et des données de composition (fibres, fruits et légumes, légumes secs).
L’affichage de ce logo est recommandé par les pouvoirs publics sans obligation pour les entreprises. 
```

### Add a section header for computation details
```Détails du calcul du Nutri-Score :```

deveshidwivedi:
Hi @teolemon! Do we want the explainer like the given format (red color for the description, green for the link) and position (positioned below all Nutri-score details)?

teolemon:
We have fixed this in the revamped Nutri-Score panel

PR #9689, opened 2024-01-24
feat: new Nutri-Score v2 2023 knowledge panels

stephanegigandet:
- Small refactor of nutrition traffic lights (now included in a new parent panel, not expanded by default on the app)
- Refactor of the Nutri-Score computation explanation, to use knowledge panels.
- New Nutri-Score "New formula" logos with a blue banner
- New explanations for components of the Nutri-Score
- New description of the Nutri-Score (see #9659)

Initially I wanted to use modified nutrition traffic lights as sub panels of the Nutri-Score, in order to explain the points, but the scales really don't match between the UK traffic lights and the new Nutri-Score thresholds. It's very difficult as some nutrients have a maximum of very few points (e.g. 4, 5 or 6 for positive points), while some others can be 10, 15 or 20 (for salt).

I started a table to compare the Nutri-Score thresholds of the traffic lights points if you want to take a look:
https://docs.google.com/spreadsheets/d/16wzjt5yBghmhaA1Ztrh8Rq1N1Esc8br6Wl6mzSuVD-4/edit#gid=0

So instead I use squares so that it's easy to see at a glance what contributes to a good or bad Nutri-Score, added to the design @teolemon created on mockups on Figma: https://www.figma.com/file/nFMjewFAOa8c4ahtob7CAB/Mobile-App-Design-(Quentin)?type=design&node-id=4189-13160&mode=design

image

----

image

----

image


teolemon:
Have you deployed it on .dev yet ? @stephanegigandet  
I'd like to test how it feels in the app.
One easy thing SpF would like solved is this: https://github.com/openfoodfacts/openfoodfacts-server/issues/9659

stephanegigandet:
There is now an almost finished version on https://world.openfoodfacts.dev

@teolemon Note that the new panels are only served to admins, but it seems the app is not authenticating when reading products, so it doesn't get the new Nutri-Score panel.

teolemon:
### The padding needs to be harmonized on the top and bottom
image

### Reword
```
⚠️ Avertissement : Le taux de fruits, légumes et noix n'est pas indiqué sur l'étiquette, il a été estimé en fonction de la liste des ingrédients : 0
>>
⚠️ Avertissement : Le taux de fruits, légumes et noix (0%) n'est pas indiqué sur l'étiquette, il a été estimé en fonction de la liste des ingrédients.
```

### Put in the protein subpanel ?
Les points pour les protéines ne sont pas comptés car les points négatifs sont supérieurs ou égaux à 11.

### Put in one line at the beginning ?
```
Score nutritionnel : 18 (18 - 0)

Nutri-Score : D
```
>>
```Ce produit est Nutri-Score D, avec un score nutritionnel de 18 (18 - 0).```

**Comment by teolemon** [comment]
### Turn this into a button
image

**Comment by stephanegigandet** [comment]
> The padding needs to be harmonized on the top and bottom

@teolemon I can't replicate this, which browser / settings / zoom / resolution etc. are you using?

This is what I have in Chrome and Firefox, and it works with all resolutions / zoom levels I tried.

image

**Comment by stephanegigandet** [comment]
> Turn this into a button

@teolemon Well it's a link, not a button. There's currently no concept of link  buttons in knowledge panels. We could add one, but it would have to be supported in the app as well.

**Comment by stephanegigandet** [comment]
> ### Put in the protein subpanel ?
> Les points pour les protéines ne sont pas comptés car les points négatifs sont supérieurs ou égaux à 11.

Well there's no protein subpanel if we don't count them.

>Put in one line at the beginning ?
>Score nutritionnel : 18 (18 - 0)
>Nutri-Score : D
>Ce produit est Nutri-Score D, avec un score nutritionnel de 18 (18 - 0).

I think it's better as-is, less text to read. This is much less useful than the visualization of the points for each component, so we can keep at the end. The numeric score is more for debugging, it's not directly interpretable.

**Comment by teolemon** [comment]
Generally speaking, it's still a bit messy, my suggestion above are also around more structure, and a cleaner panel

**Comment by stephanegigandet** [comment]
I'm now hiding the computation details, as those are not useful to end users, and are only useful to verify the score, for producers etc.

image

File: lib/ProductOpener/KnowledgePanels.pm [hunk 1]
```diff
@@ -882,6 +887,65 @@ sub create_nutriscore_panel ($product_ref, $target_lc, $target_cc, $options_ref)
...
+		$panel_data_ref->{title} = lang_in_other_lc($target_lc, "attribute_nutriscore_not_applicable_title");
+	}
+	else {
+		$panel_data_ref->{title}
+			= sprintf(lang_in_other_lc($target_lc, "attribute_nutriscore_grade_title"), uc($grade));
+		if ($panel_data_ref->{nutriscore_unknown_reason_short}) {
+			$panel_data_ref->{subtitle} = $panel_data_ref->{nutriscore_unknown_reason_short};
+		}
+		$panel_data_ref->{subtitle} = lang_in_other_lc($target_lc,
+			"attribute_nutriscore_" . $panel_data_ref->{nutriscore_grade} . "_description_short");
```
alexgarel, line 913: Aren't we missing a `else` here ? (otherwise you are over writing the subtitle attribute defined just above)

lib/ProductOpener/Food.pm
```diff
@@ -1610,7 +1612,8 @@ sub compute_nutriscore_data ($product_ref, $prepared, $nutriments_field, $versio
 		if ($is_fat_oil_nuts_seeds) {
 			# Add the fat and saturated fat / fat ratio
 			$nutriscore_data_ref->{fat} = $nutriments_ref->{"fat" . $prepared . "_100g"};
-			$nutriscore_data_ref->{saturated_fat_ratio} = saturated_fat_ratio($nutriments_ref, $prepared);
+			$nutriscore_data_ref->{saturated_fat_ratio}
+				= round_to_max_decimal_places(1, saturated_fat_ratio($nutriments_ref, $prepared));
```
alexgarel, line 1616: Arguments inversion !
```suggestion
				= round_to_max_decimal_places(saturated_fat_ratio($nutriments_ref, $prepared), 1);
```

lib/ProductOpener/Nutriscore.pm
```diff
@@ -707,6 +739,8 @@ sub compute_nutriscore_score_2023 ($nutriscore_data_ref) {
 				$nutriscore_data_ref->{$nutrient . "_points"}++;
 			}
 		}
+
+		$nutriscore_data_ref->{$nutrient . "_max"} = scalar @{$points_thresholds_2023{$nutrient_threshold_id}};
```
alexgarel, line 743: `_max` is a bit misleading here, maybe `max_points` would be more easy to understand.

Reading this code I though you made a mistake by not taking the last value of the array.

At least we should put a comment:
```suggestion
		# max number of squares in the gauge we will display to user for this nutrient
		$nutriscore_data_ref->{$nutrient . "_max"} = scalar @{$points_thresholds_2023{$nutrient_threshold_id}};
```

scripts/images/gen_svg_images_for_nutriscore_new_formula.pl
alexgarel, line 1: Thanks for putting in an adequate sub-folder.

scripts/images/gen_svg_images_for_nutriscore_new_formula.pl
```diff
@@ -0,0 +1,195 @@
...
+);
+
+# Read the Nutri-Score SVG images in the html/images/attributes/src directory
+# and add a "New formula" banner to the Nutri-Score logos.
+
+my $dir = "$www_root/images/attributes/src";
+
+my @files = glob "$dir/nutriscore-*.svg";
+
+foreach my $lc ("en", "fr") {
```
alexgarel, line 154: Only french and english ?

scripts/images/gen_svg_images_for_nutriscore_new_formula.pl
alexgarel, line 1: For generated images, we could also consider not putting them in the git, but putting them in a data folder ?
We could also use javascript with gulp for this kind of things (so that files ends in the dist folders, and developer experience is still easy).

scripts/images/nutriscore-new-formula-source.svg
alexgarel, line 1: Is it normal to have a serif-font for "New formula" ? (seems a bit less integrated to me).

templates/web/panels/panel.tt.html
```diff
@@ -106,8 +106,30 @@ <h5>[% panel.title_element.subtitle %]</h5>
                         <div class="large-8 small-12 columns">
                 [% END %]
 
+                [% IF panel_group_element.icon_url.defined %]
+                <img src="[% panel_group_element.icon_url %]"
+                    style="[% IF panel_group_element.icon_size == 'very_small' %]height: 24px;[% ELSIF panel_group_element.icon_size == 'small' %]height:36px;[% ELSE %]height:72px;[% END %]float:left;margin-right:1rem;"
```
alexgarel, line 111: couldn't we factor this kind of logic in a function ?

like `choose_size(panel_group_element.icon_size, "24px", "36px", "72px")`

alexgarel (approved):
Thanks for this move in the right direction !

Two blockers and some more minor comments, otherwise LGTM.

po/common/en.po
```diff
@@ -6673,3 +6673,144 @@ msgctxt "weight_percent"
...
+msgid "Presence"
+msgstr "Presence"
+
+msgctxt "absence"
+msgid "Absence"
+msgstr "Absence"
+
+msgctxt "nutriscore_is_water"
+msgid "This product is considered to be water for the calculation of the Nutri-Score."
+msgstr "This product is considered to be water for the calculation of the Nutri-Score."
```
alexgarel, line 6726: why isn't there nutriscore_is_beverage ?

alexgarel: ok it was already there

lib/ProductOpener/Food.pm
stephanegigandet: oops, good catch, thank you!

lib/ProductOpener/KnowledgePanels.pm
stephanegigandet: Good point, fixing.

scripts/images/gen_svg_images_for_nutriscore_new_formula.pl
stephanegigandet: Currently yes.

templates/web/panels/panel.tt.html
stephanegigandet: It's an option. Or in that case it might be simpler to just create CSS classes, and make the size part of the class name.

stephanegigandet: Filed https://github.com/openfoodfacts/openfoodfacts-server/issues/9752 so that we can take care of it later.

lib/ProductOpener/Nutriscore.pm
stephanegigandet: Good point, I renamed max to points_max

PR #9705, opened 2024-01-26
fix: last_modified_t needs to be a number and not a string

### LINKED ISSUE #9686: The "Recently modified" view is broken on the home page
### What
- The "Recently modified" view is broken on the home page
- It has been like that for 2 to 3 days at least
- The same products are at the top. I have disabled custom ranking. I have refreshed my browser.

stephanegigandet:
fix #9686

### DIFF
--- a/scripts/update_all_products.pl
+++ b/scripts/update_all_products.pl
@@ -139,6 +139,7 @@ my $fix_nutrition_data = '';
 my $compute_main_countries = '';
 my $prefix_packaging_tags_with_language = '';
 my $fix_non_string_ids = '';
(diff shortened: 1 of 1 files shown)


alexgarel (approved):
LGTM but why did it happened ? Isn't there any point in Product.pm to fix ?

stephanegigandet:
> LGTM but why did it happened ? Isn't there any point in Product.pm to fix ?

it happened in a recent change in update_all_products.pl, it's fixed in this PR, and the PR also has a new option to fix the products that were affected

PR #9706, opened 2024-01-26
test: Test coverage for integration tests

alexgarel:
This should bring coverage from the running backend server during integration tests.


docker/dev.yml
```diff
@@ -25,6 +25,10 @@ x-backend-conf: &backend-conf
   # e.g. when developing with Robotoff or openfoodfacts-query
   extra_hosts:
     - "host.docker.internal:host-gateway"
+  environment:
+    - PRODUCT_OPENER_COVERAGE
+    - COVER_OPTS
```
alexgarel, line 30:
```suggestion
    - DEVEL_COVER_OPTIONS
```

alexgarel:
I though it was an easy one, but it's not !

The problem is that apache is running scripts independently in some way, but codecoverage initialize only once. Even forcing apache to only one process does not work.

I can [force cgi scripts to report](https://github.com/pjcj/Devel--Cover/issues/244#issuecomment-742776545), but I end up with always 100% coverage, which is not true, and no info on branches, conditions, etc.

[This thread](https://www.nntp.perl.org/group/perl.qa/2007/06/msg8746.html) has a link to a working coverage apache example, but it's using ApacheTest mechanism.

alexgarel:
I'll close this because I didn't have the time to understand why it fails to get coverage server side…

PR #9707, opened 2024-01-26
feat: dq_category_specific_ingredient_percent_3

benbenben2:
### What
Continuity of previous PR https://github.com/openfoodfacts/openfoodfacts-server/pull/9606.
Now, for all jams, extra-jams, jellies.
According to EU regulations: https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A02001L0113-20131118
For existing categories in the taxonomy only.

### Screenshot
Screenshot_20240126_230542

### Related issue(s) and discussion
- Fixes #1414

PR #9709, opened 2024-01-27
taxonomy: correction sweet spreads

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/taxonomies/categories.txt
+++ b/taxonomies/categories.txt
@@ -93075,7 +93075,11 @@ nl:Rozeblaadjesjams
 
 <en:Sweet spreads
 <en:Plant-based spreads
-en:Ginger preserves, ginger jams
+en:Ginger preserves
+fr:Tartinades de gingembre
+
+<en:Ginger preserves
+en:Ginger jams
 fr:Confitures de gingembre
 
 <en:Plant-based spreads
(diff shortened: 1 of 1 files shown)


taxonomies/categories.txt
```diff
@@ -93462,10 +93491,6 @@ pt:Doces de figo
 sv:Fikonsylter
 wikidata:en:Q65545224
 
-<en:Jams
-en:ginger jams
-fr:Confitures de gingembre
-
```
benbenben2, line 93468: Is it possible to keep separated "jams" vs "preserves" for ginger?

The reason is that ginger jams in EU should have 150g of fruit per 1000g. Which is not the case for preserves. (see: https://github.com/openfoodfacts/openfoodfacts-server/pull/9707/files and https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A02001L0113-20131118)

moon-rabbitOFF: Ah yes, didn't see that sorry. Should then put ginger jams under ginger preserves?

benbenben2: Yes, that should be fine like this.

benbenben2 (approved):
Super. Thanks @moon-rabbitOFF for the correction.

PR #9721, opened 2024-01-30
fix: fix cache on countries list

alexgarel:


lib/ProductOpener/Web.pm
```diff
@@ -351,25 +351,30 @@ A reference to a list of hashes with every country code and their label in the $
 
 sub get_countries_options_list ($target_lc, $exclude_world = 1) {
 	# if already computed send it back
+	my @countries_list = ();
 	if (defined $countries_options_lists{$target_lc}) {
-		return $countries_options_lists{$target_lc};
+		@countries_list = @{$countries_options_lists{$target_lc}};
```
stephanegigandet, line 356: I think this creates a copy of the list, previously we just returned the reference.

Can you explain what this PR is about? Is it fixing a bug, or it's just refactoring? Just reading the code it sounds to me it's basically achieving the same thing, but with more operations on subsequent calls, what am I missing?

stephanegigandet: Ah, is it the exclude_world parameter? Then it that case I think we should cache both versions, that way we don't have to do the grep on every call when it's cached.

alexgarel: Is the grep costly ? I though it was not a big deal to grep again, that why !

stephanegigandet: It's not costly, not a big deal.

stephanegigandet (approved):

PR #9722, opened 2024-01-30
fix: fix_dq_facet_all_values_identical

benbenben2:
### What
Bug reported on Slack
Sometimes the unit test 'all identical values and above 1 in the nutrition table 2' in dataqualityfood.t fails.

It happens because the array "@major_nutriments_values" (see Ingredients.pm) can have values in random order. Sometimes the first value is "0.8" and since we have the condition "@major_nutriments_values[0] > 1" facet is not raised. And test fails.

See the code for suggested solution

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -1091,7 +1091,7 @@ sub check_nutrition_data ($product_ref) {
 			}
 		}
 
-		foreach my $nid (keys %{$product_ref->{nutriments}}) {
+		foreach my $nid (sort keys %{$product_ref->{nutriments}}) {
 			$log->debug("nid: " . $nid . ": " . $product_ref->{nutriments}{$nid}) if $log->is_debug();
 
 			if ($nid =~ /_prepared_100g$/ && $product_ref->{nutriments}{$nid} > 0) {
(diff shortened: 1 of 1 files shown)


stephanegigandet (changes requested):
In addition, I think you should add a sort in there:

foreach my $nid (keys %{$product_ref->{nutriments}}) {

->

foreach my $nid (sort keys %{$product_ref->{nutriments}}) {

So that we always go in the same order.

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -1164,21 +1164,25 @@ sub check_nutrition_data ($product_ref) {
...
 		# all values but one - because sodium and salt can be automatically calculated one depending on the value of the other - are identical
-		# and values (check first value only) are above 1 (see issue #9572)
+		# and values (check salt (should not check sodium which could be lower)) are above 1 (see issue #9572)
 		# and at least 4 values are input by contributors (see issue #9572)
 		if (
 			(
-				($nutriments_values_occurences_max_value == scalar @major_nutriments_values)
+				(
+					$nutriments_values_occurences_max_value == scalar @major_nutriments_values
+					and (@major_nutriments_values[0] > 1)
```
stephanegigandet, line 1176:
```suggestion
					and ($major_nutriments_values[0] > 1)
```

This currently causes a warning in the main branch:

Scalar value @major_nutriments_values[0] better written as $major_nutriments_values[0] at /opt/product-opener/lib/ProductOpener/DataQualityFood.pm line 1171.

benbenben2: done.

stephanegigandet (approved):
Thanks!

PR #9724, opened 2024-01-31
fix: move script syncing images to s3 to product-opener

raphael0202:
- move from openfoodfacts-infrastructure to openfoodfacts-server repo
- fix the script after server migration
- add systemd service and timer configuration files

The script runs correctly on production, but the service & timer config files have not been tested.

### DIFF
--- a/conf/systemd/sync_images_s3@.service
+++ b/conf/systemd/sync_images_s3@.service
@@ -9,5 +9,5 @@ Type=oneshot
 User=off
 Group=off
 WorkingDirectory=/srv/%i/scripts/sync-s3-images
-ExecStart=.venv/bin/python3 sync_s3_images.py /mnt/off/images/products
+ExecStart=.venv/bin/python3 sync_s3_images.py /mnt/%i/images/products
 KillMode=process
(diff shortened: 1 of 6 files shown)


conf/systemd/sync_images_s3@.service
```diff
@@ -0,0 +1,13 @@
...
+Description=Synchronize images to AWS S3 %i
+# __ will be replaced by @ in email-failures@.service
+OnFailure=email-failures@sync_images_s3__%i.service
+
+[Service]
+Type=oneshot
+User=off
+Group=off
+WorkingDirectory=/srv/%i/scripts/sync-s3-images
+ExecStart=.venv/bin/python3 sync_s3_images.py /mnt/off/images/products
```
alexgarel, line 12: Maybe we don't need to use a "@" here as you don't seem to really use the %i for now :-)

Or either we use it here:
```suggestion
ExecStart=.venv/bin/python3 sync_s3_images.py /mnt/%i/images/products
```

scripts/sync-s3-images/README.md
alexgarel, line 1: :+1:

alexgarel (approved):
Only a small comment.

On deployment we will have to setup the systemd service.

PR #9765, opened 2024-02-06
fix: nutriscore misc_tags removal

stephanegigandet:
small fix so that we correctly remove Nutri-Score related misc_tags (and only those tags) when we recompute Nutri-Score with update_all_products.pl


alexgarel (approved):
LGTM

PR #9767, opened 2024-02-06
fix: code stored as string

stephanegigandet:
Something very strange is happening:

$product_ref->{code} .= '';   worked if it was just before the store() call, but not inside the make_sure_numbers_as_stored_as_numbers() function.

This change seems to work, but I don't know why.


alexgarel (approved):
LGTM ! I hope it works !

Issue #9781, opened 2024-02-12
Warnings when computing Nutri-Score


stephanegigandet:
```
updating product code: 0746742400417  (1223969 / 3062367)Argument "absence" isn't numeric in sprintf at /srv/off/lib/ProductOpener/Numbers.pm line 166.
Use of uninitialized value $points_max in addition (+) at /srv/off/lib/ProductOpener/Nutriscore.pm line 796.
```

benbenben2:
Fixed by #9795?

PR #9795, opened 2024-02-13
fix: Nutri-Score v2 fixes and improvements to knowledge panels

stephanegigandet:
- use "undef" in the nutridscore data instead of 0 for fiber, fruits/vegetables/legumes values etc. when they are not known
- in the Nutri-Score panel, display "unknown" instead of 0 when a nutrient is unknown
- count the number of non-nutritive sweeteners (for beverages) and display it, instead of just indicating absence / presence
- fixes some warnings including #9781
- add some missing strings in en.po
- some bug fixes for special cases (like spring water with no nutrition data, where we do display Nutri-Score A anyway)

image

image


teolemon:
- Can we put the "Nutri-Score not applicable" at the top of the hierarchy ? (example: https://world.openfoodfacts.org/product/4009900484695/wrigley-s-orbit) 
image

alexgarel (approved):
Good !

Issue #9798, opened 2024-02-14
Upgrade Agribalyse version from 3.1 to 3.1.1


stephanegigandet:
There are strange values for chocolate bars in 3.1

PR #9799, opened 2024-02-14
fix: update Agribalyse version from 3.1 to 3.1.1

stephanegigandet:
the main change I noticed is that it fixes strange values for chocolate bars

fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/9798


alexgarel (approved):
Great !

Issue #1405, opened 2018-10-14
Add button to revert whole product to previous revision


teolemon:
### What

- [ ] Add button to revert whole product to previous revision

![image](https://user-images.githubusercontent.com/1689815/46919051-a6e9e780-cfd9-11e8-971d-9adb389bdc56.png)
Useful when only the last edit is involved.

### Part of
- #5538

alexgarel:
I would be in favor of keeping the history, that is act as a revert in git.

That is : keep the faulty edit, but add an edit on top of that to be back on.

stephanegigandet:
Yes, that's definitely how it should work. In practice: just read the old revision, and then use Products::store_product() to write it again. (that would have the drawback of losing images uploaded in between, we could be a bit more clever and add uploaded images (but not selected images) from the current "images" field.

CharlesNepote:
We could also do this on the client side with JS. We could do a first prototype with the Power User Script, with the following advantages:
* the JS functions to use the API (write) would be reusable for other needs or projects
* this feature would be reusable if we decide one day to change the language used for the backend
* it enforces us to better separate backend and frontend
* it enforces us to produce a better HTML code, with classes and ids to manipulate the document (style, dynamics, etc.)

alexgarel:
I don't think it would be handy to do this in JS. In our current settings it will be hard to test, easy to break. For a function like this, you really make your life easier by going as much close as possible to the database (which are the `.sto`, here)

CharlesNepote:
Ok, I understand.

teolemon:
@CharlesNepote would you be able to come up with a gross estimate of the time this feature could save per week. I've put 100 min (50 reverts * 2 min)

CharlesNepote:
100 min per week, at least, seems reasonable. Maybe there are fewer products to revert but I think it takes more time than you expect. What I can say is that there are ~5-10 vandalisms per week detected thanks  to Mirabelle and probably ~20-30 detected via the moderation Slack channel.

The worst part is that sometimes it's very difficult and/or long to revert manually.

The more we grow, the more we will have vandalism. I would say that this is a good P1 candidate.

PR #9800, opened 2024-02-14
feat: Revert product to a previous revision (API + upcoming website integration for moderators)

stephanegigandet:
Will fix https://github.com/openfoodfacts/openfoodfacts-server/issues/1405

- adds a new /api/v3/product_revert API
- API tests
- new centralized place (Permissions.pm) to check if a user has permission to do something, instead of having checks for admin or moderator status everywhere in the code
- integration in website (in product history on product page + product edit form), includes a confirmation

image

Still missing:
- API doc


teolemon:
Can we have some kind of log/notification when a moderator reverts a product ?

stephanegigandet:
> Can we have some kind of log/notification when a moderator reverts a product ?

I'm sending a mail to admins, we can revisit once we have a better system in place for events / notifications / patrol.

html/js/product-history.js
```diff
@@ -0,0 +1,54 @@
...
+            $.ajax({
+                url: '/api/v3/product_revert',
+                type: 'POST',
+                contentType: "application/json; charset=utf-8",
+                dataType: "json",
+                data: JSON.stringify({
+                    code: code,
+                    rev: rev,
+                    fields: "rev"
+                }),
```
alexgarel, line 40: Shan't there be the lc and cc parameters ? (so that we get back localized product name ?)

lib/ProductOpener/Permissions.pm
```diff
@@ -0,0 +1,104 @@
...
+
+sub has_permission_product_revert ($request_ref) {
+
+	my $has_permission = 0;
+
+	if ($request_ref->{admin} or $request_ref->{moderator}) {
+		$has_permission = 1;
+	}
+
+	return $has_permission;
```
alexgarel, line 63: Why make it so long ?

```suggestion
	return ($request_ref->{admin} or $request_ref->{moderator});
```

lib/ProductOpener/Permissions.pm
alexgarel, line 1: Good idea to add this module ! This will bring clarity.

lib/ProductOpener/APIProductRevert.pm
```diff
@@ -0,0 +1,181 @@
...
+	$error += check_user_permission($request_ref, $response_ref, "product_revert");
+
+	if (not $error) {
+
+		# Load the product
+		my ($code, $ai_data_string) = &normalize_requested_code($request_body_ref->{code}, $response_ref);
+		my $rev = $request_body_ref->{rev};
+
+		# Check if the code is valid
+		if ($code !~ /^\d{4,24}$/) {
```
alexgarel, line 104: We really have to centralize this kind of checks in Product.pm as we may extend codes in a few months !
So make it a method !

lib/ProductOpener/Display.pm
```diff
@@ -854,6 +861,20 @@ sub init_request ($request_ref = {}) {
...
+	else {
+		$request_ref->{moderator} = 0;
+	}
+
+	if ($User{pro_moderator}) {
+		$request_ref->{pro_moderator} = 1;
+	}
+	else {
+		$request_ref->{pro_moderator} = 0;
+	}
```
alexgarel, line 876:
```suggestion
	$request_ref->{moderator} = !!$User{moderator};
	$User{pro_moderator} = !!$request_ref->{pro_moderator};
```

lib/ProductOpener/Display.pm
```diff
@@ -10563,9 +10586,25 @@ sub display_product_history ($code, $product_ref) {
...
+		$scripts .= <<SCRIPTS
+<script>var revert_confirm_message = "$revert_confirm";</script>
+<script src="$static_subdomain/js/dist/product-history.js"></script>
+SCRIPTS
+			;
+
+		$initjs .= <<JS
+activate_product_revert_buttons_in_history();
+JS
+			;
```
alexgarel, line 10605: Don't we want to avoid this kind of JS inclusion (I find it quite annoying) ?

Why isn't it included in the template. Or we could maybe just use a specific class to trigger the JS activity.

taxonomies/nutrient_levels.txt
```diff
@@ -138,7 +138,7 @@ qu:Wira in low quantity
 rm:Fat in low quantity
 rn:Fat in low quantity
 ro:Grăsimi în cantitate mică
-ru:Жиры в небольшое количество
+ru:Жиры в в малом количестве
```
alexgarel, line 141: ? why is the `в` repeated ?

(see also below)

tests/integration/api_v3_product_revert.t
```diff
@@ -0,0 +1,139 @@
...
+		body => '{
+			"fields" : "updated",
+			"product": { 
+				"product_name_en": "Test product updated",
+				"brands_tags": ["Test brand updated"],
+				"categories_tags": ["en:coffees"],
+				"countries_tags": ["en:france"]
+			}
+		}',
+	},
```
alexgarel, line 77: I would not consider that a test ! We should at least add a comment that it's a setup.

tests/integration/api_v3_product_revert.t
```diff
@@ -0,0 +1,139 @@
...
+	{
+		test_case => 'revert-product-no-code',
+		method => 'POST',
+		path => '/api/v3/product_revert',
+		body => '{
+			"rev": 1,
+			"fields": "code,rev,product_name_en,brands_tags,categories_tags"
+		}',
+		ua => $moderator_ua,
+	},
```
alexgarel, line 88: I don't see a specific return code. Does this mean we send back a 200 ? I think this is important to return a non 20x code (to keep REST like and ease handling by a lot of languages, JS included).

alexgarel (commented):
Thanks for this great PR !

Overall it's good, but it needs some minor fixes IMO.

html/js/product-history.js
stephanegigandet: The localized status / error name you mean? (we don't display the product name)
It will get the right default value from the subdomain:

image
(tested with a translation added manually in fr.po)

lib/ProductOpener/Display.pm
stephanegigandet: Here I wanted to set the value to 0 specifically, but we can use undef instead (!! 0 or !! undef seems to be equal to undef)

html/js/product-history.js
alexgarel: ok !
Maybe add a comment to explain it ? (for the case we evolve it later) ?

taxonomies/nutrient_levels.txt
alexgarel: FYI, it seems to come from taxonomy generations…

stephanegigandet: @alexgarel I made a PR to remove this file from GitHub: https://github.com/openfoodfacts/openfoodfacts-server/pull/9851

tests/integration/api_v3_product_revert.t
stephanegigandet: It's a test set up indeed, I find it clearer to have it here in sequence with the tests, compared to creating and editing the product beforehand with edit_product().

I'm adding a setup => 1 option and removing the test_case so that the results don't get saved and checked.

html/js/product-history.js
stephanegigandet: Sure

lib/ProductOpener/APIProductRevert.pm
stephanegigandet: Good point.

lib/ProductOpener/Display.pm
stephanegigandet: OK, I'm moving it.

tests/integration/api_v3_product_revert.t
stephanegigandet: Good point, I'm changing to 40x codes for errors

lib/ProductOpener/APITest.pm
```diff
@@ -402,208 +405,229 @@ Note: this setting can be overriden for each test case by specifying a "ua" fiel
...
+
+sub execute_api_tests ($file, $tests_ref, $ua = undef) {
 
+	my ($test_id, $test_dir, $expected_result_dir, $update_expected_results) = (init_expected_results($file));
+
+	$ua = $ua // LWP::UserAgent->new();
+
+	foreach my $test_ref (@$tests_ref) {
+
+		my $response = execute_request($test_ref, $ua);
```
alexgarel, line 623: You did an incredible job at refactoring. Sorry if it feels picky, but I still have a request: I would check the expected code response even in case of setup, it can help understanding that a test fails because a setup test did go wrong.

alexgarel (approved):
You did an incredible job ! Thank you.

lib/ProductOpener/APITest.pm
stephanegigandet: Good idea. So for setup test cases, I now still call check_requests_response (so that we can check expected status code), but I don't save or compare the actual response body.

alexgarel:
:tada:

PR #9801, opened 2024-02-15
taxonomy: add origins translations

ArturLange:
### What
Added Polish origin translations for some of the most popular origins

### DIFF
--- a/taxonomies/countries.txt
+++ b/taxonomies/countries.txt
@@ -44965,7 +44965,7 @@ pam:Pisanmetung a Ka-arian, United Kingdom
 pap:Reino Uni
 pcd:Roéyôme-Uni
 pih:Yunitid Kingdum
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Thank you!

stephanegigandet:
@ArturLange it seems there is the same translation for both UK and Great Britain, could you change one of them?

ERROR - pl:wielka-brytania already is associated to en:united-kingdom (origins) - pl:wielka-brytania cannot be mapped to entry en:great-britain

ArturLange:
I thought that might be a problem, but I couldn't find it in `origins.txt`. Should be fixed now.

stephanegigandet:
> I thought that might be a problem, but I couldn't find it in `origins.txt`. Should be fixed now.

Thank you!

Issue #9820, opened 2024-02-21
Account removal is broken


alexgarel:
One user told us he can't click on the remove my account button.

The user.pl javascript seems broken.

If the console: 
`Uncaught SyntaxError: missing ) after argument list`

alexgarel:
I did a hotfix in prod with [% lang("delete_confirmation") | replace("'", "\\'") %].

The complete fix is #9822

Issue #9533, opened 2023-12-15
Unable to delete user (for admins), JS error due to unescaped apostrophe in French translation


stephanegigandet:
Reported by @raphael0202 :

It looks like I can't delete a user on prod (https://fr.openfoodfacts.org/cgi/user.pl?type=edit&userid=EMAIL), is this issue known?

stephanegigandet:
There's a translation that contains a ' and it's unescaped:  "onclick="if ('email'.localeCompare(prompt('Cela supprimera vos informations d'utilisateur et anonymisera toutes vos contributions. Veuillez ressaisir votre nom d'utilisateur pour confirmer.'),"

stephanegigandet:
Fixed by @alexgarel https://github.com/openfoodfacts/openfoodfacts-server/pull/9822

PR #9822, opened 2024-02-21
fix: js and json lang quotes escape

alexgarel:
Translations coming from crowdin might contain single quotes or double quotes. So whenever we use lang in a json or javascript string we should escape quotes.

I created edq and esq (Escape Single/Double Quotes) function for that (1st commit) and applied it in a lot of place (2nd commit). This should be the pattern to use from now on. 

fixes: 
- #9820
- #9533

**Note:** I will do a hotfix in prod for the remove account case, because it's important to fix quickly (GDPR).

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -398,7 +398,7 @@ sub process_template ($template_filename, $template_data_ref, $result_content_re
 	$template_data_ref->{f_lang} = \&f_lang;
 	# escaping quotes for use in javascript or json
 	# using short names to favour readability
-	$template_data_ref->{esq} = sub {escape_char(@_, "\'")};    # esq as escape_single_quote
+	$template_data_ref->{esq} = sub {escape_char(@_, "\'")};    # esq as escape_single_quote_and_newlines
 	$template_data_ref->{edq} = sub {escape_char(@_, '"')};    # edq as escape_double_quote
 	$template_data_ref->{lang_sprintf} = \&lang_sprintf;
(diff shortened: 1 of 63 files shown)


alexgarel:
This is a big PR ! I still have to double read my changes to be sure.

templates/api/knowledge-panels/contribution/data_quality_tags.tt.json
```diff
@@ -58,7 +58,7 @@
                 {
                     "element_type": "action",
                     "action_element": {
-                        "html": `[% lang(action_name) %]`,
+                        "html": `[% edq(lang(action_name)) %]`,
```
alexgarel, line 61: it's not needed here… as we are using back-thick.

I have to check in other places.

alexgarel:
Ideally we should add a check that we don't use lang / f_lang without edq or esq if we are inside quotes. (but I don't think of an easy way to do this).

alexgarel:
This is ready for review & merge @stephanegigandet

lib/ProductOpener/Display.pm
```diff
@@ -5736,12 +5740,21 @@ sub search_and_export_products ($request_ref, $query_ref, $sort_by) {
 	return;
 }
 
+sub escape_char($s, $char) {
```
stephanegigandet, line 5743: Looks good. Maybe this function (and escape_single_quote) could be moved to Text.pm

stephanegigandet (approved):
Looks good, thank you!

PR #9841, opened 2024-02-26
chore: Test taxonomy sort

alexgarel:
This is a test to see how taxonomies moves if we lint them with sort_each_taxonomy_entry.pl


alexgarel:
I opened https://github.com/openfoodfacts/openfoodfacts-server/pull/10156 which is up to date.

PR #9842, opened 2024-02-26
fix: fix Redis utf-8 encoding issue

raphael0202:
We should encode values to UTF8, otherwise we get a UTF8 decode error when there are accents the values (user_id, comment,...)

Fixes
https://openfoodfacts.sentry.io/issues/5013496912/?project=4505924518543360&query=is%3Aunresolved+is%3Afor_review+assigned_or_suggested%3A%5Bme%2C+my_teams%2C+none%5D&referrer=issue-stream&sort=inbox&statsPeriod=30d&stream_index=1


alexgarel (approved):

raphael0202:
Indeed! I just tested on staging, it works as expected. I'm deploying the fix now in prod

PR #9818, opened 2024-02-20
feat: revamped taxonomy sort script

alexgarel:
to really sort everything (within an entry)

PR #9843, opened 2024-02-26
chore: move taxonomies related scripts to their own folder

alexgarel:
I didn't move sort_taxonomies yet as it is done in https://github.com/openfoodfacts/openfoodfacts-server/pull/9818


benbenben2 (approved):
LGTM

PR #9844, opened 2024-02-26
fix: display email warning for pro also on public platform

alexgarel:
Display the warning about email address also on public platform.


templates/web/pages/user_form/user_form.tt.js
```diff
@@ -1,7 +1,14 @@
-\$('#pro').change(function() {
+
+\$(document).ready( function () {
+
+	\$('#pro').change(function() {
 	if (\$(this).prop('checked')) {
 		\$('.pro_org_display').show();
 		\$('#teams_section').hide();
+		if ( ! \$('#pro-email-warning').length ) {
+			let pro_email_warning = '<div style="color: red; font-weight: 600;" id="pro-email-warning">🚨 [% lang("email_warning") %]</div>';
```
stephanegigandet, line 9: This fails in French because the translation contains a single quote, so maybe merge https://github.com/openfoodfacts/openfoodfacts-server/pull/9822 first

stephanegigandet (approved):
Thanks!

templates/web/pages/user_form/user_form.tt.js
alexgarel: I will use backthick, it's even better.

alexgarel: or not, finally, (because in this case we should escape back thick !)

PR #9854, opened 2024-03-01
refactor: support for obf, opf, opff flavors [start]

stephanegigandet:
- Removed the Config.pm symbolic link and created a Config.pm module that loads Config_obf.pm etc. at runtime based on the value of the PRODUCT_OPENER_FLAVOR_SHORT environment variable
- Refactored the setenv.sh script to make it more generic
- Moved some files in the /conf directory to make it more clear which config files are specific to the docker install

Setup needed if not running through Docker:

To be able to run scripts from the off user:

in the .profile file of the off user, add:
```
cd /srv/off/
source env/setenv.sh off
cd
```


stephanegigandet:
Tested on https://world.openfoodfacts.dev/ (running debian 11 bullseye directly without docker)

conf/apache-2.4/sites-available/obf.conf
alexgarel, line 1: Beware that those file might not be up to date compared to obf-main for example (it's more true for nginx files).

docker-compose.yml
```diff
@@ -64,8 +64,10 @@ x-backend-conf: &backend-conf
     - ./logs/apache2:/var/log/httpd
 
     # Apache conf
-    - ./conf/apache.conf:/etc/apache2/sites-enabled/product-opener.conf
-    - ./conf/apache-mpm_prefork.conf:/etc/apache2/mods-avalable/mpm_prefork.conf
+    - ./conf/apache2-docker/apache.conf:/etc/apache2/sites-enabled/product-opener.conf
+    - ./conf/apache2-docker/apache-mpm_prefork.conf:/etc/apache2/mods-enabled/mpm_prefork.conf
+    - ./conf/apache-2.4/modperl.conf:/etc/apache2/conf-enabled/modperl.conf
```
alexgarel, line 69: cool that it's the same file as production !

docker/docker-entrypoint.sh
```diff
@@ -7,7 +7,7 @@ ln -sfT /opt/product-opener/po/${PRODUCT_OPENER_FLAVOR} /mnt/podata/po/site-spec
 ln -sfT /opt/product-opener/lib/ProductOpener/SiteLang_${PRODUCT_OPENER_FLAVOR_SHORT}.pm /opt/product-opener/lib/ProductOpener/SiteLang.pm
 
 # Link Config.pm and Config2.pm
-ln -sfT /opt/product-opener/lib/ProductOpener/Config_${PRODUCT_OPENER_FLAVOR_SHORT}.pm /opt/product-opener/lib/ProductOpener/Config.pm
+#ln -sfT /opt/product-opener/lib/ProductOpener/Config_${PRODUCT_OPENER_FLAVOR_SHORT}.pm /opt/product-opener/lib/ProductOpener/Config.pm
```
alexgarel, line 10: We can remove the line :-)

lib/ProductOpener/Config_opff.pm
```diff
@@ -147,35 +153,17 @@ use ProductOpener::Config2;
...
-	kyzh
-	lafel
-	lucaa
-	mbe
-	moon-rabbit
+	manoncorneille
 	raphael0202
-	sebleouf
-	segundo
+	sarazine-ouattara
```
alexgarel, line 163: I think we can remove sarazine !

lib/ProductOpener/GS1.pm
```diff
@@ -342,6 +342,7 @@ foreach my $tag (sort keys %{$gs1_maps{allergenTypeCode}}) {
 		$log->error("gs1_maps - entry not in taxonomy",
 			{tagtype => "allergens", tag => $gs1_maps{allergenTypeCode}{$tag}})
 			if $log->is_error();
+		print STDERR "tag: $tag - canon_tag: $canon_tag\n";
```
alexgarel, line 345: do you want to keep it ?

env/setenv.sh
alexgarel, line 1: Cool pattern.

env/env.off
```diff
@@ -0,0 +1,7 @@
+# NOTE: this is an addition to .env file, not a standalone file !
+# It should be loaded with: source env/setenv.sh <env>
+
+COMPOSE_PROJECT_NAME=po_off
```
alexgarel, line 4: Should we put po_off as default also in .env ?

alexgarel (approved):
Great move !

alexgarel:
There is a work to be done on the nginx files. (the one in prod differs from the one in this branch).

I think if we could unite them through some sensible includes, it would be great (also with the docker one, to avoid changing one and not the other)

lib/ProductOpener/GS1.pm
stephanegigandet: I'm moving it to the die() statement.

env/env.off
stephanegigandet: Done.

docker/docker-entrypoint.sh
stephanegigandet: Done.

PR #9860, opened 2024-03-02
taxonomy: improve stop words in HR

benbenben2:
### What
improve stop words in HR

- Fixes #-none-

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -4931,9 +4931,10 @@ my %phrases_after_ingredients_list = (
 
 	hr => [
 		'(č|Č)uvati|(č|Č)uvajte',    # store in...
+		'izvaditi',    # remove from the refrigerator half an hour before consumption
 		'najbolje upotrijebiti do',    # best before
 		'nakon otvaranja',    # after opening
-		'neotvoreno', # not opened can be stored etc.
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -4930,13 +4930,11 @@ my %phrases_after_ingredients_list = (
 	],
 
 	hr => [
-		'bez konzervans',    # without preservatives
```
stephanegigandet, line 4933: Just curious: why all the removals? they were not relevant / useful?

benbenben2: I added them long time ago.

'bez konzervans' and 'izvor dijetalnih vlakana' are labels. If I understand well, they can be extracted and put to labels tags

čuvati is problematic. It seems not to be working all the time. Hence, I will try to add both upper and lower cases for now. Is there a way to test it locally? (when I upload picture locally, I cannot extract text from it). Additionally, čuvati literally means "to store", which will not appears in the middle of ingredients, so, this is why I removed optional text that appears after: (zatvoreno na|pri sobnoj temperaturi|u zamrzivaču na).

Finally, the first line, it also contain čuvati. Optional text before is not clear, it seems to me that it could be added as new line. I will remove it for now, and wait to see again these occurences in the future and eventually add them again.

PR #9864, opened 2024-03-05
refactor: different taxonomy source files for flavors

stephanegigandet:
I think this is the simplest way to allow Product Opener flavors (OFF, OBF, OPF, OPFF) to have some taxonomies that are different.

- I created /food and /beauty to put taxonomies that are specific to a product type.
- At this point I moved only the ingredients.txt and categories.txt file, and let all other taxonomies in the root directory (even if they are specific to food). This is to avoid breaking in flight PRs. Similarly, I didn't create a /common directory. We can revisit this later and move more taxonomies.
- Resulting taxonomies are build and cached as usual (in particular the resuting files will be in /taxonomies/ even for specific taxonomies).

The drawback of this approach is that one instance of Product Opener is tied to a specific product type, it can't load taxonomies for multiple types. If at some point we want to have one Product Opener instance serve multiple product types, we can revisit it as well.


alexgarel:
> If at some point we want to have one Product Opener instance serve multiple product types, we can revisit it as well.

Yes it seems to me that in our plans we spoke of having such an instance… to do later :-)

lib/ProductOpener/Config.pm
```diff
@@ -23,10 +23,16 @@ package ProductOpener::Config;
...
-die $@ if $@;
-eval("ProductOpener::Config_$flavor->import;");
-die $@ if $@;
+
+if (not defined $flavor) {
+	die("The PRODUCT_OPENER_FLAVOR_SHORT environment variable must be set.");
+}
+
+use Module::Load;
+autoload("ProductOpener::Config_$flavor");
```
alexgarel, line 36: Nice :-)

alexgarel (approved):
Great.

alexgarel:
BTW @stephanegigandet you should not fear moving file, git will be intelligent enough to follow changes through move if you use `git mv`.

PR #9865, opened 2024-03-05
fix: expose off-web html and fix nginx static resources serving

alexgarel:
First thing is that this PR fix a long standing bug that was missed: we were using multiple locations for same files, one hiding the other. Changed the way we handle expiry so that it really works as intended, using nginx best practices.

Second openfoodfacts-web as an html/ folder that is currently not exposed, while it should… this is done through try_file directive in nginx configuration.


scripts/deploy/verify-deployment.sh
alexgarel, line 1: I renamed lan-default to web-default/lang to be coherent with off-web layout.

alexgarel:
NOTE: I alrdeay did those commands on off and off-pro to insure next deployment won't fail:

```bash
sudo ln -s /srv/off/conf/nginx/snippets/expiry-headers.include /etc/nginx/snippets/
sudo -u off ln -s /srv/openfoodfacts-web/html /srv/off/html/off_web_html
```

```bash
sudo ln -s /srv/off/conf/nginx/snippets/expiry-headers.include /etc/nginx/snippets/
sudo -u off ln -s /srv/openfoodfacts-web/html /srv/off-pro/html/off_web_html
```

conf/nginx/sites-available/obf
```diff
@@ -66,17 +69,19 @@ server {
 
 	location /data/ {
 		include snippets/off.cors-headers.include;
+		include snippets/exiry-headers.include;
```
stephanegigandet, line 72:
```suggestion
		include snippets/expiry-headers.include;
```

The typo is repeated 20 times. The actual file is named correctly.

conf/nginx/snippets/expires-no-json-xml.conf
```diff
@@ -8,32 +8,65 @@
...
+  # Media: images, icons, video, audio, HTC
+  "~*\.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|htc|ico)$" public;
 }
 
-# CSS and Javascript
-location ~* \.(?:css|js)$ {
-  expires 1d;
-#  expires 1y;
-  access_log off;
+map $uri $productopener_access_file_prefix {
```
stephanegigandet, line 33: good idea to have different access log files

conf/nginx/snippets/expiry-headers.include
```diff
@@ -0,0 +1,6 @@
+# defines expiry healers for static resources
```
stephanegigandet, line 1:
```suggestion
# defines expiry headers for static resources
```

conf/nginx/sites-available/obf
alexgarel: :grimacing: thanks !

alexgarel: fixed.

stephanegigandet (approved):

Issue #9173, opened 2023-10-19
Add a "create product" button to the barcode search result page


aleene:
### Problem
The barcode search result page is often a dead end, allow the user to create a new product for the barcode he just searched. Now the user has to open another page, reenter the barcode and start again.
<img width="1106" alt="Screenshot 2023-10-19 at 08 42 07" src="https://github.com/openfoodfacts/openfoodfacts-server/assets/1329905/5e2f9d32-858f-4a55-a113-171e010bf367">

### Proposed solution
Add a line with a button to create the product (in edit-mode).

### Number of products impacted
Anytime the user wants to see the data for a product that does not exist yet

### Time per product
10 seconds(?) per search.

alexgarel:
Yes it also help discover how to add a product !

Shrutakeerti:
hi @aleene i want to contribute in this issue , I request you to assign me this issue

alexgarel:
@Shrutakeerti just tell you are working on it, we don't asign issue, it's useless.

stephanegigandet:
A clarification on what is needed:

For this issue, we want to change the output of /cgi/search.pl

so that it adds a link to the page to create a product (e.g. https://world.openfoodfacts.lorg/cgi/product.pl?code=42424243243255&type=edit )

when a user searches for a barcode (e.g. 42424243243255 ) and we don't have any result for it.

komalverma04:
@stephanegigandet I am starting working on this issue.

tchen0125:
If you are not no longer working on this, me and my group will work on this. @komalverma04

PR #9868, opened 2024-03-06
feat: Add "create product" button to barcode search result page

Shrutakeerti:
#9173 solves this issue

- Fixes #[ISSUE NUMBER]


stephanegigandet (changes requested):
Hi @Shrutakeerti 

For this issue, we want to change the output of /cgi/search.pl

so that it adds a link to the page to create a product (e.g. https://world.openfoodfacts.lorg/cgi/product.pl?code=42424243243255&type=edit )

when a user searches for a barcode (e.g. 42424243243255 ) and we don't have any result for it.

Issue #9364, opened 2023-11-22
Small issues in the nutrition facts table for us.openfoodfacts.org


CharlesNepote:
The specific table for the US is a very good idea. I have noticed few issues though:

- [ ] 1. When editing old products, salt is often displayed but is not automatically computed after entering sodium
- [ ] 2. Potassium don't have `% DV` unit option (and it is often only displayed with %DV units on the packaging) 
- [ ] 3. When adding a new field, a new one isn't created (we can't add more than one field)
- [ ] 4. Suggest as you type for new nutrient only contain `Vitamin A`
- [ ] 5. Potassium should be displayed after Iron

CharlesNepote:
All these issue really have a bad impact on data quality. I often fix data quality issues that are directly related to these issues.

stephanegigandet:
There's a JS issue that causes at least some of those issues, I will fix it.

```
Uncaught TypeError: Cannot read properties of undefined (reading 'replace')
    at check_nutrient (product.pl?type=edit&code=80177616:5525:59)
    at product.pl?type=edit&code=80177616:5585:2
    at Array.forEach (<anonymous>)
    at HTMLDocument.<anonymous> (product.pl?type=edit&code=80177616:5581:23)
    at l (jquery.js? [sm]:24:22328)
    at Object.fireWith [as resolveWith] (jquery.js? [sm]:24:23140)
    at Function.ready (jquery.js? [sm]:24:24936)
    at HTMLDocument.o (jquery.js? [sm]:14:962)
```

CharlesNepote:
Thanks a lot @stephanegigandet.

PR #9882, opened 2024-03-07
fix: fix JS issue for US product edit form + refactor

stephanegigandet:
This solves a JS error that broke all dynamic JS functions on the product edit form for US product: #9364 

While I was at it, I moved a bunch of JS code that was dynamically generated from templates/web/pages/product_edit/product_edit_form_display.tt.js to the static product-multilingual.js file.


alexgarel (approved):
Thanks, also for renaming variables !

PR #9885, opened 2024-03-07
fix: generate and build nutrient_levels.txt

stephanegigandet:
fix of https://github.com/openfoodfacts/openfoodfacts-server/pull/9883


stephanegigandet:
Not needed, closing

PR #9888, opened 2024-03-08
refactor: remove site-specific translations for flavors

stephanegigandet:
This PR removes the site-specific .po files in po/open(.*)facts/
Entries in the files are either:
- added to Config_*.pm %options if they are not to be translated (e.g the site name: we keep Open Beauty Facts everywhere)
- moved to the common .po files and suffixed with either the flavor (e.g. "tagline_obf") or the product type (e.g. ""categories_example_beauty")

Also:
- add SVG / PNG logos at the right sizes and with no padding for the flavors (started with OBF, will need to add the others)

To do:
- add a new lang_with_optional_suffix("stringid", [flavor or product type]) that checks if there is stringid_obf or stringid_beauty
- replace the calls to lang to use %options or lang_with_optional_suffix()
- remove windows phone icons and links (store is closed)
- move strings for OPF, OPFF
- remove site specific .po files

Work in progress.

image


lib/ProductOpener/Display.pm
```diff
@@ -7044,7 +7054,7 @@ sub display_page ($request_ref) {
 	my $type;
 	my $id;
 
-	my $site = "<a href=\"/\">" . lang("site_name") . "</a>";
+	my $site = "<a href=\"/\">" . $options{site_name} . "</a>";
```
alexgarel, line 7057: why not:
```suggestion
	my $site = "<a href=\"/\">" . lang($options{site_name}) . "</a>";
```
?

alexgarel: I see we drop it also below, so maybe there is a good reason ? Is that because you consider it's not to be translated (Open Food Facts).

po/common/common.pot
```diff
@@ -6874,6 +6822,82 @@ msgctxt "description"
...
+msgctxt "footer_tagline_off"
+msgid "A collaborative, free and open database of food products from around the world."
+msgstr "A collaborative, free and open database of food products from around the world."
+
+#. make sure the text file exists for your language, otherwise ask @teolemon 
+msgctxt "get_the_app_link_off"
+msgid "/open-food-facts-mobile-app"
+msgstr "/open-food-facts-mobile-app"
+
+# OBF
```
alexgarel, line 6846:
```suggestion
# OBF - beauty
```

po/common/common.pot
```diff
@@ -6874,6 +6822,82 @@ msgctxt "description"
 msgid "Description"
 msgstr "Description"
 
+# Overrides by flavor (e.g. obf) or product type (e.g. beauty)
+
+# OFF
```
alexgarel, line 6827:
```suggestion
# OFF - food
```

lib/ProductOpener/Display.pm
stephanegigandet: Right, we should not translate Open Food Facts.

PR #9889, opened 2024-03-08
chore: more category doubles

aleene:
### What
More category doubles taken from the taxonomy editor


stephanegigandet (approved):
Thank you!

PR #9902, opened 2024-03-12
taxonomy: fix taxonomies for lines double for same langage

alexgarel:


taxonomies/food/ingredients.txt
```diff
@@ -37380,7 +37375,7 @@ et:jõhvikas, harilik jõhvikas
 fa:کرنبری
 fi:karpalo, karpalot
 fo:tranuber
-fr:airelles, cranberries, cranberry, cranberrys, canneberge
+fr:airelles, cranberries, cranberry, cranberrys, canneberge, Canneberge crue, Cranberry crue
```
moon-rabbitOFF, line 37378: airelles and canneberges are different fruits

moon-rabbitOFF (changes requested):

stephanegigandet (approved):
Awesome, thanks for repairing all those issues. :)

taxonomies/food/ingredients.txt
stephanegigandet: @moon-rabbitOFF We currently mix "airelles" and "canneberges" before this PR, so I suggest we merge it as-is, and it would be great if you could create another PR that would make the distinction between airelles and canneberges. From reading https://fr.wikipedia.org/wiki/Vaccinium I'm completely lost..

Issue #9793, opened 2024-02-13
Automatically unselect images when source image is deleted


teolemon:
### What
- Automatically unselect images when source image is deleted

stephanegigandet:
Notes to solve this:

Only admins can delete images, you need to edit lib/ProductOpener/Config_off.pm to add your userid to the list of admins, and then create this user on your local dev environment.

Related code is is in lib/ProductOpener/Images.pm 

the function called when images are deleted is:
sub process_image_move ($user_id, $code, $imgids, $move_to, $ownerid) {

Selected images are stored in $product_ref->{images} , once an image is deleted, you will need to check the images to see which ones have the "imgid" field equal to the id of the image deleted, and then call process_image_unselect.

e.g. https://world.openfoodfacts.org/api/v3/product/4056489637295/flammkuchen-vegetarisch-duc-de-coeur?fields=images

himanshi-154:
I would like to work on this issue!

stephanegigandet:
Steps to verify that the problem is solved:
- Create a user in your local product opener
- Add your userid to %admins in Config_off.pm , and run "make restart" to restart the backend
- Add or edit a product, upload an image, and crop it and select it in the ingredients section for instance
- As your are logged in as an admin, you can delete the source image (click on "Manage images" at the top of the product edit form)
- Once the source image is deleted, reload the product page
- You should not see the cropped ingredients image, as it should have been deselected

komalverma04:
@stephanegigandet can i work on this issue.

himanshi-154:
@komalverma04 I have already done some work on this issue, please refer this [PR](https://github.com/openfoodfacts/openfoodfacts-server/pull/9950) and let me know if you find something that I need to improve in it!

PR #9906, opened 2024-03-12
feat: issue#9793 Automatically unselect image when the source image is deleted

himanshi-154:
Made changes in Config.pm to add my github userid in admins list and then made changes in Images.pm to solve the [issue #9793 ](https://github.com/openfoodfacts/openfoodfacts-server/issues/9793) of automatically unselecting images when source image is deleted.
fix:  First checked if the images that have the "imgid" field equal to the id of the image deleted, then called the process_image_unselect.

### DIFF
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -1181,6 +1181,13 @@ sub process_image_move ($user_id, $code, $imgids, $move_to, $ownerid) {
 						result => $ok
 					}
 				);
+
+            # Unselect images with matching imgid
+        	foreach my $selected_imgid (keys %{$product_ref->{images}}) {
+          		if ($selected_imgid eq $imgid) {
+            			process_image_unselect($user_id, $code, $selected_imgid);
+          			}
+        		}
 			}
 
 			# Don't delete images to be moved if they weren't moved correctly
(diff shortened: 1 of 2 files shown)


alexgarel:
@himanshi-154 you have to use Conventional Commits for your PR title.

Think also of those reviewing the list of PRs, a descriptive title is better than a number ! (still put a reference to the issue in the description)

lib/ProductOpener/Images.pm
```diff
@@ -1181,6 +1181,13 @@ sub process_image_move ($user_id, $code, $imgids, $move_to, $ownerid) {
 						result => $ok
 					}
 				);
+
+            # Unselect images with matching imgid
+        	foreach my $selected_imgid (keys %{$product_ref->{images}}) {
+          		if ($selected_imgid eq $imgid) {
```
stephanegigandet, line 1187: Hi @himanshi-154 , what we want is to unselect entries that have the "imgid" subfield equal, to $imgid. Here you are unselecting the source image, not the selected images.

I added some steps in the issue description so that you can verify if the issue is solved: https://github.com/openfoodfacts/openfoodfacts-server/issues/9793#issuecomment-1994223630

PR #9909, opened 2024-03-13
fix: remove JS script loaded twice

stephanegigandet:
I made a mistake and loaded the product-history script twice, which caused a JS error.


alexgarel (approved):
:+1:

PR #9912, opened 2024-03-13
fix: added example row to the existing spreadsheet

### LINKED ISSUE #9874: Add an example line to the import spreadsheet template on the producers platform
This issue is part of #9872 , please see the parent issue for context and guidance on how to work on this issue.

We want to add 2 lines to the import template (generated by https://world.pro.openfoodfacts.org/cgi/generate_sample_import_file.pl ):

- a description line that should contain what we currently have in the comments in the header line (and then we can remove the comments)
- an example line

This issue is about adding the example line.

We also need add one first column that just contains "description" and "example" for those 2 lines, and nothing for the the other rows.

See the template proposal: https://docs.google.com/spreadsheets/d/148PWIsSfDRsj8gLJ7cHbQtdapIEscUwX/edit#gid=1221199414 :

We will need to change the importing so that rows that contain something in the first column are skipped.

Examples should be translated in different languages.
https://fr.pro.openfoodfacts.org/cgi/generate_sample_import_file.pl should have examples in French.

The examples translations are stored in a TSV file in GitHub: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/conf/pro-platform/Import%20template%20-%20Example%20translations%20-%20Import%20sheet.tsv

The TSV file has the same columns as the English template, and one row per language, with the language code in the first column. That way it will be easy to edit the examples, add more languages etc. 

When generating the import template, we will need to:

- load the conf/pro-platform/Import%20template%20-%20Example%20translations%20-%20Import%20sheet.tsv TSV file
- select the line with language corresponding to the language of the import template we are generating ( $lc contains the language code). If there isn't a matching language (e.g. Italian is requested, but we only have English and French: use English).
- For each column of the generated import template, check if we have a corresponding column (same header name) in the example file. We cannot rely in the order of the columns, because it's possible that we will be adding / removing some columns. e.g. we will soon generate a "simple" version of the import template with less columns.

Shruti-S0405:
### What

I've added an example line in the already existing template by editing `cgi/generate_sample_import_file.pl`. 

### Screenshot
image

- Fixes #9874


stephanegigandet (changes requested):
Hi @Shruti-S0405 , thank you! This works, but we want to do things a bit differently:
- the example line should be read from the .tsv file, not harcoded in the .pl file
- the example line depends on the language (If I generate the import template in French, it should have French examples)
- the columns should be matched using the header name as they may not be in the same order than the template

I added more details in the issue:

Examples should be translated in different languages.
https://fr.pro.openfoodfacts.org/cgi/generate_sample_import_file.pl should have examples in French.

The examples translations are stored in a TSV file in GitHub: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/conf/pro-platform/Import%20template%20-%20Example%20translations%20-%20Import%20sheet.tsv

The TSV file has the same columns as the English template, and one row per language, with the language code in the first column. That way it will be easy to edit the examples, add more languages etc.

When generating the import template, we will need to:

load the conf/pro-platform/Import%20template%20-%20Example%20translations%20-%20Import%20sheet.tsv TSV file
select the line with language corresponding to the language of the import template we are generating ( $lc contains the language code). If there isn't a matching language (e.g. Italian is requested, but we only have English and French: use English).
For each column of the generated import template, check if we have a corresponding column (same header name) in the example file. We cannot rely in the order of the columns, because it's possible that we will be adding / removing some columns. e.g. we will soon generate a "simple" version of the import template with less columns.

TheSussex:
Hi @alexgarel I believe you can close this PR since the issue has been resolved.

stephanegigandet:
Thanks @Shruti-S0405 , @TheSussex has fixed the issue in https://github.com/openfoodfacts/openfoodfacts-server/pull/9975 so I'm closing this PR.

PR #9918, opened 2024-03-14
fix: add some user info in /cgi/auth.pl body 

### LINKED ISSUE #9910: Return user information in auth.pl if sent with a specific parameter (eg. ?info=json)
We need to know some information like:

* is a user a moderator
* what is it's preferred language

for services like open-prices / nutripatrol / hungergames

It should be quite easy to add this functionality to auth.pl so that if there is a ?info=json parameter, we would return a json with useful information (maybe avoid revealing email for now)

stephanegigandet:
fixes #9910

### DIFF
--- a/cgi/auth.pl
+++ b/cgi/auth.pl
@@ -55,6 +55,11 @@ if (defined $User_id) {
 		user => {
 			email => $User{email},
 			name => $User{name},
+			initial_lc => $User{initial_lc},
+			preferred_language => $User{preferred_language},
+			country => $User{country},
+			moderator => $User{moderator} ? 1 : 0,
+			admin => is_admin_user($User_id) ? 1 : 0,
 		},
 	};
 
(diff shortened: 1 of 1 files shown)


cgi/auth.pl
```diff
@@ -55,6 +55,11 @@
 		user => {
 			email => $User{email},
```
alexgarel, line 56: Shan't we remove it then ?

alexgarel (approved):
Cool.

cgi/auth.pl
stephanegigandet: @alexgarel It might be used by the mobile app, I'm not really sure... https://github.com/openfoodfacts/openfoodfacts-server/pull/7455

stephanegigandet: I'm removing the email and asking about it in https://github.com/openfoodfacts/openfoodfacts-dart/issues/893#issuecomment-1999390759

PR #9919, opened 2024-03-14
fix: nginx config changes for prod

stephanegigandet:
Notes:

- each site conf had a include /etc/nginx/snippets/expires-no-json-xml.conf which is in fact server wide, it's now moved to the conf.d directory
- the whole conf.d directory is symlinked to the conf.d dir in GitHub (instead of having symlinks for each file in it)

root@off:/etc/nginx# ls -lrt | grep conf.d
drwxr-xr-x 2 root root    3 Aug 17  2023 conf.d.old
lrwxrwxrwx 1 root root   26 Mar 14 15:01 conf.d -> /srv/off/conf/nginx/conf.d

- changed the way we determine which log file to use (static-access-log or proxy-access-log) using location instead of file extensions. This is because we can't use variables in log file names + buffers:

```
-nginx: [emerg] buffered logs cannot have variables in name in /etc/nginx/sites-enabled/off:36
nginx: configuration file /etc/nginx/nginx.conf test failed
access_log /var/log/nginx/${productopener_access_file_prefix}off-access.log proxied_requests buffer=256K flush=1s;
```

### DIFF
--- a/conf/nginx/conf.d/expires-no-json-xml.conf
+++ b/conf/nginx/conf.d/expires-no-json-xml.conf
@@ -1,0 +1,1 @@
+../snippets/expires-no-json-xml.conf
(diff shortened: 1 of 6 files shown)


alexgarel (approved):
Great.

Maybe we should change docker config akin to prod (I try to keep as close as possible).

alexgarel (approved):
Thanks !

PR #9931, opened 2024-03-15
fix: docker nginx config

stephanegigandet:
[Yesterday's changes for nginx prod](https://github.com/openfoodfacts/openfoodfacts-server/pull/9919) broke nginx docker, harmonizing.


alexgarel (approved):
Thanks !

PR #9945, opened 2024-03-18
fix: reorder CSV export and compression

stephanegigandet:
We currently generate CSV and RDF exports without compressing them right away, in between we do MongoDB exports, which can take 4 hours.

This PR is to compress the CSV / RDF exports right away.

Today's timestamps:

```
-rw-r--r--   1 off  off   9167563154 Mar 18 05:42 en.openfoodfacts.org.products.csv
-rw-r--r--   1 off  off   4675878115 Mar 18 05:42 en.openfoodfacts.org.products.rdf
-rw-r--r--   1 off  off     18676807 Mar 18 07:54 fr.openfoodfacts.org.products.bad-chars.log
-rw-r--r--   1 off  off   9340189179 Mar 18 08:07 fr.openfoodfacts.org.products.csv
-rw-r--r--   1 off  off   4675987404 Mar 18 08:08 fr.openfoodfacts.org.products.rdf
-rw-r--r--   1 off  off   7311036936 Mar 18 09:36 openfoodfacts-products.jsonl.gz
-rw-r--r--   1 off  off   8859132762 Mar 18 10:28 openfoodfacts-mongodbdump.gz
-rw-r--r--   1 off  off           95 Mar 18 10:30 gz-sha256sum
-rw-r--r--   1 off  off           63 Mar 18 10:30 gz-md5sum
-rw-r--r--   1 off  off   8932229671 Mar 18 11:42 openfoodfacts-mongodbdump.tar.gz
-rw-r--r--   1 off  off           99 Mar 18 11:44 sha256sum
-rw-r--r--   1 off  off           67 Mar 18 11:45 md5sum
drwxr-xr-x   2 off  off           17 Mar 18 11:45 delta
-rw-r--r--   1 off  off    949672708 Mar 18 12:16 openfoodfacts_recent_changes.jsonl.gz
-rw-r--r--   1 off  off    441327604 Mar 18 12:19 en.openfoodfacts.org.products.rdf.gz
-rw-r--r--   1 off  off    441316211 Mar 18 12:21 fr.openfoodfacts.org.products.rdf.gz
-rw-r--r--   1 off  off    892120182 Mar 18 12:27 en.openfoodfacts.org.products.csv.gz
-rw-r--r--   1 off  off    902549626 Mar 18 12:33 fr.openfoodfacts.org.products.csv.gz
```

In a separate PR, I plan to remove the openfoodfacts-mongodbdump.tar.gz export, it makes no sense to export the exact same data twice in different formats (and on top of that we also have a 3rd format with the jsonl.. )


scripts/gen_feeds_daily_obf.sh
alexgarel, line 1: @stephanegigandet you are aware that it does not change anything on obf right now ?

scripts/gen_feeds_daily_off.sh
```diff
@@ -11,8 +11,9 @@ export PERL5LIB="../lib:${PERL5LIB}"
 ./remove_empty_products.pl
 ./gen_top_tags_per_country.pl
 #./compute_missions.pl
+
+# Generate CSV exports
```
alexgarel, line 15: maybe move up also gziping csv vs gzipping rdf (less used) ?

alexgarel (approved):
LGTM

scripts/gen_feeds_daily_off.sh
stephanegigandet: Good idea, done.

CharlesNepote (approved):
Thanks a lot @stephanegigandet.
CSV gzip takes nearly 8 minutes on my computer. Due to the facts that it is running at night, and that we have many big files to compress, we might use parallel gzip (such as pigz) working on several threads.

scripts/gen_feeds_daily_off.sh
foodietim: As gzip is not really multithreaded, would it make sense to either move to a compression command that allows using multiple cores (such as [pigz](https://zlib.net/pigz/)) or maybe spin off multiple gzip processes into the background (eg, by using `&` beside the gzip commands)?

Files compressed with pigz are compatible with the gzip command.

stephanegigandet: @foodietim we could, but the compression takes only 5 minutes per file max. One thing we should do is to compress to a new file and move it once compressed, in order to avoid someone downloading a broken file during those 5 minutes.

CharlesNepote (approved):
Fine!

scripts/gen_feeds_daily_off.sh
```diff
@@ -11,18 +11,21 @@ export PERL5LIB="../lib:${PERL5LIB}"
...
-./mongodb_dump.sh /srv/off/html openfoodfacts 10.1.0.102 off
 
 cd $OFF_PUBLIC_DATA_DIR
-gzip < en.openfoodfacts.org.products.rdf > en.openfoodfacts.org.products.rdf.gz
-gzip < fr.openfoodfacts.org.products.rdf > fr.openfoodfacts.org.products.rdf.gz
-
-gzip < en.openfoodfacts.org.products.csv > en.openfoodfacts.org.products.csv.gz
-gzip < fr.openfoodfacts.org.products.csv > fr.openfoodfacts.org.products.csv.gz
+for export in en.openfoodfacts.org.products.csv fr.openfoodfacts.org.products.csv en.openfoodfacts.org.products.rdf fr.openfoodfacts.org.products.rdf; do
+   nice pigz < $export > new.$export.gz
```
alexgarel, line 20: you should add apt install pigz in Dockerfile as for me it also serve as a documentation of needed packages.

stephanegigandet: @alexgarel Done, thanks!

PR #9946, opened 2024-03-18
fix: keep only one (or two?) mongodb export

stephanegigandet:
We currently generate 3 MongoDB exports, each of them takes more than one hour to generate every day:

```
-rw-r--r--   1 off  off   4675987404 Mar 18 08:08 fr.openfoodfacts.org.products.rdf
-rw-r--r--   1 off  off   7311036936 Mar 18 09:36 openfoodfacts-products.jsonl.gz
-rw-r--r--   1 off  off   8859132762 Mar 18 10:28 openfoodfacts-mongodbdump.gz
-rw-r--r--   1 off  off           95 Mar 18 10:30 gz-sha256sum
-rw-r--r--   1 off  off           63 Mar 18 10:30 gz-md5sum
-rw-r--r--   1 off  off   8932229671 Mar 18 11:42 openfoodfacts-mongodbdump.tar.gz
```

I suggest that we completely remove openfoodfacts-mongodbdump.tar.gz which has exactly the same usage as openfoodfacts-mongodbdump.gz

I'm not sure if we should keep the jsonl. As I recall, it was added for Robotoff. Do we still need it? Or do we still want to provide it to others, so that's it's possible to get JSON data without having MongoDB?

### DIFF
--- a/docs/dev/how-to-quick-start-guide.md
+++ b/docs/dev/how-to-quick-start-guide.md
@@ -272,7 +272,7 @@ Use the Git Bash shell to run the make commands in windows so that programs like
 When running `make import_prod_data`.
 
 ```console
-process_begin: CreateProcess(NULL, wget --no-verbose https://static.openfoodfacts.org/data/openfoodfacts-mongodbdump.tar.gz, ...) failed.
+process_begin: CreateProcess(NULL, wget --no-verbose https://static.openfoodfacts.org/data/openfoodfacts-mongodbdump.gz, ...) failed.
 make (e=2): The system cannot find the file specified.
 ```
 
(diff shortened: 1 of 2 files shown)


stephanegigandet:
See also https://github.com/openfoodfacts/openfoodfacts-server/pull/9946

We will need to update the documentation to remove references to the removed dumps.

hangy:
Should we put up a deprecation notice for a month or so before removing this export?

alexgarel (approved):
I think it's ok. We had both for about one year so I think we can move on !

alexgarel:
@stephanegigandet before merging, maybe, amend the data page + be sure to remove the old archive in production (better have a 404 than serving an old archive forever).

stephanegigandet:
@alexg : "before merging, maybe, amend the data page + be sure to remove the old archive in production (better have a 404 than serving an old archive forever)."

Date page update PR: https://github.com/openfoodfacts/openfoodfacts-web/pull/563

I'll remove the old archive when this is deployed.

PR #9950, opened 2024-03-18
feat: Automatically unselect images when source image deleted Issue#9793

himanshi-154:
**##What?**
Issue#9793 that states that images should be automatically unselected when their source image is deleted

**# Why?**
This saves time for contributors on problematic images, and avoid forgetting to unselect it.

**# How?**
First checked if selected image id is equal to image id and then deleted the image id as the siurce image is already deleted

### DIFF
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -1213,6 +1213,12 @@ sub process_image_move ($user_id, $code, $imgids, $move_to, $ownerid) {
 
 				delete $product_ref->{images}{$imgid};
 
+				# Check if any selected image's imgid equals the deleted imgid
+				foreach my $selected_imgid (keys %{$product_ref->{images}}) {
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Images.pm
```diff
@@ -1213,6 +1213,12 @@ sub process_image_move ($user_id, $code, $imgids, $move_to, $ownerid) {
 
 				delete $product_ref->{images}{$imgid};
 
+				# Check if any selected image's imgid equals the deleted imgid
```
stephanegigandet, line 1216: We want to check if the value of the "imgid" field of a selected field is equal to $imgid.

To verify the problem, see steps: https://github.com/openfoodfacts/openfoodfacts-server/issues/9793#issuecomment-1994223630

stephanegigandet:
Hello,

PR #9952, opened 2024-03-19
fix: Add org name directly in url for a logged in user on producer platform

### LINKED ISSUE #9685: Add /org/[organisation name] facet directly in URL on the producers platform
Today we have an hidden facet on the producers platform for the organization name. It would be much better to make that facet directly visible in the URL (like other facets).

This would allow:
- org admins to easily access different organizations
- producers to have access to several organizations (e.g. for sub brands)
- sharing URLs that specify the organization

TheSussex:
### What

In the "cgi/login.pl" file, I created a url_suffix variable that concatenates organisation id with existing url if user has an organisation id and uses the existing url if user has not updated their account with organisation details. This url is displayed once the user is logged in.

This will help users see at first glance their organisation name.

- Fixes #9685

### DIFF
--- a/cgi/login.pl
+++ b/cgi/login.pl
@@ -47,11 +47,18 @@ my $r = shift;
 my $redirect = single_param('redirect');
 $template_data_ref->{redirect} = $redirect;
 if (defined $User_id) {
-	my $loc = $redirect || $formatted_subdomain . "/cgi/session.pl";
-	$r->headers_out->set(Location => $loc);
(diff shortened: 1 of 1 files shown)


stephanegigandet:
@TheSussex Thanks for the PR, but the issue is about something really different, it's something that I will have to do as it requires lots of changes internally.

TheSussex:
Oh okay then.

PR #9954, opened 2024-03-19
fix: Add timestamp in log4perl logs

### LINKED ISSUE #9840: Add timestamp in log4perl
There are time stamps in perl error but not in log4perl. This is annoying when trying to understand a past event…

TheSussex:
### What

I update `conf/log.conf` file by adding timestamp to the ConversionPattern for both LOGFILE and MONGODB_LOGFILE appenders. This ensures each log has specific timestamp and makes debugging easier during development.

### Screenshot
Screenshot 2024-03-19 at 10 28 12 AM

- Fixes #9840


alexgarel (approved):
Great.

PR #9956, opened 2024-03-19
fix: run off-pro through docker

stephanegigandet:
Somes changes to try to get "make dev" work after running 

The sftp folder is created in order to avoid having an error at startup:

`Log::Any::Proxy=HASH(0x55a4ed4a7fe0)[Tue Mar 19 13:09:54.189902 2024] [perl:error] [pid 17] FATAL: Some important directories are missing: /mnt/podata/sftp at /opt/product-opener/lib/startup_apache2.pl line 159.\nCompilation failed in require at (eval 2) line 1.\n`

And we skip the import_sample_data


hangy (approved):

Makefile
```diff
@@ -205,9 +205,13 @@ refresh_product_tags:
 	${DOCKER_COMPOSE} run --rm backend perl /opt/product-opener/scripts/refresh_postgres.pl ${from}
 
 import_sample_data:
-	@echo "🥫 Importing sample data (~200 products) into MongoDB …"
-	${DOCKER_COMPOSE} run --rm backend bash /opt/product-opener/scripts/import_sample_data.sh
-
+	@ if [ "${COMPOSE_PROJECT_NAME}" = "po_off" ]; then \
```
alexgarel, line 208: I have a settings where I use po2 as prefix (to work in parallel on two branches), so it won't work for me.

Maybe:

```suggestion
	@ if [ "${PRODUCT_OPENER_FLAVOR_SHORT}" = "off" &&  "${PRODUCERS_PLATFORM}" != "1" ]; then \
```

alexgarel (approved):
Thanks !

I have a suggestion to fix the Makefile to support multiple prefixes.

Makefile
alexgarel: (I tested it locally)

PR #9957, opened 2024-03-19
fix: Add lang tag to import template title

TheSussex:
### What

I changed the title of the import template spreadsheet from `openfoodfacts_import.xlsx` to `openfoodfacts_import_template_en.xlsx`

The latter title is more descriptive and can help producers differentiate between files especially if they download in multiple languages.


stephanegigandet (approved):
Thanks!

PR #9960, opened 2024-03-19
fix: Added remove tags from field function 

### LINKED ISSUE #9453: Add the possibility to remove a tag from taxonomized fields
Cleanly removing a tag in taxonomized field is currently a mess:
* we need to use `xxx_hierarchy` (eg `labels_hierarchy`) to construct what you send in the labels field -- because `xxx_hierarchy` are non-normalized but with language prefixes
* then we need to join the tags in a comma separated list
* then change whichever string you want
* then send it back in `xxx` field, making sure all entries have language prefixes

## Need
The need is to do this simply in one API call: eg. `remove_categories=something&add_categories=whatever`

anaritadauane:
### What
Added a function to remove a tag/tags from taxonomized fields

I achieved this by: 
- Added a function remove_tags_from_fields, a function that removes tags from fields
- Updated the update_tags_fieds function, now a field can be updated by removing or adding tags
- Creating a test case to remove tags from a field

- Fixes #9453

### DIFF
--- a/tests/integration/expected_test_results/api_v3_product_write/patch-tags-fields-remove.json
+++ b/tests/integration/expected_test_results/api_v3_product_write/patch-tags-fields-remove.json
@@ -1,0 +1,99 @@
+{
+   "code" : "1234567890100",
+   "errors" : [],
+   "product" : {},
+   "status" : "success_with_warnings",
+   "warnings" : [
+      {
+         "field" : {
+            "default_value" : "en",
+            "id" : "tags_lc"
+         },
+         "impact" : {
+            "id" : "warning",
+            "lc_name" : "Warning",
+            "name" : "Warning"
+         },
+         "message" : {
+            "id" : "missing_field",
+            "lc_name" : "Missing field",
+            "name" : "Missing field"
+         }
+      },
+      {
+         "field" : {
+            "id" : "categories_tags_remove",
+            "value" : [
+               "en:tea"
+            ]
+         },
+         "impact" : {
+            "id" : "warning",
+            "lc_name" : "Warning",
+            "name" : "Warning"
+         },
+         "message" : {
+            "id" : "unrecognized_field",
+            "lc_name" : "Unrecognized field",
+            "name" : "Unrecognized field"
+         }
+      },
+      {
+         "field" : {
+            "id" : "countries_tags_fr_remove",
+            "value" : [
+               "Italie",
+               "en:spain"
+            ]
+         },
+         "impact" : {
+            "id" : "warning",
+            "lc_name" : "Warning",
+            "name" : "Warning"
+         },
+         "message" : {
+            "id" : "unrecognized_field",
+            "lc_name" : "Unrecognized field",
+            "name" : "Unrecognized field"
+         }
+      },
+      {
+         "field" : {
+            "id" : "labels_tags_remove",
+            "value" : [
+               "végétarien",
+               "Something unrecognized in French"
+            ]
+         },
+         "impact" : {
+            "id" : "warning",
+            "lc_name" : "Warning",
+            "name" : "Warning"
+         },
+         "message" : {
+            "id" : "unrecognized_field",
+            "lc_name" : "Unrecognized field",
+            "name" : "Unrecognized field"
+         }
+      },
+      {
+         "field" : {
+            "id" : "stores_tags_remove",
+            "value" : [
+               "Carrefour",
+               "Mon Ptit magasin"
+            ]
+         },
+         "impact" : {
+            "id" : "warning",
+            "lc_name" : "Warning",
+            "name" : "Warning"
+         },
+         "message" : {
+            "id" : "unrecognized_field",
+            "lc_name" : "Unrecognized field",
+            "name" : "Unrecognized field"
+         }
+      }
+   ]
+}


lib/ProductOpener/APIProductWrite.pm
```diff
@@ -213,9 +213,15 @@ sub update_tags_fields ($request_ref, $product_ref, $field, $add_to_existing_tag
 		);
 	}
 	else {
-		# Generate a comma separated list of tags, so that we can use existing functions to add tags
+		# Generate a comma separated list of tags, so that we can use existing functions to add tags or remove tags
 		my $tags_list = join(',', @$value);
 
+		if ($remove_tags && ref($remove_tags) eq 'ARRAY') {
+			foreach my $tag (@$remove_tags) {
+				remove_tags_from_field($product_ref, $tags_lc, $field, $tag);
```
alexgarel, line 221: Didn't you mispelled it ?
```suggestion
				remove_tag_from_field($product_ref, $tags_lc, $field, $tag);
```

lib/ProductOpener/Tags.pm
```diff
@@ -4613,6 +4613,49 @@ sub add_tags_to_field ($product_ref, $tag_lc, $field, $additional_fields) {
 	return;
 }
 
+sub remove_tag_from_field($product_ref, $tag_lc, $field, $tags_to_remove) {
```
alexgarel, line 4616: You except a list of tag to remove so:
* use plural in function name (remove_tags_from_field)
* your last parameter should be a reference to that list

```suggestion
sub remove_tags_from_field($product_ref, $tag_lc, $field, $tags_to_remove_ref) {
```

you can then get the list with:
```perl
my @tags_to_remove = @$tags_to_remove_ref;
```

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -213,9 +213,15 @@ sub update_tags_fields ($request_ref, $product_ref, $field, $add_to_existing_tag
...
+			}
```
alexgarel, line 222: There is a mistmacht between what you do here and you implementation which expect a list reference (which is the right way I think).

lib/ProductOpener/Tags.pm
```diff
@@ -4613,6 +4613,49 @@ sub add_tags_to_field ($product_ref, $tag_lc, $field, $additional_fields) {
...
+			$tagid = get_taxonomyid($tag_lc, canonicalize_taxonomy_tag($tag_lc, $field, $tag));
+		}
+		else {
+			$tagid = get_string_id_for_lang($tag_lc, $tag);
+		}
+		if (exists $existing{$tagid}) {
+			delete $existing{$tagid};
+			push @removed_tags, $tag;
+		}
+	}
```
alexgarel, line 4645: Instead of this I propose this implementation to you:
* create a hashmap of tags to remove (as you did with existing)
* get the list of tags (and keep it's length for reference)
* use [grep](https://perldoc.perl.org/functions/grep) to filter the list of tags
* update the field if the list length as changed

tests/integration/api_v3_product_write.t
```diff
@@ -662,6 +662,21 @@ my $tests_ref = [
...
+		body => '{
+			"fields" : "updated",
+			"product": { 
+				"categories_tags_remove": ["en:tea"],
+				"stores_tags_remove": ["Carrefour", "Mon Ptit magasin"],
+				"countries_tags_fr_remove": ["Italie", "en:spain"],
+				"labels_tags_remove": ["végétarien", "Something unrecognized in French"]
+			}
+		}',
+	},
```
alexgarel, line 679: Really good that you add an integration test.

But it would be great to also add some unit tests (in tags.t) to test remove_tag_from_field (you product ref can be a hashmap with only needed fields) so that you can test more case: 
* regular case (removing 2 tags in a list of 4)
* removing values that are not in the list (should not change)
* empty list of tags to remove
* removing all tags

alexgarel (commented):
You did quite well for an issue that is not the easiest !

Here are some inputs to let you continue.

lib/ProductOpener/Tags.pm
alexgarel: Sorry ignore this comment, I read a bit quickly and ignored some of the complexity that you instead handled well !
Your implementation is better.

lib/ProductOpener/Tags.pm
alexgarel: Here again I misread… you have a string which is a comma separated list of values (maybe it's a good indication that you should document the function ;-) )

So appart from the function name, it's all ok :-)

```suggestion
sub remove_tags_from_field($product_ref, $tag_lc, $field, $tags_to_remove) {
```

tests/integration/api_v3_product_write.t
alexgarel: Also add unit test case where you use a tagid and a value for a tag, etc.
It's really worth it.

anaritadauane: Thanks! 
I've been following the api logic and found Stéphane's tips really helpful. I'll make the necessary modifications to the code right away

lib/ProductOpener/APIProductWrite.pm
anaritadauane: I don't understand, are you saying this because of the for loop or because the $tag variable is an element of the reference list?

tests/integration/api_v3_product_write.t
anaritadauane: I made some changes, but I am a bit confused about the **is_deeply** and **is** methods. I think the **is** method is more appropriate for the remove_tags_from_field unit tests

alexgarel: **is** is for scalar values

**is_deeply** for the others (to compare two arrays or two hashmaps etc.)

tests/unit/tags.t
```diff
@@ -622,6 +622,41 @@ is_deeply(
...
 
+# test remove_tags_from_fields
+
+# removing 2 tags in a list of 4
+$product_ref = {
+	lc => "fr",
+	categories => "pommes, bananes, en:pears, fr:fraises, es:limones",
+};
+
+remove_tag_from_field($product_ref, "fr", "categories", "bananes, fraises");
```
alexgarel, line 633: you forgot to change the name here.

Better run your tests locally before committing:
* make test-unit test=tags.t

(and make test-int test=youfile.t for the integration)
See https://openfoodfacts.github.io/openfoodfacts-server/dev/how-to-write-and-run-tests/

tests/unit/tags.t
```diff
@@ -622,6 +622,41 @@ is_deeply(
...
+is_deeply($product_ref->{categories}, "pommes, oranges");
```
alexgarel, line 634: This is a scalar, so you should use **is**

alexgarel: Why would you expect "pommes, oranges" here ? I would expect "pommes, en:pears, es:limones")

lib/ProductOpener/APIProductWrite.pm
alexgarel: Your code is working but you are doing one call per field, while it could be grouped (or am I missing something ?)

Why not:
```suggestion
			remove_tags_from_field($product_ref, $tags_lc, $field, join(',', @$remove_tags));
			
```

tests/integration/api_v3_product_write.t
alexgarel, line 679: also you should have a json file resulting from this test. I let you see the documentation https://openfoodfacts.github.io/openfoodfacts-server/dev/how-to-use-gitpod/

alexgarel (commented):
Some more comments !

lib/ProductOpener/APIProductWrite.pm
anaritadauane: Yes, you are right. I did that because it is already handled by the remove_tags_from_field method

anaritadauane: It is one call per tag to be removed

tests/unit/tags.t
anaritadauane: Yes, it should! It was a mistake.

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -196,7 +196,7 @@ Update packagings.
 
 =cut
 
-sub update_tags_fields ($request_ref, $product_ref, $field, $add_to_existing_tags, $tags_lc, $value) {
+sub update_tags_fields ($request_ref, $product_ref, $field, $add_to_existing_tags, $remove_tags, $tags_lc, $value) {
```
stephanegigandet, line 199: If you change the parameters of a n existing function, you need to look for all the uses of that function to update the function calls to send the right parameters.

anaritadauane: Thanks, I made the changes but I'm having trouble generating a json file with the expected results. I'm running this command `make test-unit="api_v3_product_write.t --update-expected-results"`

tests/integration/api_v3_product_write.t
anaritadauane: I have made a few changes to the code and want to test it, but I'm having trouble getting a json file with the expected results. I'm running this command: `make test-unit="api_v3_product_write.t --update-expected-results"`

anaritadauane: **

**

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -217,9 +217,7 @@ sub update_tags_fields ($request_ref, $product_ref, $field, $add_to_existing_tag
 		my $tags_list = join(',', @$value);
 
 		if ($remove_tags && ref($remove_tags) eq 'ARRAY') {
-			foreach my $tag (@$remove_tags) {
-				remove_tags_from_field($product_ref, $tags_lc, $field, $tag);
-			}
+			remove_tags_from_field($product_ref, $tags_lc, $field, $remove_tags);    #join(',', @$remove_tags)
```
alexgarel, line 220: you have a ref to an array but then in remove_tags_from_field you treat it like a string… it wont work ! Why don't you use a join ?

tests/integration/api_v3_product_write.t
```diff
@@ -673,7 +673,7 @@ my $tests_ref = [
 				"categories_tags_remove": ["en:tea"],
 				"stores_tags_remove": ["Carrefour", "Mon Ptit magasin"],
 				"countries_tags_fr_remove": ["Italie", "en:spain"],
-				"labels_tags_remove": ["végétarien", "Something unrecognized in French"]
```
alexgarel: aren't you testing tags removal ?

tests/unit/tags.t
```diff
@@ -627,34 +627,36 @@ is_deeply(
...
-	categories => "pommes, bananes, en:pears, fr:fraises, es:limones",
+	categories => "pommes, bananes, pears, fraises",
 };
 
-remove_tag_from_field($product_ref, "fr", "categories", "bananes, fraises");
-is_deeply($product_ref->{categories}, "pommes, oranges");
+# my @tags_to_remove = split(/,\s*/, "bananes, fraises");
+
+# remove_tags_from_field($product_ref, "fr", "categories", @tags_to_remove);
+remove_tags_from_field($product_ref, "fr", "categories", "bananes, fraises");
```
alexgarel, line 636: it's not the way you function work… please read about parameters in functions, and lists in perl. You are messing things.

lib/ProductOpener/APIProductWrite.pm
anaritadauane: Because I'm calling the join function inside the remove_tags_from_field function

tests/integration/api_v3_product_write.t
anaritadauane: Yes, I am

tests/unit/tags.t
anaritadauane: You mean this way `remove_tags_from_field($product_ref, "fr", "categories", "bananes, fraises");`?

tests/integration/api_v3_product_write.t
anaritadauane: Yes I can run the website locally

tests/unit/tags.t
alexgarel: Sorry I think it's me who misread :blush:

lib/ProductOpener/APIProductWrite.pm
alexgarel: oooh, this is `make test-unit test="api_v3_product_write.t --update-expected-results"`

anaritadauane: Yes, I corrected the code but still 
I am going to try one more time

tests/integration/expected_test_results/api_v3_product_write/patch-tags-fields-remove.json
alexgarel, line 1: I pushed it, but as you see it's not what we expect.

It's important that you are able to run tests locally otherwise it will be very complicated to fix your PR.

lib/ProductOpener/APIProductWrite.pm
alexgarel: @anaritadauane changed a bit the command (we have to add :: now because of yath the new  test runner)

.DS_Store
hangy, line 1: I think this is some Apple-specific file? Should this be in `.gitignore` or is it actually useful?

aleene: Yes that is an apple file related to the directory.

stephanegigandet:
Hi @anaritadauane The PR is inactive, so closing. Please  reopen if you still want to work on it. Thank you!

PR #9964, opened 2024-03-19
refactor: Tags.pm: do not build taxonomies on the fly / at module load - remove special case for nutrient_levels.txt

stephanegigandet:
This PR is to simplify taxonomy building and loading, to remove potential chicken and egg problems (to generate the nutrient_levels.txt source file), and to remove potential race conditions when multiple processes try to load and build taxonomies in parallel.

Changes:
- taxonomies/nutrient_levels.txt is now generated only on demand, with script/taxonomies/generators/gen_nutrient_levels_taxonomy.pl : it is commited in the GitHub repository, and we can generate it from time to time (when the nutrients.txt taxonomy changes, or when we have new translations). It does not have to be up to date, this taxonomy is used only for display, so it does not matter if it is slightly out of date.
- the above removes the need for handling nutrient_levels.txt as a special case
- new init_taxonomies() function, which runs when the Tags.pm is loaded (loads available taxonomies only, skips unavailable ones) or when load_data() is called (e.g. in startup script), in a mode that dies if some needed taxonomies are not there
- when loading taxonomies, they are not automatically built on the fly, this is error prone and can lead to race conditions. Whenever we deploy a new version, we should build taxonomies once.
- initialization code in modules like Food.pm, EcoScore.pm, GS1.pm etc. moved to dedicated init_fuctions
- added a die() statement in canonicalize_taxonomy_tag if it is called for a taxonomy that is not loaded
- some code fixes


alexgarel:
Did you forget the script in scripts/taxonomies/generators/ ?

alexgarel:
Also we should eventually change the apache startup script so that it call every init_ that needs to be called to avoid calling them again and again in children process.

alexgarel (changes requested):
I think we miss the script :-)

Makefile
```diff
@@ -246,7 +246,7 @@ checks: front_build front_lint check_perltidy check_perl_fast check_critic
 lint: lint_perltidy
 # TODO: add lint_taxonomies when taxonomies ready
 
-tests: build_lang_test unit_test integration_test
+tests: build_lang_test build_taxonomies_test unit_test integration_test
```
alexgarel, line 249: Oooooh good catch, now that we stores built taxonomies as data, this is needed…

lib/ProductOpener/Ecoscore.pm
```diff
@@ -1127,14 +1127,24 @@ my @production_system_labels = (
...
+		$label_ref->[0] = canonicalize_taxonomy_tag("en", "labels", $label_ref->[0]);
+	}
+	$production_system_labels_initialized = 1;
 }
 
 sub compute_ecoscore_production_system_adjustment ($product_ref) {
 
+	if (not $production_system_labels_initialized) {
+		init_production_system_labels();
+	}
```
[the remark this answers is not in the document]
stephanegigandet: Sounds good, thanks :)

stephanegigandet:
> Also we should eventually change the apache startup script so that it call every init_ that needs to be called to avoid calling them again and again in children process.

it's done through the ProductOpener::LoadData::load_data() function called from the startup script (and other scripts)

Makefile
```diff
@@ -95,7 +95,7 @@ edit_etc_hosts:
 create_folders:
 # create some folders to avoid having them owned by root (when created by docker compose)
 	@echo "🥫 Creating folders before docker compose use them."
-	mkdir -p logs/apache2 logs/nginx debug html/data || ( whoami; ls -l . ; false )
+	mkdir -p logs/apache2 debug html/data || ( whoami; ls -l . ; false )
```
alexgarel, line 98: We must not remove logs/nginx

It might be that we must add create_folders as a dependency to another entry if you have a permission error on that.

scripts/taxonomies/generators/gen_nutrient_levels_taxonomy.pl
```diff
@@ -0,0 +1,32 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+use Modern::Perl '2017';
+use utf8;
+use strict;
```
alexgarel, line 25: Why not `use ProductOpener::PerlStandards;` ?

alexgarel (approved):
Cool !

PR #9966, opened 2024-03-20
fix: Migrated the JS from template form to real JS

### LINKED ISSUE #9106: Convert JS in Perl template into real JS

MonalikaPatnaik:
Fixes #9106 
Migrated the code of togglepassword.tt.js and user_form.tt.js scripts to /html/js/ as static JS files.


stephanegigandet (approved):
Looks good, thank you @MonalikaPatnaik !

MonalikaPatnaik:
@stephanegigandet could you pls merge this pull request.
Thanks!

templates/web/pages/user_form/user_form_page.tt.html
```diff
@@ -29,9 +29,8 @@
 
     inputfield.value = value;
     }
-    [% INCLUDE web/pages/user_form/togglepassword.tt.js %]
 </script>
-
+<script type="text/javascript" src="/js/togglepassword.js"></script>
```
hangy, line 33: Could you add both JS scripts to [gulp](https://github.com/openfoodfacts/openfoodfacts-server/blob/34a46b891c1a7195a5d424cc991d309c8f269b31/gulpfile.ts#L13), so that they're properly minified for deployment. You'll have to load them from `/js/dist/togglepassword.js` to use that effect.

hangy: We also sometimes use `static_subdomain` to ensure the JS files are loaded from a canonical URL: https://github.com/openfoodfacts/openfoodfacts-server/blob/34a46b891c1a7195a5d424cc991d309c8f269b31/templates/web/common/site_layout.tt.html#L409

MonalikaPatnaik: Hi @hangy, yes sure, added now👍
Thanks!

templates/web/pages/user_form/user_form_page.tt.html
```diff
@@ -257,4 +255,6 @@ <h3>[% lang("you_can_also_help_us") %]</h3>
     <p>type: [% type %] action: [% action %] userid: [% userid %]</p>
 [% END %]
 
+<script type="text/javascript" src="/js/user_form.js"></script>
```
hangy, line 258: This include is probably not necessary any more with https://github.com/openfoodfacts/openfoodfacts-server/blob/c1ef2354207a96fb31015a58bbb783cadf092841/templates/web/common/site_layout.tt.html#L417 being in place?

hangy (approved):

templates/web/common/site_layout.tt.html
```diff
@@ -411,6 +411,8 @@ <h3 class="title-5 text-medium">[% lang('footer_discover_the_project') %]</h3>
 <script src="[% static_subdomain %]/js/dist/display.js"></script>
 <script src="[% static_subdomain %]/js/dist/stikelem.js"></script>
 <script src="[% static_subdomain %]/js/dist/scrollNav.js"></script>
+<script src="[% static_subdomain %]/js/dist/togglepassword.js"></script>
```
stephanegigandet, line 414: Those 2 scripts are only needed on the sign-in and user page, can you load them only on those pages instead of all pages of the website?

stephanegigandet (changes requested):
I updated the tests results and saw that the new scripts are loaded on all pages, I think it would make more sense to load them only on  the sign in page and user edit page.

tests/integration/expected_test_results/web_html/user-register.html
```diff
@@ -1639,33 +1620,12 @@ <h3 class="title-5 text-medium">Discover the project</h3>
 <script src="//static.openfoodfacts.localhost/js/dist/display.js"></script>
 <script src="//static.openfoodfacts.localhost/js/dist/stikelem.js"></script>
 <script src="//static.openfoodfacts.localhost/js/dist/scrollNav.js"></script>
+<script src="//static.openfoodfacts.localhost/js/dist/togglepassword.js"></script>
+<script src="//static.openfoodfacts.localhost/js/dist/user_form.js"></script>
 
 <script>
 $(function() {
 
-$(document).ready( function () {
```
stephanegigandet, line 1646: I think this should not have been removed: when I check the "this is a pro account" checkbox, I'm not shown the warning anymore:

image

This is what we have in production:

image

stephanegigandet:
I'm closing this PR as we are going to replace all the user management system by an external system (Keycloak).

PR #9967, opened 2024-03-21
refactor: templatize display_product_api from "Display.pm"

deveshidwivedi:
### What
Templatizing `display_product_api` method from `Display.pm`

- Issue #6927

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -10346,55 +10346,9 @@ sub display_product_api ($request_ref) {
 		}
 		$response{status} = 0;
 		$response{status_verbose} = 'product not found';
-		if (single_param("jqm")) {
-			$response{jqm} = <<HTML
-$Lang{app_please_take_pictures}{$lang}
-<button onclick="captureImage();" data-icon="off-camera">$Lang{app_take_a_picture}{$lang}</button>
-<div id="upload_image_result"></div>
-<p>$Lang{app_take_a_picture_note}{$lang}</p>
-HTML
-				;
-			if ($request_ref->{api_version} >= 0.1) {
-
-				my @app_fields = qw(product_name brands quantity);
-
-				my $html = <<HTML
-<form id="product_fields" action="javascript:void(0);">
-<div data-role="fieldcontain" class="ui-hide-label" style="border-bottom-width: 0;">
-HTML
-					;
-				foreach my $field (@app_fields) {
-
-					# placeholder in value
-					my $value = $Lang{$field}{$lang};
-
-					$html .= <<HTML
-<label for="$field">$Lang{$field}{$lang}</label>
-<input type="text" name="$field" id="$field" value="" placeholder="$value">
-HTML
-						;
-				}
-
-				$html .= <<HTML
-</div>
-<div id="save_button">
-<input type="submit" id="save" name="save" value="$Lang{save}{$lang}">
-</div>
-<div id="saving" style="display:none">
-<img src="loading2.gif" style="margin-right:10px"> $Lang{saving}{$lang}
-</div>
-<div id="saved" style="display:none">
-$Lang{saved}{$lang}
-</div>
-<div id="not_saved" style="display:none">
-$Lang{not_saved}{$lang}
-</div>
-</form>
-HTML
-					;
+#template
 				$response{jqm} .= $html;
 
-			}
 		}
 	}
 	else {
--- a/templates/web/common/includes/display_product_api.tt.html
+++ b/templates/web/common/includes/display_product_api.tt.html
@@ -1,0 +1,32 @@
+
+[% IF single_param("jqm") %]
+    [%- INCLUDE 'path/to/lang_file.tt.html' -%]
+    <button onclick="captureImage();" data-icon="off-camera">[% Lang.app_take_a_picture.$lang %]</button>
+    <div id="upload_image_result"></div>
+    <p>[% Lang.app_take_a_picture_note.$lang %]</p>
+[% END %]
+
+[% IF request_ref.api_version >= 0.1 %]
+    <form id="product_fields" action="javascript:void(0);">
+        <div data-role="fieldcontain" class="ui-hide-label" style="border-bottom-width: 0;">
+            [% FOREACH field IN app_fields %]
+                [% value = Lang.$field.$lang %]
+                <label for="[% field %]">[% Lang.$field.$lang %]</label>
+                <input type="text" name="[% field %]" id="[% field %]" value="" placeholder="[% value %]">
+            [% END %]
+        </div>
+
+        <div id="save_button">
+            <input type="submit" id="save" name="save" value="[% Lang.save.$lang %]">
+        </div>
+        <div id="saving" style="display:none">
+            <img src="loading2.gif" style="margin-right:10px"> [% Lang.saving.$lang %]
+        </div>
+        <div id="saved" style="display:none">
+            [% Lang.saved.$lang %]
+        </div>
+        <div id="not_saved" style="display:none">
+            [% Lang.not_saved.$lang %]
+        </div>
+    </form>
+[% END %]


deveshidwivedi:
Hi @alexgarel, I am trying to work on this task but need a little clarity on a few things. I will be glad if you could give an idea regarding-
- Now that we are creating a separate template for this method, in what form are we going to pass it in `Display.pm`? For example, in [this PR](https://github.com/openfoodfacts/openfoodfacts-server/pull/7003),  we define `my $template_data_ref` with different keys, will we follow a similar approach here too?
- There are different HTML code bits in the method, which I have put in the same .tt.html file currently, so how will we point to these code bits in the original .pm file?
Thank you!

stephanegigandet:
> Hi @alexgarel, I am trying to work on this task but need a little clarity on a few things. I will be glad if you could give an idea regarding-
> 
> * Now that we are creating a separate template for this method, in what form are we going to pass it in `Display.pm`? For example, in [this PR](https://github.com/openfoodfacts/openfoodfacts-server/pull/7003),  we define `my $template_data_ref` with different keys, will we follow a similar approach here too?
> * There are different HTML code bits in the method, which I have put in the same .tt.html file currently, so how will we point to these code bits in the original .pm file?
>   Thank you!

Hi @deveshidwivedi , you need to create a $template_data_ref with all the data / variables that the template will need. Then to generate the html from this data + the template, you need to call process_template()

You can have a look at other parts of the Display.pm module that do it, for instance:

```
					process_template('web/pages/org_profile/org_profile.tt.html',
						$user_template_data_ref, \$profile_html)
						or $profile_html
						= "<p>web/pages/org_profile/org_profile.tt.html template error: " . $tt->error() . "</p>";
```

deveshidwivedi:
Thank you @stephanegigandet! Could you please take a look at the recent changes?

lib/ProductOpener/Display.pm
```diff
@@ -10324,6 +10324,14 @@ sub display_product_api ($request_ref) {
 
 	my $product_ref;
 
+	my $html;
+
+	my $template_data_ref = {
+		lang => $lang,
+		Lang => \%Lang,
+		request_ref => $request_ref,
```
alexgarel, line 10332: it might be better to only pass api_version value.

templates/web/common/includes/display_product_api.tt.html
```diff
@@ -0,0 +1,32 @@
...
+    [% Lang.app_please_take_pictures.$lang %]
+    <button onclick="captureImage();" data-icon="off-camera">[% Lang.app_take_a_picture.$lang %]</button>
+    <div id="upload_image_result"></div>
+    <p>[% Lang.app_take_a_picture_note.$lang %]</p>
+[% END %]
+
+[% IF request_ref.api_version >= 0.1 %]
+    <form id="product_fields" action="javascript:void(0);">
+        <div data-role="fieldcontain" class="ui-hide-label" style="border-bottom-width: 0;">
+            [% FOREACH field IN app_fields %]
```
alexgarel, line 12: I think you missed app_fields in template_data_ref

alexgarel (commented):
Seems almost good, but did you test it (I mean in a browser) ?

deveshidwivedi:
> Seems almost good, but did you test it (I mean in a browser) ?

I did try to test it and other than the initial compilation errors (now resolved), there weren't any errors. I am able to look for products using codes. Are there more ways to test?

deveshidwivedi:
@stephanegigandet @alexgarel could you please review the changes?

lib/ProductOpen
alexgarel, line 1: @deveshidwivedi you must not rename the module !

deveshidwivedi: I am sorry, I am not sure when I did this mistake but I'll take care next time

alexgarel:
@deveshidwivedi I am terribly sorry but while trying to test, it appeared to me that this html variable is not used any more.

@stephanegigandet it seems dead code to me, can you confirm ?

@deveshidwivedi it would mean we should remove the code generating this piece of html, and no need of a template !

deveshidwivedi:
I understand @alexgarel

alexgarel:
@deveshidwivedi would you like to however submit a PR where you just remove that dangling code ? (it's very useful to remove unused code !).

deveshidwivedi:
Yes I'll do that! Is it just the bit that I tried to templatize from this method that we are talking about?

stephanegigandet:
@alexgarel @deveshidwivedi It's not dead code, it's broken code. (I broke it a while ago I guess)

You can execute it with URLs like this: (from the old OFF app)

http://world.openfoodfacts.localhost/api/v0/product/3661344653573/yaourt-cremeuh-cafe-les-2-vaches.jqm  (with .jqm at the end)

but currently we have an error:

Undefined subroutine &ProductOpener::Display::data_to_display_nutriscore_and_nutrient_levels called at /srv/off/lib/ProductOpener/Display.pm line 8514.

That being said, I do hope to be able to remove most of the jqm (jquery mobile) API soon...

stephanegigandet:
I fixed the error with jquery mobile and pushed the fix to this PR.

templates/web/common/includes/display_product_api.tt.html
```diff
@@ -0,0 +1,32 @@
+
+[% IF single_param("jqm") %]
+    [% Lang.app_please_take_pictures.$request_lc %]
+    <button onclick="captureImage();" data-icon="off-camera">[% Lang.app_take_a_picture.$request_lc %]</button>
+    <div id="upload_image_result"></div>
+    <p>[% Lang.app_take_a_picture_note.$request_lc %]</p>
+[% END %]
+
+[% IF request_ref.api_version >= 0.1 %]
```
alexgarel, line 9: It should be api_version directly now !

```suggestion
[% IF api_version >= 0.1 %]
```

lib/ProductOpener/Display.pm
```diff
@@ -10181,56 +10193,10 @@ sub display_product_api ($request_ref) {
...
-<div id="saved" style="display:none">
-$Lang{saved}{$request_lc}
-</div>
-<div id="not_saved" style="display:none">
-$Lang{not_saved}{$request_lc}
-</div>
-</form>
-HTML
-					;
-				$response{jqm} .= $html;
```
alexgarel, line 10230: you have to keep this line @deveshidwivedi ! (otherwise $html is not used !)

templates/web/common/includes/display_product_api.tt.html
```diff
@@ -0,0 +1,32 @@
+
+[% IF single_param("jqm") %]
```
alexgarel, line 2: `single_param` function is not available in templates @deveshidwivedi, so better compute the value in Display.pm and add the value it as a parameter (eg. has_jqm_param).

templates/web/common/includes/display_product_api.tt.html
```diff
@@ -0,0 +1,32 @@
...
+    <div id="upload_image_result"></div>
+    <p>[% Lang.app_take_a_picture_note.$request_lc %]</p>
+[% END %]
+
+[% IF api_version >= 0.1 %]
+    <form id="product_fields" action="javascript:void(0);">
+        <div data-role="fieldcontain" class="ui-hide-label" style="border-bottom-width: 0;">
+            [% FOREACH field IN app_fields %]
+                [% value = Lang.$field.$request_lc %]
+                <label for="[% field %]">[% Lang.$field.$request_lc %]</label>
```
alexgarel, line 14: The Lang.$field.$request_lc does not work, because you forgot to pass %Lang as parameter !

alexgarel (commented):
See my comments above @deveshidwivedi 

So to make it clear: this is worthfull to refactor.

But your current code is not yet working (see my comments).

You can test it by accessing a non existing product, with v1 (to get edit fields) and v0 (without edit fields).

Eg. 
* http://world.openfoodfacts.localhost/api/v1/product/3661344653574/not-existing.jqm
* http://world.openfoodfacts.localhost/api/v0/product/3661344653574/non-existing.jqm

alexgarel (approved):
Great ! Thank you @deveshidwivedi

PR #9970, opened 2024-03-21
fix: canonicalize food_group field

stephanegigandet:
for some reason the tests have values like "en:cheese" for the "food_groups" field, but we generate "en:Cheese". I don't understand why we had en:cheese before and why it recently changed.

This PR canonicalizes the food_groups field, so that we do get "en:cheese".


alexgarel (approved):
Let's do it !

PR #9971, opened 2024-03-21
fix: Fix ecoscore origins

stephanegigandet:
Fix for food_groups + fix for Eco-Score origins

The test was failing probably because the origins:en property for the category en:calvados changed. I'm not sure how the corresponding change was merged without the test failing though.


alexgarel (approved):
LGTM

PR #9972, opened 2024-03-21
chore: week12 taxonomy updates

aleene:
### What
various requests and imrpovements


benbenben2 (approved):
Super!
Thank you @aleene

PR #9975, opened 2024-03-22
fix: Add example line to import template

### LINKED ISSUE #9874: Add an example line to the import spreadsheet template on the producers platform

TheSussex:
### What

- I added format options for the example row
- I imported a .tsv file that contains example values for the import template columns and populated the columns with values based on the requested language.

- Fixes #9874


cgi/generate_sample_import_file.pl
```diff
@@ -59,13 +59,35 @@
...
+chomp($header_line);
+my @headers = split("\t", $header_line);
+while (my $line = <$example_fh>) {
+    chomp($line);
+    my @values = split("\t", $line);
+    for my $i (0 .. $#headers) {
+        next if $headers[$i] eq 'lc';
+        $example_values_by_header{$headers[$i]} = $values[$i] if defined $values[$i];
+    }
+    last;
```
stephanegigandet, line 88: Since you have "last;", you are only reading one line (the first one: the one for English)

stephanegigandet: You should read all the lines, and populate a structure that contains example values for different languages. First you read the "lc" into $line_lc, and then you can create a structure $example_values_by_language_and_header{$line_lc} = {}, and you populate the headers in it.

TheSussex:
I have updated the code as suggested but the behaviour is still not as expected. When I select french as the language, some fields like Ingredients and labels columns get populated with the french values but other columns remain empty even though there are french examples for them. 

@stephanegigandet

cgi/generate_sample_import_file.pl
```diff
@@ -169,18 +191,8 @@
...
-			my $example_title = lang("example");
-
-			# Several examples?
-			if ($example =~ /,/) {
-				$example_title = lang("examples");
-			}
-			$comment .= $example_title . " " . $example . "\n\n";
-		}
+		my $line_lc = lc($request_ref->{lc} || 'en');
+    my $example_value = $example_values_by_language_and_header{$line_lc}{$field_ref->{text}} // '';
```
stephanegigandet, line 195: The issue is that $field_ref->{text} can be in another language than English, but in the sample configuration file we have English.
The best solution is to add a row with field ids in the example configuration file. I made a PR based on yours here: https://github.com/openfoodfacts/openfoodfacts-server/pull/9985

stephanegigandet:
> I have updated the code as suggested but the behaviour is still not as expected. When I select french as the language, some fields like Ingredients and labels columns get populated with the french values but other columns remain empty even though there are french examples for them.
> 
> @stephanegigandet

@TheSussex I think the best way to solve it is to add a row with field ids to the examples configuration spreadsheet:
https://github.com/openfoodfacts/openfoodfacts-server/pull/9985

alexgarel:
@TheSussex thanks a lot for the PR, finally we merged https://github.com/openfoodfacts/openfoodfacts-server/pull/9985 that @stephanegigandet built upon your PR.

TheSussex:
> @TheSussex thanks a lot for the PR, finally we merged #9985 that @stephanegigandet built upon your PR.

Thank you so much @stephanegigandet @alexgarel

I have gone through Stephane's fix and I understand now how the changes to the example spreadsheet helped solve the problem. Open source contributing is so educating and interesting!😊

PR #9981, opened 2024-03-23
fix: GitPod clarification

teolemon:
### What
fix: GitPod clarification

### DIFF
--- a/docs/dev/how-to-use-gitpod.md
+++ b/docs/dev/how-to-use-gitpod.md
@@ -9,7 +9,6 @@ of setting up local environments and IDEs with Perl, Docker and plugins, making 
 OpenFoodFacts Server to get started in minutes instead of hours!
 
 
-
 Note that while this how-to is tailored for Gitpod, using alternatives like [GitHub Codespaces][github-codespaces] should be
 similar.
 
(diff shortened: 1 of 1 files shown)


docs/dev/how-to-use-gitpod.md
```diff
@@ -44,17 +45,18 @@ running on the remote machine is not accessible on your local network interface.
...
 
-Once you execute the altered command in your terminal, you should be able to access OpenFoodFacts
+Once you execute the altered command in your terminal, you should be able to access Open Food Facts
 on http://openfoodfacts.localhost just as documented in the quickstart guide!
 
 [gitpod-ssh-guide]: https://www.gitpod.io/blog/copy-paste-ssh-workspace-access
 [github-codespaces]: https://github.com/features/codespaces
 
 **Remark:** for some Linux distributions, the port 80 is reserved. A workaround is to switch to port 8080: in gitpod, open the .env file and replace the line PRODUCT_OPENER_PORT=80 by PRODUCT_OPENER_PORT=8080, then replace -L 80:localhost:80 by -L 8080:localhost:8080. **Rollback the changes on .env before to make a pull request!***  
+**Remark:** on MacOS, when trying to bind, you might see a bind fail. Remember that anyport below 1024 is reserved for the sudo user. You might want to add sudo to the ssh intruction. It will then prompt you whether you you want to trust the fingerprint.
```
alexgarel, line 59: We need a line return to separate paragraphs
```suggestion

**Remark:** you will need to trust the fingerprint at first connection. 

**Remark:** on MacOS or Linux, when trying to bind, you might see a bind fail. Remember that any port below 1024 is reserved for the root user. You might want to add `sudo` just before the ssh command.
```

alexgarel (approved):
LGTM with a suggestion, I will commit now :-)

PR #9987, opened 2024-03-23
fix: drop Encode::Punycode (no longer used)

eseyman:
It appears ProductOpener::Store does not use Encode::Punycode to normalize strings but uses the tr and s instructions.


stephanegigandet (approved):
Thank you!

PR #9989, opened 2024-03-23
test: POC for running the unit test suite with Test2::Harness

smonff:
### What

Makes possible to [preload specified modules once](https://metacpan.org/pod/App::Yath#PRELOAD-MODULES) and not for each test:

```
yath -PProductOpener::Tags -POtherModule tests/unit/*
```

We want to check that the test suite can be run with `Test2::Suite`. It's not supposed to be a problem as Test2::Suite is compatible with `Test::More`.  

From the [Perl / Open Food Facts Paris Hackathon](https://forum.openfoodfacts.org/t/hackathon-perl-open-food-facts-in-paris/530), no specific ticket.


alexgarel:
@smonff, first thanks for the experimentation.
 
We are at 22 minutes instead of 18 minutes to run tests.

Apart for perfs there might be some advantages to go for yath ? What would you say ?

smonff:
Sounds weird about the timing. When I tried it in the dev container with the `-j` option, it drastically reduced the run time.

About the other advantages, I'd recommend to take a look at this Chad Granum's presentations:

https://youtu.be/7lNtM2FmTow

He also gave previous presentations explaining other aspects (easy to find on YouTube or possibly on his GitHub account).

I'll try to summarize a couple of ideas here.

Performances are great to have but only a small part of the improvements. Test2 is an unified test toolkit, available directly in Perl's core, redesigned from decades long Perl excellence in testing.

Something Test2 improve a lot is human-readability of the reports and diagnoses in case of test failures. It was already great before, but it's even clearer now.

E.g. `Test2::Tools::Compare` provide improved diagnostics compared to `cmp_ok`. 

It also provide a rspec like toolkit (`Test2::Tools::Spec`), allowing to describe, run "before" and "after" tasks, etc. Not sure it was available in core modules before.

yath provide a much more "human readable" output than prove. For "machine readability", it writes a log file (JSON, I guess). The log file allow to "replay" tests (!).

yath also provides a web-ui if necessary.

Looks like the future of Perl tests.

alexgarel:
@smonff ok, I totally buy you arguments, I will fix the PR (it's messy right now, I don't know why) and merge it.

alexgarel (dismissed):

stephanegigandet (approved):
Looks good to me, thank you @smonff and @alexgarel!

Makefile
```diff
@@ -279,19 +280,21 @@ test-stop:
...
+# you can use TEST_CMD to change test command, like TEST_CMD="perl -d" to debug a test
+# you can also add args= to pass more options to your test command
 test-unit: guard-test create_folders
 	@echo "🥫 Running test: 'tests/unit/${test}' …"
 	${DOCKER_COMPOSE_TEST} up -d memcached postgres mongodb
-	${DOCKER_COMPOSE_TEST} run --rm backend perl ${args} tests/unit/${test}
+	${DOCKER_COMPOSE_TEST} run --rm -e PO_EAGER_LOAD_DATA=1 backend ${TEST_CMD} ${args} tests/unit/${test}
 
 # usage:  make test-int test=test-name.t
 # to update expected results: make test-int test="test-name.t --update-expected-results"
```
stephanegigandet, line 291:
```suggestion
# to update expected results: make test-int test="test-name.t :: --update-expected-results"
```

stephanegigandet:
I made a small update to the comment in the Makefile for updating the results of one test:

# to update expected results: make test-int test="test-name.t :: --update-expected-results"

(we now need :: before the test arguments)

PR #9990, opened 2024-03-23
docs: light copyediting: docs/dev/explain-packaging-data.md

pauamma:
Change to "Only the value for the main product of the language is currently analyzed." as discussed with stephanebiz.


stephanegigandet (approved):
Thanks!

PR #9991, opened 2024-03-23
refactor: remove old ingredients classes for additives, ingredients from palm oil

stephanegigandet:
This PR removes some old code that was used to display product pages before we introduced knowledge panels.
We displayed in different columns ingredients that were in specific classes: additives (now in their own panel), vitamins, minerals, nucleotides, ingredients from palm oil and ingredients that may be from palm oil (now in their own panel).


alexgarel (approved):
Great !

PR #9993, opened 2024-03-23
ci: compile perl modules faster for checks

alexgarel:


alexgarel:
This is not the right way to do it… I close it.

PR #10003, opened 2024-03-24
docs: factcheck and copyedit docs/dev/how-to-use-gitpod.md

pauamma:
Note on the "500 credits" bit:
It's no longer present in https://www.gitpod.io/pricing that I can see, so (after discussion with Alex G.) I removed it. Feel free to re-add it if I missed it.


alexgarel (approved):
Thank you so much !

PR #10008, opened 2024-03-24
refactor: remove $lang global variable

stephanegigandet:
For historical reasons, Lang.pm exports 2 global variables $lang and $lc that in theory should always contain the same value.
In the different modules, cgi and scripts, we have been using one or the other interchangeably.

This PR:
- removes the $lang global variable, and keep only $lc (which contains a language code like "en" or "fr")
- in some functions that have $request_ref as a parameter, it replaces the $lang global variable by $request_ref->{lc}

The eventual goal will be to also remove the $lc global variable at some point.

Note that in some files we still have a $lang variable that is declared locally (and that masked the global $lang variable).


lib/ProductOpener/Display.pm
```diff
@@ -1492,7 +1488,7 @@ sub display_mission ($request_ref) {
 
 	my $missionid = $request_ref->{missionid};
 
-	open(my $IN, "<:encoding(UTF-8)", "$BASE_DIRS{PUBLIC_DATA}/missions/$lang/$missionid.html");
+	open(my $IN, "<:encoding(UTF-8)", "$BASE_DIRS{PUBLIC_DATA}/missions/" . $request_ref->{lc} . "/$missionid.html");
```
alexgarel, line 1491: FYI this works also:
```suggestion
	open(my $IN, "<:encoding(UTF-8)", "$BASE_DIRS{PUBLIC_DATA}/missions/$request_ref->{lc}/$missionid.html");
```

alexgarel (approved):
Cool ! That was a big refactor :astonished:

PR #10022, opened 2024-03-25
fix: sort_by param in the current url being mixed up with "?" instead of "&"

### LINKED ISSUE #9392: Sort by button in search results does not work
### What
- After searching products in the search box, the sort by dropdown does not work:
  - Results are not sorted
  - There is a blank value in the dropdown

Resulting url has 2 sort_by fields, with the last one with a ? instead of & : 

http://world.openfoodfacts.localhost/cgi/search.pl?action=process&search_terms=coca&sort_by=unique_scans_n&page_size=24?sort_by=last_modified_t

image

Reported by Olaf Görlitz on Slack:
https://openfoodfacts.slack.com/archives/CUZK65DGE/p1700749241876539 (includes video)

### Part of
- #5492

himanshisrestha:
### What
After searching for a product, and then applying a sort_by filter, the resulting url contained  2 sort_by params and "?" instead of "&"
- A new subroutine add_params($url, $param) was added to the Web.pm which checks whether we have a "?" in the current url and if yes then concatenates the sort_by param with "&" else with "?"
 
 The urls now looks like this:
- http://world.openfoodfacts.localhost/?sort_by=nutriscore_score&page_size=100
- http://world.openfoodfacts.localhost/cgi/search.pl?action=process&search_terms=sugar&sort_by=nutriscore_score&page_size=24

- Fixes #9392


alexgarel:
https://github.com/openfoodfacts/openfoodfacts-server/pull/10024 is the active PR.

PR #10023, opened 2024-03-25
refactor: add integration tests that check the HTML content of different types of pages

stephanegigandet:
I'm adding those tests so that it's easier to spot errors that could go unnoticed when we refactors big parts of the code (e.g. when [removing the SiteLang directory,](https://github.com/openfoodfacts/openfoodfacts-server/pull/9888) or [removing the $lang global variable](https://github.com/openfoodfacts/openfoodfacts-server/pull/10008)).


hangy:
As per today's meeting: Maybe something like [`Test::HTML::Differences`](https://metacpan.org/pod/Test::HTML::Differences) could help in checking the diff

tests/integration/expected_test_results/web_html/fr-decouvrir.html
alexgarel, line 1: Isn't this one in openfoodfacts-web ?
If so it won't work in integration tests, you have to add the file you want to use for tests in web-default/lang.

stephanegigandet: Good catch, I'll remove it

stephanegigandet:
> As per today's meeting: Maybe something like [`Test::HTML::Differences`](https://metacpan.org/pod/Test::HTML::Differences) could help in checking the diff

Thanks, I've been trying a bunch of modules to test differences, but I'm not getting very good results. Test::Differences (without ::HTML) works but still outputs the whole HTML file, I couldn't find a way to get only the context.

https://metacpan.org/pod/Test::File::Contents gives only context but I couldn't get it to work well yet, I'll keep trying

damil:
> > As per today's meeting: Maybe something like [`Test::HTML::Differences`](https://metacpan.org/pod/Test::HTML::Differences) could help in checking the diff
> 
> Thanks, I've been trying a bunch of modules to test differences, but I'm not getting very good results. Test::Differences (without ::HTML) works but still outputs the whole HTML file, I couldn't find a way to get only the context.
> 
> https://metacpan.org/pod/Test::File::Contents gives only context but I couldn't get it to work well yet, I'll keep trying

I don't know if this can help, but for info here is my local test script used to check differences in panel.tt.html. Not polished, but it can be an inspiration
```

use 5.38.0;
use utf8;
use strict;
use warnings;
use JSON;
use YAML;
use Path::Tiny;
use Template;
use Test::More;
use Text::Diff;


my $data = decode_json(path("sample.json")->slurp_raw);

my $template_orig = 'web/panels/panel.tt.ORIG.html';
my $template_new  = 'web/panels/panel.tt.html';



my $tt = Template->new(
	{
		INCLUDE_PATH => '../../../templates',
		INTERPOLATE => 1,
		EVAL_PERL => 1,
		STAT_TTL => 60,    # cache templates in memory for 1 min before checking if the source changed
		COMPILE_EXT => '.ttc',    # compile templates to Perl code for much faster reload
		COMPILE_DIR => "/home/off/tmp/templates",
		ENCODING => 'UTF-8',
	}
);

my $panels = $data->{product}{knowledge_panels};

foreach my $name (sort keys %$panels) {
	my $content = $panels->{$name};

	my $output_orig = "";
	$tt->process($template_orig, {panel_id => $name, panels => $panels}, \ $output_orig) or die $tt->error;

	my $output_new = "";
	$tt->process($template_new, {panel_id => $name, panels => $panels}, \ $output_new) or die $tt->error;

	# remove non-significant differences before doing the diff
	$output_orig =~ s/\.ORIG//g;
	s/\n(\h*\n)+/\n/g for $output_new, $output_orig;

	# suppress initial and final spaces
	s/^\h+//mg 		   for $output_new, $output_orig;
	s/\h+$//mg 		   for $output_new, $output_orig;
	s{<!-- end templates/web/panels/panel.tt.html - panel_id: .*? -->}{<!-- NO DIFF ON END COMMENT}g for $output_new, $output_orig;


	$output_orig =~ s{<h3  class}{<h3 class}g;
	$output_orig =~ s{class="([^"]+)*"}{'class="' . replace_nl_in_class($1) . '"'}eg;

	$output_new =~ s{(class="[^"]+)\h+"}{$1"}g;
	$output_new .= "\n";



	my $diff = diff \$output_new, \$output_orig;

	ok !$diff, $name;
	if ($diff) {
		note $diff;
		$name =~ s/:/_/g;
		path("diffs/$name.orig.html")->spew_utf8($output_orig);
		path("diffs/$name.new.html")->spew_utf8($output_new);
	}
}

done_testing;


sub replace_nl_in_class {

	my $txt = shift;
	$txt =~ s/\n\h*/ /g;
	$txt =~ s/ $//;
	return $txt;
}

```

tests/integration/expected_test_results/api_v3_product_write/options-code-test.html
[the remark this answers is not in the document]
stephanegigandet: I'll change the type

stephanegigandet:
Replaced by https://github.com/openfoodfacts/openfoodfacts-server/pull/10043

PR #10024, opened 2024-03-25
fix: sort_by param issue in the url with "?" instead of "&" 

### LINKED ISSUE #9392: Sort by button in search results does not work

himanshisrestha:
### What
After searching for a product, and then applying a sort_by filter, the resulting url contained 2 sort_by params and "?" instead of "&"

- A new subroutine add_params($url, $param) was added to the Web.pm which checks whether we have a "?" in the current url and if yes then concatenates the sort_by param with "&" else with "?"
The urls now looks like this:

- http://world.openfoodfacts.localhost/?sort_by=nutriscore_score&page_size=100
- http://world.openfoodfacts.localhost/cgi/search.pl?action=process&search_terms=sugar&sort_by=nutriscore_score&page_size=24

- Fixes #9392

### DIFF
--- a/cgi/search.pl
+++ b/cgi/search.pl
@@ -665,11 +665,11 @@ elsif ($action eq 'process') {
 		}
 	}
 
-	if (defined $sort_by) {
-		$current_link .= "&sort_by=$sort_by";
-	}
+	# if (defined $sort_by) {
+	# 	$current_link .= "&sort_by=$sort_by";
+	# }
 
-	$current_link .= "\&page_size=$limit";
+	# $current_link .= "\&page_size=$limit";
 
 	# Graphs
 
--- a/lib/ProductOpener/Web.pm
+++ b/lib/ProductOpener/Web.pm
@@ -60,7 +60,8 @@ BEGIN {
 		&display_knowledge_panel
 		&get_languages_options_list
 		&get_countries_options_list
-	);    #the fucntions which are called outside this file
+		&add_params
+	);    #the functions which are called outside this file
 	%EXPORT_TAGS = (all => [@EXPORT_OK]);
 }
 
@@ -376,6 +377,35 @@ sub get_countries_options_list ($target_lc, $exclude_world = 1) {
 		@countries_list = grep {$_->{value} ne "world"} @countries_list;
 	}
 	return \@countries_list;
+}
+
+=head2 add_params( $url, $params )
+
+Adds parameters to a URL while properly handling the existing query string.
+
+=head3 Arguments
+
+=head4 $url - The URL to which parameters need to be added.
+
+=head4 $params - The parameters to be added in the format "key=value".
+
+=head3 Return value
+
+The modified URL with added parameters.
+
+=cut
+
+sub add_params ($url, $params) {
+
+	# Check if the URL already contains parameters
+	if ($url =~ /\?/) {
+		$url .= "&$params";    # Append using '&' if parameters already exist
+	}
+	else {
+		$url .= "?$params";    # Append using '?' if no parameters exist
+	}
+
+	return $url;
 }
 
 1;
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -5002,13 +5002,13 @@ sub search_and_display_products ($request_ref, $query_ref, $sort_by, $limit, $pa
 		push @{$template_data_ref->{sort_options}},
 			{
 			value => "popularity",
-			link => $request_ref->{current_link} . "?sort_by=popularity",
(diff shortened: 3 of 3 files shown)


cgi/search.pl
```diff
@@ -665,11 +665,11 @@
...
 
-	if (defined $sort_by) {
-		$current_link .= "&sort_by=$sort_by";
-	}
+	# if (defined $sort_by) {
+	# 	$current_link .= "&sort_by=$sort_by";
+	# }
 
-	$current_link .= "\&page_size=$limit";
+	# $current_link .= "\&page_size=$limit";
```
alexgarel, line 672: I'm not sure we want to remove it because we may let user sort on whatever they want through the API.

alexgarel: In search.pl we also have different sort possibilities.

So I would say: use add_param here also, and maybe verify first that sort_by or page_size is not already in current_link.

himanshisrestha:
@alexgarel I've tried to replicate but I'vent seen what you mentioned. 
The Permanent link from staging is here: 
https://world.openfoodfacts.net/cgi/search.pl?action=process&search_terms=sugar&sort_by=unique_scans_n&page_size=20

and the local host is here:
http://world.openfoodfacts.localhost/cgi/search.pl?action=process&search_terms=sugar&sort_by=created_t&page_size=20
(Here sort_by =created_t because in search.pl url I added a filter for that)

lib/ProductOpener/Web.pm
```diff
@@ -378,4 +379,33 @@ sub get_countries_options_list ($target_lc, $exclude_world = 1) {
...
+
+=head4 $params - The parameters to be added in the format "key=value".
+
+=head3 Return value
+
+The modified URL with added parameters.
+
+=cut
+
+sub add_params ($url, $params) {
```
stephanegigandet, line 398: Can you put a more specific name, like add_params_to_link?

lib/ProductOpener/Web.pm
```diff
@@ -378,4 +379,33 @@ sub get_countries_options_list ($target_lc, $exclude_world = 1) {
...
+
+	# Check if the URL already contains parameters
```
stephanegigandet, line 400: It would be good to check if the URL already contains the same parameter we want to add, in which case we want to remove the old one before adding the new one.

stephanegigandet: It would be good to add unit tests for this function with various cases: param already there, other params, no other params etc.

hangy: > It would be good to check if the URL already contains the same parameter we want to add, in which case we want to remove the old one before adding the new one.

Do we always explicitly limit it to one value when reading the parameters? [cgi](https://metacpan.org/pod/CGI#Fetching-the-value-or-values-of-a-single-named-parameter) can do arrays

himanshisrestha: @stephanegigandet where should the test cases be written .. Is it in a file in tests/unit or should I create a .t file there for the same ?

lib/ProductOpener/Web.pm
```diff
@@ -377,4 +378,33 @@ sub get_countries_options_list ($target_lc, $exclude_world = 1) {
...
+
+=head4 $params - The parameters to be added in the format "key=value".
+
+=head3 Return value
+
+The modified URL with added parameters.
+
+=cut
+
+sub add_params ($url, $params) {
```
alexgarel, line 397: @himanshisrestha we can even consider having


`sub add_params_to_link ($url, @params) {`

@params is a list of parameters you want to add.

This enables calling the function using:

`add_params_to_link($my_url, "test=foo", "othe=bar");` with as many parameters as we want.

lib/ProductOpener/Web.pm
alexgarel: You can create a tests/unit/web.t file.

alexgarel:
@himanshisrestha we would be glad if you can finish this PR !

himanshisrestha:
@alexgarel Sure I'm working on it !! 😄

PR #10026, opened 2024-03-25
taxonomy: ingredients for Contribalim

stephanegigandet:
Ingredients needed to get the Nutri-Score on https://contribalim.openfoodfacts.org


taxonomies/food/ingredients.txt
```diff
@@ -85784,6 +85793,15 @@ ko:쇼트브레드
 xx:shortbread
 zh:奶油酥餅
 wikidata:en:Q652695
+wikidata:en:Q2915670
```
hangy, line 85796: According to Wikidata, `Sablé (Q2915670) has part shortbread (Q652695)` so it might be better to split up the ingredients? Do we support duplicate keys at all?

stephanegigandet: @hangy we don't support duplicate keys, it was a mistake.

I added a Sablé entry separate from Shortbread

teolemon (approved):

hangy (approved):

teolemon:
@stephanegigandet merging

PR #10028, opened 2024-03-25
refactor: avoid recursion in loading panel.tt.html and divide the logic into blocks

damil:
### What
Here is a proposal to refactor panel.tt.html using more advanced features of the Template Toolkit.
- no longer INCLUDE _files_, bug INCLUDE _blocks_ -- this avoids  recursive loading of modules
- divide the logic into separate steps
- use wrappers when a use case needs a specific beginning and a specific end (instead of testing the same condition at both places)
- use the CHOMP features of TT2 to remove useless spaces between directives (using tildes or dashes at beginning or end of directives)
- use SWITCH / CASE when there were too many ELSIF

### Disclaimer
Unfortunately I don't have the infrastructure for doing full tests.
I just checked that the HTML produced by the old and new version are compatible, modulo space variations; but I'm not totally confident that all cases where compared, so there might be some minor fixes to do.

Also, it's a totally different way of doing TT2, so I hope it won't be too perturbing for other members of the team. But if you like this way of templating, you will see it brings benefits for understanding the structure of the page. If you don't like it ... well, don't merge, we could revert to a much simpler change by just inserting one single block for avoiding the recursion.

Last word : this change is quite radical, so the blame history will be perturbed.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -300,7 +300,6 @@ $tt = Template->new(
 		COMPILE_EXT => '.ttc',    # compile templates to Perl code for much faster reload
 		COMPILE_DIR => $data_root . "/tmp/templates",
 		ENCODING => 'UTF-8',
-		RECURSION => 1,    # Needed for the knowledge panels that contain subpanels
 	}
 );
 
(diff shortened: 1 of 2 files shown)


stephanegigandet:
Thanks @damil ! I would like to finish https://github.com/openfoodfacts/openfoodfacts-server/pull/10023 first so that we can compare the generated HTML.

alexgarel:
I close this as this is now in https://github.com/openfoodfacts/openfoodfacts-server/pull/10050

PR #10041, opened 2024-03-27
feat: added unit for sodium in India

### LINKED ISSUE #10038: In India, make the default unit for sodium mg instead of g on product edit form on web and mobile app
[Reported by krishanti on Slack](https://openfoodfacts.slack.com/archives/CDX0VQ0TS/p1711414340376119): "Please consider setting the default unit for Sodium as mg instead of grams as that is the unit commonly seen on nutrition labels. While editing products I've seen that sometimes the unit is mistakenly ignored and the nutriscore becomes inaccurate"

in the taxonomies/nutrients.txt, we can define default units for ingredients for a specific country. e.g. this states that for energy, the default unit is kj in all countries except Canada and the US:

unit:en: kj
unit_ca:en: kcal
unit_us:en: kcal

There is some code in product_multilingual.pl that then uses it to display the right unit on the product edit form on the web site:

```
		if (exists_taxonomy_tag("nutrients", "zz:$nid")) {
			$nutriment_ref->{name} = display_taxonomy_tag($lc, "nutrients", "zz:$nid");
			# We may have a unit specific to the country (e.g. US nutrition facts table using the International Unit for this nutrient, and Europe using mg)
			$unit = get_property("nutrients", "zz:$nid", "unit_$cc:en")
				// get_property("nutrients", "zz:$nid", "unit:en") // 'g';
		}
```

This code should be moved to a function in Food.pm, and it should also be called in cgi/nutrients.pl so that we get the right unit for countries. /cgi/nutrients.pl is used by the OFF mobile app.

### Part of
- #6540

himanshisrestha:
### What
- Added unit_in:en: mg key value pair to the nutrients taxonomy
- Made a new subroutine in Food.pm named get_nutrient_unit

I'vent been able to see effects that the changes made 

- Fixes #10038


stephanegigandet:
> I'vent been able to see effects that the changes made

You will see the changes when you edit a product (the default unit should be mg for sodium), and when you view the product page.

Make sure to run "make build_taxonomies" and "make restart" first.

Sodium is automatically converted to salt, so you should put the same unit for salt.

Can you try and see if it works as expected?

Also, could you look in Food.pm, there is a table that defines which nutrients are listed by default on nutrition facts table for specific countries: %nutriments_tables = (

Could you try to add a table for India, so that it shows the nutrients that are printed on food packages in India? (for instance sodium instead of salt, but also added sugars, cholesterol etc.)

himanshisrestha:
Sure.. I'll try and update you soon

himanshisrestha:
Screenshot 2024-03-28 at 9 52 52 PM

So as expected, I'm getting this in the edit product of a mango drink from India whose dummy data and couple of pictures I was adding to test. 
Also I've added the table for India as suggested in the %nutriments_tables

lib/ProductOpener/Food.pm
```diff
@@ -87,6 +87,8 @@ BEGIN {
 
 		&check_nutriscore_categories_exist_in_taxonomy
 
+		&get_nutrient_unit
```
hangy, line 90: I think it would be great if you could add a unit test in [food.t](https://github.com/openfoodfacts/openfoodfacts-server/blob/bf084bbe8ce263cea2eee7f53dcae634fcf1f2eb/tests/unit/food.t)

stephanegigandet (approved):
It looks good, thank you @himanshisrestha 

As @hangy suggests, could you add a simple test for the get_nutrient_unit function in food.t ?

Maybe one test for India to get the unit_in, and one test for a country for which we don't have a unit, to check that we get the default unit.

himanshisrestha:
Sure ... Working on it !!

PR #10043, opened 2024-03-28
refactor: add integration tests that check the HTML content of different types of pages + fix issues

stephanegigandet:
This PR merges the https://github.com/openfoodfacts/openfoodfacts-server/pull/10023 PR with the main branch, and made corrections that were spotted by the new integration tests that compare the generated HTML:

- fixed missing language code lc in display_error pages
- fixed JS issues that were spotted by Sonar Cloud in the expected test results for the generated HTML
- removed useless select2 entries that were generated but not used any more (spotted by Sonar Cloud)
- made the genererated HTML / JSON always the same by sorting arrays and keys in hashes
- removed Old Church Slavonic language that had the same language codes than Church Slavonic (made generated HTML different)
- simplified the code to generate JSON (e.g. JSON that was converted to UTF8 binary and then converted back to Perl Unicode just after)

We should be able to catch more types of errors with those new tests going forward.


teolemon (commented):
isn't this PR going to break hackish parsing done on the assumption that error messages are in English ?

tests/integration/expected_test_results/export/0850032917148.json
alexgarel, line 1: while were those file removed ? It's a clean up of non existing test ?

alexgarel (approved):
Great.

I'm a bit sceptic that we don't remove spaces in html (reformat html to a well defined format) because it will induce changes on very basic changes in the template, but it's ok.

We could also consider adding something like https://metacpan.org/pod/Test::HTML::Tidy (or https://metacpan.org/pod/HTML::T5 ?)

stephanegigandet:
> isn't this PR going to break hackish parsing done mobile side on the assumption that error messages are in English ?

Error messages are not in English. e.g. in prod: https://fr.openfoodfacts.org/fdsfdsfsd/FSDQfsqfdsqfedsq/fdsfqsfd

This PR repairs display_error() that was broken by https://github.com/openfoodfacts/openfoodfacts-server/pull/10008 because we did not have such HTML tests.

You can see the issue on staging right now: view-source:https://fr.openfoodfacts.net/fdsfdsfsd/FSDQfsqfdsqfedsq/fdsfqsfd   : it's missing the language code in some generated HTML, like "newsletter-" : `<a href="https://link.openfoodfacts.org/newsletter-">S'abonner à notre newsletter</a>`

tests/integration/expected_test_results/export/0850032917148.json
stephanegigandet: I did a rm -rf on expected_tests_results to do some cleanup, and those files were removed and not regenerated.
It seems that they were moved to /openfoodfacts-server/tests/integration/expected_test_results/export/rows

alexgarel: We have to do it from time to time then ;-)

PR #10044, opened 2024-03-28
fix: Disable RequireFilenameMatchesPackage for Config*_.pm

hangy:
Disable the `Modules::RequireFilenameMatchesPackage` rule for config files, because we have several symlinked variants of those.


alexgarel (approved):
:sunglasses:

PR #10047, opened 2024-03-28
docs: updated quick start guide with prerequisites for WSL 

andrewhuanggg:
-->
### What

- added prerequisites for installing dev environment on WSL (I was struggling to set up the dev environment on Windows, but all the errors I had went away after installing on WSL, so this can help other people struggling to setup on Windows machines)


alexgarel (approved):
Thank you !

PR #10050, opened 2024-03-29
refactor: avoid recursion in loading panel.tt.html and divide the logic into blocks (+ HTML test) 

stephanegigandet:
Updated the test results (generated HTML) of @damil 's PR https://github.com/openfoodfacts/openfoodfacts-server/pull/10028 to see the differences. (merged main-html into the original branch to do so)


damil (approved):
looks fine as far as I can tell

templates/web/panels/panel.tt.html
```diff
@@ -1,6 +1,65 @@
...
+[%
+   # check that the panel_id arg is present
+   THROW "panel.tt.html was called without a panel_id arg" IF ! panel_id;
+
+   # preamble: build a map of evaluation values to presentation features
+   SET show_evaluation = {	good 	 => {img_color => 'green',  icon => 'check', td_color => 'green' },
+        					bad  	 => {img_color => 'red',    icon => 'cancel',td_color => 'red'   },
+        					average  => {img_color => 'orange', icon => 'check', td_color => ''      },
+        					neutral  => {img_color => 'grey',   icon => 'check', td_color => 'orange'},
+        					unknown  => {img_color => 'grey',   icon => 'help'}, td_color => 'grey'  };
```
alexgarel, line 10: Why don't we put it in the perl code ?

stephanegigandet: I guess we put the generation of data in the Perl code, but we leave everything that is a presentation choice to the template

alexgarel:
Updated branch as we fixed the perl critic problem.

PR #10058, opened 2024-03-30
fix: make the texts on edit organization profile buttons visible

### LINKED ISSUE #10057: Make Button Texts Visible on Edit Organization Profile Page
### What
- In the 'Edit your organization profile' of the producer platform, the action buttons of the forms do not have a visible value/text.

1. Go to 'https://world.pro.openfoodfacts.org/'
2. Log in to with you producer details
3. Click on the dropdown beside the User's name in the far right and select 'Edit your organization profile'
4. You will see that the buttons on that page have no text.

Here is [a link to the conversation on Slack](https://openfoodfacts.slack.com/archives/C02LDQDDD/p1711439491211779).

### Screenshot
Screenshot 2024-03-25 at 12 46 31 PM

### Type of device
- Browser: Google Chrome (Version 123.0.6312.87)

TheSussex:
### What

I changed the way `process_template` method was initialised in the `org.pl file` to match the implementation in other files. This appears to be preventing the `lang() `method from picking corresponding language values.

### Screenshot
Screenshot 2024-03-30 at 3 47 28 PM

- Fixes #10057


alexgarel (commented):
Cool that you spotted the reason of the bug.

But we should instead correct edq, because we want to escape double quotes !

TheSussex:
> Cool that you spotted the reason of the bug.
> 
> But we should instead correct edq, because we want to escape double quotes !

I'm not sure what you meant by 'correct edq'. Please expantiate.

alexgarel:
Sorry @TheSussex it's my frenglish :-D I should have said "fix edq".

What I mean is that edq should only replace '"' by '\"', so it should not break the functionality… therefore I suppose the implementation of edq is broken so we should fix it.

TheSussex:
> Sorry @TheSussex it's my frenglish :-D I should have said "fix edq".
> 
> What I mean is that edq should only replace '"' by '"', so it should not break the functionality… therefore I suppose the implementation of edq is broken so we should fix it.

Frenglish😁
I get your point now, I'll try to figure out why it's breaking.

alexgarel (approved):
Perfect !

PR #10063, opened 2024-03-31
style: created custom tagify scss file and updated gulp file

### LINKED ISSUE #10030: Use tagify SCSS file for styling of tag inputs
### Description
We currently include the default layout of [tagify](https://github.com/yairEO/tagify/) by copying the original CSS file from the npm package [in gulp](https://github.com/openfoodfacts/openfoodfacts-server/blob/1acd2ff884e176aa2434d5527fd05f4eb7ecd717/gulpfile.ts#L136). However, the tagify package also provides a [source SCSS file](https://github.com/yairEO/tagify/blob/v4.22.2/src/tagify.scss). It would be interesting to use the SCSS file instead and adapt the styling to our new website design.

### Proposed Solution
1. Use tagifies' SCSS file as a modifiable base instead of the pre-built CSS file
1. Update the styling to match our new website design.

### Benefits
- Consistent styling with our website design.
- Easier maintenance and customization using SCSS.

### Implementation Steps
1. Create a new SCSS file, which included the tagify SCSS file.
1. Use that SCSS file instead of the CSS file in Gulp, so it's copied to `css/dist/tagify.css` during build
1. Update the styling as needed.

### Additional Notes
- Ensure compatibility with other components and styles in our project.

IsaiahLevy:
### What
Implemented a custom SCSS file for Tagify inputs and modified the Gulpfile to include these styles in the build process, allowing for customizable appearance and integration with the project's frontend.

- Fixes #10030


alexgarel (commented):
This is great, but I think we loose too much contrast on hover:

image

Also I would better use the brown which of the main theme instead of introducing a color which is not the right one.

scss/custom-tagify.scss
```diff
@@ -0,0 +1,34 @@
...
+  background-color: $text-color-dark;
+}
+
+// Hover state for the remove button in tags
+.tagify__tag .tagify__tag__removeBtn:hover {
+  background-color: $chocolate; // Same 'chocolate' color for consistency
+}
+
+// Hover state for dropdown items
+.tagify__dropdown__item--active,
```
alexgarel, line 26: this rule does not seems to work, but I don't know why.

alexgarel (changes requested):
@IsaiahLevy I still have a rule which does not work (on firefox)

But I found that there is a far better way to handle this:

You can just do:
```css
@import "../node_modules/@yaireo/tagify/src/tagify.scss";

:root {
  $tagify-dd-color-primary: #341100 !important;
  --tagify-dd-color-primary: #341100 !important;
}
```
and it works (it does not work if I use `$chocolate` but I don't know why).
Do you want to try with that (there might be another variable to override)

scss/custom-tagify.scss
IsaiahLevy: Hey thank you for the feedback. Just pushed changes to fix this.

scss/custom-tagify.scss
alexgarel, line 1: @IsaiahLevy your file has merging problems !

gulpfile.ts
```diff
@@ -133,16 +134,17 @@ function copyCss() {
     "./node_modules/leaflet/dist/leaflet.css",
     "./node_modules/leaflet.markercluster/dist/MarkerCluster.css",
     "./node_modules/leaflet.markercluster/dist/MarkerCluster.Default.css",
-    "./node_modules/@yaireo/tagify/dist/tagify.css",
+    // "./node_modules/@yaireo/tagify/dist/tagify.css", // This line should be commented out or removed
```
hangy, line 137:
```suggestion
```

lib/ProductOpener/APIProductServices.pm
hangy, line 1: This change looks out of place

scss/app-ltr.scss
```diff
@@ -1,4 +1,6 @@
 @import "variables";
+@import "custom-tagify";
+@import "variables";
```
hangy, line 3: ` "variable"` is already imported in line 1. Does it have to be included here, again?

templates/web/common/site_layout.tt.html
```diff
@@ -19,6 +19,7 @@
     <link rel="stylesheet" href="[% static_subdomain %]/css/dist/app-[% edq(lang('text_direction')) %].css?v=[% css_timestamp %]" data-base-layout="true">
     <link rel="stylesheet" href="[% static_subdomain %]/css/dist/jqueryui/themes/base/jquery-ui.css" data-base-layout="true">
     <link rel="stylesheet" href="[% static_subdomain %]/css/dist/select2.min.css">
+	<link rel="stylesheet" href="[% static_subdomain %]/css/dist/custom-tagify.css" data-base-layout="true">
```
hangy, line 22: We only really need Tagify in https://github.com/openfoodfacts/openfoodfacts-server/blob/24e957738c1eccd4341e21fd2411069aaa933c4c/cgi/product_multilingual.pl#L777, so this probably needs to be removed, and the import in `cgi/product_multilingual.pl` should be updated.

hangy (changes requested):

alexgarel:
@IsaiahLevy, still you are overriding CSS rule, while I think we should only give values to variables as proposed [in my comment](https://github.com/openfoodfacts/openfoodfacts-server/pull/10063#pullrequestreview-1983220652)

alexgarel (approved):
Great !

alexgarel:
@IsaiahLevy you have to fix linting, your changes to gulp file do not concord with lint policy.

make lint might fix it.

hangy:
Obsoleted by #12944

PR #10077, opened 2024-04-02
fix: export obsolete products only once and include "obsolete" field in CSV export

### LINKED ISSUE #9684: Obsolete products appear as duplicates in CSV export
As of today, [2600+ products are duplicates in the CSV export](https://mirabelle.openfoodfacts.org/products?sql=--+identify+duplicates%0D%0Aselect+code%2C+url%2C+count%28*%29+as+%22count%22+from+%5Ball%5D+group+by+code+having+count%28*%29+%3E+1%3B). It appears these products are obsolete products. Eg. https://world.openfoodfacts.org/product/7614800709822/duo-florentin-kambly

Ideally:
* we should not export them twice
* we should export `obsolete` field in the CSV export

deveshidwivedi:
### What

- This PR implements a fix to ensure that obsolete products are exported only once and do not appear as duplicates .
- Introduces a new field "obsolete" in the CSV export.

- Fixes #9684

### DIFF
--- a/scripts/export_database.pl
+++ b/scripts/export_database.pl
@@ -265,10 +265,6 @@ XML
 		$csv .= "${nid}_100g" . "\t";
 	}
 
-	if (grep { $_ eq 'products_obsolete' } @collections) {
-		$csv .= "obsolete\t";
-	}
-
 	$csv =~ s/\t$/\n/;
 	print $OUT $csv;
 
@@ -278,9 +274,9 @@ XML
 		get_products_collection({obsolete => 1, timeout => 3 * 60 * 60 * 1000})
 	);
 
-	push @export_fields, "obsolete";
 	my $count = 0;
 	my %ingredients = ();
+	my %exported_products = ();
 
 	foreach my $collection (@collections) {
 
@@ -355,9 +351,7 @@ XML
 					$field_value = $product_ref->{obsolete} ? 1 : 0;
 				}
 
-				else {
-					$field_value = ($product_ref->{$field} // "");
-				}
+				$field_value //= "";
 
 				# Add field value to CSV file
 				$csv .= $field_value . "\t";


scripts/export_database.pl
```diff
@@ -265,6 +265,10 @@ sub sanitize_field_content {
 		$csv .= "${nid}_100g" . "\t";
 	}
 
+	if (grep { $_ eq 'products_obsolete' } @collections) {
+		$csv .= "obsolete\t";
+	}
```
alexgarel, line 270: collections is defined below !

I think you can just add the column in every cases.

scripts/export_database.pl
```diff
@@ -298,6 +303,12 @@ sub sanitize_field_content {
 			my $url = "http://world-$lc.$server_domain" . product_url($product_ref);
 			my $code = ($product_ref->{code} // '');
 
+			if ($product_ref->{obsolete}) {
+				if ($exported_products{$code}) {
+					next;
+				}
```
alexgarel, line 309: The idea is correct but you must define $exported_products variable (`my %exported_products = ()`)

scripts/export_database.pl
alexgarel: If fact you don't need this if you correctly add obsolete to @export_fields (in Config_xxx.pm)

scripts/export_database.pl
```diff
@@ -274,6 +278,7 @@ sub sanitize_field_content {
 		get_products_collection({obsolete => 1, timeout => 3 * 60 * 60 * 1000})
 	);
 
+	push @export_fields, "obsolete";
```
alexgarel, line 281: This is not the right place to add it.

Add it to Config_off.pm (and obf/opff/opf for provision).

scripts/export_database.pl
```diff
@@ -339,6 +350,15 @@ sub sanitize_field_content {
...
 				}
 
+				#Obsolete field
+				if ($field eq "obsolete") {
+					$field_value = $product_ref->{obsolete} ? 1 : 0;
+				}
+
+				else {
+					$field_value = ($product_ref->{$field} // "");
+				}
```
alexgarel, line 360: Why put it in a "else" ?

Normalizing undefined values is a good idea, but it's a bit strange to do it in a else here. Just do it plainly, obsolete field_value is always defined, so it won't change it's value.

deveshidwivedi: Updated, thank you!

PR #10080, opened 2024-04-02
bug: description of additives disappeared from the additive panels

stephanegigandet:
It looks like a recent change to the main branch made the description of additives disappear... 

The tests used to pass on https://github.com/openfoodfacts/openfoodfacts-server/pull/10043 , now they don't.

This branch is a clone of 10043 with main merged and tests results updated (but wrong).

To be investigated, maybe a recent change to the taxonomy build process.


stephanegigandet:
Solved by https://github.com/openfoodfacts/openfoodfacts-server/pull/10082

PR #10079, opened 2024-04-02
taxonomy: linting

alexgarel:
Using our new taxonomy linter to lint taxonomies !

PR #10081, opened 2024-04-02
fix: differentiate duplicate line vs multiple lines in taxonomy lint

stephanegigandet:
Trying to build on https://github.com/openfoodfacts/openfoodfacts-server/pull/10079 to differentiate duplicate line (same value) and multiple lines (2 different values for the same language).

When I run "make lint_taxonomies" the entries order change though, there's something weird with the sorting by line number.


stephanegigandet:
Replaced with https://github.com/openfoodfacts/openfoodfacts-server/pull/10083

PR #10090, opened 2024-04-03
ci: Exclude generated test results from code analysis

hangy:
### What

The generate integration tests results do not need to be included in the analysis. At best, they'll report redundant issues, at works false positives because of test configuration.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,3 @@
+name: "CodeQL config"
+paths-ignore:
+  - 'tests/integration/expected_test_results/**/*'
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -29,6 +29,8 @@ jobs:
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v3
+      with:
+        config-file: ./.github/codeql/codeql-config.yml
       # Override language selection by uncommenting this and choosing your languages
       # with:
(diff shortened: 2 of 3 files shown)


stephanegigandet:
@hangy well I did find and fix issues thanks to sonarcloud on the generated HTML in the expected results: https://github.com/openfoodfacts/openfoodfacts-server/pull/10050/commits/5a9286e4e227290bbd0968e44b4a6e8eeb9eb940
(in particular there was a big select2 array that was initialized by was not used anymore as we switched to dynamic suggestions for some fields)

So I would prefer to keep them, unless there is a significant downside on keeping them.

stephanegigandet:
> The security tab is littered with [js/functionality-from-untrusted-source](https://github.com/openfoodfacts/openfoodfacts-server/security/code-scanning?query=is%3Aopen+branch%3Amain+rule%3Ajs%2Ffunctionality-from-untrusted-source) warnings. I don't know if it's possible to suppress that rule for the integration tests path specifically

One possibility could be to change the function that compares/saves HTML test results, to turn http://static.openfoodfacts.localhost/js/dist/jquery.js to https://static.openfoodfacts.localhost/js/dist/jquery.js

alexgarel:
@stephanegigandet 
> One possibility could be to change the function that compares/saves HTML test results, to turn http://static.openfoodfacts.localhost/js/dist/jquery.js to https://static.openfoodfacts.localhost/js/dist/jquery.js

Or, better, to "//static.openfoodfacts.localhost/js/dist/jquery.js"

alexgarel:
@hangy I propose we close your PR, I can instead propose a PR to format urls to "//" only prefix.

hangy:
Works for me

PR #10108, opened 2024-04-08
taxonomy: taxonomy_add_croatian_ingredients

benbenben2:
### What
add more HR words in taxonomy
+ stopwords in HR and other languages


teolemon (approved):
Sounds reasonable and tests are passing, but a second look by @stephanegigandet or @alexgarel might be advisable. 
Some stuff like 'embalado' might have unforeseen effects.

PR #10112, opened 2024-04-09
fix: remove warning in KnowledgePanelsReportProblem.pm, remove report problem card from pro platform

stephanegigandet:
Removes a warning that $panel_data_ref is redeclared


teolemon (approved):
Could you add a simple link to signalconso in addition to the one on the image ? It should work for tomorrow

teolemon (approved):

PR #10114, opened 2024-04-09
fix: base_paths_loading_script needs to be exported

stephanegigandet:
This issue is preventing some shell scripts to work correctly, as they don't have paths exported.


alexgarel (approved):
Thank you !

CharlesNepote (approved):
Thanks!

PR #10126, opened 2024-04-14
taxonomy: romanian

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]


benbenben2 (approved):
**Reminder:** 

> Check if it is a phrase we want to ignore
> NB: If these match, the whole ingredient is ignored, so they're not suitable for ignoring *part* of an ingredient.

This means that if "in proporţie variabilă" is not separated from the ingredient (example: walnut in variable proportion), it will ignore the whole sentence, including the ingredient (walnut in my example). It is fine if you in this language you always have for example "walnuts, in variable proportionm". Keep an eye on it in the future. If it appears to ignore ingredients (first described case) other option would be to add "in proporţie variabilă" in the stopword in the ingredients.txt file, instead.



As for the error, it is due to the formatting of the file (missing or too much spaces or indentations). It can be fixed by running "make lint" before to commit and push. I just did it.

PR #10136, opened 2024-04-15
fix: disable the Invite Users button until at least one email address has been entered

### LINKED ISSUE #10130: Disable Invite Users button on organisation page until email address is entered
### What
When a producer clicks on the` Invite Users ` button without entering any email in the form, it redirects to the home page

1. Go to 'https://world.pro.openfoodfacts.org/'
2. Log in to with you producer details
3. Click on the dropdown beside the User's name in the far right and select 'Edit your organisation profile'
4. Click on the `Invite Users` button without entering any email address in the box, it will redirect you to the organisation's home page.

Here is [a link to the conversation on Slack](https://openfoodfacts.slack.com/archives/C02LDQDDD/p1712567629630809).

### Expected behavior
The ` Invite Users ` button should be disabled until at least one email address has been entered.

### Why
This will help improve user experience because with the current flow, the producer might not know that no action was performed.
 
### Type of device
Browser: Google Chrome (Version 123.0.6312.87)

TheSussex:
### What
- I added a script to the `templates/web/pages/org_form/org_form.tt.html` file that toggles off the Invite Users button until at least one email address has been entered.
- This script also uses Regex to validate the value of each email address entered to ensure users are inviting valid email addresses and not just random texts.

- Fixes #10130

### DIFF
--- a/templates/web/pages/org_form/org_form.tt.html
+++ b/templates/web/pages/org_form/org_form.tt.html
@@ -177,7 +177,7 @@
     <p>&rarr; <a href="[% profile_url %]">[% profile_name %]</a></p>
 [% END %]
 
-<script src="https://static.openfoodfacts.org/js/dist/jquery.js"></script>
+<script src="[% static_subdomain %]/js/dist/jquery.js" data-base-layout="true"></script>
 <script>
     \$(document).ready(function() {
         \$('#email_list').on('input', function(event) {
--- a/templates/web/pages/org_form/org_form.tt.html
+++ b/templates/web/pages/org_form/org_form.tt.html
@@ -70,7 +70,7 @@
 
 	<form method="post" action="/cgi/org.pl" enctype="multipart/form-data" style="margin-bottom: 20px;">
 		<p>[% lang('enter_email_addresses_of_users') %]</p>
-		<textarea id="email_list" name="email_list" style="height:100px;width:50vw" oninput="toggleInviteButton()"></textarea>
+		<textarea id="email_list" name="email_list" style="height:100px;width:50vw"></textarea>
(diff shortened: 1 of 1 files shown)


templates/web/pages/org_form/org_form.tt.html
```diff
@@ -68,13 +68,13 @@ <h2>[% lang("organization_members") %]</h2>
 	</div>
 	[% END %]
 
-		<form method="post" action="/cgi/org.pl" enctype="multipart/form-data" style="margin-bottom: 20px;">
+	<form method="post" action="/cgi/org.pl" enctype="multipart/form-data" style="margin-bottom: 20px;">
 		<p>[% lang('enter_email_addresses_of_users') %]</p>
-		<textarea id="email_list" name="email_list" style="height:100px;width:50vw"></textarea>
+		<textarea id="email_list" name="email_list" style="height:100px;width:50vw" oninput="toggleInviteButton()"></textarea>
```
alexgarel, line 73: oninput does not exists in HTML ! Some JS framework provide it but they transform it to an event listener.

The way to do it is to use JQuery, something like:
```javascript
$(document).ready(function () {
  // observe email entry
  $("#email_list").on("input", function(event) {
    // handle event, eg. toggle Invite button
    // you can get the textarea with event.target
  });
});
```

see https://api.jquery.com/on/

TheSussex: Oh okay, I'll check this out. Thanks

stephanegigandet (approved):
Thank you!

alexgarel (approved):
Great ! Also tested and it works as expected.

PR #10147, opened 2024-04-18
fix: improvements to tags suggestions, from contributors feedback

stephanegigandet:
Some changes to tags suggestions (e.g. categories, labels), based on feedback on Slack: https://openfoodfacts.slack.com/archives/C02KVRT2C/p1713363168918539

- changed the number of saved recent entries to 10 instead of 3 (those are shown before the user types something)
- fixed the order of the suggestions, so that we do display them in the order set by the server (e.g. showing first the entries that match at the beginning)
- removed the automatic highlighting of the first suggestion

before:

image

after:

image

and with the sort order fixed:

image

### DIFF
--- a/html/js/product-multilingual.js
+++ b/html/js/product-multilingual.js
@@ -535,13 +535,14 @@ function initializeTagifyInputs() {
         forEach((input) => initializeTagifyInput(input));
 }
 
-const maximumRecentEntriesPerTag = 3;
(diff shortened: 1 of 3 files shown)


alexgarel (approved):
Cool.

ArturLange (approved):
Awesome, thanks 🙂

PR #10150, opened 2024-04-19
Fix sync_images_s3@.service

raphael0202:
The conf file was not correct in production, I updated it in prod.


stephanegigandet (approved):
Thanks!

PR #10152, opened 2024-04-19
fix: add cc field to auth.pl

stephanegigandet:
add cc field requested by @monsieurtanuki in https://github.com/openfoodfacts/openfoodfacts-dart/pull/903#event-12404120731

note that cc can also be "world"

also removed some fields if their value is undef or empty

### DIFF
--- a/cgi/auth.pl
+++ b/cgi/auth.pl
@@ -30,12 +30,14 @@ use ProductOpener::Display qw/init_request single_param/;
 use ProductOpener::HTTP qw/write_cors_headers/;
 use ProductOpener::Users qw/$User_id %User is_admin_user/;
 use ProductOpener::Lang qw/:all/;
+use ProductOpener::Tags qw/country_to_cc/;
 
 use Apache2::Const -compile => qw(OK);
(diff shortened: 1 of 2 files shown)


alexgarel (approved):
Cool.

monsieurtanuki:
@stephanegigandet Is it already working in PROD?
I've just got this response, with `country` but no `cc`:
```json
{
  "status":1,
  "status_verbose":"user signed-in",
  "user":{
    "admin":0,
    "country":"en:france",
    "moderator":1,
    "name":"(my name)",
    "preferred_language":"fr"
  },
  "user_id":"monsieurtanuki"
}
```

after a call like that:
```dart
    final Uri uri = uriHelper.getPostUri(
      path: '/cgi/auth.pl',
    );
    final Response response = await HttpHelper().doPostRequest(
      uri,
      <String, String>{'body': '1'},
      user,
      uriHelper: uriHelper,
      addCredentialsToBody: true,
    );
```

teolemon:
Deployed to .net currently @monsieurtanuki

PR #10156, opened 2024-04-22
chore: sort taxonomies

alexgarel:
This sort each taxonomy entry:
- parents
- synonyms sorted by languages
- properties sorted by property name

After that taxonomies will have to remain sorted.

This is a pre-requisite to putting taxonomy editor in production as is (otherwise we should change it's behaviour to recall line orders precisely).


taxonomies/ingredients_processing.txt
```diff
@@ -887,17 +887,17 @@ wikidata:en:Q1470757
 
 #<en:pulp
 ca:sense polpa
```
stephanegigandet, line 889: ca: should not be the first language, but it was wrong before

taxonomies/languages.txt
```diff
@@ -18899,10 +18899,10 @@ ne:पश्चिमी फ्रिजी
 nl:Westerlauwers Fries, Fries, Frysk, Westlauwers Fries, Nieuwfries, Friese taal
 nn:Vestfrisisk, Vestfrisisk språk
 pl:Język zachodniofryzyjski
-pt:Frísio ocidental
 pt-br:Frísio ocidental
```
stephanegigandet, line 18902: not very important: I don't understand why pt-br is here, between pl and pms

taxonomies/packaging_shapes.txt
```diff
@@ -182,16 +182,16 @@ description:nl:Een veelal kartonnen balkvormige verpakking
 #Boxes for eggs
 <en:Box
 nl:Eierdoos
```
stephanegigandet, line 184: nl: should not be first, but was broken before

stephanegigandet (approved):
Thanks @alexgarel 
I haven't looked at all the changes, but what I reviewed looks good.
+ the tests are passing, so that's a very good sign

taxonomies/ingredients_processing.txt
alexgarel: We could add an error in taxonomy linter if we have en or xx translation and it is not the first language.

stephanegigandet: Maybe as an option first, or maybe we could add some configuration flags in the taxonomy files, to list which languages can be the first languages, and which properties are supported.

alexgarel: Ok we could add a warning instead of an error.

PR #10162, opened 2024-04-22
fix: Automatically unselect images when source image is deleted #9793

### LINKED ISSUE #9793: Automatically unselect images when source image is deleted

lilymedway:
### What

I fixed the issue where cropped images were not being automatically unselected when the source image was deleted.

- Fixes #9793


lilymedway:
I would like to request a code review for my pull request. I made changes to the [lib/ProductOpener/Images.pm] file to ensure that once an image is deleted it is also unselected so it does not continue to show under that product.

PR #10171, opened 2024-04-24
refactor: specific processes depending on product type

### LINKED ISSUE #10277: Remove old undocumented additives debug fields from the code and products database, update documentation for current fields
[From Slack](https://openfoodfacts.slack.com/archives/C043X1X90/p1715108866237309):

```
I’m looking into additives data returned by the API. However, there is basically has [no documentation](https://openfoodfacts.github.io/openfoodfacts-server/api/ref-v2/#cmp--schemas-product) for the fields. Can somebody explain how those values are filled?
additives_n - fortunately, this one is obvious
additives_old_n - what does old mean here? is this used when the additives changed? under which circumstances would they change? what if additives would be “changed” more than once?
additives_original_tags - what does original mean here? is that tied to additives_old_n ? if so, why is it named differently?
additives_tags - This seems to be the current list of identified additives!? Or does “tags” imply that it is just labels? Can I assume that tags are always normalized to “en:name” or can this be some arbitrary string?
additives_debug_tags - I guess, I can ignore this.
additives_prev_original_tags - looks like “previous” values - but previous to what original? what condition would lead to having previous original tags?
new_additives_n - huh? what does “new” mean? I thought additives_n is the current additives count?
```

stephanegigandet:
Move some code from Product.pm to FoodProduct.pm, BeautyProduct.pm etc.

Also fixes some issues with additives fields that were not removed when ingredients were removed. (fixes #10277)


tests/integration/expected_test_results/convert_and_import_excel_file/packagings-mousquetaires/products/3250390017165.json
```diff
@@ -525,17 +525,14 @@
    "packagings_materials" : {
       "all" : {
          "weight" : 21.01,
-         "weight_100g" : 9.13478260869565,
```
4nt0ineB, line 528: This field was removed

4nt0ineB (approved):

PR #10192, opened 2024-04-28
docs: updated quick start guide with fix to permission issue on macOS. 

tchen0125:
### What
I was having trouble building my local installation due to lacking proper permissions recently. I have added the steps to remedy permission issues on macOS. I added the commands that would allow for user to check which files lack permissions than changing such permissions via terminal.


stephanegigandet (approved):
Thanks

PR #10193, opened 2024-04-29
fix: Move all images to the deleted folder when product is deleted without deleting individual images first 

### LINKED ISSUE #9794: Move all images to the deleted folder when product is deleted without deleting individual images first
### What
- Move all images to the deleted folder when product is deleted without deleting individual images first

tchen0125:
### What

I've added code that uses the fields given for productOpener to make a call after the checks to move deleted product images to deleted.images. It calls the process_image_move() from Images.pm. 

### Screenshot
Screen Shot 2024-04-28 at 9 51 04 PM
Screen Shot 2024-04-28 at 9 56 22 PM

- Fixes #9794


lib/ProductOpener/Config_off.pm
```diff
@@ -188,6 +188,7 @@ $flavor = 'off';
 	stephane
 	tacinte
 	teolemon
+	tchen1
```
alexgarel, line 191: You must not commit that !

lib/ProductOpener/Products.pm
```diff
@@ -1469,13 +1470,30 @@ sub store_product ($user_id, $product_ref, $comment) {
...
+	# Assuming you're within the store_product function and the product is marked as deleted
 	if ($product_ref->{deleted}) {
+    # Check if there are any images associated with the product
+		if (defined $product_ref->{images}) {
+			my $move_to = 'trash';  # Define the destination for deletion
+
+			# Iterate over each image ID stored in the product's images hash
+			foreach my $imgid (keys %{$product_ref->{images}}) {
+				# Ensure the image ID is numeric or follows your expected format
+				if ($imgid =~ /^\d+$/ || $imgid =~ /^(front|ingredients|nutrition)/) {
```
alexgarel, line 1482: There is also packaging !
I would not check it.

lib/ProductOpener/Products.pm
```diff
@@ -1469,13 +1470,30 @@ sub store_product ($user_id, $product_ref, $comment) {
...
+					# Call the process_image_move function using each image ID
+					process_image_move($user_id, $product_ref->{code}, $imgid, $move_to, $product_ref->{owner});
```
alexgarel, line 1484: I would do this in an eval and eventually log an error, because it's important to avoid failing.
(eval used as a form of try / catch in perl).

lib/.DS_Store
alexgarel, line 1: Avoid adding this kind of file :-)
We could add it to .gitignore

alexgarel (changes requested):
Great job ! Some refinement needed still.

lib/ProductOpener/Config_off.pm
tchen0125: Sorry! I will get on the fixes hopefully this weekend or tuesday

teolemon:
@tchen0125 there are some merge conflicts that only you have the permissions to solve. Could you do that ?

kavana262:
Hi @tchen0125, is this still being worked on? It's been quiet for a while and has merge conflicts now happy to pick this up if it's been dropped.

PR #10217, opened 2024-05-02
feat: on the pro platform, detect Nutri-Score 2023 improvement opportunities

stephanegigandet:
On the producers platform, we systematically try to reduce salt/sugar/saturated fat and to increase proteins, fiber and fruits/vegetables/legumes to see if it can improve the Nutri-Score.

This PR:
- adds supports for the new Nutri-Score 2023 formula
- adds some tests


alexgarel (approved):
Great !

PR #10222, opened 2024-05-03
feat: create opportunity upon org validation in CRM

### LINKED ISSUE #7351: Create an entry for validated org in CRM (and it's first user)
Create an org entry and a user entry in CRM as we validate an org (see #7350)

Here are some fields to be copied.

For orgs, create a company entry.
- org_id --> should go in a specific field (off_org_id)
- name / phone / email / website if they exists
- add a tag "pro_platform" (it identify entries managed by pro platform)

Attach to this org a contact which information is based on user info:
- userid --> should go in a specific field (off_user_id)
- name / email
- add a tag "pro_platform" (it identify entries managed by pro platform)

### Part of
- #7296

### LINKED ISSUE #7352: Create an entry for validated user of existing org in CRM
When a user tell us, at subscription time, that he is part of an existing org, we do add a validation step.

Currently this require to manually edit the user data.

Eventually make it a view, akin to #7351

On user org membership validation, create a contact entry in the CRM:
- linked to the company entry retrieved in CRM thanks to org_id (off_org_id)
- with the user data as in #7351 

### Part of
- #7296

### LINKED ISSUE #7353: Create an opportunity on new org creation
When we create a new org (#7351) we should also create an opportunity in the CRM to follow acquisition of this producer.

- on the validation form of #7351, add a checkbox (on by default) to also create an opportunity
- create the opportunity in the CRM
  - linked to the user and the org
  - with a tag "pro_platform" (to identify entries managed by pro platform)
  - name: "<org name> - new org"
  - opportunity should be in the stage corresponding to subscription in the pipeline

### Part of
- #7296

4nt0ineB:
### What
When an organization is validated on the pro platform, associated contact and org are created in the Odoo CRM and linked to a new opportunity. 

### Related issue(s)

- Fixes #7351
- Fixes #7352
- Fixes #7353


lib/ProductOpener/Orgs.pm
```diff
@@ -160,9 +162,33 @@ sub store_org ($org_ref) {
...
+			my $user_ref = retrieve_user($org_ref->{creator});
+			my $crm_user_id = create_contact($user_ref);
+			die "Failed to create contact" 
+			if not defined $crm_user_id;
+
+			die "Failed to add contact to company" 
+			if not defined add_contact_to_company($crm_user_id, $crm_org_id);
+
+			die "Failed to create opportunity" 
+			if not defined create_opportunity($user_ref, $org_ref);
```
stephanegigandet, line 182: Maybe store the opportunity id as well, so that we can update it later, for instance if the org imports products.

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,257 @@
...
+	my $contact = {
+		name => 			$user_ref->{name}, 
+		x_off_username => 	$user_ref->{userid},
+		email => 			$user_ref->{email},
+		phone => 			$user_ref->{phone}, 
+		category_id => 		[$odoo_tags{Producter}],
+	};
+	
+	# find country code id in odoo
+	my $user_country_code = uc $country_codes_reverse{$user_ref->{country}} || 'EN';
```
alexgarel, line 129: I would prefer the use of parenthesis for every function call (clearer for many contributors)

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,257 @@
...
+
+=head4 $org_ref
+
+=head3 Return values
+
+the id of the created contact
+
+=cut
+
+sub create_contact ($user_ref) {
```
alexgarel, line 119: I imagine it comes with another PR but it would be important  to check user existence for the email in odoo.

Duplicates on contact are really annoying to the producers team.

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,257 @@
...
+	my $country_id = odoo('res.country', 'search_read', [[['code', '=', $user_country_code]]], { 'fields' => ['code', 'id']});
+	$contact->{country_id} = $country_id->[0]{id};
+
+	# find spoken language's code in Odoo
+	my $odoo_lang = odoo('res.lang', 'search_read', [[['iso_code', '=', $user_ref->{preferred_language}]]], { 'fields' => ['code']});
+	my $found_lang = $odoo_lang->[0];
+	$contact->{lang} = $found_lang->{code} // 'en_US'; # default to english
+	if(defined $found_lang->{id}) {
+		$contact->{x_off_languages} = [$found_lang->{id}];
+	}
```
alexgarel, line 139: There is a problem with Odoo because it will refuse to display a non-activated language in many to many fields (that's deep in the code), that's why I also created the [x_off_languages_str](https://github.com/openfoodfacts/odoo-openfoodfacts/blob/main/skill-pool/ir_model_fields.yml#L76) field to store language values directly

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,257 @@
...
+=head4 $org_ref
+
+=head3 Return values
+
+the id of the created opportunity
+
+=cut
+
+sub create_opportunity ($user_ref, $org_ref) {
+	my $opportunity_id = odoo('crm.lead', 'create', [{name => "$org_ref->{name} - new org", partner_id => $user_ref->{crm_user_id}}]);
```
alexgarel, line 191: It would be great to check titles, etc with Manon

alexgarel (commented):
Good work !

I did a quick review and added a comment !

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,257 @@
...
+	%EXPORT_TAGS = (all => [@EXPORT_OK]);
+
+}
+use vars @EXPORT_OK;
+
+use ProductOpener::Tags qw(%country_codes_reverse);
+use XML::RPC;
+use Log::Any qw($log);
+use Data::Dumper;
+use feature 'say';
```
hangy, line 62: `Data::Dumper` and `say` aren't actually used, right? If they are, it _might_ be better to use a proper `$log` method call instead

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,257 @@
...
+
+=head3 Return values
+
+the response or undef if an error occured
+
+=cut
+
+sub odoo(@params) {
+	if(not defined $xmlrpc) {
+		my $api_url = $ENV{ODOO_CRM_URL} . '//xmlrpc/2/';
```
hangy, line 209: We don't really use `$ENV` for dynamic configuration like this, outside of [`Config2_docker.pm`](https://github.com/openfoodfacts/openfoodfacts-server/blob/ece6f77e9333e5e35f727ac71f7b1f272f5f7cbb/lib/ProductOpener/Config2_docker.pm). The "proper" way to add optional odoo configuration is probably adding it to `Config.pm` and to `Config2.pm`

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,257 @@
...
+		my $username = $ENV{ODOO_CRM_USER};
+		my $db = $ENV{ODOO_CRM_DB};
+		my $pwd = $ENV{ODOO_CRM_PASSWORD};
+
+		eval {
+			$xmlrpc = XML::RPC->new($api_url . 'common');
+		};
+		if($@) {
+			$log->warn("odoo", {error => $@, reason => "Could not connect to Odoo CRM"});
```
hangy, line 218: For local development or some test environments, we won't always have "odoo" installed or configured. It might be a good idea to guard against the object being created again and again if no odoo connection is configured. For example during integration tests, this could avoid some noisy error logging (and possibly reduce execution duration)

Issue #7267, opened 2022-08-29
Barcode not displayed for products with fake barcodes attributed by OFF, breaks Folksonomy Engine


CharlesNepote:
### What

- The paragraph "Barcode: xxxxxxxxxxxx" is not displayed on this product: https://world.openfoodfacts.org/product/2000000009452/pate-d-olives-noires-prior
- The following test seems to fail:
- https://github.com/openfoodfacts/openfoodfacts-server/blob/main/templates/web/pages/product/product_page.tt.html#L57
- It makes Folksonomy Engine failing.

### To Reproduce

- Just open: https://world.openfoodfacts.org/product/2000000009452/pate-d-olives-noires-prior

### Part of
- Folksonomy Engine

stephanegigandet:
The product does not have a real barcode, so we don't display it. We could have a hidden <span id="barcode">.

teolemon:
The solution is either:
- to handle Folksonomy Engine server side
- to change the way the FE JS scripts gets the barcode (hidden in the DOM, or from the URL ?)

alexgarel:
The way to go IMO is to always have barcode but with display none (and no schema if it's fake)

teolemon:
I have added the number of impacted products to the GH project: https://world.openfoodfacts.org/code/2xxxxxxxxxxxx

PR #10223, opened 2024-05-03
fix: Add non EAN barcode in product page

alexgarel:
hidden, for folksonomy engine to work.

fixes: #7267


stephanegigandet (approved):
Thanks

PR #10242, opened 2024-05-06
fix: added barcode-svg

### LINKED ISSUE #1728: Show barcode image on product page

bazylevnik0:
### What
Added support of the barcode SVGs in the page

### Screenshot
Screenshot from 2024-05-06 20-08-14

- Fixes #1728
- It is my first PR, was autonomous as can, but please provide more information about changes if needed


templates/web/common/site_layout.tt.html
```diff
@@ -26,6 +26,7 @@
         [% styles %]
     </style>
     [% google_analytics %]
+	<script src="https://cdn.jsdelivr.net/jsbarcode/3.3.7/JsBarcode.all.min.js"></script>
```
teolemon, line 29: We avoid dependancies on CDNs, could you add the library directly as a dependancy

html/js/JsBarcode.all.min.js
hangy, line 1: Could you add [this dependency](https://www.npmjs.com/package/jsbarcode) as a [npm dependency](https://github.com/openfoodfacts/openfoodfacts-server/blob/ba512066776172c2177cbeb93d7e6b63c1aca0ed/package.json#L39) and then add it to the [`copyJs()` gulp task](https://github.com/openfoodfacts/openfoodfacts-server/blob/ece6f77e9333e5e35f727ac71f7b1f272f5f7cbb/gulpfile.ts#L60), please?

hangy: Thank you! :)

templates/web/common/site_layout.tt.html
```diff
@@ -490,6 +490,14 @@ <h3 class="title-5 text-medium">[% lang('footer_discover_the_project') %]</h3>
 [% END %]
 
 </body>
+<script src="[% static_subdomain %]/js/dist/JsBarcode.all.min.js"></script>
+<script>
+	//init barcode SVGs if exist on the page
+	let barcode_svg_elements = document.getElementsByClassName("barcode-svg"); 
+	if(barcode_svg_elements.length!=0){
+		JsBarcode(".barcode-svg").init();
```
hangy, line 498: Since you already have the array, this should be enough:
```suggestion
		JsBarcode(barcode_svg_elements).init();	
```
`JsBarcode` seems to do the same by calling [`getRenderProperties`](https://github.com/lindell/JsBarcode/blob/5ed2a2b9da5f82da3c6159eb47a21d06f1cf797d/src/help/getRenderProperties.js#L21) internally. Might save some ms by not running the selector again

templates/web/common/site_layout.tt.html
```diff
@@ -490,6 +490,14 @@ <h3 class="title-5 text-medium">[% lang('footer_discover_the_project') %]</h3>
 [% END %]
 
 </body>
+<script src="[% static_subdomain %]/js/dist/JsBarcode.all.min.js"></script>
+<script>
+	//init barcode SVGs if exist on the page
```
hangy, line 495: I'm not sure what the best place for this snippet is, but in general, we tried to remove most direct JS from the templates in the last few months. Maybe [display-product.js](https://github.com/openfoodfacts/openfoodfacts-server/blob/21681e9e25fcae0eba17ddd82da466149e522834/html/js/display-product.js)? cc @alexgarel @stephanegigandet

stephanegigandet: display-product.js would be the right place I think, we don't need to load this script on every page, just product pages.

In fact we probably want to display the barcode only on desktop (not mobile), as on mobile it takes a lot of space for little added value. On desktop it's very useful so that it can be scanned with a mobile app (like ours, or others).

I'm not sure if we could load the script conditionnaly if the screen is big enough though.

templates/web/pages/product/product_page.tt.html
```diff
@@ -163,8 +162,15 @@ <h2 class="title-1" property="food:name" itemprop="name">[% title %]</h1>
                     <!-- Display UPC code if the EAN starts with 0 -->
                         [% IF upc_code == 'defined' %]
                             <p id="barcode_paragraph">
-                                [% lang("barcode") %]:  <span id="barcode" property="food:code" itemprop="gtin13" style="speak-as:digits;">[% code %]</span> [% upc %]
+                                [% lang("barcode") %]:  <br>
+                                <span id="barcode" property="food:code" itemprop="gtin13" style="speak-as:digits;">[% code %]</span>
+                                [% upc %]
                             </p>
+                            <svg class="barcode-svg"
```
stephanegigandet, line 169: Nice!

Could you change the display so that it is displayed only on desktop, and at the right of the barcode number, so that we don't lose vertical space? The barcode is very nice to have in order to be able to scan it with a mobile app, but we don't want to push the real content too much down.

e.g. something like that:

image

templates/web/pages/product/product_page.tt.html
```diff
@@ -162,12 +162,36 @@ <h2 class="title-1" property="food:name" itemprop="name">[% title %]</h1>
...
+                                #barcode_div_code,#barcode_div_svg{
+                                    display: inline-block;
+                                    vertical-align: top;
+                                }
+                                @media only screen and (max-width: 600px) {
+                                    #barcode_div_svg,#barcode_br {
+                                        display: none;
+                                    }
+                                }
+                            </style>
```
alexgarel, line 190: we should put it in the main _off.scss file.

alexgarel (commented):
Thanks @bazylevnik0, this is great.

Moving to _off.scss seems the only remaining item to me !

alexgarel:
I updated tests results.

alexgarel:
@bazylevnik0 thanks for this great contribution.

Sorry for we were quite under an heavy load last week, and it's only now that I saw it was failing since a long time.

I just add to run "make update_test_results" to have all test results updated.

alexgarel (approved):
Great, thanks !

PR #10259, opened 2024-05-09
taxonomy: translate categories in Croatian

benbenben2:
### What
translate categories in Croatian

- Fixes #-

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -75,6 +75,7 @@ stopwords:de:und,mit,von
 
 en:Bulk
 fr:Vrac
+hr:rasuti
 nl:Verpakkingsvrij
 
 en:Artisan products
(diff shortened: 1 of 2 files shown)


benbenben2:
> The new categories are in lowercase, is that intended? Just curious, did you use some kind of automated script / AI to translate and add the categories?

Yes, a script was used for that.

I updated to start by upper case for first character

benbenben2:
I just tried to update integration test manually because I could not do it, nor run tests inside the container:
```
docker exec -it po_off-backend-1 bash
perl tests/integration/web_html.t
```
> Can't locate Test/File/Contents.pm in aINC (you may need to install the Test::File::Contents module) 
(can't install the package, Permission denied)

Also outside the container:
```
make test-int test="web_html.t"
```
> OCI runtime exec failed: exec failed: unable to start container process: exec: "yath": executable file not found in $PATH: unknown
make: *** [Makefile:299 : test-int] Erreur 126

stephanegigandet:
@benbenben2 You may need to run "make dev" again to install the new modules that have been added in recent weeks.

stephanegigandet (approved):
Thank you!

PR #10261, opened 2024-05-09
taxonomy: translate ingredients in Croatian

benbenben2:
### What

- Fixes #-


stephanegigandet (approved):
Thanks

PR #10266, opened 2024-05-10
feat: Pull MongoDB and Redis from shared-services repo

### LINKED ISSUE #10265: Update Product Opener to fetch Redis and MongoDB Dependencies from openfoodfacts-shared-services repo
### Description

We want to move Redis and MongoDB into their own repo so that other projects that depend on these, like off-query, can just pull in these dependencies without needing the whole of Product Opener.

### Acceptance criteria

Can develop on a related project, like off-query, without running ProductOpener locally.

### What would a demo look like

With a clean directory tree, clone off-query and make deps to show that MongoDB and Redis are available.

### Notes

At the moment MongoDB exposes the html/data directory for imports and its data/db directory for running scripts. It is proposed that the scripts that PO needs should now run in Perl (create_mongodb_indexes) and that the full product import should be moved into the shared_services repo.

_Should we also make a test SMTP server a shared service?_

### Tasks
- [x] Create shared-services repo with MongoDB and Redis containers
- [x] Update PO create_mongodb_indexes to run without direct access to a MongoDB volume
- [ ] Move the import_prod_data target to shared_services (PO version can call out to this)
- [x] Remove MongoDB and Redis container definitions from PO (except for when running integration tests)
- [ ] Move GitHub action to deploy MongoDB from PO to shared_services
- [ ] Update off-query to use shared_services for MongDB and Redis
- [ ] Update off-auth to use shared_services for Redis
- [ ] Add branch protection to openfoodfacts-shared-services

john-gom:
### What

Create a test.yml for integration tests but otherwise MongoDB and Redis come from the shared-services repo.
https://github.com/openfoodfacts/openfoodfacts-shared-services

- Fixes #10265


docker/test.yml
alexgarel, line 1: I am a bit sceptical about duplicating so much to avoid pulling deps.

Why not have instead "test_deps" and "deps" in the Makefile  to avoid pulling too much, still keeping things DRY and consistent ?

john-gom: Hi @alexgarel , I've tweaked it so the deps are cloned for tests but not run so that we can reference the docker yml in the test targets. Let me know what you think. Note there are changes to shared-services that go with this, but clone deps should do a pull anyway

john-gom:
This is mostly done, although I noticed the devcontainer.json is still mentioning redis and mongodb so assume this might need some changes to cater for shared-services. @alexgarel, @stephanegigandet , @teolemon I'm not very familiar with how devcontainers work - who is the resident expert?

john-gom:
> This is mostly done, although I noticed the devcontainer.json is still mentioning redis and mongodb so assume this might need some changes to cater for shared-services. @alexgarel, @stephanegigandet , @teolemon I'm not very familiar with how devcontainers work - who is the resident expert?

I have made edits to the devcontainer.json and I think it works now. I basically added an initializeCommand to start the dependencies first (make run_deps). @hangy , I notice you've appeared in a lot of devcontainers conversations so perhaps you could check I've got this right?

alexgarel:
> This is mostly done, although I noticed the devcontainer.json is still mentioning redis and mongodb so assume this might need some changes to cater for shared-services. @alexgarel, @stephanegigandet , @teolemon I'm not very familiar with how devcontainers work - who is the resident expert?

I think this is @hangy. I personally don't use it yet but would be glad to have an introduction also :-D

Makefile
```diff
@@ -196,46 +204,33 @@ reset_owner:
 
 init_backend: build_taxonomies build_lang
 
-create_mongodb_indexes:
+create_mongodb_indexes:run_deps
 	@echo "🥫 Creating MongoDB indexes …"
-	docker cp conf/mongodb/create_indexes.js $(shell docker compose ps -q mongodb):/data/db
-	${DOCKER_COMPOSE} exec -T mongodb //bin/sh -c "mongo off /data/db/create_indexes.js"
+	${DOCKER_COMPOSE_RUN} run --rm backend perl /opt/product-opener/scripts/create_mongodb_indexes.pl
```
alexgarel, line 209: :+1:

Makefile
```diff
@@ -196,46 +204,33 @@ reset_owner:
...
-	  LEFT=$$(($$LEFT/1024)); \
-	  if [[ $$LEFT -lt $$NEEDED ]]; then >&2 echo "NOT ENOUGH SPACE LEFT ON DEVICE: $$NEEDED MB > $$LEFT MB"; exit 1; fi
-	wget --no-verbose https://static.openfoodfacts.org/data/openfoodfacts-mongodbdump.gz -P ./html/data
-	wget --no-verbose https://static.openfoodfacts.org/data/gz-sha256sum -P ./html/data
-	cd ./html/data && sha256sum --check gz-sha256sum
-	@echo "🥫 Restoring the MongoDB dump …"
-	${DOCKER_COMPOSE} exec -T mongodb //bin/sh -c "cd /data/db && mongorestore --quiet --drop --gzip --archive=/import/openfoodfacts-mongodbdump.gz"
-	rm html/data/openfoodfacts-mongodbdump.gz && rm html/data/gz-sha256sum
+import_prod_data: run_deps
+	@cd ${DEPS_DIR}/openfoodfacts-shared-services && $(MAKE) -e import_prod_data
```
alexgarel, line 233: You can remove it from this file… it's basically used by the daily task, it concerns mongodb only and it does not make much sense keeping it here (it belongs to shared-net repository)

docker/run.yml
```diff
@@ -0,0 +1,17 @@
...
+      shared_network:
+  minion:
+    networks:
+      # Needed to access MongoDB and Redis
+      shared_network:
+
+networks:
+  # This network allows access to shared services like MongoDB and Redis
+  shared_network:
+    name: shared_network
```
alexgarel, line 16: Shan't we add a project name or something like that ?

Otherwise you can't work on two clones of the repo at the same time.

I think the COMPOSE_PROJECT_NAME is shared with dependencies, right ?

In this case, prefix it:
```suggestion
    name: ${COMPOSE_PROJECT_NAME}_shared_network
```

Note: do it only if it's simple enough.

docker/test.yml
alexgarel: So now do we still need this file ?

alexgarel (approved):
Thank @john-gom, looks good to me. I have one or two remarks.

I think you can remove docker/test.yml

hangy:
> > This is mostly done, although I noticed the devcontainer.json is still mentioning redis and mongodb so assume this might need some changes to cater for shared-services. @alexgarel, @stephanegigandet , @teolemon I'm not very familiar with how devcontainers work - who is the resident expert?
> 
> I think this is @hangy. I personally don't use it yet but would be glad to have an introduction also :-D

I haven't used devcontainers with OFF in a while, but the changes look fine to me. They devcontainers config is also used for GitPod/GitHub Codespaces, IRC.

Makefile
john-gom: I kept it as I have used it myself before to get all data locally and it is also referenced in docs.

docker/test.yml
john-gom: Removed

PR #10299, opened 2024-05-16
fix: Banner accessibility

### LINKED ISSUE #10298: [a11y] donation banner has two main concerns
The donation banner is displayed only with a mouse click: `<span id="hide-donate-banner" class="material-icons" onclick="DonationButton();">close</span>`.

The analyzer we use reports two main issues:
1.  The mouse event attribute doesn't have a keyboard equivalent: see https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR20
2.  The element is a forced user interface control without the proper role attribute: see https://www.w3.org/WAI/WCAG21/Techniques/failures/F59

For the first one, it should be possible to open it thanks to keyboard. The solution seems to only add a keyboard event:
```html
<span id="hide-donate-banner" class="material-icons" onclick="DonationButton();" onkeypress="DonationButton();" role="link">close</span>
```

The second one is more obscure...

Valimp:
### What
Accessibility for donate banner
Give role to spans
Alt to images
Make the close button accessible

- Fixes #10298


templates/web/common/includes/donate_banner.tt.html
```diff
@@ -17,9 +17,9 @@
       <a class="donation-banner__donate" href="[% link %]?utm_source=off&utf_medium=web&utm_campaign=donate-2023-a&utm_term=en-text-button">[% lang("donation_cta") %]</a>
     </div>
   </div>
-  <div class="donation-banner__image"></div>
+  <div class="donation-banner__image" aria-label="Last meeting in Paris"></div>
   <div class="donation-banner__close">
-    <span id="hide-donate-banner" class="material-icons" onclick="DonationButton();">close</span>
+    <button id="hide-donate-banner" class="material-icons modest" onclick="DonationButton();" onkeypress="DonationButton();">close</button>
```
stephanegigandet, line 22: What's the "modest" class?

stephanegigandet:
Thanks for the PR!

If there are any visual differences, could you add before / after screenshots?

Valimp:
Before: 
Capture d’écran 2024-05-16 à 12 42 49



After : 
Capture d’écran 2024-05-16 à 12 42 00


there is a small difference at the button, is this problematic?

stephanegigandet:
> there is a small difference at the button, is this problematic?

Thanks, looks fine to me. The language is different though ;)

CharlesNepote (approved):

PR #10303, opened 2024-05-17
chore: add ClaudeBot to robots.txt

stephanegigandet:
See also #10302


stephanegigandet:
From @raphael0202 :


Raphaël
  il y a 12 minutes
Well it's not really necessary, if we deny access (which is the case for ClaudeBot now), the following robots.txt will be displayed:
User-agent: *
Disallow: /

PR #10316, opened 2024-05-18
taxonomy: translate ingredients in Croatian

benbenben2:
### What


stephanegigandet (commented):
for ingredients (but not categories), the convention is to put them not capitalized

same question than for the categories: did you review a sample (or all) translations?

benbenben2:
> for ingredients (but not categories), the convention is to put them not capitalized
> 
> same question than for the categories: did you review a sample (or all) translations?

Same comment as #10317


I updated file to have everything starting by lower case for hr:

stephanegigandet (approved):
Thank you!

PR #10317, opened 2024-05-18
taxonomy: translate categories in croatian

benbenben2:
### What

translate categories in Croatian

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -8692,6 +8692,7 @@ wikidata:en:Q132675
 
 <en:Herbal teas
 en:Boldo
+xx:Boldo
 de:Boldo
 es:Boldo
 fr:Boldo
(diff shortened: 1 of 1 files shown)


stephanegigandet:
@benbenben2 I'm not sure how we can review this. Can you indicate the review you did? Did you review all the translations, or maybe looked at a sample?

benbenben2:
> @benbenben2 I'm not sure how we can review this. Can you indicate the review you did? Did you review all the translations, or maybe looked at a sample?

These are google translations for "en:"
Then, I reviewed everything (hovered over to check if something looked weird in the translation).
Then, I got rid of the error messages of duplicates (that's why you can also see some deletions). 

I noticed for example for the name of some fish or some fruits that the google translation weren't correct. I had to find some Croatian websites and also using latin names to find correct names.

stephanegigandet (approved):
Thank you!

PR #10322, opened 2024-05-20
fix: Disallow: / in robots.txt for denied crawlers

stephanegigandet:
This is to send Disallow: / for denied crawlers.


raphael0202 (approved):
Great, thank you!

PR #10334, opened 2024-05-20
fix: import new sample import template with description and example rows

### LINKED ISSUE #10313: The new sample import file for producers on the pro platform cannot be imported
We added a column and a description and sample rows:

https://world.pro.openfoodfacts.org/cgi/generate_sample_import_file.pl

We need to add support for this new format (basically ignoring the extra column and sample rows), while making sure we can still import the old format (without the extra column and rows)

Related code is in Import.pm in import_csv_file()

stephanegigandet:
We recently added a description row and an example row to the import template generated by https://world.pro.openfoodfacts.org/cgi/generate_sample_import_file.pl

This PR
- enables loading back those files that have 2 extra rows + 1 extra column
- adds an integration test for it

Fixes #10313


4nt0ineB (approved):

manon-corneille:
There may be a few things to fix with the new file: 
* we experience much more 504 than with the previous file
* when refreshing the page, the lines of description that should be ignored are interpreted as column headers, which makes mapping go wrong (cf image) 

image

PR #10336, opened 2024-05-21
fix: optimize MongoDB queries and caching for lists of products

stephanegigandet:
We have a mechanism in place to cache the results of MongoDB queries in Product Opener, but for most lists of products, we are unable to cache the results, as they are too big to fit in Memcached.

e.g. for 100 products, the size of the MongoDB response is:
(total_size: 8 139 666 > max size (1 048 576))

We can try to limit further the fields we retrieve from MongoDB, in order to reduce the size of data we get back. The challenge is to get back enough data to be able to compute product attributes.

By being more precise on Nutri-Score and Eco-Score fields, we can reduce the size by about half:
(total_size: 4 491 352 > max size (1 048 576))

It's still too big to cache, but it's a progress. (+ it should speed a bit the MongoDB queries)

The bulk of the size in in the "images" structure. For web queries, we can remove it after we compute image_front_url, and before caching. This yields a drastic reduction of the size of the object to cache:

717 576

So for the same list of 100 products, we have a reduction in size of 90%. This is before compression.

After Alex's suggestion to cache results, it turns out that the memcached module already tries to compress the data. But if the uncompressed data is bigger than 1M (the max size of compressed items in memcached), we don't try to cache. I think we can safely assume that compression will give us at least a 50% reduction in size, so I increased the limit by a factor of 2.

Removing fields like images etc. before caching is quite tricky as it's easy to forget about some cases where we do need the fields that we removed (e.g. for some queries like [https://
](http://fr.openfoodfacts.localhost/categorie/caf%C3%A9s.json?fields=brands) -> Initially I had removed the "brands" after computing the "product_display_name" (basically product name + brand + quantity), but then the extra ?fields=brands couldn't get it.

I added some tests to minimize the risk that caching breaks something. I also added a flush of the cache in the tests, as running integration tests in a different order could yield different results (we have the same query for all France products twice in different tests).


alexgarel:
https://metacpan.org/pod/Compress::Zstd could do for compression

stephanegigandet:
@alexgarel I looked into it, but in fact Memcached clients (including the Perl module we use) already have compression built-in: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Cache.pm#L52

But we currently do not try to store items that are larger than the max item size (1M), and we do that check before compression... I'm guessing we could at least multiply that check by some factor where we are pretty sure that compression will bring us lower than the limit. What we want to avoid is compressing data and then not even caching it because it's too large even when compressed.

stephanegigandet:
Some stats (not good) on the current hit rate we have for MongoDB queries:

```
root@off:/srv/off/logs# tail -n 1000000 mongodb_log4perl.log | ./cache_stats.pl 
Stats for PRODUCTS_COUNT queries:
- get: 168623
- get - hit: 21194 (12.6%)
- get - miss: 147056 (87.2%)
- set: 142173
- set - updated: 142173 (100.0%)
- set - skipping: 0 (0.0%)
Stats for PRODUCTS queries:
- get: 184914
- get - hit: 8455 (4.6%)
- get - miss: 176067 (95.2%)
- set: 169704
- set - updated: 119980 (70.7%)
- set - skipping: 49724 (29.3%)
```

PR #10340, opened 2024-05-22
feat: First step to the CRM integration

### LINKED ISSUE #735: Offer to add suggested languages in multilingual countries
In multilingual countries such as Switzerland, products are often in 2 or 3 languages. 

Offer to add suggested languages more easily than typing the language manually. (Eg Italian and German as greyed tabs you can click)

### LINKED ISSUE #7351: Create an entry for validated org in CRM (and it's first user)

### LINKED ISSUE #7352: Create an entry for validated user of existing org in CRM

### LINKED ISSUE #7353: Create an opportunity on new org creation

### LINKED ISSUE #7354: Avoid duplicate org and users in CRM
### What
- #7351 and #7352 are discussing creating users and orgs in the CRM.

But as part of the user / org acquisition will be an upstream process, it would be important to avoid duplicating entries for org and users.

As we create user and org we should use some strategies to try to verify if there is one already matching our entry.

1. if there is a matching entry with off_org_id / off_user_id set (it might have been set by producer platform team), use this one
2. if there is a user with matching email (and no off_user_id), use this one and if it's linked to an org (without an off_org_id), match it to the org
3. if there is an org with the exact same name (and no off_org_id) use this one

If we find a matching entry for org and/or user, we do not create a new one, but we may add data that is not present, and we add the pro_platform tag.

### Part of
- #7296

4nt0ineB:
### What

First step in integrating the CRM with the producer platform.
- Added Odoo CRM credentials as environment variables.
- An organization is 'unreviewed' by default and can be 'validated' or 'rejected'. The “scripts/migration/2024_06_org_validation_move_field.pl” script **must** be run to change the old value in the .sto of orgs.

- When an org is validated, it is added to the CRM. Its creator is also added and linked to an opportunity. A strategy is put in place to avoid duplicates and associate them with entities already in the CRM.
- A main contact has been added to the organization, being the person who will receive the emails and be the center of the workflow in the CRM. It can be modified in the pro platform.
- Adding a contact to a validated organization in the pro platform adds it to CRM.
- Updates the products last import and export date of the corresponding org in the CRM when performed from the producer platform.

- Fixes #7351 
- Fixes #7352 
- Fixes #7353 
- Fixes #7354


cgi/import_file_process.pl
```diff
@@ -125,6 +126,8 @@
 
 my $started_t = time();
 my $import_id = $started_t;
+# sync CRM
+update_last_import_date($Org_id, $started_t);
```
stephanegigandet, line 130: We also have nightly imports through Equadis and Agena3000, they usually call import_csv_file.pl
See for instance https://github.com/openfoodfacts/openfoodfacts-server/blob/main/scripts/imports/agena3000/run_agena3000_import.sh
The issue is that they contain multiple organizations, so it would need to be done based on the organizations actually present in the updated products.

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,572 @@
...
+ProductOpener::CRM - manages integration with the Odoo CRM
+
+=head1 SYNOPSIS
+
+C<ProductOpener::CRM> contains functions to interact with the Odoo CRM
+
+
+=head1 DESCRIPTION
+
+For clarity:
```
stephanegigandet, line 32: Good idea to document this here.

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,572 @@
...
+use vars @EXPORT_OK;
+
+use ProductOpener::Config2;
+use ProductOpener::Tags qw/%country_codes_reverse/;
+use ProductOpener::Users qw/retrieve_user store_user/;
+use ProductOpener::Orgs qw/retrieve_org is_user_in_org_group/;
+use XML::RPC;
+use Log::Any qw($log);
+
+my %Odoo_tags = (Producer => '10',);
```
stephanegigandet, line 69: Maybe add a comment to indicate where to find the id of a tag in Odoo

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,572 @@
...
+
+=head3 Return values
+
+1 if success, undef otherwise
+
+=cut
+
+sub link_user_with_contact($user_ref, $contact_id) {
+	my $req = Odoo('res.partner', 'write',
+		[[$contact_id], {x_off_username => $user_ref->{userid}, category_id => [[4, $Odoo_tags{Producer}]]}]);
```
stephanegigandet, line 119: What is 4? Maybe we should put it in a constant a the top of the module if it could be different on another Odoo instance

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,572 @@
...
+	# 1. & 3. merged in one query
+	my $companies = Odoo(
+		'res.partner',
+		'search_read',
+		[
+			[
+				'&', ['is_company', '=', 1],
+				'|', ['x_off_org_id', '=', $org_ref->{org_id}],
+				'&',
+				['name', '=', $org_ref->{name}],
```
stephanegigandet, line 297: Is this case sensitive?

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,572 @@
...
+
+=head3 Return values
+
+1 if success, undef otherwise
+
+=cut
+
+sub add_contact_to_company($contact_id, $company_id) {
+	# set contact in children of company
+	# 4 means add an existing record
```
stephanegigandet, line 377: Ah, here is my answer for the 4. :) We could set some constants for actions like this.

scripts/org_validation_move_field.pl
```diff
@@ -0,0 +1,47 @@
...
+# to a 3 state field: unreviewed, accepted, rejected
+
+use ProductOpener::Store qw/store/;
+use ProductOpener::Orgs qw/list_org_ids retrieve_org/;
+use ProductOpener::Paths qw/%BASE_DIRS/;
+
+foreach my $org_id (list_org_ids()) {
+	my $org_ref = retrieve_org($org_id);
+	# print "org_id: $org_id, is: $org->{valid_org}\n";
+	if (exists $org_ref->{valid_org} and defined $org_ref->{valid_org}) {
```
stephanegigandet, line 36: Should we also put the creator as the main_contact? Or maybe we could at least list all the orgs that do not have a main contact (in some cases OFF members have created the org, but they should not be main contacts)

templates/web/pages/org_form/org_form.tt.html
```diff
@@ -62,12 +62,24 @@ <h2>[% lang("organization_members") %]</h2>
...
 				<input type="hidden" name="action" value="process" />
 				<input type="hidden" name="type" value="admin_status" />
+				<input type="hidden" name="orgid" value="[% orgid %]">
 				<input type="submit" name="grant_remove_admin_status" class="button" style="margin-bottom: 10px;" value= "[% edq(lang('grant_remove_admin_status')) %]"/>
 			</form>
 		</div>
 	</div>
 	[% END %]
 
+	[% IF has_main_contact %]
```
stephanegigandet, line 72: If there is no main contact, we don't display the form to add one?

lib/ProductOpener/CRM.pm
```diff
@@ -0,0 +1,572 @@
...
+			[
+				['is_company', '=', 0],
+				'|',
+				['x_off_username', '=', $user_ref->{userid}],
+				['email', '=', $user_ref->{email}]
+			]
+		],
+		{fields => ['name', 'id', 'parent_id'], order => 'create_date ASC', limit => 1}
+	);
+	my $contact = $req->[0];
```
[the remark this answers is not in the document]
4nt0ineB: Why would this cause an error? If $req is undef then it should autovivify to an empty array ref so it would always be undef, or some compile flags are set that I don't know about?

Yes, if the search query has no results, it returns an empty array.

tests/integration/expected_test_results/create_pro_user/org-after-validation.json
```diff
@@ -14,12 +14,12 @@
...
+   "main_contact" : "tests",
    "members" : {
       "tests" : 1
    },
    "name" : "Acme Inc.",
    "org_id" : "acme-inc",
    "protect_data" : "",
-   "valid_org" : "1",
-   "validated" : 0
+   "valid_org" : "1"
```
stephanegigandet, line 24: create_pro_user.t should be changed so that it does not send "1"

stephanegigandet (approved):
Looks good to me, thank you!

PR #10344, opened 2024-05-23
fix: warnings modperl_error_log

benbenben2:
### What
attempt to fix warnings modperl_error_log

I believe it should fix

```
106	-e: Use of uninitialized value $title in concatenation (.) or string at /srv/off/lib/ProductOpener/Display.pm line 1481.
122	-e: Use of uninitialized value $code in pattern match (m//) at /srv/off/lib/ProductOpener/Products.pm line 412.
126	-e: Argument "" isn't numeric in numeric gt (>) at /srv/off/templates/api/knowledge-panels/environment/ecoscore/origins_of_ingredients.tt.json line 57.
126	-e: Argument "" isn't numeric in numeric le (<=) at /srv/off/templates/api/knowledge-panels/environment/ecoscore/origins_of_ingredients.tt.json line 40.
126	-e: Argument "" isn't numeric in numeric gt (>) at /srv/off/templates/api/knowledge-panels/environment/ecoscore/origins_of_ingredients.tt.json line 45.
138	-e: Use of uninitialized value in numeric le (<=) at /srv/off/lib/ProductOpener/KnowledgePanels.pm line 1117.
162	-e: Wide character (U+6210) in ucfirst at /usr/local/lib/x86_64-linux-gnu/perl/5.32.1/Template/Filters.pm line 57.
174	-e: Wide character (U+5206) in ucfirst at /usr/local/lib/x86_64-linux-gnu/perl/5.32.1/Template/Filters.pm line 57.
184	-e: Invalid conversion in sprintf: end of string at /srv/off/lib/ProductOpener/Display.pm line 9017. -> 9050 (+43)
191	-e: Argument "" isn't numeric in division (/) at /srv/off/templates/api/knowledge-panels/environment/ecoscore/origins_of_ingredients.tt.json line 92.
223	-e: Argument "" isn't numeric in division (/) at /srv/off/templates/api/knowledge-panels/environment/origins_of_ingredients.tt.json line 88.
229	-e: Argument "" isn't numeric in numeric le (<=) at /srv/off/templates/api/knowledge-panels/environment/origins_of_ingredients.tt.json line 50.
393	-e: Use of uninitialized value $canon_tag in string eq at /srv/off/lib/ProductOpener/Tags.pm line 3448.
393	-e: Use of uninitialized value $tag in ucfirst at /srv/off/lib/ProductOpener/Tags.pm line 3450.
[...]

PR #10352, opened 2024-05-26
fix: warnings modperl error log

benbenben2:
### What
Similar to #10344.
Trying to prevent warnings in logs.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -3808,7 +3808,7 @@ sub display_tag ($request_ref) {
 
 	my $weblinks_html = '';
 	my @wikidata_objects = ();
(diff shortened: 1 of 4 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -3808,7 +3808,7 @@ sub display_tag ($request_ref) {
 
 	my $weblinks_html = '';
 	my @wikidata_objects = ();
-	if (    ($tagtype ne 'additives')
+	if (    (define $tagtype && $tagtype ne 'additives')
```
alexgarel, line 3811: typo:

```suggestion
	if (    (defined $tagtype && $tagtype ne 'additives')
```

alexgarel (approved):
Cool ! Just spotted a typo !

PR #10375, opened 2024-05-31
fix: enable caching for logged-in users, use Cache-Control: no-cache

stephanegigandet:
Caching of MongoDB queries was disabled for logged in users, so that contributors could see the results of their changes immediately.

This PR changes the caching behavior so that MongoDB  queries are cached for all users (even logged in users). But it enables bypassing caching when the HTTP Cache-Control: no-cache header is set (sent by browsers when doing Ctrl-Shift-R).

Also added a test.


teolemon (approved):

4nt0ineB (approved):

hangy (commented):
How do you want to handle usability impacts? For example, it could be very confusing to search for products that don't have a specific country, add that country to the product, open the same search, and still see the product that I just modified in the result set

john-gom:
Couldn't we change it so that logged in user queries are not cached but go to off-query instead of MongoDb?

As long as redis is working people shouldn't notice the difference

stephanegigandet:
> How do you want to handle usability impacts? For example, it could be very confusing to search for products that don't have a specific country, add that country to the product, open the same search, and still see the product that I just modified in the result set

We will just tell contributors to refresh the page with Ctrl Shift R. If they don't refresh the page, they will see cached results for 1 hour.

alexgarel (approved):
Cool !

PR #10390, opened 2024-06-04
taxonomy: translate ingredients in croatian 3

benbenben2:
### What


stephanegigandet (approved):
Thanks

PR #10391, opened 2024-06-04
taxonomy: translate categories in croatian 3

benbenben2:
### What

translate categories in croatian 3


stephanegigandet (approved):
Thank you!

PR #10392, opened 2024-06-05
chore: opposite categories

aleene:
### What
Adding opposite characteristics to categories, as discussed in the data quality meeting. Based on the list on [https://wiki.openfoodfacts.org/Data_quality_missions](https://wiki.openfoodfacts.org/Data_quality_missions).

Some on the list are not included as these are better designated by the characteristic exclusive (can not be combined with another category)..


CharlesNepote:
Great!

To go further, could you explain deeper the "exclusive" characteristic? What I understand is that an "exclusive" category means that the product can't have another category than itself and its parents.
Eg. Plant-based foods and beverages > Plant-based foods | Condiments > Spices > Peppers > White Peppers

But could you also provide an example of a category that is NOT exclusive?

Regarding to [the list here](https://wiki.openfoodfacts.org/Data_quality_missions#Categories), could you make two lists: one containing the opposites and one containing the exclusives?

aleene:
We could add tests:
- the opposite category name should exist as a category in order to prevent spelling errors
- the opposite category should also have an opposite category that refers back

taxonomies/food/categories.txt
```diff
@@ -1141,6 +1164,7 @@ hr:Pića i pripravci za piće
 it:Bevande e preparati per bevande
 nl:Dranken en drankbereidingen
 description:en:This mixes ready to drink beverages and not-ready to drink beverages
+opposite:en: en:meals
```
stephanegigandet, line 1167: The opposite of "Beverages" would be "Food" I think, not "meals". Of course that would mean putting most parent categories under "Food"

stephanegigandet (approved):
I added a comment for beverages vs meals. Otherwise it looks good to me. I think we could try it, and add some data quality warnings based on it, to see what product it flags, and then we could adjust.

PR #10394, opened 2024-06-05
fix: utf8 orgid support in url query string

### LINKED ISSUE #10177: Greek alphabet support on the Pro platform
A Greek producer created a Pro account using Greek alphabet letters, but it doesn’t seem to be supported by the Pro platform. I am missing something? Should I create an issue? 

image

4nt0ineB:
### What

- Fixed the issue where the orgid from the URL query string was not correctly decoded in Perl's internal string representation when it contained utf8 characters. On the professional user's side, the org .sto file could not be found for the same reason.

### Screenshot

Screenshot from 2024-06-05 10-57-29

- Fixes #10177

### DIFF
--- a/cgi/org.pl
+++ b/cgi/org.pl
@@ -53,7 +53,7 @@ my $request_ref = ProductOpener::Display::init_request();
 my $orgid = $Org_id;
 
 if (defined single_param('orgid')) {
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Thanks!

PR #10404, opened 2024-06-06
taxonomy: remove duplicate split peas entry in ingredients

stephanegigandet:
There were 2 entries for split peas, which broke properties for it.


benbenben2 (approved):
Lgtm

PR #10422, opened 2024-06-10
feat: use more the allergens:en property of the ingredients taxonomy

stephanegigandet:
We currently have many synonyms of allergens listed in the allergens.txt taxonomy, but very often those are not actual synonyms, but more specific ingredients. (e.g. for the en:fish allergen, we list different species of fish).

This does not scale very well. Instead we now have the allergens:en: property in the ingredients taxonomy. It is currently used to identify allergens listed as ingredients.

This PR extends the use of the allergens:en: property so it can be used to normalize the allergens and traces that are listed manually in the "allergens" and "traces" field.

There are still other uses of the synonyms in allergens.txt (e.g. for parsing phrases like "may contain...". We should try to use the allergens:en: property instead, so that we can ultimately remove the "non synonyms" from allergens.txt

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -6330,6 +6330,9 @@ sub preparse_ingredients_text ($ingredients_lc, $text) {
 	# Traces de lait, d'oeufs et de soja.
 	# Contains: milk and soy.
 
+	# TODO: we should use the allergens:en: property from the ingredients.txt taxonomy instead of relying
+	# on having extensive "non synonyms" (like fish species) in allergens.txt
+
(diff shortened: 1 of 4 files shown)


aleene:
If we could move the allergens.txt entries into ingredient.txt, it would be great. Can start to do that when this one is live?

stephanegigandet:
> If we could move the allergens.txt entries into ingredient.txt, it would be great. Can start to do that when this one is live?

Not yet, we will need more code changes to remove the synonyms (e.g. fish species) from allergens.txt. But we are getting closer.

4nt0ineB (approved):

PR #10428, opened 2024-06-11
fix: change mpm_prefork configuration for off and off-pro to reduce memory usage

stephanegigandet:
This reduces MaxConnectionsPerChild from 500 to 100 on OFF (and activates it on off-pro) in order to reduce the size of the apache mod_perl processes that grow over time.

The processes start at 1.5Gb, but on off-pro they grew up to 2.5Gb.

I also reduced the number of processes on off-pro, as traffic should be very small.

@raphael0202 I accidentally checked in a change to off-log.conf that was on the off container on off2, I assume it's a wanted change?

### DIFF
--- a/conf/apache-2.4/off-mpm_prefork.conf
+++ b/conf/apache-2.4/off-mpm_prefork.conf
@@ -11,7 +11,7 @@
 	MaxSpareServers		 10
 	MaxRequestWorkers	  50
 	ServerLimit		55
-	MaxConnectionsPerChild   500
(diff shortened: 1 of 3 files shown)


CharlesNepote (approved):

raphael0202:
@stephanegigandet Sorry I missed this mention, yes it was a hot-change made in prod, so I could track the rate-limiter.
But I've created (and merged) a PR adding these changes on Product Opener, so we're good.

PR #10434, opened 2024-06-12
fix: improve logging

raphael0202:
Improve logging, especially for the rate-limiter.

Replace '%r' (=time since starting the apache2 process) by %d (datetime) in the logs, as we currently don't know in the log files when the log record was created.
Also transform some info log in debug in Redis.pm.


stephanegigandet (approved):
Thanks!

PR #10437, opened 2024-06-13
fix: detect_allergens_from_ingredients typo

foodietim:
`detect_allergens_from_ingredients` bugfix, allows listing all of the allergens that are automatically tagged from ingredients existing to show up, rather then the last allergen that was analyzed.

This should allow us to automatically detect many more allergens automatically from user input, without requiring the user to manually enter them.


stephanegigandet:
Thank you @foodietim ! The tests in tests/unit/allergens_tags.t are failing, but it's just because the results need to be updated, as the new ones are all correct I think.

stephanegigandet (approved):
Thanks!

stephanegigandet:
@foodietim I'm trying to make the test pass

PR #10440, opened 2024-06-13
feat: new sync data in CRM 

### LINKED ISSUE #7356: Sync data in CRM on org and user updates
With #7351 and #7352, we creates users and orgs in the CRM.

But we need to keep data up to date in the CRM. Also we want to have in the CRM some data from the pro platform that can help target right producers, follow their journey and eventually create emails workflows.

For org we would like to track:
- date of last export to public plateform
- typology of last import (manual / csv)  (postpone if not easy) / use a tag import_from_csv/manual
- number of products exported and in progress
- pim it's imported via (use tags, like import_from_xxxx)

For a user we would like to track:
- date of last connexion

PS: remember to always prefix specific fields in the CRM with `off_`

4nt0ineB:
### What

Add new sync data in CRM:
- last import / export date
- Last date the xlsx import template was downloaded
- last connected user (and date)
- set "onboarding" tag on opportunity of new accepted orgs
- When importing, add corresponding category of its typology : EQUADIS, AGENA3000, CSV, Manual import 

- Fixes #7356

### DIFF
--- a/lib/ProductOpener/CRM.pm
+++ b/lib/ProductOpener/CRM.pm
@@ -553,7 +553,6 @@ sub change_company_main_contact($org_ref, $user_id) {
 	return $req_company;
 }
 
-
 =head2 update_last_import_date ($org_id, $time)
 
 Update the last import date of a company in Odoo, associeted to the given org
@@ -584,7 +583,6 @@ sub _update_company_last_action_date($org_id, $time, $field) {
 		if $log->is_debug();
 	return make_odoo_request('res.partner', 'write', [[$org_ref->{crm_org_id}], {$field => $date_string}]);
 }
-
 
 =head2 update_company_last_contact_login_date ($org_ref, $user_ref)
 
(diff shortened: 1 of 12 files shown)


4nt0ineB:
I took the existing labels for the tags and categories (AGENA3000, EQUADIS, ...) in the CRM. Should I follow the naming convention suggested in the issue instead ? import_from_agena3000, import_from_csv, etc..

lib/ProductOpener/CRM.pm
```diff
@@ -81,7 +84,7 @@ my $crm_data;
 my @required_tag_labels = qw(onboarding);
 # Category (res.partner.category) must be defined in Odoo :
```
stephanegigandet, line 85: The meaning of "category" is not very intuitive, could you rename the field to "data_sources" instead?

stephanegigandet (approved):
Looks good to me, thank you!
Do you want to merge it now, or are you still working on it?

PR #10447, opened 2024-06-17
chore: add script to gather organization data daily

### LINKED ISSUE #10443: Create a script to gather organisation data nightly from the products database
### Task
Create a script to gather organisation data nightly from the products database.

We will do this using a cron job.

Here is [a link to the project document](https://docs.google.com/document/d/137Jvu2W8KR0Qgbfkey96G3mOjScto_eO1LT32G12h0Q/edit?pli=1)

### Part of story
- #9737

TheSussex:
### What

I added a script to the scripts/migrations folder, this script fetches organzaition products based on the product owner id.

- Fixes #10443


scripts/update_product_data_daily.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
stephanegigandet, line 1: Could you rename this script to gen_feeds_daily_off_pro.sh ? (similar to  the existing gen_feeds_daily_off.sh)

scripts/migrations/2024_06_save_org_product_data_daily.pl
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/perl -w
```
stephanegigandet, line 1: the /migrations/ folder is only for scripts that we are running once (e.g. to update the format of the product data). Could you remove the date from the file name, and move it outside of the migrations/ folder?

scripts/save_org_product_data_daily.pl
```diff
@@ -0,0 +1,110 @@
...
+
+sub gather_org_data {
+    my @orgs = $orgs_collection->find()->all();
+	my $count = scalar @orgs;
+	my $i = 0;
+
+    foreach my $org (@orgs) {
+        my $org_id = $org->{'org_id'};
+        print "Processing organization $i/$count: $org_id\n";
+        update_org_data($org_id);
```
alexgarel, line 103: It would be better to do this in an eval (the try/except for Perl) so that if we fail for one org, we can compute values for the others.

Also we should log in case of error.

```suggestion
        eval { update_org_data($org_id) };
        my $org_error = $@;
        print(*STDERR "Error computing data for org $org_id: $org_error\n") if $org_error;
```

scripts/save_org_product_data_daily.pl
```diff
@@ -0,0 +1,110 @@
...
+# This script is run daily to gather organisation data 
+# such as number of products, number of products with errors etc,
+
+ensure_dir_created($BASE_DIRS{ORGS});
+
+my $products_collection = get_products_collection();
+my $orgs_collection = get_orgs_collection();
+
+sub update_org_data {
+    my $org_id = shift;
```
alexgarel, line 43: Please follow our standards and use named parameters.

scripts/save_org_product_data_daily.pl
```diff
@@ -0,0 +1,110 @@
...
+
+    my $data = {
+        'products' => {
+            'number_of_products' => $org_data->{number_of_products} // 0,
+            'number_of_data_quality_errors' => $org_data->{number_of_data_quality_errors} // 0,
+            'number_of_data_quality_warnings' => $org_data->{number_of_data_quality_warnings} // 0,
+            'number_of_products_without_nutriscore' => $org_data->{number_of_products_without_nutriscore} // 0,
+            'percentage_of_products_with_nutriscore' => $percentage_of_products_with_nutriscore
+        },
+    };
```
alexgarel, line 79: It would be great to isolate this part in a function, returning $data.

alexgarel (commented):
Great, that's going in the right direction.

scripts/save_org_product_data_daily.pl
TheSussex: I'm not sure what you mean @alexgarel

TheSussex: Never mind, I get it now.

cgi/product_multilingual.pl
```diff
@@ -271,6 +271,7 @@ ($product_ref)
 					if $log->is_info();
 				$product_ref = init_product($User_id, $Org_id, $code, $country);
 				$product_ref->{interface_version_created} = $interface_version;
+				$product_ref->{owner} = $Owner_id;
```
stephanegigandet, line 274: This should only be set if we are on the producer platform and $Owner_id is defined. When a contributor adds a product on the public platform, we should not set $product_ref->{owner}

scripts/save_org_product_data_daily.pl
```diff
@@ -0,0 +1,126 @@
...
+
+sub gather_org_data {
+	my @orgs = $orgs_collection->find()->all();
+	my $count = scalar @orgs;
+	my $i = 0;
+
+	foreach my $org (@orgs) {
+		my $org_id = $org->{'org_id'};
+		print "Processing organization $i/$count: $org_id\n";
+		update_org_data($org_id);
```
stephanegigandet, line 119: Looks good, but you did not apply @alexgarel 's suggestion for the eval:

https://github.com/openfoodfacts/openfoodfacts-server/pull/10447#discussion_r1674182902

scripts/save_org_product_data_daily.pl
```diff
@@ -0,0 +1,128 @@
...
+sub get_org_data ($org_id) {
+	my $org_data = $products_collection->aggregate(
+		[
+			{'$match' => {'owner' => "org-" . $org_id}},
+			{
+				'$group' => {
+					'_id' => '$owner',
+					'number_of_products' => {'$sum' => 1},
+					'number_of_data_quality_errors' => {
+						'$sum' => {'$size' => '$data_quality_errors_tags'}
```
stephanegigandet, line 51: in my dev account, I got this error when running the script:

MongoDB::DatabaseError: The argument to $size must be an array, but was of type: missing

Can you check that the array exist before calling $size? There is an example here: https://www.mongodb.com/docs/manual/reference/operator/aggregation/size/#example

(same for other array below)

scripts/save_org_product_data_daily.pl
```diff
@@ -0,0 +1,128 @@
...
+					},
+					'number_of_data_quality_warnings' => {
+						'$sum' => {'$size' => '$data_quality_warnings_tags'}
+					},
+					'number_of_products_without_nutriscore' => {
+						'$sum' => {
+							'$cond' => [{'$in' => ['en:nutriscore-not-computed', '$misc_tags']}, 1, 0]
+						}
+					},
+					'opportunities_to_improve_nutriscore' => {
```
stephanegigandet, line 61: This field is only available on the pro platform, so I think we should run this script on the pro platform.

Can you also add the number of products with states_tags = en:exported , and states_tags = en:to-be-exported ?

scripts/save_org_product_data_daily.pl
```diff
@@ -0,0 +1,128 @@
...
+							'$cond' => [{'$regexMatch' => {input => '$states', regex => 'en:to-be-exported'}}, 1, 0]
+						}
+					},
+					'date_of_last_update' => {'$max' => '$last_modified_t'}
+				}
+			}
+		]
+	)->next;
+
+	my $number_of_products = $org_data->{number_of_products} // 0;
```
stephanegigandet, line 77: In another PR, we will need to make another query for number_of_products, but on the public platform, in order to get the number of products from the pro platform that have been exported to the pro platform.

stephanegigandet (approved):
Thanks

PR #10458, opened 2024-06-18
feat: sync the right opportunity salesperson

4nt0ineB:
### What

Set the salesperson of the opportunity corresponding to the off admin account that validates the org, based on the email.


stephanegigandet (approved):
Looks good, thanks.

Issue #8282, opened 2023-04-03
Disable edition of protected data sent by producers in the web product edit form, with explanations


stephanegigandet:
Some data imported from producers through the producers platform cannot be edited by users (unless they are moderators). We should visually disable the edition of those fields in the product edit form, with an explanation, and a mechanism to contact us in case the user believes something protected is innaccurate.

See also:

- https://github.com/openfoodfacts/openfoodfacts-server/issues/6816
- https://github.com/openfoodfacts/openfoodfacts-server/issues/2012
- https://github.com/openfoodfacts/openfoodfacts-server/issues/6721

- Note: when https://github.com/openfoodfacts/openfoodfacts-server/pull/8274 is merged, edits of protected fields through the web form will be silently ignored, but we need to warn users beforehand.

PR #10465, opened 2024-06-19
feat: show which information is provided by manufacturers in product edit form

stephanegigandet:
This PR changes the product edit form on the web so that we warn contributors that some information has been provided by producers, and that it cannot be updated (unless they are moderators).

Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/8282

When editing products, if some data has been provided by a manufacturer, we display a warning:

image

TODO: add details on how to report issues (once Nutri-Patrol is deployed).

The exact fields are displayed (with a checkmark that can be hovered on) and greyed out / inactive:

image

Moderator view:

We keep the checkmark, but the fields are not disabled.

image

The PR also fixes an issue with tags fields that could not be changed if the producer sent a value (e.g. for category or brand). As this info may not be complete / precise, it is not protected.

TODO: also specify which nutrient is protected.


stephanegigandet:
Nutrition facts:

image

4nt0ineB (approved):

teolemon (commented):
We might want to use a factory icon, to not confuse with a potential "field checked" feature, an extension of our existing "product checked" feature.

stephanegigandet:
> We might want to use a factory icon, to not confuse with a potential "field checked" feature, an extension of our existing "product checked" feature.

Good idea

PR #10472, opened 2024-06-20
fix: update paths for EAN8 and short barcodes (padding with zeroes) - DO NOT MERGE

stephanegigandet:
I closed the very old PR https://github.com/openfoodfacts/openfoodfacts-server/pull/3915 and I am opening this new one to put it on the first page.

Original issue: https://github.com/openfoodfacts/openfoodfacts-server/issues/3818

We will need to move the file structures on all flavors (OFF, OBF, OPF, OPFF) roughly at the same time, as otherwise we will have issues if some products are moved from one flavor to another.

So the migration script is using "old" conventions so that it can run on the old obf / opf / opff code (without needing newer modules).

The migration script is first used to assess the situation (how many products would be moved, how many products have conflicts etc.) on all flavors.

PR in infrastructure repo to detail the migration: https://github.com/openfoodfacts/openfoodfacts-infrastructure/pull/363


scripts/move_ean8_products_to_new_path.pl
alexgarel, line 1: can we put it in a migration sub folder ?

scripts/move_ean8_products_to_new_path.pl
```diff
@@ -0,0 +1,347 @@
...
+
+my @products = ();
+my $move = 0;
+
+GetOptions('products=s' => \@products, 'move' => \$move);
+@products = split(/,/, join(',', @products));
+
+my $d = 0;
+
+open(my $log, ">>", "$data_root/logs/move_ean8_products_to_new_path.log");
```
alexgarel, line 157: nit: you should normally be using the Path.pm module !

stephanegigandet: I need the script to work on the old code (so that I can run it in dry-run mode on OBF, OPF, OPFF, to see which products are affected), that's why it does not use some of the newer things like Path.pm

scripts/move_ean8_products_to_new_path.pl
stephanegigandet: yes, I will move it

lib/ProductOpener/Products.pm
```diff
@@ -433,16 +452,47 @@ Example: 1234567890123  :-  123/456/789/0123
...
 
 		$log->info("invalid code", {code => $code}) if $log->is_info();
 		return "invalid";
 	}
 
-	# First splits into 3 sections of 3 numbers and the ast section with the remaining numbers
+	# Pad code with 0s if it has less than 13 digits
+	while (length($code) < 13) {
+		$code = "0" . $code;
+	}
```
alexgarel, line 493:
```suggestion
	if (length($code) < 13) {
		$code = "0" x (13 - length($code)) . $code;
	}
```

lib/ProductOpener/Products.pm
```diff
@@ -289,11 +290,49 @@ sub normalize_code ($code) {
...
+		return $code;
+	}
+
+	# Remove leading zeroes
+	$code =~ s/^0+//;
+
+	# Add leading zeroes to have at least 13 digits
+	while (length($code) < 13) {
+		$code = "0" . $code;
+	}
```
alexgarel, line 325:
```suggestion
	if (length($code) < 13) {
		$code = "0" x (13 - length($code)) . $code;
	}
```

lib/ProductOpener/Products.pm
```diff
@@ -289,11 +290,49 @@ sub normalize_code ($code) {
...
-=head2 normalize_code_with_gs1_ai()
+=head2 normalize_code_zeroes($code)
+
+On disk, we store product files and images in directories named after the product code, and we add leading 0s to the paths.
+So we need to normalize the number of leading 0s of product codes, so that we don't have 2 products for codes that differ only by leading 0s.
+
+This function normalizes the product code by:
+- removing leading zeroes,
+- adding leading zeroes to have at least 13 digits,
+- removing leading zeroes for EAN8s to keep only 8 digits
```
alexgarel, line 306: I don't understand why we keep this rule ?

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
```diff
@@ -0,0 +1,348 @@
...
+
+sub normalize_code_zeroes($code) {
+
+	# Remove leading zeroes
+	$code =~ s/^0+//;
+
+	# Add leading zeroes to have at least 13 digits
+	while (length($code) < 13) {
+		$code = "0" . $code;
+	}
```
alexgarel, line 67:
```suggestion
	if (length($code) < 13) {
		$code = "0" x (13 - length($code)) . $code;
	}
```

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
```diff
@@ -0,0 +1,348 @@
...
+
+	# Require at least 4 digits (some stores use very short internal barcodes, they are likely to be conflicting)
+	if (not is_valid_code($code)) {
+		return "invalid";
+	}
+
+	# Pad code with 0s if it has less than 13 digits
+	while (length($code) < 13) {
+		$code = "0" . $code;
+	}
```
alexgarel, line 96:
```suggestion
	if (length($code) < 13) {
		$code = "0" x (13 - length($code)) . $code;
	}
```

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
```diff
@@ -0,0 +1,348 @@
...
+
+GetOptions('products=s' => \@products, 'move' => \$move);
+@products = split(/,/, join(',', @products));
+
+my $d = 0;
+
+open(my $log, ">>", "$data_root/logs/move_ean8_products_to_new_path.log");
+print $log "move_ean8_products_to_new_path.pl started at " . localtime() . "\n";
+
+if (scalar $#products < 0) {
```
alexgarel, line 164:
```suggestion
if ((scalar @products) == 0) {
```

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
```diff
@@ -0,0 +1,348 @@
...
+
+	my $dh;
+
+	opendir $dh, "$data_root/products" or die "could not open $data_root/products directory: $!\n";
+	foreach my $dir (sort readdir($dh)) {
+		chomp($dir);
+
+		if ($dir =~ /^\d\d\d$/) {
+
+			# We can have products with 9 to 12 digits that have a split path but that were not padded with 0s
```
alexgarel, line 175: if it's 9 digits, shan't you check that there are no sto files in $dh4 ?

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
```diff
@@ -0,0 +1,348 @@
...
+				push @products, $dir;
+				$d++;
+				(($d % 1000) == 1) and print STDERR "$d products - $dir\n";
+			}
+		}
+	}
+	closedir $dh;
+}
+
+my $count = $#products;
```
alexgarel, line 225: nit:
```suggestion
my $count = scalar @products;
```

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
```diff
@@ -0,0 +1,348 @@
...
+					# Another option is to call the system mv command.
+
+					# In this script, we want to avoid creating copies as we are using zfs, so we use File::Copy move()
+
+					print STDERR ("moving product data $data_root/products/$old_path to $data_root/products/$path\n");
+					print $log ("moving product data $data_root/products/$old_path to $data_root/products/$path\n");
+
+					if (not move("$data_root/products/$old_path", "$data_root/products/$path")) {
+						print STDERR ("could not move product data: $!\n");
+						print $log ("could not move product data: $!\n");
```
alexgarel, line 292: Shan't we count that as an error, and do a move-- and errors++

Also is it a good idea to continue with image if we failed for product data ?

alexgarel, line 292: I would prefix real errors in error log with a `ERR:` to easily find them afterwards (if we need manual corrections).

This has to be done for the other entries.

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
```diff
@@ -0,0 +1,348 @@
...
+			print STDERR "new path exist, not moving $old_path to $path\n";
+			print $log "new path exist, not moving $old_path to $path\n";
+			$not_moved++;
+		}
+
+		if ($new_code ne $code) {
+			$changed_code++;
+			print STDERR "changed code from $code to $new_code\n";
+			print $log "changed code from $code to $new_code\n";
+		}
```
alexgarel, line 334: It's not true, in the case it was not moved… I would put that above !

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
```diff
@@ -0,0 +1,348 @@
...
+
+	}
+	else {
+		print STDERR "new $path is the same as old $old_path\n";
+		print $log "new $path is the same as old $old_path\n";
```
alexgarel, line 339: do you really need to log it ?
It will make your log very verbose for no reason, IMO

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
alexgarel, line 1: Not to be changed, but just a general opinion: when I write a migration script, I always try to first find what's to be done and then do it.
Here you could first have made a csv of all the move to be done (or not) and then in the second part process them.

lib/ProductOpener/Products.pm
alexgarel: of course this is nitpicking :grin:

alexgarel (commented):
A lot of nitpicking but some important comments.

lib/ProductOpener/Products.pm
stephanegigandet: it's to keep barcodes closer to how they are written on packages. while it's possible to refer to a ean9 with 5 leading 0s, on the package, the 0s are not there (that's whole point of ean8).

as you noted, that way we won't change barcodes for all EAN8s

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
stephanegigandet: Good point, fixing

stephanegigandet: good idea

scripts/migrations/2024_10_move_ean8_products_to_new_path.pl
stephanegigandet: done

alexgarel (approved):
Thanks for this important work.

PR #10474, opened 2024-06-20
refactor: global variables removal

4nt0ineB:
### What

The goal is to get rid of the global variables as suggested by the various 'Todo' in the code, to limit side effects.  
These will be accessed through `$request_ref ` : 
- `$admin`
- `$styles`
- `$initjs`
- `$cc`
- `$bodyabout`

About `$lc` and `$country`I'm not sure if I it's a good idea. For a part I can easily apply the same thing as for the other ones, because it's just a matter of putting them in the `request_ref` (in the .pl cgi). But for the rest, they are used at a very nested level in the call stack. I'd have to add one or two params to many functions to get to them, and where they are called also. It's not unfeasible though.


stephanegigandet (approved):
Looks good to me, thank you!

PR #10480, opened 2024-06-21
fix: Use a separate make target for production up to not create dependencies

john-gom:
### What

Use a different target for docker compose up in production so that dependencies aren't also started in dev mode.


alexgarel (approved):
LGTM

PR #10481, opened 2024-06-21
taxonomy: german again

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]


taxonomies/food/ingredients.txt
```diff
@@ -22855,11 +22854,16 @@ it:aroma naturale di mandorle
...
 
 <en:natural flavouring
-en:natural sour almond flavouring
+en:natural bitter almond flavouring
 de:natürliches Bittermandelaroma, natürliches Bittermandel-Aroma
 fr:arôme naturel d'amande amère
 hr:prirodna aroma kiselog badema
 
+<en:natural flavouring
+en:bitter almond extract
```
stephanegigandet, line 22863: I don't remember if we have already discussed it, but it would be a good thing to decide how we classify "XYZ extracts", whether they should be under "natural flavouring", under "flavouring", or maybe under a new category which would be "flavourings and extracts".

This has an implication on things like NOVA: currently we haven't put "extracts" as markers of NOVA 4 (but I'm thinking we probably should).

stephanegigandet (approved):

PR #10483, opened 2024-06-22
taxonomy: Rusks

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -35880,7 +35880,8 @@ lt:Spelta džiūvėsiai
 nl:Speltbeschuiten, Speltbeschuit
 
 <en:Rusks
-fr:Biscottes de froment, Biscottes au froment
+en:Classic common wheat rusks
+fr:Biscottes de froment classiques, Biscottes de froment, Biscottes au froment
 de:Weizenzwieback
 nl:Tarwebeschuiten, Tarwebeschuit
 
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -35880,7 +35880,8 @@ lt:Spelta džiūvėsiai
 nl:Speltbeschuiten, Speltbeschuit
 
 <en:Rusks
-fr:Biscottes de froment, Biscottes au froment
+en:Classic common wheat rusks
```
stephanegigandet, line 35883: I'm not too sure about the "Classic" part. Without context it's quite unclear. I'm guessing that would be the "salted" commen wheat rusks? But classic could mean lots of things: like not wholemeal, or not without gluten.

Maybe we could keep "common wheat rusks", and just have "unsalted common wheat rusks" under it.

moon-rabbitOFF: Okay, I can remove the classic. The idea was to have a category with all the rusks sold as "wheat rusks" without more specifications, separated from rusks with specifications, so as you said wholemeal, without gluten, or without salt. We have a category "Classic hummus" for example to separate the ones just called "hummus" from the ones with added ingredients (the possibilities with hummus are infinites, e.g. with harrissa, with avocado, betroot, spices...)

stephanegigandet: Ah ok, we can have "Classic common wheat rusks", but then it should be under "Common wheat rusks" I think. And "Biscottes de froment" should not be a synonym for the "classic"

PR #10498, opened 2024-06-26
fix: make facets URLS /nova-group/unknown work

### LINKED ISSUE #10493: Make facets URLS /nova-group/unknown and /nova_group/not-applicable work
From @raphael0202 : "A strange bug that a contributor asked me about by mail:
on the NOVA group of the UK country: https://uk.openfoodfacts.org/nova-groups
there are 2 "Unknown" (one corresponding to "unknown" NOVA, one to "not-applicable" NOVA.
the counts are not the same depending between the tag list page and when we click on the tag (85k vs 1209) for "unknown" and 318 vs 0 for "not-applicable""

For the different "unknown" count, it's because by default, for tags fields, when we search for "unknown", we look for products with no value (e.g. for /category/unknown). But a few tags, like NOVA, have an "unknown" value, we need to disable this behavior for them.
(in function sub add_params_to_query ($request_ref, $query_ref) { )

We probably need to do the same for a few other fields like food_groups, Eco-Score and Nutri-Score.

4nt0ineB:
### What

Few tags like NOVA, Eco-Score and Nutri-Score can have an "unknown" value. Instead of looking for the presence or absence of the tag, we look for the value to be "unknown" (or not-applicable).

- Fixes #10493


lib/ProductOpener/Display.pm
```diff
@@ -4819,6 +4819,8 @@ sub add_params_to_query ($request_ref, $query_ref) {
 						if $log->is_debug();
 
 					# if the value is "unknown", we need to add a condition on the field being empty
+
+					my @tagtype_allowing_unknown_as_value = qw(nutrition_grades nova_groups ecoscore);
```
stephanegigandet, line 4823: Can you add pnns_groups_1, pnns_groups_2 and food_groups?

lib/ProductOpener/Display.pm
```diff
@@ -4831,7 +4833,7 @@ sub add_params_to_query ($request_ref, $query_ref) {
 							)
 						)
 						and ($tagtype !~ /^pnns_groups_/)
```
stephanegigandet, line 4835: pnns_groups_1 / 2 can be included in @tagtype_allowing_unknown_as_value

stephanegigandet (approved):
Thank you!

PR #10500, opened 2024-06-26
feat: add user in pending group when requests to join an org

### LINKED ISSUE #8781: Create a new "pending" group of organization members
### What
- When users create an account on OFF, they can indicate which organization they want to join.
- If the organization already exists, we currently don't add them to the organization, we just send an email to OFF producers admins.
- Instead we could create a "pending" group of users in each organization, and add users to it. That way we could then display the list of pending users to the org admins, and allow them to add or remove pending users to the organization.

### Part of
- #5521

4nt0ineB:
### What

- This feature is enabled for validated org
- The user is added to the 'pending' group of an org when he requests to join it.
- The org admin can accept or reject. 
- If an other org is requested by the user while waiting for approval, he is removed from the previous one.

- Fixes #8781


po/common/common.pot
```diff
@@ -7044,4 +7044,10 @@ msgctxt "open_in_crm"
 msgid "Open in CRM"
 msgstr "Open in CRM"
 
+msgctxt "deny"
+msgid "Deny"
+msgstr "Deny"
 
+msgctxt "users_requested_to_join_org"
+msgid "These users have requested to join the organization :"
```
stephanegigandet, line 7052:
```suggestion
msgid "These users have requested to join the organization:"
```

stephanegigandet (approved):
Thank you!

PR #10501, opened 2024-06-28
fix: various fixes to data sync with crm

4nt0ineB:
### What

- The salesperson user id was not properly matched
- Import type 'CSV' was overriden by undefined value
- Added 'SFTP' tag for imports that are fetched from sftp dir


lib/ProductOpener/Import.pm
```diff
@@ -2772,10 +2772,27 @@ sub import_csv_file ($args_ref) {
 	}
 
 	# sync CRM
+	my @csv_from_sftp_dir = qw(
+		https://www.carrefour.fr
+		https://www.intermarche.com/
+		https://bayard.com/
```
stephanegigandet, line 2778: Bayard is a GS1 Datapool, like Agena3000 and Equadis.

lib/ProductOpener/Import.pm
```diff
@@ -2772,10 +2772,26 @@ sub import_csv_file ($args_ref) {
...
+	my @csv_from_sftp_dir = qw(
+		https://www.carrefour.fr
+		https://www.intermarche.com/
+	);
+
 	foreach my $org_id (keys %{$stats_ref->{orgs_existing}}) {
 		update_import_date($org_id, $time);
 		if (exists $args_ref->{source_id}) {
-			update_last_import_type($org_id, uc($args_ref->{source_id}));
+			if ($args_ref->{source_id} eq 'agena3000') {
```
stephanegigandet, line 2783: Can you add Bayard there? It's similar to agena3000 and equadis

stephanegigandet (approved):

lib/ProductOpener/Import.pm
4nt0ineB: I add the GS1 tag ? @stephanegigandet

PR #10502, opened 2024-06-28
feat: prevent user from deletion if main contact or last admin of an org

4nt0ineB:
### What

Prevents deletion of the user's account if he is the main contact or only administrator of an organization.

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -6929,8 +6929,8 @@ msgid "You cannot remove the main contact. Change it first."
 msgstr "You cannot remove the main contact. Change it first."
 
 msgctxt "cant_delete_user_main_contact"
-msgid "This is user can't be deleted, he is the main of his organization"
-msgstr "This is user can't be deleted, he is the main of his organization"
+msgid "This is user can't be deleted, he is the main contact of his organization"
+msgstr "This is user can't be deleted, he is the main contact of his organization"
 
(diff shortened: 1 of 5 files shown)


po/common/common.pot
```diff
@@ -6928,6 +6928,14 @@ msgctxt "cant_delete_main_contact"
 msgid "You cannot remove the main contact. Change it first."
 msgstr "You cannot remove the main contact. Change it first."
 
+msgctxt "cant_delete_user_main_contact"
+msgid "This is user can't be deleted, he is the main contact of his organization"
```
stephanegigandet, line 6932:
```suggestion
msgid "This user can't be deleted, he is the main contact of his organization"
```

(same in other strings below)

po/common/fr.po
```diff
@@ -6934,4 +6934,36 @@ msgstr "Statut vérifié"
 
 msgctxt "creation_date"
 msgid "Date de création"
-msgstr "Date de création"
\ No newline at end of file
+msgstr "Date de création"
+
+msgctxt "enter_main_contact_username"
+msgid "Nom du contact principal :"
```
stephanegigandet, line 6940: the msgid should be the English text, otherwise it won't work
(same in strings below)

hangy:
This will conflict with Keycloak #9251 taking over user management

PR #10507, opened 2024-07-02
feat: a script to check if existing org can be sync with the crm

4nt0ineB:
### What

The script create a csv file that tells what will be matched in the crm and what will be created. 
It's like a verbose dry run of accepting the org.

### Screenshot
image


stephanegigandet (approved):
Awesome

PR #10510, opened 2024-07-03
feat: add filters to orgs table

TheSussex:
### What

1. Added the datatables script to the already existing template that displays organisations table
2. Added filter options (name and verified status) to the table
3. Added search form to take values of the query params so users won't have to manually edit the url

### Screenshot
Screenshot 2024-07-04 at 5 45 27 PM


stephanegigandet:
Looks good, thanks @TheSussex 

Can you make the "reviewed" field a dropddown, like in the org edit form?
image

with also the option to see orgs with all reviewed status

stephanegigandet (approved):
Thank you!

PR #10513, opened 2024-07-04
feat: add an optional condition for routes to be used or not

### LINKED ISSUE #10508: New routing refactor issue with texts matching other route (e.g. /labels)
In production we have (bogus) text files in lang/ar/ : labels.html , categories.html etc.

With the new routing introduced in https://github.com/openfoodfacts/openfoodfacts-server/pull/10432 , those texts register the "labels", "categories" etc. route patterns and associate them with the text_route

It prevents URLs like http://es.openfoodfacts.localhost/labels to work properly.

We could try to make the text route less greedy, so that the patterns are language specific.

4nt0ineB:
### What

Added a dynamic routing mechanism, by allowing to provide a condition that will need to be true for the route to be used after being matched  - or else to pass to the remaining regex patterns. 
It uses an anonymous sub that'll be called during the route match, with the request's context as parameters.
For example, this feature makes it possible to hide certain routes for some languages and allow them for others.

- Fixes #10508

### DIFF
--- a/lib/ProductOpener/Routing.pm
+++ b/lib/ProductOpener/Routing.pm
@@ -94,7 +94,8 @@ a route is registered with:
 		  Use named captures to store the arguments in $request_ref->{param}
 
 		- onlyif: a sub($request_ref, @components) that will be called to check if the route should be used
-			Its a dynamic routing, using context of the request
+			Its a dynamic routing, using context of the request.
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Looks good, thank you

PR #10523, opened 2024-07-08
feat: generate historical events from product revisions

4nt0ineB:
### What

Add script that goes through the product revisions and generate historical events
`scripts/product_revision_to_historical_events.pl`

### Issues
- #10526


scripts/product_revision_to_historical_events.pl
```diff
@@ -0,0 +1,106 @@
...
+			$action = 'deleted';
+		}
+
+		print $file encode_json(
+			{
+				ts => $change->{t},
+				barcode => $product->{code},
+				userid => $change->{userid},
+				comment => $change->{comment},
+				flavor => $options{current_server},
```
stephanegigandet, line 76: Can we add product_type too?

lib/ProductOpener/Redis.pm
```diff
@@ -131,6 +131,7 @@ sub push_to_redis_stream ($user_id, $product_ref, $action, $comment, $diffs) {
 				# We let Redis generate the id
 				'*',
 				# fields
+				'ts', $timestamp,
```
stephanegigandet, line 134: Maybe just "t" ? we have been using _t as a suffix for timestamps, or maybe "timestamp" is more explicit

scripts/product_revision_to_historical_events.pl
john-gom: Can I assume that this means "flavor" is being deprecated?

4nt0ineB: That's my understanding yes
 [fix: add product_type to redis events (#10550)](https://github.com/openfoodfacts/openfoodfacts-server/pull/10550#issue-2409093248)

scripts/product_revision_to_historical_events.pl
```diff
@@ -0,0 +1,118 @@
...
+				barcode => $product->{code},
+				userid => $change->{userid},
+				comment => $change->{comment},
+				product_type => $options{product_type},
+				action => $action,
+				diffs => $change->{diffs}
+			}
+		) . "\n";
+
+		# push_to_redis_stream(
```
john-gom, line 94: At the moment off-query will fetch a product from MongoDB and update all of its data for every event it receives, which will mean that the initial import will take a very long time. Is there some way that we can flag these updates so that I can just load the event and not do the product update?

4nt0ineB: I guess I can add a flag in the diffs, like `diffs->{initial_import} = 1` ?

john-gom:
I've updated the branch to push to redis and not the JSON file and am testing it locally with a full set of products that is a few months old. Based on current progress it looks like it's going to take about 50 hours to do everything. off-query seems to be keeping up, so the Redis queue length is averaging around 10.

If we can't optimise significantly then we may need to do a cleanup of duplicate events (as we will be receiving live updates to products while the import is running).

scripts/product_revision_to_historical_events.pl
```diff
@@ -0,0 +1,135 @@
...
+	my $rev = 0;    # some $change don't have a 'rev'
+	foreach my $change (@{$changes}) {
+
+		$rev++;
+
+		my $action = 'updated';
+		if ($rev eq 1) {
+			$action = 'created';
+		}
+		elsif ( $rev == $product->{rev}
```
john-gom, line 64: On a few products get:
`Use of uninitialized value in numeric eq (==) at scripts/product_revision_to_historical_events.pl line 64.`
I'm guessing the product doesn't have a rev. Maybe would be safer to get the count of $changes at the start and then assume we are on the last one where $rev == $change_count?

lib/ProductOpener/Display.pm
```diff
@@ -4020,6 +4020,15 @@ HTML
 						;
 				}
 
+				if ($packager_codes{$canon_tagid}{cc} eq 'ie') {
```
john-gom, line 4023: I'm not sure how this Ireland packager code branch got mixed up in your PR. Any ideas?

scripts/product_revision_to_historical_events.pl
4nt0ineB, line 64: Yes I agree 
```suggestion
    my $change_count = @{$changes};
	my $rev = 0;    # some $change don't have a 'rev'
	foreach my $change (@{$changes}) {
		$rev++;
		my $action = 'updated';
		if ($rev eq 1) {
			$action = 'created';
		}
		elsif ( $rev == $change_count
```

4nt0ineB: Yes I agree

lib/ProductOpener/Display.pm
4nt0ineB: No idea, my branch was clean locally, until I pull now.

scripts/product_revision_to_historical_events.pl
```diff
@@ -0,0 +1,135 @@
...
+			$change->{comment},
+			$change->{diffs},
+			$change->{t}
+		);
+	}
+
+	return 1;
+}
+
+# because getting products from mongodb won't give 'deleted' ones
```
stephanegigandet, line 109: It's quite possible that when we run this in production on tens of millions of files, that it will take a long time, and possibly fail (e.g. killed by the kernel).

In which case, we will have to re-run it, which would cause events to be sent multiple times.

Could we add some kind of mechanism to avoid that? (e.g. storing somewhere the last processed product + revision, and using that to skip already processed changes?)

scripts/product_revision_to_historical_events.pl
```diff
@@ -28,42 +28,55 @@
...
 	# JSONL
+	my $code = product_id_from_path($path);
+	my $change_count = @$changes;    # some $product don't have a 'rev'
 	my $rev = 0;    # some $change don't have a 'rev'
-	foreach my $change (@{$changes}) {
 
+	foreach my $change (@{$changes}) {
 		$rev++;
 
+		my $bop = retrieve($path . "/$rev.sto");
```
stephanegigandet, line 65: $bop does not seem used anywhere? is that a line you forgot to remove?

scripts/product_revision_to_historical_events.pl
```diff
@@ -119,17 +134,44 @@ sub find_products {
...
 		if (-d $file_path) {
-			find_products($file_path,"$code$entry");
+			find_products($file_path, "$code$entry");
 			next;
 		}
+
 		if ($entry eq 'product.sto') {
-			process_file($dir);
-		} 
+			if ($can_process or ($last_processed_path and $last_processed_path eq $dir)) {
```
stephanegigandet, line 142: readdir() can return files in any order, so you need to add a sort in the foreach my $entry (@files) loop

this would also allow to restart without exploring all the paths again, you could just check if the path matches the start of the path of the last processed product. That way if the last one is 808/400/300/2001, you don't need to go inside all the directories from 001 to 807

scripts/product_revision_to_historical_events.pl
```diff
@@ -0,0 +1,221 @@
...
+use ProductOpener::PerlStandards;
+use utf8;
+
+use ProductOpener::Config qw/%options $query_url/;
+use ProductOpener::Store qw/store retrieve/;
+use ProductOpener::Paths qw/%BASE_DIRS/;
+use ProductOpener::Redis qw/push_to_redis_stream/;
+use ProductOpener::Products qw/product_id_from_path/;
+use Path::Tiny;
+use JSON;
```
hangy, line 32: Should this be

```suggestion
use JSON::MaybeXS;
```

per #10578?

scripts/product_revision_to_historical_events.pl
```diff
@@ -0,0 +1,221 @@
+#!/usr/bin/perl -w
+
+# This file is part of Product Opener.
+#
+# Product Opener
+# Copyright (C) 2011-2023 Association Open Food Facts
```
hangy, line 6:
```suggestion
# Copyright (C) 2011-2024 Association Open Food Facts
```
😉

PR #10530, opened 2024-07-09
feat: add new product event types for redis queue

### LINKED ISSUE #10527: Add the "creation" event type in product events Redis queue
A product events Redis queue exists but does not differentiate simple edit from product creation. 
We should add this to have finer insights.

4nt0ineB:
### What

- Added event types : created / imported / archived 

- Fixes #10527


lib/ProductOpener/Products.pm
```diff
@@ -1501,12 +1501,25 @@ sub store_product ($user_id, $product_ref, $comment) {
 
 	$log->debug("store_product - done", {code => $code, product_id => $product_id}) if $log->is_debug();
 
-	my $update_type = $product_ref->{deleted} ? "deleted" : "updated";
+	my $action = "updated";
+
+	if (exists $product_ref->{obsolete}
+		and $product_ref->{obsolete} eq "on")
+	{
+		$action = "archived";
```
stephanegigandet, line 1509: It's a bit tricky: "archived" products can be "unarchived", and they can also be "updated" (without the obsolete parameter necessarily changing).

I think we should trigger the "archived" event only when the obsolete field is changed to "on", and that we should add an "unarchived" event if it's removed.

lib/ProductOpener/Products.pm
```diff
@@ -1501,12 +1501,25 @@ sub store_product ($user_id, $product_ref, $comment) {
...
+	}
+	if ($rev == 1) {
+		$action = defined $product_ref->{owner} ? "imported" : "created";
```
stephanegigandet, line 1512: I don't remember what's the use case for "imported"? Why don't we use "created" even for products created through the producers platform?

4nt0ineB: It was a requirement defined during the brainstorm, and I simply guessed it was about the pro platform. But since we have a flavor field already

stephanegigandet (approved):
Looks good, thanks

PR #10531, opened 2024-07-10
feat: allow orgs to not have a main contact if imported from an aggregator

4nt0ineB:
### What

- Added a script to remove orgs main contact that are not regular users. By default last migration, during the CRM integration, used the creator of the org to set the main contact. This caused the main contact to be agena3000/equadis/.. or even empty for some orgs.
- Allow orgs to not have a main contact if it was imported from an aggregator such as agena3000 or equadis
- New members can still be added to the orgs later ; it will be synced with the crm if the orgs has been accepted.

This script must be run in prod: `scripts/migrations/2024_07_change_main_contact.pl`

### DIFF
--- a/lib/ProductOpener/Orgs.pm
+++ b/lib/ProductOpener/Orgs.pm
@@ -175,28 +175,30 @@ sub store_org ($org_ref) {
 
 		# We switched to validated, update CRM
 		my $main_contact_user = $org_ref->{main_contact};
-		my $user_ref = retrieve_user($main_contact_user);
 
 		eval {
-			my $contact_id = find_or_create_contact($user_ref);
-			defined $contact_id or die "Failed to get contact";
-			$user_ref->{crm_user_id} = $contact_id;
-			store_user($user_ref);
+			my $partner_id;
+			if (defined $main_contact_user) {
+				my $user_ref = retrieve_user($main_contact_user);
+				$partner_id = find_or_create_contact($user_ref);
+				defined $partner_id or die "Failed to get contact";
+				$user_ref->{crm_user_id} = $partner_id;
+				store_user($user_ref);
+			}
 
-			my $company_id = find_or_create_company($org_ref, $contact_id);
+			my $company_id = find_or_create_company($org_ref, $partner_id);
 			defined $company_id or die "Failed to get company";
 
-			defined add_contact_to_company($contact_id, $company_id) or die "Failed to add contact to company";
+			if (defined $partner_id) {
+				defined add_contact_to_company($partner_id, $company_id) or die "Failed to add contact to company";
+			}
 
-			# admin validates the org, used to link the right salesperson
+			# The off admin who validates the org is the salesperson in crm
 			my $my_admin = retrieve_user($User_id);
 			$log->debug("store_org", {myuser => $my_admin}) if $log->is_debug();
 
-			my $opportunity_id = create_onboarding_opportunity(
-				"$org_ref->{name} - new",
-				$company_id, $user_ref->{crm_user_id},
-				$my_admin->{email}
-			);
+			my $opportunity_id
+				= create_onboarding_opportunity("$org_ref->{name} - new", $company_id, $partner_id, $my_admin->{email});
 			defined $opportunity_id or die "Failed to create opportunity";
 
 			$org_ref->{crm_org_id} = $company_id;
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Orgs.pm
```diff
@@ -175,28 +175,30 @@ sub store_org ($org_ref) {
...
 
 		eval {
-			my $contact_id = find_or_create_contact($user_ref);
-			defined $contact_id or die "Failed to get contact";
-			$user_ref->{crm_user_id} = $contact_id;
-			store_user($user_ref);
+			my $partner_id;
+			if (defined $main_contact_user) {
+				my $user_ref = retrieve_user($main_contact_user);
+				$partner_id = find_or_create_contact($user_ref);
```
alexgarel, line 183: Why don't you do it only if $user_ref{crm_user_id} is undef ?
```suggestion
				$partner_id = $user_ref->{crm_user_id} // find_or_create_contact($user_ref);
```

scripts/migrations/2024_07_change_main_contact.pl
```diff
@@ -0,0 +1,54 @@
...
+
+# This file is used
+# to change the main_contact field of orgs that are not users
+# to an undef value
+
+use ProductOpener::Store qw/store/;
+use ProductOpener::Orgs qw/list_org_ids retrieve_org/;
+use ProductOpener::Paths qw/%BASE_DIRS/;
+
+my @not_users = qw(agena3000 equadis codeonline bayard database-usda countrybot);
```
alexgarel, line 34: Don't you want to add any @openfoodfacts.org user ?

scripts/migrations/2024_07_change_main_contact.pl
```diff
@@ -0,0 +1,54 @@
...
+
+foreach my $org_id (list_org_ids()) {
+	my $org_ref = retrieve_org($org_id);
+
+	# if main_contact is empty, blank or contains multiple blank or invisble characters or is equal to one of the bad contacts, set it to empty
+	if (
+		not defined $org_ref->{main_contact}    # undefined
+		or $org_ref->{main_contact} =~ /^\s*$/    # empty or only whitespace
+		or $org_ref->{main_contact} =~ /[\p{Z}\p{C}]/    # contains unicode whitespace or control characters
+		or grep {$org_ref->{main_contact} eq $_} @not_users
```
alexgarel, line 44: It's better to parenthesis maybe (it's what we generally do, better be consistent)
```suggestion
		(not defined $org_ref->{main_contact})    # undefined
		or ($org_ref->{main_contact} =~ /^\s*$/)    # empty or only whitespace
		or ($org_ref->{main_contact} =~ /[\p{Z}\p{C}]/)    # contains unicode whitespace or control characters
		or (grep {$org_ref->{main_contact} eq $_} @not_users)
```

scripts/migrations/2024_07_change_main_contact.pl
```diff
@@ -0,0 +1,54 @@
...
+		)
+	{
+
+		$org_ref->{main_contact} = undef;
```
alexgarel, line 48: In the case of an empty contact, don't you wan't to see if there are people in the admins group ? (and maybe, just log for now, and report this to producers team).

alexgarel (approved):
Some small points to maybe fix @4nt0ineB 

I already approve though so you can merge when it's ok.

lib/ProductOpener/Orgs.pm
4nt0ineB: Sure we can, I just have to add a way to verify that it is a valid odoo record because this field can be changed manually by off admins in the user form. I'll do this in a new PR.

scripts/migrations/2024_07_change_main_contact.pl
4nt0ineB: Yes good point.

scripts/migrations/2024_07_change_main_contact.pl
4nt0ineB: Ah thanks, I completely forgot, that's what I wanted to do initially :)

scripts/migrations/2024_07_change_main_contact.pl
```diff
@@ -0,0 +1,83 @@
...
+
+		# take the first admin as main contact if available
+		if (defined $org_ref->{admins} and @{$org_ref->{admins}}) {
+			# find the first admin that is not in the list of users that are not users
+			# and set it as main contact
+
+			my $admin = undef;
+			foreach my $admin_id (@{$org_ref->{admins}}) {
```
stephanegigandet, line 66: @{$org_ref->{admins} is a hash, changing to sort keys %{$org_ref->{admins}}

Issue #10554, opened 2024-07-16
Import GS1 GDSN data format from Alnatura GmbH


stephanegigandet:
We received a sample GDSN message from Alnatura, which is a bit different from Equadis. We need to change the GS1.pm module to support it.

First differences spotted:
- a lot of fields are specified without the module name, e.g. foodAndBeverageIngredientModule vs food_and_beverage_ingredient:foodAndBeverageIngredientModule
- some fields can be arrays where we expect a hash:
 "transaction" : {
         "documentCommand" : {
            "catalogue_item_notification:catalogueItemNotification" : [

PR #10556, opened 2024-07-16
fix: GDSN import from Alnatura

stephanegigandet:
Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/10554 :

GDSN XML files can be formatted in different ways, leading to different JSON files after conversion.

This PR:
- adds support for multiple notification inside a command
- adds support for tags that don't include the namespace


tests/unit/expected_test_results/import_gs1/agena3000_andros.json
alexgarel, line 1: would it be possible to sort names in hashmaps to avoid those changes ?

alexgarel (approved):
LGTM

tests/unit/expected_test_results/import_gs1/agena3000_andros.json
stephanegigandet: @alexgarel we use https://metacpan.org/pod/XML::XML2JSON to convert the GS1 XML files to JSON, but I don't see an option to sort the keys... 

I regenerated the XML by mistake in fact, the test works on the converted JSON.

PR #10560, opened 2024-07-17
fix: pro_moderator_owner not stored for the admin/moderator user through org/[orgid] facet

4nt0ineB:
### What

Fixes this problem:

Pro moderator and admins experienced issues with the org/[orgid] facet in the url. The feature that allows them to navigate and act as they are part of the org is not properly working.


alexgarel (approved):

alexgarel:
Pro moderator and admins experienced issues with the org/[orgid] facet in the url. The feature that allows them to navigate and act as they are part of the org is not properly working.

PR #10572, opened 2024-07-21
taxonomy: cereal bars

moon-rabbitOFF:


moon-rabbitOFF:
@stephanegigandet I would like to remove the intermediate category "en:Cereal bar with almonds or hazelnuts" which I don't find useful. Should I put the ciqual link to the parent of the children categories? I put it with the parent category now

stephanegigandet (approved):
Thanks

stephanegigandet:
@moon-rabbitOFF you can remove the "almonds or hazelnuts" and put the properties on both children, + on the parent (but as a proxy as it could be other nuts)

PR #10577, opened 2024-07-22
ci: on demand PR actions

alexgarel:
Adding some actions that can be triggered from PR comments...


stephanegigandet (approved):
Nice!

alexgarel:
@stephanegigandet still in progress :-) 
It will work only when merged to main.

PR #10578, opened 2024-07-22
refactor: use JSON::MaybeXS instead of JSON::PP

stephanegigandet:
In the past we had issues with the JSON::XS module, so we we switched to JSON::PP for safety.
Maybe things work better with Cpanel::JSON::XS, so trying to replace all instances with https://metacpan.org/pod/JSON::MaybeXS (which loads Cpan::JSON::XS or JSON::XS if present)


alexgarel (approved):
:crossed_fingers:

hangy (approved):

stephanegigandet:
One difference is that integers are changed from 0 and 4 to 0.0 and 4.0 in API results:

```
-   "ingredients_with_unspecified_percent_sum" : 100,
+   "ingredients_with_unspecified_percent_sum" : 100.0,

-            "fruits_vegetables_legumes" : 0,
+            "fruits_vegetables_legumes" : 0.0,
```

PR #10580, opened 2024-07-22
build: fix working tree removing shared-services

alexgarel:
the deps subdir was wrongly added to the repository.


hangy (approved):

stephanegigandet (approved):
Thanks!

PR #10588, opened 2024-07-24
fix: org name input field not displaying when checkbox is checked in pro pl…

4nt0ineB:
…atform

### What

This pr fixes that issue:
The org name input field is not displaying on pro platform even if the checkbox is checked.

### Screenshot
Peek 2024-07-24 11-23


stephanegigandet (approved):
Thanks!

PR #10599, opened 2024-07-26
feat: add improvement to product navbar

### LINKED ISSUE #10357: Add an entry in product navigation to jump to product improvements
### Problem

We display product improvement opportunity in the producer platform to help producers have better products.
The corresponding knowledge panel is displayed in the middle of the page and not easy to reach

### Proposed solution
Add a "improvement" entry in the navbar that scrolls to the improvements knowledge panel 

### Code pointers

It should be easy to do, it's about adding the entry if the knowledge panel exists and have the right id on the panel for navigation to work.

### Mockups

image

TheSussex:
### What
Updated the product navbar with improvements entry, this scrolls to the improvements knowledge panel

- Fixes #10357


alexgarel:
My comment above `/update_tests_results` does not work because right now it's implemented only for this repository (not branches comming from external repositories).

So @TheSussex you have to `make update_tests_results`.

stephanegigandet (approved):
Looks good, thank you!

PR #10610, opened 2024-07-26
taxonomy: add Norwegian translations for ingredients analysis

tkbremnes:
### What

Adds Norwegian translations for no, nn and nb.


tkbremnes:
I was basing this on a fairly ancient local version with a massive merge conflict. Will re-apply and resubmit later :-)

PR #10615, opened 2024-07-29
taxonomy: German soy protein product / concentrated

stephanegigandet:
Used by a German manufacturer who sends us data


teolemon (approved):
Is de: Sojaeiweiß otherwise recognized ?

PR #10627, opened 2024-07-30
taxonomy: paprika for color

stephanegigandet:
Handles more synonyms used in the US for paprika used as a colour


benbenben2 (approved):
LGTM

PR #10630, opened 2024-08-03
taxonomy: add and fix italian taxonomy terms

McCio:
Added various Italian versions for taxonomy elements


stephanegigandet (approved):
Awesome, thank you @McCio

taxonomies/food/ingredients.txt
```diff
@@ -27568,8 +27658,6 @@ yi: מעל
 zh: 麵粉
 carbon_footprint_fr_foodges_ingredient:fr: Farine de blé
 carbon_footprint_fr_foodges_value:fr: 1.2
-ciqual_proxy_food_code:en: 9410
-ciqual_proxy_food_name:fr: Farine de blé tendre ou froment T110
```
alexgarel, line 27572: Why did you remove this ?

taxonomies/food/ingredients.txt
```diff
@@ -27714,8 +27802,6 @@ vi: Bột mì
 zh: 小麥麵粉
 carbon_footprint_fr_foodges_ingredient:fr: Farine de blé
 carbon_footprint_fr_foodges_value:fr: 1.2
-ciqual_proxy_food_code:en: 9410
-ciqual_proxy_food_name:fr: Farine de blé tendre ou froment T110
```
alexgarel, line 27718: Why did you remove this ?

alexgarel (changes requested):
@McCio thank you so much for this PR, but can you see why you removed the ciqual proxy for two entries ?

taxonomies/food/ingredients.txt
McCio: Because the generic "flour" is not the specific french flour type 110, and this causes all types of flour to inherit that specific type (when a more specific ciqual code is not defined)

taxonomies/food/ingredients.txt
McCio: The same here for the wheat flour

taxonomies/food/ingredients.txt
stephanegigandet: @McCio This is why the property is "ciqual_proxy_food_code" and not "ciqual_food_code": we know it's not the specific type, but it is a reasonable enough proxy that we can use for some applications. It's important to keep the proxies so that we can estimate nutrients etc.

McCio: Thanks for explaining and for the patience, @stephanegigandet and @alexgarel, now I understand the logic behind it.
Fixed and rebased all over the updated main

stephanegigandet (approved):
Thank you very much @McCio !

McCio:
@stephanegigandet I checked why the tests were failing, and I did remove another ciqual proxy food code.
Now I've [added it back](https://github.com/openfoodfacts/openfoodfacts-server/compare/20ea9b7fb95e15c539973356f51c93f9fc2bf5ab..f4c14b53f41df5e4ede39efff62aa6e1750158ea) and the tests pass again (locally at least)

aleene (commented):
LGTM

alexgarel:
Thanks @McCio for your work ! I'm launching the test locally.

FYI @stephanegigandet AFAIK the PR actions do not work on PRs from forked repository (I'm not sure if it's because of the way I wrote them or a limitation of PR actions, but I thinks it's the later)

stephanegigandet:
For the record, current stats for Italian ingredients parsing (before this PR):

https://it.openfoodfacts.org/ingredients?stats=1

Type	Unique tags	Occurrences
[known](https://it.openfoodfacts.org/ingredients?status=known)	2562 (4.22%)	418180 (82.81%)
[unknown](https://it.openfoodfacts.org/ingredients?status=unknown)	58180 (95.78%)	86807 (17.19%)
[all](https://it.openfoodfacts.org/ingredients?stats=1)	60742 (100.00%)	504987 (100.00%)

PR #10635, opened 2024-08-04
taxonomy: add “from” Danish+Swedish translation

Freso:
### What

Added translations of <q>from</q> (`ProductOpener::Ingredients::from`) to Danish and Swedish.


stephanegigandet (approved):
Thanks

PR #10648, opened 2024-08-06
feat: added a geoip api endpoint

### LINKED ISSUE #10594: Expose the GeoIP API for the mobile app
Hi everyone,

In the OFF mobile application, we’d like to identify the user’s country based on the IP (it’s not perfect, but it’s good enough).
I know a GeoIP implementation is used on the server.

Would it be possible to expose it as an API?

4nt0ineB:
### What

- Expose a GeoIP API endpoint
- some refactor in the API.pm 

### Screenshot
image

- Fixes #10594

### DIFF
--- a/lib/ProductOpener/API.pm
+++ b/lib/ProductOpener/API.pm
@@ -73,6 +73,7 @@ use ProductOpener::KnowledgePanels qw/create_knowledge_panels initialize_knowled
 use ProductOpener::Ecoscore qw/localize_ecoscore/;
 use ProductOpener::Packaging qw/%packaging_taxonomies/;
 use ProductOpener::Permissions qw/has_permission/;
+use ProductOpener::GeoIP qw/get_country_for_ip_api/;
 
 use ProductOpener::APIProductRead qw/read_product_api/;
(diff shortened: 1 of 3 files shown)


4nt0ineB:
The endpoint is totally public, is that ok ?
I put the endpoint handler in the GeoIP.pm, would it be better to have a dedicated module ? There are already APIProduct(Read/Revert/Services), etc.

stephanegigandet:
@4nt0ineB what is the need for the GeoIP endpoint? Who is going to use it? Is it only for internal use (if so, which project?).

teolemon:
There's a link to the issue: https://github.com/openfoodfacts/openfoodfacts-server/issues/10594

lib/ProductOpener/API.pm
```diff
@@ -384,6 +385,43 @@ Reference to the request object.
 
 =cut
 
+# Dipatch table for API actions
+my $dispatch_table = {
```
alexgarel, line 389: Great refactor :+1:

alexgarel (commented):
LGTM

alexgarel (approved):

PR #10652, opened 2024-08-07
fix: intermarche import

stephanegigandet:
Fixing an issue that prevented importing files from Intermarché


teolemon (commented):
I was discussing retailers yesterday evening with @manon-corneille. do we have over data imports broken or pending? @stephanegigandet

alexgarel (approved):
LGTM, thanks for documenting the change.

PR #10662, opened 2024-08-08
feat: taxonomize brands with a language less xx: taxonomy - proof of concept

### LINKED ISSUE #5208: Make brands a taxonomized field
### What

Primary use cases:
- Canonicalize brand names that can be abbreviated/spelled in different ways (e.g. Marks&Spencer / Marks and Spencer, McCain / Mc Cain, St Mamet / Saint Mamet) to a unique id
- Display properly capitalized / accented brand names in the /brands facet

Secondary use cases:
- possibly add properties to brands
- possibly make sub brands children of the parent brand (but only for brands, not owner companies etc.)

### Part of
- #8865 

### Screenshot
image

stephanegigandet:
This PR is a proof of concept to make brands a taxonomized field using a language less taxonomy (all entries use the xx: prefix).

To taxonomize existing products:

```
~/openfoodfacts-server$ docker exec -it po_off-backend-1 bash
www-data@a01d17ec9489:/opt/product-opener$ ./scripts/update_all_products.pl --fields brands 
```

Fixes #5208

This PR is a proof of concept, it lacks tests, documentation etc.

Also the "brands" taxonomy is hardcoded in Tags.pm, it would probably be better to make this configurable in case we want to do something similar with other taxonomies in the future.

### DIFF
--- a/taxonomies/brands.txt
+++ b/taxonomies/brands.txt
@@ -1,0 +1,12 @@
+# All brands in the brands taxonomy are language less and use the xx: prefix
+# Brands are not translated, we record what is written on the package, and we display it back as-is
+
+xx: Aldi
+
+xx: Marks & Spencers, Marks and Spencers
+
+xx: Marque Repère
+
+xx: Nestlé
+
+xx: Trader Joe's
--- a/lib/ProductOpener/Config_off.pm
+++ b/lib/ProductOpener/Config_off.pm
@@ -726,6 +726,7 @@ $options{replace_existing_values_when_importing_those_tags_fields} = {
 	periods_after_opening
 	data_quality data_quality_bugs data_quality_info data_quality_warnings data_quality_errors data_quality_warnings_producers data_quality_errors_producers
 	improvements
+	brands
 );
 
 # tag types (=facets) that should be indexed by web crawlers, all other tag types are not indexable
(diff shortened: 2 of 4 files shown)


aleene:
Sounds good. Anything to kickstart this taxonomy. And we can grow from there.

taxonomies/brands.txt
```diff
@@ -0,0 +1,12 @@
+# All brands in the brands taxonomy are language less and use the xx: prefix
+# Brands are not translated, we record what is written on the package, and we display it back as-is
+
+xx: Aldi
```
raphodn, line 4: so we could add a property like `wikidata:en:Q125054` below ?

aleene: Sounds ok as it is a standard element.

stephanegigandet: yes, because https://www.wikidata.org/wiki/Q125054 is for both Aldi

we would need to be careful with brands that use the same name (e.g. "Pacific" for alcohol and for other stuff: https://world.openfoodfacts.org/brand/pacific ).

stephanegigandet:
This branch is deployed on the .dev server (login and pass: off )

https://world.openfoodfacts.dev/brand

You can edit products etc. on https://world.openfoodfacts.dev/ to test the new behavior of the brands taxonomy.

teolemon:
@stephanegigandet we could update test results to reduce amount of conflicts ?

alexgarel (approved):
ok for me

stephanegigandet:
replaced by https://github.com/openfoodfacts/openfoodfacts-server/pull/11606

PR #10667, opened 2024-08-09
feat: Variations of strawberries and sweet potatoes

teolemon:
### What
- Found on the wiki, and started reformatting
- feat: Variations of strawberries and sweet potatoes
- Could also be interesting to add to ingredients @aleene


stephanegigandet:
We should discuss it in #taxonomies, but I think it would be a good idea to add some criterias for the eligibility of taxonomy entries and document them on https://wiki.openfoodfacts.org/Global_categories_taxonomy (and at the top of the taxonomy files)

I think a good rule would be to create taxonomy entries (categories, ingredients, packagings etc.) only if we have corresponding products. Empty taxonomy entries are problematic for multiple reasons:
- they have an impact on performance (building and loading time for taxonomies, memory footprint etc., taxonomy suggestions etc.)
- they create unneeded work (e.g. people will add translations for taxonomies that are in fact not used)
- they create noise (in taxonomy suggestions, in lists of children taxonomies etc.)
- they create risks of unwanted matchings (e.g. all the first names and place names used as synonyms in all the strawberry varieties). Especially in some categories used in parsing like ingredients or labels. If there are place names as synonyms in the ingredients taxonomy, this will break a lot of things.

stephanegigandet:
Referenced the discussion in #taxonomies : https://openfoodfacts.slack.com/archives/C02VDSWHT/p1723217151433759

raphael0202:
The category taxonomy is also used (extensively) in Open Prices to refer to raw product categories. So I would use as a rule, either we have a product on Open Food Facts or a raw price on Open Prices.

Freso (changes requested):
The PR as it is just outright removes Blackcurrants (and Camu-Camu) from categories, which is probably not desired.

I also don’t think the “X variety” and “X” are synonyms we need to have to start off with. As Stephane noted, those seem like prime candidates from name clash conflicts. I do like the idea of the PR though.

teolemon:

raphael0202:
In my opinion, a taxonomy PR that adds 2700 lines cannot be reviewed thoroughly by anyone.
I would advise closing this PR, and creating new PRs on a smaller scale.

PR #10671, opened 2024-08-09
taxonomy: Have `grape sugar` inherit from `glucose` rather than `monosaccharide`

Freso:
### What

Have `grape sugar` inherit from `glucose` rather than `monosaccharide`.

Grape sugar is always glucose (and in many languages the two terms are mostly interchangeable) and glucose has the `vegan` and `vegetarian` flags set, while monosaccharide does not.

Glucose also inherits from monosaccharide, so grape sugar should still inherit anything from there that is needed for classifications.

### Screenshot

[unknown vegan status](https://se.openfoodfacts.org/product/4056489412236/vitasia-rice-cracker-spicy-sweet-chili-flavour-lidl#health)

- https://se.openfoodfacts.org/product/4056489412236/vitasia-rice-cracker-spicy-sweet-chili-flavour-lidl#health

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -31355,7 +31355,7 @@ nl: biologische kokosbloesem siroop
 < en:coconut sugar
 es: nectar del brote de la flor del coco
 
-< en:monosaccharide
+< en:glucose
 en: grape sugar, raisin sugar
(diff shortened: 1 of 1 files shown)


benbenben2 (approved):
LGTM. Thanks @Freso

PR #10673, opened 2024-08-11
taxonomy: Add Swedish terms for hydrolysed soy protein and ginger extract

Freso:
### What

Adds Swedish terms for hydrolysed soy protein and ginger extract.

I added “hydrolyserat” both to `food/ingredients.txt` and to `ingredients_processing.txt`. Not sure if I should only do the latter, or if it’s fine to have both since “hydrolysed soy protein” is already in `food/ingredients.txt`.

### Screenshot

[unknown ingredients](https://se.openfoodfacts.org/product/7340083469077/heta-chilin%C3%B6tter-eldorado#health)

- https://se.openfoodfacts.org/product/7340083469077/heta-chilin%C3%B6tter-eldorado#health

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -33386,6 +33386,7 @@ hu: hidrolizált szójafehérje
 it: proteine di soia idrolizzate
 nl: gehydroliseerd soja-eiwit, gehydrolyseerd soja-eiwit
 pl: hydrolizat białka sojowego, hydrolizowane białko sojowe
+sv: hydrolyserat sojaprotein
 
 < en:soy protein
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Thank you

stephanegigandet:
> I added “hydrolyserat” both to `food/ingredients.txt` and to `ingredients_processing.txt`. Not sure if I should only do the latter, or if it’s fine to have both since “hydrolysed soy protein” is already in `food/ingredients.txt`.

We introduced the processings relatively recently, so we have entries in ingredients that could be covered by processings. We'll probably gradually remove them.

PR #10679, opened 2024-08-12
feat: make improvements to org table

TheSussex:
### What
- added responsiveness to the table by making it horizontally scrollable.
- added custom visibility option, so users can choose which column they want to see/not see on the table.
- updated `templates/web/common/site_layout.tt.html` file to include a link only visible to admins and moderators in the homepage menu, this link leads to the organization table page.
- updated `/cgi/display_org_table.pl` file to persist search queries values in the form after submission. 
- replaced `creator` field with `creator_email`
- created a script that populates the creator email for existing orgs

### Screenshot
Screenshot 2024-08-12 at 9 33 25 AM

Screenshot 2024-08-12 at 9 33 41 AM


cgi/display_org_table.pl
```diff
@@ -73,22 +76,48 @@
...
 $request_ref->{scripts} .= <<SCRIPTS
 <script src="https://static.openfoodfacts.org/js/datatables.min.js"></script>
+<script src="https://cdn.datatables.net/buttons/2.3.6/js/dataTables.buttons.min.js"></script>
+<script src="https://cdn.datatables.net/buttons/2.3.6/js/buttons.colVis.min.js"></script>
 SCRIPTS
 	;
 
 $request_ref->{header} .= <<HEADER
 <link rel="stylesheet" href="https://static.openfoodfacts.org/js/datatables.min.css">
+<link rel="stylesheet" href="https://cdn.datatables.net/buttons/2.3.6/css/buttons.dataTables.min.css">
```
hangy, line 103: Could you create a new version of the [datatables.css](https://github.com/openfoodfacts/openfoodfacts-server/blob/1cd785affc00db64a8d27cd7d85e9d726bec5d92/html/js/datatables.css) and so on, which include the buttons? Alternatively, move DataTables to be downloaded from `npm` and bundled with `webpack`.  We should ensure we use a consistent version, and we want to avoid using CDNs our assets.

TheSussex: I'm not sure how to go about this @hangy

hangy: @TheSussex When you look at https://github.com/TheSussex/openfoodfacts-server/blob/9e1dd240ce9a1b4e51e7a3dcd8ac6f03aeaaf57b/html/js/datatables.css#L7-L10, you can see that the datatables we currently publish with some specific settings. If you go to that link, you can select the buttons component to be bundled, too. The simplest version would be to simply download the modified CSS and JS files from the datatables site and then replace the ones currently in git.

TheSussex: I did this and it's not showing the column visibility button anymore. Am I supposed to rebuild or what?

TheSussex: @hangy

hangy: > I did this and it's not showing the column visibility button anymore. Am I supposed to rebuild or what?

Yes, can you try rebuilding, please? Just the frontend (`make front_build` if you're using docker) should be enough.

TheSussex: This did not work @hangy

stephanegigandet (approved):
Thank you @TheSussex !

cgi/display_org_table.pl
stephanegigandet: @hangy @TheSussex I updated the JS files and included the ordering etc. buttons. They were also loaded with a hardcoded path to static.openfoodfacts.org, that's why they were not picked in a local dev environment.

TheSussex: Oh nice. Thanks for the update @stephanegigandet

PR #10684, opened 2024-08-12
feat: sync number of product public/pro in CRM

4nt0ineB:
### What

The daily script `(save_org_product_data_daily_off_pro.pl)` which updates orgs data now sync the number of public/prod products with CRM.


stephanegigandet (approved):
Thank you!

PR #10688, opened 2024-08-12
taxonomy: Add Swedish steaming and rolling

Freso:
### What

Processing by steaming and/or rolling wasn’t recognised in ingredients lists in Swedish. They should be after this PR.

- https://se.openfoodfacts.org/product/7340083480638/fiberhavregryn-garant#health


stephanegigandet (approved):
Thank you

PR #10691, opened 2024-08-13
taxonomy: updates to NOVA 4 markers

stephanegigandet:
Some adjustments based on details from the NOVA team:

> Vitamins and minerals (e.g., flour enriched with iron, orange juice enriched with calcium, breakfast cereals with minerals and vitamins): Vitamins and minerals are not considered markers of UPF.
> 
> Vegetable fiber (e.g., potato fiber, corn fiber): Soluble or insoluble fiber is considered a food substance with little or no culinary use and is primarily used in the manufacture of UPFs. Therefore, they are markers of UPFs.
> 
> Vegetable extracts (e.g., pepper extract, spice extract, vanilla extract, onion extract, yeast extract): Since yeast extracts (such as Vegemite and Marmite) typically contain colorants and flavorings, we classify them as UPFs. While 100% vanilla extract is not considered a UPF, commercially available vanilla extract often contains other substances that generally classify it as a UPF. 100% vanilla extract appears to be obtained through a maceration process, similar to other spice extracts. In contrast, onion extract involves an enzymatic extraction process from onion cells, although I couldn't find it listed in the Codex or as a commercially available culinary ingredient. If an extract functions as a cosmetic additive according to the Codex (such as annatto extract, grape skin extract, and paprika extract), it could be regarded as a marker of UPFs. In cases of doubt, it is better not to consider it as a UPF marker.
> 
> Glucose, fructose: Fructose is a marker of UPF. Glucose: see my comment below regarding corn syrup. It is the same.
> 
> Corn syrup, maple syrup, agave syrup: We typically state that honey extracted from combs and syrup from maple trees are not UPF markers. However, corn syrup and agave syrup are commercially available for use as culinary ingredients and are listed as 100% corn syrup or agave syrup. Although the processing is different, which could classify them as UPF markers, we believe it is more prudent to be conservative and not consider them as UPF markers.

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -30845,6 +30845,7 @@ pl: Cukier inwertowany
 pt: açúcar invertido
 ro: zahăr invertit, zahar invertit
 sv: invertsocker
+nova:en: 4
 wikidata:en: Q412082
 wikipedia:en: https://en.wikipedia.org/wiki/Inverted_sugar_syrup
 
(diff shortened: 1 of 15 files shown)


benbenben2 (approved):
LGTM

alexgarel:
Great :-)

PR #10692, opened 2024-08-13
fix: Automatically unselect images when source image is deleted

### LINKED ISSUE #9793: Automatically unselect images when source image is deleted

4nt0ineB:
### What

Automatically unselect images when source image is deleted

- Fixes #9793


lib/ProductOpener/Images.pm
```diff
@@ -1210,10 +1210,17 @@ sub process_image_move ($user_id, $code, $imgids, $move_to, $ownerid) {
 
 				delete $product_ref->{images}{$imgid};
 
+				if ($move_to eq 'trash') {
+					foreach my $related_img (keys %{$product_ref->{images}}) {
+						if ($product_ref->{images}{$related_img}{imgid} eq $imgid) {
+							_process_image_unselect($product_ref, $related_img);
```
stephanegigandet, line 1216: We want to unselect the image, but we also want to move the selected image files (similar to the source file with the code above) so that they are not visible online any longer as well.

4nt0ineB: ok, I made changes. I use a queue for the images to be moved and If an image has to be unselected because the source is trashed, we add it to the queue to trash it too.

stephanegigandet: Perfect, thank you!

stephanegigandet (approved):
Thank you!

PR #10695, opened 2024-08-13
taxonomy: fix specifying Norwegian as Bokmål (labels.txt)

Freso:
### What

Fix `no:` entries in `labels.txt` to be `nb:`.

Norwegian has two (primary) variants: Nynorsk (`nn`) and Bokmål (`nb`). It seems like Norwegian OFF uses `nb` for its language rather than the generic `no` language code—e.g., looking at https://no.openfoodfacts.org/label/n%C3%B8kkelhullet I get directed to answer robotoff questions related to `nb:nøkkelhullet`.

### Screenshot

[“Answer robotoff questions about label nb:nøkkelhullet for country Norway”](https://no.openfoodfacts.org/label/n%C3%B8kkelhullet)

- https://no.openfoodfacts.org/label/n%C3%B8kkelhullet

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -829,8 +829,8 @@ it: HACCP, Analisi dei rischi e controllo dei punti critici
 ja: HACCP
 ko: HACCP
 mk: HACCP
+nb: HACCP
 nl: HACCP
-no: HACCP
 pl: HACCP, Analiza zagrożeń i krytyczne punkty kontroli
 ps: HACCP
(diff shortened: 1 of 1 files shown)


alexgarel:
@stephanegigandet I need you point of view on this ! Won't it break ?

stephanegigandet:
We do use nb: for the no.openfoodfacts.org site.

There are some discussions about it in #norway on Slack: https://openfoodfacts.slack.com/archives/C18T4PS8K/p1646593318705489

stephanegigandet (approved):
Thanks

PR #10700, opened 2024-08-14
feat: geoip/ api use the remote address if ip not provided as param

### LINKED ISSUE #10699: Using the IP  from the request for the GeoIP endpoint
Thanks for the PR.

However, getting the public IP from Dart is impossible without an external service (ex : IPify).
Would it be possible to change the endpoint from:
`https://world.openfoodfacts.net/api/v3/geoip/85.214.132.117`
to
`https://world.openfoodfacts.net/api/v3/geoip/`

Basically, if the IP is missing, use the one from the request?

_Originally posted by @g123k in https://github.com/openfoodfacts/openfoodfacts-server/issues/10594#issuecomment-2285703502_

4nt0ineB:
### What

If the geoip endpoint is called without parameter `/api/v3/geoip` the remote address is used.

- Fixes #10699

### DIFF
--- a/lib/ProductOpener/Routing.pm
+++ b/lib/ProductOpener/Routing.pm
@@ -319,8 +319,8 @@ sub api_route($request_ref) {
 		param("tagid", $components[4]);
 		$request_ref->{tagid} = $components[4];
 	}
-	elsif ($api_action eq "geoip") {    # api/v3/geoip/[ip]
-		$request_ref->{geoip_ip} = $components[3] // remote_addr();
+	elsif ($api_action eq "geoip") {    # api/v3/geoip/
+		$request_ref->{geoip_ip} = remote_addr();
 	}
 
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Routing.pm
```diff
@@ -320,7 +320,7 @@ sub api_route($request_ref) {
 		$request_ref->{tagid} = $components[4];
 	}
 	elsif ($api_action eq "geoip") {    # api/v3/geoip/[ip]
-		$request_ref->{ip} = $components[3];
+		$request_ref->{geoip_ip} = $components[3] // remote_addr();
```
alexgarel, line 323: As seen, remove the geoip_ip parameter from url, we use remote_addr() only.

alexgarel (approved):
Great

PR #10701, opened 2024-08-14
feat: knowledge API to specify which panels should be returned

### LINKED ISSUE #9369: Extend knowledge API to specify which panels should be returned or excluded
Similar to the fields parameter, we could have a parameter to specify which knowledge panels to return (and/or possibly to do the reverse: exclude some panels).

### Example
- knowledge_panels_included=toto,tata,titi

### Related
- Moderator KP
- Experimental KPs

### Part of
- #8919

4nt0ineB:
### What

Added a `knowled_panels_included=x,y` and `knowled_panels_excluded=x,y` parameter to the API, to specify which panels should be in the response.

- Fixes #9369

### DIFF
--- a/docs/api/ref/api-v3.yml
+++ b/docs/api/ref/api-v3.yml
@@ -11,7 +11,7 @@ info:
   contact:
     name: Open Food Facts
     url: "https://slack-ssl-openfoodfacts.herokuapp.com/"
-    email: contact@openfoodfacts.org
+    email: reuse@openfoodfacts.org
   license:
     name: "License (MIT, Apache 2.0, etc)"
     url: "https://opendatacommons.org/licenses/odbl/summary/index.html"
--- a/docs/api/ref/api-v3.yml
+++ b/docs/api/ref/api-v3.yml
@@ -7,22 +7,22 @@ info:
     As a developer, the Open Food Facts API allows you to get information
     and contribute to the products database. You can create great apps to
     help people make better food choices and also provide data to enhance the database.
-  termsOfService: 'https://openweathermap.org/terms'
+  termsOfService: "https://openweathermap.org/terms"
   contact:
     name: Open Food Facts
-    url: 'https://slack-ssl-openfoodfacts.herokuapp.com/'
+    url: "https://slack-ssl-openfoodfacts.herokuapp.com/"
     email: contact@openfoodfacts.org
   license:
-    name: 'License (MIT, Apache 2.0, etc)'
-    url: 'https://opendatacommons.org/licenses/odbl/summary/index.html'
-  version: '3'
+    name: "License (MIT, Apache 2.0, etc)"
+    url: "https://opendatacommons.org/licenses/odbl/summary/index.html"
+  version: "3"
 servers:
-  - url: 'https://world.openfoodfacts.org'
+  - url: "https://world.openfoodfacts.org"
     description: prod
   - description: dev
-    url: 'https://world.openfoodfacts.net'
+    url: "https://world.openfoodfacts.net"
 paths:
-  '/api/v3/product/{barcode}':
+  "/api/v3/product/{barcode}":
     get:
       tags:
         - Read Requests
@@ -37,14 +37,14 @@ paths:
           explode: false
           schema:
             type: string
-            example: '3017620422003'
-        - $ref: '#/components/parameters/cc'
-        - $ref: '#/components/parameters/lc'
+            example: "3017620422003"
+        - $ref: "#/components/parameters/cc"
+        - $ref: "#/components/parameters/lc"
         - schema:
             type: string
           in: query
           name: tags_lc
-          description: '2 letter language code to request names of tags in a specific language.  For READ requets: if passed, all taxonomized tags of the response will include a lc_name property with the translation in the requested language, if available. Otherwise, the property value will contain the name in the original language, prefixed by the 2 language code and a colon.'
+          description: "2 letter language code to request names of tags in a specific language.  For READ requets: if passed, all taxonomized tags of the response will include a lc_name property with the translation in the requested language, if available. Otherwise, the property value will contain the name in the original language, prefixed by the 2 language code and a colon."
         - schema:
             type: string
           in: query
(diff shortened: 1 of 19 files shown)


docs/api/ref/api-v3.yml
```diff
@@ -7,22 +7,22 @@ info:
     As a developer, the Open Food Facts API allows you to get information
     and contribute to the products database. You can create great apps to
     help people make better food choices and also provide data to enhance the database.
-  termsOfService: 'https://openweathermap.org/terms'
+  termsOfService: "https://openweathermap.org/terms"
```
teolemon, line 10: what the heck ?

docs/api/ref/api-v3.yml
```diff
@@ -7,22 +7,22 @@ info:
...
   contact:
     name: Open Food Facts
-    url: 'https://slack-ssl-openfoodfacts.herokuapp.com/'
+    url: "https://slack-ssl-openfoodfacts.herokuapp.com/"
```
teolemon, line 13: https://slack.openfoodfacts.org/

docs/api/ref/api-v3.yml
teolemon: https://world.openfoodfacts.org/terms-of-use"

docs/api/ref/api.yml
```diff
@@ -5,28 +5,28 @@ info:
     As a developer, the Open Food Facts API allows you to get information
     and contribute to the products database. You can create great apps to
     help people make better food choices and also provide data to enhance the database.
-  termsOfService: 'https://world.openfoodfacts.org/terms-of-use'
+  termsOfService: "https://world.openfoodfacts.org/terms-of-use"
   contact:
     name: Open Food Facts
-    url: 'https://slack.openfoodfacts.org/'
+    url: "https://slack.openfoodfacts.org/"
     email: contact@openfoodfacts.org
```
teolemon, line 12:
```suggestion
    email: reuse@openfoodfacts.org
```

docs/api/ref/api-v3.yml
```diff
@@ -7,22 +7,22 @@ info:
...
     email: contact@openfoodfacts.org
```
teolemon, line 14:
```suggestion
    email: reuse@openfoodfacts.org
```

docs/api/ref/api-v3.yml
```diff
@@ -7,22 +7,22 @@ info:
...
+    name: "License (MIT, Apache 2.0, etc)"
+    url: "https://opendatacommons.org/licenses/odbl/summary/index.html"
+  version: "3"
 servers:
-  - url: 'https://world.openfoodfacts.org'
+  - url: "https://world.openfoodfacts.org"
     description: prod
   - description: dev
-    url: 'https://world.openfoodfacts.net'
+    url: "https://world.openfoodfacts.net"
```
alexgarel, line 23: Thanks for those fixes !

docs/api/ref/api-v3.yml
```diff
@@ -37,14 +37,14 @@ paths:
...
-        - $ref: '#/components/parameters/lc'
+            example: "3017620422003"
+        - $ref: "#/components/parameters/cc"
+        - $ref: "#/components/parameters/lc"
         - schema:
             type: string
           in: query
           name: tags_lc
-          description: '2 letter language code to request names of tags in a specific language.  For READ requets: if passed, all taxonomized tags of the response will include a lc_name property with the translation in the requested language, if available. Otherwise, the property value will contain the name in the original language, prefixed by the 2 language code and a colon.'
+          description: "2 letter language code to request names of tags in a specific language.  For READ requets: if passed, all taxonomized tags of the response will include a lc_name property with the translation in the requested language, if available. Otherwise, the property value will contain the name in the original language, prefixed by the 2 language code and a colon."
```
alexgarel, line 47:
```suggestion
          description: "2 letter language code to request names of tags in a specific language.  For READ requests: if passed, all taxonomized tags of the response will include a `lc_name` property with the translation in the requested language, if available. Otherwise, the property value will contain the name in the original language, prefixed by the 2 language code and a colon."
```

docs/api/ref/api-v3.yml
```diff
@@ -58,8 +58,25 @@ paths:
...
+            When knowledge_panels are requested, you can specify which panels should be in the response. All the others will be excluded.
+        - schema:
+            type: string
+            example: "health_card, environment_card"
+          required: false
+          in: query
+          name: knowledge_panels_excluded
+          description: |-
+            When knowledge_panels are requested, you can specify which panels to exclude from the response. All the others will be included.
+            If a panel is both excluded and included (with the knowledge_panels_excluded parameter), it will be excluded.
```
alexgarel, line 77: Kudos, this is exactly the spec I would have proposed :slightly_smiling_face:

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -106,6 +106,19 @@ sub initialize_knowledge_panels_options ($knowledge_panels_options_ref, $request
 	}
 	$knowledge_panels_options_ref->{knowledge_panels_client} = $knowledge_panels_client;
 
+	my $included_panels = single_param('knowledge_panels_included') || '';
+	my %included_panels = map {$_ => 1} split(/,/, $included_panels);
+	my $excluded_panels = single_param('knowledge_panels_excluded') || '';
+	# excluded overrides included
+	$included_panels{$_} = 0 for split(/,/, $excluded_panels);
```
alexgarel, line 113: Nit: AFAIK in Perl, True is 1, False is undef, so
```suggestion
	$included_panels{$_} = undef for split(/,/, $excluded_panels);
```
might be more perlish

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -106,6 +106,19 @@ sub initialize_knowledge_panels_options ($knowledge_panels_options_ref, $request
...
+	$knowledge_panels_options_ref->{knowledge_panels_includes} = sub {
+		my $panel_id = shift;
+		return (
+				   (exists $included_panels{$panel_id}) and ($included_panels{$panel_id} eq 1)
+				or (not(exists $included_panels{$panel_id}) and not $included_panels)
```
alexgarel, line 118: The condition is not that easy to read.
If we use undef above why not:
```suggestion
					not $included_panels or (defined $included_panels{$panel_id})
```

tests/integration/api_v3_product_read.t
```diff
@@ -173,6 +173,20 @@ my $tests_ref = [
...
+		query_string => '?fields=knowledge_panels&knowledge_panels_included=health_card,environment_card',
+		expected_status_code => 200,
+	},
+	{
+		test_case => 'get-fields-knowledge-panels-knowledge-panels_excluded-environment_card',
+		method => 'GET',
+		path => '/api/v3/product/4260392550101',
+		query_string => '?fields=knowledge_panels&knowledge_panels_excluded=environment_card',
+		expected_status_code => 200,
+	},
```
alexgarel, line 189: would be great to also add a include + exclude example

alexgarel (approved):
Great work, some changes proposed.

lib/ProductOpener/KnowledgePanels.pm
4nt0ineB: It didn't work, in facts its the 'exists' that is really important. Finally I did this, but if you see a better way ... 
https://github.com/openfoodfacts/openfoodfacts-server/blob/3ec9ae8bb5cdad82277b581f21becee500016b5a/lib/ProductOpener/KnowledgePanels.pm#L116-L117

PR #10703, opened 2024-08-14
Fix the OpenAPI definitions and create a GitHub Actions workflow for SDKs

teolemon:
---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/openfoodfacts/openfoodfacts-server?shareId=55339fb7-743a-4a75-bec6-e34eb208f29c).

### DIFF
--- a/docs/api/ref/api-v3.yml
+++ b/docs/api/ref/api-v3.yml
@@ -1,4 +1,4 @@
-openapi: 3.1.0
+openapi: 3.0.0
 x-stoplight:
   id: 0x3135wk4xq0t
 info:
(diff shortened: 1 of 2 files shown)


docs/api/ref/api-v3.yml
```diff
@@ -1,4 +1,4 @@
-openapi: 3.1.0
+openapi: 3.0.0
```
stephanegigandet, line 1: we are using 3.1.0

stephanegigandet:
No idea in which state this is in, was this just an experiment?
We probably do not want autogenerated SDKs unless we are sure they are correct and can be used by others, otherwise it will be very confusing for reusers.

Closing for now, please reopen if this is still current / wanted.

PR #10708, opened 2024-08-17
fix: remove unused jquery-ui safe-active-element module

hangy:
Remove [no more existing](https://jqueryui.com/changelog/1.14.0/) jQuery UI file from Gulp build.


stephanegigandet (approved):
Thanks

jayaddison:
Unfortunately I think we may want to restore this dependency - it would be nice to remove it, but following the rollback to `jquery-ui` 1.13.3, it seems that it is used for some JavaScript functionality (ref #10861) in the webapp.

I've opened a pull request with the revert in #10924.

PR #10711, opened 2024-08-18
feat: update script to correct wrong languages for tags

benbenben2:
update script to correct wrong languages for tags

### What
Rework of the previous script to tackle tags in wrong languages for different tags type: categories, countries, labels, origins, traces. Not for allergens as updates seems not be working for those, and ingredients as it is not simple tags like others fields

Update due to result per page on api call. Make more file during the process to be able to restart the script if it fails.

Previous script:
https://github.com/openfoodfacts/openfoodfacts-server/pull/9581


benbenben2:
improvements
- sort/group by barcode, and update per barcode. That way, if barcode has 3-4 wrongs tags it will make only one post request. 
- ignore when xx:<tag> is among the results (search unknown tag in the taxonomy). 
- duplicated rows are removed in the output files after extracting all products for tags we want to update
- create a new file 'products to update' for tags that are not unknown anymore but because some products have not been updated for long time they are still considered as having this unknown tag
- reworked first function api call: https://world.openfoodfacts.org/countries?status=unknown 
- handle case when tag found in the taxonomy is in the same language as the unknown tag, it means that the tag is known but simply need to be updated

example for countries
update_tags_per_languages_countries_exist -> 154 lines
```
current tag;new tag;products
en:francia;an:francia;162
en:deutschland;de:deutschland;135
en:belgique;fr:belgique;118
en:schweiz;da:schweiz;106
en:frankreich;de:frankreich;106
fr:francia;an:francia;68
en:nederland;af:nederland;36
en:espagne;fr:espagne;34
en:algerie;nb:algerie;31
en:belgica;co:belgica;24
...
```

update_tags_per_languages_countries_new -> 462 lines
```
current tag;products
en:turkiye;192
en:en;145
fr:angleterre;77
en:francia-espana;66
en:worldwide;37
en:england;30
ru:nyl;29
en:angleterre;21
en:indonesie;21
en:emirats-arabes-unis;21
...
```

update_tags_per_languages_countries_to_update -> 2 lines
```
current tag;products
en:republic-of-macedonia;96
en:xk;2
```

benbenben2:
more improvements 

when tags occurs more than 1 time in the same field (https://world.openfoodfacts.net/product/7610095231406/protein-paprika-snack-vaya) it will remove the duplicates before to update the field

For origins, include countries taxonomy.

PR #10720, opened 2024-08-21
ci: add a NO_DEPLOY flag to temporarily block deployments

alexgarel:
This enable long running operations on staging (eg. testing a migration).


alexgarel:
Note that the deployment fails on purpose.

teolemon:
approved @alexgarel

PR #10723, opened 2024-08-22
fix: remove robotoff webhook trigger

raphael0202:
Robotoff now uses Redis Stream to get information about product updates.


stephanegigandet (approved):
Thanks

PR #10734, opened 2024-08-27
feat: 99 French "indice de durabilité" Label images

teolemon:
### What
- feat: 99 French "indice de durabilité" Label images
- not the most urgent - they made one SVG for every decimal of this shelved score (for now)
- being able to display it experimentally is aspirational for the French gov and partners (HOP)

### Part of
- #5555


stephanegigandet (changes requested):
Thank you. Could you rename the files to add a dash (so that it corresponds to the label we will have), and add the size?
e.g.
indice-de-durabilite-00.svg -> indice-de-durabilite-0-0.108x90.svg

stephanegigandet (approved):
The height in filenames must be 90px, I renamed the files

PR #10738, opened 2024-08-27
feat: add rate-limiting on facet pages

raphael0202:
Rate-limit of on facet pages, such as:
- /categories
- /label/organic
- /ingredient/salt/category/breads

The rate limit depends on the user:
- unregistered users: 5 req/min
- registered user: 10 req/min
- crawling bots: 10 req/min. Rate-limiting is higher for crawling bots than for unregistered users, as otherwise web indexer bots will slow-down indexing for the full website

This mechanism also provides an easy way to block completely some resource-intensive pages, such as facets, if the server is unresponsive due to heavy load (by setting the rate-limit value for the corresponding bucket to 0).


lib/ProductOpener/Config_off.pm
```diff
@@ -1634,6 +1634,12 @@ $options{sample_product_code} = "093270067481501";    # A good product for you -
 
 # Number of requests per minutes for the search API
 $options{rate_limit_search} = 10;
+# Number of requests per minutes for all facets for anonymous users
+$options{rate_limit_facet_unregistered} = 2;
```
stephanegigandet, line 1638: 2 requests seem very small. e.g. you do a first one to explore by brands, you click on a brand, that's already 2 facets...

lib/ProductOpener/Routing.pm
```diff
@@ -545,6 +551,7 @@ sub facets_route($request_ref) {
 	}
 
 	$request_ref->{canon_rel_url} .= $canon_rel_url_suffix;
+	$request_ref->{rate_limiter_bucket} = "facet";
 
 	if (defined $request_ref->{groupby_tagtype}) {
 		set_request_stats_value($request_ref->{stats}, "route", "facets_tags");
```
stephanegigandet, line 557: Maybe set the rate_limiter_bucket to facets_tags. Those are very different queries: currently the list of products "facets" goes to mongodb, and the list of tags "facets_tags" goes to off-query

raphael0202: Right, I will add a distinct bucket for `facet_tags`.

lib/ProductOpener/Config_off.pm
raphael0202: I increased it to 5.

stephanegigandet (approved):
Thank you!

PR #10743, opened 2024-08-28
fix: remove non-JS list of products

stephanegigandet:
As discussed on the product opener meeting on Monday, this removes the HTML list of products that that is displayed when browsers have turned off Javascript. This is done for performance reasons, to avoid having browsers download the data twice. It also makes the responses smaller and easier to cache.


lib/ProductOpener/Display.pm
```diff
@@ -5645,7 +5646,8 @@ sub search_and_display_products ($request_ref, $query_ref, $sort_by, $limit, $pa
 	my $products_json = '[]';
 
 	if (defined $request_ref->{structured_response}{products}) {
-		$products_json = $json->encode($request_ref->{structured_response}{products});
+		# We indent the JSON in the generated HTML so that we can easily see diffs in integration tests outputs
+		$products_json = $json_indent->encode($request_ref->{structured_response}{products});
```
alexgarel, line 5650: :+1:

templates/web/common/includes/list_of_products.tt.html
```diff
@@ -113,20 +113,9 @@
...
-        [% IF product.product_name == '' %]
-          <a href="[% product.url %]" title="[% product.code %]">
-          [% ELSE %]
-          <a href="[% product.url %]" title="[% product.product_name %]">
-          [% END %]
-            <div>[% product.img %]</div>
-            <span>[% product.product_name %]</span>
-          </a>
-        </li>
-        [% END %]
```
alexgarel, line 127: Maybe add a message (to make it clear for people not having JS activated in some way):
```html
     [% ELSE %]
       <div>Product list loading from javascript…</div>
     [% END %]
```

alexgarel (approved):
Great.

I made a suggestion.

PR #10750, opened 2024-08-30
fix: import_prod_data not working

### LINKED ISSUE #10749: make import_prod_data is not working
### What
Get the following error:

stat /home/john/openfoodfacts-server/deps/openfoodfacts-shared-services/docker/dev.yml: no such file or directory

1. Make sure that the DEPS_DIR content is empty so that a fresh clone of off-shared-services is obtained
2. Run make import_prod_data
3. See error

### Expected behavior
- Should not reference files outside of the root folder of the dependency

john-gom:

### What

Can't use -e flag when calling make targets in dependencies

Fixes #10749

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -232,7 +232,7 @@ refresh_mongodb:run_deps
 
 # this command is used to import data on the mongodb used on staging environment
 import_prod_data: run_deps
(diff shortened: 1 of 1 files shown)


teleboas (approved):
LGTM

PR #10754, opened 2024-08-30
fix: block more AI bots

raphael0202:
Following Cloudflare analysis on
https://blog.cloudflare.com/declaring-your-aindependence-block-ai-bots-scrapers-and-crawlers-with-a-single-click/


teleboas (approved):
Yes! Yes! Yes!

PR #10766, opened 2024-09-03
chore: have two apache servers

alexgarel:
See https://github.com/openfoodfacts/openfoodfacts-infrastructure/pull/399

Part of:
- https://github.com/openfoodfacts/openfoodfacts-infrastructure/issues/396


conf/apache-2.4/ports.conf
```diff
@@ -2,6 +2,7 @@
 # have to change the VirtualHost statement in
 # /etc/apache2/sites-enabled/000-default.conf
 
-Listen 8002
+# APACHE2_PORT as to be passed using -D directive in systemd/apache2.service.d/override.conf
```
stephanegigandet, line 5:
```suggestion
# APACHE2_PORT has to be passed using -D directive in systemd/apache2.service.d/override.conf
```

conf/nginx/sites-available/off
```diff
@@ -13,6 +13,21 @@ server {
 	}
 }
 
+# map to decide if we go to the main (8004) or secondary service (8805)
```
stephanegigandet, line 16:
```suggestion
# map to decide if we go to the main (8004) or secondary service (8005)
```

conf/nginx/sites-available/off
```diff
@@ -13,6 +13,21 @@ server {
 	}
 }
 
+# map to decide if we go to the main (8004) or secondary service (8005)
+map $request_method$uri $apache_port {
+	default 8005;
+
+	# home pages
+	"~*GET/$" 8004;
```
stephanegigandet, line 21: We should handle also the HEAD and OPTIONS methods (used by browsers for CORS)

conf/nginx/sites-available/off
```diff
@@ -13,6 +13,21 @@ server {
...
+	# product read / write
+	"~*GET/(mountaj|məhsul|продукт|gynnyrch|produkt|product|product|product|produkto|producto|toode|produkto|produit|produto|termék|produk|製品|afaris|өнім|생성물|berhem|उत्पादन|produk|produkt|उत्पादन|product|product|product|produkt|produkt|produit|produto|produto|produto|продукт|product|proizvod|produkto|ürün|продукт|产品|產品|產品)/" 8004;
+	"~*GET/cgi/product.pl" 8004;
+	"~*POST/cgi/product.pl" 8004;
+	# product API read / write
+	"~*GET/api/v./product/" 8004;
+	"~*POST/api/v./product/" 8004;
```
stephanegigandet, line 28: For API v3, we use the PATCH method too

conf/nginx/sites-available/off
```diff
@@ -13,6 +13,21 @@ server {
 	}
 }
 
+# map to decide if we go to the main (8004) or secondary service (8005)
```
stephanegigandet, line 16: "secondary" sounds like it's not a priority. I almost expected that we would send slow requests that we care a bit less about to the "secondary" Apache.

I think it would be clearer to use a name that gives the intent, like "fasttrack" or "fastlane".

env/env.obf.main
```diff
@@ -0,0 +1,3 @@
+# This is used by ports.conf in apache2
+APACHE_ARGUMENTS=-D APACHE2_PORT=8002
```
stephanegigandet, line 2: While we are changing those files, could we use the same port numbers for all flavors? We don't run them in the same server anymore, so we could just use 8001 (and 8002 if there's a fast track Apache) for all flavors.

stephanegigandet (approved):
Looks good to me, thank you.

I added a few comments, I think we should at least add the HEAD/OPTIONS/PATCH methods (or maybe just remove the method altogether, I can't think of a case where we would want to send a request to different Apache servers based on the method)

conf/nginx/sites-available/off
```diff
@@ -13,6 +13,21 @@ server {
 	}
 }
 
+# map to decide if we go to the priority (8002) or standard service (8801)
```
alexgarel, line 16:
```suggestion
# map to decide if we go to the priority (8002) or standard service (8001)
```

PR #10783, opened 2024-09-10
Fix: YAML syntax in schema files.

jforget:
### What

Some schema files have incorrect YAML syntax. When a key-value pair is split on two lines, the colon separating the value from the key should be followed by a pipe. While most YAML files contain the pipe chars when required, some of them miss one or two pipe chars.


stephanegigandet (approved):
Thank you!

PR #10786, opened 2024-09-12
fix: incorrect self-reference in ingredient.yaml

that1:
"#/" refers to a property with empty name. "#" refers to the root.


stephanegigandet (approved):
Thanks

PR #10791, opened 2024-09-13
fix: Determine obsolete and deleted from the product revision

### LINKED ISSUE #10789: Script to load historic events into off-query doesn't recognise deleted products
### What
- product_revision_to_historical_events.pl is looking for hard-coded text in the comment but this isn't always the same

1. Delete a product and do not use the word "Deleting product" in the comment
2. Run the script
3. See that no deleted event is created

### Expected behavior
- Should be using the deleted property from the product

john-gom:

### What

Use the product revision to determine delete and obsolete status

- Fixes #10789


stephanegigandet (approved):
Looks good, thank you

PR #10793, opened 2024-09-13
taxonomy: Identifiers from Ecobalyse have been added for few ingredients

very-smartin:
### What

5 identifiers have been added for few ingredients : 
ecobalyse:en:
ecobalyse_labels_en_organic:en:
ecobalyse_origins_en_france:en:
ecobalyse_origins_en_european_union:en:
ecobalyse_labels_en_organic_origins_en_france:en

- Fixes #[ISSUE NUMBER]


taxonomies/food/ingredients.txt
```diff
@@ -27681,6 +27681,11 @@ carbon_footprint_fr_foodges_ingredient:fr: Farine de blé
 carbon_footprint_fr_foodges_value:fr: 1.2
 ciqual_proxy_food_code:en: 9410
 ciqual_proxy_food_name:fr: Farine de blé tendre ou froment T110
+ecobalyse:en: flour
+ecobalyse_labels_en_organic:en: flour-organic
+ecobalyse_origins_en_france:en: flour
```
stephanegigandet, line 27686: If the value is the same as the default in ecobalyse:en: , I think it's best not to add properties like ecobalyse_origins_en_france:en: . We can add only the ones that are different from the default.

taxonomies/food/ingredients.txt
```diff
@@ -57744,6 +57764,11 @@ carbon_footprint_fr_foodges_value:fr: 0.5
 ciqual_food_code:en: 20047
 ciqual_food_name:en: Tomato, raw
 ciqual_food_name:fr: Tomate, crue
+ecobalyse:en: tomato-organic-eu
```
stephanegigandet, line 57767: The default should be the "Tomate UE Conv" I think.

very-smartin: I do agree with you ! However 'tomato-eu' is not yet available :/

teolemon:
@very-smartin why isn't ecobalyse using agribalyse id instead ?

very-smartin:
> @very-smartin why isn't ecobalyse using agribalyse id instead ?

As Ecobalyse wants to let its users create products of their own, they've chosen to have 
-few ingredients (ex: apple) with few variations (fr, eu, non-eu, bio and conventionnal) AND processes (cooking, mixing, packaging...)
-rather than a wide variety of ingredients (as Agribalyse currently use)

So          [Agribalyse] : 'mandarin, peeled, raw', 'mandarin, peeled, cooked', 'mandarin, raw', ...
Become [Ecobalyse]  : 'mandarin-fr', 'mandarin-fr-bio', 'mandarine-non-eu'...

PR #10794, opened 2024-09-13
fix: new mongodb exports for obsolete products collection

stephanegigandet:
add exports of obsolete products (requested by MyLabel app)


alexgarel (approved):
Thanks !

PR #10801, opened 2024-09-16
refactor: Update Makefile to use dynamic shell path

hangy:
Update the `Makefile` to use the dynamic shell path instead of the hardcoded `/bin/bash`. This ensures that the correct shell is used across different environments. Additionally, the environment variable `SHELL` is updated to use the `which` command to locate the path of the bash shell.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ endif
 
 
 # use bash everywhere !
-SHELL := /bin/bash
+SHELL := $(shell which bash)
 # some vars
 ENV_FILE ?= .env
 NAME = "ProductOpener"


stephanegigandet (approved):
Thanks.
Was it causing issues in some environments?

hangy:
> Thanks. Was it causing issues in some environments?

Yes, `make` wasn't able to find `bash` on my laptop, which I recently migrated from Arch to [NixOS](https://nixos.org/). By using `which`, it should be able to find the most relevant shell on the `$PATH`. 😊

PR #10816, opened 2024-09-20
test: Run API test with fresh UA each time, and randomize API test execution

hangy:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/lib/ProductOpener/APITest.pm
+++ b/lib/ProductOpener/APITest.pm
@@ -78,6 +78,7 @@ use Clone qw/clone/;
 use File::Tail;
 use Test::Fake::HTTPD;
 use Minion;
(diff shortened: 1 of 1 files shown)


hangy:
This is related to what I talked about yesterday @john-gom. Interestingly, some integration tests are already failing with this change, so obviously can't merge as-is

hangy:
Looks like this is fully intended. There are several tests that depend on the order of execution of the array to test that ie. a newly created product can be retrieved via the API directly. So I'll close this as "works as planned" 😄  For the Keycloak test, I'll try to a) separate auth tests from the other API tests, and b) try to replicate what happens when a wrong password is given when a user was previously signed in with correct data.

PR #10830, opened 2024-09-26
docs: :memo: How to extract a batch of images from AWS

jeremyarancio:
After some discussions during the OFF Days, it seems it was complex for users to extract a batch of images from the OFF AWS S3. We provide the code and explanations in the documentation.


stephanegigandet (approved):
Thanks!

PR #10833, opened 2024-09-27
taxonomy: added ecobalyse properties on a few ingredients

stephanegigandet:
Based on https://github.com/openfoodfacts/openfoodfacts-server/pull/10831/files from @very-smartin


taxonomies/food/ingredients.txt
```diff
@@ -13819,13 +13819,13 @@ it: prosciutto cotto
 nl: gekookte ham
 pt: fiambre cozinhado
 sv: kokt skinka
-
```
stephanegigandet, line 13822: I moved the properties for "cooked ham" to the en:cooked-ham and removed the properties from its children, as they will be inherited.

very-smartin (approved):

PR #10852, opened 2024-10-01
fix docs Internal URL broken fix

kant:
### What
Internal URL fixed

- No related with previous issues


stephanegigandet (approved):
Thanks

PR #10853, opened 2024-10-01
fix: docs Fixed proper denomination of product in query

kant:
### What
Name of product on query: fixed

- No related with previous fixes or issues


stephanegigandet (approved):
Thanks

PR #10863, opened 2024-10-03
feat: upload DB dump to AWS S3

raphael0202:
Upload the following dump files to AWS S3, just after being created in gen_feed_daily script:
- en.openfoodfacts.org.products.csv
- en.openfoodfacts.org.products.csv.gz
- fr.openfoodfacts.org.products.csv
- fr.openfoodfacts.org.products.csv.gz
- en.openfoodfacts.org.products.rdf
- fr.openfoodfacts.org.products.rdf
- openfoodfacts-products.jsonl.gz
- openfoodfacts-mongodbdump.gz
- openfoodfacts_recent_changes.jsonl.gz

Also add redirects (HTTP 302) to AWS S3 in the off nginx configuration so that we save I/O.

We use minio client (mc) for synchronization. We expect `/home/off/.mc/config.json` to contain AWS credentials.

### DIFF
--- a/scripts/gen_feeds_daily_off.sh
+++ b/scripts/gen_feeds_daily_off.sh
@@ -26,6 +26,16 @@ cd /srv/off/scripts
 
 ./mongodb_dump.sh /srv/off/html openfoodfacts 10.1.0.102 off
 
+# Copy CSV and RDF files to AWS S3 using MinIO client
+mc cp \
+    en.openfoodfacts.org.products.csv \
+    en.openfoodfacts.org.products.csv.gz \
+    en.openfoodfacts.org.products.rdf \
+    fr.openfoodfacts.org.products.csv \
+    fr.openfoodfacts.org.products.csv.gz \
+    fr.openfoodfacts.org.products.rdf \
+    s3/openfoodfacts-ds
+
 # Small products data and images export for Docker dev environments
 # for about 1/10000th of the products contained in production.
 ./export_products_data_and_images.pl --sample-mod 10000,0 \
--- a/scripts/gen_feeds_daily_off.sh
+++ b/scripts/gen_feeds_daily_off.sh
@@ -21,11 +21,6 @@ for export in en.openfoodfacts.org.products.csv fr.openfoodfacts.org.products.cs
    mv -f new.$export.gz $export.gz
 done
 
-# Generate the MongoDB dumps and jsonl export
-cd /srv/off/scripts
-
-./mongodb_dump.sh /srv/off/html openfoodfacts 10.1.0.102 off
-
 # Copy CSV and RDF files to AWS S3 using MinIO client
 mc cp \
     en.openfoodfacts.org.products.csv \
@@ -35,6 +30,11 @@ mc cp \
     fr.openfoodfacts.org.products.csv.gz \
     fr.openfoodfacts.org.products.rdf \
     s3/openfoodfacts-ds
+
+# Generate the MongoDB dumps and jsonl export
+cd /srv/off/scripts
+
+./mongodb_dump.sh /srv/off/html openfoodfacts 10.1.0.102 off
 
 # Small products data and images export for Docker dev environments
 # for about 1/10000th of the products contained in production.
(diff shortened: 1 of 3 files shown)


scripts/gen_feeds_daily_off.sh
```diff
@@ -21,6 +21,16 @@ for export in en.openfoodfacts.org.products.csv fr.openfoodfacts.org.products.cs
    mv -f new.$export.gz $export.gz
 done
 
+# Copy CSV and RDF files to AWS S3 using MinIO client
+mc cp \
```
stephanegigandet, line 25: Some of those files are quite big and could take a long time  to upload. What happens in the mean time? Is there a temporary file on s3, and the existing file is replaced once the full file has been received, or someone could download a file that is only partially uploaded?

CharlesNepote: Also, why not doing it in the background to let the script continue, adding a `&` at the end of the command?

_If a command is terminated by the control operator &, the shell executes the command in the background in a subshell._

scripts/mongodb_dump.sh
```diff
@@ -60,4 +60,11 @@ popd > /dev/null # data/delta
 mongoexport --collection recent_changes --host $HOST --db $DB --fields=_id,comment,code,userid,rev,countries_tags,t,diffs | gzip -9 > "new.${PREFIX}_recent_changes.jsonl.gz" && \
 mv new.${PREFIX}_recent_changes.jsonl.gz ${PREFIX}_recent_changes.jsonl.gz
 
+# Copy files to AWS S3 using MinIO client
+mc cp \
+    ${PREFIX}-products.jsonl.gz \
+    ${PREFIX}_recent_changes.jsonl.gz \
+    ${PREFIX}-mongodbdump.gz \
+    s3/openfoodfacts-ds
```
CharlesNepote, line 68: Same comment: you can add `&` if you want to do it in the background.

scripts/gen_feeds_daily_off.sh
alexgarel: @CharlesNepote the problem is that if you use an '&' it's harder to know there has been an error. If we do this we have to join all children at the end and ensure they ended up correctly (or fail).

raphael0202: CSV+RDF files took 5 minutes to upload (Total: 30.18 GiB, Transferred: 30.18 GiB, Speed: 107.80 MiB/s)
Other dumps took 3 minutes to upload (Total: 16.99 GiB, Transferred: 16.99 GiB, Speed: 106.60 MiB/s)

From what I can see in the doc (and what I've observed during upload), PUT operations on an existing key are atomic: https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html#ConsistencyModel

So either the file is fully uploaded and it's available, or the old version is served.

raphael0202:
As the subject of the time taken by each command during the daily dump comes up, here is a quick analysis I did using the latest logs (it's not 100% accurate, as some tasks such as 'product_countries.js' don't log anything, but it's useful to have to big numbers:

```
2.15AM: start of the job
1. removing empty products (<1min)
2.17AM->3.29AM completed products
product stats per country (<1min)
3.30AM->6.01AM: CSV and RDF files for english (2h40)
6.01AM->9.11AM: CSV and RDF files for french (3h10)
9.11AM->9.16AM: transfer to AWS S3 of CSV+RDF
9.16AM->12.11AM: MongoDB, JSONL export (~3h)
12.11AM->12.14AM: transfoer to AWS S3 of MongoDB+JSONL dumps
12.16AM->12.19AM export_products_data_and_images.pl (3 min)
12.19AM->12.21AM: diff files (actually the command crashed: "Oct 04 12:21:21 off gen_feeds_daily_off.sh[1013528]: [Fri Oct  4 12:21:21 2024] export_csv_file.pl: MongoDB::DatabaseError: Error in $cursor stage :: caused by :: operation exceeded time limit"
```

Almost 6 hours for CSV and RDF seems like a lot to me.

raphael0202:
Some processes seem to crash in the script by the way, probably due to python command missing (replaced by python3):

`Oct 04 12:19:41 off gen_feeds_daily_off.sh[1013487]: /srv/off/scripts/gen_feeds_daily_off.sh: ./generate_dump_for_offline_apps_off.py: /usr/bin/python: bad interpreter: No such file or directory
`

PR #10877, opened 2024-10-11
fix: downgrade jquery-ui

Valimp:
### What
Downgrade jquery-ui to 1.13.3 instead of 1.14.0


hangy (commented):
This downgrades too many unrelated dependencies, IMHO

raphael0202:
@hangy When building js files locally after downgrading jquery-ui version, @Valimp encountered issues with typescript compilation.
We wanted to make sure that the problem was local only.

raphael0202 (approved):
Thank you!

PR #10904, opened 2024-10-15
feat: ingredients knowledge panels

stephanegigandet:
This PR adds knowledge panels for ingredients, to list each ingredient below the ingredient list.
Each ingredient panel can be opened to get a description from: https://github.com/openfoodfacts/openfoodfacts-web/tree/main/knowledge_panels

Known issue: most ingredients do not currently have a description, they contain an empty panel. 

image


lib/ProductOpener/KnowledgePanelsIngredients.pm
```diff
@@ -0,0 +1,154 @@
...
+sub create_ingredients_list_panel ($product_ref, $target_lc, $target_cc, $options_ref) {
+
+	$log->debug("create ingredients list panel", {code => $product_ref->{code}}) if $log->is_debug();
+
+	# Create a panel only if the product has ingredients
+
+	if ((defined $product_ref->{ingredients_tags}) and (scalar @{$product_ref->{ingredients_tags}} > 0)) {
+
+		my $ingredient_i = 0;    # sequence number for ingredients
+		my @ingredients_panels_ids
```
alexgarel, line 80:
```suggestion
		# creates each individual panels for each ingredient
		my @ingredients_panels_ids
```

lib/ProductOpener/KnowledgePanelsIngredients.pm
```diff
@@ -0,0 +1,154 @@
...
+			= create_ingredients_panels_recursive($product_ref, \$ingredient_i, 0, $product_ref->{ingredients},
+			$target_lc, $target_cc, $options_ref);
+		my $ingredients_list_panel_data_ref = {ingredients_panels_ids => \@ingredients_panels_ids};
+
+		create_panel_from_json_template(
```
alexgarel, line 85:
```suggestion
		# create the panel that reference ingredients panels
		create_panel_from_json_template(
```

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
```diff
@@ -0,0 +1,37 @@
+{
+    "level": "info",
+    "topics": [
+        "health"
+    ],
+    "size": "small",
+    [% IF panel.ingredient_description OR panel.wikipedia_abstract %]
+    "expanded": false,
+    [% ELSE %]
+    "expanded": false,
```
alexgarel, line 10: This is strange !

Shouldn't the second expanded be true ?

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
```diff
@@ -0,0 +1,37 @@
...
+    [% END %]
+    "title_element": {
+        // Note: the app displays line feeds as line feeds...
```
alexgarel, line 13:
```suggestion
        // Note: the app displays line feeds as line feeds...
        // so we have to keep whole title expression on one line !
```

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
```diff
@@ -0,0 +1,37 @@
...
+        "title": `        [% dash = "-" %]        [% dash.repeat(panel.level) %]        [% display_taxonomy_tag_name("ingredients", panel.ingredient_id) %]        [% IF panel.ingredient.percent.defined %]            ([% round(panel.ingredient.percent) %]%)        [% ELSIF panel.ingredient.percent_estimate.defined %]            ([% round(panel.ingredient.percent_estimate) %]% [% lang("estimate") %])        [% END %]        `,
```
alexgarel, line 14: Can't we eventually deport this computation in the ingredient panel creation ?

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
```diff
@@ -0,0 +1,37 @@
...
+    },        
+    "elements": [
+        [% IF panel.ingredient_description -%]
+        {
+            "element_type": "text",
+            "text_element": {
+                "html": `[% panel.ingredient_description %]`
```
alexgarel, line 21:
```suggestion
                "html": `[% edq(panel.ingredient_description) %]`
```

templates/api/knowledge-panels/health/ingredients/ingredients.tt.json
```diff
@@ -55,6 +55,15 @@
                 `,
             },
         },
-        [% END %]        
+        [% END %]       
+        [% IF 1 %]
```
alexgarel, line 59: it should be if we have feature ingredient right ?

alexgarel (approved):
LGTM but some comments and maybe a mistake ?

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
stephanegigandet: The backticks already take care of escaping quotes, so edq() is not needed:
see https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/KnowledgePanels.pm#L298

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
stephanegigandet: I just added a double backtick function so that we can have things on multiple lines that are then converted to one line without linebreaks.

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
stephanegigandet: It's temporary, in fact what we really want is to make the first one (no description) not expandable. But currently the non expandable panels are displayed in a very different way (at least on the web, I haven't checked mobile). So we will need a better way.

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
stephanegigandet: Well the idea of using Template::Toolkit was that people could change the layout / design decisions from the panels, without having to change the Perl code. So that's why I try to put things that are more design things in the template. Here I tried to show the sub-ingredients, and I'm just pushing dashes in front, which is not very clear (but it's something..). There may be other better ways to do that.

templates/api/knowledge-panels/health/ingredients/ingredients.tt.json
stephanegigandet: ah it should be if we have an ingredients_list panel, but the condition I put before didn't work, so I just put 1. I'll change that.

teolemon:
@stephanegigandet some feedback from this morning's battle testing.
- Putting the percentage estimate in the title is too proeminent (it should be in the card with a disclaimer) + it wasn't localized (eg: "6% estimation")
- Mixing English and French is confusing (on products with multiple languages)
- Sometimes there are hyphen before some ingredients (sub-ingredients ?)

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
alexgarel: ok cool, that's what I though first, but down in the same file you put edq inside backtick !

stephanegigandet:
Thanks for the feedback @teolemon 

> Putting the percentage estimate in the title is too proeminent (it should be in the card with a disclaimer) + it wasn't localized (eg: "6% estimation")

I'm not sure what you mean by "it should be in the card", if we only see the % of the ingredient when we click on it, it's less useful than being able to see the % for all ingredients. That's one of the added value compared to the original list of ingredients from the package:

image

Not sure either about the localization, today we display "Potato (6% estimate)" in English and "Pomme de terre (6% estimation)" in French, so it is localized. Maybe "estimate" takes too much room? We could do ~6%, but it's less explicit.

> Mixing English and French is confusing (on products with multiple languages)

We display the text ingredient list in the user's language if we do have it, and we display the ingredients panels in the user language in all cases. I would think it's more useful that way?

> Sometimes there are hyphen before some ingredients (sub-ingredients ?)

Yes, this is to convey that there sub ingredients (or sub sub sub ingredients). Not ideal, but better than a completely flat list I guess. We could try to indent more instead of using hyphens, but we would need to extend a bit the knowledge panels grammar.

stephanegigandet:
I changed a bit the display:

image

stephanegigandet:
I changed a bit the display:

image

stephanegigandet:
Related issue on the mobile app: https://github.com/openfoodfacts/smooth-app/issues/5700

teolemon:
>  "it should be in the card", (....) That's one of the added value compared to the original list of ingredients from the package:

Probably we should have vegan, vegetarian, icons next to them, plus whether the ingredient is an agribalyse/ecobalyse match. Your proposal regarding rounding might solve my concerns (it was looking ridiculous to have 10 ingredients+ including glucose syrup at 0%)

> Not sure either about the localization, today we display "Potato (6% estimate)" in English and "Pomme de terre (6% estimation)" in French, so it is localized. Maybe "estimate" takes too much room? We could do ~6%, but it's less explicit.

"estimation 6%" or "estimé à 6%" or "6% estimé" sounds more French than "6% estimation". or at least with an hyphen "6% - Estimation"

> Mixing English and French is confusing (on products with multiple languages)

I had the case of a product with a German ingredient list which included glucose syrup. In the app we showed the French list, but the German normalized ingredients were shown

> Sometimes there are hyphen before some ingredients (sub-ingredients ?)

Quick n Dirty fix: ↳
[17:45](https://openfoodfacts.slack.com/archives/D02KZKKAH/p1729093557098879)
https://www.alt-codes.net/arrow_alt_codes.php

stephanegigandet:
> Probably we should have vegan, vegetarian, icons next to them, plus whether the ingredient is an agribalyse/ecobalyse match.

So at this point, we are showing each ingredient in a knowledge panel title. In the current version of knowledge panels, we can display 1 icon at the left, but that's it. We can add more text, but adding icons would require changes to the knowledge panel API + its implementation on web and mobile.

> I had the case of a product with a German ingredient list which included glucose syrup. In the app we showed the French list, but the German normalized ingredients were shown

We analyze ingredients only in 1 language, and we use the ingredient analysis to display the list of ingredients, so that's the current "normal" behaviour.  

For the sub-ingredients, I'm not sure if the down + right arrow is more clear. It works well with 1 level, but multiple levels is less obvious:

image

teolemon:
- @stephanegigandet I've made further proposals there: https://github.com/openfoodfacts/openfoodfacts-server/issues/10906

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -285,6 +289,30 @@ sub convert_multiline_string_to_singleline ($line) {
...
+Helper function to allow to enter multiline strings in JSON templates.
+The function converts the multiline string into a single line string.
+
+Line breaks are converted to spaces, and multiple spaces are converted to a single space.
+
+This function is useful in templates where we use IF statements etc. to generate a single value like a title.
+
+=cut
+
+sub convert_multiline_string_to_singleline_without_line_breaks_and_extra_spaces ($line) {
```
alexgarel, line 303: Good idea :-)

templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
```diff
@@ -4,24 +4,33 @@
...
     "size": "small",
-    [% IF panel.ingredient_description OR panel.wikipedia_abstract %]
-    "expanded": false,
-    [% ELSE %]
-    "expanded": false,
-    [% END %]
     "title_element": {
-        // Note: the app displays line feeds as line feeds...
-        "title": `        [% dash = "-" %]        [% dash.repeat(panel.level) %]        [% display_taxonomy_tag_name("ingredients", panel.ingredient_id) %]        [% IF panel.ingredient.percent.defined %]            ([% round(panel.ingredient.percent) %]%)        [% ELSIF panel.ingredient.percent_estimate.defined %]            ([% round(panel.ingredient.percent_estimate) %]% [% lang("estimate") %])        [% END %]        `,
+        // double backticks: convert to a single line without newlines and extra spaces
```
alexgarel, line 8: :astonished: cool !

alexgarel (approved):

teolemon:
@stephanegigandet some merge conflicts. 
The airing of the TV show is getting extremely near.

PR #10915, opened 2024-10-21
fix: attribute for product without new nutriscore data

stephanegigandet:
Some products (in the obsolete collection) do not have new Nutri-Score data (but may have data in an older format).  In that case the attribute Nutri-Score should be unknown. This can also happen if we display an older revision of a product.

I'll also recompute the Nutri-Score for the current revision of obsolete products.

Fixes https://github.com/openfoodfacts/smooth-app/issues/5699


stephanegigandet:
"make check_critic" passes locally, no clue why it fails on github...

alexgarel (approved):

PR #10920, opened 2024-10-21
fix: allow CORS requests for product image move

raphael0202:
It's required by Nutripatrol to perform image deletion from nutripatrol.openfoodfacts.org


stephanegigandet (approved):

stephanegigandet:
@raphael0202 looks good, but it would be good to add an integration test to make sure we don't break it in the future

raphael0202:
I looked into integration tests, but I think it will take me quite some times to get it right. I'm merging it for now.

PR #10921, opened 2024-10-21
fix: cosmetic changes for new Eco-Score grades

stephanegigandet:
This is to display better the knowledge panels for Eco-Score A+ and F

### DIFF
--- a/po/common/fr.po
+++ b/po/common/fr.po
@@ -4764,6 +4764,10 @@ msgctxt "attribute_ecoscore_grade_title"
 msgid "Eco-Score %s"
 msgstr "Eco-Score %s"
 
+msgctxt "attribute_ecoscore_a_plus_description_short"
+msgid "Very low environmental impact"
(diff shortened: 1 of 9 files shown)


po/common/fr.po
```diff
@@ -4764,6 +4764,10 @@ msgctxt "attribute_ecoscore_grade_title"
 msgid "Eco-Score %s"
 msgstr "Eco-Score %s"
 
+msgctxt "attribute_ecoscore_a_plus_description_short"
```
teolemon, line 4767: Do you have those strings in common.pot ?

teolemon (approved):

po/common/fr.po
alexgarel: I verified, they exists, see https://github.com/openfoodfacts/openfoodfacts-server/blob/ecoscore2/po/common/common.pot

alexgarel (approved):
Great

PR #10929, opened 2024-10-24
taxonomy: boolean added for european countries + ecobalyse id is adapted for organic ingredients

very-smartin:
### What

In order to get more Ecobalyse ids more precisely, a boolean for european countries has been added in the taxonomy.
Furthermore, the ecobalyse id en:ecobalyse_label_organic is retrieved for organic ingredients.


taxonomies/countries.txt
```diff
@@ -3332,6 +3332,7 @@ country_code_3:en: AUT
 language_codes:en: de
 osm_relation:en: 16239
 wikidata:en: Q40
+ecobalyse_is_part_of_eu: 1
```
stephanegigandet, line 3335: for boolean, by convention in taxonomies we use properties like "vegan:en: yes"
can you replace "ecobalyse_is_part_of_eu: 1" by "is_part_of_eu:en: yes" ?

Unless Ecobalyse has specific countries it considers to be part of EU, let's make the property name is_part_of_eu:en:

very-smartin: Alright i'll change this ! 

Indeed, Ecobalyse considers specific countries to be part of EU sometimes in order to get realistic LCA (as an example the UK is not anymore in the EU but for transport impact it's considered so) so i would recommand to keep 'ecobalyse_is_part_of_eu'.

lib/ProductOpener/Ingredients.pm
```diff
@@ -3371,20 +3371,47 @@ sub get_missing_ecobalyse_ids ($ingredients_ref) {
 		# Getting properties from product
 		#my $product_origin = has_specific_ingredient_property()
 
-		# Getting the correct ecobalyse comake devde
-		my $ecobalyse_code = get_inherited_property("ingredients", $ingredient_ref->{id}, "ecobalyse:en");
-		if (defined $ecobalyse_code) {
-			$ingredient_ref->{ecobalyse_code} = $ecobalyse_code;
+		# If the ingredient is organic, the 'ecobalyse_labels_en_organic:en' id is used
+		if ($ingredient_ref->{labels} =~ /\ben:organic\b/) {
```
stephanegigandet, line 3375: This approach won't work if we have an organic ingredient for which we have a general ecobalyse match, but no organic ecobalyse match.

e.g. if we have "organic pear", in the ingredients, we don't have it in Ecobalyse:

ecobalyse:en: pear-eu
ecobalyse_origins_en_france:en: pear-fr

but we need to fallback on pear-eu.

So I suggest a different approach:

Based on the labels, origins etc. of the product, construct a list of prioritized ecobalyse suffixes.

e.g. if the product is "organic pear from france", we might want to get:

1. ecobalyse_origins_france_label_organic
2. ecobalyse_origins_european-union_label_organic
3. ecobalyse_label_organic
4. ecobalyse_origins_france
5. ecobalyse_origins_european-uinion
6. ecobalyse

(or maybe a different order, depending on what we consider more important: the origin or the label)

Then we check if we have those as ecobalyse properties for "en:pear", and we stop at the first one we find.

very-smartin: Good point indeed, i'll change my code consequently

very-smartin:
@stephanegigandet Note : for now i've only added the code in case of organic ingredients, the code for the origin will be done later

lib/ProductOpener/Ingredients.pm
```diff
@@ -3368,23 +3368,52 @@ sub get_missing_ecobalyse_ids ($ingredients_ref) {
...
+    	# ecobalyse_origins_france							(if the product comes from france)
+	    # ecobalyse_origins_european-uinion					(if the product comes from the Europe region)
+    	# ecobalyse 										(else)
+
+
+		# If the ingredient is organic...
+		if ($ingredient_ref->{labels} =~ /\ben:organic\b/) {
+
+		    # Retrieve the correct ecobalyse code
+		    my $ecobalyse_code = get_inherited_property("ingredients", $ingredient_ref->{id}, "ecobalyse_labels_en_organic:en");
```
stephanegigandet, line 3384: I think it's easier to build a list of  the possible suffixes, because even if we just want to support organic, we need to check those:

1. ecobalyse_labels_en_organic:en
2. ecobalyse_proxy_labels_en_organic:en
3. ecobalyse_en_organic:en
4. ecobalyse_proxy_en_organic:en

Here your current code checks 1. and 4.

You could add more if clauses to cover 2 and 3, but it's going to become difficult to read with a lot of duplicate code.

Instead just do something like this (pseudo code):

my @suffixes = ();
if (label organic) : push @suffixes, "_labels_organic"
push @suffixes, ''; # no suffix when we don't have anything

then a foreach loop on suffixes:

foreach my $suffix (@suffixes) {

foreach my $proxy ("", "_proxy") {

my $ecobalyse_code = get_inherited_property("ingredients", $ingredients_ref->{id}, "ecobalyse" . $proxy . $suffix)
if defined $ecobalyse_code {

..
last;
}
˚

very-smartin: Indeed it's a much more better solution, i hadn't seen it like this : new pr incoming !

very-smartin:
Note : doing the tests on my laptop takes more time than pushing the modifications and looking for mistakes on github

lib/ProductOpener/Ingredients.pm
```diff
@@ -3368,23 +3368,47 @@ sub get_missing_ecobalyse_ids ($ingredients_ref) {
...
+		foreach my $suffix (@suffixes) {
+			# Construct the property name using the proxy and suffix
+			my $property_name = "ecobalyse" . $suffix . ":en";
+
+			# Attempt to retrieve the ecobalyse code for the current property name
+			my $ecobalyse_code = get_inherited_property("ingredients", $ingredient_ref->{id}, $property_name);
+
+			if (defined $ecobalyse_code) {
+				# Assign the ecobalyse code if found
+				$ingredient_ref->{ecobalyse_code} = $ecobalyse_code;
```
stephanegigandet, line 3402: If the match is on a proxy, we want to set ecobalyse_proxy_code instead of ecobalyse_code

lib/ProductOpener/Ingredients.pm
```diff
@@ -3368,23 +3368,47 @@ sub get_missing_ecobalyse_ids ($ingredients_ref) {
...
+		# List of suffixes
+		my @suffixes = ();
+		# If the ingredient is organic...
+		if ($ingredient_ref->{labels} =~ /\ben:organic\b/) {
+			push @suffixes, "_labels_en_organic";
+		}
+		push @suffixes, '';
+
+		if ($ingredient_ref->{labels} =~ /\ben:organic\b/) {
+			push @suffixes, "_proxy_labels_en_organic";
```
stephanegigandet, line 3388: Instead of adding _proxy_* suffixes that are the same as the non proxy suffixes, it's probably better to have only the non proxy suffixes:
- we first loop on it to see if we have an exact match, and if so we set ecobalyse_code
- if we don't have a match, we loop on it again (and add the proxy_), if we set ecobalyse_proxy_code

PR #10931, opened 2024-10-24
fix: donation page now uses Matomo like the rest of OFF

teolemon:
### What
fix: donation page now uses Matomo like the rest of OFF


alexgarel (approved):
Thanks !

PR #10952, opened 2024-10-30
ci: cache taxonomies in pr workflow

McCio:
cache taxonomies in pr workflow

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -58,10 +58,16 @@ jobs:
       with:
         # needs depth to run git log below
         fetch-depth: 50
+    - uses: actions/cache@v4
+      id: cache
+      with:
+        path: ./build-cache
+        key: taxonomies-${{ hashFiles('taxonomies/**') }}
+        restore-keys: taxonomies-
     - name: build
       run: make build container=backend
     - name: push backend image as artifact
-      uses: ishworkh/docker-image-artifact-upload@v1
+      uses: ishworkh/container-image-artifact-upload@v2.0.0
       with:
         image: "openfoodfacts-server/backend:dev"
     - name: Fetch origin/main
@@ -75,7 +81,7 @@ jobs:
       run: |
         git ls-files taxonomies/ | xargs -I{} git log -1 --date=format:%Y%m%d%H%M.%S --format='touch -t %ad "{}"' "{}" | bash
     - name: Rebuild taxonomies
-      run: make build_taxonomies GITHUB_TOKEN="${{ secrets.TAXONOMY_CACHE_GITHUB_TOKEN }}"
+      run: make DOCKER_LOCAL_DATA="$(pwd)" build_taxonomies GITHUB_TOKEN="${{ secrets.TAXONOMY_CACHE_GITHUB_TOKEN }}"
 
   check_perl:
     name: 🐪 Check Perl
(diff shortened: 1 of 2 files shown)


McCio:
@alexgarel @stephanegigandet as I noted on slack some time ago, currently the github cache is not used during builds - with this PR I'm using the standard cache from github actions (with a bit of a hack copy/pasting from/to the build_cache volume).
As you can see in the timings of the builds, it's significantly faster. Also for new PRs, it will be using `main`'s cache as start point, so it should make the whole pr builds 30 minutes faster (1h less runtime, from [~1h50m ](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/11591731210/usage) to [~50m](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/11597230714/usage?pr=10952)) when not rebuilding the taxonomy.

The downpoint is that it isn't usable for local development, as far as I know.
If you prefer to keep the github token approach - feel free to close this PR without merging, I already had fun getting to know some github actions :smiley:

.github/workflows/pull_request.yml
```diff
@@ -58,10 +58,26 @@ jobs:
...
+      id: cache
+      with:
+        path: ./build-cache
+        key: taxonomies-${{ hashFiles('taxonomies/**') }}
+        restore-keys: taxonomies-
+    - if: steps.cache.outputs.cache-hit == 'true'
+      name: Prepare build_cache volumes
+      run: |
+        proj=po_off
+        docker volume create \
```
john-gom, line 71: Would it be possible to use something like [--opt device=./build-cache](https://docs.docker.com/reference/cli/docker/volume/create/#opt), i.e. mount the folder directly, to avoid having to copy files to and from the docker volume?

.github/workflows/pull_request.yml
```diff
@@ -96,8 +115,23 @@ jobs:
...
+      run: |
+        docker volume create \
+                   --label com.docker.compose.project=po_off \
+                   --label com.docker.compose.version=2.29.7 \
+                   --label com.docker.compose.volume=build_cache \
+                   po_off_build_cache
+        docker container run --rm -v ./build-cache:/from -v po_off_build_cache:/to alpine ash -c 'cd /from ; cp -av . /to'
     - name: Download backend image from artifacts
-      uses: ishworkh/docker-image-artifact-download@v1
+      uses: ishworkh/container-image-artifact-download@v2.0.0
```
john-gom, line 134: Is this needed for this particular change or is it just a general version update?

.github/workflows/pull_request.yml
```diff
@@ -58,10 +58,26 @@ jobs:
...
+    - uses: actions/cache@v4
+      id: cache
+      with:
+        path: ./build-cache
+        key: taxonomies-${{ hashFiles('taxonomies/**') }}
+        restore-keys: taxonomies-
+    - if: steps.cache.outputs.cache-hit == 'true'
+      name: Prepare build_cache volumes
+      run: |
+        proj=po_off
```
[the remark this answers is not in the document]
McCio: interesting, I'll look into that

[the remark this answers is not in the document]
McCio: done :smile:

.github/workflows/pull_request.yml
McCio: done :smile:

john-gom:
@alexgarel , @stephanegigandet changes look good to me but I'm about to go away until Monday so won't be able to test for any issues running locally until then, not sure if you will get the chance either...

McCio:
Just realigned to `main` for simplicity

alexgarel:
I tested it locally and nothing is broken !

Thank @McCio  for this PR :tada:

alexgarel (approved):

Issue #9221, opened 2023-10-27
Image transfer between OFF and OPF/OBF/OPFF


Sebleouf:
[I hope this issue doesn't already exist, I couldn't find the documentation for it.]

It is currently impossible to transfer an image from OFF to a sister project (OPF/OBF/OPFF) if a product already exists on the target site.

There are currently two cases:
- the barcode does not exist on the target site, so all the images and data on OFF can be transferred to the target site
- the barcode already exists on the target site, an error message appears "Please correct the following errors: A product already exists with the new code". It is therefore impossible to transfer an image and/or data to the target site.

This possibility needs to be unblocked, and I imagine it can be done by following the same principle of moving images from one product to another, which has worked very well on OFF for many years.

The challenge would be to empty this category (over 2,000 products) for example: https://world.openfoodfacts.org/category/open-beauty-facts; which is far from complete (I think there is more than 10k). This would allow more photos in the target projects and also improve the overall quality of the OFF database. 
Once I've solved this problem, I'll be able to transfer the photos to thousands of products myself, as this is very important to me. However, if it's already possible to automate the transfer to the OBF category, that would be great!

Thank you in advance for dealing with this issue.

Issue #7539, opened 2022-10-13
NEW DESIGN: Strange error message when moving products to a different server (e.g. Open Beauty Facts)


aleene:
### Describe the bug

After moving a product to OBF.

### To Reproduce

Try moving a product.

### Expected behavior

Something nice.

### Screenshots

<img width="382" alt="Screenshot 2022-10-13 at 16 31 12" src="https://user-images.githubusercontent.com/1329905/195626817-2c8e58cf-6403-4fdd-b159-3e2f55d725d9.png">

### Type of device

Browser

aleene:
<img width="382" alt="Screenshot 2022-10-13 at 16 31 12" src="https://user-images.githubusercontent.com/1329905/195626699-d556ac1f-2873-4100-a0d0-8f3cd9f28e59.png">

teolemon:
@aleene @stephanegigandet we can close this

Issue #1174, opened 2018-04-02
Create a new API that checks a barcode on all 4 databases


teolemon:
Create a new API that checks a barcode on all 4 databases (off, obf, opf, opff) and returns the name of the matching db, for moderation purposes.

### Part of
- #5536
- #5505 
- #5555

stephanegigandet:
One possibility could be to merge the 4 databases into one, and have a tag on each product to indicate which database(s ?) it is in.

Issues with that:
- if we want to move the sites and database to different servers, it's not possible anymore
- there will be more barcode conflicts, especially in the self allocated 2* codes

hangy:
An advantage would be that moving products would be much easier?

teolemon:
Related: https://github.com/openfoodfacts/openfoodfacts-server/issues/797

teolemon:
- Dart side issue: https://github.com/openfoodfacts/openfoodfacts-dart/issues/972

Issue #497, opened 2016-09-26
Create placeholder response if the product is present on Open (Beauty/PetFood/Products) Facts


teolemon:
#### Web: 
- Add a link to the product on Open (Beauty/PetFood/Products) Facts
- Do not make the placeholder indexable

#### Mobile:
- Send back a product with title: "Present on Open Beauty Facts" and the OBF url as URL on official site?

### Technical decision
- https://docs.google.com/document/d/1WPD23Jf-GMstjiK2gOZGayqECwxtucFqr7ijuyxDN4Q/edit

### Part of
- #5505 
- #5536 
- #5555

hangy:
Right now, this stuff runs on the same server, so some hacks could make this easy, but for long-term support (and possibly to support other future projects), this would need some "interwiki" like feature/API right? This probably also applies #319 and #320.

teolemon:
Yes, long term, we need a decent system. But a hack would be very useful to improve the mobile user experience (at the minimum propose to install Open Beauty Facts on iPhone, or even customize behaviour (on Android))

teolemon:
Just reading the .sto file and returning `cosmetic=true` would be enough for the mobile teams to solve the issue.

teolemon:
This could be a new small new route to avoid breaking anything.

teolemon:
I'm adding the technical decision taken a while ago on universal product API @stephanegigandet @hangy @g123k : https://docs.google.com/document/d/1WPD23Jf-GMstjiK2gOZGayqECwxtucFqr7ijuyxDN4Q/edit

PR #10959, opened 2024-10-31
refactor: merge the products and product images directories for OFF, OBF, OPF and OPFF

stephanegigandet:
In the last few weeks, we normalized product barcodes and deduplicated products that existed in multiple flavors (Open Food Facts, Open Beauty Facts, Open Product Facts and Open Pet Food Facts).

We are now going to merge the directories that contain product and product revision data (the .sto files) and the product images. We will keep separate MongoDB databases for each flavor.

This will:
- make it much easier to move products from one product type to another
- remove the possibility of having duplicate products on multiple flavors
- make it much easier to have read and write APIs that can be used to retrieve / update products of any type.

Deployment plan:
- stop OBF, OPF, OPFF for the duration of the migration (a couple of hours)
- update OFF code
- move products and product images dirs from OBF, OPF, OPFF to the OFF directory structure
- change the links to products and products of images on OBF, OPF, OPFF to use the OFF dirs
- restart OBF, OPF, OPFF with the new code

In this PR:
- code updates to check product_type when reading / editing a product for both the website and API. Redirect to the right flavor if the product type is different than the one of the server the request was made to.
- changed the product edit form to allow moderators to change the product_type
- keep support for moderators to put "obf" etc. in the change code field
- tests for the above
- small refactor to put into Config.pm things that are the same for all flavors and that are currently duplicated in Config_off.pm etc.

Will solve:
- #9221 
- #7539
- #1174
- #497
- #320


lib/ProductOpener/Config_off.pm
```diff
@@ -1099,30 +1084,22 @@ $options{current_server} = "off";
 $options{other_servers} = {
```
alexgarel, line 1084: why do we keep it here ?

lib/ProductOpener/Products.pm
```diff
@@ -1024,52 +1025,68 @@ sub retrieve_product_rev ($product_id, $rev, $include_deleted = 0) {
...
+				$log->info("changing code", {old_code => $product_ref->{old_code}, code => $code})
+					if $log->is_info();
+			}
 		}
 	}
 
 	return;
 }
 
+sub change_product_type ($product_ref, $new_product_type, $errors_ref) {
```
alexgarel, line 1065: Adding a POD would be interesting.

lib/ProductOpener/Products.pm
```diff
@@ -1163,22 +1180,53 @@ sub store_product ($user_id, $product_ref, $comment) {
...
+	my $delete_from_previous_products_collection = 0;
 
 	# if we have a "server" value (e.g. from an import),
 	# we save the product on the corresponding server but we don't need to move an existing product
+	if (defined $product_ref->{server}) {
+		my $new_server = $product_ref->{server};
+		# Update the product_type from the server
+		if (defined $options{flavors_product_types}{$new_server}) {
+			my $errors_ref = {};
+			change_product_type($product_ref, $options{flavors_product_types}{$new_server}, $errors_ref);
```
alexgarel, line 1192: In case of error, shan't we die ?

alexgarel (approved):
Great ! Kudos for the though work.

I have some nitpicking comments.

lib/ProductOpener/Products.pm
stephanegigandet: Diying is a bit harsh, store_product() is called from batch scripts that update / import tons of products, we don't want it to die if there's one issue with one product. I will log an error.

lib/ProductOpener/Products.pm
stephanegigandet: Good idea. I will also remove the feature of being able to change the product type by putting "obf" in the barcode field.

lib/ProductOpener/Config_off.pm
stephanegigandet: I removed it

lib/ProductOpener/APIProductWrite.pm
```diff
@@ -465,21 +482,82 @@ sub write_product_api ($request_ref) {
...
 		}
 		else {
-			# Update the product
-			update_product_fields($request_ref, $product_ref, $response_ref);
 
-			# Process the product data
-			analyze_and_enrich_product_data($product_ref, $response_ref);
+			# Change of code
+			if (    (defined $request_body_ref->{product}{code})
+				and ($product_ref->{code} ne $request_body_ref->{product}{code}))
```
alexgarel, line 488: I'm a bit disappointed that we have to do those checks here and in product_jqm_multilingual… could'nt we factor this out ?

lib/ProductOpener/Config.pm
```diff
@@ -33,8 +33,141 @@ if (not defined $flavor) {
 }
 
 use Module::Load;
+
 autoload("ProductOpener::Config_$flavor");
 
-#
+# Add values common to all flavors
```
alexgarel, line 39: Thanks for simplifying this !

lib/ProductOpener/Display.pm
```diff
@@ -10599,6 +10614,24 @@ sub display_product_api ($request_ref) {
 			$response{jqm} .= $html;
 		}
 	}
+	elsif ((defined $product_ref->{product_type}) and ($product_ref->{product_type} ne $options{product_type})) {
```
alexgarel, line 10617: there again we have a complete duplication for product_multilingual.pl, that's a bit frustrating…

lib/ProductOpener/Products.pm
```diff
@@ -106,7 +106,8 @@ BEGIN {
 		&product_data_is_protected
 
 		&make_sure_numbers_are_stored_as_numbers
-		&change_product_server_or_code
+		&change_product_code
+		&change_product_type
```
alexgarel, line 110: That's great we de-entangled both actions !

lib/ProductOpener/Products.pm
```diff
@@ -1022,54 +1023,93 @@ sub retrieve_product_rev ($product_id, $rev, $include_deleted = 0) {
...
+		if (-e "$data_root/products/" . product_path_from_id($new_code) . "/product.sto") {
+			$log->warn("cannot change product code, because the new code already exists",
+				{code => $code, new_code => $new_code})
+				if $log->is_warn();
+			return "error_new_code_already_exists";
 		}
 		else {
 			$product_ref->{old_code} = $code;
 			$code = $new_code;
 			$product_ref->{code} = $code;
```
alexgarel, line 1065:
```suggestion
			# we only register the code change,
			# file modification will happen as we save the product sto
			$product_ref->{old_code} = $code;
			$code = $new_code;
			$product_ref->{code} = $code;
```

lib/ProductOpener/Products.pm
```diff
@@ -1022,54 +1023,93 @@ sub retrieve_product_rev ($product_id, $rev, $include_deleted = 0) {
...
+
+	# Currently only called by admins and moderators
+
+	my $product_type = $product_ref->{product_type};
+
+	# Return if the product type is already the new product type, or if the new product type is not defined
+	if ((not defined $new_product_type) or ((not defined $options{product_types_flavors}{$new_product_type}))) {
+		return "invalid_product_type";
+	}
+	elsif ($product_type ne $new_product_type) {
```
alexgarel, line 1102: shan't we at least emit a warning if the new product type is the same as the old one ?

lib/ProductOpener/Products.pm
alexgarel: I think scripts should call store_product with a try / catch !
Or maybe we could add a parameter to say we wan't a fail-safe call.

lib/ProductOpener/Products.pm
```diff
@@ -1303,15 +1362,15 @@ sub store_product ($user_id, $product_ref, $comment) {
 				"moving product images",
 				{
 					source => "$BASE_DIRS{PRODUCTS_IMAGES}/$old_path",
-					destination => "$new_www_root/images/products/$path"
+					destination => "$www_root/images/products/$path"
```
alexgarel, line 1365: In this whole sections you should not use $www_root and $data_root any more…
Please use `%BASE_DIRS` values …

scripts/migrations/2024_11_move_obf_opf_opff_products_to_off_dirs.pl
```diff
@@ -0,0 +1,258 @@
...
+
+# We removed the other servers dirs from the base paths,
+# but for this script, we need /srv/off/products and /srv/off/html/images/products
+
+$BASE_DIRS{OFF_PRODUCTS} = "/srv/off/products";
+$BASE_DIRS{OFF_PRODUCTS_IMAGES} = "/srv/off/html/images/products";
+
+sub move_product_dir_to_off ($dir, $dir2, $dir3, $dir4) {
+	# move .sto files
+	print STDERR "moving $dir/$dir2/$dir3/$dir4\n";
```
alexgarel, line 78: For me, it's incredible that in this whole function you never put $dir/$dir2/$dir3/$dir4 in a variable to avoid copying it everywhere around !

scripts/migrations/2024_11_move_obf_opf_opff_products_to_off_dirs.pl
```diff
@@ -0,0 +1,258 @@
...
+			print STDERR
+				"moved /mnt/$flavor/images/products/$dir/$dir2/$dir3/$dir4 to /srv/off/html/images/products/$dir/$dir2/$dir3/$dir4\n";
+		}
+		else {
+			print STDERR
+				"could not move /mnt/$flavor/images/products/$dir/$dir2/$dir3/$dir4 to /srv/off/html/images/products/$dir/$dir2/$dir3/$dir4: $!\n";
+			die;
+		}
+	}
+	die;
```
alexgarel, line 148: why do you need this die ?

tests/integration/change_product_code_and_product_type.t
alexgarel, line 1: :+1:

alexgarel (approved):

scripts/migrations/2024_11_move_obf_opf_opff_products_to_off_dirs.pl
stephanegigandet: I launched the script product by product for the first products to verify everything was working as expected as I couldn't replicate everything in the dev environment. I removed the die() in opf etc. after that to do all the products.

scripts/migrations/2024_11_move_obf_opf_opff_products_to_off_dirs.pl
stephanegigandet: It's just a one off script that will never be used again, so I mainly wanted to get this thing done as quickly as possible.

lib/ProductOpener/Products.pm
stephanegigandet: From the API point of view, it just sends product_type like it could send product_name_en, we don't return warnings if the product name is already the same. In practice change_product_type() is called only if the product_type is different.

PR #10966, opened 2024-11-03
taxonomy: Add french term for E160b(ii)

sparunakian:
### What

Added a french term for E160b(ii). There was none.


stephanegigandet (approved):
Thanks

PR #10973, opened 2024-11-04
refactor: gen_feeds_daily.pl (exports and database dumps) for all flavors

stephanegigandet:
Now the the code bases for OFF, OPF, OBF and OPFF are unified, we can have a single gen_feeds_daily.pl file for all flavors, as the differences are minimal (only the pro platform off-pro is different).

This PR also removes hard coded paths like /srv/off/ so that the script can be run in a dev environment (some utilities like python, mongodbdump etc. are missing from dev but the rest of the script can run).


stephanegigandet:
installed missing pigz on opf, opff, obf containers

scripts/gen_feeds_daily.sh
```diff
@@ -1,14 +1,104 @@
...
+    en.$PRODUCT_OPENER_DOMAIN.products.rdf \
+    fr.$PRODUCT_OPENER_DOMAIN.products.csv \
+    fr.$PRODUCT_OPENER_DOMAIN.products.csv.gz \
+    fr.$PRODUCT_OPENER_DOMAIN.products.rdf \
+    s3/openfoodfacts-ds
+
+# Generate the MongoDB dumps and jsonl export
+cd $OFF_SCRIPTS_DIR
+
+./mongodb_dump.sh $OFF_PUBLIC_DATA_DIR $PRODUCT_OPENER_FLAVOR 10.1.0.102 $PRODUCT_OPENER_FLAVOR_SHORT
```
alexgarel, line 69: nit: could we use mongodb address from config here, instead of hardcoded ?

alexgarel (approved):
Great move

scripts/gen_feeds_daily.sh
```diff
@@ -1,14 +1,104 @@
...
+    ./export_products_data_and_images.pl --sample-mod 1000,0 \
+        --products-file $OFF_PUBLIC_EXPORTS_DIR/products.random-modulo-1000.tar.gz \
+        --images-file $OFF_PUBLIC_EXPORTS_DIR/products.random-modulo-1000.images.tar.gz \
+        --jsonl-file $OFF_PUBLIC_EXPORTS_DIR/products.random-modulo-1000.jsonl.gz \
+        --mongo-file $OFF_PUBLIC_EXPORTS_DIR/products.random-modulo-1000.mongodbdump.gz
+fi
+
+./generate_dump_for_offline_apps_off.py
+cd $OFF_PUBLIC_DATA_DIR/offline
+zip en.$PRODUCT_OPENER_DOMAIN.products.small.csv.zip en.$PRODUCT_OPENER_DOMAIN.products.small.csv
```
teolemon, line 90: @stephanegigandet can you duplicate the output for OBF, OPF and OPFF ?

teolemon (changes requested):
@stephanegigandet could you replicate the mini output for other flavors, I've made a matching PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/10977

scripts/gen_feeds_daily.sh
stephanegigandet: good idea, I will do that

stephanegigandet:
> @stephanegigandet could you replicate the mini output for other flavors, I've made a matching PR: #10977

sure, but I'll make the script generic so that we don't have have 4 scripts

scripts/generate_dump_for_offline_apps.py
```diff
@@ -0,0 +1,26 @@
...
+    off_public_data_dir = os.getenv('OFF_PUBLIC_DATA_DIR')
+    product_opener_flavor = os.getenv('PRODUCT_OPENER_FLAVOR')
+    product_opener_flavor_short = os.getenv('PRODUCT_OPENER_FLAVOR_SHORT')
+
+    if not os.path.exists(off_public_data_dir + '/offline'):
+        os.makedirs(off_public_data_dir + '/offline')
+        
+    import pandas
+    df = pandas.read_csv(off_public_data_dir + '/' + product_opener_flavor + '/en.' + product_opener_flavor_short + '.products.csv', sep='\t', low_memory=False)
+    colnames = ['code','product_name','quantity','brands','nutriscore_grade','nova_group','ecoscore_grade']
```
teolemon, line 21: this won't work for non off flavors

teolemon: cf my other scripts, where I remove the final 3 columns

stephanegigandet: the app can just ignore the columns it doesn't need

stephanegigandet:
> @stephanegigandet could you replicate the mini output for other flavors, I've made a matching PR: #10977

@teolemon can you approve the PR and remove the change request? we can work out the generation of the small CSV in a new PR, so that it doesn't hold fixing the OBF etc. exports.

teolemon (approved):

PR #10974, opened 2024-11-04
feat: update Italian packager codes adding script

McCio:
### What

* added the script to update Italian packager codes
* added make action to update .sto files about packager codes
* actually updated Italian packager codes and .sto files


McCio:
As a small note, it shows more removed rows than added because in the old csv many companies were duplicated. I manually verified that all the old ones are still present.

stephanegigandet (approved):
Thank you!

PR #10977, opened 2024-11-05
feat: Generate miniaturized dumps of Open Beauty, Pet Food and Products Facts

teolemon:
### What
- feat: Generate miniaturized dumps of Open Beauty, Pet Food and Products Facts, to let the mobile app know which products exist.


alexgarel:
We are doing this in https://github.com/openfoodfacts/openfoodfacts-exports

PR #10978, opened 2024-11-06
ci: issue template improvements

teolemon:
### What
- issue template improvements


stephanegigandet (approved):
We should discuss and decide on the use (or the non use) of emojis in types / labels

PR #10991, opened 2024-11-07
feat: nutripatrol product report

alexgarel:
First Nutri-Patrol integration.

image


tests/integration/expected_test_results/api_v2_product_read/get-fields-attribute-groups-all-knowledge-panels.json
```diff
@@ -1992,34 +1992,10 @@
...
+            "json_debug_url" : "http://static.openfoodfacts.localhost/files/debug/knowledge_panels/incomplete_or_incorrect_data0200000000034.json",
+            "json_error" : "malformed JSON string, neither tag, array, object, number, string or atom, at character offset 519 (before \"`\\n                I...\") at /opt/product-opener/lib/ProductOpener/KnowledgePanels.pm line 411.\n",
```
stephanegigandet, line 1997: There's a typo in the template, fixing it.

lib/ProductOpener/Products.pm
```diff
@@ -2828,14 +2829,25 @@ my %actions_urls = (
...
 
 sub product_action_url ($code, $action) {
 
-	my $url = "/cgi/product.pl?type=edit&code=" . $code;
-
+	my $url;
 	if (defined $actions_urls{$action}) {
-		$url .= $actions_urls{$action};
+		my $action_url = $actions_urls{$action};
+		if ($action_url =~ /^#/) {
```
stephanegigandet, line 2841: The "edit_product" action has an empty url "" because before we always started with the edit form URL. So we need a change like this:

```suggestion
		if (($action eq '') or (($action_url =~ /^#/)) {
```

alexgarel: well spotted !

stephanegigandet (approved):
Thank you!

stephanegigandet (commented):
@alexgarel I was surprised that the integration test expected files were not changed, I think it may be due to the fact that the environment file used in tests doesn't have the NUTRIPATROL_URL.

Shouldn't it be added to https://github.com/openfoodfacts/openfoodfacts-server/blob/main/.env ?

alexgarel:
@teolemon, @Valimp and @g123k this PR adds a new action "report_product_to_nutripatrol" to knowledge panels, which is not yet implemented in the mobile app.

Is it ok, or should I only display the panel for the web ?

stephanegigandet (approved):
Perfect, thanks @alexgarel

alexgarel:
@stephanegigandet we don't merge immediately, we need to hide the panel for mobile until they implement the action.

PR #10999, opened 2024-11-08
fix: improve parsing of 'category (type 1, type 2..)' ingredients WIP

stephanegigandet:
PR to better handle things like "vegetal oil (palm, rapeseed)":

- instead of turning "vegetal oil (palm, rapeseed)" to "palm vegetal oil", "rapeseed vegetal oil", we now turn it to "vegetal oil (palm vegetal oil, rapeseed vegetal oil)", as keeping a parent ingredient is better for ingredient percent estimation
- improved the definition of all the variations of "huile et stéarine végétales non hydrogénées (colza, palme)" to have better coverage
- added support for percentages like "huiles végétales 54% (colza, palme)"

Work in progress, some tests will need to be updated.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -174,6 +174,10 @@ my $separators_except_comma = qr/(;|:|$middle_dot|\[|\{|\(|\N{U+FF08}|( $dashes 
 
 my $separators = qr/($stops\s|$commas|$separators_except_comma)/i;
 
+# Symbols to indicate labels like organic, fairtrade etc.
+my @symbols = ('\*\*\*', '\*\*', '\*', '°°°', '°°', '°', '\(1\)', '\(2\)', '¹', '²');
+my $symbols_regexp = join('|', @symbols);
+
 # do not add sub ( ) in the regexps below as it would change which parts gets matched in $1, $2 etc. in other regexps that use those regexps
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -4548,7 +4571,20 @@ sub normalize_fr_a_de_b ($a, $b) {
...
+# This function removes labels like "organic" from ingredients, so that we can check if they exist
+# with canonicalize_taxonomy_tag. The labels can be parsed out when doing ingredients analysis.
+
+sub remove_parsable_labels ($ingredients_lc, $ingredient) {
+	if ($ingredients_lc eq "en") {
+		$ingredient =~ s/(?:organic |fair trade )*//ig;
+	}
+	elsif ($ingredients_lc eq "fr") {
+		$ingredient =~ s/(?: bio| biologique| équitable|s|\s|' . $symbols_regexp . ')//ig;
+	}
```
alexgarel, line 4583: (future improvement) Maybe we could use a "used_in_ingredients" property in labels taxonomy to get them.

It's a bit of a pity we don't find this in german etc.

At least we could use taxonomy entries for organic and fair trade ?

alexgarel (approved):
I did a review but it's a bit cryptic to me in the hard parts, so I will believe the tests !

PR #11008, opened 2024-11-12
fix: allow to query tags without normalization

stephanegigandet:
This is to be able to run commands with update_all_products.pl like:

./scripts/update_all_products.pl --query categories_tags=en:some-old-entry-that-is-now-a-synonym!

the ending ! means that the value should not be normalized with the taxonomy, but instead should be searched for as specified in mongodb


teolemon (approved):
Very nice one. It would be nice to have a 2nd script that gathers all changed tags between 2 releases, and automatically runs this script on affected products

teolemon:
Just created an issue about it with pseudo code: https://github.com/openfoodfacts/openfoodfacts-server/issues/11011

baslia (approved):

PR #11012, opened 2024-11-13
feat: openApi refactor fields parameter to support multiple values with enum validation (getProductByBarCode/search)

shinjigi:
### What
This PR improves the OpenAPI specification for the `fields` parameter in the product and search endpoint.

The changes make the API specification more accurate and provide better validation and documentation for the fields parameter, which is critical for controlling response payload size and structure.

- Improves API documentation accuracy and completeness
- Makes the fields parameter behavior more predictable and better documented
- Helps API consumers understand available field options

### Implementation details
1. Created new file `parameters/product_available-fields.yaml` to house fields parameter definition
2. Updated main API spec to reference the new parameter definition
3. Removed redundant fields parameter definition from individual endpoints
4. Added comprehensive enum of available field values
5. Added validation rules and examples

### Testing
- Verified that the OpenAPI spec validates correctly
- Checked that the parameter definition matches actual API behavior
- Confirmed that examples are valid and helpful

Note: This change is purely to the API specification and does not affect the actual API implementation.


docs/api/ref/parameters/product_available-fields.yaml
```diff
@@ -0,0 +1,224 @@
...
+            - ecoscore_tags
+            - editors_tags
+            - emb_codes
+            - emb_codes_tags
+            - entry_dates_tags
+            - expiration_date
+            - food_groups
+            - food_groups_tags
+            - generic_name
+            - generic_name_it
```
stephanegigandet, line 74: all the fiields like generic_name_it, product_name_it, ingredients_text_it etc. are for Italian, but they can be requested in any language, e.g. product_name_en for English
So having a comprehensive enum is difficult (unless we list all the hundreds of different possibilities, but I'm not sure that's a good idea)

shinjigi: You are right. I replaced the definition with enum with one with string pattern. try to see if it could work.
Thanks a lot
Luigi

stephanegigandet (approved):
Thank you!

PR #11029, opened 2024-11-19
refactor: remove unused code after merging of paths

stephanegigandet:
This PR removes code does is not used anymore now that the products data and images dirs have been merged for all flavors.


alexgarel (approved):
Thanks

PR #11032, opened 2024-11-21
fix: refactor + ignore 'null' product_type sent by app

stephanegigandet:
I'm hoping this PR will fix https://github.com/openfoodfacts/smooth-app/issues/5864

If we receive product_type=null, we just ignore it.

Also refactored the code to remove some duplication.


alexgarel (approved):
great

teolemon:
As of this morning, I verified the fix on my phone for OPF

PR #11035, opened 2024-11-22
feat: NOVA 4 attribute and knowledge panel improvements

stephanegigandet:
Changes that were recommended by scientists for the U-TRANS study that will use the OFF app:

- Added the number of ultra-processing markers for NOVA 4 directly in the NOVA 4 attribute description, to make it easier to differentiate foods that are slightly ultra-processed (e.g. just one flavour) and very ultra-processed with many markers
- Added a recommendation to limit the consumption of NOVA 4 foods. Currently a bit hidden inside the NOVA knowledge panel, but we could decide to display it higher (without needing to open the NOVA panel) like we do for alcohol.

image

image


alexgarel (approved):
LGTM

john-gom:
As discussed in the Product Opener meeting is we might want to consider using the "chips" approach proposed by Ben to show simplified details of "sub" panels at the top level, e.g.

image

PR #11036, opened 2024-11-22
fix: labeler.yml for Estonia packager code

stephanegigandet:
This PR fixes a bug from https://github.com/openfoodfacts/openfoodfacts-server/pull/11017 that makes tests fail for all PRs.

```
 750 | :
-------^
 751 | - changed-files:
 752 |   - any-glob-to-any-file: 'scri ...
Error: incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line (750:1)
```


teolemon (approved):
thanks @stephanegigandet

PR #11042, opened 2024-11-23
taxonomy: add raw fruits/vegetables categories

raphael0202:
For Open Prices.

Related to https://github.com/openfoodfacts/open-prices-frontend/issues/1048


taxonomies/food/categories.txt
```diff
@@ -68225,6 +68238,31 @@ fr: Poires Conférence, Conférence
 hr: Konferencijske kruške
 nl: Conference peren
 
+< en:Pears
+en: Guyot pear
```
stephanegigandet, line 68242:
```suggestion
en: Guyot pears
```

By convention all category names are in plural, with singular as a synonym (except if the synonym is just adding a final s, which should be supported automatically)

stephanegigandet: (same for all the pears / poires below)

stephanegigandet (approved):
Thank you!

PR #11058, opened 2024-11-26
feat: Add script to remove nearly empty products with quality issues

CharlesNepote:


scripts/remove_nearly_empty_products.pl
```diff
@@ -0,0 +1,77 @@
...
+
+use ProductOpener::Config qw/:all/;
+use ProductOpener::Products qw/retrieve_product/;
+use ProductOpener::Store qw/:all/;
+use ProductOpener::Data qw/get_products_collection/;
+
+# This script is run daily to remove empty products (without data or pictures)
+# in particular products created by the button to add a product without a barcode
+
+my $cursor = get_products_collection()->query({data_quality_errors_tags => { '$ne' => ''} })->fields({code => 1});
```
stephanegigandet, line 35: This query is quite broad (any product with data quality errors), it will fetch ~ 150k products, that you will then read individually. You could change your query to put some of the filters in.
e.g. adding state_tags  en:photos-to-be-uploaded would divide the number of results by 10.

teolemon (commented):
What's the difference with : https://github.com/openfoodfacts/openfoodfacts-server/blob/main/scripts/remove_empty_products.pl ?

scripts/remove_nearly_empty_products.pl
```diff
@@ -0,0 +1,77 @@
...
+
+	no warnings 'uninitialized';
+
+	if ((defined $product_ref) and ($code ne '')) {
+	
+		if (1
+			#and (defined $product_ref->{last_image_t})
+			and ($product_ref->{last_image_t} eq '')
+			and ($product_ref->{owner} eq '')
+			and ($product_ref->{creator} ne 'usda-ndb-import')
```
stephanegigandet, line 55: If you put fields like creator, completeness etc. in the retrieved fields of the mongodb query, then you can test their values before you call retrieve_product($code), which saves disk reads

scripts/remove_nearly_empty_products.pl
```diff
@@ -0,0 +1,77 @@
...
+			and ($product_ref->{creator} !~ /^org-.*/)
+			and ($product_ref->{completeness} < 0.4)
+			and (time() > $product_ref->{last_modified_t} + (60 * 60 * 24 * 30))
+			) {
+
+			print "updating product $code, $product_ref->{creator}, $product_ref->{data_quality_errors_tags}, $product_ref->{completeness}...\n";
+			#$product_ref->{deleted} = 'on';
```
stephanegigandet, line 62: it's missing the actual save of the product, undeleting this line will not delete the product yet

scripts/remove_nearly_empty_products.pl
CharlesNepote: I have tried `last_image_t => { '$eq' => '' }` but it failed to load (time out I guess).

I haven't so much trust in `state_tags => 'en:photos-to-be-uploaded'` because sometimes (rarely) computation of state_tags seems to fail. I better trust `last_image_t` which is raw data, not computed, isn't it?

scripts/remove_nearly_empty_products.pl
CharlesNepote: Understood, but how can I be sure that it won't time out?

scripts/remove_nearly_empty_products.pl
CharlesNepote: Neither [remove_empty_products.pl](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/scripts/remove_empty_products.pl). Is this script unused? Should I open a bug? or should we delete it?

I don't know how to save a product, is the following code ok?
```
my $comment = "[remove_nearly_empty_products.pl] automatic removal of product with a data quality issue, few information & without images at all (excluding imports)";
store_product($product_ref, $comment );
```

stephanegigandet: to delete it, you need both $product_ref->{deleted} = 'on'; + store_product()

CharlesNepote: Ok, done.

scripts/remove_nearly_empty_products.pl
CharlesNepote: I added `state_tags => 'en:photos-to-be-uploaded'`.

CharlesNepote:
@teolemon 
> What's the difference with : https://github.com/openfoodfacts/openfoodfacts-server/blob/main/scripts/remove_empty_products.pl ?

```
# This script is run to remove products with a data quality issue, 
# few information & no image at all (excluding imports)
```
See: https://openfoodfacts.slack.com/archives/C03H290LF/p1732539345020489

stephanegigandet (approved):

PR #11065, opened 2024-11-28
fix: pro platform product writes to the public platform MongoDB database

### LINKED ISSUE #11064: Product changes on the pro platform incorrectly written to the public platform MongoDB database
Bug reported by @manon-corneille on Slack: https://openfoodfacts.slack.com/archives/C01E2SPNQJW/p1732725011612459

There is a bug that was introduced in the recent code change for better support of product types that resulted in the pro platform writing products to the public platform MongoDB database.

We use MongoDB as a query cache and not at the source of truth, so there's no data loss, but we will need to clean MongoDB to reindex products from the pro platform.

stephanegigandet:
fixes #11064

debug notes in the issue: https://github.com/openfoodfacts/openfoodfacts-server/issues/11064#issuecomment-2506412589

### DIFF
--- a/cgi/product_jqm_multilingual.pl
+++ b/cgi/product_jqm_multilingual.pl
@@ -118,8 +118,12 @@ else {
 	else {
 		# There is an existing product
 		# If the product has a product_type and it is not the product_type of the server, redirect to the correct server
+		# unless we are on the pro platform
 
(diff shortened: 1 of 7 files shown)


alexgarel (approved):
Thanks

PR #11076, opened 2024-12-01
fix: data quality, false positive, nutrition sum with lower symbol

### LINKED ISSUE #11075: Do not take into account values less than
### What
Error is: `Nutrition-soluble-fiber-plus-insoluble-fiber-greater-than-fiber`

see: https://world.openfoodfacts.org/product/0856651002104/chili-lime-chickpeas-the-good-bean

### Expected behaviour
no error. the algorithm should ignore values with a < sign

benbenben2:
### What
Error is: Nutrition-soluble-fiber-plus-insoluble-fiber-greater-than-fiber should not be raised for 

> Fiber = 5
> insoluble fiber < 1
> soluble fiber = 5

and similar cases.

The PR covers only Fibers, insoluble fiber and soluble fiber.
The PR covers only if symbol is "<"

- Fixes #11075

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -1319,30 +1319,30 @@ sub check_nutrition_data ($product_ref) {
 		}
 
 		# sum of nutriments that compose fiber can not be greater than the value of fiber
+		# ignore if there is "<" symbol (example: <1 + 5 = 5, issue #11075)
 		if (defined $product_ref->{nutriments}{fiber_100g}) {
-			my $soluble_fiber
-				= defined $product_ref->{nutriments}{'soluble-fiber_100g'}
-				? $product_ref->{nutriments}{'soluble-fiber_100g'}
-				: 0;
-			my $insoluble_fiber
-				= defined $product_ref->{nutriments}{'insoluble-fiber_100g'}
-				? $product_ref->{nutriments}{'insoluble-fiber_100g'}
-				: 0;
+			my $soluble_fiber = 0;
+			my $insoluble_fiber = 0;
+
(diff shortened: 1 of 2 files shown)


aleene (commented):
Great

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -1319,25 +1319,31 @@ sub check_nutrition_data ($product_ref) {
...
+				? $product_ref->{nutriments}{'soluble-fiber_100g'}
+				: 0;
+			my $insoluble_fiber
+				= defined $product_ref->{nutriments}{'insoluble-fiber_100g'}
+				? $product_ref->{nutriments}{'insoluble-fiber_100g'}
+				: 0;
+			my $total_fiber = $soluble_fiber + $insoluble_fiber;
+
+			if ($total_fiber > $product_ref->{nutriments}{fiber_100g} + 0.001) {
+				# ignore if there is "<" symbol (example: <1 + 5 = 5, issue #11075)
```
stephanegigandet, line 1334: I think it would be better to add $soluble_fiber and $insoluble_fiber to $total_fiber only if there isn't a corresponding < modifier.

That way we will trigger the warning if for instance we have 5 g fiber, 10 g insoluble fiber, and < 1g soluble fiber).

stephanegigandet (approved):
Thanks @benbenben2 !
Approving the PR, but I suggested an improvement.

lib/ProductOpener/DataQualityFood.pm
benbenben2: Good idea. I applied these changes. If you could review it again, just to be sure. Thanks

benbenben2: @stephanegigandet, forgot to push 1 file

PR #11086, opened 2024-12-03
fix: display of usage in scripts/import_csv_file.pl

dolmen:
### What

Fix typo that blocked display of usage message in case of incorrect options.

(Hackathon des Données Alimentaires)


stephanegigandet (approved):
Thanks @dolmen !

Issue #9593, opened 2023-12-24
bug: e124 is synonym of e120


benbenben2:
### What
somehow, e124 is synonym of e120, but it should not but, when ingredient is e124, it should be detected as e124 not e120.

### Steps to reproduce the behavior
Look at ingredient parsing for this product
https://world.openfoodfacts.org/product/7622201798260/cadbury-gems

### Expected behavior
E124 -> E124 (not E120)

### additional information
There is a problem in the build of the taxonomy, it sets e124 as a synonym of e120

https://world.openfoodfacts.org/ingredient/e120 -> ok
https://world.openfoodfacts.org/ingredient/e124 -> no ok

I tried following changes in additives taxonomy and rebuild.
rm all 120 -> fixed
leave only en line -> bug
put only "e120" for en and keep other lang -> fixed!
rm "Cochineal Red" (120) only from en -> fixed
rm ""cochineal red a" only (124) -> fixed
looked in the logs when we run make build_taxonomy -> nothing
remove "a" stopword from FR -> nothing

It seems that for some reasons COCHINEAL RED and COCHINEAL RED A become synonyms
`en:E120, Cochineal, carminic acid, carmines, Natural Red 4, COCHINEAL RED`
`en:E124, Ponceau 4r, COCHINEAL RED A, CI Food Red 7, Brilliant Scarlet 4R, Ponceau`

alexgarel:
@benbenben2 it could be a problem when we do synonym expansion [second phase of build_tags_taxonomy in Tags.pm](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Tags.pm#L1504)

benbenben2:
Yes, I thought so but when I comment it out, and run: 
`make lint && make checks && make build_taxonomies && make rebuild_taxonomies && make build_lang && make build_lang_test && docker-compose restart`

It still remains the same. 

Do I miss a make command to recreate the tags?

alexgarel:
@benbenben2 you missed nothing !

benbenben2:
We can see in **ingredients.results.txt**

> en:E120, Cochineal, carminic acid, carmines, Natural Red 4, Cochineal Red, E124, Ponceau 4r, cochineal red a, CI Food Red 7, Brilliant Scarlet 4R, Ponceau

and **in additives.txt**
> en:E120, Cochineal, carminic acid, carmines, Natural Red 4, Cochineal Red
> [...]
[...]

Issue #11014, opened 2024-11-14
E124 seems to be considered as not vegetarian


CharlesNepote:
In [this product](https://world.openfoodfacts.org/product/8901725012335/dream-cream-20-sunfest), E124 seems to be considered as non vegetarian:
![Image](https://github.com/user-attachments/assets/4602d360-608f-4d89-bbb2-203b32c49e66)

Some websites tend to consider it as vegetarian:
* https://www.les-additifs-alimentaires.com/E124-ponceau-4r.php (French)
* https://www.passeportsante.net/nutrition/additifs-alimentaires?doc=e124-ponceau-4r-faut-savoir-sur-colorant (French)

But E120 seems to be not vegetarian: is there a difference between the both that could explain one is vegetarian and the other is not?

It is triggering a data quality error: `Vegetarian label but non vegetarian ingredient`.

PR #11095, opened 2024-12-06
fix: letter A at end of string is not a stopword

stephanegigandet:
I think this may be the reason for https://github.com/openfoodfacts/openfoodfacts-server/issues/11014

E120 and E124 have synonyms "Cochineal Red" and "Cochineal Red A".
In the ingredients taxonomy, "a" is a stopword.
E124 becomes a synonym of E120.

Solution: we don't consider A / a to be a stopword when it is at the end of a string, as in English, "a" comes before a noun.


benbenben2:
fixes #9593

benbenben2 (approved):
Great that you solved this :1st_place_medal:

PR #11096, opened 2024-12-06
fix: new images path

stephanegigandet:
new images symlinks were mistakenly created in /srv/off/products/new_images/

I will move them to /srv/off/new_images/ so that they get processed


alexgarel (approved):
LGTM

PR #11098, opened 2024-12-06
fix: data quality, false positive, nutrition sum with lower symbol for milk below the table

### LINKED ISSUE #10715: Nutrition - Fructose plus glucose plus maltose plus lactose plus sucrose greater than sugars
### What
When a limit is given to the amount of lactose, it is taken into account for the total carbohydrates, thus inducing an error.

See https://world.openfoodfacts.org/product/2302691103212/ekte-hvit-geitost-snøfrisk

### Expected behaviour
Anything with < should not be used in the calculation

### Why
Gives false errors

benbenben2:
### What
error should not be raised when 
sugar = 0
lactose < 0.01

In the product given in the issue, lactose is given below the table. Similar to #11076, lactose with symbol "<" will be ignore for the error.

+ fix some non declared variable warnings like 

> dataqualityfood.t: Use of uninitialized value in string eq at /opt/product-opener/lib/ProductOpener/DataQualityFood.pm line 1040.

- Fixes #10715

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -1037,7 +1037,7 @@ sub check_nutrition_data ($product_ref) {
 			if ((not defined $product_ref->{serving_size}) or ($product_ref->{serving_size} eq '')) {
 				push @{$product_ref->{data_quality_errors_tags}}, "en:nutrition-data-per-serving-missing-serving-size";
 			}
-			elsif ($product_ref->{serving_quantity} eq "0") {
+			elsif (defined $product_ref->{serving_quantity} and $product_ref->{serving_quantity} eq "0") {
 				push @{$product_ref->{data_quality_errors_tags}}, "en:nutrition-data-per-serving-serving-quantity-is-0";
 			}
 		}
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -1280,33 +1280,33 @@ sub check_nutrition_data ($product_ref) {
...
+			my $lactose = 0;
+			my $sucrose
+				= defined $product_ref->{nutriments}{sucrose_100g} ? $product_ref->{nutriments}{sucrose_100g} : 0;
+
+			if (defined $product_ref->{nutriments}{lactose_100g}) {
+				my $lactose_modifier = $product_ref->{nutriments}{'lactose_modifier'};
+				if (!defined $lactose_modifier || $lactose_modifier ne '<') {
+					$lactose = $product_ref->{nutriments}{lactose_100g};
+				}
+			}
```
alexgarel, line 1302: Just a small move of a line for readability

```suggestion
			my $sucrose
				= defined $product_ref->{nutriments}{sucrose_100g} ? $product_ref->{nutriments}{sucrose_100g} : 0;

			# ignore lactose when having "<" symbol
			my $lactose = 0;
			if (defined $product_ref->{nutriments}{lactose_100g}) {
				my $lactose_modifier = $product_ref->{nutriments}{'lactose_modifier'};
				if (!defined $lactose_modifier || $lactose_modifier ne '<') {
					$lactose = $product_ref->{nutriments}{lactose_100g};
				}
			}
```

alexgarel (approved):
Thanks a lot @benbenben2

PR #11100, opened 2024-12-09
docs: some more pre-requisite for docker desktop

alexgarel:
After de


dolmen (approved):
Approving: the bits about host networking is something that I discovered on my setup on MacOS at the _Hackathon des Données Alimentaires_.

PR #11108, opened 2024-12-10
taxonomy: taking into account the ingredient origin when retrieving the corresponding Ecobalyse code

very-smartin:
### What

Goal : being able to take into account the origin when retrieving the corresponding ecobalyse code for a given ingredient


lib/ProductOpener/Ingredients.pm
```diff
@@ -3512,6 +3543,35 @@ sub get_missing_ecobalyse_ids ($ingredients_ref) {
...
+
+reference to the name of the country 
+
+=head3 Return values
+
+=head4 $ecobalyse_area
+
+=cut
+
+sub get_geographical_area ($country_code) {
```
stephanegigandet, line 3562: We use country code to refer to the 2 letters country codes, could you rename $country_code to $originid ? in the origins field you will get things like "en:france" or "en:britanny".

lib/ProductOpener/Ingredients.pm
```diff
@@ -3472,11 +3472,42 @@ sub get_missing_ecobalyse_ids ($ingredients_ref) {
 
 		# List of suffixes
 		my @suffixes = ();
+		# If the ingredient is both organic and French...
+		if ((defined $ingredient_ref->{labels}) and 
+		    ($ingredient_ref->{labels} =~ /\ben:organic\b/) and 
+		    (defined $ingredient_ref->{origins}) and 
+		    (get_geographical_area($ingredient_ref->{origins}) eq "fr")) {
+
+		    push @suffixes, "_labels_en_organic_origins_en_france";
```
stephanegigandet, line 3481: If get_geographical_area() returns "fr", we need to push both suffixes: the "origins_en_france" one + the "origins_en_european_union_one" as we may not have a FR property in Ecobalyse, but we could have a EU one.

very-smartin:
@stephanegigandet i've got some issues on this code and i don't know how to solve them... What should i do ?

stephanegigandet (approved):
Perfect, thank you!

PR #11109, opened 2024-12-10
build(deps): use jquery v2.2.4 to unblock jquery-ui v1.14.1 upgrade

jayaddison:
### What

The `jquery-ui` upgrade suggested in #10970 is incompatible with `jquery` v2.2.0 up-to-not-including v2.2.4 due to a fix that was applied for https://github.com/jquery/jquery/issues/2466

In our `package-lock.json` file, it appears that we use `jquery` v3.7.1 -- but in fact we copy a vendored copy from the [`foundation` UI framework](https://get.foundation/), during the [`gulpfile.ts` `copyJs`](https://github.com/openfoodfacts/openfoodfacts-server/blob/59bb2042947a1ce12e7b1ed4ddb4e4072a742fd1/gulpfile.ts#L63) function.

I believe that `jquery` v2.2.4 is compatible -- and have tested this locally.  `TODO` I want to add a note about why exactly that version is compatible again.

- Unblocks #10970.


gulpfile.ts
```diff
@@ -104,7 +107,6 @@ function buildjQueryUi() {
     "./node_modules/jquery-ui/ui/position.js",
     "./node_modules/jquery-ui/ui/keycode.js",
     "./node_modules/jquery-ui/ui/unique-id.js",
-    "./node_modules/jquery-ui/ui/safe-active-element.js",
```
jayaddison, line 107: NB: `safe-active-element.js` was removed from `jquery-ui` in https://github.com/jquery/jquery-ui/commit/bb49bd794bc8ea4238162725b518fb46234f3cf9 (released in v1.14.0).

jayaddison:
From local testing: `jquery` v3.7.1 does appear to work OK too.  Let's upgrade to that instead.

jayaddison:
Note: I have only manually tested the nutrition autosuggest functionality so far (it's the main affected component I'm aware of and familiar with), but I think this pull request is ready for review.

jayaddison:
I'm encountering a JavaScript error when attempting to close the preferences-edit panel using a build from fd7fb7fc840ffe94b365a7d17cc4668d749b4d32.

It seems likely that the `foundation` code is not forwards-compatible with such a significant `jquery` version upgrade.

image

alexgarel:
@jayaddison thanks for working on this !

stephanegigandet:
Thank you so much @jayaddison ! I'll do my best to test the PR locally.

stephanegigandet:
What's the best way to have the gulpfile.ts changes applied in the dev containers? (ideally without running "make dev")

I tried "make front_npm_update", "make front_build" followed by "make restart" but I still have jquery ui 1.13.3

I also tried docker compose up -d --build frontend
but I'm getting an error:

```
 > [openfoodfacts-server/frontend:dev builder 13/13] RUN npm run build:
#0 0.409 
#0 0.409 > product-opener@1.0.0 build
#0 0.409 > gulp
#0 0.409 
#0 1.030 [11:36:08] Requiring external module ts-node/register
#0 2.230 TypeError: Unknown file extension ".ts" for /opt/product-opener/gulpfile.ts
```

jayaddison:
> What's the best way to have the gulpfile.ts changes applied in the dev containers? (ideally without running "make dev")

There may be a better way, but so far I have been running `make down`, `make clean` and then subsequently `SKIP_SAMPLE_IMAGES=1 make dev` (I noticed the env var in the `Makefile` as a way to save some time on each iteration) to bring up a fresh environment.  This does also require recreating a user account each time.

stephanegigandet:
> > What's the best way to have the gulpfile.ts changes applied in the dev containers? (ideally without running "make dev")
> 
> There may be a better way, but so far I have been running `make down`, `make clean` and then subsequently `SKIP_SAMPLE_IMAGES=1 make dev` (I noticed the env var in the `Makefile` as a way to save some time on each iteration) to bring up a fresh environment. This does also require recreating a user account each time.

Thanks @jayaddison . I did that, but after I'm missing jquery, the file http://static.openfoodfacts.localhost/js/dist/jquery.js does not exist for some reason on my dev environment.

gulpfile.ts
```diff
@@ -79,7 +79,10 @@ export function copyJs() {
     "./node_modules/jsvectormap/dist/maps/world-merc.js",
     "./node_modules/select2/dist/js/select2.min.js",
     "./node_modules/jsbarcode/dist/JsBarcode.all.min.js",
-  ]).
+  ], {
```
stephanegigandet, line 82: We need to add ""./node_modules/jquery/dist/jquery.js"" in the list above, otherwise it's not copied and product opener does not find jquery

jayaddison: Ah, sorry!  Yes, I had that problem once or twice during local testing too.  I committed a fix but must have dropped it during a rebase.  I'll push that fix in a few moments.

jayaddison: Ok, d606852fc30e2f3083ced004b0c7b265f5ecd257 applies that change.

jayaddison: Note: in a [previous comment](https://github.com/openfoodfacts/openfoodfacts-server/pull/11109#issuecomment-2532716651) I reported test results from `jquery` v3.7.1 -- a much more recent and preferable version if we can use it.  But the commit ID I mentioned (fd7fb7fc840ffe94b365a7d17cc4668d749b4d32) also lacks this `gulpfile.ts` copy-list fix.

jayaddison: (since yesterday, I've reconfirmed with the fix in place that indeed, v2.2.4 is OK - but v3.7.1 is not)

hangy (approved):

jayaddison:
Happy New Year - and also, my apologies for the force pushes after PR approval.  I wouldn't usually do that, but because I included the contents of branch #10970 in this one originally (so that they would both get merged together), I've been rebasing this branch to follow the rebases on that one (otherwise I worry that the commit history could get confused).

Anyway: I'll re-test this again later today (the behaviour shouldn't have changed since last time, but there's no harm in re-checking).

jayaddison:
> Anyway: I'll re-test this again later today (the behaviour shouldn't have changed since last time, but there's no harm in re-checking).

Yep, all still seems OK to me (testing from 345d0a5bb03f589851642891a0d08dd4ddd69977).  Confirmed `jquery` v2.2.4  and `jquery-ui` v1.14.1 were in use, and that nutrient autosuggest and user preferences worked as expected.

I was not able to edit/test folksonomy properties on a product due to a CORS issue, but I think that is separate (I'll attempt to replicate and then file a bug for that soon).

jayaddison:
> I was not able to edit/test folksonomy properties on a product due to a CORS issue, but I think that is separate (I'll attempt to replicate and then file a bug for that soon).

The main problem seems to be that I did not have the `openfoodfacts-nodejs` service running; the fact that a CORS error occurred when attempting to access it is, I think, a distraction from that.

alexgarel:
@jayaddison still working on it ! Kudos for your patience and your tenacy :orange_heart:  That's really appreciated.
What is the status on this PR according to you ? Ready to merge ?

jayaddison:
Thanks @alexgarel!  Personally, I think it's OK to merge.  However, I am quite risk averse and so I'll attempt to provide some reasons _not_ to:

  * I haven't re-tested the latest commit yet -- if it's preferable, we can wait until me and/or others find time to do that.
  * It arguably makes the Foundation framework in use non-pristine (some contents modified).  If that is a concern to OFF for any reason (perhaps it could affect commercial support, for example) then I'd be OK exploring alternatives.

On a more positive note: if it is merged (with testing or otherwise) I'll make sure to be available at release-time to help with any problems encountered and/or revert if necessary.

alexgarel:
@jayaddison I will try to find the time to test it.

No problem for foundation, we have no maintenance or so… :-)

jayaddison:
@alexgarel thank you, I'd be very glad for independent confirmation.

I did find some time this evening to test it from a966702965bd8e46a94a8632dd3dbefa66ab676f (latest) - from there, nutrient autosuggest works successfully, and also the preferences panel (both in OpenFoodFacts).

I've _not_ been able to test the Folksonomy feature (that also relies on `jquery` + `jquery-ui`) on any version so far yet -- so I have no additional results to report for that.

jayaddison:
PS: I did notice one unexpected console error that appeared in the browser during retesting; after initial sign-in with my test user account, the following message appeared:

```
Uncaught TypeError: checkbox is null
    checkboxChange user.pl:672
    <anonymous> user.pl:686
    jQuery 8
user.pl:672:6
```

jayaddison:
I didn't intend to amend the dependabot commit here; I forgot that rebasing against `main` would do that.

jayaddison:
I think this is ready; sorry for making a mess of the rebase.

For maintainers: if I understand correctly, pull requests here have their commits squashed during merge -- so for this PR, I think the commit history _might_ be cleanest if dependabot #10970 would be merged first, and then subsequently this one -- but please use your preferred approach.

jayaddison:
Rebasing again, this time against #11714, in a few moments.

jayaddison:
> PS: I did notice one unexpected console error that appeared in the browser during retesting; after initial sign-in with my test user account, the following message appeared:
> 
> ```
> Uncaught TypeError: checkbox is null
>     checkboxChange user.pl:672
>     <anonymous> user.pl:686
>     jQuery 8
> user.pl:672:6
> ```

I think I've managed to locate the source of this console error; it does not relate to this branch, and may be a small fix.  I will report it as a separate bug in a few moments, and then will mark this and that comment as off-topic.

stephanegigandet (approved):
Thank you for all the work @jayaddison !

Let's merge this, and we can test it more on staging.

jayaddison:
Sounds good - thank you, @stephanegigandet!

PR #11117, opened 2024-12-12
refactor: rename ecoscore/eco-score fields to "environmental score" - ready for review

stephanegigandet:
We will soon rename the Eco-Score to a new name.

To ease the transition to a new name and be more future proof, we will rename all Eco-Score internal fields to a generic "environmental score" field.

Important notes:
- the internal attribute id "ecoscore" is not renamed, as it is stored in clients.
- for the time being, we keep the internal facet id "ecosore" and we copy environmental_score_tags to ecoscore_tags so that MongoDB and off-query can continue to work

The name and images displayed to users will still be Eco-Score for now, but it will be changed to a new name at some point.

Plan discussed in https://docs.google.com/document/d/1LL8-aiSF482xaJ1o0AKmhXB5QWfVE0_jzvYakq3VXys/edit?tab=t.0#heading=h.pirt3xscsuwq


external-data/environmental_score/data/Eco_score_Calculateur.csv.11
alexgarel, line 1: Name still contains Eco_score

po/common/en.po
```diff
@@ -4827,6 +4827,19 @@ msgctxt "attribute_group_environment_name"
...
+msgstr "Low environmental impact (Eco-Score)"
+
+msgctxt "attribute_environmental_score_setting_note"
+msgid "The Eco-Score is an environmental score from A to E which makes it easy to compare the impact of food products on the environment."
+msgstr "The Eco-Score is an environmental score from A to E which makes it easy to compare the impact of food products on the environment."
+
+# Note: the Eco-Score is renamed to Environmental Score, but we keep the ecoscore identifier as it is stored in clients
 msgctxt "attribute_ecoscore_name"
 msgid "Eco-Score"
 msgstr "Eco-Score"
```
alexgarel, line 4845: Is this ok to change that, if the attribute name does not change ?

stephanegigandet: @alexgarel I add the new names, but I kept the old name as well, hopefully we can change the name later.

stephanegigandet:
Deployed through https://github.com/openfoodfacts/openfoodfacts-server/pull/11142 closing

PR #11134, opened 2024-12-16
fix: switch main nutriscore version to 2023

stephanegigandet:
This switches the nutriscore version used for the nutriscore facet to be 2023.
Nutri-Score will need to be recomputed once this is deployed.


alexgarel (approved):
Thanks.

PR #11136, opened 2024-12-16
feat: add subnutriment in nutrition table if missing and subsub is provided

### LINKED ISSUE #8241: Add allulose to nutrients?
Here is a product displaying [allulose](https://en.wikipedia.org/wiki/Psicose) (aka psicose) in its nutrients: https://world.openfoodfacts.org/product/0850017468009/chocolate-vanilla-keto-sandwich-cookies

Allulose should not be counted as a normal carbodydrate or sugar or polyol as it is low calories (like erythritol), see #7789.
As such, it should be excluded from energy computation.

As of 2023-03-23, at least [97 products contain allulose or psicose](https://mirabelle.openfoodfacts.org/products?sql=select+rowid%2C+code%2C+url%2C+creator%2C+created_datetime%2C+last_modified_datetime%2C+last_modified_by%2C+product_name%2C+ingredients_text%0D%0Afrom+%5Ball%5D%0D%0Awhere+ingredients_text+REGEXP+%22allulose%7Cpsicose%22%0D%0Alimit+500).

### LINKED ISSUE #9187: Add isomalt to nutrients?
Here at least one example of a producer using `isomalt` as a nutrient: https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=4002590140926
image

Isomalt is a polyol. Should we add it while polyols already exists?

### LINKED ISSUE #10759: Maltitol is a polyol, and should be treated as such
In the following product, the nutrient facts mention Maltitol 83g.
https://world.openfoodfacts.org/product/9324956000763/sugar-free-cola-bottles-double-d

Maltitol is a polyol: https://fr.wikipedia.org/wiki/Maltitol

When data quality errors are computed, the given product triggers an error because it does not recognize Maltitol as polyol.

benbenben2:
### What
- Added some nutrients (isomalt, allulos/psicose, maltitol, sorbitol)
- Added sub-nutrients when sub-sub-nutrient is provided only. Example: when **erythritol** is provided only, the direct parent (**polyol**) is now shown as well between **erythitol** and **carbohydrates**

### Screenshot
PRODUCT BEFORE
nutrition_product_before

PRODUCT AFTER
nutrition_product_after

EDIT _BEFORE_ (EDIT **AFTER**)
nutrition_edit_after

EDIT _AFTER_ (EDIT: **BEFORE**)
nutrition_edit_before

- Fixes #10759 
- Fixes #8241 
- Fixes #9187


benbenben2:
corresponding logs: https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/12772999683/job/35603858977

jayaddison:
@benbenben2 a small detail: are the screenshots for the nutrient-edit table (images 3 and 4) listed in the reverse order?

lib/ProductOpener/Display.pm
```diff
@@ -10272,6 +10212,285 @@ CSS
...
+	}
+
+	# Special handling for nutrition scores (optional, based on request_ref)
+	if ($nid =~ /^nutrition-score-(.*)$/) {
+		# Always show FR score and Nutri-Score
+		if (($request_ref->{cc} ne $1) and (not($1 eq 'fr'))) {
+			return 0;
+		}
+
+		# Experimental: hide nutrition scores for now
```
jayaddison, line 10274: Can we add a hyperlink to help decide if+when these should be displayed in future?

benbenben2:
> ERROR: process "/bin/sh -c set -x &&     ( ( [ ! -e /var/cache/apt/pkgcache.bin ] || [ $(($(date +%s) - $(stat --format=%Y /var/cache/apt/pkgcache.bin))) -gt 3600 ] ) &&       apt update || true     ) &&     cpanm --notest --quiet --skip-satisfied --local-lib /tmp/local/ \"Apache::Bootstrap\" &&     cpanm $CPANMOPTS --notest --quiet --skip-satisfied --local-lib /tmp/local/ --installdeps .     || ( for f in /root/.cpanm/work/*/build.log;do echo $f\"= start =============\";cat $f; echo $f\"= end =============\"; done; false )" did not complete successfully: exit code: 1


> 2025-04-05T16:11:33.0175583Z 171.3 You have CLDR::Number (0.19)
> 2025-04-05T16:11:33.0175838Z 171.3 -> FAIL Installing the dependencies failed: Module 'GS1::SyntaxEngine::FFI' is not installed
> 2025-04-05T16:11:33.0175952Z 171.3 -> FAIL Bailing out the installation for ..

benbenben2:
> @benbenben2 a small detail: are the screenshots for the nutrient-edit table (images 3 and 4) listed in the reverse order?

Yes, correct. I just edited the comment. Thanks @jayaddison

jayaddison:
Thanks @benbenben2!  Hm; updating the labels does help - even so I reckon it might be easier to understand in `BABA` (before, after, ...) order, instead of the current `BAAB`.

lib/ProductOpener/Display.pm
```diff
@@ -10272,6 +10212,285 @@ CSS
...
+sub get_nutrient_level {
+	my ($nutriment) = @_;
+	if ($nutriment =~ /^!?--/) {
+		return 2;    # Sub-sub-nutrient (level 2)
+	}
+	elsif ($nutriment =~ /^!?-/) {
+		return 1;    # Sub-nutrient (level 1)
+	}
+	return 0;    # Top-level nutrient (level 0)
+}
```
jayaddison, line 10311: I am +1 for the refactorings in this branch; however I wonder whether it might make sense to apply those as a separate PR before the fix itself (the `add_parent_if_missing` step).

If I can understand it already, then perhaps someone else reading this later could too -- so maybe it would be unnecessary additional effort.

The other annoying detail that would add even more work, though, would be if I pointed out that `get_nutrient_level` and `should_show_nutrient` now seem to be perfect candidates for unit test coverage (small, self-contained logic, where breakage could be problematic).  Could we add some test coverage to them?

benbenben2:
Many thanks @stephanegigandet

cgi/product_multilingual.pl
```diff
@@ -805,6 +805,154 @@ ($product_ref, $field, $language, $request_ref)
...
+=head3 Return values
+
+Returns an integer indicating the nutrient's level.
+- 0 for main nutrients
+- 1 for sub-nutrients
+- 2 for sub-sub-nutrients, etc.
+
+=cut
+
+sub get_nutrient_level {
```
stephanegigandet, line 829: We use signatures to pass function parameters:

```suggestion
sub get_nutrient_level($nutriment) {
```

And then you remove the line my ($nutriment) = @_;

cgi/product_multilingual.pl
```diff
@@ -805,6 +805,154 @@ ($product_ref, $field, $language, $request_ref)
...
+The language code, used when displaying taxonomy tags to get the correct localized name of the nutrient.
+
+=head3 Return values
+
+None. The function modifies the nutrients array in-place by adding a parent nutrient if it's missing.
+
+=cut
+
+sub add_parent_if_missing {
+	my ($nutriment, $nutriment_table, $nutriments, $class, $prefix, $lc) = @_;
```
stephanegigandet, line 879: Use signatures to pass parameters insted of @_

stephanegigandet: Also use $target_lc instead of $lc, as $lc is also a global variable

lib/ProductOpener/Display.pm
```diff
@@ -10332,6 +10272,285 @@ CSS
 	return $template_data_ref;
 }
 
+=head2 should_show_nutrient ( $nutriment, $product_ref, $request_ref )
```
stephanegigandet, line 10275: Good idea to make it a function

lib/ProductOpener/Display.pm
```diff
@@ -10332,6 +10272,285 @@ CSS
...
+=head3 Return values
+
+Returns an integer indicating the nutrient's level.
+- 0 for main nutrients
+- 1 for sub-nutrients
+- 2 for sub-sub-nutrients, etc.
+
+=cut
+
+sub get_nutrient_level {
```
stephanegigandet, line 10362: If you have get_nutrient_level in Display.pm, then you don't need it defined as well in product_multilingual.pl

lib/ProductOpener/Display.pm
```diff
@@ -10332,6 +10272,285 @@ CSS
...
+
+=head4 Template data reference $template_data_ref
+
+A reference to the data structure that holds the nutrition table rows.
+
+=head4 Added nutrients reference $added_nutrients_ref
+
+A hash reference tracking nutrients that have already been added to avoid duplication.
+
+=head4 Nutriments array reference $nutriments
```
stephanegigandet, line 10397: By convention, we use a _ref prefix for all variables that are references, so that would be $nutriments_ref or $nutriments_list_ref instead of $nutriments

lib/ProductOpener/Display.pm
```diff
@@ -10332,6 +10272,285 @@ CSS
...
+
+=cut
+
+sub add_parent_if_missing {
+	my ($nutriment, $level, $template_data_ref, $nutriments, $product_ref) = @_;
+
+	# top nutrients don't have parents
+	return if $level == 0;
+
+	# Locate the parent nutrient by checking the previous nutrients in @nutriments
```
stephanegigandet, line 10413: It feels very cumbersome to do this at runtime, going backward to get the parent etc.

A simpler approach would be to create a hash of all nutrients at initialization time, that would store the level and parent of all nutrients. That way at runtime we don't need to recompute it, and the code will be much shorter.

Search for "# Compute the list of nutriments that are not shown by default so that they can be suggested" in Food.pm, we already do something similar.

tests/integration/expected_test_results/api_v2_product_read/get-fields-attribute-groups-all-knowledge-panels.json
```diff
@@ -3185,6 +3185,21 @@
                               }
                            ]
                         },
+                        {
+                           "values" : [
+                              {
+                                 "level" : 0,
+                                 "style" : "max-width:15rem",
+                                 "text" : "Alcohol"
```
stephanegigandet, line 3193: There are many tests that have changed values that are probably not expected. Like the alcohol being added.

PR #11140, opened 2024-12-17
fix: nutriscore messages

stephanegigandet:
suggestions from Santé publique France


raphael0202 (approved):

alexgarel (approved):
Thanks.

teolemon:
I don't like the new wordings

po/common/common.pot
```diff
@@ -4613,13 +4613,15 @@ msgctxt "attribute_nutriscore_c_description_short"
...
-msgstr "Poor nutritional quality"
+msgid "Lower nutritional quality"
+msgstr "Lower nutritional quality"
 
+# lower: translate to "less good" in some languages (e.g. French: "Moins bonne qualité nutritionnelle")
 msgctxt "attribute_nutriscore_e_description_short"
-msgid "Bad nutritional quality"
-msgstr "Bad nutritional quality"
+msgid "Lower nutritional quality"
+msgstr "Lower nutritional quality"
```
jayaddison, line 4624: Are these strings intentionally the same?  (I'm wondering whether this introduces a risk that they will unintentionally be translated to a single output text in other languages, if so)

On a separate note, I also don't love the phrasing - depending on how the reader parses it, it could lack relativity (e.g. "it says that it is lower.. lower than what?").

I find that to be more of a problem when I put myself in my internal `en_GB` (native) mode, while when I place my mind into `en_US` mode, I find the relative-term-without-reference-point more common, yet still confusing.  ("they're an older citizen" - initially phrases such as this confused me)

PR #11142, opened 2024-12-18
feat: rename ecoscore fields to environmental_score (internal) and new external name and logos

stephanegigandet:
Continuation of https://github.com/openfoodfacts/openfoodfacts-server/pull/11117

We will soon rename the Eco-Score to a new name.

To ease the transition to a new name and be more future proof, we will rename all Eco-Score internal fields to a generic "environmental score" field.

Important notes:

- the internal attribute id "ecoscore" is not renamed, as it is stored in clients.
- for the time being, we keep the internal facet id "ecosore" and we copy environmental_score_tags to ecoscore_tags so that MongoDB and off-query can continue to work
- in order for attributes and knowledge panels to work before Eco-Score has been recomputed (and new environmental_scores fields have been created), the attributes and knowledge panels will use ecoscore_data if environmental_data is not present.

Plan discussed in https://docs.google.com/document/d/1LL8-aiSF482xaJ1o0AKmhXB5QWfVE0_jzvYakq3VXys/edit?tab=t.0#heading=h.pirt3xscsuwq


html/images/misc/nutriscore/new-nutriscore-header-en.jpg
alexgarel, line 1: why do we have this in this PR ?

alexgarel (approved):
Great

PR #11144, opened 2024-12-18
fix: promo images for new nutri-score

stephanegigandet:


teolemon (approved):
It should have been on openfoodfacts-web, but why not

PR #11161, opened 2024-12-21
fix: allow serving size to be hyphenated

### LINKED ISSUE #10961: Big: Serving size is missing digits to strict
### What
When I want to specify that a product does not mention a portion/serving size, I would like to enter a dash (`-`). Thius now results in an error.

Just enter a dash ("-") in the field and save

### Expected behavior
No error

### Why
Entering missing data is also data

### Type of device
- Browser: Safari

Many

benbenben2:
### What
allow serving size to be hyphen ("-")

- Fixes #10961


hangy (approved):

alexgarel:
Great @benbenben2 !

Issue #11013, opened 2024-11-14
ci: GitHub Actions runner out of disk space when testing `make dev`


hangy:
I noticed an interesting issue with the GitHub Action "Test make dev" in the Keycloak branch. The [issue](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/11834780863/job/32976647075) seems to be unrelated to the business logic related stuff, as the runner seems to run out of space when while downloading [images for the random products](https://static.openfoodfacts.org/exports/products.random-modulo-10000.images.tar.gz).

![Image](https://github.com/user-attachments/assets/a4729e7b-e618-456b-b5c3-6dd475b27ce1)

One reason for the issue popping up in that branch is most likely that it requires additional container images (Keycloak, Postgres 16) to be pulled, but it's not necessarily the root cause.

The downloaded `products.images.tar.gz` is around 1.16 GB compressed, 1.17 GB (>8000 files) extracted. As the GitHub workflow probably only needs to verify that `make dev` works in principle, using a simpler product export with 5-10 products instead of hundreds of products could solve that issue for a while.

PR #11162, opened 2024-12-21
ci: Delete downloaded container image

hangy:
Delete the downloaded container image from the temporary path; the GitHub action has already loaded it into docker.

This annoying workaround should help with build runs still failing with #11013.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -109,9 +109,17 @@ jobs:
         key: taxonomies-${{ hashFiles('taxonomies/**') }}
         restore-keys: taxonomies-
     - name: Download backend image from artifacts
+      id: downloadbackendimage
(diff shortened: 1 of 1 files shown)


benbenben2:
Only the job "tests_dev" in the "workflow pull_request.yml" (Pull Requests Checks) is problematic. Hence, removing image only in tests_dev job, might be enough.

benbenben2 (approved):

hangy:
> Only the job "tests_dev" in the "workflow pull_request.yml" (Pull Requests Checks) is problematic. Hence, removing image only in tests_dev job, might be enough.

ACK. I was actually wondering if I should just suggest a change to the GitHub action that we use to download the image. I imagine that a lot of people that use the action to download an image don't need the downloaded file afterwards

alexgarel:
Thanks !

PR #11163, opened 2024-12-21
ci: Speed up build by using pre-built image from previous stage for `make dev`-like check

hangy:
See https://openfoodfacts.slack.com/archives/C02LDQDDD/p1734814872519649 for discussion.

Pro: Speeds up the `make dev` check by a lot, and it use less disk space, reducing the risk of the build agent running out of space.
Con: While similar, it doesn't run `make build` as part of the `make dev_no_build` check, making it not 100% identical to the local `make dev`

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -216,7 +216,7 @@ jobs:
         echo "export USER_GID=$(id -g)" >> .envrc
     - name: Test make dev
       run: |
-        make DOCKER_LOCAL_DATA="$(pwd)" SKIP_SAMPLE_IMAGES=1 dev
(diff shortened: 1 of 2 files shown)


teolemon (commented):
- That's great 👍 
- We could have an option to require a full build on demand.
- Also, any idea of the cases where a full build is necessary, eg the potential holes it creates

hangy:
> Also, any idea of the cases where a full build is necessary, eg the potential holes it creates

Not really. The build is done in nearly the same environment, just in separate steps. It's just not completely identical to a local `make dev`.

> We could have an option to require a full build on demand.

Good idea. That way, the normal CI build could be sped up using this change, and a full `make dev` would (optionally) still be available.

teolemon (approved):
Approving @hangy

alexgarel:
Thanks @hangy  !

PR #11167, opened 2024-12-23
taxonomy: Fix category taxonomy warnings

raphael0202:
Fix taxonomy errors and warning spotted using the Taxonomy editor:

Capture d’écran du 2024-12-23 08-50-17

The only one remaining is this one: `WARNING: Entry with same id fr:terrines-foie-volaille already exists, duplicate id in file at line 116496. The two nodes will be merged, keeping the last values in case of conflicts.`

It looks like `Terrines de foie de volaille` and `Terrines au foie de volaille` are merged into the same entity due to stopwords, I'm not sure how to fix this one.

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -63905,18 +63905,6 @@ hr: Pšenični škrob
 lt: Kviečių krakmolas
 nl: Tarwezetmelen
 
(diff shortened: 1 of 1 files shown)


teolemon (approved):
Looks good, thanks @raphael0202

PR #11173, opened 2024-12-27
feat: update product_type

### LINKED ISSUE #11094: Automatically change the product type of non-food categorized products
### Problem
- Contributors + Robotoff are able to detect non-food products and categorize them
- https://world.openfoodfacts.org/category/non-food-products
- Automatically change the product type of non-food categorized products

benbenben2:
### What
Automatically change the product type of non-food categorized products

Copy pasted all non-*something* on top of each categories taxonomies.
Also copy pasted categories from food to other flavors (for example, en: Dry Dog Food is moved from taxonomies/food/categories.txt to taxonomies/petfood/categories.txt

### Remark
The following, I could not add, but would be a nice to have:
 - rm tag from previous flavor/project after migration. For example, after changing product_type from food to beauty we would like to remove
 1) 'en:non-food-products' and 'en:open-beauty-facts'
but also categories from food
 2) "en:desserts", "en:pies", "en:sweet-pies", "en:apple-pies" for example

I tried to do something with **remove_tag** function from **Tags.pm**. It removes element of the **categories_tags** array, but **categories** (string) and **categories_hierarchy** (array) remain. And tags are still visible on the website. As I did not want to do some update of the **categories** string, I decided not to include that in the present PR.

- Fixes  #11094


lib/ProductOpener/Products.pm
```diff
@@ -2955,6 +2955,83 @@ sub compute_languages ($product_ref) {
...
+				$error = change_product_type($product_ref, "beauty");
+			}
+			elsif (has_tag($product_ref, "categories", "en:open-food-facts")) {
+				$error = change_product_type($product_ref, "food");
+			}
+			elsif (has_tag($product_ref, "categories", "en:open-pet-food-facts")) {
+				$error = change_product_type($product_ref, "petfood");
+			}
+		}
+	}
```
alexgarel, line 3026: Thanks @benbenben2 that's great.

That said, I would have first computed "expected_product_type" using "has_tag" and if it's defined then, I would have compared it to $product_ref->{product_type}.

It would allow less repetitions in the code. 

Are you willing to refactor ?

taxonomies/food/categories.txt
```diff
@@ -82,6 +82,70 @@ stopwords:de: und, mit, von
...
+< en:Open Products Facts
+en: detergents
+hr: deterdženti, deterdžent
+ja: 洗剤
+nl: Detergenten
+tr: Deterjan
+wikidata:en: Q334637
+
+# REAL Cigarettes
+< en:Non food products
```
alexgarel, line 134:
```suggestion
< en:Open Products Facts
```

lib/ProductOpener/Products.pm
benbenben2: Sure. I applied some changes. Is it what you expected?

alexgarel: it was more:

```perl
	my $expected_type;
	if (has_tag($product_ref, "categories", "en:open-food-facts")) {
		$expected_type = "food";
	}
	elsif (has_tag($product_ref, "categories", "en:open-pet-food-facts")) {
		$expected_type = "petfood";
	}
	elsif (has_tag($product_ref, "categories", "en:open-products-facts")) {
		$expected_type = "product";
	}
	elsif (has_tag($product_ref, "categories", "en:open-beauty-facts")) {
		$expected_type = "beauty";
	}
	if ($expected_type and ($product_ref->{product_type} ne expected_type)) {
		$error = change_product_type($product_ref, expected_type);
	}
```

lib/ProductOpener/Products.pm
```diff
@@ -3678,6 +3721,11 @@ sub analyze_and_enrich_product_data ($product_ref, $response_ref) {
 
 	compute_languages($product_ref);    # need languages for allergens detection and cleaning ingredients
 
+	# change the product type of non-food categorized products (issue #11094)
+	if (has_tag($product_ref, "categories", "en:incorrect-product-type")) {
+		review_product_type($product_ref);
```
stephanegigandet, line 3726: There's something tricky: change_product_type() updates the product_type field and adds a old_product_type, so that store_product() can then remove the product from the old MongoDB collection.

But that means that store_product() must absolutely be called when the product_type is changed.

What you did works for changes done through the website or API, because we call store_product().   But we don't always call store_product(), there is an exception in scripts/update_all_products.pl : when we reprocess ingredients, scores etc. 

So we need to change update_all_products.pl so that we do call store_product() if we have a $product_ref->{old_product_type} value. We can use an user like "update-product-type-bot".

See https://github.com/openfoodfacts/openfoodfacts-server/blob/main/scripts/update_all_products.pl#L1405

benbenben2: I tried to do something, that might not be what you expect. Let me know

stephanegigandet (approved):
I made small changes

teolemon:
@benbenben2 
ERROR - duplicate_synonym - 37 - en:off already is a synonym of en:open-food-facts for entry en:open-food-facts (categories) - en:off cannot be mapped to entry en:open-pet-food-facts / en:open-pet-food-facts
building taxonomy for categories - publish: 1
ERROR - duplicate_synonym - 38 - xx:off already is a synonym of xx:open-food-facts for entry en:open-food-facts (categories) - xx:off cannot be mapped to entry en:open-pet-food-facts / xx:open-pet-food-facts
Errors in the categories taxonomy definition at /opt/product-opener/lib/ProductOpener/Tags.pm line 1[62](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/13267520896/job/37038639676?pr=11173#step:9:63)2.

teolemon:
merge conflicts @benbenben2

PR #11185, opened 2025-01-03
taxonomy: Add missing ingredients from mymuesli

github-throwaway:
### What
Add missing ingredients from mymuesli product


aleene (changes requested):
Better comment out the compound parent. It is not yet official.

stephanegigandet (approved):
Thank you!

aleene (approved):

PR #11188, opened 2025-01-03
feat: petfood-nutrition-facts

benbenben2:
### What
Based on EU regulations:
https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A02009R0767-20181226&qid=1734788661215

Nutrition for pet food products is somewhat different than for human food.

There are 3 types of nutriments in pet food nutrition facts
- analytical constituents -> always in % 
- nutritional additives -> in mg/kg or UI/kg (i.e. not per 100g)
- zootechnical additives -> in mg/kg

Suggested solution is to use existing nutrition tables as defined in Food.pm (see %nutriments_tables) but split into 2 off_ and opff_.
A few `if petfood ... else ...` have been introduced. Another option might have been to make it more generic and use setup in ProductsFeatures, but that would have been more complex.

There might be future minor improvements (for the future) such as:
  - vitamin for pet food is in UI and not mg (for now contributors can change)
  - iron II sulphate monohydrate and iron might be used interchangably for pet food. i.e. iron for pet food might not be same as iron for food.
  - crude-protein for pet food might or might not be identical to protein for food. Same for crude-fibre and fibre 

Since the nutriments not given in % are given in mg/kg, nutriments suffix can be _1kg and not only _100g (see screenshot of the API below). Also compute_nutrition_data_per_100g_and_per_serving is renamed compute_nutrition_data_per_xxg_and_per_serving, 

Fixes:
- fixed bug pet_food -> petfood (x2)
- fixed warning in API.pm

comment:
	There is a bit of confusion in variables naming because product_type is used for both
	- flavor product type (petfood, food, beauty, products)
	- nutrition fact quantity/measurement type like in `foreach my $product_type ("", "_prepared") {`

### Screenshot
BEFORE
Screenshot_20250103_234833

AFTER
Screenshot_20250103_234753
Screenshot_20250103_234951
Screenshot_20250103_234911

- Fixes #-none-


stephanegigandet:
@benbenben2 awesome work! Thank a lot for looking into this!

I'm a bit wary of adding the _1kg suffix. I think an alternative would be to keep the _100g internally, so that we can keep the code closer to the food code. Maybe we could discuss it in the product opener meeting, or in another call?

benbenben2:
> @benbenben2 awesome work! Thank a lot for looking into this!
> 
> I'm a bit wary of adding the _1kg suffix. I think an alternative would be to keep the _100g internally, so that we can keep the code closer to the food code. Maybe we could discuss it in the product opener meeting, or in another call?

front for product as well as for edit remains per 1 kg. API is rolled back to be per 100g. There is changes in Food and Display to switch from front to api (divide by 10) and from api to front (multiply by 10), respectively. 

function name is rolled back from xxg to 100g


Screenshot_20250106_214034
Screenshot_20250106_214051
Screenshot_20250106_214125

stephanegigandet:
@benbenben2 I made a few changes because I couldn't see the per 100g / per serving checkboxes on OFF with this branch.

I deployed it on OPFF:
https://fr.openpetfoodfacts.org/produit/3560070387687/multicroquettes-carrefour

benbenben2:
Thanks @stephanegigandet

PR #11196, opened 2025-01-06
taxonomy: adding a new feature in the API to compute the Ecobalyse score

very-smartin:
### What

A new service in the API allows the computation of the Ecobalyse score

- Fixes #[ISSUE NUMBER]


lib/ProductOpener/APIProductServices.pm
```diff
@@ -51,6 +51,7 @@ Currently implemented services:
 - extend_ingredients : extend the ingredients object with additional information
 - estimate_ingredients_percent : compute percent_min, percent_max, percent_estimate for each ingredient in the ingredients object
 - analyze_ingredients : analyze the ingredients object and return a summary object
+- estimate_environmental_cost_ingredients : estimate the environmental cost of a given product (see Ecobalyse)
```
stephanegigandet, line 54: I think the service should be estimate_environmental_impact, without speciffying ingredients: at first we will look only at ingredients, but maybe in the next iteration, we will also add the packaging data, and maybe some processing implied by the category (e.g. frozen vegetables: freezing).

cost -> I think it may be more generic to use "impact" instead of "cost"

So we could create a separate EnvironmentalImpact.pm module.

stephanegigandet:
Error from the tests:

( FAILED )  job 17    tests/unit/environmental_score_ecobalyse.t
< REASON >  job 17    Test script returned error (Err: 111)
< REASON >  job 17    No plan was declared, and no assertions were made.

stephanegigandet:
To see why the test is failing, you can open a bash container in docker, and run it manually:

www-data@81b6c881293a:/opt/product-opener$ perl tests/unit/environmental_score_ecobalyse.t 

# Response content: <h1>Software error:</h1>
# <pre>Undefined subroutine &amp;ProductOpener::EnvironmentalImpact::estimate_environmental_impact_service called at /opt/product-opener/lib/ProductOpener/APIProductServices.pm line 222.

stephanegigandet (approved):
Thank you @very-smartin !

PR #11206, opened 2025-01-07
fix: Invalid OFF dark icon

g123k:
Hi everyone!

A really tiny change, but the dark variant of the OFF logo is flipped.
We use it in dark mode in the app.


hangy (approved):

stephanegigandet (approved):
Thanks!

PR #11207, opened 2025-01-08
taxonomy: Add german and italian words for semi-skimmed milk

github-throwaway:
### What
Add german and italian words for semi-skimmed milk


stephanegigandet (approved):
Thanks

Issue #192, opened 2016-01-19
Compress all static assets


teolemon:
Compressing resources with gzip or deflate can reduce the number of bytes sent over the network.
Enable compression for the following resources to reduce their transfer size by 702.6KiB (75% reduction).
(That’s the homepage)

> Compressing http://static.openfoodfacts.org/js/jquery-ui-1.11.4/jquery-ui.min.js could save 171.7KiB (73% reduction).
> Compressing http://static.openfoodfacts.org/foundation/js/foundation.min.js could save 154.8KiB (79% reduction).
> Compressing http://static.openfoodfacts.org/foundation/css/app.css could save 139.3KiB (87% reduction).
> Compressing http://static.openfoodfacts.org/foundation/js/vendor/jquery.js could save 53.4KiB (64% reduction).
> Compressing http://world.openfoodfacts.org/ could save 37.6KiB (75% reduction).
> Compressing http://static.openfoodfacts.org/js/jquery.fileupload.js could save 28.7KiB (77% reduction).
> Compressing http://static.openfoodfacts.org/js/jquery-ui-1.11.4/jquery-ui.css could save 26.4KiB (76% reduction).
> Compressing http://staticxx.facebook.com/connect/xd_arbiter.php?version=42 could save 22.4KiB (66% reduction).
> Compressing https://staticxx.facebook.com/connect/xd_arbiter.php?version=42 could save 22.4KiB (66% reduction).
> Compressing http://static.openfoodfacts.org/…on/foundation-icons/foundation-icons.css could save 15.5KiB (81% reduction).
> Compressing http://static.openfoodfacts.org/foundation/js/vendor/modernizr.js could save 6.5KiB (58% reduction).
> Compressing http://slack.openfoodfacts.org/iframe could save 5.7KiB (57% reduction).
> Compressing http://static.openfoodfacts.org/js/jquery.iframe-transport.js could save 5.5KiB (72% reduction).
> Compressing http://slack.openfoodfacts.org/slackin.js could save 4.6KiB (68% reduction).
> Compressing http://static.openfoodfacts.org/js/jquery.fileupload-ip.js could save 4.4KiB (71% reduction).
> Compressing https://platform.twitter.com/…tton.b059c0eb61ea902a882e5c5b3c66a17a.js could save 2.8KiB (65% reduction).
> Compressing http://static.openfoodfacts.org/js/load-image.min.js could save 720B (56% reduction).
[...]

PR #11208, opened 2025-01-08
feat: gzip js, css, SVG icons assets

stephanegigandet:
As pointed by @github-throwaway in https://github.com/openfoodfacts/smooth-app/discussions/5583#discussioncomment-11742451 , we currently serve uncompressed assets (CSS, JS etc.).

We should add .gz versions so that they can be served by nginx. We need to keep the non .gz files as well as we use try_files, which returns a 404 if the uncompressed file does not exist:

```
                try_files resources/$uri $uri $uri/ =404;
                gzip_static always;
                gunzip on;
```


stephanegigandet:
Works locally, but in GitHub tests I get this error:

```
#22 [dynamicfront builder 13/13] RUN npm run build
#22 0.329 
#22 0.329 > product-opener@1.0.0 build
#22 0.329 > gulp
#22 0.329 
#22 0.885 [14:18:40] Requiring external module ts-node/register
#22 1.955 TypeError: Unknown file extension ".ts" for /opt/product-opener/gulpfile.ts
#22 1.955     at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:218:9)
#22 1.955     at defaultGetFormat (node:internal/modules/esm/get_format:[244](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/12672402831/job/35316357560?pr=11208#step:4:245):36)
#22 1.955     at defaultLoad (node:internal/modules/esm/load:122:22)
#22 1.955     at async ModuleLoader.load (node:internal/modules/esm/loader:570:7)
#22 1.955     at async ModuleLoader.moduleProvider (node:internal/modules/esm/loader:443:45)
#22 1.955     at async ModuleJob._link (node:internal/modules/esm/module_job:106:19) {
#22 1.955   code: 'ERR_UNKNOWN_FILE_EXTENSION'
#22 1.955 }
#22 ERROR: process "/bin/sh -c npm run build" did not complete successfully: exit code: 1
```

There's a huge thread about this error here: https://github.com/TypeStrong/ts-node/issues/1997

teolemon:
I was convinced we had already enabled that in the past @stephanegigandet 
- #192

stephanegigandet:
@teolemon it probably was working before because we had the nginx gzip module enabled in https://github.com/openfoodfacts/openfoodfacts-server/blob/main/conf/nginx/nginx.conf

but that file is not used anymore in production apparently, not sure if it's by design or not.

stephanegigandet:
That being said, it's better to have pre-compressed static assets I think. We could remove the try_files (I'm not sure they are really useful, it doesn't make much sense to me to try multiple locations...)

```
                # First attempt to serve request from resource, then as file,
                # then as directory, then fall back to displaying a 404.
                try_files resources/$uri $uri $uri/ =404;
                gzip_static always;
                gunzip on;
```

If we remove the try_files, then we don't need to keep the uncompressed assets, as gunzip on will work.

PR #11219, opened 2025-01-08
taxonomy: more flavorings translated to Polish

Teiron:
### What

Added more missing flavorings in Polish


stephanegigandet (approved):
Thank you!

PR #11225, opened 2025-01-10
feat: new import_images.pl script for pro platform

stephanegigandet:
The import of images in the pro platform is currently tied to the import of product data: we import images only of they come with product data.

Some producers can send us product photos in bulk, without sending product data at the same time.

The new image_import.pl script is to load those photos.  The script essentially uses existing code from Import.pm (which has been moved a bit in separate functions).

### DIFF
--- a/lib/ProductOpener/Import.pm
+++ b/lib/ProductOpener/Import.pm
@@ -328,6 +328,7 @@ sub upload_images_for_product($args_ref, $images_ref, $product_ref, $imported_pr
 
 			my $imagefield_with_lc = $imagefield;
 
+			# we might have more than one column for image_other
 			# image_other_url.2 -> remove the number
 			$imagefield_with_lc =~ s/(\.|_)(\d+)$//;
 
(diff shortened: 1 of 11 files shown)


lib/ProductOpener/Import.pm
```diff
@@ -239,6 +261,275 @@ sub import_images_from_dir ($image_dir, $stats) {
...
+			$imagefield_with_lc =~ s/(\.|_)(\d+)$//;
+
+			if ($imagefield_with_lc !~ /_\w\w/) {
+				$imagefield_with_lc .= "_" . $product_ref->{lc};
+			}
+
+			# upload the image
+			my $file = $images_ref->{$imagefield};
+
+			# Skip PDF file has we have issues to convert them, and they are sometimes not images about the product
```
teolemon, line 340:
```suggestion
			# Skip PDF file, as we have issues to convert them, and they are sometimes not images about the product
```

alexgarel (approved):
Thanks.

PR #11238, opened 2025-01-13
feat: determine packaging components in contact with food

stephanegigandet:
Some simple rules (to start) to determine which packaging components are in contact with food.

See https://docs.google.com/document/d/1GFWM7TBQwBQ2UJ_WSgmk4VYkHl0W7wmdTPHprwJlsxY/edit?tab=t.0#heading=h.i1wspmuo16w4 for discussion.

Exposed as a separate service.

Currently computed and displayed only for moderators, for testing.


lib/ProductOpener/PackagingFoodContact.pm
```diff
@@ -0,0 +1,271 @@
...
+
+Determine if packaging components are in contact with the food.
+
+=head3 Parameters
+
+=head4 $packagings_ref packaging data
+
+=cut
+
+sub determine_food_contact_of_packaging_components ($packagings_ref) {
```
alexgarel, line 206: I appreciate that you separate the _service and the action by itself !

alexgarel (approved):
Great !

PR #11240, opened 2025-01-15
chore: make more requests goes to priority apache

alexgarel:
We added the priority apache.
But initially we did add only a few path to it.
As it's working fine, make more requests goes through it as they are important for product contribution.

A [graph is visible here](https://prometheus.openfoodfacts.org/graph?g0.expr=apache_workers%7Bstate%3D%22busy%22%2Cenv%3D%22prod%22%7D&g0.tab=0&g0.stacked=0&g0.show_exemplars=0&g0.range_input=2h).


stephanegigandet (approved):
Looks good

PR #11244, opened 2025-01-15
fix: Green-Score attributes/panels titles and subtitles

stephanegigandet:
A few fixes/changes:
- we still had a "A-PLUS" instead of "A+" in a knowledge panel title
- separated title and subtitle
- for not-applicable, changed the "short description" so that it includes the non-applicable category (e.g. "sodas"), and display it directly in the attribute subtitle. (it's a bit longer, but probably worth displaying directly).

### DIFF
--- a/po/common/common.pot
+++ b/po/common/common.pot
@@ -5385,12 +5385,12 @@ msgid "User is being deleted. This may take a few minutes."
 msgstr "User is being deleted. This may take a few minutes."
 
 msgctxt "attribute_environmental_score_not_applicable_title"
-msgid "Green-Score not yet applicable"
-msgstr "Green-Score not yet applicable"
+msgid "Green-Score not applicable"
+msgstr "Green-Score not applicable"
 
 msgctxt "attribute_environmental_score_not_applicable_description_short"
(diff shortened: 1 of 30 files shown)


po/common/common.pot
```diff
@@ -5385,12 +5385,12 @@ msgid "User is being deleted. This may take a few minutes."
...
-msgid "Green-Score not yet applicable"
-msgstr "Green-Score not yet applicable"
+msgid "Green-Score not applicable"
+msgstr "Green-Score not applicable"
 
 msgctxt "attribute_environmental_score_not_applicable_description_short"
-msgid "Not yet applicable for the category"
-msgstr "Not yet applicable for the category"
+msgid "Not applicable for the category"
+msgstr "Not applicable for the category"
```
alexgarel, line 5393: Are we still using it then ?

alexgarel (approved):
Thanks !

po/common/common.pot
stephanegigandet: yes, this just changes the message for some categories like sodas and water. We might get a Green-Score for them at some point, but we don't necessarily want people to think it will come soon.

PR #11245, opened 2025-01-15
feat: serve content without header and footer for mobile app

stephanegigandet:
This PR is for better display of content pages in the mobile app. It removes the header or the footer if:
- the user agent contains "Smoothie" (can also be faked by passing the ?user_agent=smoothie)
- we have a ?content_only=1 parameter


lib/ProductOpener/Display.pm
```diff
@@ -7618,7 +7618,12 @@ sub display_page ($request_ref) {
 	$template_data_ref->{request} = $request_ref;
 
 	my $html;
-	process_template('web/common/site_layout.tt.html', $template_data_ref, \$html, $request_ref)
+	my $template = 'web/common/site_layout.tt.html';
+	# ?content_only=1 -> only the content, no header, footer, etc.
+	if (($user_agent =~ /smoothie/) or (single_param('content_only'))) {
+		$template = 'web/common/content_only.tt.html';
+	}
```
alexgarel, line 7625: nitpick: I think it would be worth isolating in a get_site_layout method

templates/web/common/content_only.tt.html
alexgarel, line 1: Can't we avoid a full rewrite of the content template ?

Isn't it possible to separate common parts in another template and include it in both ?

alexgarel (approved):
Great idea.

I approve to avoid slowing you down but I have some complains still ;-)

templates/web/common/content_only.tt.html
stephanegigandet: it's possible, but then it makes tons of small files.. I think it would be better to use [% IF ! content_only %] directives to remove what we don't want in fact.

stephanegigandet: @alexgarel I think it's better that way: https://github.com/openfoodfacts/openfoodfacts-server/pull/11245/commits/ebee79c86135f620cff593c91cb6befe5491e914

PR #11250, opened 2025-01-16
feat: more cosmetic brands for more Open Beauty Facts cleanup

teolemon:
### What
- feat: more cosmetic brands for more Open Beauty Facts cleanup
- it triggers the quality facet that allows us to move products quickly to Open Beauty Facts.
- it is complementary to Robotoff detection (sometimes it predicts the en:Open Beauty Facts category)
- ideally, the mechanism would use the brand taxonomy from Open Beauty Facts.


CharlesNepote:
What are your sources?

Isn't it a bit dangerous? Some of these brands are also food brands. Some of these brands might represent dozens of food products. What is the impact of such a list? Are there automatically tagged as OBF?

Freso:
Is this still relevant with the taxonomisation of brands?

teolemon:
- the current cosmetics détection by robotoff is not perfect
- The real game changer that might make this obsolete is the deployment of Robotoff to OBF, and we better train it to detect cosmetics 
- My source is a selection of most popular brands on Open beauty facts 
- it is still relevant as long as we don't have an efficient way to detect well and thus move quickly cosmetics out of open food facts
- We need to check brands that good have false positives, but that's just a quality facet, so we have a margin for errors

teolemon:
@Freso 
- it triggers the quality facet that allows us to move products quickly to Open Beauty Facts.
- it is complementary to Robotoff detection (sometimes it predicts the en:Open Beauty Facts category)
- ideally, the mechanism would use the brand taxonomy from Open Beauty Facts.

PR #11251, opened 2025-01-16
fix: better analysis of "oil (rapeseed, something unrecognized)" + separation of additive class + additive

stephanegigandet:
In the test set for the estimation of the % of ingredients, we have 43 products out of 1000 with the ingredients "rapeseed". It turns out that a lot of that is due to our inability to correctly parse things like "vegetable oil (rapeseed, something that we don't recognize as oil)".

We have an ingredient preparsing algorithm that tries to recognize things like [category of ingredients] ([enumeration of types of ingredients]).  e.g. the preparsing turns  "vegetable oils (palm, rapeseed, soy)" in "vegetable oils (palm vegetable oils,  rapeseed vegetable oils, soy vegetable oils)".

This only works if we can identify all the oil types in the enumeration, and that we have corresponding oils in the taxonomy. So it fails a lot.

This PR introduces an alternative to the preparsing, with a more general approach:

When an ingredient has a parent ingredient, we check if there is a known ingredient "parent + child" in the taxonomy (e.g. for "oil (palm)", we check if we have a known ingredient "palm oil". In other languages, we reverse the order "huiles (palme)" -> checks for "huile palme".

It would have to be tested, but we could potentially keep this and completely remove the equivalent function in the preparsing, which requires to hardcode all types of oils, flavours etc. and fails when we miss one in the enumeration.

This PR also removes some entries from the ingredients taxonomy, like "emulsifier soy lecithin" which is incorrect but common in ingredients list. This is because we don't want "emulsifier (soy lecithin)" to be converted to "emulsifier soy lecithin". We have a function to turn "emulsifier : soy lecithin" in the preparsing. It didn't work in this specific case because "soy lecithin" is in the ingredients taxonomy but not the additives taxonomy. I changed the function to use the ingredients taxonomy instead.

There may be unwanted false positives, it is useful to look at all the tests.


lib/ProductOpener/Ingredients.pm
```diff
@@ -2992,6 +3021,47 @@ sub parse_ingredients_text_service ($product_ref, $updated_product_fields_ref, $
...
+					my $parent_plus_child_ingredient_text;
+
+					if ($ingredients_lc eq "en") {
+						# oil (palm) -> palm oil
+						$parent_plus_child_ingredient_text = $ingredient . ' ' . $parent_ingredient_text;
+					}
+					else {
+						# huile (palme) -> huile palme
+						$parent_plus_child_ingredient_text = $parent_ingredient_text . ' ' . $ingredient;
+					}
```
alexgarel, line 3042: Why not test both for every language ?

lib/ProductOpener/Ingredients.pm
```diff
@@ -5943,12 +6037,13 @@ my %ingredients_categories_and_types = (
...
-				"olive vierge", "olive extra vierge", "olive vierge extra", "palme",
-				"palmiste", "pépins de raisin", "sal", "sésame",
-				"soja", "tournesol", "tournesol oléique",
+				"coco", "colza", "coprah", "coton",
+				"graines de colza", "illipe", "karité", "lin",
+				"mangue", "noisette", "noix", "noyaux de mangue",
+				"olive", "olive extra", "olive vierge", "olive extra vierge",
+				"olive vierge extra", "palme", "palmiste", "pépins de raisin",
+				"sal", "sésame", "soja", "tournesol",
+				"tournesol oléique",
```
alexgarel, line 6046: I though you where removing that !

Ain't there a way we could generate the list for the various languages from the english names ?

Or isn't it the case to make a small taxonomy for this kind of things ? (so that we can scale with languages).

taxonomies/food/ingredients.txt
```diff
@@ -578,12 +578,6 @@ en: non-gmo sunflower lecithin
 hr: ne-gmo suncokretov lecitin
 it: lecitina di girasole non ogm
 
-# <en:emulsifier
-< en:sunflower lecithin
-fr: émulsifiant lécithine de tournesol
-fi: emulgointiaine auringonkukkalesitiini
-# ingredient/fr:émulsifiant-lécithine-de-tournesol has 100 products in 6 languages @2019-02-22
```
alexgarel, line 585: why won't those entries appear again ?

I would leave the entry commented with a comment about why we should not add it

lib/ProductOpener/Ingredients.pm
alexgarel, line 1: I think it would really be worth having a script that takes the JSONL (or parquet) and run analysis on every entry and then points the difference with current analysis.
Or even better (because you don't know when analysis was run on each entry) run two version of the analyzer on the parquet file and point at the differences.

This could be fast enough (you could also do it on a sample), and this would help really know the impact of a change in the analyzer.

alexgarel (approved):
LGTM

taxonomies/food/ingredients.txt
stephanegigandet: They should not appear again because I fixed the parsing so that they are recognized, see this new test: https://github.com/openfoodfacts/openfoodfacts-server/pull/11251/files#diff-ab86aa197f60d1dd875cf1f147d97168ac999c9c3de967adaf5b27c41aa4ec09R89

lib/ProductOpener/Ingredients.pm
stephanegigandet: I am planning to remove that, but it will need a lot of testing and probably adjustments to check that the new way (introduced in this PR) supports those cases. Much more work is needed I think.

lib/ProductOpener/Ingredients.pm
stephanegigandet: Well the more combinations we create, the greater the risk for false positives. e.g. in French, at least for oils, there's no sense in testing "[fruit / grain] huile" because it's always in the other order. We could test more combinations, but then I would prefer to wait until we have something like you suggest to see the effect of the changes.

lib/ProductOpener/Ingredients.pm
stephanegigandet: We could try something like that. I'm not sure where it should live exactly. If the sample is small enough (e.g. 1000 products or less), one option could be to store the analysis results in GitHub just like we do for many tests.

Or it could be a standalone script, to run manually.

PR #11254, opened 2025-01-17
taxonomy: polish translations for flour and pasta

Teiron:
### What

Add more translations for Polish language - mainly for flour and pasta


stephanegigandet (approved):
Thank you!

PR #11259, opened 2025-01-17
fix: Update donation year to 2025 in banner

### LINKED ISSUE #11227: Update fundraiser text to 2025
### Task

**Current text:**
"Please give to our 2024 Fundraiser"

Should be updated to 2025.

**Misc:**
I don't know how reliant OFF is on donations but in terms of UX/UI there is potential for improvement. 

Take a look at this great analysis of the Wikipedia donation banner:
https://medium.com/@hugopezot91/wikipedia-donation-ux-analysis-how-to-ask-for-money-the-right-way-9a4480a40f8a

### Part of story
- #5529

github-throwaway:
This also updates the year in the banner.

- Fixes #11227


github-throwaway:
The amount of files that need to be touched to update the year seems somewhat crazy to me... 🙈

github-throwaway:
Are there different test on the server?

```
Failed test 'get-auth-bad-user-password'
2025-01-17T19:43:05.8318511Z (  DIAG  )  job  4      at /opt/product-opener/lib/ProductOpener/Test.pm line 454.
2025-01-17T19:43:05.8522338Z (  DIAG  )  job  4    --- tests/integration//expected_test_results/api_v2_product_read/get-auth-bad-user-password.html	Fri Jan 17 19:27:17 2025
2025-01-17T19:43:05.8527100Z (  DIAG  )  job  4    +++ tests/integration//expected_test_results/api_v2_product_read/get-auth-bad-user-password.html.test	Fri Jan 17 19:43:05 2025
```

stephanegigandet:
> The amount of files that need to be touched to update the year seems somewhat crazy to me... 🙈 Maybe there is a way to fill it in automatically?

In practice the only string that needs to change is this one:

```
msgctxt "donation_title_2024"
msgid "Please give to our 2024 Fundraiser"
msgstr ""
```

In practice we do not need to change the msgctxt, it's just an id, it's not displayed.

Usually, we only update common.pot and en.po, and the other translations come from crowdin.

But yes, definitely, we should make the year a variable. Usually we define variables like this:

```
# variable names between { } must not be translated
msgctxt "f_pro_moderator_owner_set"
msgid "You are currently viewing products from {organization}."
msgstr "You are currently viewing products from {organization}."
```

But it was not used in this string, it should definitely be changed. That way when we need to change the year, we only change it at one place, in the template, when we call the string.

github-throwaway:
@stephanegigandet where do you usually define these variables? `lib/ProductOpener/Config_off.pm`? I'm somewhat unsure of the process here..

alexgarel:
@github-throwaway here the variable should be something we place in the context of the template that renders the banner. And then we must use a function in the template to render the message with the variable. (using f_lang instead of lang)

github-throwaway:
Yeah but where do you actually define the variable? In what file

github-throwaway:
@stephanegigandet any idea why the backend build fails?

alexgarel:
@github-throwaway would you be able to fix the merge conflicts ? Or should we close it and re-open one ?

If you open a PR from a fork, the `/update_tests_results` does not work, and we have to update them locally with `make update_tests_results`.

github-throwaway:
Yeah, let's close for now. I don't have the time to figure out why its not working 🙈

PR #11262, opened 2025-01-18
feat: Update the tagline to force the app upgrade

g123k:
Hi everyone!

We don't use these 2 JSON files for the tagline in the mobile app (hosted on GitHub instead).
That's why we force the old users to update the app.


teolemon:
What about F-Droid for Android ?
how about openfoodfacts.app ?

g123k:
This tagline is for versions < 4.17.
We only support Android/iOS, that's why the link is for Google Play (more users)

PR #11263, opened 2025-01-18
feat: add common categories found on re-commerce websites

teolemon:
### What
feat: add common categories found on re-commerce websites


stephanegigandet (approved):
Thanks

PR #11265, opened 2025-01-18
taxonomy: translate and add ingredients

McCio:


taxonomies/food/ingredients.txt
```diff
@@ -25303,6 +25303,19 @@ fr: préparation pain d'épices
...
+
+< en:emmer
+en: emmer flakes
+it: fiocchi di grano farro
+
+< en:emmer flakes
+< en:wholemeal emmer
+en: whole grain emmer flakes
+it: fiocchi di grano farro integrale
+
```
benbenben2, line 25318: I have a suggestion here

Instead of adding the ingredient + processing forms in ingredients.txt, -> add only the ingredient in ingredients.txt (in this particular case, it is already added) + in ingredients_processing.txt add the processing (for example, add "it: integrale" under "en: whole")

Then, there is a function in ProductOpener: https://github.com/openfoodfacts/openfoodfacts-server/blob/c0518c16562ca4b1ee92b458effed749fe3b654d/lib/ProductOpener/Ingredients.pm#L1417 that is combining the two.

It has 2 advantages:
1) if tomorrow you have "\<something else\> integrale" you will not need to add it in the taxonomy if that \<something else\> ingredient is already in ingredients.txt
2) the processing is extracted, the API will show that this is ingredient + processing

McCio: Thanks for the suggestion! I made an update :+1:

stephanegigandet (approved):
Grazie!

PR #11273, opened 2025-01-21
fix: separate granulated sugars from white sugars

stephanegigandet:
probably needs more work, to make ingredients and categories taxonomies converge


aleene:
I am ok with the idea that we need a separate sugar tree for grain size and one for the whiteness. Probably we need another one for the humidity level.

stephanegigandet:
replaced by https://github.com/openfoodfacts/openfoodfacts-server/pull/11325

PR #11281, opened 2025-01-23
taxonomy: add Polish translations

Teiron:
### What
- add pork tripe (important part of Polish dish https://en.wikipedia.org/wiki/Flaki)
- add honey flavouring
- add more Polish assorted translations of taxonomy


benbenben2 (approved):
Great work!

PR #11288, opened 2025-01-25
fix: Import format_subdomain

### LINKED ISSUE #11287: Undefined format_subdomain called at product_jqm2.pl line 128.
### What
- We cannot save some products in Smoothie because of a bug. Probably non food products, that would need redirection.
- Looks like function `format_subdomain` is not reachable.
```perl
# There is an existing product
# If the product has a product_type and it is not the product_type of the server, redirect to the correct server
# unless we are on the pro platform

if (    (not $server_options{private_products})
	and (defined $product_ref->{product_type})
	and ($product_ref->{product_type} ne $options{product_type}))
{
	redirect_to_url($request_ref, 307,
		format_subdomain($subdomain, $product_ref->{product_type}) . '/cgi/product_jqm.pl?code=' . $code);
}
```

### Part of
- https://github.com/openfoodfacts/smooth-app/issues/6279

hangy:
### What

This pull request includes updates to the `cgi/product_jqm_multilingual.pl` file, including a copyright year update and an additional module import.

General updates:

* [`cgi/product_jqm_multilingual.pl`](diffhunk://#diff-92f19641303749fe28f85e1fac66e3beb133831802fa003c74c4abec2c290b7dL6-R6): Updated the copyright year from 2023 to 2025.

Code enhancements:

* [`cgi/product_jqm_multilingual.pl`](diffhunk://#diff-92f19641303749fe28f85e1fac66e3beb133831802fa003c74c4abec2c290b7dR46): Added `format_subdomain` from `ProductOpener::URL` to the list of imported modules.

- Fixes #11287

### DIFF
--- a/cgi/product_jqm_multilingual.pl
+++ b/cgi/product_jqm_multilingual.pl
@@ -3,7 +3,7 @@
 # This file is part of Product Opener.
 #
 # Product Opener
-# Copyright (C) 2011-2023 Association Open Food Facts
+# Copyright (C) 2011-2025 Association Open Food Facts
 # Contact: contact@openfoodfacts.org
 # Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
 #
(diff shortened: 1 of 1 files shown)


alexgarel:
I will apply it to OFF production directly… because it seems important enough.

PR #11306, opened 2025-01-28
taxonomy: canned foods - vinegars - cereals additions

OpenChris:


stephanegigandet (approved):
Thanks a lot @OpenChris !

PR #11344, opened 2025-02-07
feat: use ingredients processing for NOVA classification

stephanegigandet:
This PR adds support for using ingredients processing for the NOVA classification.

e.g. if we have "modified strawberry", en:modified will be an ingredient processing, and it will trigger NOVA 4.


alexgarel (approved):
Great.

It would be great to document taxonomies properties somewhere ! (I think an OpenAPI like description would be great)

PR #11347, opened 2025-02-10
chore: taxonomy changes 25:5

aleene:
### What
More request, etc. Continued integration for Prices.


stephanegigandet (approved):
Thanks

stephanegigandet (commented):
A lot of translations and entries seem to have been removed, is that intended?

aleene:
I often move things around a bit to put similar things together.

aleene:
But in this case it is caused by the merges

aleene:
@stephanegigandet a test fails. I do not know how to repair it.

aleene:
To many errors in the file

PR #11357, opened 2025-02-11
ci: tentative repair Imager::File::WEBP install

alexgarel:

PR #11358, opened 2025-02-12
ci: tentative repair Imager::File::WEBP install - 2

stephanegigandet:
duplicate of https://github.com/openfoodfacts/openfoodfacts-server/pull/11357 with fix from @hangy


alexgarel (commented):
Great ! Thanks @hangy :-)

PR #11362, opened 2025-02-12
docs: openapi for reading and writing tags fields

stephanegigandet:
This in an attempt to improve the documentation for reading tags fields, and updating them (replacing or adding to them) using API v2 or API v3.

At this time I just listed the categories tags, so that we can finalize it before adding the other tag fields.

There are a lot of complexities due to the difference we make for categories_hierarchies (which contains the "truth" values for the categories of the product) and categories_tags (which is a lossy version where entries that could not be matched to a taxonomy have been deaccented and normalized).

I think we could simplify things by having only one field, that we would name categories_tags but that would contain the content of the current categories_hierarchy field. And we would not have a specific field for indexing only.


docs/api/ref/schemas/tags/tags-write.yaml
```diff
@@ -0,0 +1,27 @@
...
+    items:
+      $ref: ./input_taxonomy_tag_entry.yaml
+  categories_tags_add:
+    type: array
+    description: An array of categories tag entries that will be added to existing categories
+    items:
+      $ref: ./input_taxonomy_tag_entry.yaml
+
+patternProperties:
+  categories_tags_(?<language_code>\w\w):
```
stephanegigandet, line 18: I added support for things like categories_fr_tags = "Gâteaux", but it means we need patternProperties. Maybe we could deprecate that feature. The alternative is to send tags_lc=fr + categories_tags = "Gâteaux", which is not very complex either.

docs/api/ref/schemas/tags/tags-write.yaml
```diff
@@ -0,0 +1,27 @@
...
+title: Tags (WRITE)
+description: |
+  Products have list of tags of different types (e.g. categories, labels) that can be replaced or added to with the following properties in API v3.
+properties:
+  categories_tags:
+    type: array
+    description: An array of categories tag entries that will replace existing categories
+    items:
+      $ref: ./input_taxonomy_tag_entry.yaml
+  categories_tags_add:
```
stephanegigandet, line 11: Instead of having a _add suffix for all tag properties, we could remove that, and instead have a default parameter like tags_mode which would be "add" or "replace". Similiar to the tags_lc parameter that indicate the default language for tags.

john-gom:
I would prefer it if the tags fields were the exact cross-reference to a taxonomy entry (even if that taxonomy entry doesn't exist yet) as it will make it easier to join the data together in queries, e.g. if we were to add taxonomies to off-query so that we can do things like translation in the query itself.

john-gom:
> That's why we have categories_hierarchy, which contains "en:cakes" + "fr:Tartes aux pêches". That's what I would prefer to keep.

I guess that could work as long as the one where the taxonomy does exist is the exact taxonomy tag.

What is the current process to update the categories_hiearchy to be "en:cakes" + "en:peach-tarts" when the taxonomy entry for "Tartes aux pêches" is created?

docs/api/ref/schemas/product_tags.yaml
```diff
@@ -13,17 +13,22 @@ properties:
 
   categories:
     type: string
+    description: |
+      Comma separated list of categories (not taxonomized), in the last language used to edit it (recorded in categories_lc)
+      This field is mostly used for debugging and testing purposes. Do not use it for display purposes.
   categories_hierarchy:
     type: array
+    description: An array of categories tag entries
```
alexgarel, line 21:
```suggestion
    description: |
      An array of categories tag entries.
      
      It only contains categories found in the taxonomy.
```

docs/api/ref/schemas/product_tags.yaml
```diff
@@ -13,17 +13,22 @@ properties:
...
     items:
-      type: string
+      $ref: ./tags/taxonomy_tag_entry.yaml      
   categories_lc:
     type: string
-    description: Categories language code
+    description: Language code of the last edit for categories
   categories_tags:
     type: array
+    description: An array of indexed categories tag entries (lossy, for search only)
```
alexgarel, line 29: I don't think it's clear enough.

```suggestion
    description: |
      An array of indexed categories tag entries.
      
      That is the id of categories found in taxonomy + 
      categories not found in taxonomy (with case / accents / spaces normalized).
      
      This is mostly used for search.
```

docs/api/ref/schemas/tags/tags-write.yaml
alexgarel: Yes would be better to show it as deprecated.

alexgarel (approved):
Almost good to me, I made some suggestions.

docs/api/ref/schemas/product_tags.yaml
stephanegigandet: @alexgarel it's not correct, categories_hierarchy contains all categories entries, including those that are not in the taxonomy. The difference with categories_tags is that those entries that are not in the taxonomy are not normalized (e.g. we keep spaces, accents etc.)

stephanegigandet: I'll try to reformulate

stephanegigandet:
> I guess that could work as long as the one where the taxonomy does exist is the exact taxonomy tag.

Yes, is is the canonical tag.

> What is the current process to update the categories_hiearchy to be "en:cakes" + "en:peach-tarts" when the taxonomy entry for "Tartes aux pêches" is created?

We need to run update_all_products.pl --fields categories again, possibly limited with --query categories_tags=fr:tartes-aux-peches

PR #11373, opened 2025-02-12
taxonomy: fix bananas

stephanegigandet:
This is to fix a build error we have on the main branch:

ERROR - unknown_parent - en:bananas-fresh-whole-red has an undefined parent 

+ some changes to make the bananas categories similiar to other categories (no "Bananas (something)")


aleene:
I see you did revert some changes I made. Esp Bananas (X) to X Bananas. The reason I did make these changes is that it is much easier to see in the web popup.

aleene:
Also see here: https://wiki.openfoodfacts.org/Category/Fruits/Bananas

alexgarel (approved):

PR #11374, opened 2025-02-12
taxonomy: fix taxonomy build error: unexisting whole-bananas parent

stephanegigandet:
fix so that taxonomy can build, but more changes would be good, see: https://github.com/openfoodfacts/openfoodfacts-server/pull/11373


aleene:
At some stage I added the whole adjective, but later removed it. Forgot to do this one. I guess.

teolemon (approved):

Issue #11334, opened 2025-02-06
Store number of unique scans by country for the last year in the product object


stephanegigandet:
We currently have one popularity sort key that is the same for the whole world. So the order of products by popularity is the same in all countries (but some are filtered if they are not sold in the country).

We currently store scans statistics in a JSON file next to the product .sto files:
- https://github.com/openfoodfacts/openfoodfacts-server/issues/5083

In order to be able to show popular products by country, we could store the raw number of scans for each country in the last year directly in the product object.

e.g. unique_scans : { world : 1000, fr : 750, be: 200, nl: ... } etc.

- This would also allow to show those numbers to users https://github.com/openfoodfacts/openfoodfacts-server/issues/8668

- and to producers https://github.com/openfoodfacts/openfoodfacts-server/issues/9741

We could also remove the popularity key (which is a bit complex as it uses unique scans, top scans, and date of last image), if we can find a way to get sorted products by unique_scans in a country (primary key) and by another field like last_modified_t (secondary key) so that we have a unique ordering of products (in order to have products by page).

This unique_scans by country could be saved in off-query, so that we can sort on it to retrieve lists of products.

john-gom:
Proposed approach is to create a product_country table which would have a record for every row in product_countries_tag.

We would store the country-specific sort here along with the world sort to have sensible sorting where products have a small number of country-local scans. Table definition:

`create table product_country (
  product_id int not null,
  country_id int not null,
  country_scans int not null,
  all_scans int not null,
  constraint "product_country_pkey" primary key ("product_id", "country_id"));
create index product_country_ix1 on product_country (country_id, country_scans DESC, all_scans DESC, product_id);
`

On a database with just over 3 million products this table and indexes take up about 365 MB of storage.

The best way to query is something like this:

`select * from product p 
join product_country pc on pc.product_id = p.id and pc.country_id = 372 
where id in (
  select product_id from product_country pc
  where pc.country_id = 372
  order by pc.country_scans desc, pc.all_scans desc, pc.product_id 
  limit 50 offset 100000)
order by pc.country_scans desc, pc.all_scans desc, pc.product_id
;`

The key here is to put the limit and offset inside the sub-query so that the PostgreSQL only needs to perform the LIMIT and OFFSET operation on the inner table, which has a very small record size. I found that I had to supply the country_id to get an efficient query,  doing something like `pc.country_id IN (select id from country where code = 'fr')` gave a much worse query plan.

[...]

PR #11378, opened 2025-02-12
feat: Script to upload scans to off-query

john-gom:
### What

Script to load scans into off-query

- Part of #11334


stephanegigandet (approved):
Looks good!

Issue #11241, opened 2025-01-15
Make facets start with a prefix


alexgarel:
### Problem

It's currently very difficult to tell if a url is a "facet" search.

### Proposed solution

1. Prefix facets uri with a "f/".
   - `/categories` -- becomes -> `f/categories`
   - `/category/speculaas` -- becomes -> `f/category/speculaas`

2. Make old facets redirect uri to the new uris (permanent redirect)

I think the redirection can be done by apache (it is as fast as checking if first uri part is a taxonomy name), because generating all urls for nginx might be a bit cumbersome.

### Additional context

This can help with:
* adding rate limiting on facet on nginx side
* better handling of priorities

Generally it's a good thing to be able to easily tell what kind of content a URL is.

stephanegigandet:
If we change URLs, then maybe it would be good to think a bit more about what URLs we want.

Adding a prefix for facets is a good idea. Another one that we floated around was to get rid of the translation of facet type.

e.g. instead of having https://fr.openfoodfacts.org/categorie/bananes , we use /category/

In fact I think it would be even better to get rid of the plural / singular distinction.

Today:

/categories -> list of facets, that links to /category/bananas (or /categorie/bananes) in French.

I think it would make sense to always use the plural tag types:

/f/categories
/f/categories/bananas

and always with the English name for the tag type (but keeping the local name for the tag value I think).

e.g.

in French: /f/additives/e330-acide-citrique , (but not /f/additives/e330-citric-acid )

alexgarel:
I agree with it @stephanegigandet.

Letting the translation of the facet name goes makes me a bit less enthusiast, but if it really cuts complexity, let's do it.

alexgarel:
A minor question is: should /f alone redirect to the home page or list available facets.

I would prefer redirect to the home page for now, but it might help discoverablity to list available facets.

teolemon:
What about /c/bananas ? Also, one thing for multi-languages, and stability is usually adding a stable id

/c/bananas-id12345
/c/bananes-id12345
/c/new-principal-name-of-bananas-id12345

alexgarel:
I don't understand you point @teolemon 

It does not help identify facets if you don't put /f or /facets before.

For the id, what would it be ? We don't have such an id yet !

stephanegigandet:
Personal preference: understandable and readable / editable / guessable urls. :)

[...]

PR #11387, opened 2025-02-12
feat: adding a prefix to facets

alexgarel:
+ some cleanup in product routing

fixes: #11241


stephanegigandet:
Cool feature and refactoring!

I'm testing it locally, a few things I observed:

- we still display links like http://fr.openfoodfacts.localhost/label/bio
- the link above is redirected to http://fr.openfoodfacts.localhost/facets/labels/fr:bio-europeen/1  --> we should remove the fr: and the 1
- we still display links like http://world.openfoodfacts.localhost/nova-groups
- URLs like http://world.openfoodfacts.localhost/labels are not redirected to /facets/labels
- http://world.openfoodfacts.localhost/facets/label/2 -> infinite redirect loop

alexgarel:
Thanks for the reports @stephanegigandet I will add tests for those cases.

alexgarel:
@stephanegigandet I fixed most of the bugs.

I have to look closely at [this commit](https://github.com/openfoodfacts/openfoodfacts-server/pull/11387/files/37c147e12d1f2d38700451980fe0f88637a75349..be2a5acee43c654286907adcd08c37e0bb37c0bd) to verify.

I think there are still links without /facets (hard to find).

lib/ProductOpener/Display.pm
```diff
@@ -1276,7 +1275,7 @@ sub display_index_for_producer ($request_ref) {
 		if ($count > 0) {
 			push @{$template_data_ref->{facets}},
 				{
-				url => "/" . $tag_type_plural{$tagtype}{$lc},
+				url => "/facets" . $tag_type_plural{$tagtype}{$lc},
```
stephanegigandet, line 1278: not tested, but I think we may be missing a /

```suggestion
				url => "/facets/" . $tag_type_plural{$tagtype}{$lc},
```

lib/ProductOpener/Routing.pm
```diff
@@ -454,34 +464,51 @@ sub redirect_text_route($request_ref) {
...
 
 sub facets_route($request_ref) {
 
+	my $is_obsolete_url = undef;
+
 	my $target_lc = $request_ref->{lc};
 	$request_ref->{canon_rel_url} = '';
 	my $canon_rel_url_suffix = '';
 
+	# check if we use the facet prefix, and however add it to canonical_url
```
stephanegigandet, line 503: the "however" is not clear, I'd suggest something more verbose like:

```suggestion
	# add the facets prefix to the canonical_url
	# the facets prefix may not be in older facet urls (before we prefixed them with /facets), in which case we don't consume the first component and mark the url obsolete
```

tests/integration/expected_test_results/page_crawler/normal-user-access-nested-facet-page.html
```diff
@@ -10,7 +10,7 @@
 	<meta property="fb:app_id" content="219331381518041">
     <meta property="og:type" content="food">
     <meta property="og:title" content="">
-    <meta property="og:url" content="//world.openfoodfacts.localhost/category/en:hazelnut-spreads/brand/nutella">
+    <meta property="og:url" content="//world.openfoodfacts.localhost/facets/categories/en:hazelnut-spreads/brands/nutella/1">
```
stephanegigandet, line 13: not a big deal, but we probably need the same rules you added to remove the en: and the 1. (the en: issue preexisted before your change)

tests/integration/expected_test_results/product_read/get-existing-product.html
```diff
@@ -25,7 +25,7 @@
 <meta name="theme-color" content="#ffffff">
 
 	<meta name="apple-itunes-app" content="app-id=588797948">
-    <link rel="canonical" href="//world.openfoodfacts.localhost/product/0200000000034/some-product">
+    <link rel="canonical" href="//world.openfoodfacts.localhost/product/2000000000034/some-product">
```
stephanegigandet, line 28: that one is strange, i can't reproduce it locally

alexgarel: It's me that change the code, the test code was a 12 digits code, and now I issue a redirect for these.

tests/integration/expected_test_results/page_crawler/normal-user-access-nested-facet-page.html
alexgarel: Honestly for a canonical ur, I think it make sense to be really explicit.

stephanegigandet:
(making some changes for pro platform)

stephanegigandet:
tested locally on pro platform, looks good

PR #11391, opened 2025-02-13
fix: redirect pro.off to world.pro.off

stephanegigandet:
reported by @github-throwaway on Slack

we have code in Apache to redirect to world, but nginx wasn't configured to answer for the pro.off subdomain


alexgarel (approved):
Thanks !

PR #4818, opened 2021-02-03
feat: Try to detect foreign languages in ingredient lists (2019 PR on fresh code)

teolemon:
https://github.com/openfoodfacts/openfoodfacts-server/pull/1862/files
Reapply 2019 PR on fresh code (multilang ingredient lists)
Try to detect foreign languages in ingredient lists

PR #11392, opened 2025-02-13
feat: more data quality warnings for ingredients language mismatch

stephanegigandet:
refactored some code + new data quality warnings + tests

replaces https://github.com/openfoodfacts/openfoodfacts-server/pull/4818


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -1504,6 +1504,17 @@ Checks related to the ingredients list and ingredients analysis.
 
 =cut
 
+# note: we need common ingredients words that exist only in 1 language
+my %ingredients_in_languages_regexps = (
+	"en" => qr/\b(sugar|water|salt|flour)\b/i,
```
benbenben2, line 1509:
```suggestion
	"en" => qr/\b(sugar|salt|flour)\b/i,
```
water is also water in NL
https://www.wikidata.org/wiki/Q283

benbenben2:
We could relate this with this work: https://github.com/openfoodfacts/openfoodfacts-server/pull/11267
For which, a possible future dev might be - for some languages (rich enough in the taxonomies) - if less than let's say 10% of ingredients are known only, and if first let's say 3 ingredients are found in the taxonomy in another language, then we can assume that the language is wrong.

benbenben2 (approved):
LGTM

PR #11407, opened 2025-02-13
feat: dq correct wrong language for tags v2

### LINKED ISSUE #7838: Create a script to correct all taxonomized values for categories that are stored in an incorrect language by storing them in the right language

benbenben2:
### What
Try a different approach to correct languages for tags.
- categories
- labels
- origins

### Screenshot
Example of log table in duckdb:
Screenshot_20250213_213757

Just to give an idea, the intermediate table tags_to_update_df:
Screenshot_20250213_222343

- Fixes #7838


scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+        lang TEXT,
+        old_tags_text TEXT,
+        new_tags_text TEXT,
+        updated_tags_text BOOLEAN,
+        updated BOOLEAN
+    )
+    ''')
+    for _, row in tags_to_update_df.iterrows():
+        code = row[0]
+        lang = row[1]
```
alexgarel, line 269: can't you use column names in the dataframe ?

scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+        
+        existing_row = conn_db.execute(f"SELECT * FROM products_to_update_{tag_type} WHERE code = '{code}'").fetchone()
+        
+        if existing_row:
+            # Update the existing row
+            # existing_row[2] is tags as text already updated
+            # row[3] is unknown tag with lc
+            # row[4] is unknown tag
+            # row[8] is row[4] tag found in another language
+            new_tags_text, updated_tags_text = update_tags(existing_row[3], row[3], row[4], row[8])
```
alexgarel, line 279: Using named parameters helps readability

```suggestion
            # row[3] is unknown tag with lc
            # row[4] is unknown tag
            # row[8] is row[4] tag found in another language
            new_tags_text, updated_tags_text = update_tags(tags_text=existing_row[3], old_tag_with_lc=row[3], old_tag=row[4], new_tag=row[8])
```

scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+
+
+    updated_tags_text = ", ".join(tags_list)
+
+    return updated_tags_text, updated
+
+
+time_start = datetime.now()
+
+for tag_type in tag_types:
```
alexgarel, line 199: It would be better to have some fuctions we call here, to keep things simple:
* one that create the tags_to_update_df
* one that creates and populate products_to_update_{tagtype}
* one that run the modifications

scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+
+    for row in all_rows:
+        code = row[0]
+        lang = row[1]
+
+        try:
+            country = mapping_languages_countries[lang]
+        except KeyError:
+            print(f"ERROR: language {lang} is not referenced in mapping_languages_countries")
+            sys.exit()
```
alexgarel, line 313: do you need a sys.exit() ? isn't an error log enough ?

scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+    all_rows = conn_db.execute(f"SELECT * products_to_update_{tag_type}").fetchall()
+
+    for row in all_rows:
+        code = row[0]
+        lang = row[1]
+
+        try:
+            country = mapping_languages_countries[lang]
+        except KeyError:
+            print(f"ERROR: language {lang} is not referenced in mapping_languages_countries")
```
alexgarel, line 312:
```suggestion
            print(f"ERROR: language {lang} is not referenced in mapping_languages_countries", file=sys.stderr)
```

scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+import duckdb
+import re
+import requests
+import sys
+import unicodedata
+
+data = {
+    'user_id': "TODO",
+    'password': "TODO",
+}
```
alexgarel, line 34: call is BASE_DATA and underneath use:

data = dict(BASE_DATA, myvalue=1, othervalue="foo")

scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+        try:
+            country = mapping_languages_countries[lang]
+        except KeyError:
+            print(f"ERROR: language {lang} is not referenced in mapping_languages_countries")
+            sys.exit()
+
+        data.update({
+            'code': code,
+            tag_type: row[3],
+        })
```
alexgarel, line 318: See my comment above on BASE_DATA, it's dangerous to reuse the same dict (and it's hard to think it comes from top of the file !)

scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+        try:
+            post_call_url_res = requests.post(
+                post_call_url.format(country=country),
+                data=data,
+                headers=headers,
+            )
+
+            if post_call_url_res.status_code != 200:
+                print(f"ERROR: when updating product {code}. Received {post_call_url_res.status_code} status code")
+                sys.exit()
```
alexgarel, line 328: Why exit, we might continue indeed (but if we have a 429 maybe ?)

scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+
+            conn.execute(f'''
+                UPDATE products_to_update_{tag_type} 
+                SET updated = true
+                WHERE code = ?
+            ''', [code])
```
alexgarel, line 334: you do that, but at the end you do not use this table ?

scripts/tags_and_languages.py
```diff
@@ -0,0 +1,338 @@
...
+    ON products_unknown_tags.{tag_type}_tag = taxonomy_tags.lc_tag_name
+    WHERE taxonomy_tags.lc_tag_name IS NOT NULL
+    AND taxonomy_tags.lc != 'xx'
+    AND products_unknown_tags.{tag_type}_lc_and_tag != taxonomy_tags.lc_tag_name_concat
+    ''').df()
+    # remark:
+    #   AND products_unknown_tags.{tag_type}_lc_and_tag != taxonomy_tags.lc_tag_name_concat -> to prevent fr:angleterre and fr:angleterre
+
+    print(f"prepare log table - time {datetime.now()-time_start}")
+    conn_db = duckdb.connect(database='tags_and_languages.db', read_only=False)
```
alexgarel, line 253: maybe use a different name from above, it's more fail safe !

eg: `updated_db`

alexgarel (approved):
Fantastic work @benbenben2 !

I leaved some comments mostly on the form.

scripts/tags_and_languages.py
benbenben2: Instead of using dataframe, it will use duckdb result of query (tuple).
It is better for performances. But column names cannot be used. Instead the variables names have been renamed to be more explicit.

scripts/tags_and_languages.py
benbenben2: the code interrupt here
the user will have to manually update mapping_languages_countries
and resume the job

scripts/tags_and_languages.py
benbenben2: Let's see how it goes (I commented that part while running it locally for now) with post requests limitation.

scripts/tags_and_languages.py
benbenben2: Idea is to be able to resume the update of products after a failure.
See last modifications.

alexgarel (approved):
Thanks @benbenben2 !

PR #11409, opened 2025-02-14
fix: always show as sold nutrition in petfood edit form

stephanegigandet:
issue reported by @benbenben2 

fix applied in opff prod:
https://hr.openpetfoodfacts.org/cgi/product.pl?type=edit&code=8716811001809

### DIFF
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -1395,7 +1395,10 @@ CSS
 	}
 
 	# In all cases, if we have data, we will check the checkbox.
-	if ($nutrition_data_exists{""}) {
+	# We also check the "as sold" checkbox for petfood products,
+	# as we don't display the checkboxes to indicate the presence of nutrition data for "as sold" and "prepared" for petfood products
+	# (they can only have "as sold" nutrition data)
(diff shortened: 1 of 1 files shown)


benbenben2:
It might not be as simple as that.

For example for food: https://hr-en.openfoodfacts.org/cgi/product.pl?type=edit&code=3850334406262
You cannot edit and modify the values in the nutrition table for this product.
You have to do something (workaround) like check the box above and uncheck, to make values fields clickable.

For pet food, I am afraid that even if we change the default check or unchecked it will remain non-clickable with no possible workaround.

stephanegigandet:
> It might not be as simple as that.
> 
> For example for food: https://hr-en.openfoodfacts.org/cgi/product.pl?type=edit&code=3850334406262 You cannot edit and modify the values in the nutrition table for this product. You have to do something (workaround) like check the box above and uncheck, to make values fields clickable.

The nutrition table edit is showing for me for that product:

image

> For pet food, I am afraid that even if we change the default check or unchecked it will remain non-clickable with no possible workaround.

Well I applied the fix to OPFF production, do you see products where it does not work?

alexgarel (approved):
This cannot harm I think (although it might be better to do this kind of check just before saving products).

PR #11412, opened 2025-02-14
docs: add minimal git version 2.25.0 in quick start doc

nmetifiot:
### What

I've add one ligne in doc quick start to specify that the minimal version of git is 2.25.0


stephanegigandet (approved):
Thank you!

alexgarel (approved):
Thanks !

hangy (commented):
Thanks for the suggestion! Just a note: Git 2.25.0 is a five years old release. Do we need to list minimum versions that should be common sense, because anything that old probably has severe CVEs, too? 🤔

nmetifiot:
In my opinion it is very confortable to know all minimum version when I install my dev setup. Because I don't upgrade all my tools systematically. Sometimes more informations in documentation is better than less informations ☺️
But this PR is just a proposition.

PR #11431, opened 2025-02-14
taxonomy add stopword in Croatian

benbenben2:
### What
add **čuvati** stopword

- Fixes #-none-


hangy (commented):
The change seems redundant as  `(č|Č|c|C|ć|Ć)uvati|(č|Č|c|C|ć|Ć)uvajte` should already match `čuvati`. If it it doesn't match, that might be caused by `[prije otvaranja ]`; I don't quite understand why it's written using a character set.

stephanegigandet (approved):

PR #11432, opened 2025-02-15
taxonomy: translate fruits, jams and misc

McCio:


McCio:
I don't know why the old PR was closed (I closed it apparently), but this should be the same with the fix you noted @stephanegigandet

stephanegigandet (approved):
Thank you @McCio !

raphael0202:
@McCio @stephanegigandet I believe we need the `en:` language prefix after the wikidata property: `wikidata:en: XXX`.
The error popped on taxonomy editor when I tried to import the taxonomy.
I'm submitting a PR to fix the error.

PR #11436, opened 2025-02-16
fix: typo

teolemon:
### What
fix: typo

### DIFF
--- a/po/common/en.po
+++ b/po/common/en.po
@@ -4685,7 +4685,7 @@ msgid "Processed foods"
 msgstr "Processed foods"
 
 msgctxt "attribute_nova_4_description_short"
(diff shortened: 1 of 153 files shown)


po/common/en.po
```diff
@@ -4685,7 +4685,7 @@ msgid "Processed foods"
 msgstr "Processed foods"
 
 msgctxt "attribute_nova_4_description_short"
-msgid "Ultra processed foods"
+msgid "Ultra-processed foods"
```
stephanegigandet, line 4688: changing the msgid does not work if it's not changed in common.pt too (in which case we lose the translations though..)
I replaced it in common.pot and all .po files

stephanegigandet (approved):

PR #11441, opened 2025-02-17
build: run apt update in needed in Dockerfile at various steps

alexgarel:
Because first step that runs apt update might be taken from cache.


stephanegigandet (approved):
Thank you!

PR #11453, opened 2025-02-18
build: Better handling of apache2 in docker

alexgarel:
the entrypoint should enable graceful stop

(note: taken from #9706)

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -62,10 +62,12 @@ then
   # handle graceful shutdown of apache on sigterm
   _term () {
     echo "Caught SIGTERM signal!"
-    while [[ ! -e /run/apache2/apache2.pid ]]
+    i=5  # don't wait too much, hard kill is still an option
+    while [[ ( ! -e /run/apache2/apache2.pid && i -gt 0 ) ]]
     do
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):

docker/docker-entrypoint.sh
```diff
@@ -56,7 +56,25 @@ chown www-data:www-data /tmp/env-export.sh && chmod 0400 /tmp/env-export.sh
...
 
-exec "$@"
+if [ "$1" = 'apache2ctl' ]
+then
+  # handle graceful shutdown of apache on sigterm
+  _term () {
+    echo "Caught SIGTERM signal!"
+    while [[ ! -e /run/apache2/apache2.pid ]]
+    do
+      echo "Waiting for apache to be fully started..."
```
hangy, line 67: Is there any case where this loop might deadlock if apache never starts successfully, or is that a non-issue?

alexgarel: We could add a max wait time indeed.

alexgarel:
I tested and this does not work … I don't know why...

with docker compose > 2.30 we could send a pre_stop command.

Closing for now.

PR #11454, opened 2025-02-18
fix: gen_feeds_daily.sh: failure email for export_database.pl

stephanegigandet:
As suggested by @alexgarel , this PR adds a mechanism to see if some of the commands failed, and list them at the end so that we get them in the email failure.

If this is ok, we can add checks for the other commands (and possibly make those commands return a non 0 status on errors if they don't already).

### DIFF
--- a/scripts/gen_feeds_daily.sh
+++ b/scripts/gen_feeds_daily.sh
@@ -31,6 +31,9 @@ fi
 
 cd $OFF_SCRIPTS_DIR
 
+ERRORS=0
+FAILED_COMMANDS=""
+
 # off-pro flavor: we don't generate most exports
 # but we have some special processing
 if [ "$PRODUCT_OPENER_FLAVOR_SHORT" == "off-pro" ]; then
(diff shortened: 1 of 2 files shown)


scripts/gen_feeds_daily.sh
```diff
@@ -40,13 +43,22 @@ if [ "$PRODUCT_OPENER_FLAVOR_SHORT" == "off-pro" ]; then
...
 ./remove_empty_products.pl
 ./gen_top_tags_per_country.pl
 
 # Generate the CSV and RDF exports
 ./export_database.pl
+RETURN=$?
+
+if [ $RETURN -ne 0 ];
+then
+    echo "export_database.pl not executed successfully - return value: $RETURN"
```
alexgarel, line 55:
```suggestion
    >&2 echo "export_database.pl not executed successfully - return value: $RETURN"
```

scripts/gen_feeds_daily.sh
```diff
@@ -113,4 +125,17 @@ then
...
 fi
 
+# If there were commands that resulted in errors,
+# echo the list of commands so that it is included in the
+# failure e-mail sent to root
+if [ $ERRORS -gt 0 ];
+then
+    echo "$ERRORS ERROR(S) DURING EXECUTION OF gen_fields_daily.sh"
+    echo "FAILED COMMANDS:
+$FAILED_COMMANDS"
```
alexgarel, line 135:
```suggestion
    >&2 echo "$ERRORS ERROR(S) DURING EXECUTION OF gen_fields_daily.sh"
    >&2 echo "FAILED COMMANDS:
$FAILED_COMMANDS"
```

alexgarel (approved):
LGTM

I added suggestions to echo on stderr instead of stdout

PR #11460, opened 2025-02-19
fix: Fix resume from checkpoint code in export scans

john-gom:

### What

Fix resume from checkpoint when exporting scans to off-query

- #11334


stephanegigandet (approved):
Thanks!

PR #11471, opened 2025-02-20
fix: compute Nutri-Score stats for categories - WIP

stephanegigandet:
In December 2021 we removed the feature that was showing the average Nutri-Score for a category. I don't remember why.

There is some corresponding dead code. We may want to put the feature again.

gen_top_tags_per_country.pl is very outdated, we probably should improve it a bit.

This is a work in progress.


stephanegigandet:
Old PR, not sure it will be finished, closing for now

PR #11476, opened 2025-02-20
taxonomy: Translation of the misc taxonomy

teolemon:
### What
taxonomy: Translation of the misc taxonomy


taxonomies/misc.txt
```diff
@@ -1,18 +1,79 @@
 en: EcoScore extended data not computed
```
benbenben2, line 1:
```suggestion
en: EcoScore - Extended data not computed
```

PR #11479, opened 2025-02-20
refactor: Move Slack webhooks to common module

hangy:
### What

Moves Slack hook calls to reusable module to avoid redundant similar code. Also makes the Slack Hook URL an environment option, so that it can be kept out of source control.


lib/ProductOpener/Config2_docker.pm
```diff
@@ -158,4 +159,17 @@ $crm_username = $ENV{ODOO_CRM_USER};
...
+if ((defined $ENV{SLACK_HOOK_URLS}) and ($ENV{SLACK_HOOK_URLS} ne '')) {
+	foreach my $kvp (split(',', $ENV{SLACK_HOOK_URLS})) {
+		if (not($kvp =~ s/^(?<channel>.+)=(?<url>https?.+)$//)) {
+			next;
+		}
+
+		$slack_hook_urls{$+{channel}} = $+{url};
+	}
+}
+
```
alexgarel, line 174: Why put it there ? 

We have a lib/ProductOpener/ConfigEnv.pm for configurations that depends on environment, and it will be available everywhere.

.env
```diff
@@ -71,6 +71,7 @@ ODOO_CRM_URL=
 ODOO_CRM_DB=
 ODOO_CRM_USER=
 ODOO_CRM_PASSWORD=
+# SLACK_HOOK_URLS=\#infrastructure=https://example.com,\#team=https://example.org
```
alexgarel, line 74: If we want it also in prod, we should add it to env/env.off (and other flavors ?)

alexgarel (approved):
Cool.

I don't think those checks are running in production. Should we have a systemd timer + service to run them ?

See also my comments  before merging.

lib/ProductOpener/Config2_docker.pm
hangy: I wasn't aware of that file TBH. 😄

hangy: Fixed in c7ffc6c29c905468c998a531a7afd3485cbbd645

.env
hangy: @alexgarel Can you expand on how the secrets should be handled for the production config? It doesn't seem like the hook-URLs should be in git

alexgarel: @hangy (sorry if it's dead posting !).

In production secrets are in Config2_oxf.pm, which is not in git.
Personally I would prefer everything in environments, but for now that seems a bit complicated !

teolemon:
@alexgarel @hangy can we merge ?

.env
alexgarel: So my proposal:

Keep the env variable for docker, as you did, but read it in Config2_docker. When we deploy we can add it if Config2_oxf.pm on each server (we can do it before deploying indeed).

When we deploy in production,

alexgarel (approved):

PR #11480, opened 2025-02-20
fix: improve GS1Encoder loading

hangy:
### What

Suppress the useless `only used once` warning from `FFI::Platypus` from polluting logs.


hangy:
Doesn't have the desired effect

PR #11487, opened 2025-02-25
taxonomy: taxonomies-enhancer-croatia

benbenben2:
### What
apply taxonomies enhancer suggestion found for food products in Croatia

- Fixes #-none-


alexgarel (approved):
You are the one to best know !
So let's go :-)

PR #11494, opened 2025-02-26
ci: Update CODEOWNERS for less notification spam

teolemon:
### What
disable openfoodfacts-server as a codeowner to reduce notification spam (reviews will have to be requested manually)


john-gom:
Should we perhaps create an openfoodfacts-server-owners group in GitHub so that people who still want these emails can get them?

hangy (commented):
`CODEOWNERS` are not only about notification, but also about making a review from a group required. Should that not be the case any more?

stephanegigandet (changes requested):
There is an error message:

Unknown owner on line 7: make sure the team @openfoodfacts/taxonomy-reviewers exists, is publicly visible, and has write access to the repository
taxonomies/       @openfoodfacts/taxonomy-reviewers

teolemon:
thanks @stephanegigandet, updated with the new name of the team

teolemon:
still broken

john-gom (approved):

hangy (approved):

PR #11495, opened 2025-02-26
fix: add missing product_type field to products, 

### LINKED ISSUE #11492: Products without a product_type
Reported by @teolemon on Slack: https://openfoodfacts.slack.com/archives/C02LDQDDD/p1740566174871269

Some products still have no product_type field.

```
off> db.products.find({ product_type: { $exists : false }}, {code: 1})
[
  { _id: '12112788', code: '12112788' },
  { _id: '05571154', code: '05571154' },
  { _id: '2000000039454', code: '2000000039454' },
  { _id: '2000000005970', code: '2000000005970' },
  { _id: '2000000009975', code: '2000000009975' },
  { _id: '2000000033079', code: '2000000033079' },
  { _id: '20901722', code: '20901722' },
  { _id: '20004101', code: '20004101' },
  { _id: '20024765', code: '20024765' },
  { _id: '20024901', code: '20024901' },
  { _id: '20047214', code: '20047214' },
  { _id: '80122029', code: '80122029' },
  { _id: '20181611', code: '20181611' },
  { _id: '20841706', code: '20841706' },
  { _id: '20248321', code: '20248321' },
  { _id: '20499471', code: '20499471' },
  { _id: '20621483', code: '20621483' },
  { _id: '20641627', code: '20641627' },
  { _id: '20657284', code: '20657284' },
  { _id: '24039758', code: '24039758' }
]
Type "it" for more
off> db.products.count({ product_type: { $exists : false }})
DeprecationWarning: Collection.count() is deprecated. Use countDocuments or estimatedDocumentCount.
1805

```

It seems some products never had a product_type: https://world.openfoodfacts.org/product/05571154/danette-chocolat-delice-100-g

And in at least one instance, it was removed, apparently by scanbot (weird, scanbot just reads the product and adds scans): https://world.openfoodfacts.org/product/12112788/softer-butter-marks-spencer-250-g?rev=27
-> in this instance, scanbot also removed one category (butters), very strange

stephanegigandet:
will need to be run again on off, opf, obf, opff, off-pro to fix missing product_types

fixes #11492

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -1161,6 +1161,12 @@ sub store_product ($user_id, $product_ref, $comment) {
 		delete $product_ref->{server};
 	}
 
+	# If we do not have a product_type, we set it to the default product_type of the current server
+	# This can happen if we are reverting a product to a previous version that did not have a product_type
+	if (not defined $product_ref->{product_type}) {
+		$product_ref->{product_type} = $options{product_type};
+	}
+
 	# In case we need to move a product from OFF to OBF etc.
 	# we will have a old_product_type field
 
--- a/scripts/update_all_products.pl
+++ b/scripts/update_all_products.pl
@@ -329,6 +329,11 @@ if ($fix_string_last_modified_t) {
 	$query_ref->{last_modified_t} = {'$type' => "string"};
 }
 
+# Query products that don't have a product_type field
+if ($add_product_type) {
+	$query_ref->{product_type} = {'$exists' => 0};
+}
+
 # On the producers platform, require --query owners_tags to be set, or the --all-owners field to be set.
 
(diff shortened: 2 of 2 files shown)


teolemon (approved):
Any idea what could be causing the thing. 
Is there a scenario where product_type could be unset, or nether set (arbitrary product creation, older versions of the API, corner case in the current API (eg images uploaded on product which doesn't exist)… ?

stephanegigandet:
> Any idea what could be causing the thing. Is there a scenario where product_type could be unset, or nether set (arbitrary product creation, older versions of the API, corner case in the current API (eg images uploaded on product which doesn't exist)… ?

It may be multiple causes. At least a few instances are when a product was reverted to an earlier version without product_type. I added a fix for that, which might also fix other causes. We'll add the product types and verify later if we have new cases.

PR #11502, opened 2025-02-27
fix: Makefile improvements for better DX

rpowis:

- Fixes #[ISSUE NUMBER]


alexgarel (approved):
Ok ! I tend to use grep -P because in a perl project, it's easier to read for everyone ;-)

rpowis:
> Ok ! I tend to use grep -P because in a perl project, it's easier to read for everyone ;-)

I thought that was the case 😄  Unfortunately all greps aren't made equal 😢

rpowis:
@alexgarel this can this be merged now, unless you need me to make any changes 👍

PR #11528, opened 2025-03-03
fix: Override no_cache parameter if database is explicitly set to off-query

john-gom:

### What

When testing popularity queries via off-query we need to set the no_cache parameter but this supresses using off-query for some counts

- Part of #11334


stephanegigandet (approved):
Maybe we should rename can_use_query_cache() to can_use_off_query(). It's not really a cache now.

PR #11535, opened 2025-03-04
chore: use a .log extension for apache2 logs

alexgarel:
To have logrotate works correctly

(Already deployed to production)


teolemon (approved):

teolemon (commented):
Do we have any scripts, processes, 3rd party services listening to that specific location ? @alexgarel

stephanegigandet (approved):

PR #11555, opened 2025-03-07
fix: added skip tests while docs updations

JagjeevanAK:
### What
Added skip test feature when docs are updated only.
Also, Upgraded the node version from 12.x to 20.x, as Nodejs 12 is no longer maintained and has been deprecated.


rpowis:
Haha. I was also doing this on a branch with a bunch of other workflow
improvements.

It should be bumped to 22 though, as that’s what’s specified in the
dockerfile.

rpowis (approved):
LGTM 👍 I can confirm it works [here](https://github.com/rpowis/openfoodfacts-server/actions/runs/13702168645/job/38318512889)

.github/workflows/pull_request.yml
```diff
@@ -19,7 +19,7 @@ concurrency:
 
 jobs:
   filter:
-    name: "Filter changed paths"
+    name: "Doc file changed"
```
rpowis, line 22: Hmmm. I'm not sure that's accurate as it's filtering out more than just docs (css, js, etc.)

JagjeevanAK: what should be updated according to you?

rpowis: does it need changing?

JagjeevanAK: I have reverted the change i thought it could be good to understand directly what is it's primary job in frist place.

.github/workflows/pull_request.yml
```diff
@@ -85,7 +108,7 @@ jobs:
     name: 🐪 Check Perl
     needs: build_backend
     runs-on: ubuntu-latest
-    if: github.event_name == 'pull_request'
+    if: github.event_name == 'pull_request' && needs.filter.outputs.code_changed == 'true'
```
alexgarel, line 111: No, you should not do this because this check must pass for PR to be merged.

Currently
image

I imagine we could skip the steps instead ?

alexgarel (changes requested):

.github/workflows/pull_request.yml
alexgarel: Or am I wrong ?
from https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks
> A job that is skipped will report its status as "Success". It will not prevent a pull request from merging, even if it is a required check.

alexgarel (approved):
Hope this work, this would be a huge improvement.

@JagjeevanAK we have more "careful" rules to craft. For example if we only change taxonomies, we could avoid rebuilding the container and use the "main" image to test taxonomies.

.github/workflows/pull_request.yml
alexgarel: Indeed, I see now this was the case for this PR !

(maybe some times ago it was not the case, because I remember being block by that in the past…)

.github/workflows/pull_request.yml
```diff
@@ -88,8 +113,27 @@ jobs:
     if: github.event_name == 'pull_request'
     steps:
     - uses: actions/checkout@v4
+      # only run if code changes
+      if: env.run_tests == 'true'
```
alexgarel, line 117: You can't test the variable before you created it !

You have to put this stop under the one where you compute run_tests (that's why your tests are currently failing)

.github/workflows/pull_request.yml
```diff
@@ -88,8 +113,27 @@ jobs:
...
+        else
+          echo "Skipping Check Perl as no code changes detected."
+          echo "run_tests=false" >> $GITHUB_ENV
+        fi
+
+    - name: Exit Successfully if No Code Changes
+      if: env.run_tests == 'false'
+      run: |
+        echo "No relevant code changes detected. Marking as success."
+        exit 0
```
alexgarel, line 135: Only this step will exit… so this step does not bring anything. Your step computing the run_tests variable is already exiting 0, so you don't need this.

alexgarel:
@JagjeevanAK you used the step by step skipping approach, but it seems that skipping the whole job was also ok, so why don't you use it for Perl tests ?

JagjeevanAK:
Hey Alex, I have not completed my work after paper I will complete that.

.github/workflows/pull_request.yml
JagjeevanAK: Hey @alexgarel, I was in a hurry that day due to my exams, so I couldn't complete my work. I understand my mistake now, and I'll make the necessary improvements.

PR #11557, opened 2025-03-07
taxonomy: Make the Misc Facets consistent with their populated tags

teolemon:
### What
taxonomy: Make the Misc Facets consistent with their populated tags


benbenben2 (approved):
LGTM

Issue #11568, opened 2025-03-07
Report product link points to incorrect domain


areebahmeddd:
The "Report this product" link currently uses `nutripatrol.localhost`, which is incorrect. It should be updated to `nutripatrol.openfoodfacts.net` so that users are directed to the correct domain.

areebahmeddd:
Had to close the old PR because my formatter made unnecessary changes.

hangy:
The link is set in `.env` files, so that should just be a configuration issue:
 - Local dev: https://github.com/openfoodfacts/openfoodfacts-server/blob/2df34cc553cfd5f9e8280c1a54458784026fcc6f/.env#L77
 - Other deployments: https://github.com/openfoodfacts/openfoodfacts-server/tree/c4fa8289c722af5cad5811e93d67cd4a01826803/env

areebahmeddd:
I'm not quite sure I follow. If the URL is set in `.env`, why is it still hardcoded in the `href` across the files?

hangy:
> I'm not quite sure I follow. If the URL is set in `.env`, why is it still hardcoded in the `href` across the files?

The HTML files in `tests/integration` are the expected test results that are used to test if the code worked correctly. Therefore, `http://nutripatrol.localhost` should be correct there, for the integration test scenario.

areebahmeddd:
Looks good? @hangy 

Made changes: https://github.com/openfoodfacts/openfoodfacts-server/pull/11570/files

PR #11570, opened 2025-03-07
fix: update report product links to use the correct domain

areebahmeddd:
**Signed-off-by:** Areeb Ahmed [hi@areeb.cloud](mailto:hi@areeb.cloud)  

## Summary of Changes
**Fixes:** #11568  

- Updated incorrect report product links by updating the domain to `nutripatrol.openfoodfacts.net`.

### DIFF
--- a/.env
+++ b/.env
@@ -74,7 +74,7 @@ ODOO_CRM_PASSWORD=
 
 # No need to have it running, it's just to compose a URL
 # you might also use the .net service
-NUTRIPATROL_URL=http://nutripatrol.localhost
+NUTRIPATROL_URL=https://nutripatrol.openfoodfacts.org/
 
 BUILD_CACHE_REPO=openfoodfacts/openfoodfacts-build-cache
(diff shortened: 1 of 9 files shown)


.env
```diff
@@ -74,7 +74,7 @@ ODOO_CRM_PASSWORD=
 
 # No need to have it running, it's just to compose a URL
 # you might also use the .net service
-NUTRIPATROL_URL=http://nutripatrol.localhost
+NUTRIPATROL_URL=https://nutripatrol.openfoodfacts.org/
```
hangy, line 77: I do believe that `http://nutripatrol.localhost` was chosen here intentionally for local development. If there's wrong URL in production, can you show me where? May be able to give better advice then.

areebahmeddd: Umm, what I tried to do was report a product to the moderators, but it redirected me to a dead link instead. That was the issue I was trying to address in the PR.

image

PR #11577, opened 2025-03-09
docs: Add note about regenerating expected results for integration tests.

### LINKED ISSUE #11576: No mention of build_lang_test in how-to-write-and-run-tests.md
**What**

When working on this [PR](https://github.com/openfoodfacts/openfoodfacts-server/pull/11575). I noticed that even after updating expected results for integration tests (using test-int) locally I was failing integration tests while making PR. 

The problem was that translatable strings ( [% msgctxt %] ) were getting rendered as empty string in expected result files instead of default (i.e from en.po).

Solution is to run `make build_lang_test` before regenerating the expected results, so that language files are rebuild in test environment.

Nothingspecial99:
### What

Added info in docs to avoid failing integration tests in workflow when making changes to .pot, .po files.

- Fixes #11576

### DIFF
--- a/docs/dev/how-to-write-and-run-tests.md
+++ b/docs/dev/how-to-write-and-run-tests.md
@@ -94,6 +94,8 @@ If you made a change that affects stored expected results, you can use:
 
 If you regenerate test results, be sure to check carefully that the changes in your commit are expected.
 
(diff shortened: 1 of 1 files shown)


hangy (approved):

stephanegigandet (approved):
Thanks!

PR #11586, opened 2025-03-10
feat: New Category Tree for Open Products Facts

teolemon:
### What
feat: New Category Tree for Open Products Facts

### DIFF
--- a/taxonomies/product/categories-new.txt
+++ b/taxonomies/product/categories-new.txt
@@ -1,0 +1,16824 @@
+# en: Root
+en: Arts & Entertainment
+
+
+<en: Arts & Entertainment
+en: Event Tickets
+
+
+<en: Arts & Entertainment
+en: Hobbies & Creative Arts
+
+
+<en: Hobbies & Creative Arts
+en: Arts & Crafts
+
+
+<en: Arts & Crafts
+en: Art & Craft Kits
+
+
+<en: Art & Craft Kits
+en: Candle Making Kits
+
+
+<en: Art & Craft Kits
+en: Drawing & Painting Kits
+
+
+<en: Art & Craft Kits
+en: Fabric Repair Kits
+
+
+<en: Art & Craft Kits
+en: Incense Making Kits
+
+
+<en: Art & Craft Kits
+en: Jewelry Making Kits
+
+
+<en: Art & Craft Kits
+en: Mosaic Kits
+
+
+<en: Art & Craft Kits
+en: Needlecraft Kits
+
+
+<en: Art & Craft Kits
+en: Scrapbooking & Stamping Kits
+
+
+<en: Art & Craft Kits
+en: Toy Craft Kits
+
+
+<en: Arts & Crafts
+en: Art & Crafting Materials
+
+
+<en: Art & Crafting Materials
+en: Art & Craft Paper
+
+
+<en: Art & Craft Paper
+en: Cardstock & Scrapbooking Paper
+
+
+<en: Art & Craft Paper
+en: Construction Paper
+
+
+<en: Art & Craft Paper
+en: Craft Foil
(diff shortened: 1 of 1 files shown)


teolemon:
@moon-rabbitOFF @aleene @benbenben2 @TaciteOFF @alexgarel @stephanegigandet would love your input on this new taxonomy proposal for Open Products Facts

stephanegigandet:
Quite frankly, I think it's best to use a bottom up product driven approach (like we did for OFF) to create a product taxonomy that is useful for the type of products that currently are in Open Products Facts. If we use an extensive list of categories such as this one, we will have a huge % of categories that won't have any product. Potential downsides:
- we display a category tree with lots of empty categories
- we don't know which categories are worth translating, adding properties, mappings to wikidata etc.
- if we translate all those categories, then we get a huge taxonomy file that needs to be loaded in memory (and slows a bunch of things like automated tests in GitHub)

raphael0202:
For Open Product Fact, I believe it would be really valuable to have an already comprehensive taxonomy to rely on.

We don't have the same community engagement on Open Product Facts than on Open Food Facts, and building this kind of taxonomy from the ground-up takes 1000s of hours of work.

My points of concern are the following:

- what's the licence of this taxonomy? Do we need to credit Google, and if yes how?
- from what I see, some very obvious categories are missing (T-shirt for examples). What is the coverage of this product category taxonomy?

What I like about this taxonomy is that there is already a wikidata mapping available, and that translations are available as well (https://tiago.bio.br/google_product_taxonomy_reference/5010).

While Stephane points are valid, in my opinion we can find a way to accomodate those:
- `we display a category tree with lots of empty categories` -> that's totally valid
- `we don't know which categories are worth translating, adding properties, mappings to wikidata etc.`: we can use other metrics such as the number of products in each category to know those worth working on.
- `if we translate all those categories, then we get a huge taxonomy file that needs to be loaded in memory (and slows a bunch of things like automated tests in GitHub)`: we can maybe use a smaller test taxonomy for the automated testing?

Another middle-ground approach could be to filter this taxonomy to only keep the most popular/common ones.

teolemon:
@raphael0202 @stephanegigandet 

### License
- I'm doing some additional research on the licence. Probably public domain, especially since it's shared among companies
- Wikidata has included it (de facto duplicated it) without any issues: https://www.wikidata.org/wiki/Wikidata:Property_proposal/Google_Product_Taxonomy_ID
- https://www.wikidata.org/wiki/Property:P11302
- Facebook is advising to use it and reference it: https://www.facebook.com/business/help/526764014610932?id=725943027795860
- Shopify as well: https://help.shopify.com/en/manual/online-sales-channels/facebook-instagram-by-meta/product-categories
- Also a lot of great insights on Google Merchant page, we could draw inspiration from in terms of data structure: https://support.google.com/merchants/answer/7052112?hl=en

### Coverage
- The latest version seems to have: "Apparel & Accessories > Clothing > Shirts & Tops"

stephanegigandet:
If we do use the Google products taxonomy, then what's the plan for updating it? Do we still allow contributors to change the taxonomy (e.g. adding entries, adding synonyms, adding properties etc.)

teolemon:
I would be in favor yes, our taxonomies have always been collaborative, and I see it as a seed. 
I don't see the upstream taxonomy evolving that much, and we'll probably be able to update our taxonomies with novel categories as they appear. @raphael0202 WDYT ?

raphael0202:
I agree, we should expect the Google taxonomy to update slowly. We can always perform mapping to Google taxonomy such as what Wikidata does when we add new categories of our own.

teolemon:
Closing in favor of the better: https://github.com/openfoodfacts/openfoodfacts-server/pull/12457

PR #11587, opened 2025-03-10
taxonomy: Add “Från Sverige” sub-labels

Freso:
### What

Kött från Sverige logo from:
https://fransverige.se/produkt/kott-fran-sverige-market-1080x1080/
Mjölk från Sverige logo from:
https://fransverige.se/produkt/mjolk-fran-sverige-market/

Both logos were processed by `oxipng`[0] version 9.1.4.

[0] `oxipng --opt max --strip safe --fix --zopfli`


taxonomies/labels.txt
```diff
@@ -3727,6 +3727,20 @@ label_categories:en: en:Geographic label
 origins:en: en:sweden
 web:sv: https://fransverige.se/
 
+< sv:Från Sverige
+xx: Kött från Sverige
```
stephanegigandet, line 3731: Could you also add the en: translations?

Freso: I mean, the translation would be “Meat from Sweden”, but there’s no “Meat from Sweden” label (that I know of), so even in English it should be displayed as “Kött från Sverige”, no?

Even in the English blurb on https://fransverige.se/other-languages/ it gives the labels using the Swedish names (emphasis mine):
> When you choose **‘Från Sverige’** (From Sweden) labelled raw materials, food, and plants, you get products loaded with added Swedish value. […] Both small and large companies label their products with **‘Från Sverige’** (From Sweden), **‘Kött från Sverige’** (Meat from Sweden), and **‘Mjölk från Sverige’** (Milk from Sweden) – making it easy to choose Swedish-produced goods. […]

stephanegigandet: That's what we do for many other labels that have logos, like "Produit en Bretagne", we make "en: Produced in Brittany" the canonical entry name. We usually do this for all labels that have a name like "Made in [country]", or "[ingredient] from [country]", things like "British Eggs" etc.  It makes more sense for people who don't know the language.

stephanegigandet (approved):

Issue #11277, opened 2025-01-22
Error: environmental_score_details.tt.html: not found


g123k:
Hello,

I've noticed a bug on this page: https://world.openfoodfacts.org/cgi/product_image.pl?code=8033028910213&id=front_en
There's this error at the end:
```
file error - web/pages/product/includes/environmental_score_details.tt.html: not found
```

<img width="1122" alt="Image" src="https://github.com/user-attachments/assets/3f3be9f2-da46-482a-89cb-a1df8286ae97" />

AdityaP700:
Well is anyone working in this issue ? It looks a concerning bug to me which needs to be fixed . Well if no one is assigned ,Can i take the lead of trying and fixing it out.?

JagjeevanAK:
Hey, @g123k Could you kindly provide an example that reproduces this issue?

g123k:
> Hey, [@g123k](https://github.com/g123k?rgh-link-date=2025-02-28T22%3A10%3A03.000Z) Could you kindly provide an example that reproduces this issue?

Mmm what's the link in the description? 🧐

JagjeevanAK:
> > Hey, [@g123k](https://github.com/g123k?rgh-link-date=2025-02-28T22%3A10%3A03.000Z) Could you kindly provide an example that reproduces this issue?
> 
> Mmm what's the link in the description? 🧐

Hey I have figured out the issue but what is was saying that only by clicking the link i am seeing that issue  and if  I directly go to website then how I can see without clicking the link. what are the steps to recreate.

g123k:
Ah ok, just open an image

JagjeevanAK:
The issue has finally been resolved, @g123k! You can check it out once the new version is released.

PR #11591, opened 2025-03-11
fix: Missing Template Error: Update ecoscore_details.tt.html to environmental_score_details.tt.html 

JagjeevanAK:
### What

This PR updates references to `ecoscore_details.tt.html`, which was **renamed** as part of an earlier commit [`#11142`](https://github.com/openfoodfacts/openfoodfacts-server/pull/11142), to the new file name `environmental_score_details.tt.html`.  

#### **Issue Fixed**  
The website was displaying an error on this page:  https://world.openfoodfacts.org/cgi/product_image.pl?code=8033028910213&id=front_en
```
file error - web/pages/product/includes/environmental_score_details.tt.html: not found
```
This happened because:  
- The `ecoscore_details.tt.html` template was renamed to `environmental_score_details.tt.html` as part of the Eco-Score transition.  
- Some parts of the code (Perl controllers, config files, or dynamic includes) were still trying to reference the old filename.  

#### **Changes Made**  
- Updated all occurrences of `ecoscore_details.tt.html` to `environmental_score_details.tt.html` in:  
  - Template files (`.tt.html`)  
- Also checked below file for occurrences of `ecoscore_details.tt.html` but there was no such thing present,.
  - Perl controllers (`.pm`)  
  - Configurations (`.conf`, `.json`, `.yml`, etc.)  
- Verified that the `[% INCLUDE ... %]` statements correctly reference the new template.  
- Restarted the web server locally to confirm the fix.  

#### **Feature required changes**  
- If any frontend dependencies reference `ecoscore_details.tt.html`, they may need updates as well.  
- This PR does **not** rename database fields or API endpoints—only the template reference.  

---

This should be good for submission! Let me know if you need refinements. 🚀

``

#### Original 
Screenshot 2025-03-01 at 11 59 14 AM

#### Updated change
Screenshot 2025-03-01 at 12 02 20 PM

#11142: *Rename Eco-Score fields to Environmental Score*  

I above issue there was about renaming of  ecoscore to environmental_score by [stephanegigandet](https://github.com/stephanegigandet) but file names in templates was not made.
- Fixes [#11277](https://github.com/openfoodfacts/openfoodfacts-server/issues/11277)

If there is an issue in my Pull request then I am open to make necessary changes to it.


JagjeevanAK:
Hey @alexgarel and @stephanegigandet ,
Do we require the api end point `https://staging-ecobalyse.incubateur.net/api/food` because it is causing problems of timeout and which is causing to fail my test case.

### This are my finding 
**1. My error**
Screenshot 2025-03-11 at 6 31 41 PM
  
**2. Root cause of error**
Screenshot 2025-03-11 at 6 30 49 PM
  
**3.API endpoint which is mentioned does not exist and returns the error**
Screenshot 2025-03-11 at 6 31 00 PM  
<br>
**If we require this api then we need to update the things according to it as there [docs](https://ecobalyse.beta.gouv.fr/#/api) suggest different endpoints.**

tests/integration/expected_test_results/web_html/fr-search-form.html
```diff
@@ -2027,7 +2027,7 @@ <h3>Résultats sur un graphique</h3>
     </form>
 </div>
 
-<!-- end templates/web/pages/search_form/search_form.tt.html --><p>file error - web/pages/product/includes/environmental_score_details.tt.html: not found</p>
+<!-- end templates/web/pages/search_form/search_form.tt.html --><p></p>
```
stephanegigandet, line 2030: What I don't understand is why the search form somehow loads the environmental_score_details.tt.html template: it should be only on product pages. So your fix definitely get rids of the message, but I don't understand the root cause of the issue. We probably should not load this template for this page, and we should not have this empty <p></p> at the end.

JagjeevanAK: > So your fix definitely get rids of the message, but I don't understand the root cause of the issue.

Thanks for the change you suggested to test file As I want to tell you that is you last update failed to update the file name from ecoscore .. to environmental ... with other changes that where made for this feature.  
So some added `file error - web/pages/product/includes/environmental_score_details.tt.html: not found` in test file as you can see to overcome the issue for short while instead of going deep into it to solve the main problem.  

<br>
I don't know about why env score is added on search page rather than keep it only specific to product page need to understand form the person who has integrated this feature on search page. If not then we need to create a new Issue to solve this problem or to remove this form the search page and make specific to product page.

stephanegigandet (approved):
Thank you!

PR #11593, opened 2025-03-11
fix: no_cache + no_off_query

stephanegigandet:
This PR:
- disables Memcached for results on the pro platform
- makes no_cache only affect Memcached: no_cache=1 queries can now go to off-query
- adds a new no_off_query parameter to indicate we want results from MongoDB


john-gom (approved):
The options are getting a bit confusing but hopefully we can simplify once the most recent changes have been fully tested

PR #11596, opened 2025-03-11
fix: require user identification in API v2 WRITE requests

stephanegigandet:
In the documentation, we state that API WRITE queries must be done with an authenticated user, but in practice it was not checked.


cgi/product_jqm_multilingual.pl
```diff
@@ -101,7 +101,12 @@ =head1 DESCRIPTION
 
 $code = normalize_code($code);
 
-if (not is_valid_code($code)) {
+if (not defined $User_id) {
+	$log->info("no user credentials", {code => $code, original_code => $original_code}) if $log->is_info();
```
hangy, line 105: Should we return a `401 Unauthorized` HTTP status code or might that break clients?

stephanegigandet: I'm not sure, so for V2 I'm keeping the 200 HTTP status code.

stephanegigandet:
Note: I hope I repaired the tests. A few of them were broken because we were creating test products without a logged in user, or because the user creation (and thus login) failed because we were not removing existing users, and the test failed because the user was already created.

teolemon:
- This is a big one @stephanegigandet @hangy 
- The mobile team needs to be informed @monsieurtanuki @g123k @primaelquemerais 
- For 3rd party app, and the legacy Android apps, it's potentially Russian Roulette. We used to let people add fields when they did not exist, not edit them if they exist.

cgi/product_jqm_multilingual.pl
hangy: I'm just wondering if some clients might give users a false sense of their data being saved, because they don't really check the response status, just the HTTP response code. But it's definitely a breaking change, so needs to be considered carefully.

monsieurtanuki:
@stephanegigandet For the record in Smoothie and off-dart I'm not sure we qualify for "logged in" users.
Basically we call `/cgi/auth.pl` at the init of the app but it's more like a password check.
And for each WRITE operation, we pass user data, but it's not like a cookie or a bearer token.

teolemon:
@monsieurtanuki it's also the case of product addition for non-logged in users.

monsieurtanuki:
> @monsieurtanuki it's also the case of product addition for non-logged in users.

@teolemon Non logged-in users are considered logged-in as `smoothie-app` for WRITE operations.

tests/integration/api_v2_product_write.t
```diff
@@ -25,7 +25,7 @@ create_user($ua, \%create_user_args);
 # Note: expected results are stored in json files, see execute_api_tests
 my $tests_ref = [
 	{
```
alexgarel, line 27:
```suggestion
	# Test anonymous contribution, this should fail
	{
```

alexgarel (approved):
Great.
Just a few nitpicks and comments.

teolemon:
@monsieurtanuki What about the dart package, do we have a failover user ? Or does the package prevent you from doing unauthed queries ?

monsieurtanuki:
> @monsieurtanuki What about the dart package, do we have a failover user ? Or does the package prevent you from doing unauthed queries ?

The `User` parameter is mandatory for both off-dart "save product" methods (normal `/cgi/product_jqm2.pl` and packagings only `/api/v3/product/$barcode`). It will add fields like `user_id` and `password` to the http query.

package-lock.json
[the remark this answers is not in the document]
stephanegigandet: it's not intended in fact, I don't know why that file changed

stephanegigandet:
> @stephanegigandet For the record in Smoothie and off-dart I'm not sure we qualify for "logged in" users. Basically we call `/cgi/auth.pl` at the init of the app but it's more like a password check. And for each WRITE operation, we pass user data, but it's not like a cookie or a bearer token.

if you pass userid and password, or a valid session cookie, then the user is authenticated, so we should be fine

PR #11597, opened 2025-03-11
fix: allow all logged in users to change product_type

stephanegigandet:
Previously only moderators could change the product_type.

This PR also tries to generalize a bit the use of Permissions.pm / has_permission(), so that we keep them in one place.


hangy (approved):
Great idea!

PR #11606, opened 2025-03-12
feat: taxonomize brands with a language less xx: taxonomy

stephanegigandet:
This PR is intended to replace https://github.com/openfoodfacts/openfoodfacts-server/pull/10662 which got broken with the merges.

This PR is a proof of concept to make brands a taxonomized field using a language less taxonomy (all entries use the xx: prefix).

To taxonomize existing products:

~/openfoodfacts-server$ docker exec -it po_off-backend-1 bash
www-data@a01d17ec9489:/opt/product-opener$ ./scripts/update_all_products.pl --fields brands 
Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/5208

This PR is a proof of concept, it lacks tests, documentation etc.

Also the "brands" taxonomy is hardcoded in Tags.pm, it would probably be better to make this configurable in case we want to do something similar with other taxonomies in the future.

Read API requests should still get the same values in the "brands" field (but there will be new fields brands_tags and brands_hierarchy).

Write API requests should continue to use the "brands" field to write brands.


teolemon:
@stephanegigandet 
tests/unit/tags.t: Begin Error Output Stream
tests/unit/tags.t: assertion failure: '--assert-tidy' is set but output differs from input
tests/unit/tags.t: 
tests/unit/tags.t: Input   file length has 36469 chars in 917 lines
tests/unit/tags.t: Output  file length has 36470 chars in 91[8](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/13817106065/job/38653854037?pr=11606#step:10:9) lines
tests/unit/tags.t: Files first differ at character 85 of line 244
tests/unit/tags.t:  243:is get_taxonomyid("en", "some brand"), "some-brand";
tests/unit/tags.t: <244:...s", "some brand, xx:some-other-brand")], ['xx:some brand', 'xx:some-o...
tests/unit/tags.t: >244:...s", "some brand, xx:some-other-brand")],
tests/unit/tags.t:                                                 ^
tests/unit/tags.t:

benbenben2 (approved):
LGTM

lib/ProductOpener/Tags.pm
```diff
@@ -2584,6 +2584,8 @@ sub generate_tags_taxonomy_extract ($tagtype, $tags_ref, $options_ref, $lcs_ref)
 
 sub retrieve_tags_taxonomy ($tagtype, $die_if_taxonomy_cannot_be_loaded = 0) {
 
+	print STDERR "retrieve_tags_taxonomy - tagtype: $tagtype\n";
```
jayaddison, line 2587: @JagjeevanAK @stephanegigandet this `print` statements seems to emit a large number of messages to stderr during build logs.  Are we using that to help locate a problem, or could/should we remove it?

jayaddison: I've filed #11765 as a cleanup task to remove the print statement; please let me know if you have any objections.

JagjeevanAK: Sorry @jayaddison, I have not seen at this comment.

jayaddison: No problem @JagjeevanAK - thanks for letting me know that you found it :)

jayaddison: The cleanup will be applied by #11765 (only in CI so far, not yet released) - so please let me know if you have any concerns about that.

PR #11607, opened 2025-03-12
fix: TaxonomiesEnhancer division by zero

stephanegigandet:
We had an issue to publish products from the Systeme U import, it took quite a bit of time to debug, so I wrote a little doc of what I did (maybe not  the best way to do it). It turns out to be a small issue in the logging statements of TaxonomyEnhancer.pm.

In Perl $#array is the indice of the last member of the array, which starts at 0. So if the array has 1 element, its value is 0. So on a specific product we got a division by zero.

minion_log.conf has logging at the debug level, so it may be a reason why we got this error in the minion tasks and that we didn't get division by zero (or at least we didn't see them) in the website / API.

### DIFF
--- a/docs/dev/how-to-debug-minion-import-and-export-tasks.md
+++ b/docs/dev/how-to-debug-minion-import-and-export-tasks.md
@@ -1,0 +1,102 @@
+# How to debug Minion import and export tasks
+
+For imports on the pro platform, and exports from the pro platform and imports on the public platform, we use Minion tasks.
+
+Those tasks are run by a Minion worker service, which is configured in /etc/systemd/system/minion\@off.service
+
+## Log files
+
+in the logs directory, we have:
+- minion.log - states which tasks are started
+- minion_log4perl.log - debug messages from the Perl code that implements the Minion tasks
+
+## Checking the status of Minion tasks (jobs)
+
+````
+off@off:/srv/off$ (off) ./scripts/minion_producers.pl minion job 
+[..]
+132132  inactive  pro.openfoodfacts.org  update_export_status_for_csv_file
+132131  failed    openfoodfacts.org      import_csv_file
+132130  finished  pro.openfoodfacts.org  export_csv_file
+````
+
+## Debugging failed import of pro platform data on public platform
+
+The following are notes on how a specific issue was debugged in production:
+
+We now are loading automatically Systeme U (a big French retailer) imports in the producers platform.
+
+The checkbox in the Systeme U organization (admin field) to automatically export new products is checked, but the products have not been loaded on the producers platform.
+
+We can launch the export and import manually:
+
+off@off-pro:/srv/off-pro$ (off-pro) scripts/export_and_import_to_public_database.pl --query states_tags=en:to-be-exported --owner org-systeme-u
+
+And we can see the details of the corresponding minion jobs:
+
+````
+off@off-pro:/srv/off-pro$ (off-pro) ./scripts/minion_producers.pl minion job
+[..]
+132132  inactive  pro.openfoodfacts.org  update_export_status_for_csv_file
+132131  failed    openfoodfacts.org      import_csv_file
+132130  finished  pro.openfoodfacts.org  export_csv_file
+````
+
+In the off container, there is little useful information in the logs:
+
+/srv/off/logs/minion.log:
+
+````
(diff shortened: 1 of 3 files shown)


benbenben2 (approved):
Thanks @stephanegigandet

docs/dev/how-to-debug-minion-import-and-export-tasks.md
```diff
@@ -0,0 +1,102 @@
...
+
+/srv/off/logs/minion.log:
+
+````
+import_csv_file_task - job: 132131 started - args: {"comment":"Import from producers platform","csv_file":"/srv/off-pro/export_files/org-systeme-u/export.1741792171.exported.csv","global_values":{"data_sources":"Producers, Producer - systeme-u"},"query":{"owner":"org-systeme-u","states_tags":"en:to-be-exported","data_quality_errors_producers_tags.0":{"$exists":false},"code":"3256221408515"},"export_job_id":132130,"source_id":"org-systeme-u","manufacturer":1,"org_id":"systeme-u","export_id":1741792171,"include_images_paths":1,"user_id":"org-systeme-u","source_name":"systeme-u","include_obsolete_products":1,"exported_t":1741792171,"owner_id":"org-systeme-u"}
+````
+
+To get more data (debug level) in minion_log4perl.log:
+
+W can stop the minion daemon
```
alexgarel, line 55:
```suggestion
We can stop the minion daemon
```

docs/dev/how-to-debug-minion-import-and-export-tasks.md
```diff
@@ -0,0 +1,102 @@
...
+
+as the root user:
```
alexgarel, line 57: A command is missing here !

```suggestion
as the root user: `systemctl stop minion@off.service`
```

docs/dev/how-to-debug-minion-import-and-export-tasks.md
```diff
@@ -0,0 +1,102 @@
...
+
+And run it manually as a normal process
+
+as the off user:
+
+source env/setenv off
+
+off@off:/srv/off/logs$ (off) TAP_LOG_FILTER=none perl scripts/minion_producers.pl minion worker -m production
```
alexgarel, line 65: Using code block can make it more readable. Also I think using sudo is more explicit as telling user:

````suggestion
as the off user:
```bash
sudo -u off bash
source env/setenv off
TAP_LOG_FILTER=none perl scripts/minion_producers.pl minion worker -m production
```
````

Comment (alexgarel) [inline docs/dev/how-to-debug-minion-import-and-export-tasks.md:65]: can you explain the TAP_LOG_FILTER=none  ?

File: docs/dev/how-to-debug-minion-import-and-export-tasks.md [hunk 4]
```diff
@@ -0,0 +1,102 @@
...
+
+I added a print STDERR in Import.pm to see if a specific product is causing the problem:
+
+Import.pm - org: systeme-u - code: 3256221408515
```
alexgarel, line 69:
````suggestion
```
Import.pm - org: systeme-u - code: 3256221408515
```
````

File: docs/dev/how-to-debug-minion-import-and-export-tasks.md [hunk 5]
```diff
@@ -0,0 +1,102 @@
...
+132132  inactive  pro.openfoodfacts.org  update_export_status_for_csv_file
+132131  failed    openfoodfacts.org      import_csv_file
+132130  finished  pro.openfoodfacts.org  export_csv_file
+````
+
+In the off container, there is little useful information in the logs:
+
+/srv/off/logs/minion.log:
+
+````
```
alexgarel, line 49: Note you should normally use 3 backtick 
the 4 backtick is reserved for the case where you must include 3 backtick inside a code block

docs/dev/how-to-debug-minion-import-and-export-tasks.md
```diff
@@ -0,0 +1,102 @@
...
+                $log->debug(
+                        "check_ingredients_between_languages > detect_missing_stop_words_after_list -   too much difference between languages to raise warning. diff/total > tolerance: $translation_difference_count / $#$ingredients1 = "
+                                . $translation_difference_count / $#$ingredients1 . " > "
+                                . $translation_difference_accepted_percentage)
+                        if $log->is_debug();
+````
+
+$#array is equal to 0 when @array contains 1 element, so it makes a division by 0 error and the task fails. But unfortunately I could not find any log where this division by zero error was reported, the only thing we get from Minion is that the task failed...
```
alexgarel, line 95:
```suggestion
`$#array` is equal to `0` when `@array` contains 1 element, so it makes a `division by 0 error` and the task fails. But unfortunately I could not find any log where this division by zero error was reported, the only thing we get from Minion is that the task failed...
```

docs/dev/how-to-debug-minion-import-and-export-tasks.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+
+
+
+
+
```
alexgarel, line 102: I would remember to re-start the minion

```suggestion

**Finally:** don't forget to **restart the minion service** !

```bash
sudo systemctl start minion@off.service
```
```

stephanegigandet:
Thanks for the detailed review and suggestions @alexgarel !

docs/dev/how-to-debug-minion-import-and-export-tasks.md
stephanegigandet: Thanks

docs/dev/how-to-debug-minion-import-and-export-tasks.md
stephanegigandet: done

PR #11609, opened 2025-03-13
fix: remove the ugly bg color for the Folksonomy Engine

teolemon:
### What
fix: remove the ugly bg color for the Folksonomy Engine


CharlesNepote (approved):
The background color was there to highlight it's still beta, but we can remove it if you prefer so.

PR #11611, opened 2025-03-13
docs: ingredients_percent_analysis

alexgarel:


stephanegigandet (approved):
Thanks

stephanegigandet:
The test failed because Ecobalyse is not responding, we will need to mock the response for the test

PR #11617, opened 2025-03-14
fix: fixed the issue of pull_request.yml file

JagjeevanAK:
### What
This was omitted from the previous commit because Now as believe it is unnecessary to build the image every time when only the documentation is updated.

- Fixes #11555


alexgarel (approved):
Thanks @JagjeevanAK !

That's always tricky to spot that kind of logical mistake.

PR #11620, opened 2025-03-14
ci: roll-back conditional runs for now

alexgarel:
Until we find the right logic to do that.

This rollback some changes from #11555 and #11617

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -18,6 +18,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # FIXME: not used yet
   filter:
     name: "Filter changed paths"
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -30,6 +31,9 @@ jobs:
       - uses: dorny/paths-filter@v2
         id: filter
         with:
+          # FIXME: we need to consider more than these files
```
JagjeevanAK, line 34: Ya you are correct I have also came across that observations form this situation. will work on that and update quickly.

JagjeevanAK (approved):
I have reviewed your request and I think You have removed all the necessary thing from the yml file to go back to previous state.

PR #11621, opened 2025-03-14
docs: add precision to direnv doc for mac users

nmetifiot:
### What

Complete dirent docs for mac users


JagjeevanAK (commented):
It's not that important, but you make a good point. Many people know which shell they're using and how to configure it. Besides Bash and Zsh, a lot of users prefer Fish or other shells. I think we should create a more generalized version of this.

JagjeevanAK:
> (updated to see which tests run after the merging of #11624 )

take a look of #11628

stephanegigandet (approved):
Thanks

stephanegigandet:
@nmetifiot are you still working on this PR, or should we merge it as-is now?

nmetifiot:
@JagjeevanAK I pushed a new commit, I've deleted the bash doc to give more importance to the official direnv documentation. Finally I've just add a bullet point to be careful at the used shell.

stephanegigandet (approved):
Thanks

PR #11624, opened 2025-03-15
ci: Updated the workflow file with refined logic 

JagjeevanAK:
### What
Updates have been made to the workflow file, as per suggestions from @stephanegigandet and @alexgarel, which were discussed in Slack.


stephanegigandet (approved):
Looks good, thank you

PR #11630, opened 2025-03-17
ci: replaced dorny/path-filter to tj-actions/changed-files

JagjeevanAK:
Hey @stephanegigandet and @alexgarel, we finally made it! 🎉    

I've tested this solution in my local repo, so there won't be any further excuses. I just want to clarify that the issue last time wasn't with the merge file solution — that was correct. The real problem was with the `dorny/path-filter` action, which turned out to be buggy.    

I've now switched to `tj-actions/changed-files`, which works perfectly and is well-maintained. ✅

### DIFF
--- a/.github/scripts/path-filter.sh
+++ b/.github/scripts/path-filter.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+set -e
+# Check each changed file
+for file in $@; do
+  if ! echo "$file" | grep -E -q '\.md$|^docs/'; then
+    echo "code_modified=true" >> "$GITHUB_OUTPUT"
+    exit 0
+  fi
+done
+
+# If only Markdown or docs files are changed
+echo "code_modified=false" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -23,7 +23,7 @@ jobs:
     name: "Filter changed paths"
     runs-on: ubuntu-latest
     outputs:
-      code_modified: ${{ steps.filter.outputs.code }}
+      code_modified: ${{ steps.filter.outputs.code_modified }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -25,17 +25,16 @@ jobs:
     outputs:
       code_modified: ${{ steps.filter.outputs.code }}
     steps:
-      - uses: dorny/paths-filter@v2
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Get changed files
+        uses: tj-actions/changed-files@v45
+        id: changed_files
+
+      - name: Filter non-markdown and non-docs files
         id: filter
-        with:
-          # FIXME: we need to consider more than these files
-          # We need to consider every file in the repository and have it in a category
-          # Maybe using negative match: https://github.com/micromatch/picomatch?tab=readme-ov-file#advanced-globbing
-          filters: |
-            code:
-            - '**/*'       # Catch everything else
-            - '!**/*.md'   # Exclude all Markdown files
-            - '!docs/**'   # Exclude docs/ folder
+        run: ./.github/scripts/path-filter.sh "${{ steps.changed_files.outputs.all_changed_files }}"
 
   lint:
     name: 🕵️‍♀️ NPM lint


stephanegigandet (approved):
Thanks

stephanegigandet:
Seems to work well :)

I refreshed https://github.com/openfoodfacts/openfoodfacts-server/pull/11621 after merging the PR, and the tests were skipped:

image

JagjeevanAK:
Hey @stephanegigandet, I tested everything on my personal repo yesterday, and it should work in all aspects. If there's an extreme case where it fails for that, I'm here to resolve it.

jayaddison:
@JagjeevanAK @alexgarel @stephanegigandet if I understand correctly, the reason for this is that it is not possible to use the GitHub Actions workflows `paths` and `paths-ignored` feature in combination with GitHub Actions CI Required Status (in other words: require success) checks, as described in the comment here: https://github.com/openfoodfacts/openfoodfacts-server/blob/db8daa9ac0230eea1b92024dc5819cd76b20c65b/.github/workflows/pull_request.yml#L5-L11

**However**: previously there [was a GitHub-documented](https://github.com/orgs/community/discussions/49124#discussioncomment-5209069) workaround for that -- the advice has disappeared from GitHub's documentation some time since Y2023 (now the page redirects to one that does not mention it).

The Y2023 copy (that the poster in the linked discussion was referring to when they posted) can today be found at: https://web.archive.org/web/20230321195500/https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks

The workaround was to create one `paths-ignore: x` entry, and then an inverted separate workflow -- with a job with the same name -- that has `paths: x` that `always()` succeeds.

Is that workaround not possible any more?

jayaddison:
> **However**: previously there [was a GitHub-documented](https://github.com/orgs/community/discussions/49124#discussioncomment-5209069) workaround for that -- the advice has disappeared from GitHub's documentation some time since Y2023 (now the page redirects to one that does not mention it).
...
> The workaround was to create one `paths-ignore: x` entry, and then an inverted separate workflow -- with a job with the same name -- that has `paths: x` that `always()` succeeds.
> 
> Is that workaround not possible any more?

I have begun testing this in openculinary/openfoodfacts-server#1.  So far, if my experiments are correct, it seems that the workaround remains feasible.

JagjeevanAK:
I've never heard of a workaround, but what I mean is that the actions I've used are widely adopted and actively maintained. If you check their repositories, you'll see their high level of activity.  

Additionally, we're currently using `tj-actions/changed-files` to skip unnecessary Docker builds and optimize our existing CI operations. So in some way they are helping more than a one thing.

jayaddison:
The workaround - previously a documented solution according to GitHub - would allow us to enable `paths` and `paths-ignore`, meaning that we do not need to use a third-party action.  The `tj-actions/changed-files` action experienced a security incident recently, so it may be worth being cautious about it.

jayaddison:
As a followup to my previous comments: it seems the built-in `paths` feature from GitHub Actions itself is not sufficient here; we want the ability to filter individual job steps.

I'm trying to work out what the blockers for usage of `dorny/paths-filter` were; I've learned from discussion in #11644 that it is due to something about that action not behaving as documented.  My first guess is that it is this comment:

https://github.com/dorny/paths-filter/blob/9b7572ffb2cddcbb0397d8a240c253352a845140/README.md?plain=1#L205

```
    steps:
    # For pull requests it's not necessary to checkout the code
    - uses: dorny/paths-filter@v2
```

Although that may have been true at the time it was written, I don't think it remains true today.  I believe that GHA workflows activated by the `pull_request` event _do_ need to perform a checkout for subsequent steps to be able to use the content of the repo.

I will begin experimenting with using `dorny/paths-filter` in a fork, using pull requests, and testing what the behaviour is with/without an `actions/checkout` step beforehand.

jayaddison:
> I will begin experimenting with using `dorny/paths-filter` in a fork, using pull requests, and testing what the behaviour is with/without an `actions/checkout` step beforehand.

The logic expressed in #11555 seems to neatly express what we want, and it uses the `dorny/paths-filter` action; that's approximately the baseline/GHA workflow definition that I will commence testing from.

JagjeevanAK:
Glob characters are not working as they are intended to work as given in doc's.

JagjeevanAK:
Mate #11555 has some logic error's so we reverted it back and then I have solved that problem by creating another PR. you can take look at it.

jayaddison:
Ok, cool, I've looked at those PRs.  Was it the `!` (not) globs that were the problem?  Files appearing modified even with those added?

jayaddison:
Again: I should answer my own question!  I'll use a fork to check.

jayaddison:
Ok, I think I have confirmed: the [`predicate-quantifier` setting in `dorny/paths-filter`](https://github.com/dorny/paths-filter/blob/de90cc6fb38fc0963ad72b210f1f284cd68cea36/README.md?plain=1#L157-L171) does not appear to function as documented.  Even after enabling that, a `!` blob to ignore some files _did not_ result in the filter having a false value when those files were modified.

For anyone curious, my testing for this was two comparative pull requests at:

  - https://github.com/openculinary/openfoodfacts-server/pull/4
  - https://github.com/openculinary/openfoodfacts-server/pull/5

jayaddison:
Hold on!  We were using `dorny/paths-filter@v2`!

The `predicate-quantifier` feature only arrived for `dorny/paths-filter@v3`: https://github.com/dorny/paths-filter/commit/cf89abdbaede7b5bcbc8e105da2aaa39fd84b997

jayaddison:
@JagjeevanAK if the problem in #11555 was that the ignore blobs (`!...`) did not work, then I think that two small fixups could resolve that:

  * Upgrade from `dorny/paths-action` version `v2` to `v3`.
  * Configure `predicate-quantifier: 'every'` on the `filter` rule -- so that _all_ of the blobs in it have to match.

I've been testing that using the following fork PR, and it was successful:

  * https://github.com/openculinary/openfoodfacts-server/pull/6

However, if the problem was something else -- let me know and I can investigate.

PR #11636, opened 2025-03-19
fix: update tests to fix import_systemeu.t test

stephanegigandet:
update tests, needed because the original import_systemeu PR had test passing before https://github.com/openfoodfacts/openfoodfacts-server/pull/11606 was merged .

The brands taxonomy changes the order of the brands in the brands field


teolemon (approved):
thanks @stephanegigandet

PR #11642, opened 2025-03-20
docs: Documentation for 'last_updated_t'.

jforget:
### docs: Documentation for 'last_updated_t'.

When I found that many documents contained an undocumented `last_updated_t` property,
I first thought it was a typo for `last_modified_t`. But after reading
[issue 9646](https://github.com/openfoodfacts/openfoodfacts-server/pull/9646),
[the change log](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/CHANGELOG.md)
and the 
[message](https://github.com/openfoodfacts/openfoodfacts-server/compare/v2.25.0...v2.26.0).
for tag 2.26.0,
I understood this is a real property.
So here is its documentation.

Some other properties need to be documented, but I do not know what to say
about them and in which file the documentation should be written. These
properties are:

* packaging_materials_tags

* packaging_recycling_tags

* packaging_shapes_tags

* packagings_materials

### DIFF
--- a/docs/api/ref/schemas/product_meta.yaml
+++ b/docs/api/ref/schemas/product_meta.yaml
@@ -67,6 +67,14 @@ properties:
     type: integer
     description: |
       Date when the product page was last modified.
+      This date is updated only when primary data is modified (data entered by the user or updated by an interface)
+  last_updated_t:
+    type: integer
+    description: |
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Thanks!

PR #11644, opened 2025-03-21
fix: replace tj-actions/changed-file by step-security/changed-files

JagjeevanAK:
`tj-action/changed-files` is compromised and I have replaced it with the best alternative of it.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -29,7 +29,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Get changed files
-        uses: tj-actions/changed-files@v45
+        uses: step-security/changed-files@v45
         id: changed_files
 
       - name: Filter non-markdown and non-docs files


jayaddison:
I think we might be able to use the `paths` filter that is built-in to GitHub Actions to achieve what `changed-files` was doing for us.  It would be less code, and that could be good: it means less code to review and that could have bugs / problems in future.

The tricky part is that this repository uses GitHub Actions required status checks.  However: I think it may be possible to have an inverted workflow that has `paths: *` where the checks always succeed (`exit 0`).  GitHub will (I think!) still prevent a PR from being merged if a check from the more specific path filter (for example, `paths: *.md`) fails.

github-throwaway:
Context:

https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised

JagjeevanAK:
Malte he knew it that was compromised and he was the first-one who updated me.

jayaddison:
@JagjeevanAK do you think we could implement the required functionality without `changed-files`?  I guess we want to handle multiple cases: changed documentation, changed Docker files, ... maybe other filetypes too?

That would be more complicated than what I've tested so far, so I'm not sure.

JagjeevanAK:
I believe you are not seeing the broader picture yet. While we can certainly test just for docs, but our main goal is to avoid rebuilding the Docker container each time. To achieve this, we need to monitor which files have been modified. I've been experimenting with this on another branch/ different pull request.

jayaddison:
Understood.  Would the GitHub Actions `paths` filter be usable to detect those docker-related changes, or are there reasons that that would not meet the requirements?

alexgarel (approved):
Thanks.

jayaddison:
@JagjeevanAK If/when you have time, could you let me know the reasons/bugs that made `dorny/paths-filter` no longer suitable?

I know my questions probably seem very repetitive; I'm trying to understand why it made sense to make the change, mainly because I think we got lucky avoiding the compromise, and also because to me, `paths-filter` seems to have fewer and smaller dependencies (so, if I were choosing between two otherwise similar components, I would probably choose the smaller/simpler one).

JagjeevanAK:
Hey @jayaddison, there is nothing called **repetitive** —we are all learning day by day. The reason behind shifting from `dorny/paths-filter` was that it was not working as it was supposed to or as described in the docs. There were some errors and inconsistencies. Though it was lightweight and faster than `tj-action/changed-files`, I thought it was a good time to move on to the second-best option.

github-throwaway:
@jayaddison the beauty of open source is you can always try for yourself. If you want to test a different implementation yourself, go for it. That's the best way to learn ☺️

jayaddison:
Thanks @JagjeevanAK @github-throwaway.  I'd like to test it, and am willing to; I'm not sure what behaviour(s) to check for, though (I have begun searching through the `paths-filter` issue tracker to try to guess at things that might not work as documented).

jayaddison:
> Thanks @JagjeevanAK @github-throwaway. I'd like to test it, and am willing to; I'm not sure what behaviour(s) to check for, though (I have begun searching through the `paths-filter` issue tracker to try to guess at things that might not work as documented).

I'll add any additional relevant notes in #11630, because that seems to be the more relevant location.  I have one or two ideas to examine and test.

PR #11623, opened 2025-03-15
fix: Producers on large-screen

Pawan8538:
### What

- The Producers platform is now visible on large screens.
- The GET THE APP button no longer collapses.
- The OPENFOODFACTS logo no longer shrinks.

### Fixed Video

https://github.com/user-attachments/assets/f0242fd4-efd6-441e-a3e3-32c109a31cec
 
### Related issue(s) and discussion

- Fixes [#11461]

@alexfauquette @teolemon Any suggestions, or are the above changes satisfactory?

PR #11651, opened 2025-03-23
feat: Keep the link to the producer platform visible

### LINKED ISSUE #11461: Link to Producer platform not visible on small screen
### Problem

The link to "Producteur" disappears with a small screen. A producer with a 13" mac book will never see this link

https://github.com/user-attachments/assets/ce3c1b8f-e940-41e6-b299-5ee7c52afcc5

@manon-corneille 

### Proposed solution

This is due to the `class="show-for-xlarge-up"` which could be removed.

https://github.com/user-attachments/assets/66179333-d366-4538-8027-27ef6769b7b4

The drawback is the logo OFF that shrinks. Should it be replaced by the small one.

### Code pointers

site_layout.tt.html

alexfauquette:
Alternative to #11623

### What

The idea is to gain pace by replacing the long logo by a short one instead of removing the "producer" link

I noticed a weird behavior. The class `.expanded` seem to be related to a which between the top bar is divided in 2 lines or only one. But the breaking point does not correspond to large or medium breaking points defined in the SCSS config 🙃

So for this PR, I went with the logo shrink being implemented only id `:not(.expanded)`

### Screenshot

https://github.com/user-attachments/assets/743dd338-cc35-4007-a891-23f6be377433

- Fixes #11461


teolemon:
\update_test_results

stephanegigandet:
Small issue on the Pro platform where we add "Professionals" under the logo:

image

we should hide "Professionals" when we don't display the full logo:

image

stephanegigandet:
There's one window size where we don't have enough space for the "install the app": 
image

Not very common probably though. But we probably have some languages where the text is longer than the French text for Contribute, Discover and Professionals.

alexfauquette:
I gave it a second try

- I replaced the "Professional" by "Pro" is space is short
- I allowed the "get the app" the wrap if needed. With German we still have the issue but on a tiny screen-size range
- I remove the link to producer platform if user are already on the producer platform 


https://github.com/user-attachments/assets/b8a51f6c-b134-44e0-bc7f-5d6e7c071cc5


https://github.com/user-attachments/assets/202b47d6-7f79-414f-b020-776923f566ea

alexfauquette:
\update_test_results

PR #11668, opened 2025-03-25
feat: Promote Write API

Pawan8538:
### What

This PR updates the edit log to display a promotional message when a product is added by an external app. If the edit comes from an app, the app name is shown along with a link to the Write API documentation.


lib/ProductOpener/Display.pm
```diff
@@ -4225,8 +4225,14 @@ HTML
 				my $app_name = deep_get(\%options, "apps_names", $appid) || $appid;
 				my $app_user = f_lang("f_app_user", {app_name => $app_name});
 
+				# Append promotional text to encourage use of the Write API
+    			my $display_creator_promo = $app_user;
+    			my $promo = " (app – <a href='https://docs.openfoodfacts.org/write-api' target='_blank'>learn how your app can contribute too</a>)";
```
stephanegigandet, line 4230: We want to display this only when the creator userid is the userid of an app, not for all users. 

There's no really easy way to tell if the creator userid is an app. There aren't that many app so we could have an hardcoded list, but it's not very clean.

Pawan8538: Hey @stephanegigandet ,
I thought appid, app_name and uuid are only used for an app. 
>There's no really easy way to tell if the creator userid is an app.

If so than it's not good idea to show API docs to every creator.

teolemon:
per @stephanegigandet, let's close @Pawan8538

PR #11669, opened 2025-03-25
taxonomy: Tighten up “vegan” requirements

Freso:
A number of ingredients that are on their surface vegan actually undergo some kind of treatment or processing using animal products, which would make them not vegan. In many cases, this would get reflected in ingredient lists, like gelatin in jams and candy etc., but in some cases this is done in ways that are not going to be reflected in an ingredient list, e.g., because there won’t be one—like for fruits.

This also means that just because something is “vegetable” or “extracted from plants”, it doesn’t automatically make them vegan, so those labels have been moved to be synonyms of vegetarian rather than vegan.

And finally, `en:Non-vegetarian` did not imply/cause `en:Non-vegan` previously, which also doesn’t make sense. If something is explicitly not vegetarian, it also won’t be vegan.


Freso:
Is there a way to easily regenerate/mass-edit the `tests/unit/expected_test_results/**.json` files to fix the failing tests? Maybe it’s time I finally learn to use `jq`… :sweat_smile:

R13ose:
There are many ingredients that are maybe vegan and you don't know.  As a vegan, I see the ingredient natural flavoring which could be derived from plant or animal material.

Freso:
@R13ose `en: natural flavors` in `taxonomies/labels.txt` already doesn’t imply/set vegan (or even vegetarian) status, and `en: natural flavouring` in `taxonomies/food/ingredients.txt` have both vegan and vegetarian status set to `maybe`. Are you seeing products on OFF that are being wrongly set as vegan? Are there any of my changes here that you disagree with, or are there more changes you think should be included?

R13ose:
Thanks for pointing me in the right direction.

To all the questions, nothing that needs change that I can see.

benbenben2 (approved):
LGTM

Freso:
I’m going to close this and will try revisiting as a series of smaller PRs. (If others don’t beat me to it, like https://github.com/openfoodfacts/openfoodfacts-server/pull/13147 :) )

PR #11678, opened 2025-03-27
refactor: Convert additives tests to JSON format for better maintainability

### LINKED ISSUE #11560: Move more of the tests to JSON-based outputs for better explainability and maintenability
### What
- Move more of the tests to JSON-based outputs for better explainability and maintenability
- JSON outputs allow seeing what changed in the case of taxonomy updates, and allow for easier debugging
- A good candidate is additives.t

abderhman487:
### **What**

I've transformed the food additive tests from hardcoded data to a flexible JSON-based approach, significantly improving maintainability and testability. This makes the tests more organized and easier to maintain/extend.

**Key changes:**
- Created `additives_test_data.json` containing structured test data
- Modified `additives.t` to read from JSON file instead of hardcoded data
- Added special handling for Spanish vitamin tests
- Added diagnostic outputs for better visibility during test execution
- Fixed perltidy styling issues to comply with project standards
- Improved test maintenance and expandability

- Fixes #11560


stephanegigandet (changes requested):
Hi @abderhman487 ,
Thanks a lot for the PR. It is an improvement, but it's different from what we do in other tests.
What we want to do is to use the technique from other tests like ingredients.t : https://github.com/openfoodfacts/openfoodfacts-server/blob/main/tests/unit/ingredients.t 
There is a function to compare test results to expected JSON files, and there is an --update-test-results option that will update those expected JSON files.
This is very useful when we change the code of the ingredient parsing etc. We can easily see the diffs, but we can also easily update the expected test results.
Could you change the PR to use this approach?.

abderhman487:
Hi @stephanegigandet,

Thanks for taking the time to share your views. 

I have updated the PR to follow the approach used in ingredients.t. Now, the tests compare results to expected JSON files, and I've also added support for the `--update-expected-results` option to allow easier updates when needed.
I've also included code to remove temporary OCR fields that could cause test failures, ensuring more consistent test results.

Let me know if there's anything else you'd like to adjust!

tests/unit/additives.t
```diff
@@ -7,992 +7,625 @@ use Test2::V0;
...
+	extract_additives_from_text($product_ref);
+
+	# If there's a language test, perform it
+	if (exists $product_ref->{ingredients_text_fr}) {
+		compute_languages($product_ref);
+		clean_ingredients_text($product_ref);
+		extract_additives_from_text($product_ref);
+	}
+
+	# Remove any temporary OCR fields that might cause test failures
```
stephanegigandet, line 608: I don't understand how there could be OCR fields, as all the inputs are specified in this additives.t file, and they don't have OCR fields.

stephanegigandet: ah ok found it, it can be added by clean_ingredients_text()

stephanegigandet, line 608:
```suggestion
	# Remove any temporary OCR fields that might cause test failures
	# Those fields can be added by clean_ingredients_text()
```

stephanegigandet (approved):
Looks good, thank you!

abderhman487:
Thanks a lot for the review and the merge. @stephanegigandet

PR #11681, opened 2025-03-27
fix: replace minus with dash to separate product name and brand

### LINKED ISSUE #1558: Hyphen or en or em dash or something else to separate product name, brand and quantity
On OFF, the title of a product page is a combination of "Product name" + "Brand" + "Quantity", separated by the minus / hyphen character.

One question raised by @hangy is that we may want to use another character (such as dash). And that character may need to be different by language.

This bug is to discuss this and see what solution to implement.

Original discussion:

@hangy 
I'm not a typography expert, but as far as I know, the hyphen-minus (as present on most keyboards) is only used as a separator if necessary. Also, changing the msgid of a translation that has already been localized in Crowdin seems like a bad idea (unless strictly necessary). IMHO em dash is probably the better character according to this Wikipedia article. cc @teolemon?

  @stephanegigandet
stephanegigandet 3 days ago  Member
I'm a bit afraid of making punctuation translatable in fact, I don't really see how someone could translate " - " to a different dash. It might make sense for non roman alphabets, but for European languages we are going to get a random collection of translations...

So my intent was to get the smaller minus dash, because the long dash really looked ugly (in my opinion of course, I'm not an expert at all, and I have my own tastes :-) ) :

image

I'm ok with the slighly smaller dash: – if we want to avoid the minus -.

But to force it, we have to change the msgid I think, we can't leave the huge dash as the msgid.

stephanegigandet:
fixes #1558 reported by @hangy 7 years ago. Thanks for your patience! :)

I'm removing the title_separator string from the common.pot file, as in all languages it was set to " - ".


hangy (approved):
LGTM

PR #11686, opened 2025-03-28
fix: enable brands taxonomy for OBF and OPF

stephanegigandet:


teolemon (approved):
Do you prefer a global file for all 4 projects @stephanegigandet ?
Can we add samples for OBF and OPF in the current brands.txt

PR #11688, opened 2025-03-28
fix: OBF brands categories, ingredients panels

stephanegigandet:
all brands must be in xx:

note that the current OBF brands taxonomy has parents for groups like Unilever and Henkel, which is probably not wanted


teolemon (approved):
Great @stephanegigandet 👌 thank you very much 💄 🧴 🪒

PR #11696, opened 2025-03-29
taxonomy: mv duplicates to xx

benbenben2:
### What
add xx: something when there is more than 4 times the same translation.

- Fixes #-none-


taxonomies/origins.txt
```diff
@@ -196,16 +196,12 @@ zh: 未指定，不可用，未指定
 # World regions
 
 en: Africa
-af: Afrika
+xx: afrika
```
Freso, line 199: Shouldn’t it be capital-A “Afrika”? (and similar for other entries)

taxonomies/food/ingredients.txt
```diff
@@ -176,21 +176,10 @@ stopwords:zh: :配料
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
 en: imazalil
-da: imazalil
-de: Imazalil
```
Freso, line 180: German capitalises all nouns, so if the `xx:` entry isn’t capitalised (and it probably shouldn’t be in this case), the `de:` entry should still be there so it won’t show up lower-cased in German locales.

taxonomies/food/ingredients.txt
```diff
@@ -864,19 +842,9 @@ fr: mono- et diglycérides d'acides gras et stéaroyl-2-lactylate de sodium
 
 < en:E511
 en: nigari, bittern
+xx: nigari
 bg: Нигари
-ca: Nigari
-cs: nigari
-de: Nigari
```
Freso, line 870: Again with the special `de:` capitalisation.

taxonomies/food/ingredients.txt
```diff
@@ -1885,23 +1846,20 @@ pt: coalho de ovelha
 
 < en:enzyme
 en: cellulase
+xx: cellulase
 ar: سليولاز
 bg: целулаза
 bs: Celulaza
 ca: cellulasa
-de: Cellulase
```
Freso, line 1892: `de:` continues to be a special case :)

taxonomies/food/ingredients.txt
```diff
@@ -5916,27 +5822,18 @@ wikipedia:en: https://en.wikipedia.org/wiki/Smetana_(dairy_product)
 
 < en:sour cream
 en: Crème fraîche
-af: Crème fraîche
+xx: crème fraîche
```
Freso, line 5825: Could probably include unaccented version too, to clean up the list even further.
```suggestion
xx: crème fraîche, creme fraiche
```

taxonomies/food/categories.txt
```diff
@@ -70088,19 +69880,15 @@ ciqual_food_name:fr: Fruit de la passion ou maracudja, pulpe et pépins, cru
 
 < en:Tropical fruits
 en: Pineapple
+xx: ananas
```
stephanegigandet, line 69883: I think it makes sense to use xx: when there is really a term that works almost universally. e.g. when it's a name that is very distinctive (because it's origially from one specific country), like Calvados, Whisky, Gnocchi, Spaghetti, Paella, Chili con carne, that is understood in most countries. But I would not use xx: for common nouns that have different translations / endings in different languages, like "ananas", "papaya", "mandarina".

taxonomies/origins.txt
benbenben2: This is not case sensitive
This being said, I updated to have upper case everywhere. To be coherant with the rest of the taxonomy.

benbenben2:
Thank you @Freso and @stephanegigandet for taking the time to review the changes and impact of the changes.

You raised very good points. I  added condition that **en:** should be same as **xx:** and that **xx:** is added for more than 5 times the same translation (it was 4 previously).

taxonomies/food/ingredients.txt
benbenben2: I removed this one, for now

Freso (commented):
There’s still a massive amount of interesting capitalisation changes in the ingredients file. `categories` entries I feel should generally be capitalised, as those are titles. `origins` should also typically be capitalised, as those will be proper nouns. `ingredients` (and `_processing`), however, are mostly regular nouns (or verbs/adjectives for `_processing`!) and will most frequently be lowercased in most languages, except for German (for nouns) and in some special cases. For cases where localised versions of words are displayed, it will look jarring if/when the capitalisation is off. With this in mind, I’d like to still stress that when the capitalisation of a translation differs from the `xx` entry, the translation should be kept. This way, the system knows to use the proper capitalisation for the given locale.

benbenben2:
Thanks for the additional comments @Freso 

Here is what has been changed:
- **categories** and **origins** will start by upper case, regardless of the language code
  **ingredients**, **de** starts by upper case, other languages start by lower case
- update only if there are > 5 identical translations + it represents more than 50% of all translations (that would exclude Quarks for example)

taxonomies/food/ingredients.txt
```diff
@@ -49806,45 +49699,26 @@ wikipedia:en: https://en.wikipedia.org/wiki/Bogdanu%C5%A1a
...
-de: Cabernet franc
-es: Cabernet franc
-et: Cabernet franc
-fi: Cabernet franc
-fr: Cabernet franc
+en: cabernet franc
+xx: cabernet franc
+bg: каберне фран
+cv: каберне фран
+de: Cabernet Franc
```
Freso, line 49714: idk if german wants both words here to be uppercased, i could see the argument of "franc" being an adjective to "Cabernet". my many years old german school knowledge is not good enough to know for this case. :)

taxonomies/food/ingredients.txt
```diff
@@ -49853,56 +49727,33 @@ wikipedia:en: https://en.wikipedia.org/wiki/Cabernet_Franc
...
-et: Cabernet sauvignon
-fi: Cabernet sauvignon
-fr: Cabernet Sauvignon, Sauvignon
-gl: Cabernet sauvignon
+en: cabernet sauvignon
+xx: cabernet sauvignon
+az: kabernet sauviqnon
+bg: каберне совиньон, cabernet sauvignon
+cv: каберне совиньон
+de: Cabernet Sauvignon
```
Freso, line 49735: see `de: Cabernet Franc`

Freso (approved):
seems overall good to me now. i have some questions about german, but that can easily be fixed if any german speakers should complain. also, i think some ingredients should maybe retain being uppercased due to being a placename or similar, but, again, i can’t at a glance tell for which of these this would apply and for which it wouldn’t (and probably nothing that could be algorithm’d), so probably fine to do this now and if anything gets complained about, it can be corrected then.

stephanegigandet (approved):
Thank you!

Issue #11633, opened 2025-03-18
Make sure the donation message is not featured on social media link previews


stephanegigandet:
Currently when sharing some OFF website pages on LinkedIn, we get the donation message in the link preview:

![Image](https://github.com/user-attachments/assets/e4621e76-d2a4-4b9a-91b3-c17807a45494)

We may have similar issues on other social networks (to be investigated)

It would be great to find a way to have better previews.

One issue is that the `<h1>` tags are not generated anymore, e.g. https://world.openfoodfacts.org/new-nutriscore

areebahmeddd:
I'll take a look here :)

teolemon:
- Related: https://github.com/openfoodfacts/openfoodfacts-server/issues/7422

JagjeevanAK:
Hey @teolemon it will be good if we close one of the issue #7422 or #11633

star-warrior:
I'll start working on it :)

areebahmeddd:
Hey @star-warrior,

I was stuck with making changes in other open PRs in another OFF repository. Apologies for the delay. I will be shortly raising a PR on this.

Thank you.

JagjeevanAK:
Hey @areebahmeddd, If you are busy with other PR's, I think you should give-up this issue to @star-warrior. He will also get chance to explore the codebase, and you can use your productive time elsewhere.

star-warrior:
thanks for giving me chance on working on this issue. I'll start working on this ASAP

hangy:
Does `make dev` exit with an error? If it doesn't, everything should be up. You can also try `make up` afterwards

star-warrior:
```
#79 DONE 0.0s
 backend  Built
 dynamicfront  Built
 frontend  Built
 incron  Built
 minion  Built
openfoodfacts-shared-services
/usr/bin/sh: line 4: [: too many arguments
/usr/bin/sh: line 10: cd: too many arguments
make: *** [Makefile:539: clone_deps] Error 1
```

So this is how it ends, i dont think so this is an error that breaks everything but, @hangy here you go.

star-warrior:
Also while creating a branch I am greeted with this error

```
$ git switch donation-link-preview-sw
error: The following untracked working tree files would be overwritten by checkout:
        .check_perl_excludes
        .devcontainer/devcontainer.json
        .dockerignore
        .editorconfig
        .eslintignore
        .eslintrc.js
        .gemini/config.yaml
[...]

PR #10143, opened 2024-04-17
fix: Social media link preview missing website description (fixes #7422)

Josckar:
### What
Added the line to pass `canon_description` to `site_layout.tt.html`
Mainly taken from closed PR #7589 (could not get in contact with user @ferb7o2 to understand why it was closed).

### Related issue(s) and discussion

- Fixes #7422

PR #11697, opened 2025-03-29
Link preview issue solved and also description added to Needy pages

star-warrior:
### What

Solved Issue number #10143 and #11633 about link previews.

This issue only occurred on static HTML pages and not dynamic generated pages. Some pages has blank descriptions which are done for good and mostly on products page.

Added appropriate title and description for Link preview, the Image is kept the same which is Open Food Facts Logo.

- Fixes #[ISSUE NUMBER]
#10143 , #11633

### DIFF
--- a/tests/integration/expected_test_results/api_v2_product_read/get-auth-bad-user-password.html
+++ b/tests/integration/expected_test_results/api_v2_product_read/get-auth-bad-user-password.html
@@ -1,625 +1,948 @@
 <!-- start templates/web/common/site_layout.tt.html -->
 
-<!doctype html>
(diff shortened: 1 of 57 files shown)


JagjeevanAK:
Can you share the preview, How it is working?

star-warrior:
I dont think it can be previewed without proper deployment.

hangy (commented):
The code in `tests/integration/expected_test_results` is used to validate that the backend code works as expected, meaning that the backend code returns the correct HTML with substituted variables, and so on. They "static" HTML in there is not used to be returned from the webserver. (In fact, it doesn't even have to be deployed to production.)

PR #11727, opened 2025-04-01
feat: Update donation campaign banner

alexfauquette:
Replace the 2024 by 2025. And make it a variable such that we don't ask translators to update the year.

image


stephanegigandet (approved):
Looks good to me, thank you!

For translations we usuallly let Crowdin do them cc @teolemon

Valimp (approved):

PR #11738, opened 2025-04-03
fix: templates: string-quoting audit / fixups

jayaddison:
### What
From manually auditing the templates directory (searching for the literal string ` lang(`, basically), I think there are a few places where the templates should be updated to wrap the inputs in `edq(...)` (escape-double-quotes) because the results are interpolated inside double-quotes.

E.g. where a template contains `x = "[% lang('something') %]";` we should instead use `x = "[% edq(lang('something')) %]";`

Tests are likely going to fail here initially; I'll update the test expectations soon.

- Follow-up to #9822.

### DIFF
--- a/templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
+++ b/templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
@@ -35,8 +35,8 @@
             "element_type": "text",
             "text_element": {
                 "html": "<strong>[% edq(panel.wikipedia_title) %][% sep %]:</strong> [% edq(panel.wikipedia_abstract) %]",
-                "source_text": `Wikipedia`,
-                "source_url": `[% panel.wikipedia_url %]`,
+                "source_text": "Wikipedia",
+                "source_url": "[% panel.wikipedia_url %]",
                 "source_lc": "[% panel.wikipedia_url_lc %]",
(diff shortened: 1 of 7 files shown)


templates/api/knowledge-panels/health/ingredients/ingredient.tt.json
```diff
@@ -34,9 +34,9 @@
         {
             "element_type": "text",
             "text_element": {
-                "html": `<strong>[% edq(panel.wikipedia_title) %][% sep %]:</strong> [% edq(panel.wikipedia_abstract) %]`,
-                "source_text": `Wikipedia`,
-                "source_url": `[% panel.wikipedia_url %]`,
+                "html": "<strong>[% edq(panel.wikipedia_title) %][% sep %]:</strong> [% edq(panel.wikipedia_abstract) %]",
+                "source_text": "Wikipedia",
+                "source_url": "[% panel.wikipedia_url %]",
```
jayaddison, line 39: Perhaps we should add `edq(...)` here too, despite the absence of a call to `lang(...)`.

```suggestion
                "source_url": "[% edq(panel.wikipedia_url) %]",
```

stephanegigandet (approved):
Thank you!

jayaddison:
Thank you @stephanegigandet!

PR #11742, opened 2025-04-04
docs: OpenAPI spec for new images schema - WIP

stephanegigandet:
This PR contains OpenAPI specification for the new images schema for product images.

It is based on the discussion we had for the Pomme d'API project, see: https://docs.google.com/document/d/1IJ--JRNBtmUixm23KUsT_z9pWLudZKHytaYk12h7q-Q/edit?tab=t.0

The new schema:
- cleanly separates uploaded images and selected images
- removes some duplicate URLs for selected images

I created separate schemas for API v3, so that we can keep the documentation for the old schema in API v2.

The OpenAPI contains patternProperties. Those are now well supported in redocly.
For the review, I included a generate HTML file built with redocly:

`redocly build-docs docs/api/ref/api-v3.yml
`


JagjeevanAK (approved):
Everything looks good.

redocly.yaml
```diff
@@ -0,0 +1,21 @@
+theme:
+  openapi:
+    schemaDefinitionsTagName: Schemas
+apis:
+  main:
+    root: docs/api/ref/api.yml
+  test:
+    root: docs/api/ref/api-v3.yml
```
alexgarel, line 8: why not v2 and v3 ?

docs/api/ref/schemas/product_images_v3.yaml
```diff
@@ -0,0 +1,126 @@
...
+                    '(?<language_code>\w\w)':
+                      type: object
+                      description: |
+                        Front image in the language given by the 2 letter 'language_code'.
+                      $ref: "./image_selected.yaml#/components/schemas/ImageSelected"
+                ingredients:
+                  description: |
+                    Cropped images of the ingredients list in languages shown on the packaging.
+                  patternProperties:
+                    '(?<language_code>\w\w)':
```
alexgarel, line 65: shan't it be an opportunity to support `\w\w(_\w\w)?` codes ?

docs/api/ref/schemas/product_images_v3.yaml
```diff
@@ -0,0 +1,126 @@
...
+                      type: object
+                      description: |
+                        Packaging / recycling information image in the language given by the 2 letter 'language_code'.
+                      $ref: "./image_selected.yaml#/components/schemas/ImageSelected"
+          
+        selected_images:
+          type: object
+          title: Selected images URLs
+          description: |
+            URLs of selected images, generated at runtime.
```
alexgarel, line 93: So it does not depends on the current language ?

As this is for commodity, I would have provided a simpler structure with selected images only in current language.

docs/api/ref/schemas/product_images_v3.yaml
```diff
@@ -0,0 +1,126 @@
...
+          patternProperties:
+            "(?<image_type>front|ingredients|nutrition|packaging)":
+              description: |
+                Type of the selected image: front (front of product), ingredients (ingredients list), nutrition (nutrition facts), packaging (packaging/recycling information).
+              type: object
+              patternProperties:
+                '(?<language_code>\w\w|best)':
+                  type: object
+                  description: |
+                    Packaging / recycling information image in the language given by the 2 letter 'language_code'.
```
alexgarel, line 103: > Packaging / recycling information image 

Is inexact

docs/api/ref/schemas/image_uploaded.yaml
```diff
@@ -0,0 +1,15 @@
...
+      description: |
+        Image uploaded by a user or provided by a manufacturer identified by an integer 'imgid'.
+      properties:
+        uploader:
+          type: string
+          description: |
+            userid of the user who uploaded the image.
+          example: "stephane"
+        sizes:
+          $ref: "./image_sizes.yaml#/components/schemas/ImageSizes"
```
alexgarel, line 15: Where is the image uploaded time ?

docs/api/ref/schemas/image_selected.yaml
```diff
@@ -0,0 +1,78 @@
+components:
+  schemas:
+    ImageSelected:
+      type: object
+      title: Selected image
+      description: |
+        Images selected (and possibly cropped or rotated) to show an area of interest
+        (front of product, ingredients, nutrition facts, packaging info) in a specific language.
+      properties:
```
alexgarel, line 9: where is the time of image selection ? The user who selected would also be valuable.

alexgarel, line 9: Instead of having url directly here, we have to seek for selected_images field ?
That would be useful to explain this in the description.

docs/api/ref/redocly/api-v3.redoc-static.html
alexgarel, line 1: I put the rapidoc file in docs/assets I think it's better. The doc generation script can then copy it.

alexgarel (commented):
Great, I have some comments though.

PR #11753, opened 2025-04-08
fix: integration tests fail when ecobalyse service is unavailable

### LINKED ISSUE #11751: bug: Integration tests fail when Ecobalyse service is unavailable
### What
- Integration tests fail when the Ecobalyse service is unavailable

### Expected behavior
- Integration tests should not depend on external services

### Why
- We don't want flaky tests

john-gom:
### What

Convert Ecobalyse test to be a unit test so we can mock the service response

- Fixes #11751


tests/unit/environmental_impact.t
```diff
@@ -0,0 +1,140 @@
...
+	}
+}
+
+# Run the test
+my $updated_product_fields_ref = {};
+my $errors_ref = [];
+
+estimate_environmental_impact_service($product_ref, $updated_product_fields_ref, $errors_ref);
+
+# Save the result
```
stephanegigandet, line 106: I think pretty much everything below can be replaced by something like:

is($mock_calls, 1, "Ecobalyse mock called");

compare_to_expected_results($product_ref, "$expected_result_dir/$testid.json", $update_expected_results);
..

stephanegigandet (approved):
Thank you!

PR #11766, opened 2025-04-09
fix: compute Nutri-Score for coffee using prepared nutrition facts

stephanegigandet:
Reported by @TaciteOFF on Slack

Moving ground coffee in beverage preparations ensures that we compute Nutri-Score only if we have prepared nutrition facts (so we won't have an estimated Nutri-Score from nutrition facts derived from "ground coffee" ingredient).

Also removes some code that resulted in a "not applicable" Nutri-Score for coffees missing prepared nutrition facts.


alexgarel (approved):
Thanks !

Issue #11759, opened 2025-04-09
Some facet URLs with some non latin characters results in 400 bad requests


stephanegigandet:
Reported by @teolemon on Slack: https://hu.openfoodfacts.org/facets/c%C3%ADmk%C3%A9k/Hozz%C3%A1adott%20cukor%20n%C3%A9lk%C3%BCl

This is for languages where we don't remove accents like Hungarian, Greek etc.

https://hu.openfoodfacts.org/facets/c%C3%ADmk%C3%A9k/Hozz%C3%A1adott%20cukor%20n%C3%A9lk%C3%BCl

Strangely it works in dev mode:

```
stephane@stephane-jaguar-NS50-70MU:~/openfoodfacts-server$ wget -S 'https://hu.openfoodfacts.org/facets/c%C3%ADmk%C3%A9k/Hozz%C3%A1adott%20cukor%20n%C3%A9lk%C3%BCl'
--2025-04-09 09:57:45--  https://hu.openfoodfacts.org/facets/c%C3%ADmk%C3%A9k/Hozz%C3%A1adott%20cukor%20n%C3%A9lk%C3%BCl
Résolution de hu.openfoodfacts.org (hu.openfoodfacts.org)… 213.36.253.214
Connexion à hu.openfoodfacts.org (hu.openfoodfacts.org)|213.36.253.214|:443… connecté.
requête HTTP transmise, en attente de la réponse… 
  HTTP/1.1 400 Bad Request
  Server: nginx/1.18.0
  Date: Wed, 09 Apr 2025 07:57:45 GMT
  Content-Type: text/html; charset=iso-8859-1
  Content-Length: 309
  Connection: keep-alive
2025-04-09 09:57:45 erreur 400 : Bad Request.

stephane@stephane-jaguar-NS50-70MU:~/openfoodfacts-server$ wget -S 'http://hu.openfoodfacts.localhost/facets/c%C3%ADmk%C3%A9k/Hozz%C3%A1adott%20cukor%20n%C3%A9lk%C3%BCl'
--2025-04-09 09:58:01--  http://hu.openfoodfacts.localhost/facets/c%C3%ADmk%C3%A9k/Hozz%C3%A1adott%20cukor%20n%C3%A9lk%C3%BCl
Résolution de hu.openfoodfacts.localhost (hu.openfoodfacts.localhost)… ::1, 127.0.0.1
[...]

PR #11768, opened 2025-04-09
fix: spaces in nginx ProxyPass URLs lead to Bad Request

stephanegigandet:
Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/11759

Although I don't remember exactly, but I think we had a discussion about $request_uri vs $uri$is_args$args but I didn't find it.

### DIFF
--- a/conf/nginx/sites-available/off
+++ b/conf/nginx/sites-available/off
@@ -165,7 +165,12 @@ server {
 		access_log /var/log/nginx/proxy-off-access.log proxied_requests buffer=256K flush=1s;
 		error_log /var/log/nginx/proxy-off-error.log;
 		# Note: as soon as we use a variable in proxy_pass, $uri and the rest must be explicit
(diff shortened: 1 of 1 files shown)


conf/nginx/sites-available/off
```diff
@@ -175,8 +180,6 @@ server {
 		real_ip_recursive on;
 		access_log /var/log/nginx/proxy-off-access.log proxied_requests buffer=256K flush=1s;
 		error_log /var/log/nginx/proxy-off-error.log;
-		# Note: as soon as we use a variable in proxy_pass, $uri and the rest must be explicit
-		proxy_pass http://127.0.0.1:$apache_port$uri$is_args$args;
-	}
+		proxy_pass http://127.0.0.1:$apache_port?$request_uri;
```
alexgarel, line 183: The ? is wrong here.

stephanegigandet: fixed, thanks

alexgarel (approved):
Thanks.

PR #11776, opened 2025-04-12
chore: taxonomy updates

aleene:
### What
Some small taxonomy upddates


benbenben2 (approved):
Looks good to me.

Not 100% sure for "прахов протеин" being egg white powder becuase "протеин" looks like "protein". But I am not expert in Bulgarian.

Thanks @aleene

PR #11778, opened 2025-04-13
taxonomy: remove duplicates in taxonomies and put back error on build_taxonomies

benbenben2:
### What
remove duplicates in taxonomies and put back error on build_taxonomies

- Fixes #-none-


taxonomies/additives.txt
```diff
@@ -11655,21 +11867,23 @@ wikipedia:en: https://en.wikipedia.org/wiki/Tara_spinosa#Uses
 #description:en:TARA GUM is produced by separating and grinding the endosperm of T. spinosa seeds. Tara gum is used as a thickening agent and stabilizer in a number of food applications.
 # ingredient/fr:gomme-tara has 254 products in 6 languages -@2019-05-12
 
-en: E418, Gellan gum, gellan
+en: E418, Gellan gum, gellan, E-418, E 418, INS418, INS-418, INS 418
```
stephanegigandet, line 11870: in theory the extra E-418, E 418, INS418, INS-418, INS 418 should not be needed, as we normalize the e-numbers when doing ingredients parsing

taxonomies/food/ingredients.txt
```diff
@@ -68,10 +70,6 @@ synonyms:en: semi-wholemeal, semi-whole, semi-whole grain, semi-wholegrain, semi
 
 synonyms:en: semi-sweet, semisweet
 
-synonyms:es: zumo, jugo
-
-synonyms:es: jarabe, sirope
```
stephanegigandet, line 73: why did you remove this synonym and the ones below?

for zumo / jugo, we have it as an ingredient en: juice, but for en: syrup we don't have sirope as synonym of jarabe

tests/unit/expected_test_results/additives/fr-emulsifiant-chlorure.json
```diff
@@ -14,6 +12,8 @@
    "lc" : "fr",
    "minerals_tags" : [],
    "nucleotides_tags" : [],
-   "other_nutritional_substances_tags" : [],
+   "other_nutritional_substances_tags" : [
+      "en:choline-chloride"
```
stephanegigandet, line 16: here it was used an additive

tests/unit/expected_test_results/additives/fr-vitamine-c-with-function.json
```diff
@@ -1,17 +1,15 @@
...
-   ],
-   "additives_tags" : [
-      "en:e300"
-   ],
+   "additives_n" : 0,
+   "additives_original_tags" : [],
+   "additives_tags" : [],
    "amino_acids_tags" : [],
    "ingredients_lc" : "fr",
    "ingredients_text" : "Eau, sucre, antioxydant: vitamine C",
```
stephanegigandet, line 7: Here it is used as an additive

stephanegigandet (commented):
There are a couple of tests that have changed because vitamin c and chlorure de choline are no longer found as additives, even when an additive class is before them. With the current system, I'm not sure how it could be solved though.

taxonomies/food/ingredients.txt
benbenben2: That was the error:

ERROR - duplicate_synonym - 69571 - es:jarabe already is associated to es:jarabe (ingredients) - es:jarabe cannot be mapped to entry en:syrup

tests/unit/expected_test_results/additives/fr-emulsifiant-chlorure.json
benbenben2: understood. It was duplicates between additives.txt and other_nutritional_substances.txt, the solution was to keep it only in the latter one as there is already a block "en: Choline chloride" there. If Choline chloride is a synonym or a child of E1001, it could be reorganized in the taxonomy. But it might be out of the scope of this PR for now.

tests/unit/expected_test_results/additives/fr-vitamine-c-with-function.json
benbenben2: understood. It was duplicates between additives.txt and vitamins.txt, the solution was to keep it only in vitamins.txt. Although we expect that "fr-vitamine-c-with-function" to produce different expect result :-(

benbenben2:
> There are a couple of tests that have changed because vitamin c and chlorure de choline are no longer found as additives, even when an additive class is before them. With the current system, I'm not sure how it could be solved though.

should we accept it as is in the PR but create an (or more) issue(s) to handle these cases? Do you think it can be handled in Ingredients.pm instead?

taxonomies/additives.txt
stephanegigandet: In all languages, we have some preparsing code to change all variations like that to E418:

https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Ingredients.pm#L6738

stephanegigandet:
> > There are a couple of tests that have changed because vitamin c and chlorure de choline are no longer found as additives, even when an additive class is before them. With the current system, I'm not sure how it could be solved though.
> 
> should we accept it as is in the PR but create an (or more) issue(s) to handle these cases? Do you think it can be handled in Ingredients.pm instead?

I think we can keep the PR as-is, and hopefully there won't be too many cases like that.

Thanks for all the work to remove all the conflicts @benbenben2 !

stephanegigandet (approved):

PR #11781, opened 2025-04-14
Feat: github action pinning with Ratchet

JagjeevanAK:
### What

Most CI/CD systems are one layer of indirection away from `curl | sudo bash`. Unless you are specifically pinning CI workflows, containers, and base images to checksummed versions, _everything_ is mutable: GitHub labels are mutable and Docker tags are mutable. This poses a substantial security and reliability risk.

**Few Week earlier the most popular action was compromised with `curl` command pointing to an python script :** [Link](https://alexwlchan.net/2025/github-actions-audit/)

### DIFF
--- a/.github/workflows/auto-assign-pr.yml
+++ b/.github/workflows/auto-assign-pr.yml
@@ -17,4 +17,4 @@ jobs:
     name: 'Pull request assignment'
     runs-on: ubuntu-latest
     steps:
-      - uses: toshimaru/auto-author-assign@v2.1.1
+      - uses: toshimaru/auto-author-assign@16f0022cf3d7970c106d8d1105f75a1165edb516 # ratchet:toshimaru/auto-author-assign@v2.1.1
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -19,35 +19,35 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@v4
-      with:
-        # We must fetch at least the immediate parents so that if this is
-        # a pull request then we can checkout the head.
-        fetch-depth: 2
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # ratchet:actions/checkout@v4
+        with:
+          # We must fetch at least the immediate parents so that if this is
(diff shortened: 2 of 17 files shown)


stephanegigandet:
Sounds like a good idea to me.

JagjeevanAK:
I appreciate @teolemon’s point that our CI actions aren’t updated very frequently and that there might not seem to be an urgent need for them. However, I respectfully disagree. Relying on older versions of GitHub Actions can be risky because they’re more vulnerable to security exploits – their hashes and SHAs can be modified. That’s why using Ratchet is so important for keeping things secure.

teolemon:
You're misrepresenting what I said yesterday.  I take updating actions seriously and that's why we have Dependabot, and I'm wary of switching to a 3rd party solution to do that.
Basically:
- we need a way to get automatic update notifications of dependencies, with a clear changelog of what will change. Dependabot does that. I don't know if Ratchet does that. Dependabot is maintained by GitHub, which also has dedicated teams monitoring supply chain attacks. 
- we don't use that many actions, and most actions we use are well-maintained (or even GitHub official actions), and not very likely to be subject to a supply chain attack ([introducting tj-files 2 days after the CVE was probably bad luck](https://github.com/openfoodfacts/openfoodfacts-server/pull/11630) )
- The real risk, and the priority should review existing dependancies, replace any abandoned ones by more popular ones if the cases happens. Pinning to the commit will bring little real additional safety compared to pinning on version.

JagjeevanAK:
@teolemon , Really sorry for that I was not doing it by any intention in meeting I got your point differently. Your point's are also valid.

alexgarel:
@JagjeevanAK as decided on slack, the next step on this PR is to test on a toy project how Rachet and dependabot are interacting:
* can dependabot still creates update PRs
* can we run an action or so to automatically transform dependabot versions to commit hash (even on merge, maybe, or in the PR itself).

alexgarel:
@JagjeevanAK I'm putting this PR as draft, until you are able to run the tests with dependabot.

To transform dependabot versions into rachet style version, we could try to do it with a manual action (action like /update_tests_results). This way we could launch it on dependabot PR's to transform links.
Also a check could verify that we are using rachet style versions on actions and fail otherwise.

dolmen (changes requested):
This patch should not change indent.

teolemon:
@JagjeevanAK conflicted pr

PR #11782, opened 2025-04-14
fix: Configured Apache log rotation in docker deployment scenario

### LINKED ISSUE #5970: Configure Apache log rotation in docker deployment scenario

We do not configure Apache log rotation with docker deployment, it can leads to potential problems at a moment in the future.

It seems the best for now could be setting up logrotate system wide (ensure it in deploy script ?), as system logrotate has access to volume containing log and can sens signal to apache process running say `docker-compose exec -T backend killall -s SIGHUP apache2`).

Normally we should send all logs to docker log handling and harvest them in a log facility… but that's for later on.

I didn't see any way Apache can auto-rotate logs by itself, see https://httpd.apache.org/docs/2.4/logs.html#rotation

### Additional context

_No response_

_No response_

### Time per product

_No response_

JagjeevanAK:
Signed-off-by: Jagjeevan Kashid [jagjeevandev97@gmail.com](mailto:jagjeevandev97@gmail.com)
### What
Implemented a log-rotation according to described solution by @alexgarel in the issue

I would suggest maintainer to change the config file of logrotate according to there need current setting keep 14 rotated logs (i.e., 2 weeks) and other things.  

- Fixes #5970


conf/logrotate/apache2
alexgarel, line 1: This one is not used in docker. It's for production only (proxmox containers).
It's not *.log but *log on purpose and it have to work for all flavors.

So never touch this for docker.

Dockerfile
```diff
@@ -198,6 +199,9 @@ ARG USER_GID
 RUN usermod --uid $USER_UID www-data && \
     groupmod --gid $USER_GID www-data
 
+# Setup logrotate cron job
+RUN echo "0 0 * * * /usr/sbin/logrotate /etc/logrotate.conf" > /etc/cron.d/logrotate && \
+    chmod 0644 /etc/cron.d/logrotate
```
alexgarel, line 204: This will not work as there is no cron daemon in docker containers (and that's better)

That's why I was expecting either to use an external cron (on docker staging VM) or to have Apache rotate it's log by itself (this is a capability found in some logging framework, I don't know if it's the case for apache2.

The first option means you might not be able to resolve this issue, but in the -infrastructure repository, and with a manual deployment.

alexgarel:
@JagjeevanAK I'm closing this PR as I think there is no good way to do this here.

JagjeevanAK:
100% you are correct I was about to say this thing but forgot to mention on this PR.

PR #11783, opened 2025-04-14
feat: add diet info from GS1 imports + match categories with gpc_category_code:en

stephanegigandet:
GS1 has ~ 20 diets (e.g. vegan, vegetarian etc.) that product can be tagged with. This PR adds support to add those diets as labels.

Also included: we now match categories using the gpc_category_code:en property in the categories taxonomy. Previously we only did a match on the GPC category name, and not the code.

+ some tests


lib/ProductOpener/Tags.pm
```diff
@@ -5245,6 +5247,43 @@ sub get_knowledge_content ($tagtype, $tagid, $target_lc, $target_cc) {
...
+
+sub create_property_to_tag_mapping_table ($tagtype, $property) {
+	my %mapping = ();
+
+	if (exists $properties{$tagtype}) {
+		foreach my $tagid (keys %{$properties{$tagtype}}) {
+			my $value = $properties{$tagtype}{$tagid}{$property};
+			if (defined $value) {
+				$value =~ s/^\s+//;
+				$value =~ s/\s+$//;
```
alexgarel, line 5278: why do we do that ? A comment is welcome.

I don't think the second line is useful. Or should the first line be `value =~ s/^\s+/\s/;` ?

lib/ProductOpener/GS1.pm
```diff
@@ -520,6 +560,28 @@ my %gs1_message_to_off = (
...
+
+=head2 load_gpc_category_codes_from_categories_taxonomy()
+
+Loads the GPC category codes from the categories taxonomy (in the gpc_category_code:en: property) and stores them in %gs1_gpc_category_codes_to_off.
+
+=cut
+
+sub load_gpc_category_codes_from_categories_taxonomy() {
+
+	return if %gs1_gpc_category_codes_to_off;
```
alexgarel, line 579:
```suggestion
	# exit if already loaded
	return if %gs1_gpc_category_codes_to_off;
```

alexgarel (approved):
Great.
I approve but give a look at my comment before merging !

lib/ProductOpener/Tags.pm
stephanegigandet: I think we can remove both lines in fact

PR #11795, opened 2025-04-17
fix: keep query parameters in facet redirect URLs

stephanegigandet:
Continuation of https://github.com/openfoodfacts/openfoodfacts-server/pull/11794 to fix ?json=1 and add support for all other query parameters.

### DIFF
--- a/lib/ProductOpener/HTTP.pm
+++ b/lib/ProductOpener/HTTP.pm
@@ -49,7 +49,7 @@ BEGIN {
 		&get_http_request_headers
 		&set_http_response_header
 		&write_http_response_headers
-		&add_extension_and_query_parameters_to_redirect_url
+		&extension_and_query_parameters_to_redirect_url
 		&redirect_to_url
 		&single_param
 		&request_param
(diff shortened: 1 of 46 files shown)


lib/ProductOpener/HTTP.pm
```diff
@@ -218,6 +219,32 @@ sub get_http_request_header($header_name) {
...
+	if ($request_ref->{extension}) {
+		$request_ref->{redirect} .= '.' . $request_ref->{extension};
+	}
+	# Add the query parameters to the redirect URL
+	if ($request_ref->{query_parameters}) {
+		$request_ref->{redirect} .= '?' . $request_ref->{query_parameters};
+	}
+
+	return;
+}
```
alexgarel, line 246: just for the discussion:

My approach was returning a string because it makes it a [pure function](https://en.wikipedia.org/wiki/Pure_function), I think we should try to do more of this when possible.
Here it could have returned a string, that would be then concatenated to {redirect}, it makes the function more general and with no side effects (make the code easier to read, making explicit where mutations happens)

lib/ProductOpener/Routing.pm
```diff
@@ -780,7 +780,7 @@ sub sanitize_request($request_ref) {
 
 			param($parameter, 1);
 			$request_ref->{query_string} =~ s/\.$parameter(\b|$)//;
-			$request_ref->{response_type_as_extension} = 1;
+			$request_ref->{extension} = $parameter;
```
alexgarel, line 783: Good idea.

alexgarel (approved):
Great

lib/ProductOpener/HTTP.pm
stephanegigandet: another version I tried was like this:

`$request_ref->{redirect} .= extension_and_query_parameters_to_redirect_url($request_ref->{extension}, $request_ref->{query_parameters});
`

that had the advantage to have a function that didn't need to know anything about $request_ref
but it felt a bit weird to have all those parameters, it felt more natural to just use the $request_ref object.   

another option is:

`$request_ref->{redirect} .= extension_and_query_parameters_to_redirect_url($request_ref);
`
and it's questionable if we need a function, we could also just do something like:

`$request_ref->{redirect} .= ($request_ref->{extension} ? '.' . $request_ref->{extension} : '') . ($request_ref->{query_parameters} ? '?' . $request_ref->{query_parameters} : '')
`

but we would use it twice, and maybe later update it in one place and forget the other one.

And there was another option:  add a parameter to redirect_url() to indicate that we want to add the extension and query parameter.

More than one way to do it. I'm fine with all of them.

alexgarel: For me there is no doubt that `$request_ref->{redirect} .= extension_and_query_parameters_to_redirect_url($request_ref);`
is the best option !

The function is pure, without too much hassle (encapsulation).
And it's easier to add some subtle changes in the future, that might be linked to other aspects, and it will be applied everywhere (separation of concerns, single responsability).

PR #11800, opened 2025-04-18
taxonomy: Update L. plantarum ingredient

Freso:
Several lactic acid bacteria were recategorised in 2020, including Lactobacillus plantarum being reassigned to the “new” genus Lactiplantibacillus, thus becoming Lactiplantibacillus plantarum.

See
https://en.wikipedia.org/wiki/Lactobacillaceae
https://en.wikipedia.org/wiki/Lactiplantibacillus
https://en.wikipedia.org/wiki/Lactiplantibacillus_plantarum


stephanegigandet (approved):
Thank you!

PR #11803, opened 2025-04-18
ci: GitHub Actions: prefer tags/commit IDs instead of branch names

jayaddison:
### What
Three of the GitHub Actions that we use in our GitHub Actions CI workflows (via the `uses: ...` clause in the YAML files) refer to those actions based on a branch name, instead of a conventional tagged release number or fixed commit ID.

The three relevant actions are:

  * `actions/add-to-project` (used to add issues/PRs to our GitHub Project boards)
  * `appleboy/ssh-action` (used during some of our deployment workflows)
  * `mschilde/auto-label-merge-conflicts` (used to annotate PRs that currently have merge conflicts)

I noticed this during some auditing of our GitHub Actions references during [Slack discussion](https://openfoodfacts.slack.com/archives/C02LDQDDD/p1744723461030369) of the possible merits and disadvantages of [`ratchet`](https://github.com/sethvargo/ratchet/) and how it relates to our existing use of GitHub's [Dependabot](https://github.com/dependabot).

Regardless of whether we use Ratchet and/or Dependabot, I think it is better to use versioned tags or commit IDs instead of branch names when referring to actions modules, because the contents of branches can vary and may sometimes contain code that is not release-ready.

:spiral_notepad: **Detail notes**

* As an exception in the case of `mschilde/auto-label-merge-conflicts`, I chose to pin to a commit ID instead of a tag name.  This is because the functionality seems very stable, and upstream has not released a version since Y2020.  So I would expect that for minimal disruption and workload we could continue to use this precise version.

* Technically `appleboy/ssh-action` downloads the latest-known version of another project, [`appleboy/drone-ssh`](https://github.com/appleboy/drone-ssh) when it runs.  I decided _not_ to pin to a specific version -- something that is possible using the [`version` option](https://github.com/appleboy/ssh-action/blob/master/README.md?plain=1#L53) -- because I doubt that Dependabot would detect updates to that and to recommend upgrades, meaning that we might unintentionally miss important updates.

- [Slack thread](https://openfoodfacts.slack.com/archives/C02LDQDDD/p1744723461030369) related to Ratchet for GitHub Actions version management.

### DIFF
--- a/.github/workflows/container-deploy.yml
+++ b/.github/workflows/container-deploy.yml
@@ -52,7 +52,7 @@ jobs:
         false
 
     - name: Check for an eventual NO_DEPLOY file to put deployment on hold
-      uses: appleboy/ssh-action@master
+      uses: appleboy/ssh-action@v1.2.2
       with:
         host: ${{ env.SSH_HOST }}
         username: ${{ env.SSH_USERNAME }}
@@ -100,7 +100,7 @@ jobs:
       run: echo Timeout && false # fail if build time out
 
     - name: Checkout git repository
-      uses: appleboy/ssh-action@master
+      uses: appleboy/ssh-action@v1.2.2
       with:
         host: ${{ env.SSH_HOST }}
         username: ${{ env.SSH_USERNAME }}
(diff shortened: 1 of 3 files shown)


jayaddison:
Thank you @hangy!

stephanegigandet (approved):
Thank you!

Issue #11798, opened 2025-04-17
Category `en: Pickles` is both pickled cucumbers and pickled vegetables more generally


Freso:
### What
- The category `en: Pickles` currently seems to be a mix of both pickled vegetables generally and also pickled cucumbers specifically.

### Expected behavior
- I would expect `en: Pickles` to be synonymous with `en: Pickled cucumbers` and pickled vegetables more generally to be something like `en: Pickled vegetables`, or maybe even `en: Pickled food`.

### Additional context

“Pickles” in English generally refer to pickled cucumbers, where most(?) other languages, “pickle” refer solely to the process and can apply to multiple different products.

This might also be something to tackle once we store user-entered categories/labels/etc. separately from the derived ones. (I think @stephanegigandet talked about this at one point in Slack not too long ago, but I can’t find the thread/post right now, and also can’t find any issues about this.)

### Number of products impacted

As of this writing, there are [9524 products under the Pickles category](https://world.openfoodfacts.org/facets/categories/pickles), with [2004 of those having the “Pickled cucumbers” category](https://world.openfoodfacts.org/facets/categories/pickled-cucumbers).

PR #11806, opened 2025-04-19
taxonomy: Untangle un/pickled cucumbers/gherkins

Freso:
Right now pickled and non-pickled cucumbers and gherkins seem to be mixed together with no clear purpose or clear separation of what should go where.

This is an attempt at starting to unravel this.

See:
https://openfoodfacts.slack.com/archives/C06A7LENM/p1744912951211729
https://en.wikipedia.org/wiki/Pickled_cucumber
https://www.wikidata.org/wiki/Q1211297
https://www.wikidata.org/wiki/Q1365891

Related (but not addressed here) issue:
https://github.com/openfoodfacts/openfoodfacts-server/issues/11798

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -51,7 +51,7 @@
 # Synonyms must be separated with an empty line.
 # If a term is also in the taxonomy, it must come first. e.g. "en:hydrolysed"
 
-# comment:en:These are words that occur inside the name of one ingredient, that can be considered as synonyms 
+# comment:en:These are words that occur inside the name of one ingredient, that can be considered as synonyms
 # for matching the ingredient with entries in the taxonomy
 # note: synonyms should not be ingredients themselves, if we have an ingredient entry with synonyms, they should not be listed in the synonyms below as well
 # for example, zumo, jugo (juice) cannot be listed here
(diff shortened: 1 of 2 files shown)


taxonomies/food/ingredients.txt
```diff
@@ -85393,12 +85394,20 @@ ciqual_food_code:en: 20210
...
+sv: smågurka, smågurkor, smågurk
+
+< en:cucumber
+en: pickled cucumber
+da: syltet agurk, syltede agurker
+wikidata:en: Q1365891
+
+< en:baby cucumber
+< en:pickled cucumber
+en: pickled gherkin, pickle
```
stephanegigandet, line 85410: This one should be removed right? We already have an entry "en: Pickled gherkins" above. Also "pickle" by itself should not be asynonym.

Freso: `en: Pickled gherkins` is a category. `pickle` by itself is already a synonym, I just moved it: https://github.com/openfoodfacts/openfoodfacts-server/pull/11806/files#diff-bf065447ce91af597ab8343c4053a9c4c17d8f4409ec243012daf06aae7f38c4L85321

stephanegigandet (approved):
Thank you!

PR #11810, opened 2025-04-20
fix: Rename label image files with `_`

Freso:
It seems like images with `_` are ignored/not considered for showing on the website, both on `facets/labels` pages and on product pages.

This renames all `image:` files referenced in `taxonomies/labels.txt` that have `_` in them to use `-` instead.

See also related dicussion (and debugging/testing) in Slack:
https://openfoodfacts.slack.com/archives/C02LDQDDD/p1745161790853179


benbenben2 (approved):
LGTM

stephanegigandet (approved):
Thanks

PR #11814, opened 2025-04-22
taxonomy: Add Swedish consumer waste recycling

Freso:
This expands the packaging/recycling taxonomy with the most common day-to-day household/consumer waste recycling types in Sweden.

Used a bunch of stuff I have in the apartment for reference as well as official Swedish sites like https://www.sopor.nu/


stephanegigandet (approved):
Thanks

PR #11821, opened 2025-04-25
ci: dependabot: ignore 'jquery' updates to v3.0.0+

jayaddison:
### What

We use `jquery` v2.x for compatibility with the Foundation UI framework, as described in #11109.

However, Dependabot isn't yet configured to know about this, so it attempts to suggest that we upgrade to `jquery` v3.x - e.g. #11820 that it opened earlier today.

So, let's [configure Dependabot to ignore](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#ignoring-specific-dependencies) those particular updates.

- Follow-up to #11109.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -38,6 +38,9 @@ updates:
         patterns:
           - "ts-node"
           - "typescript"
+    ignore:
+      - dependency-name: "jquery"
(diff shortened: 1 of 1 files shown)


hangy (approved):

stephanegigandet (approved):
Thanks

jayaddison:
Thank you both!

PR #11827, opened 2025-04-27
taxonomy: paratha

moon-rabbitOFF:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -106818,23 +106818,20 @@ fr: Tomates marinées
 en: Torshi
 fr: Torchi
 
-< en:Canned fruits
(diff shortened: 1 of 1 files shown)


Freso:
Were all the removals intentional?

moon-rabbitOFF:
@Freso Yes it was intentional. Is there a removal that you don't understand/don't agree with ?
St Moret is a brand, we don't add brand in the categories. 
We put in the category "Fruits" only products with 100% fruits. Pickled fruits are not fruits, they are fruit preparations

Freso:
I was just wondering because the commit only mentions “paratha”, so wasn’t sure if all the non-paratha changes were meant to be part of it as well. :)

Issue #11830, opened 2025-04-28
Demote products from vegan/vegetarian/palm-oil-free if they have ingredients specifically contradicting this


Freso:
### Problem

Sometimes a product may contain ingredients that directly contradict a given label. E.g., a “vegan” product might contain milk. This can be due to a variety of factors and might just be bad data (we already have data quality checks like [vegan label but not vegan ingredient](https://world.openfoodfacts.org/facets/data-quality-errors/vegan-label-but-non-vegan-ingredient)), but might also be a genuine error on the product or producer’s side.

Right now, a label will _always_ override whatever is discovered during ingredient parsing, even if they directly contradict each other. E.g., a product with “palm oil” as an ingredient and “no palm oil” label would be given as `en:no-palm-oil` which may provide false information to users/consumers.

### Proposed solution

Instead of outright claiming that a product is vegan, vegetarian, or palm oil free because there’s a label on the packaging, also check the ingredients to see if anything contradicts this statement. If so, demote to a “maybe” status to let the user/consumer know that there is something going on.

### Additional context

See this Slack discussion: https://openfoodfacts.slack.com/archives/C03H290LF/p1743584095493549

### Code pointers

* `lib/ProductOpener/Ingredients.pm`

### Number of products impacted

As of this writing, there are [779 “Vegan label but non vegan ingredient”](https://world.openfoodfacts.org/facets/data-quality-errors/vegan-label-but-non-vegan-ingredient), [244 “Vegetarian label but non vegetarian ingredient”](https://world.openfoodfacts.org/facets/data-quality-errors/vegetarian-label-but-non-vegetarian-ingredient), and [1 “Mutually exclusive tags for labels contains-palm-oil and labels no-palm-oil”](https://world.openfoodfacts.org/facets/data-quality-errors/mutually-exclusive-tags-for-labels-contains-palm-oil-and-labels-no-palm-oil) products (the last one isn’t a label/ingredient conflict though; not sure if there's a data QA test for that currently).

### Part of
-

PR #11831, opened 2025-04-28
feature: Handle label ingredient conflicts

Freso:
### What

Sometimes a product may contain ingredients that directly contradict a given label. E.g., a “vegan” product might contain milk. This can be due to a variety of factors and might just be bad data (we already have data quality checks like “vegan label but not vegan ingredient”), but might also be a genuine error on the product or producer’s side.

Right now, a label will always override whatever is discovered during ingredient parsing, even if they directly contradict each other. E.g., a product with “palm oil” as an ingredient and “no palm oil” label would be given as `en:palm-oil-free` which may (or may not!) be providing false information to users/consumers.

Instead of outright claiming that a product is vegan, vegetarian, or palm oil free because there’s a label on the packaging, this checks the ingredients to see if anything contradicts this statement. If so, this now demotes the product to a “maybe” status to let the user/consumer know that there is something going on and exercise extra caution.

- Slack discussion: https://openfoodfacts.slack.com/archives/C03H290LF/p1743584095493549
- Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/11830


stephanegigandet (approved):
Thank you!

PR #11842, opened 2025-04-29
feat: Staging documentation

CharlesNepote:
Staging auth is not documented anywhere. Here's a quick fix, with an example.


stephanegigandet (approved):
Thank you !

PR #11852, opened 2025-05-03
fix: Lacto-ovo-vegetarian taxonomy

Freso:
Currently the lacto- and ovo-vegetarian labels both derive from lacto-ovo-vegetarian, which will almost always be wrong. Ovo-vegetarians consume eggs but not dairy, while lacto-vegetarians consume dairy but not eggs—ovo-lacto-vegetarians consume both dairy and eggs.

This makes the `en:Ovo-lacto-vegetarian` labels instead be a parent of `en:Lacto-vegetarian` and `en:Ovo-vegetarian` instead, since anything that is either lacto- or oco-vegetarian will also be ovo-lacto-vegetarian. `en:Ovo-lacto-vegetarian` is additionally made a child of `en:Vegetarian` since the ovo/lacto modifier adds additional restrictions, so vegetarians who still consume non-human animal byproducts (like dairy and eggs) will be fine with this.

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -19996,20 +19996,7 @@ es: La Sociedad Vegetariana
 he: איגוד הצמחונות
 wikidata:en: Q2154954
 
-en: Ovo-vegetarian
-ca: Ovovegetarià
-de: Ovo-vegetarisch
-es: Ovo-vegetariano
-fi: ovo-vegetaarinen
-fr: Ovo-végétarien
-hu: Ovo-vegetáriánus
-it: Ovo-vegetariano
-pt: Ovo-vegetariano
-sv: Ovo-vegetarisk, ovovegetarisk, ovovegetarianer
-wikidata:en: Q3543760
-
-< en:Lacto-vegetarian
-< en:Ovo-vegetarian
+< en:Vegetarian
 en: Ovo-lacto-vegetarian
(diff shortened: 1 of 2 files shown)


stephanegigandet (commented):
I'm not too sure about the "Contains milk" label. It's not really a label, if we want to find products that contain milk, then using the ingredients en:milk tag is the way to go.

Freso:
@stephanegigandet Is it possible to have `incompatible_with:en: ingredients:en:dairy` and match that against labels with `ingredients:en: en:dairy` (ideally as well as children of `ingredients:en:dairy`)? This PR was split out of my vegan label tightening, where I’ve also marked `en: Contains milk` as a child of `en: Non-vegan`, which will, in turn, cause all of the milk labels to get marked as not vegan due to the label override code, even if they have no ingredients listed or all (non-vegan) ingredients are unknown. I _could_ mark all the milk labels with `< en:Non-vegan` directly, but this felt like a better approach.

(I also realise that this would all be better once we store derived and actually entered items separately, but I’m not sure how far off in the future that is, ie., whether to wait for this to manifest.)

Freso:
I took out the `< en:Contains milk` bits for now, to not hold up the rest of this longer than necessary.

stephanegigandet (approved):
Thank you

PR #11855, opened 2025-05-04
feat: Add weight % for single-item packaging

Freso:
### What

This handles the case where we know all the packaging (`packagings_complete` is set to true/`1`), it only consists of a single packaging item, _and_ we don’t know the weight of it. Even without knowing the actual weight, we do know that that single item makes up 100% of the packaging’s weight.


stephanegigandet (approved):
Thanks!

PR #11857, opened 2025-05-06
feat: Use multiplication symbol for packaging counts

Freso:
Rather than using the letter “x” which has multiple different meanings and connotations in different contexts, use the Unicode symbol “×” which means multiplication—which is exactly the meaning of “x” used here.

We already use proper Unicode symbols for various dashes (e.g., em‐dashes (—) in ingredient lists, en‐dashes (–) in page titles), so makes sense (to me) to also do so here.


stephanegigandet (approved):
Looks good, thank you!

PR #11859, opened 2025-05-06
fix: Pin Test2::Harness at <2

### LINKED ISSUE #11858: Tests are hanging after the first unit test has been completed
### What
- Running tests individually works fine but make tests hangs after the first test

1. make tests

If you get a shell inside the backend container then it seems to work fine.

Suspect this may be an issue with a newer version of Test2::Harness

john-gom:

### What

It looks like Test2::Harness version 2 has introduced some breaking changes for us

- Fixes #11858

### DIFF
--- a/cpanfile
+++ b/cpanfile
@@ -146,8 +146,8 @@ on 'develop' => sub {
   requires 'Devel::Cover';
   requires 'Devel::Cover::Report::Codecov';
   requires 'Devel::Cover::Report::Codecovbash';
-  requires 'Test2::Harness';
-  requires 'Test2::Harness::Renderer::JUnit';
+  requires 'Test2::Harness', '<2'; # Seems to be a problem with newer versions in Docker. See #11858
+  requires 'Test2::Harness::Renderer::JUnit', '<2'; # As above
 };
 
 feature "off_server_dev_tools", "Optional development tools" => sub {


cpanfile
```diff
@@ -146,8 +146,8 @@ on 'develop' => sub {
   requires 'Devel::Cover';
   requires 'Devel::Cover::Report::Codecov';
   requires 'Devel::Cover::Report::Codecovbash';
-  requires 'Test2::Harness';
-  requires 'Test2::Harness::Renderer::JUnit';
+  requires 'Test2::Harness', '<2'; # Seems to be a problem with newer versions in Docker. See #11858
```
hangy, line 149: The latest version I can find on CPAN is 1.000157 from 2025-05-04. Pinning CPAN modules in general does sound like a  good idea anyways (as they often don't use SemVer and can introduce breaking changes in minor version). I just don't understand how this change helps with #11858.

jayaddison: Before one of our [failed Perl tests runs](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/14832596802/job/41638320538) (6h timeout), [`Test2-Harness-2.000005` was installed](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/14832596802/job/41637025795#step:4:7156).

There's a CPAN note about a [botched trial release](https://metacpan.org/dist/Test2-Harness/changes#L3) -- perhaps the 2.x version was only available temporarily.

hangy: Yep that could be possible. I don't know how CPAN handels retracted versions

hangy (approved):

jayaddison (approved):

PR #11861, opened 2025-05-07
fix: fix redirection for owner product list in product page

### LINKED ISSUE #11860: Clicking on owner name redirects to same product page instead of owner product list
Image

RayanGoubet:
### What

Correction of the redirection of owner product in product page

Related to #11860 
- Fixes #11860

### DIFF
--- a/templates/web/pages/product/product_page.tt.html
+++ b/templates/web/pages/product/product_page.tt.html
@@ -148,7 +148,7 @@
             <!-- owner -->
             [% IF owner %]
                 <p>
-                    [% FILTER format(lang("sources_manufacturer")) %]<a href="facets/editors/[% owner %]">[% owner_org.name %]</a>[% END %]
+                    [% FILTER format(lang("sources_manufacturer")) %]<a href="/facets/editors/[% owner %]">[% owner_org.name %]</a>[% END %]
                     
                     [% IF owner_org.customer_service.email || owner_org.customer_service.link || owner_org.customer_service.phone %]
                             - [% lang("customer_service") %][% sep %]:


templates/web/pages/product/product_page.tt.html
```diff
@@ -148,7 +148,7 @@ <h2 class="title-1" property="food:name" itemprop="name">[% title %]</h1>
             <!-- owner -->
             [% IF owner %]
                 <p>
-                    [% FILTER format(lang("sources_manufacturer")) %]<a href="facets/editors/[% owner %]">[% owner_org.name %]</a>[% END %]
+                    [% FILTER format(lang("sources_manufacturer")) %]<a href="/facets/editors/[% owner %]">[% owner_org.name %]</a>[% END %]
```
jayaddison, line 151: The fixup to add the root to the URL path looks good; one extra idea: to enable localization, I think we might be able to use:

```suggestion
                    [% FILTER format(lang("sources_manufacturer")) %]<a href="/facets[% canonicalize_taxonomy_tag_link("editors", owner) %]">[% owner_org.name %]</a>[% END %]
```

Idea based on a pattern noticed at: https://github.com/openfoodfacts/openfoodfacts-server/blob/ebc8ab252c45013e96214e4cb77d011a7c05bcd7/templates/web/common/site_layout.tt.html#L168

stephanegigandet: Thanks for the PR @RayanGoubet , can you test @jayaddison 's suggestion and commit it? Thanks!

RayanGoubet: I'll try that !

stephanegigandet (approved):

templates/web/pages/product/product_page.tt.html
jayaddison: Thank you @RayanGoubet!

jayaddison:
The fix looks good to me in production (v2.67.0).  Tested for both `world.` and `es.`.

Oddly the translation of `editors` has an initial-capital in Spanish; that is an unrelated issue though, and hopefully an easy fix.

PR #11864, opened 2025-05-07
fix: Tests fail when run locally

john-gom:
It looks like the pre-loading of data with `-PProductOpener::LoadData` doesn't work with Coverage enabled (which is what the GitHub action uses), but it does work locally. This causes inconsistent results.

This PR removes the pre-loading and also includes the following:

 - Limit CPU_COUNT to 8 as otherwise can run out of memory
 - removed a large amount of "noise" generated from `print STDERR`


alexgarel (approved):
Thanks !

PR #11870, opened 2025-05-09
fix: change the reset preference message depend on flavor

### LINKED ISSUE #11228: Open Beauty Facts - Personal Search - "Use default preferences Nutri-Score, Eco-Score and food processing level (NOVA)"
### What
- "Use default preferences Nutri-Score, Eco-Score and food processing level (NOVA)" in the Edit preferences panel
- ex: https://world.openbeautyfacts.org/product/4210201410683/oral-b

Image

RayanGoubet:
### What

Change de reset preference message by an adapted message following the flavor, add a translation for the obf.
New function on product-preferences.js for take the flavor by the current url.

Related to #11228 
- Fixes #11228


html/js/product-preferences.js
```diff
@@ -23,6 +23,32 @@ function get_user_product_preferences() {
 	return user_product_preferences;
 }
 
+function getFlavorFromHostname() { // get the flavor by the url
```
stephanegigandet, line 26: This probably works, but I would prefer that we pass the the flavor from the server, so that we don't have to update the JS code when we add new flavors, or change preferences for OPF or OPFF.

See the existing mechanism to pass variables from server to JS here: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Display.pm#L4435

html/js/product-preferences.js
```diff
@@ -23,6 +23,32 @@ function get_user_product_preferences() {
...
+    }
+    
+    return "off"; // default 
+}
+
+var flavor = getFlavorFromHostname();
+
+var reset_message;
+
+if (flavor === "off") { // get the reset message depend on the flavor
```
stephanegigandet, line 43: Instead of having logic like this, I think it would be better to have logic like:
if we have a string with the id reset_preferences_details + product_type (it's better to use the "food" and "beauty" product type instead of the flavors "obf" and "off" as at some point we might merge websites) -> use it
otherwise, if there is no string defined (and not empty), use the default message.

html/js/product-preferences.js
```diff
@@ -23,32 +24,18 @@ function get_user_product_preferences() {
 	return user_product_preferences;
 }
 
-function getFlavorFromHostname() { // get the flavor by the url 
-    var host = window.location.hostname;
+var type_product;
```
stephanegigandet, line 27: Can you rename it to product_type? that's the variable name we use in other parts of the code

html/js/product-preferences.js
```diff
@@ -23,32 +24,18 @@ function get_user_product_preferences() {
...
-    return "off"; // default 
-}
-
-var flavor = getFlavorFromHostname();
-
-var reset_message;
-
-if (flavor === "off") { // get the reset message depend on the flavor
-    reset_message = lang().reset_preferences_details;
+if (flavor === "off") { // get the type of product
```
stephanegigandet, line 29: Instead of passing the flavor to the script, you can directly pass product_type, that way you don't need this logic.

po/common/common.pot
```diff
@@ -7471,6 +7471,22 @@ msgctxt "scan"
...
+
+msgctxt "reset_preferences_details_food"
+msgid "Nutri-Score, Green-Score and food processing level (NOVA)"
+msgstr "Nutri-Score, Green-Score and food processing level (NOVA)"
+
+msgctxt "reset_preferences_details_product"
+msgid "Nutri-Score, Green-Score and food processing level (NOVA)"
+msgstr "Nutri-Score, Green-Score and food processing level (NOVA)"
+
+msgctxt "reset_preferences_details_pet_food"
```
stephanegigandet, line 7486: for pet food, we just have organic and fair trade:

image

po/common/common.pot
```diff
@@ -7479,6 +7479,14 @@ msgctxt "reset_preferences_details_food"
 msgid "Nutri-Score, Green-Score and food processing level (NOVA)"
 msgstr "Nutri-Score, Green-Score and food processing level (NOVA)"
 
+msgctxt "reset_preferences_details_product"
```
stephanegigandet, line 7482: for products, we should just put "Good repairability"

stephanegigandet (approved):
Thank you!

PR #11875, opened 2025-05-10
ci: check for link rot in static resources

JagjeevanAK:
### What

This aims at checking that no link nor image is broken in the content pages.
The action is currently stuck due to the sheer size of the arglist
It should only be run periodically
We could have another action that only runs on modified files

### Related PR

It's a part of testing the existing PR by @teolemon  #5744


.github/workflows/link-rot-on-push.yml
```diff
@@ -0,0 +1,49 @@
...
+          file_list="CONTRIBUTING.md,README.md,${html_files// /,}"
+          echo "::set-output name=all_files::$file_list"
+          echo "$file_list"
+
+      - name: Run link check
+        if: steps.get_files.outputs.all_files != ''
+        uses: paramt/url-checker@master
+        id: check
+        with:
+          files: ${{ steps.get_files.outputs.all_files }}
```
alexgarel, line 39: Why do we use this action here and lycheeverse/lychee-action@v2.4.1 ir link-rot.yml ?

JagjeevanAK: both are same and I am adding some changes to it to bring consistancy.

PR #11879, opened 2025-05-12
taxonomy: Sort brands.txt

Freso:
The current `brands.txt` is _almost_ fully sorted alphabetically. This commit now commits to this and sorts all the entries, as per `sort -bi` in the `C.UTF-8` locale. This ignores case (`-f`) and leading whitespace (`-b`).

I also tried with the `--dictionary-order` and `--ignore-nonprinting` options, but these seem to fully ignore non-ASCII characters, leading to a less human-sensible ordering (e.g., “Änglamark” getting sorted with the n’s as “nglamark”).

It also adds a test in the `Makefile` to check for the brand ordering going forward.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,7 +244,7 @@ import_sample_data: run_deps
 	else \
 	 	echo "🥫 Not importing sample data into MongoDB (only for po_off project)"; \
 	fi
-	
+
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Thanks!

PR #11884, opened 2025-05-13
feat: Switch to using JSON for serialising Perl data rather than using Storable STO files

### LINKED ISSUE #11872: Switch to using JSON for serialising Perl data rather than using Storable (STO) files
### Problem
Currently the main data source in Open Food Facts, namely Products, is stored in STO files which are proprietary to Perl. This makes them difficult to access / manipulate in other languages.

### Proposed solution
Switch to using JSON for object serialisation. [Benchmarks](https://www.perlmonks.org/?node_id=11134502) would suggest that it can be even faster than using Storable and initial tests indicate that file size may also be slightly reduced.

### Additional context
This also might allow tools like the Taxonomy editor to pre-build the taxonomy files that Product Opener uses, reducing Product Opener build times

### TLDR
Comments below list benchmarks that were carried out on product and taxonomies. My conclusions from these are:
1. YAML parsers are still not as fast as JSON so not a suitable option
2. JSON::MaybeXS is faster and produces smaller files than STO if sorting and indenting are disabled, so this should be suitable for Products
3. If indenting and sorting are enabled (preferable for Taxonomies) then JSON is about 6% slower than STO for reads and takes up about 8% more space, which should be acceptable for Taxonomies as they are only read at startup.

john-gom:
### What

Use JSON format for Perl object serialization so that this data is more accessible to other platforms.

Have currently implemented for taxonomies and products. General approach:
 - Use formatted JSON for taxonomies and lean format for Products
 - Moving all logic that interacts with the file system into Store.pm (so that we could move to a non-file based approach more easily in the future)
 - File extensions are not specified by the caller
 - STO files are converted to JSON when they are written to
 - "Silent" updates (not creating a new version) to a symlink will stay in STO format

- Fixes #11872


lib/ProductOpener/Store.pm
```diff
@@ -255,6 +276,245 @@ sub retrieve ($file) {
...
 
+# Serializes an object in our preferred object store, removing it from legacy storage if it is present
+sub store_object ($path, $ref, $delete_old = 1) {
+	my $sto_path = $path . '.sto';
+	my $file_path = $path . '.json';
+	# If the file already exists then we need to first open it non-destructively so that
+	# other code doesn't read an empty file before we have written the data
+	my $READ_LOCK;
+	if (-e $file_path) {
+		open(my $READ_LOCK, "<", $file_path);
```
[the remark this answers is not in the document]
john-gom: You are correct, I had missed the flocking $READ_LOCK :-). I have also clarified the comments

lib/ProductOpener/Products.pm
```diff
@@ -1251,9 +1251,17 @@ sub store_product ($user_id, $product_ref, $comment) {
 			$product_ref->{_id} = $product_ref->{code} . '';    # treat id as string;
 		}
 
-		if (    (!-e "$BASE_DIRS{PRODUCTS}/$path")
-			and (!-e "$BASE_DIRS{PRODUCTS_IMAGES}/$path"))
-		{
+		if (object_exists("$BASE_DIRS{PRODUCTS}/$path")) {
```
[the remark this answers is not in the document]
john-gom: I've created a separate `object_path_exists` function for this now (could have made `object_exists` cope with directories too but I though a separate function would be clearer)

john-gom:
@stephanegigandet I was hoping I could remove the store_json function and replace with store_config as the former is only now used for packaging stats. However, there is a note that store_json does not use indenting (although it does sort keys). Do you think adding the single space indenting will be a significant problem? If you can let me know where I can get the current files from then I can work out what the increase in size would be

lib/ProductOpener/Store.pm
```diff
@@ -255,50 +277,269 @@ sub retrieve ($file) {
...
-	if (!-e $file) {
-		return;
+=head2 retrieve_object($path)
+
+Fetch the JSON object from storage and return as a hash ref. Reverts to STO file if no JSON file exists
+
+=cut
+
+sub retrieve_object($path) {
+	my $file_path = $path . '.json';
```
jayaddison, line 366: I'd be inclined _not_ to require the complete filename/path - e.g. not to automatically append `.json` to the file to open.

Reasoning: it can make it a lot easier (and less confusing) either operationally or during development to track down where in the code a file that appears in a log message / on disk is referenced from in the code.

scripts/migrations/2025_05_convert_sto_to_json.pl
```diff
@@ -0,0 +1,78 @@
...
+
+use ProductOpener::PerlStandards;
+use utf8;
+
+use ProductOpener::Paths qw/%BASE_DIRS/;
+use ProductOpener::Store qw/object_iter retrieve_object store_object/;
+
+# This script converts all product sto files to json
+
+my ($checkpoint_file, $last_processed_path) = open_checkpoint('2025_05_convert_sto_to_json.tmp');
```
jayaddison, line 31: NB: this file was accidentally committed to source control on the branch.  I wonder whether we could update `open_checkpoint` to place checkpoint files within `/mnt/podata/tmp`?

lib/ProductOpener/Store.pm
john-gom: Hi @jayaddison , thanks for the feedback. My reasoning for this is as follows:

Firstly, on a practical level... During the migration the file could be an sto or json file, so it could be even more confusing if the code is requesting a .json file and the data is being fetched from an .sto file.

The second reason is more philosophical in that the Store.pm is acting like a database repository layer so the caller doesn't necessarily know how the data is being serialized (it could potentially be going to a service like MinIO or PostgreSQL in the future). All the caller cares about is saving a hash reference and then getting it back again.

However, I can certainly review the logs generated in Store.pm to make sure the full file name is being given on any errors logged here. Does that sound OK?

scripts/migrations/2025_05_convert_sto_to_json.pl
john-gom: It wasn't an accident, but just not something to be run just yet. I have noticed we have a lot of scripts that use checkpoints (and those that don't possibly should) so it might be a good idea to centralise this logic somewhere. Unfortunately not something I'll be able to get to before the Perl hack weekend

lib/ProductOpener/Store.pm
jayaddison: > Firstly, on a practical level... During the migration the file could be an sto or json file, so it could be even more confusing if the code is requesting a .json file and the data is being fetched from an .sto file.

Will the code be reduced to handle a single file type (`.json`) after the migration is completed?  I noticed a few datatypes that do not appear to be involved in the migration (yet?).



> However, I can certainly review the logs generated in Store.pm to make sure the full file name is being given on any errors logged here. Does that sound OK?

This may depend on the answer to my previous question here :)

scripts/migrations/2025_05_convert_sto_to_json.pl
jayaddison: Ok, got it.  I did wonder after commenting whether I might have been mistaken and that it could have been intentional for some reason.

From what I remember: the checkpoint _code_ is reasonably centralised already, but the deployment configuration seems variable (e.g. some of the scripts rely on being started from a particular working directory, which does not seem ideal).

lib/ProductOpener/Store.pm
john-gom: Yes, the plan is to move entirely to JSON, but in the longer term we might not be storing it in the file system anymore

jayaddison: Ok, thanks - I'm still trying to weigh up my feelings about it.  Was using the existing `retrieve` method and using the file extension to determine the loading method considered?

e.g. during migration it would be possible for this code to be valid:

```pl
my $rev_product_ref = retrieve("$BASE_DIRS{PRODUCTS}/$path/$rev.sto");
my $changes_ref = retrieve_object("$BASE_DIRS{PRODUCTS}/$path/changes.json");
```

jayaddison: Drats - that should be `retrieve(...)` on both lines.

scripts/revert_changes_from_user.pl
```diff
@@ -181,38 +179,29 @@
...
-		# keep a copy of current product
-		my $cmd = "mv $BASE_DIRS{PRODUCTS}/$path/product.sto $BASE_DIRS{REVERTED_PRODUCTS}/$target";
-		print STDERR "$code - $cmd\n";
-		# move does not work for symlinks on different file systems
-		#move("$BASE_DIRS{PRODUCTS}/$path/product.sto", "$BASE_DIRS{REVERTED_PRODUCTS}/$target") or die "Could not execute $cmd : $!\n";
+		# Just remove the original product link (no point copying it)
+		print STDERR "$code - removing product\n";
 		if (not $pretend) {
-			(system($cmd) == 0) or die "Could not execute $cmd : $!\n";
+			remove_object("$BASE_DIRS{PRODUCTS}/$path/product");
```
jayaddison, line 185: This seems to be a change-in-behaviour - maybe worth relocating it into a separate PR?

lib/ProductOpener/Store.pm
john-gom: I wanted to have a unique function name so it's easier to search and replace the future. Harder to do if you have to look at the file name expression

scripts/revert_changes_from_user.pl
john-gom: Sorry, that wasn't my intention. Looking above I think it should be a move

tests/unit/store.t
```diff
@@ -49,4 +53,197 @@ is(get_string_id_for_lang("el", "string with spaces"), "string-with-spaces");
...
+# Open an orphaned link, e.g. product.sto link points to a revision that has already been converted to JSON
+remove_object("$test_path-product_revision");
+store_object("$test_path-current_revision", {id => '1'});
+symlink("$test_name-current_revision.sto", "$test_path-product_revision.sto");
+is(retrieve_object("$test_path-product_revision"), {id => '1'}, "Data is fetched from the JSON file");
+
+# If we save back to the link it creates a JSON link to the target file
+store_object("$test_path-product_revision", {id => '2'});
+ok(-l "$test_path-product_revision.json", "New path stays as a link");
+is(retrieve_object("$test_path-current_revision"), {id => '2'}, "Target file is updated");
```
jayaddison, line 124: After this step, could we load both the STO and JSON versions and confirm that their objects are equal, as an additional test assurance?  (do we support that type of object equality check?)

lib/ProductOpener/Store.pm
```diff
@@ -268,24 +268,39 @@ sub retrieve ($file) {
...
-		flock($OUT, LOCK_EX);
-		my $json = $json_for_objects->encode($ref);
-		# Strip out any nul characters as many parsers can't cope with these
-		# This doesn't seem to add too much overhead
-		$json =~ s/\000//g;
-		print $OUT $json;
-		close($OUT);
+	# If the file already exists then we need to first open it non-destructively so that
+	# other code doesn't read an empty file before we have written the data
+	my $READ_LOCK;
```
jayaddison: Interesting: is this not required for STO files because an empty file in that format would be considered invalid (and cause an error) in the read process? (or some other reason?)

lib/ProductOpener/Store.pm
```diff
@@ -369,15 +391,18 @@ sub link_object($path, $link) {
 	symlink($path . '.json', $link . '.json')
 		or $log->error("could not link", {source => $path, link => $link, error => $!});
 
+	# We normally delete a link before creating a new one but just in case make sure there is no STO link
+	unlink($link . '.sto');
```
jayaddison, line 511: Would it be worth an error/warning message here if the operation is successful?  (e.g. unexpected leftover `.sto` file found, and was deleted)

lib/ProductOpener/Store.pm
```diff
@@ -368,12 +387,35 @@ sub retrieve_object_json($path) {
 	return $json_for_objects->encode(retrieve($path . '.sto'));
 }
 
+
+=head2 object_exists($path)
+
+Indicates whether an object (STO or JSON) exists at the specified path
```
jayaddison, line 431: As far as I can tell: `object_exists` is used solely for `change_product_code`.  I am trying to think of any edge cases around this (e.g. for some reason a `.sto` file exists where we have not yet written a JSON file -- so we would consider that the object does exist, but we have not migrated it to the latest file format yet).

lib/ProductOpener/Store.pm
```diff
@@ -255,50 +277,269 @@ sub retrieve ($file) {
...
+=cut
 
-use it as an iterator:
-my $iter = sto_iter(".");
-while (my $path = $iter->()) {
-	# do stuff
+sub link_object($name, $link) {
+	# If target is a sto file then keep the link as a sto file too. Note we use relative paths for the target file
+	#11901: Remove $serialize_to_json test once production is migrated
+	if (not $serialize_to_json or -e dirname($link) . '/' . $name . '.sto') {
```
jayaddison, line 502: Should we definitely not add symlinks for existing STO files after migration begins?  I don't yet totally understand what the symlinks are for -- so maybe this is correct, but I figured I'd ask.

lib/ProductOpener/Store.pm
jayaddison: Ok.  I think what's comfortable for your workflow will be best - and after reading the commits in a bit more detail, I am not so concerned about this.  The log messaging fixup would certainly be good though :+1:

jayaddison:
Taking a review pause now - I'm becoming familiar with the code, and generally it looks good to me, but as long as there is no hurry, I think I'd like to take another review pass next week and to figure out more of the migration read/write steps that will take place (and perhaps more about the symlinks, too).

lib/ProductOpener/Store.pm
john-gom: For sto files we use the lock_store function which handles all of the locking for us, we need to do explicit locking for JSON as there isn't an equivalent function in the JSON module. The reason it is a bit convoluted to do it in Perl code is because flock only works once you have opened a file, but opening a file for writing will empty that file so there is a small window between us opening a the file (and hence clearing it out) and applying the lock where another process might try to read the file.

2025_05_convert_sto_to_json.tmp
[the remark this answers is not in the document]
john-gom: Sorry, yes

scripts/revert_changes_from_user.pl
john-gom: Actually I've just had another look at the code. When we move a revision the file is renamed to something like 123_456_7890_1.sto (where the last "1" is the revision). However, the product.sto symlink will point to just "1.sto" so moving the existing product.sto file would just create a broken symlink so there doesn't seem any point in this. @stephanegigandet what do you think?

scripts/migrations/2025_05_convert_sto_to_json.pl
jayaddison: Ah: I mistakenly opened this thread on the script, instead of the `.tmp` file.  You did resolve my concerns about it though :)  Please consider this resolved.

lib/ProductOpener/Store.pm
jayaddison: Thank you!

lib/ProductOpener/Store.pm
jayaddison: Could/should we deploy the locking change around the existing `.sto` files, independently of the JSON migration, to confirm that it is reliable?

john-gom: It doesn't work the same. `lock_store` does all of the file management / locking internally.

jayaddison: Understood - my line of thinking was that we could deploy the new locking mechanism independently and before the JSON migration; it would involve removing the use of `lock_store`.

And to explain that: I am anxious about deploying two changes (1. file locking change, 2. data format change) that I think are complicated and risky within a single release, because I think it could make it more difficult to identify problems and/or rollback.

john-gom:
I'm sensing some reluctance in merging this PR so I'm wondering if it would be better to operate in parallel for a bit, so if the SERIALIZE_TO_JSON option is not set then the code will treat the STO files as the primary source of truth, but start also duplicating the data in JSON files as products are updated. When the flag is set it will start to delete the STO files as products are updated, with a migration script to convert everything, when we are ready. Thoughts @stephanegigandet , @alexgarel , @jayaddison , @hangy ?

lib/ProductOpener/Store.pm
john-gom: I don't believe we are changing the locking behaviour. lock_store will basically lock the file, write to it and then release the lock. The only reason that we have to do it in a slightly more convoluted way for JSON is that we have to open the file for reading before we can put a lock on it, we can then re-open for writing to ensure nothing accesses the file part way through the write process.

tests/unit/store.t
john-gom: Now implemented for Level 1

lib/ProductOpener/Store.pm
john-gom: I've changed the locking to open in append mode, lock and then truncate, the same as lock_store. Also reviewed the error handling to be consistent, this being that store_object will die on error (as I think that's what lock_store would do) but retrieve_object will just carp as the existing retrieve wrapper in Store.pm catches and carps lock_retrieve errors.

lib/ProductOpener/Store.pm
```diff
@@ -308,22 +549,23 @@ sub sto_iter ($initial_path, $pattern = qr/\.sto$/i) {
 					# avoid ..
 					next if $file =~ /^\.\.?$/;
 					# avoid conflicting-codes and invalid-codes
-					next if $file =~ /^(conflicting|invalid)-codes$/;
+					next if $exclude_path_pattern and $file =~ $exclude_path_pattern;
```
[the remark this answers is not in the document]
john-gom: OK, I've added a product_iter function which does the default exclusions so hopefully people will see to use that

lib/ProductOpener/Store.pm
```diff
@@ -332,4 +607,55 @@ sub sto_iter ($initial_path, $pattern = qr/\.sto$/i) {
...
+JSON keys are sorted and indentation is used so files can be used in source control
+No locking is performed
+
+=cut
+
+sub store_config ($path, $ref, $delete_old = 1) {
+	my $sto_path = $path . '.sto';
+	my $file_path = $path . '.json';
+
+	#11901: Config files aren't shared so ignore get_serialize_to_json_level() flag. Just remove this comment when migration is complete
```
hangy, line 656: I might be missing something, but as `store_config` is needed to build taxonomies, should it not write both file types for `get_serialize_to_json_level() > 0`?

lib/ProductOpener/Checkpoint.pm
```diff
@@ -0,0 +1,73 @@
...
+	}
+	open(my $checkpoint_file, '+<', $filename) or die "Could not open file '$filename' $!";
+	my $checkpoint;
+	my $is_resume = first_index {$_ eq "resume"} @ARGV;
+	if ($is_resume > -1) {
+		seek($checkpoint_file, 0, 0);
+		$checkpoint = <$checkpoint_file>;
+		chomp $checkpoint if $checkpoint;
+		if ($checkpoint) {
+			print STDERR "Resuming from $checkpoint\n";
```
hangy, line 45: Is it assumed that the `Checkpoint` class is always instantiated in a CLI script? Can we check the environment to avoid using it in an unsupported one?

tests/unit/store.t
```diff
@@ -49,4 +54,276 @@ is(get_string_id_for_lang("el", "string with spaces"), "string-with-spaces");
...
+
+	# Make sure json file doesn't exist
+	remove_object($test_path);
+	# Create an initial test file
+	ensure_dir_created_or_die($test_root_path);
+	lock_store({id => 1}, "$test_path.sto");
+
+	ok(object_exists($test_path), "object_exists should recognize sto files");
+
+	is(retrieve_object("$test_path"), {id => 1}, "Verify retrieve copes with a sto file");
```
hangy: Is there a specific reason for this to be use interpolated strings?

lib/ProductOpener/Store.pm
john-gom: I wasn't thinking that the implementation levels would apply to build artefacts as these should get sufficient testing through the current unit and integration tests.

lib/ProductOpener/Checkpoint.pm
john-gom: I guess we could but is it really a concern? I'd rather not add further complexity if it can be avoided.

tests/unit/store.t
john-gom: No. Fixed.

lib/ProductOpener/Store.pm
jayaddison: Although I'm not confident that the logging semantics are the same between this and the Perl Storable implementation -- the updated series of file-related syscalls (open, flock) does look good to me, thank you :+1: 

Are you planning to re-run the benchmarking process?  I am wary that these and perhaps other recent changes could have invalidated the previous results.

jayaddison:
@john-gom I'm hoping to find some time on Wednesday (2025-06-04) to test this using a local development instance.  I don't have a detailed test plan for that yet, but based on what I understand about the feature, I'll gradually increase the serialization level (0, 1, 2) and perform various manual reads/writes of products, and if possible also taxonomy items, using the web ProductOpener interface.  At each stage I will stop to inspect the filesystem contents for `.sto` files, `.json` files, and symlinks as appropriate.

lib/ProductOpener/Checkpoint.pm
hangy: Nah more of a nitpick 😅

lib/ProductOpener/Store.pm
hangy: OK so taxonomies get migrated to JSON directly? Works for me. 👍

lib/ProductOpener/Store.pm
john-gom: Hi @jayaddison , yes, I re-rean the benchmarks with the new locking and results are still comparable

jayaddison:
> * "Silent" updates (not creating a new version) to a symlink will stay in STO format

@john-gom can you recommend a way to make a silent update occur for an individual product?  I can try to include that as a scenario during testing if so.

lib/ProductOpener/Store.pm
jayaddison: Brilliant - thank you!

jayaddison:
> @john-gom I'm hoping to find some time on Wednesday (2025-06-04) to test this using a local development instance. I don't have a detailed test plan for that yet, but based on what I understand about the feature, I'll gradually increase the serialization level (0, 1, 2) and perform various manual reads/writes of products, and if possible also taxonomy items, using the web ProductOpener interface. At each stage I will stop to inspect the filesystem contents for `.sto` files, `.json` files, and symlinks as appropriate.

I've completed this round of testing.  The scope of the testing was:

  * I tested baseline/existing behaviour and then each of the serialization levels, after documentation some of my expectations at each level.
  * I experimented primarily by editing a single product in my test OFF dev environment, and inspecting the filesystem contents under `/mnt/podata/products/...` in the backend container.
  * I performed a rebuild of the taxonomy `build-cache` -- but aside from that did not edit taxonomy data.

Although I could report on more of the results, I will firstly report back with the only **unexpected** result I found:

  * At serialization `level 0`, `product.json` symlinks were created for the current product version during each product edit.  Additionally, because the corresponding `<version-number>.json` files were _not_ created, these symlinks were broken (could not be followed to an openable file).

Aside from that, results were good: editing products produced the expected effects on the filesystem, historic product revisions could be viewed as expected at each serialization level, and STO files continued to be considered the source-of-truth at serialization level 1 (providing the possibility for rollback and cleanup of JSON files, if required).

Most of the testing was performed completely offline (without an Internet connection) - as a result I have a small suggested edit for our `openfoodfacts-server.git` `Makefile`, and I'll open that as a separate pull request.

jayaddison:
Note: testing was performed from commit efa0576ab28a40c9a9f0167fdf92aa3520a8a9ae and compared to existing mainline behaviour at commit 54d6f3c1d49dfc92f17fd3fcb5a72d301b8765b5.  The `SERIALIZE_TO_JSON` value was adjusted between each test by using the `.env` file.

Edit: fixup: baseline/test commits were documented the wrong way around.

john-gom:
> > * "Silent" updates (not creating a new version) to a symlink will stay in STO format
> 
> @john-gom can you recommend a way to make a silent update occur for an individual product? I can try to include that as a scenario during testing if so.

The update_all_products script is the only place I have come across it. Logic is here: https://github.com/openfoodfacts/openfoodfacts-server/blob/54d6f3c1d49dfc92f17fd3fcb5a72d301b8765b5/scripts/update_all_products.pl#L1470

john-gom:
> * At serialization `level 0`, `product.json` symlinks were created for the current product version during each product edit.  Additionally, because the corresponding `<version-number>.json` files were _not_ created, these symlinks were broken (could not be followed to an openable file).

I have fixed this. I hadn't realised that "-e" on a symlink will return false if the symlink exists but the referenced file does not

jayaddison:
> > * At serialization `level 0`, `product.json` symlinks were created for the current product version during each product edit.  Additionally, because the corresponding `<version-number>.json` files were _not_ created, these symlinks were broken (could not be followed to an openable file).
> 
> I have fixed this. I hadn't realised that "-e" on a symlink will return false if the symlink exists but the referenced file does not

Thanks! - I'll re-test that specific case within the next few hours.

I will also try to test the rollback behaviour (e.g. ramping the levels back down from 2 to 0).  It only occurred to me a few hours after testing yesterday that that would also be a useful/important scenario to check.

These should both be relatively quick for me to check now that I am more familiar with the feature and process.

jayaddison:
> These should both be relatively quick for me to check now that I am more familiar with the feature and process.

Apologies: this is going to take me longer than anticipated; I have encountered a problem, reported as #12003, when attempting to restart the application from the latest commit (775583e06a758bfc3c8bbfad454286064355f05e) here.

jayaddison:
Both the `po_off-frontend-1` (`nginx` webserver) and `po_off-backend-1` (ProductOpener `perl` web application) are configured as part of a Docker `po_off_default` network, according to their network settings.

And the DNS names `frontend` and `backend` are found as aliases for those containers within those networks:

```sh
docker inspect po_off-backend-1 | jq '.[0].NetworkSettings.Networks.po_off_default.DNSNames'
[
  "po_off-backend-1",
  "backend",
  "<container hexadecimal alias>"
]
```

jayaddison:
(!) Ek, wrong thread.

JagjeevanAK (approved):

jayaddison:
> > These should both be relatively quick for me to check now that I am more familiar with the feature and process.
> 
> Apologies: this is going to take me longer than anticipated; I have encountered a problem, reported as #12003, when attempting to restart the application from the latest commit ([775583e](https://github.com/openfoodfacts/openfoodfacts-server/commit/775583e06a758bfc3c8bbfad454286064355f05e)) here.

I have found a [workaround](https://github.com/openfoodfacts/openfoodfacts-server/issues/12003#issuecomment-2958927923) for this blocker, and will resume testing this afternoon.

jayaddison:
@john-gom from additional testing: almost everything seems to be working, although I did encounter one more quirk: after testing the following two _successfully_:

* Reverting from level 2 to level 1 and editing a product.
  * Viewing the product succeeded, as intended.

* Reverting from level 1 to level 0 and editing a product.
  * Viewing the product again succeeded, as intended.

...I then re-upgraded to level 2, and now I encounter a problem with the revert-user-changes script:

```
www-data@a31cb9cbe979:/opt/product-opener$ ./scripts/revert_changes_from_user.pl --userid testing
[Tue Jun 10 13:59:39 2025] XSLoader.pm: Name "FFI::Platypus::TypeParser::ffi_type" used only once: possible typo at /usr/lib/x86_64-linux-gnu/perl-base/XSLoader.pm line 111.
[Tue Jun 10 13:59:39 2025] XSLoader.pm: Name "FFI::Platypus::keep" used only once: possible typo at /usr/lib/x86_64-linux-gnu/perl-base/XSLoader.pm line 111.
1 products to revert
reverting product 9002355004345
9002355004345 - moving revision 900_235_500_4345_35
Status: 500
Content-type: text/html

<h1>Software error:</h1>
<pre>could not move json file from /mnt/podata/products/900/235/500/4345/35.json to /mnt/podata/reverted_products/900_235_500_4345_35.json, error: No such file or directory at /opt/product-opener/lib/ProductOpener/Store.pm line 525.
</pre>
<p>
For help, please send mail to this site's webmaster, giving this error message 
and the time and date of the error.

</p>
[Tue Jun 10 13:59:44 2025] revert_changes_from_user.pl: could not move json file from /mnt/podata/products/900/235/500/4345/35.json to /mnt/podata/reverted_products/900_235_500_4345_35.json, error: No such file or directory at /opt/product-opener/lib/ProductOpener/Store.pm line 525.
```

```sh
$ env | grep SERIALIZE_TO_JSON
SERIALIZE_TO_JSON=2
```

```sh
www-data@a31cb9cbe979:/mnt/podata/products/900/235/500/4345$ ls -lSr
total 752
lrwxrwxrwx 1 www-data www-data     7 Jun 10 13:58 product.json -> 37.json
-rw-r--r-- 1 www-data www-data   119 Feb  3 12:14 images.sto
-rw-r--r-- 1 www-data www-data   455 Feb  4 16:42 scans.json
-rw-r--r-- 1 www-data www-data  4065 Sep  7  2018 1.sto
-rw-r--r-- 1 www-data www-data  4448 Sep  7  2018 2.sto
-rw-r--r-- 1 www-data www-data  5164 Feb 29  2020 4.sto
-rw-r--r-- 1 www-data www-data  5171 Jan 14  2020 3.sto
-rw-r--r-- 1 www-data www-data  9151 Jan  5  2021 5.sto
-rw-r--r-- 1 www-data www-data 10021 Jan 16  2021 6.sto
-rw-r--r-- 1 www-data www-data 11797 Jun 10 13:58 changes.json
-rw-r--r-- 1 www-data www-data 11862 Jul 22  2021 7.sto
-rw-r--r-- 1 www-data www-data 14328 Aug  8  2021 8.sto
-rw-r--r-- 1 www-data www-data 14526 Aug  8  2021 9.sto
-rw-r--r-- 1 www-data www-data 14530 Aug  8  2021 10.sto
-rw-r--r-- 1 www-data www-data 14711 Aug  8  2021 11.sto
-rw-r--r-- 1 www-data www-data 15043 Aug  8  2021 12.sto
-rw-r--r-- 1 www-data www-data 17050 Feb  2  2022 13.sto
-rw-r--r-- 1 www-data www-data 20277 Jan 22 03:00 14.sto
-rw-r--r-- 1 www-data www-data 20713 Feb  6 17:49 16.sto
-rw-r--r-- 1 www-data www-data 20714 Feb  6 17:49 15.sto
-rw-r--r-- 1 www-data www-data 20858 Feb  6 17:49 17.sto
-rw-r--r-- 1 www-data www-data 21074 Feb  6 17:49 18.sto
-rw-r--r-- 1 www-data www-data 21400 Feb  6 17:49 19.sto
-rw-r--r-- 1 www-data www-data 21594 Feb  6 17:49 20.sto
-rw-r--r-- 1 www-data www-data 21919 Feb  6 17:49 21.sto
-rw-r--r-- 1 www-data www-data 22113 Feb  6 17:49 22.sto
-rw-r--r-- 1 www-data www-data 22436 Feb  6 17:49 23.sto
-rw-r--r-- 1 www-data www-data 22630 Feb  6 17:49 24.sto
-rw-r--r-- 1 www-data www-data 22935 Feb  6 17:49 25.sto
-rw-r--r-- 1 www-data www-data 23129 Feb  6 17:49 26.sto
-rw-r--r-- 1 www-data www-data 23472 Feb  6 17:49 27.sto
-rw-r--r-- 1 www-data www-data 23611 Feb  6 17:49 30.sto
-rw-r--r-- 1 www-data www-data 23611 Feb  6 17:49 28.sto
-rw-r--r-- 1 www-data www-data 23628 Feb  6 17:49 31.sto
-rw-r--r-- 1 www-data www-data 23659 Feb  6 17:49 29.sto
-rw-r--r-- 1 www-data www-data 25117 Feb  6 17:49 33.sto
-rw-r--r-- 1 www-data www-data 25212 Feb  6 17:49 32.sto
-rw-r--r-- 1 www-data www-data 26332 Jun 10 13:58 37.json
-rw-r--r-- 1 www-data www-data 26333 Jun 10 13:48 36.json
-rw-r--r-- 1 www-data www-data 28257 Mar 30 18:40 34.sto
```

jayaddison:
(removing the `product.json` symlink and then adding a `product.sto` symlink to the latest version is a manual fixup for that problem, but doesn't seem ideal)

john-gom:
> ..I then re-upgraded to level 2, and now I encounter a problem with the revert-user-changes script:

I think the issue here is why revision 35.sto / json wasn't there in the first place. What was the level set to when you created that revision and the one after (did you make any notes in the changes.json? @jayaddison )

john-gom:
> > ..I then re-upgraded to level 2, and now I encounter a problem with the revert-user-changes script:
> 
> I think the issue here is why revision 35.sto / json wasn't there in the first place. What was the level set to when you created that revision and the one after (did you make any notes in the changes.json? @jayaddison )

@jayaddison I fixed an issue where move_object should be moving both files if they exist. Also I made sure we always use the STO file as a preference on retrieve if it exists to cater for going up and down levels. I couldn't reproduce your problem before the fix though. The steps I follows were (with the testing user):
- Set the level to 2 and create a revision - new revision files and link are now json
- Set the level to 1 and create a revision - new revision files and link are json and sto
- Set the level to 0 and create a revision - new revision files and link just sto
- Set the level to 2 and revert the changes for the testing user
I observed the issue with some of the old revision files not being moved when I did this but didn't get any errors

jayaddison:
> > ..I then re-upgraded to level 2, and now I encounter a problem with the revert-user-changes script:
> 
> I think the issue here is why revision 35.sto / json wasn't there in the first place. What was the level set to when you created that revision and the one after (did you make any notes in the changes.json? @jayaddison )

That is a good question: I didn't notice that that file was missing during the testing.

I had previously run the revert script at level 2 -- basically to clean up all my temporary edits before downgrading to level 0.

Is it possible that running that script removed the underlying `.sto` file?

I will attempt to figure out the steps required to replicate that soon (perhaps tomorrow afternoon, or Friday at the latest).

john-gom:
> I had previously run the revert script at level 2 -- basically to clean up all my temporary edits before downgrading to level 0.

It might be the fact that move_object wasn't deleting the json file if an sto file existed could have put the data in an odd state when you did that first revert.

jayaddison:
> > ..I then re-upgraded to level 2, and now I encounter a problem with the revert-user-changes script:
> 
> I think the issue here is why revision 35.sto / json wasn't there in the first place. What was the level set to when you created that revision and the one after (did you make any notes in the changes.json? @jayaddison )

> > I had previously run the revert script at level 2 -- basically to clean up all my temporary edits before downgrading to level 0.
> 
> It might be the fact that move_object wasn't deleting the json file if an sto file existed could have put the data in an odd state when you did that first revert.

After reviewing these replies, re-reading my testing notes, and failing to replicate the issue -- I think that a test case I ran at level 1 was most likely to have caused the problem.

I wanted to check that removal of a `json` file would continue to allow the `sto` file to be used as a fallback, so that we could cleanup JSON files if we need to rollback.

To do that, I removed a single `<version>.json` file.  I can't guarantee that it was `35.json`, because I didn't note it (and don't have the shell history), but it seems likely, in retrospect.

I'm trying a few more things to replicate this, but with no luck so far.  Because it was a somewhat-artificial scenario, I don't think this should block this pull request.

Note that to replicate this, I'm continuing to test from commit efa0576ab28a40c9a9f0167fdf92aa3520a8a9ae (what I was using at the time, according to my test notes).  After either finding or giving up on replicating the issue, I'll continue on to 59e0b22cdc956734ef3052244ce75dae3c9d6a02 with the latest fixes.

jayaddison:
Ah, and about the `changes.json` question: I didn't modify any of those files manually, although I did add edit messages when editing each product through the web interface.

There does not appear to be a `changes.json` file in the `/mnt/podata/products/900/235/500/4345` directory currently (I didn't remove it, but it is not present).

jayaddison:
> It might be the fact that move_object wasn't deleting the json file if an sto file existed could have put the data in an odd state when you did that first revert.

This gives me one more idea about how to attempt to replicate the problem -- I'll report back on that soon..

jayaddison:
Nope, sorry; I'm still baffled and unable to recreate the circumstances of the could-not-move-json-file error.

That's unsatisfying and a bit frustrating, but I think it may make sense to move on to testing the latest commits.

Issue #11304, opened 2025-01-27
nutrition facts are not displayed when portion is missing


teolemon:
### What
- nutrition facts are not displayed when portion is missing

stephanegigandet:
@teolemon do you have an example product? was this bug on the product page, or the product edit form? were there nutrition facts indicated per serving? i made a few tests locally and could not reproduce

squeak9:
@stephanegigandet Here's one:

https://world.openfoodfacts.org/product/6416453015432/milk-chocolate-dumle-fazer

- Old revision works fine: https://world.openfoodfacts.org/product/6416453015432/milk-chocolate-dumle-fazer?rev=21
- Current is busted: https://world.openfoodfacts.org/product/6416453015432/milk-chocolate-dumle-fazer?rev=22

On the edit screen it looks fine right now, I bet editing it would unbreak it but I'll let you try and reproduce the error first.

PR #11888, opened 2025-05-15
fix: not displayed nutrition columns #11304

stephanegigandet:
fixes the computation of the nutrition_data and nutrition_data_prepared columns that indicates if we have nutrition data, and also add a check to display columns anyway if we have data (for old versions of products that have a bad value in nutrition_data)

fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/11304

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -9588,12 +9588,17 @@ sub data_to_display_nutrition_table ($product_ref, $comparisons_ref, $request_re
 	my @displayed_product_types = ();
 	my %displayed_product_types = ();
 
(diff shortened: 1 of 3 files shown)


teolemon (approved):
It would also be nice to have a knowledge panel action to input the portion in the right cases.

teolemon:
The following jobs failed:
+--------------------------------------+-------------------------+
| Job ID                               | Test File               |
+--------------------------------------+-------------------------+
| BF676F30-316B-11F0-A7E5-81A9B2547160 | tests/unit/food.t       |
| BF680508-316B-11F0-A7E5-81A9B2547160 | tests/unit/nutriscore.t |
+--------------------------------------+-------------------------+

PR #11893, opened 2025-05-15
docs: precise that user_id is username, not email

alexgarel:
Make it really clear that it's better not to use email to login when using the API.


stephanegigandet (approved):
Thanks

PR #11915, opened 2025-05-22
feat: Add some basic support for interfixes

Freso:
Some languages (notably North Germanic ones) can use -s- or other interfixes to combine words, which is also a frequent occurrence for ingredient lists.

This adds some very basic support for German and Dutch (-s appended to first word or s- prepended to second), Danish (s-, e-, or n- appended to second part), Swedish (s-, e-, o-, u-, or a- appended to second part), and Norwegian (s- or e- appended to second part—at least for Bokmål; couldn’t find a source for Nynorsk).

There are _some_ rules regarding whether some of the interfixes are valid or not depending on surrounding letters, but this is a start and might be enough, since the regular expressions may become too tangled/complex (with look-behinds and what-nots).

Note that the added bits to the regular expressions are using non-greedy matches, so it should prefer to keep any s’s (or other interfixes) already in the ingredient words before trying to match by inserting the interfix.

Sources:
https://en.wikipedia.org/wiki/Interfix
https://da.wikipedia.org/wiki/Bindebogstav
https://sv.wikipedia.org/wiki/Fogemorfem
https://sv.wikipedia.org/wiki/Foge-s
https://no.wikipedia.org/wiki/Fuge-s
https://no.wikipedia.org/wiki/Fuge-e
https://svenska.se/grammatik/ (search “bindesuffix”)


aleene:
Nice

teolemon:

PR #11925, opened 2025-05-24
taxonomy: Attempt at untangling blueberries

Freso:
“Blueberry” is a term that is commonly used for different (albeit similar looking and closely related) berries; namely the Eurasian bilberries (Vaccinum sect. Myrtillus and V. uliginosum) and the North American blueberries (V. sect. Cyanococcus).

While they do have different names in English (bilberry and blueberry), many languages simply use “blueberry” without specific distinction of which berry it is.

This commit makes a “Blueberry jams” category covering both, with child categories for “Bilberry jams” and “American blueberry jams” for when it is known which specific type of “blueberry” was used.

Sources:
https://en.wikipedia.org/w/index.php?title=Bilberry&oldid=1291352354
https://en.wikipedia.org/w/index.php?title=Blueberry&oldid=1291531485

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -103813,7 +103813,13 @@ ciqual_proxy_food_name:en: Jam, strawberry
 ciqual_proxy_food_name:fr: Confiture de fraise (extra ou classique)
 
 < en:Berry jams
-en: Bilberries jams
+en: Blueberry jams
+da: Blåbærsyltetøj
+nb: Blåbærsyltetøy
+sv: Blåbärssylter
+
+< en:Blueberry jams
+en: Bilberry jams, Bilberries jams, Blaeberry jams, Blueberry jams (European)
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -103813,7 +103813,13 @@ ciqual_proxy_food_name:en: Jam, strawberry
 ciqual_proxy_food_name:fr: Confiture de fraise (extra ou classique)
 
 < en:Berry jams
-en: Bilberries jams
+en: Blueberry jams
```
Freso, line 103816: Another approach could maybe be to call this something like “Generic blueberry jam” or something, and then the `en: American blueberry jams` could just be `en: Blueberry jams`?

```suggestion
en: Blueberry jams (generic)
```

stephanegigandet (approved):
Thanks

PR #11937, opened 2025-05-25
docs: Update product_nutrition.yaml with nutrients explainers

teolemon:
### What
docs: Update product_nutrition.yaml with nutrients explainers


JagjeevanAK (approved):

alexgarel (commented):
It's questionable whether we want to add all those fields ? We already have a documentation which is cumbersome to read, do we want to add all those fields that will serve very little ?

PR #11945, opened 2025-05-25
refactor: experiment to store products data in postgresql - WIP

stephanegigandet:
Experimenting with using Postgresql to store product data, instead of using JSON or STO files (done with @cquest during the Perl x OFF Hackathon)

Tested with a databased hosted by Christian.
Basic reading and writing works.

Lots of things to change:
- Docker dev setup of the database
- 2 tables in Postgres, one for the last revision, one for all revisions
- reading / writing a specific revision
- Postgresql triggers to automatically update both tables when we write a product
- checking UTF8 encoding/escaping/storing in Postgres
- using a Postgres table for the changes.sto data, so that we can store individual changes for one revision
- supporting some operations currently done with MongoDB?
- merging with MongoDB?
- etc.


john-gom:
@cquest would it be possible to post details on how the database was configured, e.g. any specific configuration parameters, triggers, indexes, etc.?

stephanegigandet:
A bit more details from the Hackathon notes from Christian:

Christian

performance test to compare reading JSON from postgresql to reading .sto from file system
1.1ms average on JSON/PG
0.9ms average on .sto

setup a test db for Stéphane
connection from ProductOpener
read a product data
write a product data
PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/11945

    Next steps:

check UTF8 encoding   
Re-import actual .sto in a fresh db (check UTF8)
add TRIGGERs to check validity on INSERT/UPDATE
check "rev" for increasing values
update last_update field automatically
extract a few useful fields out of the JSON global payload (last_editor, product_type, more ?)
which index should we add in the JSON... possibly with fulltext search capabilities (tsvector and/or pg_trgm)

PR #11956, opened 2025-05-27
fix: localization: lowercase the Spanish translation of the 'editors' resource string

jayaddison:
### What

In the localization resources under the `po` directory, it seems that all tagnames should be lowercase.

However, there is a tagname in Spanish that begins with an uppercase ASCII character.  It seems to be the only one at the moment:

```sh
$ grep -rw -A 1 'msgid "editors"' po/tags | grep msgstr | grep '\.po-.*[A-Z]'
```

This pull request modifies it to be lowercase too, for consistency.

- Discovered / noticed by https://github.com/openfoodfacts/openfoodfacts-server/pull/11861#issuecomment-2912151272

### DIFF
--- a/po/tags/es.po
+++ b/po/tags/es.po
@@ -117,7 +117,7 @@ msgstr "depuración"
 #. part of an url, put here the plural in your language, without caps, and by using hyphens instead of spaces
 msgctxt "editors:plural"
 msgid "editors"
-msgstr "Editores"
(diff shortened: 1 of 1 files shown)


jayaddison:
Thanks @RayanGoubet!

stephanegigandet (approved):
Thanks

PR #11982, opened 2025-06-01
chore(dev): Update devcontainer to use different dir as home

hangy:
The previous home was `/var/www/`, which the vscode user wasn't allowed to write to.


alexgarel (approved):

alexgarel:
Thanks @hangy !
Could you write a small documentation about how to use this devcontainer in vscode ?

PR #11983, opened 2025-06-01
chore(test): Remove old and unused test along with unused testing npm packages

hangy:
The old integration test hasn't been run in ages, but we were still having dependabot update the packages unnecessarily.


stephanegigandet (approved):
Thanks

PR #11996, opened 2025-06-04
build: allow stale dependency git repo clones

jayaddison:
### What

During a recent manual QA/testing process for #11884, I was working mostly offline (with my laptop's Internet connection turned off) in order to reduce distractions.

Attempting to restart the local development environment between tests failed at a `git pull` step in the `Makefile` -- and I'd like to suggest that we could make that step optional - with a warning if it fails - provided that the user already has a checkout of the dependency.

### Screenshot
N/A

- This suggestion is based on practical experience from a round of testing described at: https://github.com/openfoodfacts/openfoodfacts-server/pull/11884#issuecomment-2939482998

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -561,7 +561,9 @@ clone_deps:
 				https://github.com/openfoodfacts/$$dep.git ${DEPS_DIR}/$$dep; \
 			echo "Cloned $$dep"; \
 		else \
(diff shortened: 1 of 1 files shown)


jayaddison:
Thank you, @hangy!

PR #11997, opened 2025-06-04
fix: better error handling on export scripts

alexgarel:
We need to be notified when exports failed in some way.
Current error handling was not complete.

Also removed unused gen_madenearme service definition.

Add timeout to service to avoid having it hanged (as it happened this week).

### DIFF
--- a/scripts/error_report_utils.sh
+++ b/scripts/error_report_utils.sh
@@ -17,7 +17,12 @@ init_error_report () {
 function report_error () {
     return_code=$1
     script_name=$2
-    if [ "$return_code" -ne 0 ];
+    if [ -z "$script_name" ] || [ -z "$return_code" ]
+    then
+        >&2 echo "ERROR: report_error called with no arguments"
+        exit 1
+    fi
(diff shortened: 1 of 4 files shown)


alexgarel:
@hangy would you like to look at it ?

scripts/error_report_utils.sh
```diff
@@ -0,0 +1,43 @@
...
+    fi
+}
+
+# If there were commands that resulted in errors,
+# echo the list of commands so that it is included in the
+# failure e-mail sent to root
+function report_failed_commands () {
+    if [ $ERRORS -gt 0 ];
+    then
+        >&2 echo "ERROR: $ERRORS ERROR(S) DURING EXECUTION OF gen_fields_daily.sh"
```
stephanegigandet: just a small detail, this script is also now used by other scripts than gen_feeds_daily.sh like generate_madenearme_pages.sh

alexgarel: Oh yes, thanks !

PR #11998, opened 2025-06-04
fix: add an explanation for the missing value and unit to a product

### LINKED ISSUE #11322: If the serving size is incorrect, the KP should provide a better explanation
Hi!

This is a report from Slack with the following nutrition facts (= the serving size has no unit):
Image

On the KP, the reason is incorrect because all fields are filled:
Image

RayanGoubet:
### What

Add an explanation for the issues that we have a wrong message when we don't have value or unit for a product

- Fixes #11322

### DIFF
--- a/tests/unit/expected_test_results/knowledge_panels/en-nutriscore-serving-size-error.json
+++ b/tests/unit/expected_test_results/knowledge_panels/en-nutriscore-serving-size-error.json
@@ -904,7 +904,7 @@
             {
                "element_type" : "text",
                "text_element" : {
-                  "html" : "\n                For a more precise calculation of the Green-Score, you can modify the product page and add them.\n                <br><br>\n                If you are the manufacturer of this product, you can send us the information with our <a href=\"https://server_domain/a>.\n                    ",
+                  "html" : "\n                For a more precise calculation of the Green-Score, you can edit the product page and add them.\n                <br><br>\n                If you are the manufacturer of this product, you can send us the information with our <a href=\"https://server_domain/a>.\n                    ",
                   "type" : "warning"
                }
             }
@@ -1297,7 +1297,7 @@
                }
             }
          ],
-         "expanded" : true,
+         "expanded" : false,
          "level" : "info",
          "title_element" : {
             "title" : "Ingredient information"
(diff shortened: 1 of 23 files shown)


po/common/common.pot
```diff
@@ -7491,6 +7491,11 @@ msgctxt "action_report_image_to_nutripatrol"
 msgid "Report this image to our moderators"
 msgstr "Report this image to our moderators"
 
+msgctxt "missing_value_and_or_unit"
```
stephanegigandet: It's better to name it missing_serving_size_value_and_or_unit.

po/common/common.pot
```diff
@@ -7491,6 +7491,11 @@ msgctxt "action_report_image_to_nutripatrol"
...
+msgid "We are missing the serving size value and or unit to calculate the NutriScore"
```
stephanegigandet: ``
```suggestion
msgid "We are missing the serving size value and/or unit to calculate the Nutri-Score"
```

stephanegigandet (commented):
Looks good, but we should add a test that show the result of your changes. For instance in test/unit/attributes.t : add a product where we have serving size nutrients, but not the serving size.

tests/unit/attributes.t
```diff
@@ -119,6 +119,50 @@ my @tests = (
...
+		}
+	],
+
+	[
+		'en-nutriscore-2',
+		{
+			lc => "en",
+			categories => "biscuits",
+			categories_tags => ["en:biscuits"],
+			nutrition_data_per_serving => "20",
```
stephanegigandet: Should be nutrition_data_per => "serving", and then you pass serving_size => "20 g" (or "20" or nothing depending on the case you want to test)

stephanegigandet:
I wrongly suggested to add a test in attributes.t : the change you made is not visible in attributes, only in the knowledge panels...

tests/unit/attributes.t
```diff
@@ -119,6 +119,51 @@ my @tests = (
...
+				"fat_100g" => 50,
+				"saturated-fat_100g" => 9.7,
+				"sugars_100g" => 5.1,
+				"salt_100g" => 0,
+				"sodium_100g" => 0,
+				"proteins_100g" => 29,
+				"fiber_100g" => 5.5,
+			},
+		}
+	],
```
alexgarel: you added twice this test (it is already there)
```suggestion
```

alexgarel (approved):
LGTM, apart from a test copied twice.

Thanks !

PR #12005, opened 2025-06-09
feat: add Nutri-Score % of fruits / vegetables / nuts / legumes from GS1

stephanegigandet:
GS1 has added support for some fields that are necessary to compute the Nutri-Score, like adding the % of fruits / vegetables / legumes. We do get some of those from some producers (e.g Bonduelle, but for the old Nutri-Score fields it seems).

This PR adds some support + some comments and links in the code.

There are more fields we could add (which category should be used for the Nutri-Score: fats, nuts etc.). But so far I have not seen it in the data we receive yet.


stephanegigandet:
New fields in GS1 data model:

https://navigator.gs1.org/gdsn/class-details?name=NutritionalProgramIngredientTypeCode&version=12
								
As of 2025/06/01, the nutritionalProgramIngredients values seem to refer to the old Nutri-Score formula
There seems to be new fields for the new formula:

https://www.gs1.nl/kennisbank/gs1-data-source/levensmiddelen-drogisterij/welke-data/nutri-score/ (in Dutch)

alexgarel (approved):
Thanks

PR #12021, opened 2025-06-11
fix: corrections for mismatched-kj-nutrient-calculation data quality description in most languages

### LINKED ISSUE #12018: Data quality error description doesn't match title
On this [product](https://world.pro.openfoodfacts.org/cgi/product.pl?type=edit&code=4751037280731) the title of the error is : 
"Energy value in kJ does not match value computed from other nutrients" 

BUT the description of the issue is not matching the title (see screenshot below): 
_"The energy value in kJ must be about 4.2 times greater than the value in kcal. The values seems to be reversed."_ 

Image

It should be : 
_"The energy value in kJ does not match value computed from other nutrients. It is often due to an error in the nutrition facts table entered by the users. Sometimes, it's an error coming from the producer."_ 
See: https://wiki.openfoodfacts.org/List_of_data_quality_errors_(generated)

jayaddison:
### What

As described in #12018, the long-text description for the data quality error with title "Energy value in kJ does not match value computed from other nutrients" seems to be incorrect.

This pull request backfills those descriptions with search-and-replaced copies of the equivalent data quality message for nutrient energy calculation mismatch in kcal.

Three languages are commented-out where I couldn't find the text string `kcal` to search-and-replace with `kJ`.

- Fixes #12018

### DIFF
--- a/taxonomies/data_quality.txt
+++ b/taxonomies/data_quality.txt
@@ -182,26 +182,26 @@ es: El valor energético en kJ no coincide con el valor calculado a partir de ot
 fr: Valeur énergétique en kJ ne correspondant pas à la valeur calculée à partir des autres nutriments
 it: Il valore energetico in kJ non corrisponde al valore calcolato dagli altri nutrienti
 pt: O valor energético em kJ não corresponde ao valor calculado a partir de outros nutrientes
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):
Thank you!

jayaddison:
You're welcome; it's not ideal that a few languages are commented-out, but I think it's worth updating/correcting most of them.

PR #12031, opened 2025-06-15
taxonomy: Add “concentrate” as distinct ingredient

Freso:
There are currently [288 products with `en:concentrate` in their ingredients](https://world.openfoodfacts.org/facets/ingredients/concentrate), compared with [45 with `en:concentrated extract`](https://world.openfoodfacts.org/facets/ingredients/concentrated-extract). That’s “concentrate” being used more than 5 times more than the “extract” term, but only the “extract” variant is in `food/ingredients.txt`, leading to all the others to go unrecognised.

This adds “concentrate” as its own, distinct ingredient. Hopefully we’ll be able to parse it better in the future.

See also https://openfoodfacts.slack.com/archives/C06A7LENM/p1749983011998189


stephanegigandet:
I think that in most cases, those are parsing issues, or error in the ingredients list.  "concentrate" or "concentrated extract" is not an ingredient, it's a processing, so it's missing something.

e.g. Lemon Juice*, Strawberry Juice*, Organic Agave Inulin, Stevia Leaf Extract., *Concentrate
https://world.openfoodfacts.org/product/09663336/strawberry-lemon-poppi
-> we should combine it in "lemon juice concentrate" (or more likely it is "lemon juice from concentrate")

stephanegigandet:
https://world.openfoodfacts.org/product/0613008771102/iced-tea-tropical-arizona
concentrate (black carrot)
->we could try to convert it to "black carrot concentrate"

stephanegigandet:
We are missing a lot of ingredient photos to verify, but I think that in many instances we got an extra comma. https://world.openfoodfacts.org/product/0052200076248/organics-mango-carrot-beech-nut
e.g. ORGANIC MANGO, ORGANIC CARROT, WATER, ORGANIC, LEMON JUICE, CONCENTRATE. --> the last one os organic lemon juice concentrate.

stephanegigandet:
Regarding the original question: we should not make "concentrate" a synonym of "concentrated extract". For "concentrate", the only thing that has been removed should be water. But an extract is different, they can use solvent etc. 

So in the short term, we can just add a new ingredient "concentrate". In longer term, we could try to parse a bit better some of those things.

PR #12032, opened 2025-06-15
taxonomy: Fix and amend some `expected_ingredients`

Freso:
I misunderstood how `expected_ingredients` worked when I added these[1], so removing again to stop them getting flagged for quality issues!

And while I was at it, I also added one that _could_ use it. :)

[1] I thought it would just want that ingredient to be in 
    the ingredient list, not that it would want it to be
    the sole ingredient.


alexgarel (approved):
Thanks !

PR #12037, opened 2025-06-16
docs: Search-as-you-type warning

CharlesNepote:
Many people use search API to build a search-as-you-type feature, leading to be banned shortly.


teolemon (dismissed):
Can you explicitly mention searchalicious as an alternative? It is all ready mentionned in the tutorial

CharlesNepote:
> Can you explicitly mention searchalicious as an alternative? It is all ready mentionned in the tutorial

Search-a-licious is still mentioned as beta, and I don't know if search-a-licious can support as many requests as needed for this kind of feature (search-as-you-type).

Maybe @alexgarel wants to do it, otherwise I prefer to left this PR as is, to avoid reusers asking to unban them due to this.

alexgarel (approved):
Thanks !

PR #12038, opened 2025-06-16
chore: editorconfig: Don’t affect expected_test_results

Freso:
Anything in `tests/*/expected_test_results/*` should mostly be automatically generated, but sometimes it may be prudent to poke around in them or make minor changes by hand.

Right now, `.editorconfig` tells text editors to strip all trailing whitespace as well as enforce some other style settings, for all files. The HTML files that tests are compared against contain a lot of trailing whitespace which easily gets stripped by this configuration.

This adds another section to `.editorconfig` that essentially `unset`s most settings specifically for .html files under `expected_test_results`.


hangy:
Note: Build seems to be failing because of an unrelated change to a CPAN module: #12042

hangy (approved):

PR #12049, opened 2025-06-17
refactor: Use shared method to create HTTP client

hangy:
### What

Introduce a new method `create_user_agent` to instantiate `LWP::UserAgent` the same way across modules. Also replace `LWP::Simple` with `LWP::UserAgent` from that method to enforce similar behaviour.

### Future

In the long run, I also want to use this for consistent integration in the [`otel` branch](https://github.com/openfoodfacts/openfoodfacts-server/tree/otel).


alexgarel (approved):
Cool

PR #12051, opened 2025-06-17
fix: replace NutriScore image with higher resolution version

garvit-bhattt:
Replaced the low-resolution image with a high-resolution version (nutriscore) in the producers folder for better clarity and sharper rendering across devices.
Screenshot 2025-06-18 010044
nutriscore


stephanegigandet:
Hi @garvit-bhattt 
Thanks for the PR. Could you use one of the official monochrome logos from https://www.santepubliquefrance.fr/media/files/02-determinants-de-sante/nutrition-et-activite-physique/nutri-score/annexe2-charte-graphique-en ?
(e.g. this one: 
image -- the graphic charter has a vector version)
Also for this specific file, it's use in the https://world.pro.openfoodfacts.org , it needs to be in a square as you did, but can you make sure the logo is centered inside the square?
Thank you!

garvit-bhattt (commented):
I've made the suggested change to center the NutriScore logo in the updated image section. This should now align better visually across different screen sizes.

PR #11986, opened 2025-06-01
ci: Keep test coverage up-to-date

hangy:
### What

Build backend and run tests for pushes to the main branch

### Related issue(s) and discussion

- Fixes #11985

PR #12064, opened 2025-06-21
ci: Split tests run in CI

JagjeevanAK:
### What

This PR restructures the CI pipeline to split test execution across multiple jobs for improved parallelization and faster feedback. The changes reorganize how tests are run in GitHub Actions workflows to optimize build times and resource utilization.

Key changes include:
- Splitting test suites into separate CI jobs to run in parallel
- Reorganizing workflow configuration to better distribute test workloads
- Updating Docker and dependency configurations to support the new test structure
- Improving test organization for better maintainability and debugging

The split test approach will reduce overall CI runtime by allowing different test categories to execute simultaneously, providing faster feedback to developers and improving the development workflow.


.github/workflows/pull_request.yml
```diff
@@ -47,14 +46,9 @@ jobs:
...
       with:
-        node-version: '22.x'
-    - name: Cache node modules
-      uses: actions/cache@v4
-      with:
-        path: node_modules
-        key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
-        restore-keys: |
-          ${{ runner.os }}-node-
+        node-version: '24.x'
```
hangy: The version should be kept in sync with https://github.com/openfoodfacts/openfoodfacts-server/blob/d766aa247c037b5acd0d1e9a3bc2d51918061269/Dockerfile.frontend#L6, and ideally with what Stephane uses for manual deployment to production. Node 22.x is still LTS, so should be usable.

JagjeevanAK: okay got it. I will change

.github/workflows/pull_request.yml
```diff
@@ -134,11 +180,15 @@ jobs:
...
       run: make check_perl
 
-  tests:
-    name: 🐪 Perl unit tests
+  unit_tests:
+    name: 🐪 Perl Unit Tests
     needs: [filter, build_backend]
     runs-on: ubuntu-latest
-    if: ((github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'pull_request') && needs.filter.outputs.code_modified == 'true'
+    if: github.event_name == 'pull_request' && needs.filter.outputs.code_modified == 'true'
```
hangy: We need unit tests and integration tests to run on the `main` branch to keep code coverage in Codecov up-to-date, see #11986

.github/workflows/pull_request.yml
```diff
@@ -153,14 +203,24 @@ jobs:
...
     - name: Download backend image from artifacts
       id: downloadbackendimage
-      uses: ishworkh/container-image-artifact-download@v2.1.0
+      uses: ishworkh/container-image-artifact-download@v2.0.0
       with:
-        image: "openfoodfacts-server/backend:dev"
-        download_tmp_dir: ${{ runner.temp }}
-    - name: tests
+        image: "openfoodfacts-server/backend:dev"       
+    # downloadbackendimage task loads the image into docker and keeps the original file.
```
hangy: This shouldn't be necessary for `ishworkh/container-image-artifact-download@v2.1.0`

.github/workflows/pull_request.yml
```diff
@@ -153,14 +203,24 @@ jobs:
       run: ./.github/scripts/setup_git.sh
     - name: Download backend image from artifacts
       id: downloadbackendimage
-      uses: ishworkh/container-image-artifact-download@v2.1.0
+      uses: ishworkh/container-image-artifact-download@v2.0.0
```
hangy, line 407: Why the downgrade?

JagjeevanAK: It was giving error when it was v2.1.0.

Makefile
```diff
@@ -605,3 +607,67 @@ guard-%: # guard clause for targets that require an environment variable (usuall
    		echo "Environment variable '$*' is not set"; \
    		exit 1; \
 	fi;
+
+# Unit test groups for parallel execution in CI
+# Usage: make unit_test_group TEST_GROUP=1
+# Groups are balanced by number of test files
+define UNIT_TEST_GROUPS
```
hangy: Could the groups be determined algorithmically? All jobs of the actions matrix use the same source, so that the groups could be built automatically to avoid having to add each test file here in the `Makefile` individually

.github/workflows/pull_request.yml
JagjeevanAK: Sorry I was not there so I didnt knew main reason to this was I have copied some of the code from my previous PR which I will close soon

Makefile
JagjeevanAK: I have made group on there size as small are combined with large one so that run it in equal time

.github/workflows/pull_request.yml
JagjeevanAK: really ? I will see docs and will update it thanks for pointing it out.

JagjeevanAK:
hey @alexgarel,

This PR is ready to merge.

Makefile
john-gom: I don't like the idea of having hard-coded groups. It will be too easy to add a test and forget to include it which means the test will run locally but not in CI.

john-gom:
It would be good to see some kind of comparison with not using groups but just running the unit and integration tests as separate parallel tasks. That way we can see how much time the group approach is actually saving

Makefile
JagjeevanAK: good point never though of this will look into it.

JagjeevanAK:
group approach is good according to me because every time we need to spin the resource for each test and that might take time.

Makefile
JagjeevanAK: Hey @john-gom, @hangy what do you think instead of hard coding making directory inside of existing directory of tests.

alexgarel: @JagjeevanAK I think it's better to create a script that you call and gives you back the test group, it should also take the number of group as parameter.

This script would list all tests, using file system) and separate into groups in a predictable manner (eg. after sorting the list), and print the list on stdout.

Then in makefile you can just call the script to get the test group.

The script might be written in python, as it's found on most machines today.

Please put the script in a sub folder, eg. scripts/tests/

Makefile
```diff
@@ -605,3 +607,67 @@ guard-%: # guard clause for targets that require an environment variable (usuall
...
+unit_test_group: create_folders
+ifeq ($(TEST_GROUP),)
+	$(error TEST_GROUP is required. Usage: make unit_test_group TEST_GROUP=1)
+endif
+	@echo "🥫 Running unit test group $(TEST_GROUP) …"
+	@echo "🥫 Tests in group $(TEST_GROUP): $(call get_unit_group_tests,$(TEST_GROUP))"
+	mkdir -p tests/unit/outputs/
+	${DOCKER_COMPOSE_TEST} up -d memcached postgres mongodb
+	@echo "🥫 Running all tests in group $(TEST_GROUP) with parallel execution and JUnit XML generation..."
+	${DOCKER_COMPOSE_TEST} run ${COVER_OPTS} -e JUNIT_TEST_FILE="tests/unit/outputs/junit_group_$(TEST_GROUP).xml" -T --rm backend yath test --renderer=Formatter --renderer=JUnit --job-count=${CPU_COUNT} $(addprefix tests/unit/,$(call get_unit_group_tests,$(TEST_GROUP)))
```
alexgarel, line 672: Don't we loose coverage with this new run type ?

Makefile
alexgarel: Oh, in bash this is as simple as eg.:
```bash
find tests/unit -name "*.t"|sort|split -n 1/10
find tests/unit -name "*.t"|sort|split -n 2/10

```

Makefile
```diff
@@ -289,10 +289,12 @@ unit_test: create_folders
 	@echo "🥫 Running unit tests …"
 	mkdir -p tests/unit/outputs/
 	${DOCKER_COMPOSE_TEST} up -d memcached postgres mongodb
+	@echo "🥫 Running tests with yath (parallel jobs: ${CPU_COUNT})"
 	${DOCKER_COMPOSE_TEST} run ${COVER_OPTS} -e JUNIT_TEST_FILE="tests/unit/outputs/junit.xml" -T --rm backend yath test --renderer=Formatter --renderer=JUnit --job-count=${CPU_COUNT} tests/unit
```
alexgarel, line 304: Here unit tests might run in parallel directly thanks to yath, without having to run them in different github runner. Is it that CPU_COUNT is too low in each runner ?

JagjeevanAK: need to check

Makefile
JagjeevanAK: alex problem is that some test are taking time so I have mixed it with small and big.

JagjeevanAK: there are some e2e test like some what 2-3 if I am not wrong.

Makefile
JagjeevanAK: according to me it should not and this is ordered this way because each test was running double previously.

docs/dev/dynamic-test-grouping.md
```diff
@@ -0,0 +1,269 @@
+# Dynamic Test Grouping System
```
alexgarel: Please use diataxis naming
```suggestion
# Explain Dynamic Test Grouping System
```

also rename the file to explain-dynamic-test-grouping.md

.github/workflows/test-dynamic-grouping.yml
alexgarel, line 1: Why is this a workflow ?

I mean it would be better to have it in a bash script that can also be run locally (eventually make test could also call it).

.github/scripts/update_test_groups.sh
```diff
@@ -0,0 +1,152 @@
...
+    if [ ! -f ".test_groups_cache/${test_type}_groups.json" ]; then
+        echo "Cache not found for $test_type tests, generating..."
+        return 1
+    fi
+    
+    # Check if any new test files exist since last cache
+    local current_test_files
+    current_test_files=$(find "tests/$test_type" -name "*.t" -type f 2>/dev/null | sort || echo "")
+    
+    # Get cached test files
```
alexgarel: Reformulating to (maybe) make it a bit more clear
```suggestion
    # Get the list of tests contained in cached files
```

.github/scripts/update_test_groups.sh
```diff
@@ -0,0 +1,152 @@
...
+    exit 1
+fi
+
+echo "🐍 Python version: $(python3 --version)"
+
+# Create cache directory if it doesn't exist
+mkdir -p .test_groups_cache
+
+# Function to check if test groups need regeneration
+check_test_groups_validity() {
```
alexgarel, line 31:
```suggestion
# Function to check if test groups need regeneration
# Param $1: the type of test: unit or integration
# Exit with code 1 if tests needs to be regenerated
check_test_groups_validity() {
```

.github/scripts/update_test_groups.sh
```diff
@@ -0,0 +1,152 @@
+#!/bin/bash
+
+# Script to update test groups in CI based on new tests or timing changes
+# This script should be run before test execution to ensure groups are up-to-date
```
alexgarel, line 4:
```suggestion
# Script to update test groups in CI based on new tests or timing changes
# This script should be run before test execution to ensure groups are up-to-date
# It uses dynamic_test_grouper.py to generate test groups, generating
# .test_groups_cache/${test_type}_groups.json and ".test_groups_cache/${test_type}_groups.mk
```

.github/workflows/pull_request.yml
```diff
@@ -63,42 +61,206 @@ jobs:
...
+      - name: Restore BuildKit cache
+        uses: actions/cache@v4
+        with:
+          # Cache key based on files that trigger Docker rebuild (from path-filter logic)
+          key: ${{ runner.os }}-docker-${{ hashFiles('Dockerfile*', 'docker-compose.yml', 'docker/**', 'cpanfile*', 'conf/apache*') }}
+          restore-keys: |
+            ${{ runner.os }}-docker-
+          path: /tmp/.buildx-cache
+          # Enable compression for faster cache restoration
+          enableCrossOsArchive: false
```
alexgarel, line 97: Can we measure how much this cache is taking ?

Because Github only ensure [10G of cache per repo](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#usage-limits-and-eviction-policy) so if the cache size is too big, it won't work.

Also as we are using the "dev" mode to run tests (which bind mounts directory), I think we could altogether avoid building containers when we know modified files don't have a real impact (to be sure, I would do this by telling which files are safe to have changed, not the contrary). It would just be about getting the latest "main" container.

In this case we could even drop altogether this buildx optimization (which has this space problem), only PRs affecting Docker build would take longer to build (and there are only a few of them).

.github/workflows/pull_request.yml
```diff
@@ -63,42 +61,206 @@ jobs:
...
+          else
+            echo "✅ Sufficient space available for Docker save"
+          fi
+
+      - name: push backend image as artifact
+        uses: ishworkh/container-image-artifact-upload@v2.0.0
+        with:
+          image: "openfoodfacts-server/backend:dev"
+          # Add retention to save space in Actions storage
+          retention_days: 1
```
alexgarel, line 230: Is 1 day this enough ?

scripts/__pycache__/dynamic_test_grouper.cpython-39.pyc
alexgarel, line 1: __pycache__ folder should have been ignored… please, remove it.

scripts/dynamic_test_grouper.py
alexgarel, line 1: Great ! It's clear and easy to follow

scripts/generate_matrix_config.py
```diff
@@ -0,0 +1,55 @@
...
+    total_unit_time = sum(unit_grouper._get_test_time(test, unit_timings) for test in unit_test_files)
+    unit_groups = max(1, int(total_unit_time / 780) + 1)  # 780 seconds = 13 minutes
+    unit_groups = max(6, min(unit_groups, min(len(unit_test_files), 12)))  # Minimum 6 groups for unit tests
+    
+    integration_test_files = integration_grouper._discover_test_files()
+    integration_timings = integration_grouper._load_cache(integration_grouper.cache_dir / "integration_timings.json") or {}
+    
+    total_integration_time = sum(integration_grouper._get_test_time(test, integration_timings) for test in integration_test_files)
+    integration_groups = max(1, int(total_integration_time / 780) + 1)
+    integration_groups = max(9, min(integration_groups, min(len(integration_test_files), 12)))  # Minimum 9 groups for integration tests
```
alexgarel: Why do we recompute all this ?isn't it already computed by DynamicTestGrouper ?

tests/unit/expected_test_results/knowledge_panels/en-nutriscore-serving-size-error.json
alexgarel, line 1: This should not be here ! But it's not because of you :-) I think just merging main will remove this.

.github/workflows/pull_request.yml
```diff
@@ -177,8 +450,111 @@ jobs:
...
+      run: |
+        echo "🥫 Updating integration test timing data from test results..."
+        if [ -d "tests/integration/outputs" ] && [ "$(ls -A tests/integration/outputs/*.xml 2>/dev/null)" ]; then
+          python3 scripts/dynamic_test_grouper.py --type=integration --update-timings --junit-dir=tests/integration/outputs
+          echo "🥫 Integration test timing data updated successfully"
+        else
+          echo "🥫 No JUnit XML files found, skipping timing update"
+        fi
+    
+    - name: generate coverage results
```
alexgarel: I don't think code coverage will be effective after splitting tests.

We should normally upload the test coverage as an artifact, and have a separate action that get all those files and send the coverage report.

We can live without it for some time.

Can you add a comment and comment out the whole coverage thing ?

scripts/__pycache__/dynamic_test_grouper.cpython-39.pyc
JagjeevanAK: done.

.github/workflows/pull_request.yml
JagjeevanAK: yes

.github/workflows/pull_request.yml
JagjeevanAK: it's just around 1.6G some what

scripts/dynamic_test_grouper.py
JagjeevanAK: thanks for your comment

tests/unit/expected_test_results/knowledge_panels/en-nutriscore-serving-size-error.json
JagjeevanAK: Yes this was due to @RayanGoubet not correctly made changes into his PR discussion with him about this

.github/workflows/pull_request.yml
JagjeevanAK: okay I will comment out this things as code coverage will not be effective in general.

scripts/__pycache__/dynamic_test_grouper.cpython-39.pyc
JagjeevanAK: done

scripts/generate_matrix_config.py
JagjeevanAK: YUP You are absolutely correct I have update the file check it now.

JagjeevanAK:
Hey @alexgarel check this PR for final review.

docs/dev/dynamic-test-grouping.md
alexgarel: I'm renaming it, it's important because I have rules in docs/dev/.pages to set page order base on filenames.

alexgarel (approved):
Thanks @JagjeevanAK for this hard work, which will really benefits the devs.

Can I let you fix tests ?

JagjeevanAK:
> Thanks @JagjeevanAK for this hard work, which will really benefits the devs.
> 
> Can I let you fix tests ?

100% I will check the issue.

PR #12073, opened 2025-06-24
fix: Include lc in the Test Ingredients Analysis form

tuukka:
### What

The form Test Ingredients Analysis did not include the `lc` language parameter as a form field, so it was not included in the request when the form was submitted. As a result, the parameter was lost and the ingredients were analysed in English.

### DIFF
--- a/templates/web/pages/test_ingredients/test_ingredients_analysis.tt.html
+++ b/templates/web/pages/test_ingredients/test_ingredients_analysis.tt.html
@@ -7,6 +7,7 @@
     Ingredients text (language code: [% lc %]): <br/>
     <textarea id="ingredients_text" name="ingredients_text" style="height:8rem;">[% ingredients_text %]</textarea>
     <input type="hidden" name="type" value="[% type %]"/>
+    <input type="hidden" name="lc" value="[% lc %]"/>
     <input type="hidden" name="action" value='process'/>
     <input type="submit" name="submit">
 


Freso (approved):
LGTM.

tuukka:
> Actually, I wonder if this will cause `lc` to always be added to query parameters, even when `lc` matches the domain language

Good point. I tested this and it does always add the parameter, which is not ideal: if I change the domain I expect the language to change too, but with `lc` involved, it doesn't. In the same sense, `lc` is not the best way to switch the language, as there can always be a mismatch between it and the domain.

I'm not deeply familiar with either the language system of OFF or the Template Toolkit, so I'm probably not the best person to improve this from here, but I probably can, if there's no-one around who can fix this directly.

>  I think this is probably better than the current state of affairs either way. If it’s deemed too much of a nuisance to include this query parameter when it’s superfluous, then that can be changed in a future PR.

I agree - the current state is more confusing to users (the form displays the `lc` language but when submitted, the ingredients are actually parsed based on the site language).

stephanegigandet:
The best thing would be to actually add a language drop down (defaulting to the language of the site), and set it to ingredients_lc to do the analysis..

Freso:
This PR is still an improvement on the current situation, I’d argue, and better/more advanced solutions can be made later. :)

stephanegigandet (approved):
Thanks!

PR #12078, opened 2025-06-27
fix: Use testcontainer directly with simplified configuration

john-gom:
### What

Use new Keycloak testcontainer image with config built in which improves startup time

- Fixes https://github.com/openfoodfacts/openfoodfacts-auth/issues/207

Note this changes the way clients are created in Keycloak. Instead of creating them on-the-fly on startup they need to be created manually via the admin UI or API. The full process for creating a client would be as follows:

1. Create the client in Keycloak (copy from the OFF client if it needs full permission). The client id should match the Product Opener flavour in upper case
2. Get the secret and add this as a GitHub secret in openfoodfacts-server
3. Update the openfoodfacts-server image deployment script to add this secret to the environment.

For production the edits would be more manual


JagjeevanAK (approved):
Looks good to me.

Issue #12006, opened 2025-06-09
Extract "claims" data from GS1 imports and assign them to labels


stephanegigandet:
There are a number of fields we get in GS1 product data sheets that we currently do not import.

In particular there are some "claims" that producers can set, with an indication of whether the claim is also claimed on the package).

e.g.

```
<productInformationDetail>
<claimDetail>
<claimMarkedOnPackage>FALSE</claimMarkedOnPackage>
<claimTypeCode>REGULAR_LEVEL_OF</claimTypeCode>
<claimElementCode>SUGARS</claimElementCode>
</claimDetail>
<claimDetail>
<claimMarkedOnPackage>FALSE</claimMarkedOnPackage>
<claimTypeCode>FREE_FROM</claimTypeCode>
<claimElementCode>SWEETENERS</claimElementCode>
</claimDetail>
<claimDetail>
<claimMarkedOnPackage>FALSE</claimMarkedOnPackage>
<claimTypeCode>FREE_FROM</claimTypeCode>
<claimElementCode>ARTIFICIAL_SWEETENERS</claimElementCode>
</claimDetail>
<claimDetail>
<claimMarkedOnPackage>TRUE</claimMarkedOnPackage>
<claimTypeCode>NO_ADDED</claimTypeCode>
<claimElementCode>SUGARS</claimElementCode>
</claimDetail>
<claimDetail/>
<claimDetail>
<claimMarkedOnPackage>TRUE</claimMarkedOnPackage>
<claimTypeCode>CONTAINS</claimTypeCode>
<claimElementCode>NATURALLY_OCCURING_SUGARS</claimElementCode>
</claimDetail>
</productInformationDetail>
</product_information:productInformationModule>
```

We could definitely add some of them, especially the ones that are marked TRUE for claimMarkedOnPackage. We possibly could add the ones that are not marked on the package too.

PR #12084, opened 2025-06-28
feat: add no added sugars label from gs1

Vic142:
GS1 provides several claims for its products that are not yet imported. In this PR the claim "no added sugars" is now imported from the GS1 provided files into the products _labels_ field. New GS1 files were also added to test that the new claim is correctly imported.

This PR is related to the issue #12006.


lib/ProductOpener/GS1.pm
```diff
@@ -367,6 +367,10 @@ my %gs1_maps = (
 		"MON" => "month",
 		"DAY" => "day",
 	},
+
+	claimElementCode => {
```
stephanegigandet, line 371: Maybe rename claimElementCode to claimElementCodeNoAdded so that we can have labels like "contains added sugars"

PR #12087, opened 2025-06-28
chore:Extended the existing linting rules to lint the spec file

JagjeevanAK:
This PR is to extend the current existing rules of our api linting tool to follow the good practices while maintaining the our hand written Openapi spec file.

### DIFF
--- a/.spectral.yaml
+++ b/.spectral.yaml
@@ -1,1 +1,125 @@
-extends: "spectral:oas"
+# Open Food Facts API - Spectral Configuration
+# This file defines Spectral rules for linting the Open Food Facts OpenAPI specification
+
+extends: ["spectral:oas"]
+
+rules:
+  # Documentation Quality Rules
+  require-operation-description:
+    description: "All operations must have descriptions"
+    message: "Operation {{property}} is missing a description"
+    severity: error
+    given: "$.paths.*[get,post,put,patch,delete,head,options,trace]"
+    then:
+      field: description
+      function: truthy
+
+  require-parameter-description:
+    description: "All parameters must have descriptions"
+    message: "Parameter {{value.name}} is missing a description"
+    severity: error
+    given: "$.paths..parameters[*]"
+    then:
+      field: description
+      function: truthy
+
+  require-schema-description:
+    description: "All schemas should have descriptions"
+    message: "Schema {{property}} is missing a description"
+    severity: warn
+    given: "$.components.schemas.*"
+    then:
+      field: description
+      function: truthy
+
+  # Response Structure Rules
+  consistent-error-responses:
+    description: "Error responses should follow consistent structure"
+    message: "Error response should include schema definition"
+    severity: warn
+    given: "$.paths..responses['400','401','403','404','500']"
+    then:
+      field: "content.application/json.schema"
+      function: truthy
+
+  require-response-examples:
+    description: "Success responses should include examples"
+    message: "Response {{property}} should include examples for better documentation"
+    severity: info
+    given: "$.paths..responses['200','201','202']"
+    then:
+      field: "content.application/json.examples"
+      function: truthy
+
+  # Component Organization Rules
+  consistent-component-naming:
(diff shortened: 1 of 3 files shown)


.github/workflows/api-linting.yml
```diff
@@ -25,7 +27,9 @@ jobs:
           node-version: '22'
 
       - name: Run Spectral lint
+        id: spectral
+        continue-on-error: true
         uses: stoplightio/spectral-action@latest
         with:
           file_glob: 'docs/api/ref/*.yaml'
-          spectral_ruleset: .spectral.yaml
+          spectral_ruleset: .spectral.yaml
```
Freso: You might want to switch to an editor that adheres to [`.editorconfig`](https://editorconfig.org/) to not introduce random changes like this. :) (Or get a plugin for the editor you are using.)

JagjeevanAK: Freso will look into your suggestions but on this tool we have already discussed and we have seen it as effective and not random introduction I am just trying to make it more powerful.

.spectral.yaml
```diff
@@ -1 +1,125 @@
...
+      function: truthy
+
+  require-schema-description:
+    description: "All schemas should have descriptions"
+    message: "Schema {{property}} is missing a description"
+    severity: warn
+    given: "$.components.schemas.*"
+    then:
+      field: description
+      function: truthy
```
Freso, line 33: E.g., these seem like they could be replaced by `docs-descriptions` from <https://apistylebook.stoplight.io/docs/documentation#docs-descriptions>.

.github/workflows/api-linting.yml
Freso: The random change I’m referring to here is the removal of the newline at the end of the file: removal of newline at end of file

Pretty sure that is not something that you have discussed and “seen it as effective”. :)

Freso: Having a newline at end-of-files is codified in [`.editorconfig`](https://github.com/openfoodfacts/openfoodfacts-server/blob/cd881277fb47a0b11d2308c463fa8f5c3f34fb05/.editorconfig), so any editors that understand that should not randomly remove that final newline:
https://github.com/openfoodfacts/openfoodfacts-server/blob/cd881277fb47a0b11d2308c463fa8f5c3f34fb05/.editorconfig#L14

JagjeevanAK: Oh got your point sorry for that misunderstanding. I didnt kenw about that and it was removed my me personally also there I was testing some things so it was removed by mistake and not by any intensions.

.github/workflows/api-linting.yml
```diff
@@ -4,11 +4,13 @@ on:
...
+      - '.spectral.yaml'
   push:
     branches:
       - main
     paths:
       - 'docs/api/ref/**/*.yaml'
+      - '.spectral.yaml'
 
 jobs:
   lint-api:
```
stephanegigandet, line 16: How is lint-api used exactly? Is it a test that passes or fails?

If we have tests in GitHub actions, it would be great to be able to run them locally first, with a make command, similar to "make lint".

JagjeevanAK: hey @stephanegigandet there is already an make cmd in Makefile called `check_openapi_spectral` you can check it.

.spectral.yaml
alexgarel: @Freso we are already deriving from spectral:oas 

And docs-description is more strict (no short description, and so on). Let's keep it simple for now :-)

.spectral.yaml
```diff
@@ -1 +1,125 @@
...
+
+  # Ensure operation IDs are present
+  require-operation-id:
+    description: "All operations should have operation IDs"
+    message: "Operation should have an operationId"
+    severity: warn
+    given: "$.paths.*[get,post,put,patch,delete,head,options,trace]"
+    then:
+      field: operationId
+      function: truthy
```
alexgarel, line 85: I'm not sure what's this one.

PR #12095, opened 2025-06-30
fix: cache OCR, make cloud_vision default

stephanegigandet:
This PR is to implement improvements triggered by a production issue we had with the DART plugin integration tests overloading the OFF server with tesseract requests.

It:
- makes cloud vision the default if ocr_engine parameter is not passed
- enables disabling tesseract and/or cloud vision in Config2.pm
- for Cloud Vision, if we have a JSON OCR file that is less than 30 days, we serve it back immediately (will help with the requests made by the DART plugin integration tests)

Related: https://github.com/openfoodfacts/openfoodfacts-dart/issues/1109

### DIFF
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -118,6 +118,7 @@ BEGIN {
 
 		&display_image
 
+		&select_ocr_engine
 		&extract_text_from_image
 		&send_image_to_cloud_vision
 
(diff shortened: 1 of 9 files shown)


lib/ProductOpener/Images.pm
```diff
@@ -2527,6 +2584,34 @@ sub send_image_to_cloud_vision ($image_path, $json_file, $features_ref, $gv_logs
 	my $url
 		= $ProductOpener::Config::google_cloud_vision_api_url . "?key="
 		. $ProductOpener::Config::google_cloud_vision_api_key;
+
+	# If we already have a CloudVision result JSON file that is less than 1 month old, we return it
+	# instead of sending the image to the API. This is in particular useful for integration tests
+	# that make the same request multiple times.
```
alexgarel, line 2605: Note: we could do the same for tesseract, because it's not a cheap processing.

alexgarel (approved):
Thanks

Freso (commented):
I don’t see any glaring issues, but this is not an area I’m familiar enough with at this point to feel confident in “approving”. :sweat_smile:

stephanegigandet:
@alexgarel I had to change a bit the unit test as we are now caching some requests

PR #12112, opened 2025-07-02
fix: moving images to another product

stephanegigandet:
fix issue from refactor of product image upload


lib/ProductOpener/Images.pm
```diff
@@ -1399,6 +1399,21 @@ sub process_image_move ($user_id, $code, $imgids, $move_to, $ownerid) {
 	my $product_ref = retrieve_product($product_id);
 	defined $product_ref->{images} or $product_ref->{images} = {};
 
+	# New product to which the images are moved
+	my $move_to_product_ref;
+
+	if ($move_to ne "trash") {
+		# Retrieve the product to which the images are moved
+		$move_to_product_ref = retrieve_product($move_to_id);
```
alexgarel, line 1407: was this $move_to_id validated as a good id (we have a function for that) ?

PR #12114, opened 2025-07-02
fix: change product barcode

stephanegigandet:
There was a typo that prevented the box to changed the barcode of a product to be displayed.


alexgarel (approved):
Good !

PR #12115, opened 2025-07-02
fix: is_valid_code() with leading zeroes

stephanegigandet:
Short codes with 3 digits or less and leading 0s should be rejected.


alexgarel (approved):
Cool,
It's a good example why to factor code in a method, even if it's as simple as a regexp !

hangy (approved):

PR #12116, opened 2025-07-03
feat: M3 raw data knowledge panel

RayanGoubet:
### What

Milestone 3  - Introduce a raw data knowledge panel for core contributors

Capture d’écran du 2025-07-03 13-44-39

Capture d’écran du 2025-07-03 13-44-59

Capture d’écran du 2025-07-03 13-45-22


teolemon:
@RayanGoubet can you add a little preview screenshot ?

lib/ProductOpener/KnowledgePanelsProduct.pm
```diff
@@ -0,0 +1,124 @@
...
+
+=head4 options reference $options_ref
+
+
+=cut
+
+sub create_product_card_panel ($product_ref, $target_lc, $target_cc, $options_ref, $request_ref) {
+	$log->debug("create product card panel", {code => $product_ref->{code}}) if $log->is_debug();
+
+	my $base_url = "/world.openfoodfacts.org/facets/brands/";
```
stephanegigandet: To get the link and the display name, just use this function:

```
# get_taxonomy_tag_and_link_for_lang computes the display text and link
# in a target language for a canonical tagid
# It returns a hash ref with:
# - display : text of the link in the target language, or English
# - display_lc : language code of the language returned in display
# - known : 0 or 1, indicates if the input tagid exists in the taxonomy
# - tagurl : escaped link to the tag, without the tag type path component

sub get_taxonomy_tag_and_link_for_lang ($target_lc, $tagtype, $tagid) {
```

po/common/common.pot
```diff
@@ -7547,3 +7547,17 @@ msgstr "Report this image to our moderators"
 msgctxt "robots_not_served_here"
 msgid "Due to heavy abuse from robots, we are unable to serve this page to unidentified users. Please create a free Open Food Facts account to access all of Open Food Facts."
 msgstr ""
+
+msgctxt "origin_of_ingredients"
```
stephanegigandet, line 7551: We already have translated strings for those, as they are displayed in the product page already.

Check the code for the display_field() function in Web.pm

```
			my $name = lang($field);
			if ($name eq '') {
				$name = ucfirst(lang($field . "_p"));
			}
```

templates/api/knowledge-panels/product/product_card.tt.json
```diff
@@ -0,0 +1,37 @@
...
+  "topics": ["product"],
+  "title_element": {
+    "title": "[% edq(lang('product_information')) %]"
+  },
+  "elements": [
+    [% IF panel.brand_panels %]
+      [% INCLUDE "api/knowledge-panels/product/product_brands.tt.json" %]
+    [% END %]
+
+    [% IF product.origins_tags %]
```
stephanegigandet: We should have a generic template that works for all field types, instead of having one template per type.

Or maybe simpler, just have one template that loops and displays all the fields.

templates/api/knowledge-panels/product/product_countries.tt.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "panel",
+  "title": "[% edq(lang('countries')) %]",
+  "elements": [
+    [% FOREACH country IN product.countries_tags %]
+    {
+      "type": "panel",
+      "title": "[% display_taxonomy_tag('countries_tags', country) %]"
+    }[% IF !loop.last %],[% END %]
```
stephanegigandet: you can put the comma at the end of all entries, including the last one, it should be removed automatically after

templates/api/knowledge-panels/product/product_tags_panels.tt.json
```diff
@@ -0,0 +1,15 @@
+[% FOREACH panel IN panel.taxonomy_panels %]
+    {
```
stephanegigandet: Each panel contains elements, and those elements can be themselves panels, but when an element is a panel, it just lists the panel id.

You can look at the generated panels for a product to see how the panel elements are defined:
https://fr.openfoodfacts.org/api/v3/produit/7622210449283/prince-gout-chocolat-lu?fields=knowledge_panels

e.g.

```
{
elements: [
{
element_type: "panel",
panel_element: {
panel_id: "environmental_score"
}
},
```

stephanegigandet (commented):
To verify that the knowledge panels are correct, you should definitely try to display them.

If you look in Display.pm, you can see how to render the knowledge panels in HTML, then you can pass the generated HTML to the product page template and display it:

```
	$template_data_ref->{environment_card_panel}
		= display_knowledge_panel($product_ref, $product_ref->{"knowledge_panels_" . $lc}, "environment_card");
```

RayanGoubet:
Add option &raw_panel=1 to show the raw data in the product page

lib/ProductOpener/KnowledgePanelsProduct.pm
```diff
@@ -0,0 +1,158 @@
...
+		my @tags_data = map {
+			my $link = get_taxonomy_tag_and_link_for_lang($target_lc, $field_name, $_);
+			{
+				name => $link->{display},
+				url => "/facets/$field_name/" . $link->{tagurl},
+			}
+		} @$tags;
+
+		# Generation of html for product page
+		my $html = "<ul>\n";
```
stephanegigandet: The idea of having templates is that the HTML generation is done by the template, instead of having it in the Perl code. That way, if a designer wants to change the HTML, it can be done just by modifying the template. So instead of building $html, you just pass the tags_data, and the template generates the list (or a table or anything else).

tests/integration/expected_test_results/api_v2_product_read/get-fields-attribute-groups-all-knowledge-panels.json
```diff
@@ -3448,6 +3448,43 @@
...
+            ],
+            "evaluation" : "neutral",
+            "expand_for" : "large",
+            "level" : "info",
+            "title_element" : {
+               "title" : "Labels, certifications, awards"
+            },
+            "topics" : [
+               "problem"
+            ]
```
alexgarel: The topic here is strange !

alexgarel (approved):
Thank you @RayanGoubet 
Good work.

Let's see what it looks like on .net

cc: @teolemon @monsieurtanuki

PR #12119, opened 2025-07-03
taxonomy: Agaragar cooking helper

leonarf:
Category Agar agar is only in seafood for now, along side with fish and seaweed.
I added it in category cooking helpers

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -115932,6 +115932,7 @@ it: Prodotti a base di alghe
 nl: Producten op basis van algen
 
 < en:Seaweed products
+< en:Cooking helpers
 en: Agar-agar, Agar
 bg: Агар-агар
 de: Agar-Agar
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -115933,6 +115932,7 @@ it: Prodotti a base di alghe
 nl: Producten op basis van algen
 
 < en:Seaweed products
+< en:Cooking helpers
```
Freso: The automated test is now complaining about these not being sorted:
```suggestion
< en:Cooking helpers
< en:Seaweed products
```
(see <https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/16052972566/job/45349640987?pr=12119#step:7:10>)

leonarf: Fixed! Thanks for the explanation :-)

PR #12120, opened 2025-07-03
feat: OFF Days 2025 banner

stephanegigandet:
This PR replaces the top donation banner by an OFF day banner. (the code for the donation banner is still in the template so that we can put it back).

The banner design was done in Canva with elaborated button etc. so I just turned it into a PNG image (we lose the text and hover effect on button but that's ok...).

I made a less wide image for mobile, as otherwise it was not readable.

image

mobile:

image


JagjeevanAK (approved):
LGTM

PR #12131, opened 2025-07-07
feat: integrate Product Opener with recipe-estimator service

stephanegigandet:
This PR:
- allows Product Opener to make calls to external product services providers (e.g. recipe-estimator)
- adds an integration test for an external service (in the test Product Opener is in fact calling itself, but as an external service using HTTP)
- adds radio buttons in test_ingredients_analysis.pl to select the ingredient percent estimator: product_opener, recipe_estimator_glop, recipe_estimator_scipy

Caveats:
- the recipe_estimator services are not using exactly the product opener service format (the product data input and ouput is at the root of the request and response body, instead of being in a "product" key)
- URL for recipe_estimator is hardcoded, it will be changed to a configuration option --> DONE
- test_ingredients_analysis.pl will need to be modified to allow to specify key nutrition facts (currently none are set) --> DONE

image


lib/ProductOpener/APIProductServices.pm
```diff
@@ -90,11 +91,139 @@ use ProductOpener::HTTP qw/request_param/;
...
+		$body_ref->{fields} = $fields_ref;
+	}
+
+	# hack: recipe-estimator currently expects the product at the root of the body
+	if ($url =~ /estimate_recipe/) {
+		# We will send the product as the root object
+		$body_ref = $product_ref;
+	}
+
+	my $ua = create_user_agent();
```
alexgarel: Default timeout is 180s, we should really set it to a far more reasonable amount, as it will block the apache worker for that amount of time in case the service is not reachable. (10s ?)

```suggestion
	my $ua = create_user_agent(timeout => 10);
```

lib/ProductOpener/APIProductServices.pm
```diff
@@ -90,11 +91,139 @@ use ProductOpener::HTTP qw/request_param/;
...
 
+=head2 make_product_services_api_request ($request_ref, $product_ref, $url, $services_ref)
+
+Make a request to execute services on an external server using the product services API.
+
+e.g. this function is used to run the percent estimation service on the recipe-estimator server.
+
+=cut
+
+sub make_product_services_api_request ($request_ref, $product_ref, $url, $services_ref, $fields_ref = undef) {
```
alexgarel: I don't feel the name explains what we do.

What we do is that we modify the product based on an api request.

So I would call it
`add_product_data_from_external_service`

tests/integration/api_v3_product_services_request.t
```diff
@@ -0,0 +1,76 @@
...
+	]
+};
+
+# Make a call to an external product service
+# We use a Product Opener product service, but we call it as if it was an external service through HTTP
+
+my $services_url = construct_test_url("/api/v3/product_services");
+my $services_ref = ["estimate_ingredients_percent"];
+my $request_ref = {};
+init_api_response($request_ref);
```
alexgarel: FYI, in `tests/integration/run_cloud_vision_ocr.t` we use `fake_http_server` to really make an external service.

lib/ProductOpener/APIProductServices.pm
stephanegigandet: Good point, in fact even 2 seconds could be too long already, we will also call this function in update_all_products.pl in the future...

lib/ProductOpener/APIProductServices.pm
stephanegigandet: Well modifying product objects is the base concept of product services in general.
I'll rename it.

PR #12132, opened 2025-07-07
docs: OpenAPI documentation for new product nutrition data schema

Vic142:
The API v3 is going to use a new nutrition data schema for products. 
In this PR new OpenAPI documentation is created for the new product nutrition data schema (see [2025 - API - Nutrition Facts Schema and API](https://docs.google.com/document/d/19ZRrlWJraJm61E6U7AwxQ1uubPDvmSuNfl9F1oLC0Tg/edit?tab=t.0#heading=h.dooohcheu3im) for more information on the new schema).

Every nutrient in the  _nutrients.txt_ taxonomy file is added in the _nutrient_ field of the new nutrition schema documentation.


docs/api/ref/schemas/product_nutrition_v3.yaml
```diff
@@ -0,0 +1,814 @@
...
+        Most of these properties are read-only.
+
+        See [how to add nutrition data](https://openfoodfacts.github.io/openfoodfacts-server/api/ref-cheatsheet/#add-nutrition-facts-values-units-and-base)
+      properties:
+        nutrition:
+          type: object
+          description: |
+            todo
+          properties:
+            nutrient_sets_index:
```
stephanegigandet: For the selected proposal (see https://docs.google.com/document/d/19ZRrlWJraJm61E6U7AwxQ1uubPDvmSuNfl9F1oLC0Tg/edit?tab=t.0#heading=h.oqf07hovz1m5 ) we decided that we would not include a nutrients_sets_index, but that we would have a separate nutrient_set_preferred set instead

docs/api/ref/schemas/product_nutrition_v3.yaml
```diff
@@ -0,0 +1,814 @@
...
+                - the source of the data (e.g. *packaging*, *manufacturer*, *estimate*).
+
+                This structure allows capturing multiple nutritional profiles for a single product.
+              items:
+                type: object
+                properties:
+                  preparation:
+                    type: string
+                    enum:
+                      - as sold
```
stephanegigandet:
```suggestion
                      - as_sold
```

docs/api/ref/schemas/product_nutrition_v3.yaml
```diff
@@ -0,0 +1,814 @@
...
+                      For example, if the label states "per 250g", then this field should be *g*.
+                    example: "g"
+                  source:
+                    type: string
+                    enum:
+                      - packaging
+                      - manufacturer
+                      - usda
+                      - estimate
+                      - preferred
```
stephanegigandet: There's no "preferred" source for the option we selected

docs/api/ref/schemas/product_nutrition_v3.yaml
```diff
@@ -0,0 +1,814 @@
...
+                      When the data is given per serving, the actual unit that defines one serving may vary 
+                      between products and should be stored in this field.
+  
+                      This is essential to understand to which quantity values in `nutrients` apply for.
+
+                      For example, if the label states "per 250g", then this field should be *g*.
+                    example: "g"
+                  source:
+                    type: string
+                    enum:
```
stephanegigandet: I'm not sure about the enum there, as some clients could break if we add a new string (e.g. a new database). It may be better to list the possible values in the description, and state that new values may be added.

e.g. an article about how clients could try to handle unexpected valeus in enums in Flutter, but that's a lot of complexity on the client, and each client would need to do it: https://medium.com/@amjadhamdoun0/efficient-enum-handling-in-dart-with-json-converters-5d98dba17c8c

docs/api/ref/schemas/product_nutrition_v3.yaml
```diff
@@ -0,0 +1,814 @@
...
+                    type: integer
+                    description: |
+                      A timestamp indicating when this nutrition set was last updated.
+                    examples:
+                      - 1631270265
+                      - 1512153487
+                  nutrients:
+                    type: object
+                    description: |
+                      All known nutrients for the product.
```
stephanegigandet: We should indicate that new nutrients are regularly added, and that clients should not break if they encounter an unexpected nutrient.

docs/api/ref/schemas/product_nutrient_values.yaml
```diff
@@ -0,0 +1,103 @@
+type: object
+description: |
+  Quantity of sodium
```
stephanegigandet:
```suggestion
  Quantity of a nutrient
```

docs/api/ref/schemas/product_nutrient_values.yaml
```diff
@@ -0,0 +1,103 @@
...
+
+        🤓 code: see the [Units module][units-module],
+        and [Food:default_unit_for_nid function][default-unit]
+
+        [units-module]: https://openfoodfacts.github.io/openfoodfacts-server/dev/ref-perl-pod/ProductOpener/Units.html
+        [default-unit]: https://openfoodfacts.github.io/openfoodfacts-server/dev/ref-perl-pod/ProductOpener/Food.html#default_unit_for_nid_(_%24nid)
+
+        **Note**: For most use cases, you should not use this unit, as it can change depending on how the value was entered.
+        Instead use the `<nutrient>_100g` or `<nutrient>_serving` fields, and convert them to the unit you need.
+    enum:
```
stephanegigandet: I'm not sure we should list possible values in an enum, as clients may break if we add a new unit that they do not expect.

stephanegigandet: I'll ask for feedback in #api and #productopener

docs/api/ref/schemas/product_nutrient_values.yaml
```diff
@@ -0,0 +1,103 @@
...
+        "�e",
+        "�dh",
+        "gpg",
+      ]
+  modifier:
+    type: string
+    enum: ["<", "<=", "~", ">=", ">"]
+    description: |
+      todo (this property is optional)
+  source:
```
stephanegigandet: "source" and "source_per" would be only for the nutrient_set_preferred object, not for the objects in the nutrient_sets array

docs/api/ref/schemas/product_nutrient_values.yaml
```diff
@@ -0,0 +1,103 @@
...
+        "iu",
+        "mol/l",
+        "mval/l",
+        "ppm",
+        "�rh",
+        "�fh",
+        "�e",
+        "�dh",
+        "gpg",
+      ]
```
silkgh, line 87: What about foods sold as whole items? e.g. oranges etc - typically nutrition is given as "each"/ per "item" - in my own app I use "item" for this

stephanegigandet: @silkgh So far we have not encountered any source of nutrition data that gives nutrition facts per item (unless the item is equal to the serving, and in that case we have a serving quantity in g / ml). Do you have example of products where nutrition facts are listed per item without a specified weight / volume?

silkgh: While I agree that all foods should be recorded in valid, scientific measurements (e.g. per 100g), I find that pretty much all fruits are recorded per item - not saying it's right but it seems to be a thing in the wonderful world of nutrition

docs/api/ref/schemas/product_nutrient_values.yaml
Vic142: should I wait for other feedbacks on this before removing the enum ?

docs/api/ref/schemas/nutrients_v3.yaml
```diff
@@ -0,0 +1,694 @@
...
+    description: |
+      All known nutrients for the product.
+
+      You can get all possible nutrients from the
+      [nutrients taxonomy](https://static.openfoodfacts.org/data/taxonomies/nutrients.json)
+
+      New nutrients are regularly added.
+
+      Clients should not break if they encounter an unexpected nutrient to preserve compatibility.
+    properties:
```
alexgarel: Personally, here I would concentrate on main nutrients (energy, salt, sugar, etc) and let people see the rest with taxonomies, because I feel like it will be overwhelming otherwise to read this documentation), and also it makes a duplicate effort with the taxonomy to maintain this. (Otherwise we could seek to generate this file from the taxonomy, but again it would make it far too long a documentation IMO).

We can add a additionalProperty to support SDK generations tools, letting them know there are other values.

docs/api/ref/schemas/product_nutrient_values.yaml
```diff
@@ -0,0 +1,106 @@
...
+    type: number
+    description: |
+      todo
+    examples:
+      - 2
+      - 4.1
+  unit:
+    type: string
+    description: |
+        The unit of the value entered by the contributor (a user or the manufacturer), for the product as sold (not prepared).
```
alexgarel: Now it's not determined, it depends in which entry it's used.

```suggestion
        The unit of the value entered by the contributor (a user or the manufacturer), for the product.
```

docs/api/ref/schemas/product_nutrient_values.yaml
alexgarel: @Vic142 I would say no enum there, because you are creating an expectation that won't hold.

docs/api/ref/schemas/product_nutrient_values.yaml
alexgarel: so we must have a separate object product_nutrient_values_sources that we can add to preferred_set definitions (we already use a AllOf declaration)

docs/api/ref/schemas/product_nutrient_values.yaml
alexgarel: @silkgh you comment better applies to product_nutrition_properties.yaml, `per` property.

docs/api/ref/schemas/product_nutrition_properties.yaml
```diff
@@ -0,0 +1,46 @@
...
+      The nutrition data on the package can be per serving, per 100g or per 100ml.
+
+      This is essential to understand if values in the `nutrients` object apply for a serving, for 100g or for 100ml.
+  per_quantity:
+    type: integer
+    minimum: 0
+    description: |
+      The nutrition data on the package can be per serving, per 100g or per 100ml.
+      When the data is given per serving, the actual quantity that defines one serving may vary 
+      between products and should be stored in this field.
```
alexgarel: I would remove the "should be" the API defines things
```suggestion
      between products and is stored in this field.
```

alexgarel: nit: when I break line in markdown I prefer to do it in meaninful places.
Because if you edit this text afterwards, you would have to redo all the line splitting if your text does not fit anymore. So it's better to treat it as code, and make meaningfull lines.
Like:
```suggestion
      The nutrition data on the package can be per serving, per 100g or per 100ml.
      When the data is given per serving,
      the actual quantity that defines one serving may vary between products
      and should be stored in this field.
```

html/images/misc/android-apk.svg
alexgarel, line 1: why changes in this file ?

docs/api/ref/schemas/product_nutrient_values.yaml
stephanegigandet: "I find that pretty much all fruits are recorded per item" --> @silkgh are you talking about input data? i.e. that we get from the packaging that 1 apple = 50g of sugars?  or about a desired output, that you would like to get the OFF API to tell you that an apple is 50 of sugars?
Because I have never seen a input source that specifies nutrients per item. EU and US law etc. mandate that the nutrition facts are for a specific quantity (which can be equal to 1 serving which is 1 item, but it needs to specifiy how much g it is)

docs/api/ref/schemas/nutrients_v3.yaml
stephanegigandet: That sounds like a good option to me. We can specficially name only the "mandatory" nutrients for energy-kj, energy-kcal, proteins, fats, saturated-fats, carbohydrates, sugars, fibers and salt. Maybe total_carbohydrates as well, and whatever nutrients are mandatory in the US and Canada (e.g. calcium and a few others)

html/images/misc/android-apk.svg
Vic142: i didn't touch it and the changes don't appear in my branch locally so I don't know how it appeared on GitHub, I'm going to check why it says there were changes

docs/api/ref/schemas/nutrients_v3_with_source.yaml
```diff
@@ -0,0 +1,397 @@
...
+      Clients should not break if they encounter an unexpected nutrient to preserve compatibility.
+    properties:
+      added-sugars:
+        description: |
+          Quantity of added sugars
+
+          (per 100g, per 100ml or per serving) in a standard unit (g or ml)
+        allOf:
+          - $ref: "../schemas/product_nutrient_values.yaml"
+          - type: object
```
stephanegigandet: Turn this into a ref to a new product_nutrient_source.yml

docs/api/ref/schemas/product_nutrition_properties.yaml
```diff
@@ -0,0 +1,46 @@
...
+    enum:
+      - 100g
+      - 100ml
+      - serving
+    description: |
+      The nutrition data on the package can be per serving, per 100g or per 100ml.
+
+      This is essential to understand if values in the `nutrients` object apply for a serving, for 100g or for 100ml.
+  per_quantity:
+    type: integer
```
stephanegigandet: This should be a float. It's rare, but a few products have quantities with floating point. e.g. this old spray has a serving quantity of 0.25 ml: https://world.openfoodfacts.org/product/4250519660295/2-calories-oil-spray-esn

docs/api/ref/schemas/product_nutrition_v3.yaml
```diff
@@ -0,0 +1,79 @@
...
+            Nutrition fields of a product
+
+            Most of these properties are read-only.
+
+            See [how to add nutrition data](https://openfoodfacts.github.io/openfoodfacts-server/api/ref-cheatsheet/#add-nutrition-facts-values-units-and-base)
+          properties:
+            nutrient_set_preferred:
+              type: object
+              description: | 
+                A set that combines set of nutriments with the preferred sources, with normalized units.
```
stephanegigandet:
```suggestion
                A set that combines nutrient data from preferred sources, with normalized units.
```

docs/api/ref/schemas/product_nutrition_v3.yaml
```diff
@@ -0,0 +1,79 @@
...
+              items:
+                type: object
+                allOf:
+                  - $ref: "../schemas/nutrients_v3_base.yaml"
+                  - $ref: "../schemas/product_nutrition_properties.yaml"
+                  - type: object
+                    properties:
+                      source:
+                        type: string
+                        enum:
```
stephanegigandet: We should make those examples instead of enum, in case we add a source

PR #12135, opened 2025-07-08
fix: Move all extra_hosts under backend-conf in dev.yml

tuukka:
Docker Compose documentation explains:
"YAML merge only applies to mappings, and can't be used with sequences."


alexgarel (approved):
LGTM, thanks.

PR #12141, opened 2025-07-09
taxonomy: Various updates/additions

Freso:
Unrecognised ingredients etc. from a number of products:
https://us.openfoodfacts.org/product/00481403/chewable-peppermint-flavored-melatonin-trader-darwin-s
https://se.openfoodfacts.org/product/7312220096019/biscoff-lotus

and some Danish and Swedish translations I did along the way.

### DIFF
--- a/taxonomies/nutrients.txt
+++ b/taxonomies/nutrients.txt
@@ -6678,6 +6678,12 @@ unit:en: g
 wikidata:en: Q243309
 wikipedia:en: https://en.wikipedia.org/wiki/Carnitine
 
+zz: melatonin
+xx: Melatonin
+unit:en: µg
+wikidata:en: Q180912
+wikipedia:en: https://en.wikipedia.org/wiki/Melatonin
+
 zz: methylsulfonylmethane
 xx: MSM, OptiMSM
 en: Methylsulfonylmethane, Dimethyl sulfone, Methyl sulfone
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -701,6 +701,7 @@ It is a list of nutrients names with eventual prefixes and suffixes:
 			'glycemic-index-', 'water-hardness-',
 			'sulfate-', 'nitrate-',
 			'acidity-', 'carbohydrates-',
+			'melatonin-',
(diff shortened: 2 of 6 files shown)


taxonomies/nutrients.txt
```diff
@@ -6678,6 +6678,12 @@ unit:en: g
 wikidata:en: Q243309
 wikipedia:en: https://en.wikipedia.org/wiki/Carnitine
 
+zz: melatonin
+xx: Melatonin
```
jayaddison, line 6682: NB: I think the format of this taxonomy file requires an `en:`-language entry for a nutrient for it to be registered properly -- I'll file a small fixup pull request for this soon (and will check for any other affected nutrients in the file).

jayaddison: ...for context: I noticed this while attempting to add melatonin as a nutrient for product: https://world.openfoodfacts.org/product/0027917000275/blackberry-extra-strength-melatonin-5-mg-gummies-vitafusion

Freso: Huh. I don’t remember having an issue when adding it to https://world.openfoodfacts.org/product/00481403/chewable-peppermint-flavored-melatonin-trader-darwin-s

jayaddison: Hrm - ok, then perhaps I'm mistaken about the missing prefix in the taxonomy entry being the problem.  I'll try to find some time tomorrow to test it out locally.

jayaddison: (and thanks for mentioning that it seemed to work fine previously!)

jayaddison: Mystery solved; `melatonin` is available in autosuggest when using the United States domain/instance of OpenFoodFacts.  Also posted to: https://github.com/openfoodfacts/openfoodfacts-server/pull/13050#issuecomment-3891821279

Freso: Ah, yes, indeed. I did add that as part of this PR, but I only did it for US, since that was the market for the product I was adding, and I didn’t want to do “too much”. :) Glad you figured it out!

PR #12165, opened 2025-07-21
taxonomy: stopwords allergens czech

moon-rabbitOFF:
I added a line for czech stopwords with jiné (other) and výrobek (products)


stephanegigandet (approved):
Thanks

PR #12169, opened 2025-07-23
feat: add panel to identify added sugars

stephanegigandet:
New panel to identify added sugars:

image


CharlesNepote (approved):

CharlesNepote (approved):

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -68,7 +68,7 @@ use ProductOpener::Display qw/:all/;
 use ProductOpener::Images qw/data_to_display_image/;
 use ProductOpener::HTTP qw/single_param/;
 use ProductOpener::PackagerCodes qw/%packager_codes/;
-use ProductOpener::KnowledgePanelsIngredients qw/create_ingredients_list_panel/;
+use ProductOpener::KnowledgePanelsIngredients qw/:all/;
```
alexgarel, line 71: why use all here, if we only use create_ingredients_added_sugars_panel ?

alexgarel: hum sorry, I misread, there are a lot of functions so this is fair.

PR #12170, opened 2025-07-23
feat: preferred set generation

Vic142:
The new nutrition facts schema is going to have a preferred nutrients set generated from a list of nutrient sets (see [2025 - API - Nutrition Facts Schema and API](https://docs.google.com/document/d/19ZRrlWJraJm61E6U7AwxQ1uubPDvmSuNfl9F1oLC0Tg) for more information on the new schema). In this PR, a new module is created to generate the preferred nutrients set from a given nutrient sets list. 

The new module currently only generates the preferred set. Other PRs will aim to complete the module to upgrade the nutrition schema.


lib/ProductOpener/Nutrition.pm
```diff
@@ -0,0 +1,284 @@
...
+sub generate_nutrient_set_preferred_from_sets {
+    my ($nutrient_sets_ref) = @_; 
+    my @nutrient_sets = @$nutrient_sets_ref;
+    
+    my $nutrient_set_preferred_ref = {};
+    
+    if (@nutrient_sets) {
+        @nutrient_sets = sort_sets_by_priority(@nutrient_sets);
+
+        if (%{$nutrient_sets[0]}) {
```
stephanegigandet: What is the use of the confition on %{$nutrient_sets[0]} ? Is there a case where it would be not true?

lib/ProductOpener/Nutrition.pm
```diff
@@ -0,0 +1,284 @@
...
+        manufacturer => 0,
+        packaging => 1,
+        usda => 2,
+        estimate => 3,
+        _default => 4,
+    );
+
+    my %per_priority = (
+        "100g" => 0,
+        "100ml" => 0,
```
stephanegigandet: It's best to put different priority values for 100g and 100ml, so that we always have the same outcome. There are a few products (e.g. ice cream) that may have nutrition sets for both 100g and 100ml (with differing values). We need to make sure the preferred set is always computed in the same way.

lib/ProductOpener/Nutrition.pm
```diff
@@ -0,0 +1,284 @@
...
+The generated set is a combined set of nutrients with the preferred sources, per references and preparation states 
+and with normalized units.
+
+=head3 Return values
+
+The generated preferred nutrient set
+
+=cut
+
+sub generate_nutrient_set_preferred_from_sets {
```
stephanegigandet: We use signatures to pass function parameters, instead of using @_ :

```suggestion
sub generate_nutrient_set_preferred_from_sets ($nutrient_sets_ref) {
```

lib/ProductOpener/Nutrition.pm
```diff
@@ -0,0 +1,284 @@
...
+        if (    defined $nutrient_set_ref->{preparation}
+            and $nutrient_set_ref->{preparation} eq $nutrient_set_preferred_ref->{preparation}
+            and exists $nutrient_set_ref->{nutrients} 
+            and ref $nutrient_set_ref->{nutrients} eq 'HASH') 
+        {
+            foreach my $nutrient (keys %{$nutrient_set_ref->{nutrients}}) {
+                # for each nutrient, set its values if values are not already present in preferred set
+                # (ie if nutrient not present in other set with highest priority)
+                if (!exists $nutrient_set_preferred_ref->{nutrients}{$nutrient}) {
+                    $nutrient_set_preferred_ref->{nutrients}{$nutrient} = $nutrient_set_ref->{nutrients}{$nutrient};
```
stephanegigandet: When you do this, you are not creating a new copy of all the values of $nutrient_sets_ref->{nutrients}{$nutrients}, which means than the 2 functions nutrient_in_standard_unit() and nutrient_in_wanted_per() will in fact also change the input nutrient sets.

I made a new test so that you can see it: https://github.com/openfoodfacts/openfoodfacts-server/pull/12173

To solve it, you can use the clone function from
use Clone qw/clone/;

lib/ProductOpener/Nutrition.pm
```diff
@@ -0,0 +1,284 @@
...
+                    nutrient_in_standard_unit($nutrient_set_preferred_ref->{nutrients}{$nutrient}, $nutrient);
```
stephanegigandet: Could you rename the two functions so that it's clear that they are updating the passed value?
e.g. convert_nutrient_to_standard_unit()
With the current name, I was expecting that they would not modify the input, but return the converted value.

stephanegigandet (commented):
Thanks for including all the unit tests!

lib/ProductOpener/Nutrition.pm
Vic142: In case the first set is empty ({}) then I wanted to generate an empty set too

I don't know if this can happen, but I decided to test this in the unit tests. If the first set is empty after sorting the array then all the sets in the array are empty and then I wanted to return an empty set so I added this condition to add values to _preparation_, _per_, _per\_quantity_ and _per\_unit_ in the generated set.
If there can never be empty sets then this test is probably useless and I will remove it

lib/ProductOpener/Nutrition.pm
Vic142: I didn't think of that, it's a good point. I changed it to prioritize 100g then 100ml

lib/ProductOpener/Nutrition.pm
Vic142: ok thanks that's done

lib/ProductOpener/Nutrition.pm
Vic142: that's done thank you!

lib/ProductOpener/Nutrition.pm
Vic142: done

tests/unit/nutrition.t
```diff
@@ -7,1017 +7,765 @@ use Test2::V0;
...
+		# Generated set should have converted per nutrients given nutrients per in different serving units with wanted in g
+		"convert_per_nutrients_when_different_per_units_g",
+		{
+			nutrition => {
+				nutrient_sets => [
+					{
+						preparation => "as_sold",
+						per => "serving",
+						per_quantity => "100",
+						per_unit => "mg",
```
stephanegigandet, line 674: This test is a bit weird because we have 2.5 g of sodium in a 50 mg serving. Can you change it to possible values? When I was looking at the results I thought there was a mistake.

stephanegigandet (approved):
Thank you @Vic142 , it looks good to me. Could you just change the values on the tests where there are more grams of nutrients than the small 50mg serving size?

I also think we should have 2 preferred sets, 1 for 100g/100ml and one per serving, but we can discuss it and do it in another PR, it should not prevent to merge this one.

PR #12171, opened 2025-07-24
docs: a few fixes in the comments for Nutriscore

jforget:
A few fixes in the comments.

Some remarks. I do not know what is the use of the
`[nutrient]_points_max` property, so the comment contains question
marks.

In the 2023 version, I have added a comment for nutrient `salt` but I
have not removed `sodium`. The provisional comment for `salt` contains
only question marks, please write something interesting.

There is a problem with the signatures of functions
`get_value_with_one_less_negative_point_2023` and
`get_value_with_one_more_positive_point_2023`, but it is given in a
separate issue.

### DIFF
--- a/lib/ProductOpener/Nutriscore.pm
+++ b/lib/ProductOpener/Nutriscore.pm
@@ -1,7 +1,7 @@
 # This file is part of Product Opener.
 #
 # Product Opener
-# Copyright (C) 2011-2023 Association Open Food Facts
+# Copyright (C) 2011-2025 Association Open Food Facts
 # Contact: contact@openfoodfacts.org
 # Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
 #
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Nutriscore.pm
```diff
@@ -509,13 +510,15 @@ If the product is a beverage, water, cheese, or fat, it must contain a positive
...
 - is_fat_oil_nuts_seeds
+- is_red_meat_product
 
 =head4 Output keys: details of the Nutri-Score computation
 
 Returned values:
 
 - [nutrient]_value -> rounded values for each nutrient according to the Nutri-Score rules
 - [nutrient]_points -> points for each nutrient
+- [nutrient]_points_max -> ???
```
stephanegigandet:
```suggestion
- [nutrient]_points_max -> maximum number of points that can be allocated for each specific nutrient
```

stephanegigandet:
Thanks for the PR. Could you run "make lint"? The /lint command only works for branches on the OFF repo, not forked repos.

teolemon:
@jforget 2 small merge conflicts to fix

teolemon:
also, @jforget if you can comment /lint here

PR #12173, opened 2025-07-24
fix: new tests for preferred nutrition set generation

stephanegigandet:
What we are doing in many unit tests is to give as input a product object, and then we store the updated product object and compare it to the version we expect.

I just added one of the tests to show why this is useful: generate_nutrient_set_preferred_from_sets() generates the correct preferred set, but it changed the input nutrient sets.

So I think it would be best to convert all tests to this format.

### DIFF
--- a/tests/unit/expected_test_results/nutrition/convert_per_nutrients.json
+++ b/tests/unit/expected_test_results/nutrition/convert_per_nutrients.json
@@ -1,0 +1,60 @@
+{
+   "nutrition" : {
+      "nutrient_set_preferred" : {
+         "nutrients" : {
+            "sodium" : {
+               "source" : "packaging",
+               "source_per" : "serving",
+               "unit" : "g",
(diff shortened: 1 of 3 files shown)


tests/unit/expected_test_results/nutrition/convert_per_nutrients.json
```diff
@@ -0,0 +1,60 @@
...
+         "preparation" : "as_sold"
+      },
+      "nutrient_sets" : [
+         {
+            "nutrients" : {
+               "sodium" : {
+                  "source" : "packaging",
+                  "source_per" : "serving",
+                  "unit" : "g",
+                  "value" : 1.25,
```
stephanegigandet, line 32: The input value was 2.5, but it was changed to 1.25.

stephanegigandet:
Incorporated in another PR, closing this one

PR #12177, opened 2025-07-25
fix: response schema for GET /api/v3/product/[barcode]

### LINKED ISSUE #12157: OpenAPI spec for GET /api/v3/product/{barcode} has wrong return type
In OFF NodeJS SDK, we use a tool to generate api bindings from the openapi spec. 

The `GET /api/v3/product/{barcode}` has the wrong return type typing.

https://github.com/openfoodfacts/openfoodfacts-server/blob/c1837c1a1bc8f7acfdcd4cdbffbe769c2c56906b/docs/api/ref/api-v3.yaml#L90-L92

With this, we get
```js
content: {
  "application/json": components["schemas"]["response_status"] & components["schemas"]["product_v3"];
};
```

which means that the product fields are inlined with the response_status, like

```js
{
  // response fields
  status_id?: "success" | "success_with_warnings" | "success_with_errors" | "failure";
  result?: { ... };
  ... // from now on, product fields
  product_name?: string;
  ...
}
```

while instead the right type would be something like
```js
{
  // response fields
  status_id?: "success" | "success_with_warnings" | "success_with_errors" | "failure";
  result?: { ... };
  ...
  product?: {
    product_name?: string;
  }
  ...
}
```

stephanegigandet:
fixes #12157

### DIFF
--- a/docs/api/ref/api-v3.yaml
+++ b/docs/api/ref/api-v3.yaml
@@ -89,7 +89,10 @@ paths:
               schema:
                 allOf:
                   - $ref: ./responses/response-status/response_status.yaml
-                  - $ref: ./schemas/product_v3.yaml
+                  - type: object
+                    properties:
+                      product:
+                        $ref: ./schemas/product_v3.yaml
         "302":
           description: Redirect to the correct server for the product type of the requested product
           headers:


alexgarel (approved):
Thanks

PR #12178, opened 2025-07-25
fix: export images on pro platform

stephanegigandet:
The export of images from the pro platform was broken by the recent change of images schema.


stephanegigandet:
Deployed on pro platform.

lib/ProductOpener/Export.pm
alexgarel, line 1: FTR: this is a good example why this is problematic that we don't have good "encapsulation" of this kind of logics in functions.

PR #12189, opened 2025-07-30
docs: small rework on how-to-write-and-run-tests

alexgarel:
to better point at stored expected results


RayanGoubet (approved):
Thank's for that !

stephanegigandet (approved):

Issue #7249, opened 2022-08-25
Remove or move entries with non normalized codes in mongodb


alexgarel:
We have entries on the file system or in mongodb with a non normalized code, (see Products.pm::normalize_code). This leads to confusion for people using the data dump (see #7244)

Run a script to clean those entries. For each entry with a non normalized code:
- always remove the non normalized version in mongodb
- for the sto file:
  - if an entry with the normalized code exists: remove the non normalize one
  - else move the entry to its normalized code

Note: be sure to double search the non normalized codes : in mongodb and on the filesystem

alexgarel:
First dry-run on preprod, it was really really slow. I must add some progression indicators to understand what is the slowest. 
```
real    3608m29.575s
user    11m10.066s
sys     34m11.942s
```

Difference user / real indicates it spends a lot of time waiting for something (file system). I suspect we have a problem with zfs mount on preprod. (see also  #6164). Maybe it's just that we need to avoid being on an old version of the ZFS, and for that we need to do something ! (maybe mount / unmount the file system, I'm not sure).

Finally it died on mongodb stuff.

```
MongoDB::DatabaseError: Executor error during find command :: caused by :: operation exceeded t
ime limit
```

I attach the [log of advertised changes from the first phase](https://github.com/openfoodfacts/openfoodfacts-server/files/10478530/2023-01-23-logs-fix-barcodes.txt) (it's a dry run, so it's not done yet). 



I will:
* add logs to get progression (with a specific option)
* try to see if I can remount zfs and see if it brings a change
* add options to only run part of the script (that make sense) to be able to jump to one part of the script, it will help debugging

alexgarel:
Since the first try on preprod, we have moved staging closer to the ntfs mount so it should be feasible to test it on staging.

alexgarel:
> Since the first try on preprod, we have moved staging closer to the ntfs mount so it should be feasible to test it on staging.

Even like that, it's not feasible, just listing 100000 products takes more than 40 minutes
```bash
time ( find /mnt/podata/products/ -name "product.*" |head -n 100000 )
...
real    46m54.424s
user    0m3.384s
sys     0m27.930s
```

PR #12224, opened 2025-08-04
fix: enable fixing non normalized codes only in mongodb

alexgarel:
Transforming the script to, at least, be able to fix mongodb.

Part of #7249


benbenben2 (approved):
LGTM

PR #12227, opened 2025-08-05
build: add a stop_deps target

alexgarel:
Because it's annoying to have a lot of running containers

See also:
* https://github.com/openfoodfacts/openfoodfacts-auth/pull/235
* https://github.com/openfoodfacts/openfoodfacts-shared-services/pull/15
that adds the "stop" targets to dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,6 +183,10 @@ restart: run_deps
 	${DOCKER_COMPOSE} restart backend frontend
 	@echo "🥫  started service at http://openfoodfacts.localhost"
 
+stop: stop_deps
+	@echo "🥫 Stopping containers …"
+	${DOCKER_COMPOSE} stop
+
 status: run_deps
 	@echo "🥫 Getting container status …"
 	${DOCKER_COMPOSE} ps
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -598,6 +598,11 @@ prune_deps: clone_deps
 		cd ${DEPS_DIR}/$$dep && $(MAKE) prune; \
 	done
 
+stop_deps:
+	@for dep in ${DEPS} ; do \
```
alexgarel, line 606: I have the docker compose stop fallback, until we have stop target in each dependency. (but in reality, for openfoodfacts-auth it does not work, because we need the `RUN` version of docker compose...)

Makefile
```diff
@@ -183,6 +183,10 @@ restart: run_deps
 	${DOCKER_COMPOSE} restart backend frontend
 	@echo "🥫  started service at http://openfoodfacts.localhost"
 
+stop: stop_deps
+	@echo "🥫 Stopping containers …"
+	${DOCKER_COMPOSE} stop
```
john-gom, line 188: Should the stop not also call stop_deps?

alexgarel: Done, thanks (as a dependency of the stop target)

john-gom: Hi @alexgarel . I don't think you can add it as a dependency of the target as you need to stop the current service before you stop the things it depends on

PR #12231, opened 2025-08-05
feat: upgrade product version nutrition

Vic142:
The products are going to use a new nutrition schema (see [2025 - API - Nutrition Facts Schema and API](https://docs.google.com/document/d/19ZRrlWJraJm61E6U7AwxQ1uubPDvmSuNfl9F1oLC0Tg) for more information on the new schema). In this PR an update function of product version from 1002 to 1003 is created to update the nutrition schema of the products.
The downgrade function from 1003 to 1002 will be done in an other PR.


teolemon:
🥫 Checking with perltidy lib/ProductOpener/Nutrition.pm lib/ProductOpener/ProductSchemaChanges.pm lib/ProductOpener/Test.pm tests/unit/product_schema_changes.t
test -z "lib/ProductOpener/Nutrition.pm lib/ProductOpener/ProductSchemaChanges.pm lib/ProductOpener/Test.pm tests/unit/product_schema_changes.t" || COMPOSE_FILE="docker-compose.yml;docker/dev.yml" docker compose --env-file=.env  run --rm --no-deps backend perltidy --assert-tidy -opath=/tmp/ --standard-error-output lib/ProductOpener/Nutrition.pm lib/ProductOpener/ProductSchemaChanges.pm lib/ProductOpener/Test.pm tests/unit/product_schema_changes.t

tests/unit/product_schema_changes.t: Begin Error Output Stream
tests/unit/product_schema_changes.t: assertion failure: '--assert-tidy' is set but output differs from input
tests/unit/product_schema_changes.t: 
tests/unit/product_schema_changes.t: Input   file length has 14936 chars in 679 lines
tests/unit/product_schema_changes.t: Output  file length has 14329 chars in 678 lines
tests/unit/product_schema_changes.t: Files first differ at character 1 of line 18
tests/unit/product_schema_changes.t:  16:#set_fixed_time(1650000000);  # freeze time to a known epoch
tests/unit/product_schema_changes.t: <18:
tests/unit/product_schema_changes.t: >18:my ($test_id, $test_dir, $expected_result_dir, $update_expected_results)...
tests/unit/product_schema_changes.t:     ^
tests/unit/product_schema_changes.t:

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -321,3 +323,123 @@ sub convert_schema_1002_to_1001_refactor_images_object ($product_ref) {
...
+	if (defined $product_ref->{nutriments}) {
+		my %hash_nutrients = map {/^([a-z][a-z\-]*[a-z]?)(?:_\w+)?$/ ? ($1 => 1) : ()}
+			keys %{$product_ref->{nutriments}};
+
+		my @nutrients = keys %hash_nutrients;
+
+		foreach my $state (keys %$new_nutrition_sets_ref) {
+			$new_nutrition_sets_ref->{$state}{preparation} = $nutrition_preparations_ref->{$state}{state};
+			$new_nutrition_sets_ref->{$state}{source} = "packaging";
+			$new_nutrition_sets_ref->{$state}{last_updated_t} = time() + 0;
```
alexgarel: would be great (IMO) to have the same time for all preferred_sets, so I would have a `$time` variable with `$time = time()` at the start of the function.

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -321,3 +323,123 @@ sub convert_schema_1002_to_1001_refactor_images_object ($product_ref) {
...
+		"serving" => {state => "as_sold", modifier_state => ""},
+	};
+
+	if (defined $product_ref->{nutriments}) {
+		my %hash_nutrients = map {/^([a-z][a-z\-]*[a-z]?)(?:_\w+)?$/ ? ($1 => 1) : ()}
+			keys %{$product_ref->{nutriments}};
+
+		my @nutrients = keys %hash_nutrients;
+
+		foreach my $state (keys %$new_nutrition_sets_ref) {
```
alexgarel: It's a bit misleading to call it state, as we have state inside the new_nutrition_sets_ref.

Maybe call it $set_type ?

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -321,3 +323,123 @@ sub convert_schema_1002_to_1001_refactor_images_object ($product_ref) {
...
+						= $product_ref->{nutriments}{$nutrient . '_' . $state};
+
+					$new_nutrition_sets_ref->{$state}{nutrients}{$nutrient}{value_string}
+						= sprintf("%s", $product_ref->{nutriments}{$nutrient . '_' . $state});
+					$new_nutrition_sets_ref->{$state}{nutrients}{$nutrient}{unit} = set_nutrient_unit($nutrient);
+
+					if (
+						defined $product_ref->{nutriments}
+						{$nutrient . ($nutrition_preparations_ref->{$state}{modifier_state}) . "_modifier"})
+					{
```
alexgarel: can you comment a bit  the why ?

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -321,3 +323,123 @@ sub convert_schema_1002_to_1001_refactor_images_object ($product_ref) {
...
+						$new_nutrition_sets_ref->{$state}{nutrients}{$nutrient}{modifier} = $product_ref->{nutriments}
```
alexgarel: I would use more variables to avoid very long things like that which are really difficult to read (IMO).

Like having a variable for `$new_nutrition_sets_ref->{$state}{nutrients}{$nutrient}`
and for `$product_ref->{nutriments}{$nutrient . $nutrition_preparations_ref->{$state}`

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -321,3 +323,123 @@ sub convert_schema_1002_to_1001_refactor_images_object ($product_ref) {
...
+							{$nutrient . $nutrition_preparations_ref->{$state}{modifier_state} . "_modifier"};
+					}
+				}
+
+				elsif (defined $product_ref->{nutriments}{$nutrient . "_modifier"}
+					and $product_ref->{nutriments}{$nutrient . "_modifier"} eq "-")
+				{
+					push(@{$new_nutrition_sets_ref->{$state}{unspecified_nutrients}}, $nutrient);
```
alexgarel: I would read that from the start, because you should not add this nutrient in the set in this case !

lib/ProductOpener/ProductSchemaChanges.pm
alexgarel:
```suggestion
		# Generates the nutrition sets,
		# which, for old data, are all from source "packaging"
		foreach my $state (keys %$new_nutrition_sets_ref) {
```

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -321,3 +323,123 @@ sub convert_schema_1002_to_1001_refactor_images_object ($product_ref) {
...
+
+	my $new_nutrition_sets_ref = {
+		"prepared_100g" => {},
+		"100g" => {},
+	};
+	# only create serving sets if we know the serving quantity
+	if (defined $product_ref->{serving_quantity}) {
+		$new_nutrition_sets_ref->{"prepared_serving"} = {};
+		$new_nutrition_sets_ref->{"serving"} = {};
+	}
```
alexgarel: It seems to me that what to include should depends on the nutrition_data and/or nutrition_data_prepared value. Because the other values are hidden today, and might not be valid. (not 100% confident of what I'm saying).

alexgarel: Also there are cases where we have the serving set, but not the serving_quantity,
in this case we want to only generate the "serving" nutrition set (and not the "100g" one)

alexgarel: Also one last point to consider @Vic142 
If no_nutrition_data is "on" we must ignore everything ! (it says there is no packaging information)

stephanegigandet: So when we upgrade products, the per of the "source" nutrients we have are determined by the values of the fields "nutrition_data_per" and "nutrition_data_prepared_per". It can contain "100g" or "serving".

If it's 100g, then we copy the values from nutriments.salt_100g, if it's serving then we copy the values from salt_serving. We never copy both, as today we have only 1 input table (either per 100g or per serving), and the other data is automatically computed from it.

The fact that we have a serving_quantity or not is not taking into account to decide which input nutrient set to create.

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -331,115 +334,215 @@ The nutrition schema is updated to allow storing several nutrition sets
...
+				= defined $product_ref->{nutrition_data_per} && $product_ref->{nutrition_data_per} eq "serving";
+
+			if ($nutrition_given_for_100g) {
+				$new_nutrition_sets_ref->{"100g"} = {};
+			}
+			elsif ($nutrition_given_for_serving) {
+				$new_nutrition_sets_ref->{serving} = {};
+			}
+		}
+		if (!$no_nutrition_data && $nutrition_given_as_prepared) {
```
stephanegigandet: there's already a condition on $no_nutrition_data, so it cannot be false here

teolemon:
@stephanegigandet @Vic142 is this PR still relevant ? it is conflicted.

teolemon:

stephanegigandet:
Closing old PR, already merged

PR #12234, opened 2025-08-06
ci: Update labeler.yml for nutrients and units taxonomies

teolemon:
### What
ci: Update labeler.yml for nutrients and units taxonomies


Freso (approved):
LGTM.

PR #12235, opened 2025-08-06
feat: external sources of knowledge panels with optin

IWouldShiverTheWholeNightThrough:
### What
Add full integration for external knowledge panels with strict JSON ordering (JS, HTML, external-sources.json).

The architecture is generalizable and generic: it can support multiple NGOs, associations, or external resources wishing to “plug in” their knowledge panels, without requiring custom logic for each source. Any number of providers can add their own panels by updating the JSON file; everything else works out of the box.

Each external panel is grouped by section (as defined in the JSON). Preferences are rendered as checkboxes per panel, within their respective sections.

User preferences are fully decoupled: a panel in one section is managed independently of the same panel ID in another section (thanks to a unique key combining section and panel).

Selections and display update live and consistently:

The main display always shows only the panels checked by the user, grouped and sorted by section, then by panel (following the JSON order).

The sticky navbar is dynamically updated to exactly match the visible sections and their order, ensuring perfect UI synchronization.

The preferences UI is now embedded in the main user preferences form, using product-preferences.js to inject the external panels and manage all event handling.

Scope/moderator restriction is now enforced in the frontend: Panels marked with "scope": "moderators" are visible to moderators and admins only.

Includes all necessary changes in Perl (lib/ProductOpener/Display.pm) for backend/frontend data flow.

Limitations and Next Steps
Language management is not yet implemented.

### Screenshot

image

image

### DIFF
--- a/lib/ProductOpener/APITest.pm
+++ b/lib/ProductOpener/APITest.pm
@@ -339,9 +339,16 @@ Create a moderator user, implies creation of an admin
 =head3 Returns
 
 A hashmap associating user with their user agent:
+
+=over
+
 =item user: normal user
+
 =item admin: admin user
+
 =item moderator: moderator user
+
+=back
 
 =cut
 


teolemon:
image
https://www.figma.com/design/VjBLByG8mzdKjCJkruqj62/3rd-party-knowledge-panels---Open-The-App?node-id=2205-2&p=f&t=1ZyFq77NuQHr6Ddr-0

html/js/external-knowledge-panels.js
```diff
@@ -0,0 +1,272 @@
...
+    });
+    allPanelsBySection = sections;
+}
+
+/**
+ * Returns a pretty string for section names.
+ */
+function prettySectionName(sectionId) {
+    return sectionId.replace(/_/g, ' ').replace(/\b\w/g, c => c.toUpperCase());
+}
```
alexgarel, line 50: We must open a bug to fix this better (with translations).

But ok if it's not in this PR.

html/js/external-knowledge-panels.js
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Returns whether the panel should be shown (opt-in), defaulting to true.
+ */
+function getExternalKnowledgePanelsOptin(sectionId, panelId) {
+    const val = localStorage.getItem('external_panel_' + sectionId + "_" + panelId);
+    return val === null ? true : val === "true";
+}
```
alexgarel, line 105: It must default to false, that's the meaning of opt-in ! (vs opt-out)

```suggestion
/**
 * Returns whether the panel should be shown (opt-in), defaulting to false.
 */
function getExternalKnowledgePanelsOptin(sectionId, panelId) {
    const val = localStorage.getItem('external_panel_' + sectionId + "_" + panelId);
    return val === null ? false : val === "true";
}
```

html/js/external-knowledge-panels.js
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Sets the opt-in value for a panel.
+ */
+function setExternalKnowledgePanelsOptin(sectionId, panelId, enabled) {
+    localStorage.setItem('external_panel_' + sectionId + "_" + panelId, enabled);
```
alexgarel: Why use sectionId ? For me panelID should be unique in all external-sources. So adding sectionId is not necessary.

html/resources/files/external-sources.json
alexgarel, line 1: Could we add an OpenAPI description for this file (the API being a get of the file url) ? (in docs/api/ref/api-v3.yaml)

lib/ProductOpener/Display.pm
```diff
@@ -8568,6 +8568,18 @@ JS
...
 	my $html_display_product;
+
+	use JSON;
+
+	my $categories_json = encode_json($product_ref->{categories_tags});
+
+	$template_data_ref->{category} = $categories_json;
+	$template_data_ref->{product_type} = $product_ref->{product_type};
+
+	use ProductOpener::Users qw/%User is_admin_user/;
```
alexgarel: Please put it on top of file.

html/js/external-knowledge-panels.js
```diff
@@ -0,0 +1,272 @@
...
+    let insertAfter = matchSection;
+    const visibleSectionsOrdered = [];
+    allPanelsBySection.forEach(section => {
+        const visiblePanels = section.panels.filter(panel => {
+            const f = panel.filters || {};
+            const matchCat = !f.categories || !f.categories.length || f.categories.some(cat => category.includes(cat));
+            const matchCountry = !f.countries || !f.countries.length || f.countries.some(c => country.includes(c));
+            const matchLang = !f.languages || !f.languages.length || f.languages.some(l => language.includes(l));
+            const matchType = !f.product_types || !f.product_types.length || f.product_types.some(pt => product_type.includes(pt));
+            const scopeOk = !panel.scope || panel.scope !== "moderators" || window.isModerator === 1;
```
alexgarel: I was imagining 3 scopes for now:
* moderators
* users --> logged in users
* public --> everybody

html/js/external-knowledge-panels.js
```diff
@@ -0,0 +1,272 @@
...
+            return matchCat && matchCountry && matchLang && matchType && scopeOk && getExternalKnowledgePanelsOptin(section.sectionId, panel.id);
+        });
+        if (!visiblePanels.length) return;
+        const sectionDiv = document.createElement("section");
```
alexgarel: nitpik: putting all the html creation in a sub function would help reading IMO.

templates/web/pages/product/product_page.tt.html
```diff
@@ -1,5 +1,19 @@
 <!-- start templates/[% template.name %] -->
 <script src="[% static_subdomain %]/js/dist/JsBarcode.all.min.js"></script>
+
+<script>
+  window.productData = {
```
alexgarel:
```suggestion
  // this is used to filter external knowledge panels
  window.productData = {
```

lib/ProductOpener/Display.pm
```diff
@@ -8568,6 +8568,18 @@ JS
...
+	my $is_moderator = is_admin_user($User_id) || $User{moderator};
+	$template_data_ref->{is_moderator} = $is_moderator ? 1 : 0;
```
alexgarel: Add a isUser, when $User is defined (`= defined $User`)

alexgarel:
@IWouldShiverTheWholeNightThrough 

I tested it.
First of all it's very good and really what I had in mind.

Some important points (I'm leaving minor points for now and I don't repeat what is already in a comment above).
* Your CSS for knowledge panel is very different from the one of the rest of the page. Although I like the design, I don't think the team want's that, the panel should have the same design as the other panels. (though improving in some way the design of current panel could be a good idea, like your double line at bottom effect is quite nice to show what can be unfolded)
* BUG: The personal search settings is broken (and does not display) on search page (there is a `renderExternalPanelsOptinPreferences` undefined in the console logs)  
* In preferences, the description for selection of the knowledge panel should display more info (that's why we put this information in extrenla sources). eg. [org icon] "Empreinte Souffrance", "Indicateur de la quantité de souffrance animale calculée par produit." , provided by "Empreinte Souffrance" (with link to website)
* It would be important to have on each panel an "external" badge, maybe next to the title that may link to a page explaining what it is. I think this should be adde to the knowlegde panel component (maybe with a badge + badqe_url property).

But again, great job !

templates/web/pages/product/product_page.tt.html
```diff
@@ -1,5 +1,19 @@
...
+    code: "[% code %]",
+    category: [% category %],
+    country: "[% cc %]",
+    language: "[% lc %]",
+    product_type: "[% product_type %]"
```
hangy: The strings should probably be HTML encoded, just to be sure

templates/web/pages/product/product_page.tt.html
hangy: I don't think we necessarily need this data on the global `window` object. Passing the values as parameters like in this example could be better separation https://github.com/hangy/openfoodfacts-server/blob/25cf617792b1acc96c128c9c9b4cf4b12e89d32d/templates/web/pages/tags_map/map_of_tags.tt.html#L5-L11

html/js/external-knowledge-panels.js
```diff
@@ -0,0 +1,272 @@
...
+ */
+function syncNavbarExternalSections(visibleSectionsOrdered) {
+    const navbar = document.querySelector("#navbar ul.inline-list");
+    navbar.querySelectorAll("a[href^='#external_section_']").forEach(link => link.parentElement.remove());
+    const after = navbar.querySelector('[href="#match"]')?.parentElement;
+    let insertAfter = after;
+    visibleSectionsOrdered.forEach(({ sectionId, label }) => {
+        const li = document.createElement("li");
+        li.className = "product-section-button";
+        li.innerHTML = `<a class="nav-link scrollto button small round white-button" href="#external_section_${sectionId}"><span>${label}</span></a>`;
```
hangy: `sectionId` and `label` should probably be encoded

html/js/external-knowledge-panels.js
```diff
@@ -0,0 +1,272 @@
...
+/**
+ * Renders the opt-in preference UI for all external knowledge panels, grouped by section.
+ */
+function renderExternalPanelsOptinPreferences(container) {
+    if (!allPanelsBySection.length) {
+        loadPanelsMapping().then(() => renderExternalPanelsOptinPreferences(container));
+        return;
+    }
+    let html = '<div class="card" style="background:#fff;margin-top:2em;margin-bottom:2em;padding:2em 2em 1em 2em;">';
+    html += `<h2 style="margin-bottom:1em;">External Knowledge Panels (BETA)</h2>`;
```
hangy: The title should be translated using `lang`

html/js/product-preferences.js
```diff
@@ -468,8 +466,10 @@ function display_user_product_preferences(target_selected, target_selection_form
 		);
 
     $('.edit_button.close_food_preferences:last', target_selection_form).before('<div id="external_panels_prefs"></div>');
-    renderExternalPanelsOptinPreferences($(target_selection_form).find('#external_panels_prefs')[0]);
-
+    const prefEl = $(target_selection_form).find('#external_panels_prefs')[0];
+    if (prefEl && typeof window.renderExternalPanelsOptinPreferences === "function") {
+      window.renderExternalPanelsOptinPreferences(prefEl);
+    }
```
alexgarel, line 58: This does not work because we introduce renderExternalPanelsOptinPreferences later.

Also speaking with @teolemon he told be it would be better if it was a separate button (and cleaner at implementation level).
So I propose that we add it as a separate button, at the same level of preferences but on the bottom right corner.


image

alexgarel:
@IWouldShiverTheWholeNightThrough I have made a commit to:
* make it really optin (if we don't know, it's false)
* a mini refactor to put things in an include (because at first I though I had to put it in different templates)
* more important: add a /api/v3/external_sources API to have translated items

Note: there is now a section_title that is computed you should use it (I didn't have the time to make that change).

Still TODO (would do):
* add tests for my new service
* add this API to OpenAPI (instead of direct access to external-sources.json)

html/js/product-preferences.js
alexgarel: @IWouldShiverTheWholeNightThrough I let you address that.

alexgarel:
@IWouldShiverTheWholeNightThrough I added test + OpenAPI for external_sources API.

html/resources/files/external-sources.json
alexgarel, line 1: TODO @alexgarel : we have an API so we should move it to a private space.

lib/ProductOpener/APIProductServices.pm
```diff
@@ -382,4 +385,55 @@ sub product_services_api ($request_ref) {
...
 
+my %external_sources_cache = ();
+
+=head2 external_sources
+
+Get external sources but translated in target language
+
+=cut
+
+sub external_sources_api ($request_ref) {
```
alexgarel, line 407: TODO @alexgarel : add a field to tell if it applies to current user based on scope. (scope_applies: bool)

alexgarel:
@littlecoffe8 too cool to see you committing as well :hugs:

html/resources/files/external-sources.json
alexgarel: I close because this is not important.

lib/ProductOpener/APIProductServices.pm
alexgarel: done

alexgarel:
@IWouldShiverTheWholeNightThrough I had a lot of ESLint to fix (it was a bit painful, but now it's done).

alexgarel:
Thanks @IWouldShiverTheWholeNightThrough for this great work !

PR #12239, opened 2025-08-08
feat: update new nutrition schema

Vic142:
In this PR the new nutrition schema is updated based on the changes on [2025 - API - Nutrition Facts Schema and API](https://docs.google.com/document/d/19ZRrlWJraJm61E6U7AwxQ1uubPDvmSuNfl9F1oLC0Tg). The main changes are new names for the nutrient sets, simplification of the generated set object and quality check field addition. 

The OpenAPI documentation and the new features based on the new schema are updated in consequence.


docs/api/ref/schemas/product_nutrition_v3.yaml
```diff
@@ -77,3 +105,12 @@ components:
...
                         example: ["fibers"]
+            quality_checks:
+              type: object
+              description: |
+                Fields useful for quality checks.
+              properties:
+                energy_kcal_value_computed_from_other_nutrients:
+                  type: object
+                  description: |
+                    The energy in kcal computer from the other nutrients of the aggregated set.
```
stephanegigandet:
```suggestion
                    The energy in kcal computed from the other nutrients of the aggregated set.
```

stephanegigandet: I think those checks should be at the level of each nutrient set (aggregated + source sets). Also there is also the same check for enery in kJ.

lib/ProductOpener/Nutrition.pm
```diff
@@ -139,73 +148,78 @@ sub sort_sets_by_priority (@nutrient_sets) {
...
 		if (    defined $nutrient_set_ref->{preparation}
-			and $nutrient_set_ref->{preparation} eq $nutrient_set_preferred_ref->{preparation}
+			and $nutrient_set_ref->{preparation} eq $aggregated_nutrient_set_ref->{preparation}
 			and exists $nutrient_set_ref->{nutrients}
 			and ref $nutrient_set_ref->{nutrients} eq 'HASH')
 		{
 			foreach my $nutrient (keys %{$nutrient_set_ref->{nutrients}}) {
-				# for each nutrient, set its values if values are not already present in preferred set
+				# for each nutrient, set its values if values are not already present in aggregated set
 				# (ie if nutrient not present in other set with highest priority)
```
stephanegigandet:
```suggestion
				# (ie if nutrient not present in other set with higher priority)
```

lib/ProductOpener/Units.pm
```diff
@@ -322,12 +323,36 @@ Returns undef if no unit was detected.
...
+
+Unit to convert.
+
+=head3 Return values
+
+Standard unit corresponding to the given unit.
+
+=cut
+
+sub convert_to_standard_unit ($unit) {
```
stephanegigandet: Maybe rename to get_standard_unit? Just from the name convert_to_standard_unit(), it sounds like we are converting a numeric value.

PR #12246, opened 2025-08-12
docs: explanation on javascript usage and strategy

teolemon:
to explain the current state of affairs, the long term vision, and provide clarity on why we use nodejs (and how safe it is to bump it)


VaiTon (approved):

docs/dev/ref-javascript.md
```diff
@@ -0,0 +1,36 @@
+## JavaScript Usage and Strategy
```
alexgarel: @teolemon this is not a "ref", this is more a "explain" file.

And you should also start the title with the page type:
```suggestion
# Explain JavaScript Usage and Strategy
```

Also I think it's better to use h1 for main page title, and h2, etc.
```

alexgarel: @teolemon I also renamed the file as it is important because the order of the right menu depends on it. (see [docs/dev/.pages](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/dev/.pages))

PR #12251, opened 2025-08-14
fix: changes in remove_bad_nearly_empty_products_wo_images

alexgarel:
I get this change from current production.

I kept it there, but I was wondering if it's to be kept or not.

cc: @stephanegigandet

### DIFF
--- a/scripts/remove_bad_nearly_empty_products_wo_images.pl
+++ b/scripts/remove_bad_nearly_empty_products_wo_images.pl
@@ -121,7 +121,7 @@ while (my $product_ref = $cursor->next) {
 		$err .= ($creator ne 'usda-ndb-import') ? "" : "u";    # exclude USDA imports
 		$err .= ($creator !~ /^org-.*/) ? "" : "p";    # exclude orgs like "org-database-usda"
 		$err .= ($creator ne '') ? "" : "v";    #exclude products with no creator (bugs?)
-		$err .= ($completeness < 0.4) ? "" : "c";    # exclude products which have a certain level of completeness
+		$err .= ($completeness < 0.3) ? "" : "c";    # exclude products which have a certain level of completeness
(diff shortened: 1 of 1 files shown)


CharlesNepote:
Still struggling with this and I would need you advises: @stephanegigandet and @alexgarel.

The script is working like this:
1. At first, `./remove_bad_nearly_empty_products_wo_images.pl --test` is identifying products which have to be deleted, pushing the results to STDOUT
2. Then `./remove_bad_nearly_empty_products_wo_images.pl --misc` is tagging each product to be deleted.
3. Then reviewers can have a look to products that can be deleted, and validate the removal of all products.
4. The removal is finally made with `./remove_bad_nearly_empty_products_wo_images.pl --remove`

My concern is about stage 3: if reviewers identify products that should not be deleted, they should remove the tag.
At first I was using the tag `/facets/misc/en:bad-product-wo-image-to-be-deleted` BUT reviewers can't edit the `misc` field.
So in this current test, I'm using `facets/categories/en:bad-product-wo-image-to-be-deleted`:
* the reviewers can manually remove the category
* BUT adding a category increases the completeness (is it by 0.1 @stephanegigandet?)! Then many products which were supposed to be deleted, are not deleted any more due the `completeness >= 0.3`

Any advise?

This process is important (test => tag => review => remove) because we could use it to allow more mass modifications.
It allows reviewers to confirm that a mass modification is ok, before it happens.

stephanegigandet:
> My concern is about stage 3: if reviewers identify products that should not be deleted, they should remove the tag. At first I was using the tag `/facets/misc/en:bad-product-wo-image-to-be-deleted` BUT reviewers can't edit the `misc` field. So in this current test, I'm using `facets/categories/en:bad-product-wo-image-to-be-deleted`:

@CharlesNepote sorry for the late reply,.

In fact any edit will recreate the misc tags and remove the one you added with your script. So you could just tell editors to resave the product.

Issue #12256, opened 2025-08-19
Create a new `ocr_ready` Redis Stream to notify of OCR result creation


raphael0202:
- See discussion https://github.com/openfoodfacts/openfoodfacts-server/discussions/12125 for more background.

This event should contain:

- `image_id`: the ID of the image (ex: `1`)
- `product_type`: the type of the product (ex: `beauty`, `food`,...)
- `code`: the barcode of the product
- `timestamp`: timestamp of the event

raphael0202:
Done!

PR #12257, opened 2025-08-19
feat: add an `ocr_ready` Redis Stream to notify of OCR result creation

raphael0202:
Solves #12256.

Once merged, I will add documentation about Redis events in a new PR.

### DIFF
--- a/lib/ProductOpener/Redis.pm
+++ b/lib/ProductOpener/Redis.pm
@@ -427,9 +427,13 @@ The product code associated with the image.
 
 The ID of the image.
 
+=head4 String $json_url
+
+The URL where the OCR result JSON file can be found.
+
 =cut
 
-sub push_ocr_ready_to_redis ($code, $image_id, $timestamp = time()) {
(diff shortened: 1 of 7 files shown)


lib/ProductOpener/Redis.pm
```diff
@@ -411,6 +412,93 @@ sub push_product_update_to_redis ($product_ref, $change_ref, $action) {
...
+				'*',
+				# fields
+				'timestamp',
+				$timestamp,
+				'code',
+				Encode::encode_utf8($code),
+				'image_id',
+				Encode::encode_utf8($image_id),
+				'product_type',
+				$options{product_type},
```
alexgarel: Could we also put the URL of the JSON for sake of simplicity (and possible future evolution). It's quite easy to compute here, and avoid recomputing it on the other side.

alexgarel (approved):
Cool !

I have a suggestion to put the URL in the payload.

lib/ProductOpener/Redis.pm
raphael0202: I added it to the event :)

PR #12274, opened 2025-08-21
fix: build and save pro platform column names mapping

stephanegigandet:
When we upload CSV / Excel files to the pro platform, we currently recompute all possible column names to OFF field mappings for the current language (+ English), which can take quite a lot of time.

Instead we can just build all languages and store the corresponding mappings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,7 +236,7 @@ reset_owner:
 	${DOCKER_COMPOSE_BUILD} run --rm --no-deps --user root backend chown www-data:www-data -R /opt/product-opener/ /mnt/podata /var/log/apache2 /var/log/httpd  || true
 	${DOCKER_COMPOSE_BUILD} run --rm --no-deps --user root frontend chown www-data:www-data -R /opt/product-opener/html/images/icons/dist /opt/product-opener/html/js/dist /opt/product-opener/html/css/dist
 
-init_backend: build_taxonomies build_lang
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -494,6 +494,17 @@ build_taxonomies_test: create_folders
 # GITHUB_TOKEN might be empty, but if it's a valid token it enables pushing taxonomies to build cache repository
 	${DOCKER_COMPOSE_TEST} run --no-deps --rm -e GITHUB_TOKEN=${GITHUB_TOKEN} backend /opt/product-opener/scripts/taxonomies/build_tags_taxonomy.pl ${name}
 
+build_pro_platform: create_folders
+	$(MAKE) MOUNT_FOLDER=build-cache MOUNT_VOLUME=build_cache _bind_local
+	@echo "🥫 build pro platform"
+	${DOCKER_COMPOSE_BUILD} run --no-deps --rm backend /opt/product-opener/scripts/build_pro_platform_fields_columns_names.pl
```
alexgarel, line 500: You should also add the command to the deploy documentation: docs/dev/how-to-release.md

alexgarel (approved):
Thanks !

Just a documentation to update (important).

Makefile
stephanegigandet: I see you did it, thank you!

PR #12275, opened 2025-08-21
taxonomy: czech synonym soy

moon-rabbitOFF:
I added a czech synonym for soy: sóju; and added the stop word "ostatní": other.


stephanegigandet (approved):
Thanks

PR #12277, opened 2025-08-23
feat: adds folksonomy-properties-webcomponent

kirtanchandak:
### What
-  adds folksonomy-properties-webcomponent
### Screenshot
image


alexgarel (approved):
@kirtanchandak Good work.

That said, I had to change the place for off-webcomponents-configuration, otherwise it didn't apply to /properties pages and we where fetching properties from the wrong API.

PR #12279, opened 2025-08-24
feat: Make Keycloak the master source of user data at implementation level 2  - M1

john-gom:

### What

Any user PII (like ful name, email) should be fetched from Keycloak and not duplicated in Product Opener.

Data that Product Opener stores about users is referred to as user_preferences and user functions have been renamed to reflect this. A full "user" object contains the user PII plus preferences whereas user_preferences only holds the data that is stored by Product Opener

Also introduces caching of user data stored in Keycloak so we don't have to call out to Keycloak every time to fetch it.


john-gom:
Still to be discussed:

When Keycloak becomes the master source of truth for PII we will need to fetch things like the user full name from Keycloak to display this on the top right account menu, but I'm not sure we really want to do this every time a page is displayed. Here are a few options I'm thinking about:
1. Fetch the data from Keycloak every time.
 - Good: Data is always up to date
 - Bad: Additional workload and delay on every page request
2. Always keep a copy of the data in the Product Opener user.sto file
 - Good: No need to call Keycloak to get data
 - Bad: We would be replicating PII outside of Keycloak
 - Bad: Need to send events from Keycloak every time user data is edited so that PO can replicate any changes
3. Fetch the user data from the access token stored in the user session
 - Good: No additional call to Keycloak
 - Bad: We are still storing PII in the user.sto file, albeit in a more opaque format
 - Bad: User data won't be refreshed on screen if it is changed in Keycloak
4. Remove the access token from the user.sto file and just store a Keycloak session id. Add any PII we need for the user to the session cookie
 - Good: No PII stored in the user.sto file
 - Bad: Session cookie will get bigger
 - Bad: User data won't be refreshed on screen if it is changed in Keycloak until the session expires and is refreshed
 - Bad: Need to make sure any PII we need access to in the HTML templates is copied into the session cookie

Code is currently implementing option 1

hangy:
> When Keycloak becomes the master source of truth for PII we will need to fetch things like the user full name from Keycloak to display this on the top right account menu, but I'm not sure we really want to do this every time a page is displayed. Here are a few options I'm thinking about:
> 
> 1. Fetch the data from Keycloak every time.

Considering this is 1 or 2 HTTP calls *just* to display the name, that sounds kinda expensive. We also don't know how well our Keycloak instance would handle it in production, so it might also be a gamble to run in production workload (number of HTTP calls to OFF itself)

Potential 5th option: Return the user info once, and store it in the browser's `localStore`.

- Good: No PII stored in the user.sto file
- Good: Session cookie size unaffected
- Bad: User data won't be refreshed on screen if it is changed in Keycloak until the session expires and is refreshed (or refreshed manually)
- Bad: Need to make sure the `localStore` is kept in sync with the session cookie (ie. delete on sign-out)

john-gom:
> Potential 5th option: Return the user info once, and store it in the browser's `localStore`.
> 
> * Good: No PII stored in the user.sto file
> * Good: Session cookie size unaffected
> * Bad: User data won't be refreshed on screen if it is changed in Keycloak until the session expires and is refreshed (or refreshed manually)
> * Bad: Need to make sure the `localStore` is kept in sync with the session cookie (ie. delete on sign-out)

Additional bad point for this option is that all of the current templates that include user properties would need to be interpolated client side, which could be a lot of work

hangy:
> > Potential 5th option: Return the user info once, and store it in the browser's `localStore`.
> > 
> > * Good: No PII stored in the user.sto file
> > * Good: Session cookie size unaffected
> > * Bad: User data won't be refreshed on screen if it is changed in Keycloak until the session expires and is refreshed (or refreshed manually)
> > * Bad: Need to make sure the `localStore` is kept in sync with the session cookie (ie. delete on sign-out)
> 
> Additional bad point for this option is that all of the current templates that include user properties would need to be interpolated client side, which could be a lot of work

tbh, it doesn't look like there are a lot of places where the name is used (in the HTML template)
- [`"user.name"`](https://github.com/search?q=repo%3Aopenfoodfacts%2Fopenfoodfacts-server%20%22user.name%22&type=code): 3 files, only one actual HTML template match, 2 notification mails
- [`"->{email}"`](https://github.com/search?q=repo%3Aopenfoodfacts%2Fopenfoodfacts-server+%22-%3E%7Bemail%7D%22&type=code): 18 files (few of them in code that should exist after Keycloak migration)
- [`"->{name}"`](https://github.com/search?q=repo%3Aopenfoodfacts%2Fopenfoodfacts-server+%22-%3E%7Bname%7D%22&type=code): 33 files (lots of false-positives, few of them in code that should exist after Keycloak migration)

lib/ProductOpener/Keycloak.pm
```diff
@@ -276,15 +293,23 @@ Try to find a user in Keycloak by their username.
...
-A hashmap reference with user information from Keycloak.
+A hashmap reference with user information from Keycloak mapped onto PO user fields
 
 =cut
 
-sub find_user_by_username ($self, $username, $brief = 0) {
+sub find_user_by_username ($self, $userid) {
+	my $user_ref = $memd->get("user/$userid");
+	if (not defined $user_ref) {
+		my $user_ref = keycloak_to_user_ref($self->find_keycloak_user_by_username($userid));
```
stephanegigandet: I think it should be

```suggestion
		$user_ref = keycloak_to_user_ref($self->find_keycloak_user_by_username($userid));
```

lib/ProductOpener/Tags.pm
```diff
@@ -2698,6 +2699,19 @@ sub country_to_cc ($country) {
 	return;
 }
 
+sub cc_to_country($cc) {
```
stephanegigandet, line 2702: If called after init_countries(), then we can use the hash $country_codes{$cc} directly

hangy (approved):

lib/ProductOpener/APITest.pm
```diff
@@ -158,22 +160,24 @@ sub wait_server() {
...
 
 Wait for server, dynamic front, and authentication server to be ready.
 Run this at the beginning of every integration test
 
 =cut
 
-sub wait_application_ready() {
+sub wait_application_ready($file) {
+	# write_minion_log("Starting test $file");
+
```
alexgarel:
```suggestion
sub wait_application_ready($file) {
```

lib/ProductOpener/Users.pm
```diff
@@ -1146,55 +1136,95 @@ sub open_user_session ($user_ref, $refresh_token, $refresh_expires_at, $access_t
 	return $user_session;
 }
 
-sub retrieve_user ($user_id) {
+# This just fetches the local data about the user, e.g. sessions. It should not be used if you need access to data
+# that comes from Keycloak, like email, name, locale and country
+sub retrieve_user_preferences ($user_id) {
```
alexgarel, line 1141: Thank for renaming it !

lib/ProductOpener/Minion.pm
```diff
@@ -85,7 +87,22 @@ sub get_minion() {
...
+	my $job_id = get_minion()->enqueue(@_);
+
+	# Can uncomment this for debugging during integration testing but need to comment out again for normal use
+	# my $job = get_minion()->job($job_id);
+	# write_minion_log("Job $job_id for " . $job->task . " created at " . localtime($create_time) . " has created time of " . localtime($job->info->{created}));
+
+	return $job_id;
 }
 
+sub write_minion_log($message) {
```
alexgarel, line 100: thanks for this small but useful refactor

alexgarel (approved):
Thanks a lot for yet another big PR :-)

PR #12292, opened 2025-08-26
fix: block municorn-calorie-counter-app

alexgarel:
and fix edit rules implementation for simple fields ignore


teolemon (approved):
TY. It would make sense longer term to break down config_off into parts for NOVA, Edit rules…

davidpryor:
> TY. It would make sense longer term to break down config_off into parts for NOVA, Edit rules…

Hello! I have almost 0 experience with Perl, so this may be a simple question.

I was just wondering why these edit rules are built into the Perl files rather than being a config file that could be pushed to the production servers?

PR #12310, opened 2025-08-29
refactor: Nutrition data panel

Vic142:
The changes of the new products nutrition schema break some features in the project.

This PR aims at refactoring the display of the nutrition facts table in order for it to work with the new nutrition schema of the products.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -70,6 +70,7 @@ BEGIN {
 		&compute_stats_for_products
 		&compare_product_nutrition_facts_to_categories
 		&data_to_display_nutrition_table
+		&data_to_display_nutrition_table2
 		&display_nutrition_table
 		&display_product
 		&display_product_api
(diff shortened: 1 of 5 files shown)


lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -1279,8 +1279,8 @@ sub create_nutrition_facts_table_panel ($product_ref, $target_lc, $target_cc, $o
 
 		# Compare the product nutrition facts to the most specific category
 		my $comparisons_ref = compare_product_nutrition_facts_to_categories($product_ref, $target_cc, 1);
-		my $panel_data_ref = data_to_display_nutrition_table($product_ref, $comparisons_ref, $request_ref);
-
+		#my $panel_data_ref = data_to_display_nutrition_table($product_ref, $comparisons_ref, $request_ref);
+		my $panel_data_ref = data_to_display_nutrition_table2($product_ref, $comparisons_ref, $request_ref);
```
stephanegigandet: The new function now has the same name as the old function I think: 

```suggestion
		my $panel_data_ref = data_to_display_nutrition_table($product_ref, $comparisons_ref, $request_ref);
```

lib/ProductOpener/Display.pm
```diff
@@ -9991,208 +9835,95 @@ CSS
...
-						if (    (defined $product_ref->{nutriments}{$nid . $prepared . "_modifier"})
-							and ($product_ref->{nutriments}{$nid . $prepared . "_modifier"} eq '-'))
-						{
-							# The nutrient is not indicated on the package, display a minus sign
-							$value_unit = '-';
-						}
-						else {
-							$value_unit = '?';
-						}
+					if (!defined $product_ref->{nutrition}{aggregated_set}{nutrients}{$nid}) {
```
stephanegigandet, line 9853: When the nutrient is not indicated on the package, it is in the unspecified_nutrients array, but in the aggregated set we may have a value from the estimation of nutrients from ingredients. We need to check if the aggregated_set gets the unspecified_ingredients array when it is computed, if not we probably should add it.

lib/ProductOpener/Display.pm
```diff
@@ -9991,208 +9835,95 @@ CSS
...
+							# if petfood then display for 1kg if not percentage
 							if (   (defined $product_ref->{product_type})
 								&& ($product_ref->{product_type} eq "petfood")
 								&& ($unit ne "%"))
 							{
-								$value = $decf->format(
-									g_to_unit($product_ref->{nutriments}{$nid . "_" . $col_id} * 10, $unit));
+								$value
+									= $decf->format(
+									g_to_unit($product_ref->{nutrition}{aggregated_set}{nutrients}{$nid}{value}, $unit)
```
stephanegigandet, line 9876: So now for pet food, we will store the input sets (and aggregated set) per_1kg, we won't need the conversion.

PR #12311, opened 2025-08-29
ci: verify dist volumes are removed on deployment

alexgarel:
Because a running container, can prevent them from being removed, as it happened for a few weeks… making js / css etc not up to date on .net…

Also change deprecated script_stop to continue-on-error in deploy workflow

### DIFF
--- a/.github/workflows/container-deploy.yml
+++ b/.github/workflows/container-deploy.yml
@@ -273,7 +273,7 @@ jobs:
         script: |
           cd ${{ matrix.env }} && \
           make hdown && \
-          # verify dist volumes where removed
+          # verify dist volumes were removed
           ! (docker volume list |grep po_.*_dist) && \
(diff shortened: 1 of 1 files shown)


teolemon (approved):

.github/workflows/container-deploy.yml
```diff
@@ -269,10 +269,12 @@ jobs:
         proxy_username: ${{ env.SSH_USERNAME }}
         proxy_key: ${{ secrets.SSH_PRIVATE_KEY }}
         command_timeout: 15m
-        script_stop: false
+        continue-on-error: false
         script: |
           cd ${{ matrix.env }} && \
           make hdown && \
+          # verify dist volumes where removed
```
alexgarel:
```suggestion
          # verify dist volumes were removed
```

PR #12162, opened 2025-07-21
feat: M4 - add new attribute unwanted_ingredients

RayanGoubet:
### What

Introduce a new attribute to know wich ingredients we don't want in our product and make a score with that
For exemple if I don't want garlic and the product have it, the score is 0 and if it doesn't the score is 100 (true false logic)  
There no translation yet, so it's hard coded in french 

### Screenshot
Product with an ingredient we don't want
image

PR #12321, opened 2025-09-01
feat: unwanted ingredients attribute for personal search - WIP

stephanegigandet:
Continuation of https://github.com/openfoodfacts/openfoodfacts-server/pull/12162 by @RayanGoubet


teolemon:
If we can take advantage of this PR to adopt the new Attributes icons on the Web: https://github.com/openfoodfacts/smooth-app/tree/develop/packages/smooth_app/assets/fonts/food_icons

html/js/product-preferences.js
```diff
@@ -168,6 +170,176 @@ function display_use_preferences_switch_and_edit_preferences_button(target_selec
...
+    }
+    document.cookie = name + "=" + (encodeURIComponent(value) || "")  + expires + "; path=/";
+}
+
+// callback function when the unwanted ingredients input field is changed
+function unwanted_ingredients_change_callback(e) {
+    var values_json = e.tagifyValue;
+    // If tagifyValue is empty and value is not an array and is a non-empty string, it is an initial call to initialize tagify, ignore it
+    if ((!values_json || values_json.length == 0) && (!Array.isArray(e.value) && e.value && e.value.length > 0)) {
+        console.log("Unwanted ingredients changed, empty value, ignoring");
```
alexgarel: It's better to remove or comment this log.
```suggestion
        // console.log("Unwanted ingredients changed, empty value, ignoring");
```

alexgarel: You have to do this for other logs

lib/ProductOpener/Attributes.pm
```diff
@@ -1746,6 +1741,16 @@ sub compute_attributes ($product_ref, $target_lc, $target_cc, $options_ref) {
 			$attribute_ref = compute_attribute_ingredients_analysis($product_ref, $target_lc, $analysis);
 			add_attribute_to_group($product_ref, $target_lc, "ingredients_analysis", $attribute_ref);
 		}
+		# Unwanted ingredients
+		if (defined cookie("attribute_unwanted_ingredients_tags")) {
+			my @unwanted_ingredients = map {s/^\s+|\s+$//gr} split /,/, cookie("attribute_unwanted_ingredients_tags");
```
alexgarel, line 1746: I really find it a pity that we call "cookie" in such deep places in the code. It makes it harder to unit test or to re-use the code.

For me we should first translate HTTP thing into a meaningful map / object (like we do for request_ref) then use it.

I don't say that to change it now, because it already happens all over the place in this file, but just has what we should better tend to do… isolate HTTP concern and have a code that does not depends on HTTP thing (even less on Apache2 stuff)

alexgarel: that said, here at least, `compute_attribute_unwanted_ingredients` is well separated :+1:

po/common/common.pot
```diff
@@ -7615,3 +7615,23 @@ msgstr "Quantity of added sugars"
 msgctxt "search_products_in_same_category_without_added_sugars"
 msgid "Search for products in the same category without added sugars:"
 msgstr "Search for products in the same category without added sugars:"
+
+msgctxt "attribute_unwanted_ingredients_name"
+msgid "Unwanted ingredients"
+msgstr "Unwanted ingredients"
```
alexgarel, line 7621: I think we really need a note as a warning here that this settings might not be reliable, because we only check tags, that means if we didn't parse ingredients list correctly or miss synonyms, we might give a wrong indication.

```suggestion
msgctxt "attribute_unwanted_ingredients_name"
msgid "Unwanted ingredients"
msgstr "Unwanted ingredients"

msgctxt "attribute_unwanted_ingredients_note"
msgid "This preference is based on Open Food Facts understanding of ingredients list and might not be accurate, always verify the product by yourself."
msgstr "This preference is based on Open Food Facts understanding of ingredients list and might not be accurate, always verify the product by yourself."

```

lib/ProductOpener/Attributes.pm
```diff
@@ -1923,4 +1928,102 @@ sub compute_attribute_repairability_index_france ($product_ref, $target_lc, $tar
...
+			$attribute_ref->{description}
+				= lang_in_other_lc($target_lc, "attribute_unwanted_ingredients_name")
+				. lang_in_other_lc($target_lc, "sep") . ": "
+				. $unwanted_ingredients;
+			$attribute_ref->{icon_url} = "$static_subdomain/images/attributes/dist/contains-unwanted-ingredients.svg";
+		}
+		else {
+			# No unwanted ingredients detected
+
+			# Check that we do not have too many unknown ingredients, otherwise mark as unknown
```
alexgarel, line 2004: That's cool that we do this. Do we have some kind of justification for using 10% of unknown ingredients ?

stephanegigandet: Not really, we could adjust it. I didn't have time to research the stats on this, but it sounded reasonable to start with.

html/js/product-preferences.js
stephanegigandet: ok, removing the logs

lib/ProductOpener/Attributes.pm
stephanegigandet: Agreed, I will move the cookies somewhere else in a new PR

PR #12322, opened 2025-09-01
ci: Add action to include JavaScript issues in JS project

teolemon:
### What
- ci: Add action to include JavaScript issues in JS opportunities project


teolemon:
thx @hangy

PR #12332, opened 2025-09-03
fix: use added sugars value from nutrition facts in knowledge panel

stephanegigandet:
We currently always use the estimated added sugar percent from ingredient analysis for the Added sugars knowledge panel.
For some products, we can have added sugars specified in the nutrition facts, so we now use those by priority.

Before:

image

After:

image


stephanegigandet:
Fixes #12334

PR #12335, opened 2025-09-03
fix(openapi): Required fields in product_base.yml

VaiTon:
### What

Add required fields in product_base.yml to fix automatic type generation for the JS SDK.

### DIFF
--- a/docs/api/ref/schemas/product_base.yaml
+++ b/docs/api/ref/schemas/product_base.yaml
@@ -1,6 +1,9 @@
 type: object
 description: |
   Base product data
+required:
+  - code
+  - schema_version
 properties:
   abbreviated_product_name:
     type: string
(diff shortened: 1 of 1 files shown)


docs/api/ref/schemas/product_base.yaml
```diff
@@ -1,6 +1,23 @@
...
+  - nova_group
+  - nova_groups
+  - obsolete
+  - obsolete_since_date
+  - product_name
+  - product_name_en
+  - schema_version
+  - product_quantity
+  - product_quantity_unit
+  - quantity
```
alexgarel: Why would product_name_en be required ?
Also I don't think we need code and id.
nova_group should not be required, neither obsolete and obsolete_since_date.

Even lc and cc can be deduced from domain.

For me only required field are:
- code
- (and maybe) schema_version

stephanegigandet: Right, the only field that is guaranteed to always be there is "code". You will also always get product_schema on API v3.2+ 

SDKs should not fail if some of the other fields are not present (even if some of them might be generated for all product now, like codes_tags).

VaiTon:
@alexgarel @stephanegigandet changes applied. Thank you for the review.

PR #12352, opened 2025-09-08
feat: adjusted file-structure and added fumadocs sync workflow

JagjeevanAK:
Signed-off-by: Jagjeevan Kashid <jagjeevandev97@gmail.com>
### What

For doc generation I have added a workflow and created well organized dir structure for it.


JagjeevanAK:
Hey @alexgarel  look into this PR it's just restructuring of some of the docs files and workflow addittion.

alexgarel (commented):
By refactoring this way, we loose a bit the Diataxis structure.
Also I personally was ok with a more flat structure, because it's always difficult then to decide which docs goes where.

For example here you put uploading photo, in advanced feature, while it's absolutely not so… photo upload is the first thing we want from our re-users. 

So I would say, let's stick to the [Diataxis](https://diataxis.fr/) framework with this two fold: DEV and API (because it's absolutely not the same public), and inside those two folders have subfolders for tutorials / how-to / explanation / references. If there are many file in these subfolders we might split them in section (like tutorials), but we should decide together, for example it would be more easy to have a read and write section.

alexgarel:
Also an important point is that if we remove .pages and so on:
* we must remove the actual workflow that builds the documentation and push it to pages 
* we have to put redirections in place in current documentation.

For the redirections, the easiest would be to replace current pages in the gh-pages branch by html pages with a [http-equiv element using refresh](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/http-equiv#refresh) that redirects to the corresponding new documentation. It can be a one time action that we do directly on the gh-pages branch.

.github/workflows/sync-docs-to-fumadocs.yml
```diff
@@ -0,0 +1,191 @@
...
+on:
+  push:
+    branches: [ "main" ]
+    paths:
+      - "docs/**"
+      - "!docs/api/**"  
+  pull_request:
+    paths:
+      - "docs/**"
+      - "!docs/api/**"
```
alexgarel, line 12:
```suggestion
      # this is the OpenAPI documentation
      # which is handled differently
      - "!docs/api/**"
```

.github/workflows/sync-docs-to-fumadocs.yml
```diff
@@ -0,0 +1,191 @@
...
+  workflow_dispatch:
+
+
+jobs:
+  sync-docs:
+    name: Convert MD to MDX and Sync to Documentation Repository
+    runs-on: ubuntu-latest
+    # Only run on push to main, not on PRs (PRs will just validate the workflow)
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
alexgarel, line 21: No one is checking that actions on main are not failing.

So I would say that this conditions should be set on some of the steps below but not on full workflow, because otherwise we will have the documentation not updating during several weeks without noticing, and also the one breaking the workflow won't notice it.

So we must run the workflow on every PR. The only step that should be skipped is the "Check for changes and commit" step.

.github/workflows/sync-docs-to-fumadocs.yml
```diff
@@ -0,0 +1,191 @@
...
+          cd openfoodfacts-server
+          
+          # First, validate the markdown files
+          echo "Validating markdown files..."
+          npx fumadocs-transpiler ./docs --validate-only
+          
+          # Use fumadocs-transpiler to convert entire docs directory
+          # This will convert all .md files to .mdx format with proper component imports
+          echo "Converting markdown files to MDX..."
+          npx fumadocs-transpiler ./docs ./temp-mdx --verbose
```
alexgarel, line 70: Maybe we could use the "--verbose" only if we have ACTIONS_RUNNER_DEBUG set to true

.github/workflows/sync-docs-to-fumadocs.yml
```diff
@@ -0,0 +1,191 @@
...
+
+      - name: Fix MDX frontmatter and content
+        run: |
+          cd openfoodfacts-documentation/content/docs/Product-Opener
+          
+          # Process all MDX files to ensure proper frontmatter
+          find . -name "*.mdx" -type f | while read -r file; do
+            echo "Processing frontmatter for: $file"
+            
+            # Extract title from first heading if no frontmatter exists
```
alexgarel, line 143: I'm a bit surprised that the transpiler does not do that. Is there a reason ?

.github/workflows/sync-docs-to-fumadocs.yml
```diff
@@ -0,0 +1,191 @@
...
+              echo "description: \"$title documentation\"" >> "$temp_file"
+              echo "---" >> "$temp_file"
+              echo "" >> "$temp_file"
+              cat "$file" >> "$temp_file"
+              mv "$temp_file" "$file"
+            fi
+            
+            # Fix relative links to work with Fumadocs structure
+            sed -i 's|\]\(\.\/|\]\(/docs/Product-Opener/|g' "$file"
+            sed -i 's|\]\(\.\.\/|\]\(/docs/Product-Opener/|g' "$file"
```
alexgarel, line 161: How would this work in subfolders ? You are replacing a relative url with an absolute one ?

alexgarel (commented):
I reviewed the workflow, here are my comments.

.github/workflows/sync-docs-to-fumadocs.yml
JagjeevanAK: hey this workflow is only pulling .md files and not .yml files as all the files will be given to fumadocs transpiler

.github/workflows/sync-docs-to-fumadocs.yml
JagjeevanAK: okay no issues

JagjeevanAK:
hey @alexgarel I am closing this PR and opening another PR.

PR #12355, opened 2025-09-09
chore: Rearrange config so we can move production to Keycloak level 1 - M1

john-gom:

### What
The existing config did not support storing the OIDC client secrets, which have to be in Config2 in order for them to be only stored on the production servers.

Steps to enabling an instance for production are as follow:

1. Create a client id (e.g. "OFF") in Keycloak by following these instructions: https://github.com/openfoodfacts/openfoodfacts-auth?tab=readme-ov-file#internal-backend-client
2. Set the $oidc_client_id and $oidc_client_secret values in Config2.pm
3. Set the $oidc_implementation_level=1 in Config2.pm and export all of the variables
4. Make sure the redis listener service is started for each instance
5. For one instance, e.g. OFF, run the migrate_users_to_keycloak script (this will take some time)


docker/docker-entrypoint.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Link Config.pm and Config2.pm
-ln -sfT /opt/product-opener/lib/ProductOpener/Config2_docker.pm /opt/product-opener/lib/ProductOpener/Config2.pm
+ln -sfT Config2_docker.pm /opt/product-opener/lib/ProductOpener/Config2.pm
```
alexgarel, line 4: For my information, why did it need to be changed ?

alexgarel (approved):
Great !

docker/docker-entrypoint.sh
john-gom: In Dev the code is on a local mount so the absolute path for docker isn't the same as the local path, so locally you have a broken link.

PR #12361, opened 2025-09-12
taxonomy: New additive E267 - Buffered vinegar

moon-rabbitOFF:
I added a new additive accepted by the EU in 2023 https://agrinfo.eu/book-of-reports/food-additives-authorisation-of-buffered-vinegar-e-267/

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -17790,20 +17790,6 @@ fr: Vinaigre blanc distillé
 hr: bijeli destilirani ocat
 it: aceto bianco distillato
 
-< en:vinegar
-en: buffered vinegar
-hr: puferirani ocat
-it: aceto tamponato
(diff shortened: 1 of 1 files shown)


moon-rabbitOFF:
@stephanegigandet There is already an entry for "buffered vinegar" in the ingredient taxonomy. Should I delete the ingredient entry to avoid having a duplicate?

stephanegigandet:
> @stephanegigandet There is already an entry for "buffered vinegar" in the ingredient taxonomy. Should I delete the ingredient entry to avoid having a duplicate?

@moon-rabbitOFF yes, it should be in additives.txt only

moon-rabbitOFF:
There is an child ingredient "buffered vinegar powder". How should I deal with that?

aleene:
Delete? It should be catched by the processing.

moon-rabbitOFF:
Okay I deleted it

Issue #12358, opened 2025-09-12
Follow up on Unwanted Ingredients


teolemon:
### Add real names for additives
Add real names for additives
<img width="556" height="393" alt="Image" src="https://github.com/user-attachments/assets/0df6171d-7557-48c1-9c0d-949e6c7442c7" />

### Hover text in case of no detection

Even in green, be more precise about the list of ingredients we're avoiding (in that case pork) (We couldn't detect vanilla, pork and gluten in the ingredient list. Always double check yourself)
<img width="556" height="393" alt="Image" src="https://github.com/user-attachments/assets/472199f4-eb64-4190-9123-5d323e2d2155" />

### Enable on Open Beauty Facts, so that it can spark a discussion over the week-end

stephanegigandet:
For additives, it would be very difficult. We do cover the more useful case which is to show the E-number when we start to type the additive name:

<img width="285" height="217" alt="Image" src="https://github.com/user-attachments/assets/014276f0-e2ce-4b69-8aad-cbece7d325a5" />

PR #12362, opened 2025-09-12
fix: list which unwanted ingredients were not detected

stephanegigandet:
fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/12358

change message to "No unwanted ingredients detected" (instead of just "No unwanted ingredients") and explicitly list which ingredients were not detected

image


teolemon:
Do we have a server log of ingredients we did not recognize @stephanegigandet ?

teolemon:
To add more taxonomy entries

PR #12372, opened 2025-09-16
fix: deactivate OFF Days banner

stephanegigandet:
put back the donation banner


teolemon:
Tests will probably be failing since the HTML output changed. You need to update test results by running 
```make update_tests_results```, and committing updated tests.

PR #12383, opened 2025-09-17
feat: New Unwanted Ingredients product attribute + related APIs and documentation

stephanegigandet:
This PR is a refactor and documentation of the Unwanted Ingredients product attribute that was added to the OFF website before the OFF days:

- refactor to remove the reading of cookies deep in the Attributes.pm code
- new doc explain-product-attributes.md to update and replace the documentation on the wiki https://wiki.openfoodfacts.org/Product_Attributes -- In particular the doc explains how product attributes with parameters work.
- created API v3 versions of /api/v3/preferences and /api/v3.4/attribute_groups APIs : this is necessary because we do not want clients that do not handle attributes with parameters to get them listed + tests
- OpenAPI documentation for new APIs

To add support for Unwanted ingredients in the mobile app, OFF Explorer etc:

- the call to /api/v0/attribute_groups must be replaced with /api/v3.4/attribute_groups so that the unwanted_ingredients attribute is listed (the response is almost the same, with the addition of the standard v3 error reporting)
- ideally, the client should use the "parameters" returned by the attribute_groups API to ask the user for the unwanted ingredients. In practice, on the OFF site, some of it is hardcoded as we changed it before the "parameters" was added. It would be best not to hardcode it, so that any new attribute with parameters will work without changes on the client.
- unwanted ingredients should be sent as a comma separated list of canonical ingredients tags in the attribute_unwanted_ingredients_tags parameter: either as a cookie, or a URL query parameter.


docs/api/ref/api-v3.yaml
```diff
@@ -582,6 +667,146 @@ paths:
...
+                                          description: The name of the parameter.
+                                          example: "Unwanted ingredients"
+                                        tagtype:
+                                          type: string
+                                          description: The tag type of the parameter.
+                                          example: "ingredients"
+                                        type:
+                                          type: string
+                                          description: The type of the parameter. "tags" indicates a comma-separated list of canonical tags is expected.
+                                          example: "tags"
```
VaiTon: Can we make these enums?

stephanegigandet: @VaiTon yes we can, I put them as strings as it's likely there will be new types coming, but when that happens, I can bump the API version number so that it doesn't break apps.

VaiTon: I think you can even not do that, as adding values to an enum should be fairly backward compatible in my opinion, but that's on you :)

hangy: > adding values to an enum should be fairly backward compatible in my opinion

Note: New enum values can break some apps. For example, when you generate C# client code from an OpenAPI spec, the [enum](https://swagger.io/docs/specification/v3_0/data-models/enums/) will be mapped to an [enum](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/enum), where non-existing values could cause serialization issues.

VaiTon (commented):
I'm not quite sure on the structure of "parameters". Why is it an array? 

Also, shouldn't the type be specific to the entire attribute rather than the single parameter? 

In the unknown ingredients case, for example, I fail to find a meaning for the "values" field

tests/integration/expected_test_results/api_v3_taxonomy_suggestions/packaging-shapes-cc-fr-categories-en-yogurts.json
```diff
@@ -2,31 +2,7 @@
    "errors" : [],
    "status" : "success",
    "suggestions" : [
-      "Pot",
-      "Seal",
```
hangy, line 6: Any idea why this test for packaging and other tests like [categories-string-fr-tart-get-synonyms.json](https://github.com/openfoodfacts/openfoodfacts-server/pull/12383/files#diff-5a1b04492da94d05504c582bb8e52b82ec4052bd8fae7128dda08bca22689df3) for categories are affected by the change? I feel like this part of the snapshots shouldn't need to be updated for the unwanted ingredients API?

docs/api/ref/api-v3.yaml
stephanegigandet: Right, it's the same in the Dart code that is used to parse the JSON in our mobile app, returning new values will break the parsing.

stephanegigandet:
> I'm not quite sure on the structure of "parameters". Why is it an array?

It's an array so that in the future we can have attributes with more than one parameter.

e.g. we could have a "Preferred brands" attribute, with 2 lists of brands: "Brands I like" and "Brands I don't like".

> Also, shouldn't the type be specific to the entire attribute rather than the single parameter? In the unknown ingredients case, for example, I fail to find a meaning for the "values" field

```
{
  "icon_url": "http://static.openfoodfacts.localhost/images/attributes/dist/contains-unwanted-ingredients.svg",
  "id": "unwanted_ingredients",
  "name": "Unwanted ingredients",
  "parameters": [
    {
      "id": "attribute_unwanted_ingredients_tags",
      "name": "Unwanted ingredients",
      "tagtype": "ingredients",
      "type": "tags"
    }
  ],
  "setting_name": "Unwanted ingredients",
  "values": [
    "not_important",
    "important",
    "very_important",
    "mandatory"
  ]
}
```

The "type" is for the parameter: we want a list of unwanted ingredients. But we could have another parameter like "attribute_unwanted_ingredients_ignore_if_quantity_below" that could be a number, so that people could indicate that if there is for instance less than 5% of the unwanted ingredients, then they don't care.

"values" is for the attribute: in this case, users an set it to all 4 values "not_important" etc.   For some attributes like allergens, it makes more sense to make it binary: either it's "not_important" (you are not allergic to X), or "mandatory" (you are allergic to X and absolutely cannot eat it).

tests/integration/expected_test_results/api_v3_taxonomy_suggestions/packaging-shapes-cc-fr-categories-en-yogurts.json
stephanegigandet: @hangy it took me a while to understand why: it's because I changed request_param() in HTTP.pm to be able to return a value set in a cookie, but I put an empty "return;" statement instead of "return undef;" because perl_critic doesn't like return undef.

The issue is that code like this then fails in a very bad way if a parameter is not set, the keys and values get completely mixed.

```
	my $options_ref = {
		limit => request_param($request_ref, 'limit'),
		get_synonyms => request_param($request_ref, 'get_synonyms')
	};
```

Should be fixed.

docs/api/explain-product-attributes.md
```diff
@@ -0,0 +1,267 @@
...
+### List and descriptions of available product attributes
+
+Apps can get the list of all available product attributes at the `/api/v3.4/attribute_groups` endpoint for API v3,
+or at the `/api/v2/attribute_groups` or `/api/v2/attribute_groups_[language code]` endpoints for API v2.
+
+e.g. [https://world.openfoodfacts.org/api/v3.4/attribute_groups]
+
+Note that API < 3.4 does not support product attributes with parameters (such as the "Unwanted ingredients" attribute that requires a list of unwanted ingredients).
+
+It returns a JSON array of attribute groups that each contain an array of attributes.
```
alexgarel, line 63:
```suggestion
It returns a JSON array of attribute groups that each contain an array of attributes.

For a complete and accurate documentation refers to OpenAPI specification for v3.
```

I would have added a link but I don't know what the URL would be :-)

docs/api/explain-product-attributes.md
```diff
@@ -0,0 +1,267 @@
...
+                        "status": "known",
+                        "description_short": "Fair trade products help producers in developing countries.",
+                        "description": "When you buy fair trade products, producers in developing countries are paid a higher and fairer price, which helps them improve and sustain higher social and often environmental standards.",
+                        "title": "Fair trade product"
+                    }
+                ]
+            }
+        ]
+    }
+    [..]
```
alexgarel: typo ?
```suggestion
```

docs/api/explain-product-attributes.md
alexgarel, line 1: Great documentation.

I'm just not sure if it's a good idea to put the full description of fields if we have it in OpenAPI format, because will we maintain this documentation ? We could just link to OpenAPI documentation to give full description.

docs/api/ref/api-v3.yaml
```diff
@@ -582,6 +667,148 @@ paths:
         description: |
           The code and rev fields are mandatory.
       parameters: []
+  /api/v3/preferences:
+    get:
+      summary: Get List of Preference Importance Values
```
alexgarel, line 672:
```suggestion
      summary: Get List of Preference Importance Values
      description: |
        These parameters are used to compute the product preferences score.

        for an overview see [Explanation on Product Attributes](https://openfoodfacts.github.io/documentation/docs/Product-Opener/api/explain-product-attributes/)"
```

docs/api/ref/api-v3.yaml
```diff
@@ -582,6 +667,148 @@ paths:
...
+                              type: integer
+                              description: The factor associated with the preference importance (optional, not set for not_important). Indicates that the product attribute score should be multiplied by this factor when this importance is selected.
+                              example: 1
+                            minimum_match:
+                              type: integer
+                              description: The minimum match percentage required for the preference (optional, set for mandatory). Indicates that product with a lesser score for this attribute should not be considered a match.
+                              example: 20
+  /api/v3.4/attribute_groups:
+    get:
+      summary: Get List of Attribute Groups and Attributes
```
alexgarel, line 713:
```suggestion
      summary: Get List of Attribute Groups and Attributes
      description: |
        for an overview see [Explanation on Product Attributes](https://openfoodfacts.github.io/documentation/docs/Product-Opener/api/explain-product-attributes/)"
```

docs/api/explain-product-attributes.md
stephanegigandet: It's to indicate it's just an excerpt, there are more attributes. It's not at the right place though, I'll change it.

PR #12402, opened 2025-09-18
fix: HTML element ID attribute on file import page

### LINKED ISSUE #12401: bug: low severity: file import does not display record count to admin users
### What
There is a small bug in the reporting of the file import process; an intended status update is not displayed to admin users.

This is due to some script code on the page that attempts to query the HTML DOM for an element with an `id` of `#poll`:

https://github.com/openfoodfacts/openfoodfacts-server/blob/4817290a6aa4569c940672abbca205c7adb25923/templates/web/pages/import_file_process/import_file_process.tt.js#L29

...where currently the corresponding element in the HTML has an incorrect/mismatched `id` value of `span` (the name of a different type of HTML element):

https://github.com/openfoodfacts/openfoodfacts-server/blob/4817290a6aa4569c940672abbca205c7adb25923/templates/web/pages/import_file_process/import_file_process.tt.html#L9

Note: this does not prevent file upload or processing.

1. Login as an admin user
2. Use the file import to upload an import file

### Expected behavior
A record count should appear, next to the note about `shown to admins only`.

Image

### Actual behaviour
The text `shown to admins only` does appear, but no value is displayed.

### Why
- This provides improved feedback for admin users of the import process, and appears to be the intended behaviour.

See above for expected behaviour.

### Part of
N/A

### Additional context
N/A

### Type of device
- Browser: Firefox 128.14.0
- Operating System: Debian Linux (`testing` / `forky`)

N/A

### Time per product
N/A

jayaddison:
### What
The primary fix in this branch is to correct the `id` of an element on the file import page - this is commit 638d7606bb8553e9a2c93bb416d44488721bb8af.

Two smaller fixes are included:

  * Correction for a misnamed closing HTML tag.
  * Use `jQuery.text` instead of `jQuery.html` where safer and functionally equivalent.

### Screenshot
image

- Resolves #12401.


stephanegigandet (approved):
Thank you!

jayaddison:
You're welcome; thanks @stephanegigandet

PR #12403, opened 2025-09-19
feat: upgrade the product edit form and API v2 to support new nutrition schema - WIP

stephanegigandet:
See https://docs.google.com/document/d/19ZRrlWJraJm61E6U7AwxQ1uubPDvmSuNfl9F1oLC0Tg/edit?tab=t.0

This PR changes the product edit form so that it can edit the different possible nutrition input sets for 1 specific source (the "packaging" source on the public platform, or the "manufacturer" (or USDA etc.) source on the producers platform).

Checkboxes are displayed to show the different possible input set columns (some of them are hidden if we don't have data yet).

In theory the units can be different for the same nutrient in different input sets. In practice it should not be often the case, so unless we do have existing different units, only 1 global unit column is shown.

image

Otherwise we can show units by input set:

image

Compared to the current product edit form, we do not support unknown nutrients, we limit users to nutrients that exist in the taxonomy.

So the system to add lines with labels that can be typed in or selected from suggestions has been changed: there is now a button at the bottom to display additional nutrients. This has the benefit of greatly simplifying the code.

Also removed: the salt and sodium rows are now independent, we don't compute one from the other anymore.

To be fixed:
- nutrient checks (e.g. sugars <= carbohydrates) should be made to work on all columns

The new product edit form still uses the API v2 parameters. So I added new fields  like nutrition_input_sets_${preparation}_${per}_${nid}_value_string to pass values.

The same function will be used to add support for those fields in the actual API v2, so that existing apps can set values for the input sets, without switching to the API v3 (which does not handle nutrients yet).

API v2 will continue to support existing nutrition fields and "guess" in which input set they belong.

### DIFF
--- a/lib/ProductOpener/Export.pm
+++ b/lib/ProductOpener/Export.pm
@@ -103,7 +103,8 @@ use ProductOpener::Products qw/product_path/;
 use ProductOpener::Images qw/add_images_urls_to_product $valid_image_types_regexp/;
 use ProductOpener::EnvironmentalScore qw/localize_environmental_score/;
 use ProductOpener::ProductsFeatures qw(feature_enabled);
-use ProductOpener::Nutrition qw/add_nutrition_fields_from_product_to_populated_fields get_nutrition_input_sets_in_a_hash/;
+use ProductOpener::Nutrition
+	qw/add_nutrition_fields_from_product_to_populated_fields get_nutrition_input_sets_in_a_hash/;
 
 use Text::CSV;
 use Excel::Writer::XLSX;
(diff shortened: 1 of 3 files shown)


JagjeevanAK (commented):
**AI Review Summary**

                - Found hardcoded localhost URLs → replace with env vars  
                - Placeholder repo URLs need to be updated  

                Here are the suggested changes:

stephanegigandet:
replaced by https://github.com/openfoodfacts/openfoodfacts-server/pull/12475

PR #12404, opened 2025-09-19
chore: use staging folksonomy in staging

alexgarel:
We now have a folksonomy staging instance, use it in staging.


hangy (approved):

stephanegigandet (approved):
Great!

PR #12422, opened 2025-09-25
taxonomy: more Polish ingredients

Teiron:
### What

Added more translations for Polish ingredients and added new local salt for Poland (11 usages).


stephanegigandet (approved):
Thank you

PR #12423, opened 2025-09-25
taxonomy: synonym mustard czech

moon-rabbitOFF:
Added a synonym of mustard in czech based on a product containing this allergen: https://world.openfoodfacts.org/product/5051007128902/2-tortillas-mexicana-tesco
https://en.wiktionary.org/wiki/ho%C5%99%C4%8Dice


teolemon:
thanks @moon-rabbitOFF

PR #12424, opened 2025-09-26
docs: rename and update ecoscore -> green_score in AGRIBALYSE update …

### LINKED ISSUE #12376: Update Ecoscore references and syntax to Green-Score in docs/dev/how-to-update-agribalyse-ecoscore.md
### Task
- Update Ecoscore references and syntax to Green-Score in docs/dev/how-to-update-agribalyse-ecoscore.md

AnasDharar:
…guide.

Replaced Ecoscore/ecoscore mentions with Green-Score/green_score to align with the ongoing naming migration (Eco-Score -> Environmental Score -> Green-Score).

### What

The mentions of Ecoscore across the how-to-update-agribalyse-ecoscore.md is changed to Green-Score and the name of the file is renamed to how-to-update-agribalyse-green-score.md, as asked in the issue.
I assume there is a naming migration from 'ecoscore' to 'environmental_score' to 'green_score'.
The Perl modules in lib/ProductOpener/EnviromentalScore.pm, commit 8a1fec4 and external_data/environmental_score/, helped me understand the migration that took place earlier (ecoscore to environmental_score). And now, it is getting changed to green_score.
As the codebase is still in transition, some of these references
might not exactly match the current code for now. However, the goal is 
to bring the docs in line with the project’s intended terminology 
going forward.

- Fixes #12376

### DIFF
--- a/docs/dev/how-to-update-agribalyse-green-score.md
+++ b/docs/dev/how-to-update-agribalyse-green-score.md
@@ -6,11 +6,11 @@ New versions of the AGRIBALYSE database are released from time to time and this 
 
 ## Obtain and Convert the AGRIBALYSE Spreadsheet
 
-Download the AGRIBALYSE food spreadsheet from the [AGRIBALYSE](https://doc.agribalyse.fr/documentation/acces-donnees) web site (use the French site rather than English as updates on the English site may be delayed), and save it as AGRIBALYSE_vf.xlsm" in the green_score/agribalyse folder.
+Download the AGRIBALYSE food spreadsheet from the [AGRIBALYSE](https://doc.agribalyse.fr/documentation/acces-donnees) web site (use the French site rather than English as updates on the English site may be delayed), and save it as AGRIBALYSE_vf.xlsm" in the ecoscore/agribalyse folder.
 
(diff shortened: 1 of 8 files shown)


stephanegigandet:
Hi @AnasDharar 

Thanks for the PR. The Eco-Score was renamed to the Green-Score, but internally, we decided to change the field names to environmental_score instead, so that if there's a rename later, we don't have to change the database schema.

So you cannot rename all instance of eco_score to green_score, some of them are going to be environmental_score will stay that way. e.g. there is no GreenScore.pm file.

AnasDharar:
> Hi @AnasDharar
> 
> Thanks for the PR. The Eco-Score was renamed to the Green-Score, but internally, we decided to change the field names to environmental_score instead, so that if there's a rename later, we don't have to change the database schema.
> 
> So you cannot rename all instance of eco_score to green_score, some of them are going to be environmental_score will stay that way. e.g. there is no GreenScore.pm file.

Yups, so all the instances of Ecoscore in text should be Green-Score, and in syntax, it should be environmental_score? I took a look at the .pm files, there's EnvironmentalScore.pm file, and there were comments stating that ecoscore fields are removed and renamed with environmental_score

AnasDharar:
I have changed the file now. I have kept the syntax as it is.

AnasDharar:
@stephanegigandet can you please review pr once, if there are any further changes needed, do let me know.

PR #12433, opened 2025-09-29
taxonomy: correction stuffed tomatoes with rice

moon-rabbitOFF:
"Tomates farcies et riz" was badly translated as "Rice stuffed tomatoes" instead of "Stuffed tomatoes with rice". The rice is on the side. This category doesn't go in "en:Prepared vegetables" since the main ingredient is rice.


stephanegigandet (approved):
Thank you

PR #12455, opened 2025-10-02
taxonomy: modifications prepared meats

moon-rabbitOFF:
Added 2 categories of prepared meats and corrected the parent of Pork Pâtés.

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -89775,12 +89775,16 @@ fr: Charcuteries de bœuf
 en: Prepared pork meats
 fr: Charcuteries de porc
 
-< en:Chicken and its products
 < en:Prepared meats
+en: Prepared poultry meats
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -89751,12 +89751,16 @@ fr: Charcuteries de bœuf
 en: Prepared pork meats
 fr: Charcuteries de porc
 
-< en:Chicken and its products
 < en:Prepared meats
+en: Prepared poultry meats
+fr: Charcuterie de volaille
```
stephanegigandet: We seem to use the plural for "Charcuteries de poulet", dinde etc.
```suggestion
fr: Charcuteries de volaille
```

moon-rabbitOFF: Ah yes, thanks

PR #12468, opened 2025-10-06
fix: Update CGI version requirement in cpanfile

teolemon:
### What
- fix: Update CGI version requirement in cpanfile
- that allows the image to build again in my PR


alexgarel:
https://github.com/openfoodfacts/openfoodfacts-server/pull/12469 is a better way to go, I think.

PR #12475, opened 2025-10-08
feat: WIP - Support for new nutrition schema: product edit, API v2 + v3, import and export

stephanegigandet:
This PR is a work in progress to implement support for the new nutrition schema: https://docs.google.com/document/d/19ZRrlWJraJm61E6U7AwxQ1uubPDvmSuNfl9F1oLC0Tg/edit?tab=t.0

Included changes:
- new product edit form
- API v2 old nutrition fields backward compatibility
- API v2 new nutrition fields for new nutrition schema
- API v3 new nutrition fields for new nutrition schema
- export_csv_file support
- import_csv_file support

This PR changes the product edit form so that it can edit the different possible nutrition input sets for 1 specific source (the "packaging" source on the public platform, or the "manufacturer" (or USDA etc.) source on the producers platform).

Checkboxes are displayed to show the different possible input set columns (some of them are hidden if we don't have data yet).

In theory the units can be different for the same nutrient in different input sets. In practice it should not be often the case, so unless we do have existing different units, only 1 global unit column is shown.

image

Otherwise we can show units by input set:

image

Compared to the current product edit form, we do not support unknown nutrients, we limit users to nutrients that exist in the taxonomy.

So the system to add lines with labels that can be typed in or selected from suggestions has been changed: there is now a button at the bottom to display additional nutrients. This has the benefit of greatly simplifying the code.

Also removed: the salt and sodium rows are now independent, we don't compute one from the other anymore.

To be fixed:
- nutrient checks (e.g. sugars <= carbohydrates) should be made to work on all columns

The new product edit form still uses the API v2 parameters. So I added new fields  like nutrition_input_sets_${preparation}_${per}_${nid}_value_string to pass values.

The same function will be used to add support for those fields in the actual API v2, so that existing apps can set values for the input sets, without switching to the API v3 (which does not handle nutrients yet).

API v2 will continue to support existing nutrition fields and "guess" in which input set they belong.


CharlesNepote:
Great!! A small question: have you also changed CSS or are you using Power User Script? If so, I think you have to test the form without Power User Script to be sure that the form is still usable without it.

.test_groups_cache/integration_groups.json
alexgarel, line 1: @JagjeevanAK we need a way to ensure the generated file is always the same (unless a real change) to avoid those diffs ! Do you eventually have time to fix that ?

Maybe checking if:
* there is a significant difference in total time
* or a new test
before deciding to write down the change

stephanegigandet:
> Great!! A small question: have you also changed CSS or are you using Power User Script? If so, I think you have to test the form without Power User Script to be sure that the form is still usable without it.

@CharlesNepote I have not tested Power User Script on the new nutrition edit form, I suspect it might need some adaptations.

lib/ProductOpener/Nutrition.pm
```diff
@@ -40,7 +40,23 @@ use Exporter qw< import >;
...
+		&get_pers_for_product_type
+		&get_default_per_for_product
+		&get_unit_options_for_nutrient
+		&assign_nutrient_modifier_value_string_and_unit
+		&assign_nutrition_values_from_old_request_parameters
+		&assign_nutrition_values_from_request_parameters
+		&assign_nutrition_values_from_imported_csv_product
+		&assign_nutrition_values_from_imported_csv_product_old_fields
+		&assign_nutrition_values_from_request_object
+		&add_nutrition_fields_from_product_to_populated_fields
```
alexgarel, line 59: Good sign to see those small "one job" functions. It's a good sign IMO.

lib/ProductOpener/Nutrition.pm
```diff
@@ -308,20 +373,1471 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+
+=cut
+
+sub convert_nutrition_input_sets_hash_to_array($input_sets_hash_ref, $product_ref) {
+
+	$log->debug("convert_nutrition_input_sets_hash_to_array: start", {input_sets_hash_ref => $input_sets_hash_ref})
+		if $log->is_debug();
+
+	my $input_sets_ref = [];
+	if (defined $input_sets_hash_ref and ref $input_sets_hash_ref eq 'HASH') {
```
alexgarel, line 510: I would rather have 
```suggestion
	if (defined $input_sets_hash_ref) {
	    die("Expected a hash for convert_nutrition_input_sets_hash_to_array, got " . ref($input_sets_hash_ref) . " instead") unless (ref $input_sets_hash_ref eq 'HASH')
```
because I feel otherwise we are keeping silent an important error !
But maybe I'm missing something ?

lib/ProductOpener/Nutrition.pm
```diff
@@ -308,20 +373,1471 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+
+Type of the product (food, petfood, etc)
+
+=head3 Return values
+
+List of valid preparation states for the given product type
+
+=cut
+
+sub get_preparations_for_product_type ($product_type) {
```
alexgarel, line 622: It could also be a configuration value !

lib/ProductOpener/Nutrition.pm
```diff
@@ -308,20 +373,1471 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+
+Type of the product (food, petfood, etc)
+
+=head3 Return values
+
+List of valid per references for the given product type
+
+=cut
+
+sub get_pers_for_product_type ($product_type) {
```
alexgarel, line 649: There again looks like a config variable.

lib/ProductOpener/Nutrition.pm
```diff
@@ -308,20 +373,1471 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+		#
+		# nutrition.input_sets.packaging.prepared.100ml.nutrients.saturated-fat.value_string
+		# nutrition.input_sets.packaging.prepared.100ml.nutrients.saturated-fat.unit
+		#
+		# Note: the parameters are long because they mimic the structure of the product nutrition hash
+
+		# Go through all the possible input sets
+		foreach my $preparation (@preparations) {
+			foreach my $per (@pers) {
+				foreach my $source (@sources) {
```
alexgarel, line 1198: it could be interesting to keep track of processed field, keep track of field that we consider while computing sources, and at the end, compare the two list to issue a warning about not processed field, it could help spot typos and so, that otherwise will be hard to debug.

lib/ProductOpener/Nutrition.pm
```diff
@@ -308,20 +373,1471 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+											id => "nutrition.inputs_sets[$input_set_index].unspecified_nutrients",
+											value => $nid
+										},
+										impact => {id => "nutrient_ignored"},
+									},
+									200
+								);
+							}
+						}
+						if (scalar(@unspecified_nutrients) > 0) {
```
alexgarel, line 1571: Potential error: what if it was not empty and it is now… you have to remove the eventual old "unspecified_nutrients" key.

lib/ProductOpener/Nutrition.pm
```diff
@@ -308,20 +373,1471 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+			elsif ((not defined $nutrient_ref->{value_string}) or ($nutrient_ref->{value_string} eq '')) {
+				delete $input_set_ref->{nutrients}{$nid};
+			}
+			# If the modifier is undefined or empty, we remove it from the nutrient
+			elsif ((not defined $nutrient_ref->{modifier}) or ($nutrient_ref->{modifier} eq '')) {
+				delete $nutrient_ref->{modifier};
+			}
+		}
+	}
+	return;
```
alexgarel, line 1732: Isn't it worth, before returning, to verify that we still have values in the nutrient set, or at list unspecified_nutrients ? (or maybe there is another method to check that ?

lib/ProductOpener/Nutrition.pm
```diff
@@ -308,20 +373,1471 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+
+	if (not defined $product_ref->{nutrition}) {
+		return;
+	}
+
+	# Fields at the root of nutrition
+	my $item_number = 0;
+	foreach my $field ("no_nutrition_data_on_packaging") {
+		if (defined deep_get($product_ref, "nutrition", $field)) {
+			$populated_fields_ref->{$field} = $sort_key . '_0-root_' . sprintf("%02d", $item_number);
```
alexgarel, line 1829: In the lines below, the item_number is just after "$sort_key". Is it normal that it's different here ?

lib/ProductOpener/Nutrition.pm
```diff
@@ -308,20 +373,1471 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+		}
+		$item_number++;
```
alexgarel, line 1831: Is it normal that we only increment it once ? (and not in the loops below)

lib/ProductOpener/Nutrition.pm
```diff
@@ -308,20 +373,1471 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+			}
+		}
+	}
+	return;
+}
+
+=head2 add_nutrition_fields_from_product_to_populated_fields($product_ref, \%populated_fields, $sort_key)
+
+This function is used by Export.pm to generate the list of populated nutrition fields for a product.
+Export.pm can then create a CSV file with columns that have data for at least one product.
```
alexgarel, line 1738: It would be great to have a very short example of what it would produce for a product with 2 nutrient sets of 2 nutrients each, for example. Because otherwise it's hard to understand what we do below.

lib/ProductOpener/Nutrition.pm
stephanegigandet: Yes we could do that, in practice this should never happen, so maybe checking the type of the ref is overkill...

lib/ProductOpener/Nutrition.pm
stephanegigandet: I'll put it in Config.pm but keep the function to look it.

lib/ProductOpener/Nutrition.pm
stephanegigandet: Yes, it would be useful to report if some columns in a CSV file were ignored. A lot of changes would be needed though, so we can do it in another PR.

I filed an issue: https://github.com/openfoodfacts/openfoodfacts-server/issues/12556

lib/ProductOpener/Nutrition.pm
stephanegigandet: Well that's exactly what the else clause below does.

stephanegigandet: Or I am missing something?

lib/ProductOpener/Nutrition.pm
stephanegigandet: It's done in sub convert_nutrition_input_sets_hash_to_array($input_sets_hash_ref, $product_ref), but I can move the deletion of the nutrients hash if it's empty here, and I forgot the unspecified_nutrients, I'll add it

lib/ProductOpener/Nutrition.pm
stephanegigandet: Good idea, I'm adding an example.

lib/ProductOpener/Nutrition.pm
stephanegigandet: It's because after it's after the $input_set_sort_key, which is prefixed already.

As you suggested I added a sample output to make it clearer.

Essentially the goal is to have columns in a predictable order.

lib/ProductOpener/Nutrition.pm
alexgarel: ok !

cgi/product_multilingual.pl
```diff
@@ -1148,425 +1163,170 @@ ($product_ref, $field, $language, $request_ref)
...
-				and ($product_ref->{nutriments}{$nid . "_prepared"} ne ''))
-			or (    (defined $product_ref->{nutriments}{$nid . "_modifier"})
-				and ($product_ref->{nutriments}{$nid . "_modifier"} eq '-'))
-			or (    (defined $product_ref->{nutriments}{$nid . "_prepared_modifier"})
-				and ($product_ref->{nutriments}{$nid . "_prepared_modifier"} eq '-'))
-			or ($nid eq 'new_0')
-			or ($nid eq 'new_1')
+			($nutrient !~ /-$/)
+			# FIXME: add an or condition that is true if we have a value or modifier in any of the input sets for the selected source
+			or ($nid eq 'new_0') or ($nid eq 'new_1')
```
alexgarel, line 1172: I think new_0 and new_1 do not exist anymore, do they ? (because I don't see them in the foeach loop)

cgi/product_multilingual.pl
```diff
@@ -1148,425 +1163,170 @@ ($product_ref, $field, $language, $request_ref)
...
+						if (defined $input_set_nutrient_ref->{modifier}) {
+							$input_set_nutrient_ref->{modifier} eq '<' and $value_string = "&lt; $value_string";
+							$input_set_nutrient_ref->{modifier} eq "\N{U+2264}"
+								and $value_string = "&le; $value_string";
+							$input_set_nutrient_ref->{modifier} eq '>' and $value_string = "&gt; $value_string";
+							$input_set_nutrient_ref->{modifier} eq "\N{U+2265}"
+								and $value_string = "&ge; $value_string";
+							$input_set_nutrient_ref->{modifier} eq '~' and $value_string = "~ $value_string";
+							$input_set_nutrient_ref->{modifier} eq '-' and $value_string = '-';
+						}
```
alexgarel, line 1244: using the "and" notation is a bit confusing… if is better !
```suggestion
						# If we have a modifier, include it in the value_string
						if (defined $input_set_nutrient_ref->{modifier}) {
							$value_string = "&lt; $value_string" if $input_set_nutrient_ref->{modifier} eq '<' ;
							$value_string = "&le; $value_string" if $input_set_nutrient_ref->{modifier} eq "\N{U+2264}";
							$value_string = "&gt; $value_string" if $input_set_nutrient_ref->{modifier} eq '>';
							$value_string = "&ge; $value_string" if $input_set_nutrient_ref->{modifier} eq "\N{U+2265}";
							$value_string = "~ $value_string" if $input_set_nutrient_ref->{modifier} eq '~';
							$value_string = '-' if $input_set_nutrient_ref->{modifier} eq '-';
						}
```

Also I'm asking myself, appart form "-", should we add modifiers if $value_string is empty ?

cgi/product_multilingual.pl
alexgarel, line 1: It's really difficult to review this kind of code, so I just sip at it. But I really have no idea which kind of structure we define, and it's not local. Just to say I may have missed something.

cgi/product_multilingual.pl
```diff
@@ -1101,42 +1104,54 @@ ($product_ref, $field, $language, $request_ref)
...
+	$template_data_ref_display->{source} = $source;
+	$template_data_ref_display->{preparations} = \@preparations;
+	$template_data_ref_display->{pers} = \@pers;
+	$template_data_ref_display->{nutrients} = \@nutrients;
+	$template_data_ref_display->{input_sets} = \%input_sets;
 
-	my %nutrition_data_exists = ();
-	my @nutriments;
-	foreach my $nutriment (@{$nutriments_tables{$nutriment_table}}, @unknown_nutriments, 'new_0', 'new_1') {
+	foreach my $nutrient (@{$nutriments_tables{$nutriment_table}}) {
```
alexgarel, line 1149: It would be worth explaining what structure we are building. Because it's hard to understand the code.
It's optional of course.

docker-compose.yml
alexgarel, line 1: Yes I really think we can't have this dependency if we want pro platform to work.
One way to keep the dependency would be to have an alternate "fake" postgresql service for the pro platform, that just do nothing (sleep infinity).

lib/ProductOpener/Export.pm
```diff
@@ -568,6 +520,15 @@ sub export_csv ($args_ref) {
...
 				}
+				# Nutrition fields: input sets
+				elsif ($field =~ /^nutrition\.input_sets\.(.*)$/) {
+					# the field key is of the form nutrition.input_sets.<input_set_id>.<property> that exactly matches the input set keys
+					$value = deep_get($input_sets_hash_ref, split(/\./, $1));
+				}
+				# Nutrition: other fields
+				elsif ($field =~ /^nutrition\./) {
+					$value = deep_get($product_ref, split(/\./, $field));
+				}
```
alexgarel, line 531: That's smart this deep_get + split combination.

lib/ProductOpener/Food.pm
```diff
@@ -308,89 +308,14 @@ sub default_unit_for_nid ($nid) {
...
-
-	if ((defined $modifier) and ($modifier ne '')) {
-		$product_ref->{nutriments}{$nid . "_modifier"} = $modifier;
-	}
-	else {
-		delete $product_ref->{nutriments}{$nid . "_modifier"};
-	}
-
-	if ((defined $value) and ($value ne '')) {
+	## FIX ME
```
alexgarel, line 313: The well known formulation is FIXME !
```suggestion
	## FIXME
```

lib/ProductOpener/Food.pm
```diff
@@ -937,6 +862,20 @@ It is a list of nutrients names with eventual prefixes and suffixes:
...
 );
 
+# Compute a hash of all nutrients that are valid in at least one region for the site flavor (opf, off, ...)
+%valid_nutrients = ();
+
+foreach my $region (keys %nutriments_tables) {
+	# Use the flavor (off, opff) to select regions that start with the flavor
+	next if $region !~ /^$flavor\_/;
+	foreach (@{$nutriments_tables{$region}}) {
+		my $nutriment = $_;    # copy instead of alias
```
alexgarel, line 872: why not the standard notation ?
```perl
	foreach my $nutriment (@{$nutriments_tables{$region}}) {
```
Is this what the copy instead of alias means ?
It's cryptic to a non Perl practitioner…

lib/ProductOpener/Import.pm
alexgarel, line 1: Great to see improvement in readability on this file !

templates/web/pages/product_edit/product_edit_form_display.tt.html
alexgarel, line 1: Didn't read, I think the best review is to test it !

cgi/product_multilingual.pl
stephanegigandet: Good catch, removing them

cgi/product_multilingual.pl
stephanegigandet: Changing to if

If $value_string is empty we should not have modifiers.

cgi/product_multilingual.pl
```diff
@@ -1148,425 +1163,170 @@ ($product_ref, $field, $language, $request_ref)
...
-				);
-			}
-
-			if (   (defined get_property("nutrients", "zz:$nid", "dv_value:en"))
-				or ($nid =~ /^new_/)
-				or (uc($unit) eq '% DV'))
-			{
-				push @units, '% DV';
-			}
-			if (   (defined get_property("nutrients", "zz:$nid", "iu_value:en"))
```
jayaddison, line 1325: @stephanegigandet Is the International Units (IU) functionality here migrated elsewhere, or is it removed?  (I'm preparing to test #12500 locally, and have been reading more about IU today)

stephanegigandet: Hi @jayaddison , it's still in the new product edit form:

image

results in:

image

jayaddison: My mistake; both in production, and also from a checkout of #12500, I can find the `IU` field in the nutriment unit field (NB: your screenshot shows DV% - but I think it would have shown IU if the dropdown was open).

I was worried because I wasn't sure where that logic has moved to -- but I can confirm that it continues to work (I tested editing a product, too).  Sorry for the distraction!

jayaddison: NB: I didn't spend a long time exploring the updated nutriment data management UI -- only enough to confirm that I was wrong -- but from a quick usage it seemed like a nice improvement.  If+when time allows I'll test it out a bit more.

PR #12480, opened 2025-10-10
fix: make OIDC_DISCOVERY_URL to use keycloak:8080 (needed on dev environment)

stephanegigandet:
The change introduced in https://github.com/openfoodfacts/openfoodfacts-server/commit/e17c36d02b6a270704e7e387e2e49fd743b322d6 does not work on the dev environment. So I made this change back locally.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -553,10 +553,10 @@ Categories:
   - changed-files:
       - any-glob-to-any-file: "taxonomies/food/categories.txt"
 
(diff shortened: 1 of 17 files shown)


john-gom:
Hi @stephanegigandet . I have re-worked this to get the setting from Config2.pm so that it can be overridden in production.

stephanegigandet:
> Hi @stephanegigandet . I have re-worked this to get the setting from Config2.pm so that it can be overridden in production.

Thank you @john-gom !

PR #12481, opened 2025-10-10
fix: make dev for pro platform needs postgres

stephanegigandet:
A PR to be able to run the pro platform in a dev environment. Maybe not the right fix but it unblocks me.

Slack discussions: 

https://openfoodfacts.slack.com/archives/C02LDQDDD/p1759764416701789
https://openfoodfacts.slack.com/archives/C02LDQDDD/p1760090829660309

I'm still struggling to run the pro platform dev environment... If I run "make dev", I get:

`"service "incron" depends on undefined service "postgres": invalid compose project"
`
I can comment out the dependency on "postgres" in docker-compose.yml, and then "make dev" works. But when I run it and it tries to access postgres (the pro platform uses Minion to schedule tasks which uses Postgres), it fails....

`DBI connect('dbname=minion;host=postgres','productopener',...) failed: could not translate host name "postgres" to address: Temporary failure in name resolution at /opt/perl/local/lib/perl5/Mojo/Pg.pm line 73.`

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -1170,11 +1170,7 @@ sub store_product ($user_id, $product_ref, $comment, $client_id = undef) {
 		$log->info("changing product type",
 			{old_product_type => $product_ref->{old_product_type}, product_type => $product_ref->{product_type}})
 			if $log->is_info();
-		# We need to remove the product from its previous collection, unless we are on the pro platform
-		# where we have only one collection for all product types
-		if (not $server_options{private_products}) {
-			$delete_from_previous_products_collection = 1;
-		}
+		$delete_from_previous_products_collection = 1;
 		delete $product_ref->{old_product_type};
 	}
 


env/env.off-pro
```diff
@@ -3,7 +3,9 @@
 
 COMPOSE_PROJECT_NAME=po_off_pro
 # we do not want postgres on the pro side, we use the one from off
-COMPOSE_PROFILES=
```
alexgarel, line 6: I don't think this is a good idea
In fact I'm not sure you need to run "make dev" in pro shell.
you need to run docker compose up.
but you should not need to run make dev.
make dev is to be run in a normal shell.

Those profiles are necessary to avoid having two postgresql instance, which would confuse minions (not sharing the same postgresql)

john-gom:
Maybe it would be easier to just switch to using the shared services Postgres instead? Off-auth has an example of how to bootstrap a new Postgres user for the service

stephanegigandet:
Fixed by https://github.com/openfoodfacts/openfoodfacts-server/pull/13055

PR #12482, opened 2025-10-10
fix: keep all product types in mongodb off database on pro platform

stephanegigandet:
On the pro platform, we have only one mongodb database for all products of all product types.

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -1170,7 +1170,11 @@ sub store_product ($user_id, $product_ref, $comment, $client_id = undef) {
 		$log->info("changing product type",
 			{old_product_type => $product_ref->{old_product_type}, product_type => $product_ref->{product_type}})
 			if $log->is_info();
-		$delete_from_previous_products_collection = 1;
+		# We need to remove the product from its previous collection, unless we are on the pro platform
+		# where we have only one collection for all product types
+		if (not $server_options{private_products}) {
+			$delete_from_previous_products_collection = 1;
+		}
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Products.pm
```diff
@@ -1170,7 +1170,11 @@ sub store_product ($user_id, $product_ref, $comment, $client_id = undef) {
 		$log->info("changing product type",
 			{old_product_type => $product_ref->{old_product_type}, product_type => $product_ref->{product_type}})
 			if $log->is_info();
-		$delete_from_previous_products_collection = 1;
+		# We need to remove the product from its previous collection, unless we are on the pro platform
+		# where we have only one collection for all product types
+		if (not $server_options{private_products}) {
+			$delete_from_previous_products_collection = 1;
+		}
```
alexgarel, line 1177: Is this enough ? I mean, if the product is updated, changes won't be reflected in the right collection.
This means error / quality indicators won't work, and so on.

stephanegigandet: This is only for the pro platform, where we have only one collection for all product types. So when we change the product type of a product on the pro platform, the new collection is the same as the previous collection, and we don't want to delete from the previous collection.

Error quality indicators, taxonomies etc. is another story: all the data enrichment is done with code and taxonomies for the server's product type, so when we change the product type, they are not recomputed with the new product type. But that's already like that on the public platform.

alexgarel: Ok cool ! In this case, this is fine.

PR #12477, opened 2025-10-09
feat: taxonomy based recommendation panels

stephanegigandet:
New generic recommendation panel that is activated through categories taxonomy properties:

image

Some strings were taken from @teolemon 's PR https://github.com/openfoodfacts/openfoodfacts-server/pull/12476

New recommendation panels are specified in taxonomies such as the category taxonomy with properties like:

```
	recommendation_panel_id:en: recommendation_health
	recommendation_panel_evaluation:en: bad
	recommendation_panel_icon:en: arrow-bottom-right-thick.svg
	recommendation_panel_msgctxt_prefix:en: recommendation_eu_tobacco_health_warning
	recommendation_panel_source_url:en: European Union
	recommendation_panel_source_language:en: en
	recommendation_panel_source_url:en: https://health.ec.europa.eu/tobacco/product-regulation/health-warnings_en#combined-health-warnings-in-the-eu
```

The value of the property recommendation_panel_msgctxt_prefix:en: is used to find the displayable strings in the .po files:

```
	msgctxt "recommendation_eu_tobacco_health_warning_title"
	msgid "Smoking kills"
	msgstr "Smoking kills"

	msgctxt "recommendation_eu_tobacco_health_warning_subtitle"
	msgid "Tobacco seriously damages health. Don't start or quit now."
	msgstr "Tobacco seriously damages health. Don't start or quit now."

	msgctxt "recommendation_eu_tobacco_health_warning_text"
	msgid "Smoking is highly addictive. Don't start. Smoking kills – quit now. Smoking clogs the arteries and causes heart attacks and strokes. Smoking causes fatal lung cancer. Smoking when pregnant harms your baby. Protect children: don't make them breathe your smoke. Your doctor or your pharmacist can help you stop smoking. Smoking may reduce the blood flow and causes impotence. Smoking causes ageing of the skin. Smoking can damage the sperm and decreases fertility. Smoke contains benzene, nitrosamines, formaldehyde and hydrogen cyanide."
[...]

PR #12487, opened 2025-10-13
docs: How to for official warnings

teolemon:
### What

- docs:How to for official warnings
- related to #12477


docs/dev/how-to-add-generic-recommendation-panel.md
alexgarel, line 1: I think you should add it in docs/dev/how-to-add-generic-recommendation-panel.md folder… I don't think many will saw it otherwise.

Maybe you could keep this file as a symlink to the documentation.

alexgarel: Also, IMO it is not a tutorial but an "How to" (and your title says that :-) )

stephanegigandet:
Looks good to me, but please move to the name Alex suggested. We could also add a link to it in the header of the categories.txt file where we explain the different properties.

docs/dev/how-to-add-generic-recommendation-panel.md
teolemon: done

PR #12493, opened 2025-10-13
fix: filter out unknown nutrients + need convert_blessed JSON option for booleans

stephanegigandet:
perl tests/unit/product_schema_changes.t --update
was crashing because we now have booleans in the product schema.

I'm not sure why it changes the formatting of the test output though, maybe the expected results were created manually?

This PR adds some logic for unknown nutrients for the migration to the new nutrition facts schema.

### DIFF
--- a/lib/ProductOpener/Nutrition.pm
+++ b/lib/ProductOpener/Nutrition.pm
@@ -308,7 +308,7 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
 	return;
 }
 
-=head2 filter_out_nutrients_not_in_taxonomy
+=head2 filter_out_nutrients_not_in_taxonomy ($product_ref)
 
 In the old nutrition facts schema (2025 and before), we authorized users to add any nutrient they wanted, even if they did not exist in the taxonomy.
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Nutrition.pm
```diff
@@ -307,4 +308,92 @@ sub convert_nutrient_to_100g ($nutrient_ref, $original_per, $original_per_quanti
...
+	# 'en-some-unknown-nutrient_value' => 1.23,
+
+	my %hash_nutrients = map {/^([a-z][a-z\-]*[a-z]?)(?:_\w+)?$/ ? ($1 => 1) : ()} keys %{$product_ref->{nutriments}};
+
+	foreach my $nid (sort keys %hash_nutrients) {
+
+		# check that the nutrient exists in the taxonomy
+		my $nutrient_id = "zz:" . $nid;
+		if (not exists_taxonomy_tag("nutrients", $nutrient_id)) {
+			# Check if we can canonicalzie the nid to a known nutrient
```
Vic142: there is just a typo for canonicalize in this comment

stephanegigandet:
```suggestion
			# Check if we can canonicalize the nid to a known nutrient
```

Issue #12490, opened 2025-10-13
Fix lint_taxonomies.pl or make another test so that such a problem raises an error or is fixed


teolemon:
### Task
Fix lint_taxonomies.pl or make another test so that such a problem raises an error or is fixed

```

< en:Salami
Mangalitza

< en:Salami
ro: Parizer

```

alexgarel:
If I remember well, in linter I try to categorize which type of line we have. We should report an error if the line is of unknown type

alexgarel:
I'll fix this

alexgarel:
@teolemon I have proposed a PR but there is a lot to fix in some beauty taxonomies…

PR #12495, opened 2025-10-13
fix: taxonomy lint raises on unknown line types

alexgarel:
* modify linter to detect a wider range of language prefix
* add an error if we encounter a line that is of an unknown type
* fix some taxonomies consequently

fixes: #12490

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -442,7 +442,7 @@ check_critic:
 	@echo "🥫 Checking with perlcritic"
 	test -z "${TO_CHECK}" || ${DOCKER_COMPOSE_BUILD} run --rm --no-deps backend perlcritic ${TO_CHECK}
 
-TAXONOMIES_TO_CHECK := $(shell [ -x "`which git 2>/dev/null`" ] && git diff origin/main --name-only | grep -E 'taxonomies.*/.*\.txt$$' | grep -v '\.result.txt' | xargs ls -d 2>/dev/null | grep -v "^.$$")
+TAXONOMIES_TO_CHECK := $(shell [ -x "`which git 2>/dev/null`" ] && git diff origin/main --name-only | grep -P 'taxonomies/(beauty/|food/|petfood/|product/|)[^/]+\.txt$$' | grep -v '\.result.txt' | xargs ls -d 2>/dev/null | grep -v "^.$$")
 
 # TODO remove --no-sort as soon as we have sorted taxonomies
 check_taxonomies:


alexgarel:
@teolemon @stephanegigandet has is I got a lot of errors, for examples on categories.txt for:
* entries with code like "en-GB:"
* entries with code like "fr_be:"
* entries with code like "sco:" , "pam:"

I know we want to keep some of it…

What do we want to do ?
 
Be lax for now and accept all those ?
(you can test on the branch by using make lint_taxonomies TAXONOMIES_TO_CHECK=taxonomies/food/categories.txt)

alexgarel:
I committed a version handling more language patterns.

alexgarel:
Here is the log of the different errors this change will pop-up !

I don't know if someone has time to process them ?
[lint-log2.txt](https://github.com/user-attachments/files/22899997/lint-log2.txt)

scripts/taxonomies/lint_taxonomy.pl
```diff
@@ -250,8 +254,21 @@ ($lines_iter)
...
+			}
+			# undefined ! this should be rejected
 			else {
+				push(
+					@errors,
+					{
+						severity => "Error",
+						type => "Correctness",
+						line => $line_num,
+						message => "Unknown line type !\n- $line",
```
Freso:
```suggestion
						message => "Unknown line type!\n- $line",
```

I _think_ the “!” is just there as a grammatical exclamation point, right? It doesn’t denote any addition meaning? If so, let’s use English grammar and not have a space between the word and the punctuation. :)

scripts/taxonomies/lint_taxonomy.pl
```diff
@@ -88,6 +88,9 @@ ($fd)
 	}
 }
 
+# explicit regexp of the different language variant we got
+my $language_prefix_re = qr/(\w{2,3}|\w{2}[-_]\w{2,3})/;
```
Freso: Maybe
```suggestion
my $language_prefix_re = qr/(\w{2,3}(?:[-_]\w{2,3})?)/;
```

To also include language/locale codes like `bat-smg` (though that specific one should be changed to `sgs`)? Or

```suggestion
my $language_prefix_re = qr/(\w{2,3}(?:[-_]\w{2,4})?)/;
```

to include Chinese variants like `zh-han[st]`?

Maybe it’d be good to have a look at what language codes that are there and figure out what to keep and what to drop, to figure out what the regex should look like. E.g., `\w{2,3}-x-old` entries can probably be dropped just fine (and also won’t be matched with any of these regexes), but others (like the mentioned `zh-han[ts]`) might ones we’d want to keep around?

Freso:
Small cleanup of `simple:` language entries: https://github.com/openfoodfacts/openfoodfacts-server/pull/12503

teolemon:
https://github.com/openfoodfacts/openfoodfacts-server/pull/12501

Freso:
Remove/fixing `zh-min-nan:` entries: https://github.com/openfoodfacts/openfoodfacts-server/pull/12505

scripts/taxonomies/lint_taxonomy.pl
alexgarel: Yes, that's my french bias ;-)

scripts/taxonomies/lint_taxonomy.pl
alexgarel: yes let's go for the tolerant one, for now :-)

Freso:
@alexgarel Any chance for an updated lint-log now that the first few batches of fixes have gone in? :eyes:

teolemon:
@alexgarel can you rebase and send us an updated list of the reported mistakes ?

alexgarel:
@teolemon and @Freso I did some changes to better handles properties and enable ":" inside synonyms lines.

I also did some manual fixes to taxonomies.

Here is where we are at this point: 
[2025-10-21-lint-taxonomies-output.txt](https://github.com/user-attachments/files/23019423/2025-10-21-lint-taxonomies-output.txt)


As only countries.txt and languages.txt have problems now, I suggest to review and merge this PR asap.

Are you ok for commenting lines not yet passing in countries.txt and languages.txt ?

Freso:
Commenting “bad” entries in countries/languages.txt sounds fine to me.

PR #12504, opened 2025-10-14
fix: small changes to new nutrition schema downgrade

stephanegigandet:
Small changes, in particular:
- do not return an empty nutriments {} hash
- more comments to detail specific cases


alexgarel (approved):
LGTM

PR #12505, opened 2025-10-14
taxonomy: fix: Fold `zh-min-nan` into `nan`

Freso:
`zh-min-nan` is a code formerly used for Southern Min, `nan` is the more recent ISO 639-3 code: https://iso639-3.sil.org/code/nan

Preparation for https://github.com/openfoodfacts/openfoodfacts-server/pull/12495


alexgarel (approved):
Thanks !

PR #12509, opened 2025-10-15
fix: use new nutrition schema for CSV export of the full database

stephanegigandet:
This PR merges against new-nutrition-stephane which has import CSV support for new nutrition schema (needed for the export.t scripts which first imports products).

The idea is to keep the same names for the nutrients (e.g. salt_100g), but populate them from the aggregated set of the new schema.

Notes:
- we are missing the initial values as we merge against new-nutrition-stephane, if we merge against merge, we can see the different values. Check this PR against main to see the diffs between the current CSV and the new one:
https://github.com/openfoodfacts/openfoodfacts-server/pull/12511/files#diff-7f8efddece7a6648c7c820c9c4806a192845fbbf755f8b247f624f4a740e77df

- we are missing fruits vegetables estimates: those have not been migrated to the new schema yet

- we are missing the energy_100g field: that's because we currently don't compute it in the aggregated set (we probably should)

- Note: we do not upgrade the schema of the products, they are read directly from MongoDB.
This means that some old products may not have all the fields we expect (e.g. for the migration to the new nutrition schema)
One possibility could be to upgrade products read from MongoDB (but it could add a lot of overhead if we do it on all products)
For those types of changes, we can also suspend the export for a few days, and instead run upgrade_all_products.pl once.


stephanegigandet:
Merged in https://github.com/openfoodfacts/openfoodfacts-server/commit/2a7b1b87d86384b38c6d1b952c052118be64a6d1

PR #12512, opened 2025-10-15
fix: energy field in new nutrition aggregated and input sets

stephanegigandet:
In the early days, we used to only have "energy" in kJ or kcal, then we split it into energy-kj and energy-kcal, and we computed an energy field (always in kJ).

For the new nutrition schema, we should compute "energy" for the aggregated set, but not for the input sets.

This PR:
- adds the computation of "energy" for the aggregated set
- filters out "energy" when we upgrade an old nutrition schema to populate the input sets


Vic142 (approved):
looks good to me !

PR #12513, opened 2025-10-15
fix: compute nutrient levels / nutrition traffic lights with new nutrition schema

stephanegigandet:
Compute the nutrient levels from the aggregated set.

Also a bit of preparation work for the Nutri-Score computation.

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -2001,7 +2001,7 @@ sub check_availability_of_nutrients_needed_for_nutriscore ($product_ref) {
 		foreach my $nid ("energy", "fat", "saturated-fat", "sugars", "sodium", "proteins") {
 			# If we don't set the 100g figure then this should flag the item as not enough data
 			if (not deep_exists($product_ref, "nutrition", "aggregated_set", "nutrients", $nid, "value")) {
-				# we have two special case where we can deduce data
+				# we have two special cases where we can deduce data
 				next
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Food.pm
```diff
@@ -2021,10 +2025,13 @@ sub check_availability_of_nutrients_needed_for_nutriscore ($product_ref) {
...
 		# for others, display a warning when the value is missing
-		# do not display a warning if fibers are not specified on the product ('-' modifier)
-		if (    (not defined $product_ref->{nutriments}{"fiber" . $prepared . "_100g"})
-			and (not defined $product_ref->{nutriments}{"fiber" . $prepared . "_modifier"})
-			and not(has_tag($product_ref, "categories", "en:sodas")))
+		# do not display a warning if fibers are not specified on the product ('-' modifier: listed in the unspecified_nutrients array)
+		if (
+			(not deep_exists($product_ref, "nutrition", "aggregated_set", "nutrients", "fiber", "value"))
+			and (not grep {$_ eq 'fiber'}
+				@{deep_get($product_ref, "nutrition", "aggregated_set", "unspecified_nutrients") // []})
```
Vic142: I don't think the aggregated set has an _unspecified_nutrients_ field so I think this test is useless

stephanegigandet: Good point, I'll remove it. Thanks!

PR #12528, opened 2025-10-18
taxonomy: Chocolate alternatives without cocoa

moon-rabbitOFF:
Added two categories for chocolate and bonbons alternatives.


Freso (approved):
LGTM!

PR #12539, opened 2025-10-20
fix: missing folksonomy config on opf

alexgarel:
fixes: https://github.com/openfoodfacts/openfoodfacts-webcomponents/issues/338


hangy (approved):

Freso:
@teolemon I did (before leaving my previous comment). The exact same line(s) being added here is present in all of `Config_o{b,f,pf}f.pm` – `Config_opf.pm` is the only one that doesn’t have it.

alexgarel:
> Have you checked for opff and OBF?

Yes, it's working there.

PR #12541, opened 2025-10-20
ci: add additional feedback when PR action is done

Freso:
Right now there’s feedback (in the form of a reaction) when a PR action is seen (and thus begun being acted on), but there’s no feedback when it’s done being processed, esp. when there’s no resulting changes and thus no commits.

This replaces the 👀 reaction with 🚀 at the end of the action, letting the caller know the action has finished running.

(It also utilises YAML anchors to not have to repeat code.)

### DIFF
--- a/.github/workflows/pr_actions.yml
+++ b/.github/workflows/pr_actions.yml
@@ -14,7 +14,8 @@ jobs:
       (github.event.comment.body == '/lint')
     runs-on: ubuntu-latest
     steps:
-    - name: Add reaction to comment
+    - &step-reaction-seen
+      name: Add “seen” reaction to comment
       uses: peter-evans/create-or-update-comment@v4
       with:
         comment-id: ${{ github.event.comment.id }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_actions.yml
```diff
@@ -61,6 +62,13 @@ jobs:
         status_options: '--untracked-files=no'
         skip_dirty_check: false
         create_branch: no
+    - &step-reaction-completed
+      name: Add “completed” reaction to comment
+      uses: peter-evans/create-or-update-comment@v4
+      with:
+        comment-id: ${{ github.event.comment.id }}
+        reactions: rocket
```
Freso, line 70: Would 👍 be better than :rocket: ? I originally wanted to use :heavy_check_mark: but that is not a valid GitHub comment reaction. :frowning_face:

alexgarel: Both are ok :-)

PR #12542, opened 2025-10-20
taxonomy: fix: Fold `be-x-old` into `be-tarask`

Freso:
`be-x-old` is a code formerly used for Belarusian Classical Orthography (Taraškievica), which has since been given the IANA code `be-tarask`.

See https://en.wikipedia.org/wiki/Tara%C5%A1kievica and https://en.wikipedia.org/wiki/Belarusian_Wikipedia for more background.

I don’t know if we’ll want to keep `be-tarask` or if it should be further folded into `be` for purposes of O*F, but this PR takes one step toward getting rid of superfluous language codes.

Preparation for https://github.com/openfoodfacts/openfoodfacts-server/pull/12495


alexgarel (approved):
Thanks.

PR #12555, opened 2025-10-24
taxonomy: fix fetan -> feta (fr)

stephanegigandet:
Marie found a type "fetan" instead of "feta"

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -46201,7 +46201,7 @@ de: Feta
 el: Φέτα, Feta
 es: Feta
 fi: feta
-fr: Fetan
+fr: Feta
 hr: Feta
 hu: Feta sajtok
 it: Feta


taxonomies/food/categories.txt
```diff
@@ -46201,13 +46201,14 @@ de: Feta
...
 fi: feta
-fr: Fetan
+fr: Feta
 hr: Feta
 hu: Feta sajtok
 it: Feta
 ja: フェタチーズ, フェタ
 nl: Fetas, Feta's
 pl: Feta, Ser feta
+xx: Feta
```
```suggestion
```

There’s already an `xx` line. :)

taxonomies/food/categories.txt
```diff
@@ -46201,13 +46201,14 @@ de: Feta
 el: Φέτα, Feta
 es: Feta
 fi: feta
-fr: Fetan
+fr: Feta
```
Freso, line 46204: Given that there’s already an `xx: Feta` entry, this line could probably also just be removed. I know there’s been some thoughts before about the balance between keeping the taxonomy files light-ish and explicitly denoting that a word has been translated to a given language. There’s also an argument that this would be scope-creep given that removing the `fr` entry on this basis should probably also lead to a bunch of the other “Feta” entries getting removed.
```suggestion
```

Freso (approved):
Other than the duplicate `xx` line (which is causing the test to fail), this looks good to me!

I’ve added some thought (and another suggestion) in a comment for this review, but that suggestion is just that: a suggestion. My approval stands with or without it.

PR #12561, opened 2025-10-27
test: fix workflow - 2nd try

alexgarel:


.github/workflows/pull_request.yml
```diff
@@ -373,6 +373,8 @@ jobs:
         cat matrix_config.json
 
   unit_tests:
+    # for now "🐪 Perl Unit Tests" is the expected label for en of tests,
```
Freso, line 376: “en of tests”? end?

Freso (approved):
Please!

One small comment about phrasing in a comment, but that can be fixed later.

.github/workflows/pull_request.yml
alexgarel: https://github.com/openfoodfacts/openfoodfacts-server/pull/12563

PR #12565, opened 2025-10-28
fix: compute nutrition aggregated set with IU and % DV units

stephanegigandet:
Input sets can have some nutrients in IU (international unit) or % DV (daily value) for vitamins in some countries.

In the aggregated set, we convert everything to g.


Vic142 (approved):
looks good !

PR #12572, opened 2025-10-29
taxonomy: Update stores.txt with Wikidata entries

teolemon:
### What
taxonomy: Update stores.txt with Wikidata entries


Freso (approved):
Seems good to me. Might be worth noting that this is currently unused, so this won’t actually have any effect on any of the O*F sites. :)

PR #12610, opened 2025-11-02
fix(gulpfile): Return Promise correctly to make Gulp build fail if a file is missing

### LINKED ISSUE #10709: ci: Build issues in Gulp do not cause builds to fail
### What

A build failure in the Gulp (`make front_build`) step does not cause the build to fail. This is is valid for [main branch](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/10435133992/job/28898781682#step:6:392) and [PR checks](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/10435587297/job/28899715624?pr=10065#step:4:257).

Delete a file that's required by `gulpfile.ts` and submit a PR with it missig.

### Expected behavior

The PR check should fail.

### Why

A build failure in the Gulp step could break the full UI of ProductOpener. It's better to know about those early on.

hangy:
### What

Apparently, this works better if we return the `Promise` directly.

### Screenshot

N/A

- Fixes #10709

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -217,19 +217,18 @@ function copyImages() {
   return src(imagesSrc).pipe(dest("./html/css/dist"));
 }
 
-export default function buildAll() {
-  return new Promise(
-    parallel(
-      copyJs,
-      buildJs,
-      buildjQueryUi,
-      copyCss,
-      copyImages,
-      jQueryUiThemes,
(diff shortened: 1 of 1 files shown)


hangy:
> Ah great, I had issues because sometimes I broke a JS file while doing local dev, but make front_build was not complaining.

Same! 😆

PR #12613, opened 2025-11-02
build: Reenable coverage reporting for unit and integration tests in Codecov

hangy:
### What

Upload each groups' test result to Codcov with a distinct flag.

### Screenshot

N/A

N/A


stephanegigandet:
I don't remember why it was disabled?

hangy:
> I don't remember why it was disabled?

Part of #12064. I think, the assumption was, that we have to manually combine the test results. But apparently, Codecov can do that just fine, if we upload the separate results with distinct keys.

PR #12616, opened 2025-11-03
feat: simplified panels for mobile app

stephanegigandet:
This PR introduces simplified knowledge panels that are intended for the mobile app.
See discussion: https://docs.google.com/document/d/1vJ9gatmv8pCXxyOERmYD16jOKRWJpz1RaQQ5MEcTxms/edit?tab=t.psqeabp0bh82

- new simplified_root panel (includes placeholders for price_card and folksonomy_engine card)
- new simplified_health_card and simplified_environmental_card panels
- ?simplified_panels=1 option on the website to view the simplified cards instead of the normal ones (for debugging)

Notes/questions:

- No report_problem_card or product_data_card?
- No title for the environmental_score_adjustments? (used to display 2 x 2 subpanels)
- Kept the health recommendations (e.g. cigarettes, alcohol, pulses etc.)
- The simplified cards will hide some features, like added sugars detection and details about vegan / vegetarian status for instance


lib/ProductOpener/Display.pm
```diff
@@ -7940,10 +7940,24 @@ JS
...
+	$template_data_ref->{environment_card_panel} = display_knowledge_panel(
+		$product_ref,
+		$product_ref->{"knowledge_panels_" . $lc},
+		$simplified_prefix . "environment_card"
+	);
+	$template_data_ref->{health_card_panel} = display_knowledge_panel(
+		$product_ref,
+		$product_ref->{"knowledge_panels_" . $lc},
+		$simplified_prefix . "health_card"
+	);
```
alexgarel, line 8145: I don't understand what we are doing here…

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -280,6 +280,33 @@ sub create_knowledge_panels ($product_ref, $target_lc, $target_cc, $options_ref,
 		$request_ref
 	);
 
+	# Create the simplified root panel that contains simplified versions of the health and environment cards
+	# (used on mobile app)
```
alexgarel: IMO we should add them only if we asked for them.

lib/ProductOpener/Display.pm
stephanegigandet: This is to allow testing the simplified cards on the website, to see how they look like, the children panels etc. So if you send ?simplified_panels=1, then we replace the environment and health cards with the simplified versions, on the website. I'll add more comments.

lib/ProductOpener/KnowledgePanels.pm
stephanegigandet: ok, we can use a similar option like the one we have for the physical activities panel:
activate_knowledge_panel_physical_activities

lib/ProductOpener/KnowledgePanels.pm
```diff
@@ -89,6 +89,11 @@ Initialize the options for knowledge panels from parameters.
 
 sub initialize_knowledge_panels_options ($knowledge_panels_options_ref, $request_ref) {
 
+	# Activate simplified_root + simplified cards only when specified
+	if (single_param("activate_knowledge_panels_simplified")) {
+		$knowledge_panels_options_ref->{activate_knowledge_panels_simplified} = 1;
+	}
```
alexgarel: Can you add this to the OpenAPI spec ?

alexgarel (approved):
LGTM.

See my comment.

Can you also add an integration test where you ask for simplified knowledge panels ?

teolemon:
@g123k @primaelquemerais WDYT ?

PrimaelQuemerais:
Thank you for the PR @stephanegigandet !

I'm copying here my reply from Slack :
- We do in fact need `report_problem_card` for the order of tabs.
- `product_card` is not used on our side at the moment.

Everything else looks good, like you said it would be nice to be able to test it and I can provide some more feedback if further modifications are needed.

PR #12622, opened 2025-11-04
fix: attribute_unwanted_ingredients_tags parameter for search

stephanegigandet:
Fixes issue found by @monsieurtanuki : https://github.com/openfoodfacts/openfoodfacts-dart/issues/1134#issuecomment-3475865815

https://fr.openfoodfacts.org/cgi/search.pl?search_terms=pizza&langs=fr&page_size=5&page=1&fields=code%2Cbrands&json=1&attribute_unwanted_ingredients_tags=en:garlic returns 0 results because the search think it is a filter on tags as the parameter ends with _tags

Also added some tests.

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -4776,7 +4776,9 @@ sub add_params_to_query ($params_ref, $query_ref) {
 		# Tags fields can be passed with taxonomy ids as values (e.g labels_tags=en:organic)
 		# or with values in a given language (e.g. labels_tags_fr=bio)
 
-		if ($field =~ /^(.*)_tags(_(\w\w))?/) {
+		# Exception: attribute_unwanted_ingredients_tags=en:water is not a filter, it is a parameter to compute attributes
+
+		if (($field !~ /^attribute_/) and ($field =~ /^(.*)_tags(_(\w\w))?/)) {
 			my $tagtype = $1;
 			my $tag_lc = $lc;
 			if (defined $3) {
--- a/tests/integration/expected_test_results/search_v1/search-attribute-unwanted-ingredients-water-cookie.json
+++ b/tests/integration/expected_test_results/search_v1/search-attribute-unwanted-ingredients-water-cookie.json
@@ -1,0 +1,2075 @@
+{
+   "count" : 6,
+   "page" : "1",
+   "page_count" : 6,
+   "page_size" : 50,
+   "products" : [
+      {
+         "attribute_groups" : [
+            {
+               "attributes" : [
+                  {
+                     "description" : "",
+                     "description_short" : "Average nutritional quality",
+                     "grade" : "c",
+                     "icon_url" : "http://static.openfoodfacts.localhost/images/attributes/dist/nutriscore-c-new-en.svg",
+                     "id" : "nutriscore",
+                     "match" : 53.6666666666667,
+                     "name" : "Nutri-Score",
+                     "panel_id" : "nutriscore_2023",
+                     "status" : "known",
+                     "title" : "Nutri-Score C"
+                  },
+                  {
+                     "grade" : "unknown",
+                     "icon_url" : "http://static.openfoodfacts.localhost/images/attributes/dist/nutrient-level-salt-unknown.svg",
+                     "id" : "low_salt",
+                     "missing" : "Missing nutrition facts",
+                     "name" : "Salt",
+                     "panel_id" : "nutrition_facts_table",
+                     "status" : "unknown",
+                     "title" : "Salt in unknown quantity"
+                  },
+                  {
+                     "grade" : "unknown",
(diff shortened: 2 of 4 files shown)


monsieurtanuki:
@stephanegigandet Still not convinced.
https://fr.openfoodfacts.net/cgi/search.pl?search_terms=&code=6111035000058&langs=fr&page_size=5&page=1&fields=code%2Cbrands,ingredients&json=1&attribute_unwanted_ingredients_tags=en:mineral-water
```json
{
  "count": 1,
  "page": "1",
  "page_count": 1,
  "page_size": 5,
  "products": [
    {
      "brands": "Les Eaux Minérales d'oulmès,Sidi Ali",
      "code": "6111035000058",
      "ingredients": [
        {
          "ciqual_food_code": "18066",
          "ecobalyse_code": "36b3ffec-51e7-4e26-b1b5-7d52554e0aa6",
          "id": "en:mineral-water",
          "is_in_taxonomy": 1,
          "percent": 100,
          "percent_estimate": 100,
          "percent_max": 100,
          "percent_min": 100,
          "rank": 1,
          "text": "EAU MINÉRALE",
          "vegan": "yes",
          "vegetarian": "yes"
        }
      ]
    }
  ],
  "skip": 0
}
```
I would expect something visible when I ask for a mineral water without mineral water.
We've got the same result without the "no mineral water" parameter: https://fr.openfoodfacts.net/cgi/search.pl?search_terms=&code=6111035000058&langs=fr&page_size=5&page=1&fields=code%2Cbrands,ingredients&json=1

Not convinced by the feature as I understand it anyway, at least for Smoothie:
* if we wanted to flag a product as "with unwanted ingredients", we could already do it, from the ingredients field that we already use
* in Smoothie, imagine the user decides to add an unwanted ingredient - the local database would be populated with products stored with different "unwanted ingredients" values
* the server not returning products that have that ingredients would be fantastic instead - give me the top 10 soy milk without xanthan gum

stephanegigandet:
> @stephanegigandet Still not convinced. https://fr.openfoodfacts.net/cgi/search.pl?search_terms=&code=6111035000058&langs=fr&page_size=5&page=1&fields=code%2Cbrands,ingredients&json=1&attribute_unwanted_ingredients_tags=en:mineral-water

> I would expect something visible when I ask for a mineral water without mineral water. 

@monsieurtanuki That feature is about attributes that should be shown at the top of the product page, so you need to ask for the attributes too:
https://fr.openfoodfacts.net/cgi/search.pl?search_terms=&code=6111035000058&langs=fr&page_size=5&page=1&fields=attribute_groups,code%2Cbrands,ingredients&json=1&attribute_unwanted_ingredients_tags=en:mineral-water

Then you get this:

```
{
attributes: [
{
description: "Ingrédients indésirables: Eau minérale",
description_short: "Eau minérale",
grade: "e",
icon_url: "https://static.openfoodfacts.net/images/attributes/dist/contains-unwanted-ingredients.svg",
id: "unwanted_ingredients",
match: 0,
name: "Ingrédients indésirables",
panel_id: "ingredients",
status: "known",
title: "Contient des ingrédients indésirables"
}
],
id: "ingredients",
name: ""
},
```

> Not convinced by the feature as I understand it anyway, at least for Smoothie:
> 
> * if we wanted to flag a product as "with unwanted ingredients", we could already do it, from the ingredients field that we already use

The goal is to do it with the attributes, so that it is done in the similar way as the allergens, the nutriscore etc. without recoding new logic. It's also automatically taken into account in the compatiblity % score.

> * in Smoothie, imagine the user decides to add an unwanted ingredient - the local database would be populated with products stored with different "unwanted ingredients" values

The local database should be used only as a temporary fallback, we should refresh products when they are scanned or opened.

> * the server not returning products that have that ingredients would be fantastic instead - give me the top 10 soy milk without xanthan gum

That's a different feature, you can do it like this: https://world.openfoodfacts.org/api/v2/search?fields=code,product_name&categories_tags=en:soy-milk&ingredients_tags=-en:xanthan-gum

monsieurtanuki:
> @monsieurtanuki That feature is about attributes that should be shown at the top of the product page, so you need to ask for the attributes too

Understood.

> The goal is to do it with the attributes, so that it is done in the similar way as the allergens, the nutriscore etc. without recoding new logic. It's also automatically taken into account in the compatiblity % score.

Understood.
In Smoothie we compute the score from the `match` values.
In that case it's a 0/100 match value that we could process on top, in a degraded mode (e.g. no connection), right?

> The local database should be used only as a temporary fallback, we should refresh products when they are scanned or opened.

It's currently not the case (e.g. when "opened"). There are currently limitations regarding the number of queries.
Besides, it would be relatively painful to process, as it would mean refreshing all the local database from the server if we want to compare products.

> That's a different feature, you can do it like this: https://world.openfoodfacts.org/api/v2/search?fields=code,product_name&categories_tags=en:soy-milk&ingredients_tags=-en:xanthan-gum

Thank you, I didn't know that! Not sure how that could be useful in Smoothie.

PR #12625, opened 2025-11-04
taxonomy: add ingredients based on languages comparison

benbenben2:
### What

add ingredients based on languages comparison


stephanegigandet (approved):
Thank you!

PR #12627, opened 2025-11-04
fix: Correct HTML structure in barcode scanner modal

hangy:
### What

- Close `h2` correctly
- Move `span` inside header, as it was probably intended. `h2` cannot be nested into `span`

### Screenshot

N/A

N/A


stephanegigandet (approved):
Thank you!

PR #12630, opened 2025-11-05
taxonomy(beauty): merge “nail care” categories from OPF

Freso:
A number of “Nail Care” entries were added to the Products taxonomy when the Google products taxonomy was imported, several of which didn’t already exist in the Beauty taxonomy.

For categories which didn’t exist, I mostly just copied them over. For categories that already existed, I did my best to merge them.

Changes from Products taxonomy/before:

- I renamed “Nail Care” to “Nail products” since they’re not necessarily all _care_ products. I considered just calling it “Nails” to mirror the various other body part 1st level taxa we have, but went with including “products”. I did add “Nails” as a synonym though, so just shuffling them around should work if we decide on that at a later time.

- I changed cases to sentence case rather than Title Case, since that’s how we’re generally doing in the rest of the taxonomies.

- Where they weren’t already plural, I changed the categories to be plural, since that’s how we generally do category taxa elsewhere.

- For non-English/Scandinavian translations/synonyms, I made a best guess as to which was plural and/or “better”. I almost certainly didn’t make the best choice in every case. Please fix/improve or poke me to make a follow-up later. :)

- I changed `no:` entries to `nb:`.

### DIFF
--- a/taxonomies/beauty/categories.txt
+++ b/taxonomies/beauty/categories.txt
@@ -3010,21 +3010,22 @@ uz: Tozalovchi sut
 zh: 洁面乳
 
 < en:Cleansers
-en: Nail polish remover, Nail varnish remover
+< en:Nail products
+en: Nail polish removers, Nail polish remover, Nail varnish remover
 ar: مزيل طلاء الأظافر
 az: Dırnaq lakını təmizləyən vasitə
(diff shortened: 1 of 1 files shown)


teolemon (approved):
LGTM @Freso 👍

Issue #10989, opened 2024-11-07
OpenAPI definition doesn't reflect correctly some packaging fields returned by API


void0x23:
This issue is related to a coherence problem i've found in the openapi specification, in particular the packaging section, where we can see
that the elements are described with 2 fields, id and lc_name.

![Image](https://github.com/user-attachments/assets/30912c2f-2958-4455-bbae-a8d8138f3e2b)

but they seems to be returned as a single string

![Image](https://github.com/user-attachments/assets/9bec8c39-6ac7-4708-80f0-1ec5c7a554b6)

and this create problems during the automatic generation of classes in the code.

Should we change the definition to String instead ?

void0x23:
Also, in the api call to /api/v2/product/{barcode}?fields=knowledge_panels, the definition of the field action_element is String
![Image](https://github.com/user-attachments/assets/02d2e65b-f745-4459-8f64-2b2149ad3401)
but seems to be returned as an object
![Image](https://github.com/user-attachments/assets/9a19cb92-41a1-489a-9216-309cf6d60eff)
and this creates deserialization problems

stephanegigandet:
Hi @void0x23 
Sorry for my very late reply.

> This issue is related to a coherence problem i've found in the openapi specification, in particular the packaging section, where we can see that the elements are described with 2 fields, id and lc_name.

For the packaging, it's because we have a packagings section inside the environmental_score / ecoscore, which is different from the packagings section at the root of the product object. The one with id and lc_name is for the packagings at the root, while the other one is for the ecoscore packagings bonus/malus. I think the OpenAPI correctly reflects that (at least now, maybe not last year).

I will fix the issue you raised for the action_element.

Issue #8605, opened 2023-06-26
Add more parameters to Get a product by barcode API


alexgarel:
I think we at least needs to add lc and ll parameters.

Issue #7878, opened 2022-12-21
Transfer to OpenAPI : Get a list of products by barcodes  |  GET


teolemon:
### What
- Transfer to OpenAPI : Get a list of products by barcodes | GET
- https://openfoodfacts.github.io/api-documentation/#jump-3SEARCHRequests-Getalistofproductsbybarcodes
### Part of
- #7446

PR #12631, opened 2025-11-06
docs: improve openapi spec

stephanegigandet:
- Some refactoring to avoid duplicate specs in api.yaml and api-v3.yaml by moving definitions to the parameters directory
- Added missing parameters to api v2 get product, fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/8605
- Renamed {barcode} to {code} in the paths, to be consistent with the "code" field inside products
- Fixed type of action_element in knowledge panels. Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/10989
- Added code parameter for search queries. Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/7878

### DIFF
--- a/taxonomies/beauty/categories.txt
+++ b/taxonomies/beauty/categories.txt
@@ -3010,21 +3010,22 @@ uz: Tozalovchi sut
 zh: 洁面乳
 
 < en:Cleansers
-en: Nail polish remover, Nail varnish remover
+< en:Nail products
+en: Nail polish removers, Nail polish remover, Nail varnish remover
 ar: مزيل طلاء الأظافر
 az: Dırnaq lakını təmizləyən vasitə
 bg: Лакочистител
 bn: নেইল পলিশ রিমুভার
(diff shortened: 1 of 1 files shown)


docs/api/ref/api.yaml
```diff
@@ -142,7 +144,7 @@ paths:
         Open Food Facts uses optical character recognition (OCR) to retrieve nutritional data and other information from the product labels.
       parameters:
         - $ref: "#/components/parameters/id"
-        - $ref: "#/components/parameters/code"
+        - $ref: "./parameters/code.yaml#/components/parameters/Code"
```
teolemon, line 147: typo @stephanegigandet ?

stephanegigandet: no, all the other parameters are in PascalCase in the openapi spec

alexgarel: to clarify @teolemon this is the internal identifier which is PascalCase, the parameter per se is lowercase.

PR #12653, opened 2025-11-11
fix: Locale fixups and output sorting for Google Product taxonomy import

jayaddison:
### What

Addresses [an observation I'd reported](https://github.com/openfoodfacts/openfoodfacts-server/pull/12457#pullrequestreview-3435756033) in #12457, noting that although an `en-AU` Google taxonomy source file exists, it unexpectedly isn't yet in use when fetching that file for Australian taxonomy import.

Also adds sorting of the keys in the output JSON mappings, to improve determinism of the output and to make it easier to identify differences after the import is re-run.

- Follow-up to #12457


jayaddison:
There are a few differences between the contents of the two Google taxonomy files -- and not only differences in category description/translation.

For example: the `en-GB` file contains category ID `1471` for smoke detectors -- but that entry does not exist in the `en-AU` file.  However, both files do contain ID `499673` that can contain either smoke or carbon-monoxide detectors.

scripts/taxonomies/generators/01_fetch_google_product_taxonomy.py
```diff
@@ -24,7 +24,7 @@
     'fr-FR': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.fr-FR.txt',
     'de-DE': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.de-DE.txt',
     'es-ES': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.es-ES.txt',
-    'au-AU': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.en-GB.txt',
+    'au-AU': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.en-AU.txt',
```
stephanegigandet:
```suggestion
    'en-AU': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.en-AU.txt',
```

jayaddison: @stephanegigandet I did wonder about that.  But if that is the fixup here, then I think there may be other errors for other keys in this dictionary?  `at-AT`, for example -- perhaps should be `de-AT`?

scripts/taxonomies/generators/01_fetch_google_product_taxonomy.py
```diff
@@ -24,7 +24,7 @@
...
     'at-AT': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.de-DE.txt',
```
teolemon:
```suggestion
    'de-AT': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.de-DE.txt',
```

scripts/taxonomies/generators/01_fetch_google_product_taxonomy.py
```diff
@@ -24,7 +24,7 @@
...
     'be-FR': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.fr-FR.txt',
```
teolemon:
```suggestion
    'fr-BE': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.fr-FR.txt',
```

scripts/taxonomies/generators/01_fetch_google_product_taxonomy.py
```diff
@@ -24,7 +24,7 @@
...
     'be-NL': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.nl-NL.txt',
```
teolemon:
```suggestion
    'nl-BE': 'https://www.google.com/basepages/producttype/taxonomy-with-ids.nl-NL.txt',
```

teolemon (commented):
fix

jayaddison:
Thanks @teolemon - we have a few more errors too I think.  The language code for Danish (`da`) is not the same as the country code for Denmark (`dk`) but the key for that is currently `dk-DK` instead of `da-DK`.

Something seems wonky about the Swiss entries too.

jayaddison:
I should have some time tomorrow to review the keys for this dictionary in more detail.

teolemon:
Don't spend too much time on this since the taxonomy has been generated and we don't really support variants.

jayaddison:
@teolemon OK; will the script be used again in future?  If not, I'll close this.

teolemon:
@jayaddison it may, in case google massively updates its taxonomy, or adds new languages we didn't import

jayaddison:
Ok, thanks.  I think it's worth correcting the entries in that case - I'll ideally also re-run the script after the fixups to find out whether that produced an effect/diff on the results.

jayaddison:
I've applied what I believe are corrections to the dictionary keys (incuding the addition of an entry for Canadian English -- yes, it is true that they do speak it in addition to French); my next step (perhaps tomorrow) will be to re-run the import process and to compare the results.

jayaddison:
There seems to be some amount of nondeterminism in the mappings; running the `python3 00_run_all.py` script repeatedly produces some differing modifications/mappings with each run.

jayaddison:
Currently it's difficult to visually inspect some of the differences, partly because the `diff` output is across JSON files that do not have a sorted key order.  I'll look into updating the scripts to add sorting of keys -- perhaps as straightforward as adding `sort_keys=True` to some of the `json.dump` invocations -- to determine whether that helps.

jayaddison:
After adding sorting (08066ad9ec51b930cd5cecf3a6f8a1fd9f05b147) I have been able to re-run the Google Product taxonomy import process a few times and achieve a stable (e.g. deterministic / no changes to the JSON output) result.  This doesn't _guarantee_ that it'll remain stable, especially because we are importing external content from the network, but it seems more stable than without the sorting (where I found that the output changed on each invocation).

jayaddison:
> After adding sorting ([08066ad](https://github.com/openfoodfacts/openfoodfacts-server/commit/08066ad9ec51b930cd5cecf3a6f8a1fd9f05b147)) I have been able to re-run the Google Product taxonomy import process a few times and achieve a stable (e.g. deterministic / no changes to the JSON output) result. This doesn't _guarantee_ that it'll remain stable, especially because we are importing external content from the network, but it seems more stable than without the sorting (where I found that the output changed on each invocation).

Re-running the import script again today (+7 days from previous comment) from commit 08066ad9ec51b930cd5cecf3a6f8a1fd9f05b147 did not introduce any modifications to the data on the local filesystem.

jayaddison:
Thanks @teolemon!

PR #12654, opened 2025-11-11
taxonomy(beauty): more parity with OPF cosmetic categories

Freso:
Continuation of 64e4ed01411155965cc0445369f7cd3a373fcb4e ( https://github.com/openfoodfacts/openfoodfacts-server/pull/12630 ) for a couple more products that are currently in both OPF and OBF’s category taxonomies.

Most of the same considerations from 64e4ed01411155965cc0445369f7cd3a373fcb4e apply here as well.


stephanegigandet (approved):
Thanks

PR #12662, opened 2025-11-12
feat: Tools for indice de réparabilité + Folksonomy Engine

CharlesNepote:
Used to import ~1500 products: https://fr.openproductsfacts.org/facets/editeurs/org-database-fr-datagouv-indice-rep

### DIFF
--- a/scripts/imports/indice_reparabilite_france/mapping.csv
+++ b/scripts/imports/indice_reparabilite_france/mapping.csv
@@ -29,186 +29,186 @@
 "note_c5.2",,"-",,
 "note_c5.3",,"-",,
 "lien_documentation",,"link","Product Opener","Sometimes its leads to the documentation, sometimes to the producers' website"
-"nom_piece_detachee_1_reparateur","add_to_list","france:reparability_index:spare_part:name:repairer:fr",,"Nested values. Eg. bla|bli|blo"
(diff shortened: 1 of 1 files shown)


scripts/imports/folksonomy_engine/csv2folksonomyEngineTags.js
```diff
@@ -0,0 +1,1062 @@
...
+      port: urlObj.port || 443,
+      path: urlObj.pathname + urlObj.search,
+      method: options.method || 'GET',
+      headers: options.headers || {}
+    };
+    
+    // Log request options but mask Authorization header for security
+    const logOptions = { ...requestOptions };
+    if (logOptions.headers?.Authorization) {
+      logOptions.headers = { ...logOptions.headers, Authorization: 'Bearer ********' };
```
stephanegigandet, line 309: SonarCloud thinks **** is a password, but this function just strips it. Maybe passing '' instead would not trigger the warning.

alexgarel: @stephanegigandet we can tell sonar cloud it's a false positive.

alexgarel: Done (you can do this after loging in SonarCube with github account).

PR #12663, opened 2025-11-13
ci: remove `off-org` from deploy matrix

Freso:
Right now, every commit to `main` as well as every `v*.*.*` release will result in a red “x”/failed status, as the deploy workflow exits early for deploying to off.org.

This removes `off-org` from the build matrix entirely, to not run this. Hopefully this will lead us to not see all those red x’s all the time, only when there’s something _actually_ wrong. :)

Note that this _doesn’t_ simplify the workflow based on there only being one entry in the matrix, since this allows for a future easy reversion of this commit or otherwise re-insert `off-org` into the matrix.

### DIFF
--- a/.github/workflows/container-deploy.yml
+++ b/.github/workflows/container-deploy.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - main
       - deploy-*
-    tags:
-      - v*.*.*
 
 # Note on secrets used for connection
 # SSH_PRIVATE_KEY is the private key (shared between VM and host)
@@ -17,7 +15,7 @@ jobs:
     strategy:
       matrix:
         env:
-          - ${{ startsWith(github.ref, 'refs/tags/v') && 'off-org' || 'off-net' }}
+          - off-net
     environment: ${{ matrix.env }}
     concurrency: ${{ matrix.env }}
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/container-deploy.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         env:
-          - ${{ startsWith(github.ref, 'refs/tags/v') && 'off-org' || 'off-net' }}
+          - ${{ startsWith(github.ref, 'refs/tags/v') && 'off-net' }}
```
alexgarel: Nope, this is not the right way… this would break off-net deployment.

What you should do is simply remove the "tags" entry from the "on" above !

Of course, here you can simplify:
```suggestion
          - off-net
```

Freso: Right, because the statement is parsed as `(startsWith(github.ref, 'refs/tags/v') && 'off-org') || ('off-net')`, not `startsWith(github.ref, 'refs/tags/v') && ('off-org' || 'off-net')`

Freso: Should I still remove `tags: [v*.*.*]` with this fix applied?

alexgarel: Yes :-)

alexgarel: Otherwise we will deploy the release on .net, while it should not. 

Removing the `tags: [v*.*.*]` will avoid this workflow to run on releases at all.

Freso: Updated!

PR #12664, opened 2025-11-14
feat: rule to block edits on a field condition

alexgarel:
To avoid some users to select image we need to block edition based on a field matching a regexp.

- Part of: https://github.com/openfoodfacts/contributor-quality-issues/issues/18

### DIFF
--- a/lib/ProductOpener/Config_off.pm
+++ b/lib/ProductOpener/Config_off.pm
@@ -432,6 +432,7 @@ $options{users_who_can_upload_small_images} = {
 			["ignore_serving_size"],
 			# block image selection
 			["block_if_regexp_match_id", "^(front|ingredients|nutrition|packaging).*"],
+			["block_if_regexp_match_imagefield", "^(front|ingredients|nutrition|packaging).*"],
 		],
 	},
 );
--- a/cgi/product_image_upload.pl
+++ b/cgi/product_image_upload.pl
@@ -215,6 +215,15 @@ if ($imagefield) {
 		$log->info("product code already exists", {code => $code});
 	}
 
+	my $proceed_with_edit = process_product_edit_rules($product_ref);
+	$log->debug("edit rules processed", {proceed_with_edit => $proceed_with_edit}) if $log->is_debug();
+	if (not $proceed_with_edit) {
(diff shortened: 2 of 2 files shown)


alexgarel:
I still need to test it locally.

alexgarel:
Tested locally… works fine.

lib/ProductOpener/Config_off.pm
```diff
@@ -430,6 +430,8 @@ $options{users_who_can_upload_small_images} = {
 			["ignore_nutriment_alcohol"], ["ignore_nutriment_vitamin-d"],
 			["ignore_nutriment_calcium"], ["ignore_nutriment_potassium"],
 			["ignore_serving_size"],
+			# block image selection
+			["block_if_regexp_match_id", "^(front|ingredients|nutrition|packaging)_.*"],
```
stephanegigandet: It might still be possible to select an image by passing "front" without a language code.

```suggestion
			["block_if_regexp_match_id", "^(front|ingredients|nutrition|packaging).*"],
```

lib/ProductOpener/Config_off.pm
```diff
@@ -430,6 +430,8 @@ $options{users_who_can_upload_small_images} = {
 			["ignore_nutriment_alcohol"], ["ignore_nutriment_vitamin-d"],
 			["ignore_nutriment_calcium"], ["ignore_nutriment_potassium"],
 			["ignore_serving_size"],
+			# block image selection
```
stephanegigandet, line 433: The rule below will work to prevent image selection from product_image_crop.pl, but it won't prevent images from being uploaded and selected by passing image_type=front_fr to product_image_upload.pl

So possibly add a rule for image_type too, and we need to add process_product_edit_rules() to product image upload too.

alexgarel: Done

alexgarel:
tested with
```
curl -X POST -v "http://world.openfoodfacts.localhost/cgi/product_image_upload.pl" -F "user_id=municorn-calorie-counter-app" -F password=**** -F "code=0001234567890" -F imagefield="front"  -F imgupload_front=@/path/to/image.jpg  -H "User-Agent: Test alex"
```

PR #12665, opened 2025-11-14
fix: use aggregated set as input to Nutri-Score computation + other fixes and tests

stephanegigandet:
This PR:
- refactors a bit the normalization of input nutrient values and modifiers, and add a lot of unit tests for it
- makes more unit tests runnable (e.g. food.t was breaking)
- changes the NutriScore computation functions to use the new aggregated set

### DIFF
--- a/lib/ProductOpener/Food.pm
+++ b/lib/ProductOpener/Food.pm
@@ -72,9 +72,7 @@ BEGIN {
 		&compute_nutrient_levels
 		&evaluate_nutrient_level
 		&compute_units_of_alcohol
-		&compute_estimated_nutrients
 
-		&has_nutrition_data_for_product_type
 		&compare_nutriments
 
 		&extract_nutrition_from_image
@@ -336,11 +334,9 @@ It is a list of nutrients names with eventual prefixes and suffixes:
 			'iodine-', 'caffeine-',
 			'taurine-', 'methylsulfonylmethane-',
 			'ph-', 'fruits-vegetables-nuts-',
-			'fruits-vegetables-nuts-dried-', 'fruits-vegetables-nuts-estimate-',
-			'collagen-meat-protein-ratio-', 'cocoa-',
-			'chlorophyl-', 'carbon-footprint-',
-			'carbon-footprint-from-meat-or-fish-', 'nutrition-score-fr-',
-			'nutrition-score-uk-', 'glycemic-index-',
+			'fruits-vegetables-nuts-dried-', 'collagen-meat-protein-ratio-',
+			'cocoa-', 'chlorophyl-',
+			'carbon-footprint-', 'glycemic-index-',
 			'water-hardness-', 'choline-',
 			'phylloquinone-', 'beta-glucan-',
 			'inositol-', 'carnitine-',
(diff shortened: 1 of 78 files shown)


lib/ProductOpener/Food.pm
```diff
@@ -1487,42 +1393,36 @@ sub saturated_fat_ratio ($nutriments_ref, $prepared) {
...
 
 =cut
 
-sub saturated_fat_0_because_of_fat_0 ($nutriments_ref, $prepared) {
-	my $fat = $nutriments_ref->{"fat" . $prepared . "_100g"};
-	return ((!defined $nutriments_ref->{"saturated-fat" . $prepared . "_100g"}) && (defined $fat) && ($fat == 0));
+sub saturated_fat_0_because_of_fat_0 ($nutrition_ref) {
+	my $fat = deep_get($nutrition_ref, "aggregated_set", "nutrients", "fat", "value");
+	return (   (!defined deep_get($nutrition_ref, "aggregated_set", "nutrients", "saturated-fat", "value"))
+			&& (defined $fat)
```
Vic142: why do you create a variable $fat but not $saturated_fat ? Wouldn't it be easier to read if there was also a $saturated_fat variable ?

lib/ProductOpener/Food.pm
```diff
@@ -1487,42 +1393,36 @@ sub saturated_fat_ratio ($nutriments_ref, $prepared) {
...
+			&& ($fat == 0));
 }
```
Vic142, line 1399: why do you create a variable $fat but not $saturated_fat ? wouldn't it be easier to read if there was also a variable $saturated_fat ?

scripts/update_all_products.pl
```diff
@@ -1209,41 +1208,8 @@
...
-					}
-					# Reassign so that the energy-kcal field is recomputed
-					assign_nid_modifier_value_and_unit(
-						$product_ref,
-						"energy-kcal" . $product_type,
-						$product_ref->{nutriments}{"energy-kcal" . $product_type . "_modifier"},
-						$product_ref->{nutriments}{"energy-kcal" . $product_type . "_value"},
-						$product_ref->{nutriments}{"energy-kcal" . $product_type . "_unit"}
-					);
-				}
```
Vic142, line 1240: I don't understand why this part is removed

lib/ProductOpener/Food.pm
stephanegigandet: @vic142 That's a good point, I created only fat because that's the only one I needed multiple times, but it would be easier to read with $saturated_fat as well, I will change it.

scripts/update_all_products.pl
stephanegigandet: The update_all_products.pl script is used to repair products sometimes, but the lines above are not relevant anymore. The carbon footprint has been replaced by the Green-Score. And the part about energy-kcal was for an old bug. We won't have nutrition data like in $product_ref->{nutriments} so it won't work and it's not needed anymore.

lib/ProductOpener/Nutrition.pm
```diff
@@ -811,6 +815,130 @@ sub get_unit_options_for_nutrient ($nid) {
...
+	# Note: in regexps below, the first matching alternative is used, so the longest strings must be first
+	# e.g. maximum before max
+	my $inferior_regexp = "<|&lt;|inférieur|inferieur|inferior|inf|less than|less|menor|menos";
+	my $superior_regexp = ">|&gt;|greater|more than|more|superior|mayor|más";
+	my $approximately_regexp = "~|≈|about|environ|env|aprox|alrededor";
+	my $maximum_regexp = "<=|&lt;=|\N{U+2264}|maximum|maxi|max";
+	my $minimum_regexp = ">=|&gt;=|\N{U+2265}|minimum|mini|min";
+
+	# Normalize the modifier if defined
+	if (defined ${$modifier_ref}) {
```
alexgarel, line 979: good idea, it makes it more readable ! Separation of concern is a great practice :-)

alexgarel (approved):
Cool, thanks !

PR #12666, opened 2025-11-15
taxonomy: Updates for 7331642061009 and 7318690074175

Freso:
https://se.openfoodfacts.org/product/7331642061009/ren-bikarbonat-universal-kryddhuset
https://se.openfoodfacts.org/product/7318690074175/bakpulver-ica

I was adding the first, then noticed the second was wrongly categorised as a sodium bicarbonate, so that ended up getting caught up in these changes too.

Note that this breaks out “Baking powders” into its own category, as there are raising agents that are not “baking powder”, so this helps us be a bit more specific – as well as link “baking powder” to Wikidata and the Google product taxonomy. (There are also currently several products with category `en:baking-powder` that are not currently being caught: https://world.openfoodfacts.org/facets/categories/baking-powder )

### DIFF
--- a/taxonomies/additives.txt
+++ b/taxonomies/additives.txt
@@ -13689,9 +13689,6 @@ anses_additives_of_interest:en: yes
 e_number:en: 450
 wikidata:en: Q290828
 
-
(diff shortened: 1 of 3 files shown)


stephanegigandet (approved):
Thanks

PR #10570, opened 2024-07-19
feat: packaging adds Luxembourg

benbenben2:
### What
packaging codes adds Luxembourg

### Screenshot
BEFORE -> AFTER
Screenshot_20240719_172512

### Related issue(s) and discussion
Part of https://github.com/openfoodfacts/openfoodfacts-server/issues/338

More examples: #8921, #8958, #10264, #10318, #10351, #10388, #10485, #10533:
- lib/ProductOpener/Display.pm
add description (name, street, city) based on columns in the file or hardcoded
- lib/ProductOpener/PackagerCodes.pm
add country and suffix of the code
- scripts/update_packager_codes.pl
add code formatting ('country' 'code' 'suffix', for example if code does not already contain 'country' or 'suffix')
add the column name for the $code variable
- packager-codes/
add the csv file (mind the naming)
- scripts/packager-codes/
add your script
- update sto files
```
docker exec -it po_off-backend-1 bash
./scripts/update_packager_codes.pl
```

Fixes: #331

PR #9295, opened 2023-11-10
ci: CRON for packager codes

alexgarel:
to refresh automatically packager codes once a month.

PR #12668, opened 2025-11-16
feat: Rework Croatian packager codes extraction for CRON

benbenben2:
### What
Rework Croatian packager codes extraction for CRON

**Remarks:**
- common/ -> generic functions that can be used for all countries
- countries/\<country\>/ -> country specific functions (some could be move to common like generate_file_identifier(), let see in the future)
- packager_sources_config.json -> list of urls and files name to download, common to all countries 
- packager_text_replacements_config.json -> list of data cleaning, common to all countries
- geocode_cache_hr.db -> commit and persist cache db of conversion address to lat/lon, per country (to be able to replace 1 country easily if needed)
- ideally, all scripts should create a csv file with "code;name;street;city;postalcode;lat;lng" which should contain all necessary information (long term, it might allow to simplify update_packager_codes.pl and Display.pm)

Usefull links: 
- @alexgarel work in progress: https://github.com/openfoodfacts/openfoodfacts-server/pull/9295
- previous PRs for packaging: https://github.com/openfoodfacts/openfoodfacts-server/pull/10570

### DIFF
--- a/scripts/packager-codes/common/geocode.py
+++ b/scripts/packager-codes/common/geocode.py
@@ -80,6 +80,7 @@ def convert_address_to_lat_lng(debug: bool, country_name: str, country_code: str
     """
     Convert address from CSV row to latitude and longitude using Nominatim.
     
+    We use a persistent cache, through dbm to avoid resolving known addresses.
     Args:
         debug: Enable debug logging
         country_name: Full country name (e.g., "Croatia")


scripts/packager-codes/common/csv_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    
+    for pattern, replacement in replacements.items():
+        text = re.sub(pattern, replacement, text, flags=re.IGNORECASE)
+    
+    return text.strip()
+
+
+def get_data_rows(rows: List[List[str]], header_keywords: Optional[List[str]] = None) -> Tuple[List[List[str]], int]:
+    """
+    Extract data rows from CSV by finding header row using keywords.
```
alexgarel, line 57:
```suggestion
    Extract data rows from CSV by finding header row using keywords.
    
    The idea is that there might be other data above the header row,
    so it will skip those lines and the header line.
```

scripts/packager-codes/common/download.py
```diff
@@ -0,0 +1,182 @@
...
+            if not excel_link:
+                print(f"{country_name} - Error - Could not find Excel file matching keyword '{keyword}' in {url}.")
+                sys.exit(1)
+
+            current_filename = excel_link.split('/')[-1]
+            
+            if expected_file_name:
+                if current_filename == expected_file_name:
+                    print(f"{country_name} - Info - File '{current_filename}' already processed. No update needed.")
+                    sys.exit(0)
```
alexgarel: nitpick: It's a bit strange to have sys.exit directly.
Why aren't we returning (and the higher level function might sys.exit).

scripts/packager-codes/common/geocode.py
```diff
@@ -0,0 +1,176 @@
...
+    url = "https://nominatim.openstreetmap.org/search.php?"
+
+    if street:
+        url += f"street={street}&"
+    if city:
+        url += f"city={city}&"
+    if postalcode:
+        url += f"postalcode={postalcode}&"
+
+    url += f"country={country_name}&countrycodes={country_code}&format=jsonv2"
```
alexgarel: I would have kept a dict, and recompute this query to make no_results_update_query simpler.

scripts/packager-codes/common/download.py
```diff
@@ -0,0 +1,182 @@
...
+    # Restart 3 times in case of empty response to make sure it is not an issue on API-side
+    restart = True
+    i = 0
+    if debug:
+        print(f"{country_name} - Debug - Fetching from API: {url}")
+    while restart:
+        try:
+            response = requests.get(url, headers=HEADERS)
+            # 1 request per second (Nominatim usage policy) - increased to 2 seconds for safety
+            sleep(2)
```
alexgarel: This should be a parameter of the function (maybe with a default of 2).

scripts/packager-codes/common/geocode.py
```diff
@@ -0,0 +1,176 @@
...
+                iter_failures += 1
+                print(f"{country_name} - Warning - {code}: No results found (attempt {iter_failures}/3)")
+                if iter_failures <= 3:
+                    url = no_results_update_query(country_name, url, iter_failures, code)
+                    if debug:
+                        print(f"{country_name} - Debug - {code}: Retrying with modified URL: {url}")
+                else:
+                    # Fail if all attempts fail
+                    print(f"{country_name} - Error - {code}: Could not geocode after {iter_failures} attempts. Final URL was: {url}")
+                    sys.exit(1)
```
alexgarel: why sys.exit ? Would we abandon everything for a single address which is missing ?
Why not use an exception eventually ?

scripts/packager-codes/common/geocode.py
```diff
@@ -0,0 +1,176 @@
...
+
+    return [lat, lng]
+
+
+def geocode_csv(debug: bool, country_name: str, country_code: str, input_csv: str, output_csv: str):
+    """
+    Read preprocessed CSV and geocode all addresses.
```
alexgarel, line 147: can you specify what the CSV columns are ?

alexgarel:
```suggestion
    Read preprocessed CSV and geocode all addresses.
    
    proprocessed CSV contains: code, name, street, city, postalcode
```

scripts/packager-codes/common/download.py
```diff
@@ -0,0 +1,182 @@
...
+            return None
+        
+    except requests.exceptions.RequestException as e:
+        print(f"{country_name} - Error - Downloading file: {e}")
+        sys.exit(1)
+
+
+def cached_get(debug: bool, country_name: str, url: str, cache) -> list:
+    """
+    Get data from URL with caching support.
```
alexgarel, line 133:
```suggestion
    Get data from URL with caching support.
    
    This is notably used to query nominatim. 
    The cache can be a dbm database to enable persistent cache,
    allowing to only fetch new addresses as data source is updated.
```

scripts/packager-codes/common/geocode.py
```diff
@@ -0,0 +1,176 @@
...
+                simplified_city = city_value.split('-')[0]
+                url = re.sub(r"city=[^&]*", f"city={simplified_city}", url)
+        
+    return url
+
+
+def convert_address_to_lat_lng(debug: bool, country_name: str, country_code: str, row: list) -> list:
+    """
+    Convert address from CSV row to latitude and longitude using Nominatim.
+
```
alexgarel, line 82:
```suggestion
    Convert address from CSV row to latitude and longitude using Nominatim.
    
    We use a persistent cache, through dbm to avoid resolving known addresses.
```

scripts/packager-codes/countries/hr/main.py
```diff
@@ -0,0 +1,216 @@
...
+            print(f"{country_name} - Info - Updated configuration with new filename: {new_filename}")
+        
+        return True
+        
+    finally:
+        cleanup_temp_files(country_name, [excel_file, csv_raw_file, source_file])
+
+
+def main():
+    """Main function to process Croatian packager codes CSV file."""
```
alexgarel, line 176: Normally in Python this would be the right place to emit sys.exit.

Other methods can throw exception, and we decide here if we should exit or not.

scripts/packager-codes/geocode_cache_hr.db
alexgarel, line 1: I understand that as you have updates you will commit this file along.
It's quite a good idea, I just hope it won't grow too big ! Because each update will add the file size to our repo (the file being a binary).

I don't have a very nice alternative, apart from eventually sftp/scp on one of our server (I can setup an account for this).

alexgarel (commented):
Very good work @benbenben2 !

I made some suggestion but nothing mandatory.

scripts/packager-codes/geocode_cache_hr.db
benbenben2: Not sure either. We will see when doing France or Germany. Maybe a dedicated repo, and have only sto files here could be an option.

benbenben2:
Thanks @alexgarel. I did some rework to apply your suggestions. Let me know if it corresponds to what you had in mind.

PR #12687, opened 2025-11-18
fix: extract Yuka params from GET JSON body

stephanegigandet:
Yuka edits work again: https://fr.openfoodfacts.org/produit/3174064454790/compote-multifruits-sans-sucres-ajoutes

Yuka made some changes some months ago and sends us parameters in JSON format inside the body of a GET request.

- Will solve: https://github.com/openfoodfacts/contributor-quality-issues/issues/19

### DIFF
--- a/cgi/product_jqm_multilingual.pl
+++ b/cgi/product_jqm_multilingual.pl
@@ -75,7 +75,7 @@ use Log::Any qw($log);
 # Yuka sends a POSTDATA parameter in JSON:
 # "POSTDATA":"{\"code\":\"3270160874071\",\"lc\":\"fr\",\"cc\":\"FR\",\"user_id\":\"kiliweb\" [..]
 # This needs to be done before init_request() as the body contains user_id and password for authentication
-if ((user_agent() =~ /Symfony HttpClient/) and (request_method() eq 'GET') and (not param("code")) {
+if ((user_agent() =~ /Symfony HttpClient/) and (request_method() eq 'GET') and (not param("code"))) {
 
 	my $r = Apache2::RequestUtil->request();
 


cgi/product_jqm_multilingual.pl
```diff
@@ -67,10 +67,40 @@ =head1 DESCRIPTION
...
-use Encode;
+use Encode qw/decode encode /;
 use JSON::MaybeXS;
 use Log::Any qw($log);
 
+# hack to get params in body of GET requests from Yuka
+# Yuka sends a POSTDATA parameter in JSON:
+# "POSTDATA":"{\"code\":\"3270160874071\",\"lc\":\"fr\",\"cc\":\"FR\",\"user_id\":\"kiliweb\" [..]
+# This needs to be done before init_request() as the body contains user_id and password for authentication
+if (user_agent() =~ /Symfony HttpClient/) {
```
hangy: Unfortunately, this user agent is used by one of the most generic PHP clients in general: https://symfony.com/doc/current/http_client.html
This might mean that the change may work for Yuka, but could disrupt a lot of other PHP clients. I'd suggest trying to detect if the magic "POSTDATA" string exists in the request to see if this hack has to be applied

cgi/product_jqm_multilingual.pl
```diff
@@ -67,10 +67,40 @@ =head1 DESCRIPTION
...
+			$cnt = $r->read($content, 262144, $offset);
+			$offset += $cnt;
+		} while ($cnt == 262144);
+	}
+
+	my $postdata_params_ref = eval {JSON::MaybeXS->new->utf8->decode($content)};
+	if (defined $postdata_params_ref) {
+
+		foreach my $key (sort keys %$postdata_params_ref) {
+			param($key, encode('UTF-8', $postdata_params_ref->{$key}));
```
alexgarel, line 99: Does this work if $postdata_params_ref->{$key} is a list ?

cgi/product_jqm_multilingual.pl
alexgarel: @hangy I think the processing is not destructive and just add eventual params, so it should continue to work for other PHP client.
But we must definitively test !

That said @stephanegigandet we should check if it's a GET, and eventually check it has no code param:
```suggestion
if ((user_agent() =~ /Symfony HttpClient/) and (request_method() eq 'GET') and (not param("code")) {
```

cgi/product_jqm_multilingual.pl
stephanegigandet: No, but it's not a list ;) But the likelyhood that someone else would send a list as JSON in a GET request a bit remote, so it's ok if it fails I think...

cgi/product_jqm_multilingual.pl
stephanegigandet: Good point, we do want to check it's a GET, otherwise if Yuka changes to POST, we will read the body before CGI.pm and it will fail.

stephanegigandet: I'll check if the param() call is ok here, it will trigger some processing in CGI.pm, but since it's a GET request, it should not touch the body hopefully.

hangy: > I think the processing is not destructive and just add eventual params, so it should continue to work for other PHP client.

Maybe. I wasn't sure how the apache2request object handles this when I wrote the comment. Sometimes, libraries/frameworks only stream the incoming data once, unless you explicitly buffer it. In that case, it could be destructive. If Perl/CGI buffers by default, the probability of an issue is quite a bit lower

cgi/product_jqm_multilingual.pl
alexgarel: I mean things like _tags are list no ?
Oh… they are "comma" separated values I guess.

PR #12698, opened 2025-11-20
ci: GitHub Actions: migrate to dorny/paths-filter@v3 to allow simpler path filtering

jayaddison:
### What

I think that the GitHub Action provided by [`dorny/paths-filter`](https://github.com/dorny/paths-filter) seems simpler than `changed-files` - specifically the GitHub Action previously [provided by `tj-actions`](https://github.com/tj-actions/changed-files); a fork of which we are [currently using](https://github.com/openfoodfacts/openfoodfacts-server/blob/0f2e5560b33bd89e011de6aaa351247370bd145b/.github/workflows/pull_request.yml#L33-L35) as [provided by `step-security`](https://github.com/step-security/changed-files).

We encountered an obstacle when trying to use `dorny/paths-filter` previously; it didn't seem to support file exclusions as advertised.  But in fact I am confident that the problem was that we need to upgrade to `v3` of the action to enable that.

In addition, I think that we can remove the use of a shell script if we upgrade to version three of `dorny/paths-filter`.  Fewer ecosystems/moving-parts generally makes me feel more confident that a software process (in this case, a continuous integration workflow) is reliable -- because it means fewer parts to inspect, and we can more easily inspect and replace a smaller number of parts.  It also requires fewer mental context-switches and engineering areas of knowledge to maintain.

However: I am not 100% confident that this is a better path to follow.  It could be controversial because what we have is already implemented and working.

cc @JagjeevanAK 

- Provides an alternative, maybe simpler, path-filtering mechanism for #11624.
- Relates to #11630.
- Relates to #11644.

Edit: update checklist (hidden within markdown comment block).

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -27,7 +27,7 @@ jobs:
     outputs:
       code_modified: ${{ steps.filter.outputs.code }}
     steps:
-      - uses: dorny/paths-filter@v3
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
         id: filter
         with:
           # FIXME: we need to consider more than these files


jayaddison:
NB: The most important part of this changeset, and/or the easiest to review, I think, is commit with ID: 358dbc2aee500fe02c87d492597edbee203bbacb

stephanegigandet:
Looks good to me. @JagjeevanAK as you made the previous change, could you review? Thank you!

JagjeevanAK:
Hey @stephanegigandet looking into it

JagjeevanAK:
Hey @stephanegigandet, @jayaddison is correct here I had some discussion with him on this topic prior and that time there were some issues but no they are solved mostly

jayaddison:
Note: I'm sorry about this - this branch/PR has failed in GitHub Actions CI after merge as 2f123b1b103c2783a3edf63344eb919d139bd3a0: https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/19801127288/job/56728559065

I think we should revert it until we can determine the problem; I have begun investigation in openculinary/openfoodfacts-server#7 to attempt that, but it may take some time.

jayaddison:
> Note: I'm sorry about this - this branch/PR has failed in GitHub Actions CI after merge as [2f123b1](https://github.com/openfoodfacts/openfoodfacts-server/commit/2f123b1b103c2783a3edf63344eb919d139bd3a0): https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/19801127288/job/56728559065
> 
> I think we should revert it until we can determine the problem; I have begun investigation in [openculinary#7](https://github.com/openculinary/openfoodfacts-server/pull/7) to attempt that, but it may take some time.

Slightly more info: it seems to work fine on pull requests, but not on pushes to the `main` branch.

jayaddison:
I think the problem on `main` is that although a `pull_request` event on GitHub _does_ perform a clone/checkout of the `git` repository -- but a `push` event does not.  The answer for dealing with that in [`dorny/paths-filter` is hinted at in the README](https://github.com/dorny/paths-filter/blob/de90cc6fb38fc0963ad72b210f1f284cd68cea36/README.md?plain=1#L29): we must ensure there is a checkout.

And we can do that by adding `actions/checkout` -- similar to one of the example README entries (although I note that `v6` now exists, and I would recommend that we should start simple, and not use the `with:` options unless we need them), from here: https://github.com/dorny/paths-filter/blob/de90cc6fb38fc0963ad72b210f1f284cd68cea36/README.md?plain=1#L348

I will attempt that fix in the [`openculinary` fork of this repository](https://github.com/openculinary/openfoodfacts-server/) and will only suggest it here if it resolves the problem.

edit: fix hyperlink to forked repo.

jayaddison:
I know I have written a lot of comments/text here; my attempt at a short summary is that we have a choice either to:

 - Revert with #12726 
OR
 - Roll-forward with a fix from #12727 

The largest mistake I made here was that I did not test the merge to `main` in a fork on GitHub.  Doing so would have replicated the issue before we encountered it upstream here, and reduced the need to scramble and attempt to explain the situation.

PR #12714, opened 2025-11-27
fix: remove some dead code - remove JQM support for old Cordova app

stephanegigandet:
Removing some dead code:
- support for jquerymobile, used in the Cordova app
- unused code and templates for product page that have been replaced with knowledge panels

### DIFF
--- a/tests/integration/expected_test_results/web_html/fr-search-form.html
+++ b/tests/integration/expected_test_results/web_html/fr-search-form.html
@@ -2139,7 +2139,7 @@ sur les deux axes pour obtenir un nuage de produits.</p>
     </form>
 </div>
 
-<!-- end templates/web/pages/search_form/search_form.tt.html --><p>file error - web/pages/product/includes/environmental_score_details.tt.html: not found</p>
+<!-- end templates/web/pages/search_form/search_form.tt.html --><p></p>
 								</div>
 							</div>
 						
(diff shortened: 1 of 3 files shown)


alexgarel (approved):
That's so cool !

image

hangy:
> Removing some dead code:
> 
>     * support for jquerymobile, used in the Cordova app

Related: The [cordova app](https://github.com/openfoodfacts/openfoodfacts-cordova-app) should probably be marked as obsolete (potentially archived). We still get some dependabot package upgrades and GitHub action runs for that repo.

Also, this should potentially be marked as a [breaking change](https://github.com/googleapis/release-please?tab=readme-ov-file#how-should-i-write-my-commits) in the changelog, because consumers might be using the API endpoints

stephanegigandet:
I looked in the logs:
- on OFF, we do not have jqm requests at all currently
- on OBF, we had 101 jqm requests from 63 IPs in the last 15 days

stephanegigandet:
@hangy : the jqm queries were never documented as it's something we used only for our own Cordova app, so I'm not sure we need to bump the version number, nobody else should be using them.

stephanegigandet:
@hangy we should archive the cordova repo indeed, I'm adding a deprecation notice first: https://github.com/openfoodfacts/openfoodfacts-cordova-app/pull/195

PR #12719, opened 2025-11-29
taxonomy: brand Lindt

moon-rabbitOFF:


hangy (approved):

taxonomies/brands.txt
```diff
@@ -386,6 +389,9 @@ wikidata:en: Q688424
 
 xx: Maître CoQ
 
+xx: Netto, Netto MD, Netto Marken-Discount
```
hangy: Is it used explicitly as a food/product brand, not just a store?

moon-rabbitOFF: I'm realizing now that there are a French, a Danish and a German Netto that are independant. I will remove this entry for now. Why would it be an issue to enter the store name that owns the brand of a particular product? Some stores are regularly switching some of their products from one of their brand to an other or changing the name of their brand (I have noticed that in Lidl and Aldi particularly). So it is convenient to have the name of the store to find a product when the brand name changed. For instance, Aldi is currently removing its brand "Gut Bio" for a simple "Bio" on the packaging (https://world.openfoodfacts.org/product/4061459700275/bio-bergbauernkase-nussig-aldi). If someone who bought this product a few months ago would try to find this product with Gut Bio, this person will never find it if it was not linked to Aldi too.

PR #12727, opened 2025-12-01
ci: pull_requests.yml: ensure a checkout exists

jayaddison:
### What

The `push` event in GitHub Actions does not implicitly create a clone of the repository, unlike the `pull_request` event.

Since the `dorny/paths-filter` action we use to handle both event types does require a checkout, this commit adds an explicit checkout step to the job.

- Relates to https://github.com/openfoodfacts/openfoodfacts-server/pull/12698#issuecomment-3593216264
- This is a rollforward alternative to the rollback/revert suggested by #12726.
- Fix has been tested in GitHub Actions CI using a fork at `openculinary/openfoodfacts-server`: https://github.com/openculinary/openfoodfacts-server/pull/8

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -27,6 +27,7 @@ jobs:
     outputs:
       code_modified: ${{ steps.filter.outputs.code }}
     steps:
+      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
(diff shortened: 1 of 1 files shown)


JagjeevanAK:
I think we can move forward with the new clean setup though previous one was persistent and reliable we can revert it back when we need that

JagjeevanAK:
I think we can move forward with the new clean setup though previous one was persistent and reliable we can revert it back when we need that

jayaddison:
Thank you, @JagjeevanAK.

PR #12735, opened 2025-12-01
fix: rename “Suffering Fingerprint” to “Suffering Footprint” across locale files

### LINKED ISSUE #12690: Rename Suffering Fingerprint to Suffering Footprint
### Task
Rename Suffering Fingerprint to Suffering Footprint

ShrutiPandey-4:
### What
This pull request renames all occurrences of **“Suffering Fingerprint”** to **“Suffering Footprint”** across all locale `.po` files.

### Why
The term *“Suffering Fingerprint”* was identified as incorrect terminology and needs to be replaced consistently throughout all translations.

### Notes
- This PR contains **only string replacements** in locale files.
- No taxonomy changes.
- No nutrition fallback unit code.
- This is a clean PR as requested by maintainers.

### Related
Replaces the previous PR that had unrelated commits.
- Fixes #12690


taxonomies/countries.txt
```diff
@@ -11246,7 +11246,7 @@ country_code_3:en: CIV
 language_codes:en: fr
 osm_relation:en: 192779
 
-en: Democratic Republic of the Congo, DRC, DR Congo, Congo-Kinshasa, Zaire, Congo, Democratic Republic of Congo, CD, COD, Congo DR, Congo DRC
+en: Democratic Republic of the Congo, DRC, DR Congo, Congo-Kinshasa, Zaire  , Congo, Democratic Republic of Congo, CD, COD, Congo DR, Congo DRC
```
jayaddison, line 11249: I think that our Perl code _does_ remove spaces from the items in our taxonomies - so maybe the change from `Zaire` to `Zaire  ` (with two spaces) does not have an effect, but please hold on while I check that.

I think it happens here: https://github.com/openfoodfacts/openfoodfacts-server/blob/95972083a6dae27b8542aacddcdc8fe6ad22e15f/lib/ProductOpener/Store.pm#L163

(a complicated-looking line of code initially -- but in fact it is mostly a list of symbols to replace with a `-` character instead)

PR #12746, opened 2025-12-01
test: test mongo-8 dependency

alexgarel:
Runs test with Mongo-8.
No need to merge this PR, it's for running tests only.

Part of: https://github.com/openfoodfacts/openfoodfacts-shared-services/pull/21


alexgarel:
Closing as [we decided not to go for mongo 8 for now](https://github.com/openfoodfacts/openfoodfacts-server/discussions/12708)

But it's interesting to see that all test where passing !

PR #12752, opened 2025-12-02
fix: edit history for new nutrition schema

stephanegigandet:
This PR makes the computation of edit history work with the new nutrition schema.
When computing the diffs, we now upgrade each version to the newest schema.

The history looks at diffs for all input nutrient sets (except the one that may be estimated from ingredients).

It is a bit wordy:

image

Also fixes handling of no_nutrition_data (old) / no_nutrition_data_on_packaging (new)
+ add tests for the web edit form
+ somes fixes for data completion

### DIFF
--- a/lib/ProductOpener/Nutrition.pm
+++ b/lib/ProductOpener/Nutrition.pm
@@ -1184,23 +1184,23 @@ sub assign_no_nutrition_data_from_request_parameters ($request_ref, $product_ref
 	# The old nutrition_data and nutrition_data_prepared checkboxes were only used to show/hide the nutrition data sections on the web form
 	# They do not map to any field in the new product structure
 
-	foreach my $checkbox ("no_nutrition_data") {
+	my $checkbox_value = request_param($request_ref, "no_nutrition_data");
+	if (defined $checkbox_value) {
 
-		if (defined request_param($request_ref, $checkbox)) {
-			my $checkbox_value = request_param($request_ref, $checkbox);
-			if (($checkbox_value eq '1') or ($checkbox_value eq "on")) {
-				$product_ref->{nutrition}{no_nutrition_data_on_packaging} = true;
-			}
-			elsif (deep_exists($product_ref, qw/nutrition no_nutrition_data_on_packaging/)) {
-				delete $product_ref->{nutrition}{no_nutrition_data_on_packaging};
-			}
+		if (($checkbox_value eq '1') or ($checkbox_value eq "on")) {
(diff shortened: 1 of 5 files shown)


lib/ProductOpener/Nutrition.pm
```diff
@@ -2398,4 +2400,100 @@ sub convert_salt_to_sodium ($salt_value) {
...
+	}
+
+	return 0;
+}
+
+=head2 get_nutrition_data_as_key_values_pairs ( $product_ref )
+
+This function is used by ProductOpener::Product::complete_product_history_and_completeness()
+
+It serialize the nutrition data into key-value pairs for easier comparison of product history entries.
```
alexgarel:
```suggestion
It serialize the nutrition data into key-value pairs (flat hashmap) for easier comparison of product history entries.

Estimates are not returned.
```

alexgarel:
(canceled) /update_tests_results

tests/integration/expected_test_results/import_csv_file/new_nutrition/5000000000001.json
alexgarel, line 1: is it because csv import of nutrition is still broken ?

tests/integration/expected_test_results/api_v2_product_read/get-with-blame.json
alexgarel, line 1: looks good !

lib/ProductOpener/Nutrition.pm
```diff
@@ -1162,6 +1166,43 @@ sub assign_nutrient_modifier_value_string_and_unit ($input_sets_hash_ref, $sourc
...
+	#
+	# Web:
+	# - check: no_nutrition_data is passed "on"
+	# - uncheck: no_nutrition_data is not sent but no_nutrition_data_displayed is sent
+
+	# We use only the no_nutrition_data checkbox to map it to no_nutrition_data_on_packaging
+	# The old nutrition_data and nutrition_data_prepared checkboxes were only used to show/hide the nutrition data sections on the web form
+	# They do not map to any field in the new product structure
+
+	foreach my $checkbox ("no_nutrition_data") {
```
alexgarel: why a foreach on one value only ? (maybe add a comment why)

alexgarel (approved):
Thanks a lot !

lib/ProductOpener/Nutrition.pm
stephanegigandet: I'll remove it

PR #12753, opened 2025-12-02
feat: added / changed / deleted values in edit history

stephanegigandet:
PR suggested by @teolemon to add more details about what was changed in the product history.

I only list values when they are small (e.g. it doesn't list the full ingredients list unless it's very small)

image


teolemon:
It's really cool 👍
Any chance you could print the computed Nutri-Score if it changes ?

john-gom:
@stephanegigandet , please can we update the [event schema](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/docs/events/messages/product_updated.yaml) to reflect the new diffs structure?

john-gom (changes requested):
I'm not sure about this change. We are changing what is currently a well structured model, a list of field names, into space-delimited unstructured data. I think it would be better to store before and after values in a structured way too and not corrupt the existing list of changed property ids.

@raphael0202 , you mentioned that you parse the diffs in one of your projects, could this change affect you?

raphael0202:
@john-gom Yes, I use the `diff` in Robotoff to see whether we need to predict again the category (when the fields passed as inputs to the classifier change).
I'm not sure to understand the schema difference, do we break compatibility?

john-gom:
> @john-gom Yes, I use the `diff` in Robotoff to see whether we need to predict again the category (when the fields passed as inputs to the classifier change). I'm not sure to understand the schema difference, do we break compatibility?

Hi @raphael0202 yes I think it would break compatibility as the field name would now become a space delimited string unless I have misunderstood the change @stephanegigandet

stephanegigandet:
@john-gom that's right, I'm abusing the field name to show the values. I forgot that the diffs are used in events, my bad. Thanks for spotting this!

stephanegigandet:
@john-gom In order not to break compatibility, I can leave the diffs hash of arrays as it was, and create a new changed_values field:

$change_ref->{changed_values} = { old => { product_name_en => 'old product name', ... }, new => { product_name_en => 'new_product_name' } ... 

And we could decide not to pass it to the redis event (or to pass it if we think it's useful).

john-gom:
I think we could break compatibility as long as it is coordinated with Robotoff. My main concern was keeping some structure to the data. We may also need to be mindful of not making too verbose a structure.

One simple "hack" I considered was to change the list of IDs to an array of arrays where the first element of the child array is the property id and subsequent elements are changed, added, deleted values. That could also cater for changing tags where multiple values are added / removed.

E.g. diffs->add->[["product_nane_en","Vegan Nutella"],["categories_tags","nut_spreads","vegan_spreads"]]

If the new value is too long it could be truncated or omitted. In @raphael0202 's code he could look at the value and if it is a sting use that as the property name but if it is an array use the first element.

I'm thinking for changes we only show the new value to keep it simple as can always refer to the previous product revision if more detail is needed.

john-gom:
Also with an array the UI is simple as you just need to join the array with spaces

raphael0202:
I like @john-gom proposal of using nested lists to keep the diff structured while being not verbose.
I would also suggest adding a `schema_version` field to be able to spot easily this kind of schema changes now and in the future.

teolemon:

PR #12768, opened 2025-12-04
fix: data quality checks for new nutrition schema

stephanegigandet:
Updated the data quality checks and the corresponding tests.

In practice, we will run the nutrition checks on all input sets (including the estimate from ingredients).

For simplicity, at this time, I did not create separate data quality tags for each source, preparation and per.
i.e. if one of the set has energy in kj not matching energy in kcal, we will have only 1 tag en:energy-value-in-kj-does-not-match-value-computed-from-other-nutrients

The alternative would be creating tags like en:nutrition-[packaging|manufacturer|usda|estimate|...]-[as-sold|prepared]-[100g|100ml|serving]-energy-value-in-kj-does-not-match-value-computed-from-other-nutrients but there would be lots of combinations.


alexgarel:
> For simplicity, at this time, I did not create separate data quality tags for each source, preparation and per.
i.e. if one of the set has energy in kj not matching energy in kcal, we will have only 1 tag en:energy-value-in-kj-does-not-match-value-computed-from-other-nutrients


@stephanegigandet it seems to me it will make the data quality team work really hard.
Shouldn't we at least distinguish errors present in the "preferred" nutrition set from the rest (and I would put those other errors as warning rather than errors).

cc @CharlesNepote

stephanegigandet:
> > For simplicity, at this time, I did not create separate data quality tags for each source, preparation and per.
> > i.e. if one of the set has energy in kj not matching energy in kcal, we will have only 1 tag en:energy-value-in-kj-does-not-match-value-computed-from-other-nutrients
> 
> @stephanegigandet it seems to me it will make the data quality team work really hard. Shouldn't we at least distinguish errors present in the "preferred" nutrition set from the rest (and I would put those other errors as warning rather than errors).
> 
> cc @CharlesNepote

I can generate different errors for each set (and make them warnings for the input set), and generate errors for the aggregated set.

Initially I didn't do that because I didn't want too many different ids for the data quality warnings / errors: we currently have translations in the taxonomy, it's going to be quite difficult to manage if we have 10x ids... (but maybe translating the errors and warnings is not such a good idea anyway...)

stephanegigandet:
Working on a new version with different warnings / errors for each input set and for the aggregated set.

jayaddison:
A concern/suggestion about the identifiers for the aggregated errors in particular: could we keep their string values the same as the existing, single-source error identifiers?

The reason I'm thinking/asking about that is: it could make it difficult/annoying to perform historical analysis of individual error rules if we always need to remember to introduce a time-based adjustment to the labels for them.

Alternatively, we could reprocess historic data to add the aggregated prefix.

stephanegigandet:
@jayaddison I changed it so that the aggregated tags are the same as current, with the exception that I added "nutrition - " in front of the energy related errors, for consistency and grouping.

jayaddison:
Ok, thank you @stephanegigandet.  I'm doing a quick audit of the [Mirabelle `distri-qual.sh` script](https://github.com/openfoodfacts/openfoodfacts-infrastructure/blob/97b12831d29a94ddc9c1446019edb9a660b040d5/scripts/mirabelle/distri-qual.sh) to check the downstream usage of the tags.

The [`recipe-estimator-metrics`](https://github.com/openfoodfacts/recipe-estimator-metrics) and [`off-product-environmental-impact`](https://github.com/openfoodfacts/off-product-environmental-impact) projects also reference some tags by value (found via codesearch for some example tags).

jayaddison:
I'm a little busy for the rest of this week, but I will attempt to resume the downstream effect review of the tags tomorrow.

jayaddison:
Self-quoting:
> I'm a little busy for the rest of this week, but I will attempt to resume the downstream effect review of the tags tomorrow.

I didn't get around to the downstream-effect review last week; I hope to find time for it within the next day.

jayaddison:
Partial progress on downstream-effect review:

**[openfoodfacts-infrastructure](https://github.com/openfoodfacts/openfoodfacts-infrastructure)**

  * `data-quality.sh`: this solely compares `data_quality_errors_tags` for non-empty values (e.g. [here](https://github.com/openfoodfacts/openfoodfacts-infrastructure/blob/855f7c6b7e3257098ce603517f75836c9954b83a/scripts/mirabelle/data-quality.sh#L131)).  So I anticipate no problems in this script; the additional prefixes will not affect the behaviour.
  * `products_daily_update.sh`: this seems fine; it [builds an index on the error tags column](https://github.com/openfoodfacts/openfoodfacts-infrastructure/blob/855f7c6b7e3257098ce603517f75836c9954b83a/scripts/mirabelle/products_daily_update.sh#L142).
  * `distri-qual.sh`
    * There is a maximum of one error logged per product, and this is keyed on the `code` column (barcode).  That behaviour should be unaffected by this change.
    * I think the formatting of the email reasons may be a bit unusual, if we do not prepare for the change.  We replace `-` with spaces and remove the `en:` prefix -- maybe we will also want to remove the additional dataset prefixes (cc @CharlesNepote).
    * TODO: review the rest of the code in this file

I'll continue/resume this review soon.

Edit: hyperlink fixup (`openfoodfacts-infrastructure`).

CharlesNepote:
> Partial progress on downstream-effect review:
> 

@jayaddison : thanks a lot for your review. Unfortunately, the version on Github is a bit old, my bad!! The production scripts are not linked to the ones on Github yet. I'm really sorry about that, this is due to my bad skills related to git. I'm going to see if I can update them. That said, I'm not sure that there have been structural changes.

jayaddison:
@CharlesNepote you're welcome!  And that's OK - please take your time.  The process/design of the script is beginning to make sense to me.  I had intended to complete the review of it and the remaining other files in this PR this week, but haven't found the time, and instead I'm going to pause for the holiday season.

CharlesNepote:
@jayaddison I have updated the git repo with production scripts. From now git repo's scripts should be the same as production scripts. Those scripts are a bit difficult to read, I'm afraid...

stephanegigandet:
@jayaddison I'm merging this PR (in the new-nutrition-update branch, not the main branch) as we need it to continue the code migration to the new nutrition format, but we can change the data quality tags later if needed. The new nutrition schema will change quite a lot of things, so it's expected that some downstream tools will need to be changed too (not only data quality but anything that uses nutrition data).

jayaddison:
@CharlesNepote @stephanegigandet thanks both.  I am not aware of any blockers and did find that the scripts are flexible to accept changing prefixes :+1:

PR #12769, opened 2025-12-04
taxonomy: Frozen topped baguettes

moon-rabbitOFF:
I created a new categor for this kind of products https://world.openfoodfacts.org/product/4001724004325/bistro-classique-baguette-a-la-provence-dr-oetker


teolemon:
@moon-rabbitOFF I've seen "loaded" being used as well for fries.

PR #12783, opened 2025-12-09
taxonomy(food): Add tahina as tahini xx-synonym

Freso:
Also cleans up other entries that are already covered by the `xx` entry.

Needed-for: https://se.openfoodfacts.org/product/7331217014010/traditionell-hummus-tahina-sevan


Freso:
I’m not sure what’s going wrong with the tests. It doesn’t seem like the failures I looked at are actually related to this PR?

PR #12785, opened 2025-12-11
fix: add email to be able to create test user

stephanegigandet:
All PRs are currently failing because a test user cannot be created anymore without an email address.

PR #12786, opened 2025-12-11
fix: replace gmail.com with example.com in tests

stephanegigandet:
As suggested by @hangy


tests/integration/protected_images.t
```diff
@@ -20,7 +20,8 @@ my $sample_products_images_path = dirname(__FILE__) . "/inputs/upload_images";
 
 # Create an owner
 my $owner_ua = new_client();
-my %create_user_argss = (%default_user_form, (name => 'sample-owner', userid => "sample-owner"));
+my %create_user_argss
+	= (%default_user_form, (name => 'sample-owner', userid => "sample-owner", email => 'sample-owner@gmail.com'));
```
Freso, line 24:
```suggestion
	= (%default_user_form, (name => 'sample-owner', userid => "sample-owner", email => 'sample-owner@example.com'));
```
?

tests/integration/expected_test_results/create_pro_user/user-after-subscription.json
```diff
@@ -1,24 +1 @@
...
-   "preferred_language" : "en",
-   "pro" : 1,
-   "registered_t" : "--ignore--",
-   "team_1" : "",
-   "team_2" : "",
-   "team_3" : "",
-   "user_sessions" : "--ignore--",
-   "userid" : "tests"
-}
+{}
```
hangy, line 1: This change seems weird? Should it stem from the mail change?

stephanegigandet: it's because there is a test@test.com email also that doesn't validate.... I'm now trying to fix all those in https://github.com/openfoodfacts/openfoodfacts-server/pull/12785

PR #12787, opened 2025-12-11
fix: simplified environmental panels

stephanegigandet:
This PR changes the simplified environmental panels to use "name" and "value" instead of "title" and "subtitle" (with a new "string" type to avoid adding a "%" to the value).

It also makes the value string shorter.

image


stephanegigandet:
{
elements: [
{
element_type: "panel",
panel_element: {
panel_id: "environment_label_fr:ab-agriculture-biologique"
}
},
{
element_type: "panel",
panel_element: {
panel_id: "environment_label_en:eu-organic"
}
}
],
evaluation: "good",
expanded: false,
half_width_on_mobile: true,
level: "info",
title_element: {
icon_color_from_evaluation: true,
icon_size: "small",
icon_url: "http://static.openfoodfacts.localhost/images/icons/dist/agriculture.svg",
name: "Environmental labels",
type: "string",
value: "High environmental benefits"
},
topics: [
"environment"
]
},

tests/integration/expected_test_results/api_v3_product_read/get-knowledge-panels-simplified.json
```diff
@@ -310,8 +310,9 @@
                "icon_color_from_evaluation" : true,
                "icon_size" : "small",
                "icon_url" : "http://static.openfoodfacts.localhost/images/icons/dist/agriculture.svg",
-               "subtitle" : "No labels with environmental benefits",
-               "title" : "Environmental labels"
+               "name" : "Environmental labels",
+               "type" : "string",
+               "value" : "No assessed labels"
```
alexgarel: Is this normal that this is in value and not "value_string" ?

tests/integration/expected_test_results/create_pro_user/user-after-subscription.json
alexgarel, line 1: woops ?

alexgarel (commented):
I'm validating but there are problems to solve I think, hope that update tests result will solve them !

tests/integration/expected_test_results/create_pro_user/user-after-subscription.json
stephanegigandet: yeah that's because of the test users email bug...

alexgarel (approved):
Cool !

PR #12789, opened 2025-12-13
fix: add fallback unit when nutriment unit is missing

### LINKED ISSUE #12619: Negative ingredient percent estimate
Reported by @moon-rabbitOFF : https://world.openfoodfacts.org/product/4000539033100/joghurt-heidelbeer-vanille-vollmilchschokolade-lindt

ShrutiPandey-4:
### What

This PR adds a fallback unit ('g') when a nutrition value exists but the
corresponding unit field is missing.  
Previously, such cases caused incomplete or confusing nutrition displays.

The logic now ensures:
- If '_value' is present AND no '_unit' is provided → default to '"g"'.
- No other parts of the file or taxonomy are touched.

This PR is intentionally clean and isolated, following the maintainers'
guidance to avoid unrelated commits.  
It only modifies the nutrition unit fallback logic in Display.pm.

### Why

Some products include a nutrition value without specifying a unit.
The UI then shows incomplete nutrition information.  
Adding a fallback unit improves display consistency and makes the data usable.

### Related issue
Fixes #12619


ShrutiPandey-4:
Hi @teolemon @Pierre,

Just checking in 

I have created a clean and isolated fix exactly as requested earlier:
— Only the nutrition fallback logic is included
— No unrelated changes or taxonomy edits
— Branch is up to date with main and ready for review

Please let me know if anything else needs to be adjusted.  
Happy to iterate quickly on any requested updates.

ShrutiPandey-4:
Hi @maintainers 
Just a small follow-up — this PR is a simple, isolated fix and all checks have passed.
Please let me know if you need any tweaks or additional context.
Happy to adjust anything required. Thanks!

jayaddison (changes requested):
I don't think it's OK to assume that the units are grams when that field is empty.

It might be correct in many cases -- maybe even the majority, but it could also be unsafe for some quantities, such as vitamins or minerals where milligrams or micrograms can be commonplace.

@ShrutiPandey-4 do you know how many products are affected by this?  Ideally I think we should try to either:

  * Validate data that arrives with us to prevent empty units.
  * In cases where the information is directly from a producer -- find out if we can report the problem back to them so that they can improve their data quality.

PR #12792, opened 2025-12-13
fix: rename Suffering Fingerprint to Suffering Footprint in locale files

ShrutiPandey-4:
This PR cleanly renames "Suffering Fingerprint" to "Suffering Footprint" across locale files.

-Only locale files are updated  
- No unrelated taxonomy, Perl, or display changes  
- Fresh branch created as requested to avoid noise in history  

This should fully address issue #12690.

### DIFF
--- a/po/common/en.po
+++ b/po/common/en.po
@@ -7636,8 +7636,8 @@ msgid "Suffering Footprint"
 msgstr "Suffering Footprint"
 
 msgctxt "external_sources_empreinte_souffrance_description"
-msgid "Indicator of animal suffering (animal welfare impact score)"
-msgstr "Indicator of animal suffering (animal welfare impact score)"
+msgid "external_sources_empreinte_souffrance_description"
+msgstr "external_sources_empreinte_souffrance_description"
 
 msgctxt "section_animal_welfare_title"
 msgid "Animal welfare"


ShrutiPandey-4:
Hi maintainers 
All requested changes have been made and this PR is now clean with only the intended modifications.

It looks like the workflows are waiting for maintainer approval.  
Please let me know if anything else is needed from my side. Thanks!

po/common/en.po
```diff
@@ -7632,12 +7632,12 @@ msgid "Those preferences are based on Open Food Facts' understanding of the ingr
...
-msgid "Suffering Fingerprint"
-msgstr "Suffering Fingerprint"
+msgid "Suffering Footprint"
+msgstr "Suffering Footprint"
 
 msgctxt "external_sources_empreinte_souffrance_description"
-msgid "Indicator of animal suffering computed by product"
-msgstr "Indicator of animal suffering computed by product"
+msgid "Indicator of animal suffering (animal welfare impact score)"
+msgstr "Indicator of animal suffering (animal welfare impact score)"
```
jayaddison: Are the changes to adjust the `external_sources_empreinte_souffrance_description` texts also wanted?  It's not mentioned in the issue or PR description, so I think we should pause to check that it's beneficial.

ShrutiPandey-4: Thanks for pointing this out!
The changes inside external_sources_empreinte_souffrance_description were not intentional — they happened because of a global search-and-replace.

You’re right, these adjustments were not part of the issue or PR scope.
I will revert those lines so the PR contains only the required renaming changes.

ShrutiPandey-4: Thanks for the catch!  
This change was added unintentionally during the global renaming.  
I have now reverted the 'external_sources_empreinte_souffrance_description' lines back to their original values.

Please let me know if anything else needs adjustment.

ShrutiPandey-4:
Hi @maintainers 
I’ve applied the requested adjustments (including the external_sources text cleanup).
This PR only contains the intended string-rename now and is clean to review.

Please let me know if anything else should be updated.
Thanks for your time!

jayaddison:
> Hi @maintainers I’ve applied the requested adjustments (including the external_sources text cleanup). This PR only contains the intended string-rename now and is clean to review.
> 
> Please let me know if anything else should be updated. Thanks for your time!

Hi @ShrutiPandey-4 - there are still some unrelated string modifications in this branch; please check again when you have time - thank you.

jayaddison:
Superseded by #13075.

PR #12807, opened 2025-12-17
fix: remove unused 'lol' LOLcat language code

jayaddison:
### What

Remove the `lol` ([LOLcat](https://en.wikipedia.org/wiki/Lolcat)) language from the codebase, because fun is disallowed (also / alternatively: it seems broken, and was related to an unrealised idea to introduce [in-context translation](https://crowdin.com/page/in-context-localization); instead of text translation strings, there are `crwdns` strings with numbered suffixes).

### Screenshot
image

- From Slack discussion at: https://openfoodfacts.slack.com/archives/C02LDQDDD/p1765980585825599


jayaddison:
NB: there is a `data-lol` attribute in `docs/events/openfoodfacts-server.html` -- that isn't lolcat-related, it's from an [`asyncapi-react`](https://github.com/asyncapi/asyncapi-react) web component that we use there.

jayaddison:
Thanks!

PR #12809, opened 2025-12-18
fix: treat unknown aluminium packaging shape as "light" (worst case)

### LINKED ISSUE #12696: Default weight for aluminium pack is heavy --> should be light
### What
in EnvironmentalScore.pm code suggests that default material for aluminium packaging should be [light](https://github.com/openfoodfacts/openfoodfacts-server/blob/0cc5d60bd0a072b306968c74a306d4f7080e8686/lib/ProductOpener/EnvironmentalScore.pm#L1707) (worse score) when actual behaviour is [heavy](https://github.com/openfoodfacts/openfoodfacts-server/blob/0cc5d60bd0a072b306968c74a306d4f7080e8686/lib/ProductOpener/EnvironmentalScore.pm#L1714) because the [condition](https://github.com/openfoodfacts/openfoodfacts-server/blob/0cc5d60bd0a072b306968c74a306d4f7080e8686/lib/ProductOpener/EnvironmentalScore.pm#L1709) to switch to heavy (empty shape) is flawed because of prior code that sets empty shape to [unknown](https://github.com/openfoodfacts/openfoodfacts-server/blob/0cc5d60bd0a072b306968c74a306d4f7080e8686/lib/ProductOpener/EnvironmentalScore.pm#L1689).

### Expected behavior
Default should be the worse case i.e. 'light'.
[the code](https://github.com/openfoodfacts/openfoodfacts-server/blob/0cc5d60bd0a072b306968c74a306d4f7080e8686/lib/ProductOpener/EnvironmentalScore.pm#L1709) should be adapted with something like ’if shape is defined and not “en:unknown” then etc...

arybhatt4533:
### What
**Problem**
Unknown aluminium shapes were defaulting to 'heavy' logic.
**Solution**
EnvironmentalScore.pm to skip appending the shape if it is en:unknow.
**Result**
this ensure the system falls back to the default 'light' (worst case) material ID.

### Screenshot
image

- Fixes #12696
-


OpenChris:
@arybhatt4533 see my comments in the Issue [here](https://github.com/openfoodfacts/openfoodfacts-server/issues/12696#issuecomment-3790143472).
@teolemon can you reply on the specification question [here](https://github.com/openfoodfacts/openfoodfacts-server/issues/12696#issuecomment-3790143472) as well?

teolemon:
- Dupe of https://github.com/openfoodfacts/openfoodfacts-server/pull/13031

PR #12831, opened 2025-12-27
taxonomy: translation czech ingredients

moon-rabbitOFF:
Added and corrected some czech translation. Corrected notably hazelnut paste which translate to "lískooříšková pasta" and not "oříšková pasta" (which means nut pasta).


stephanegigandet (commented):
Thank you

PR #12832, opened 2025-12-27
taxonomy: correction E476

moon-rabbitOFF:
Corrected the czech translation of E476 and reordered the French translations to have most common and the direct translation of the English version as the first synonym.


stephanegigandet (approved):
Thank you

PR #12850, opened 2026-01-05
Fix redundant product_not_found error after invalid barcode validation

dharsan-devaraj:

- Fixes #[ISSUE NUMBER]

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -953,13 +953,16 @@ sub check_nutrition_data_energy_computation ($product_ref) {
 				)
 				)
 			{
-				# Compare to specified energy value with a tolerance of 30% + an additiontal tolerance of 5
-				if (   ($computed_energy < ($specified_energy * 0.7 - 5))
-					or ($computed_energy > ($specified_energy * 1.3 + 5)))
+				# Compare to specified energy value with a tolerance
+				# Relax tolerance for kcal to avoid false positives due to rounding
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -953,13 +953,16 @@ sub check_nutrition_data_energy_computation ($product_ref) {
 				)
 				)
 			{
-				# Compare to specified energy value with a tolerance of 30% + an additiontal tolerance of 5
-				if (   ($computed_energy < ($specified_energy * 0.7 - 5))
-					or ($computed_energy > ($specified_energy * 1.3 + 5)))
+				# Compare to specified energy value with a tolerance
```
stephanegigandet, line 956: this code change seems unrelated

lib/ProductOpener/APIProductRead.pm
```diff
@@ -106,7 +106,7 @@ sub read_product_api ($request_ref) {
 				impact => {id => "failure"},
 			}
 		);
-		$response_ref->{result} = {id => "product_not_found"};
+		return;
```
stephanegigandet, line 109: Thank you for your PR.

We want to keep the response result, it's different from the error.

also instead of returning here, it would be best to the rest of the function logic (starting at if ((not defined $product_ref) or (not defined $product_ref->{code})) {    ) inside the else clause where we have a valid product code.

Finally, to verify that the code change works, please run the test update locally (make update_tests_results) and commit the resulting changed test results.

PR #12851, opened 2026-01-05
fix: correctly compute nutrient levels attributes with new nutrition …

stephanegigandet:
There were a number of issues that prevented the new nutrient levels to be computed correctly with the new nutrition schema:
- we were using the old nutriments structure to compute the attribute (we display the actual value, so the nutrient_levels_tags value is not enough)
- for search, we didn't ask for the schema_version, so products had their nutrition data upgraded even if they already had the new nutrition data, which led to empty nutrition data
- attributes.t test input data still used the old nutriments

image

### DIFF
--- a/scripts/get_product_from_mongodb.pl
+++ b/scripts/get_product_from_mongodb.pl
@@ -1,7 +1,6 @@
 #!/usr/bin/perl -w
 
-use Modern::Perl '2017';
-use utf8;
+use ProductOpener::PerlStandards;
 
 use ProductOpener::Config qw/:all/;
 use ProductOpener::Store qw/:all/;


scripts/get_product_from_mongodb.pl
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/perl -w
+
+use Modern::Perl '2017';
+use utf8;
```
alexgarel: nitpick: better always `use ProductOpener::PerlStandards;` everywhere…
```suggestion
use ProductOpener::PerlStandards;
```

tests/unit/expected_test_results/attributes/en-attributes.json
alexgarel, line 1: There are a lot of changes, I don't know really why (like on forest footprint)

tests/unit/expected_test_results/attributes/en-nutriscore.json
alexgarel, line 1: On this one I clearly see improvements compared to the "broken" base branch.

tests/unit/expected_test_results/attributes/en-attributes.json
stephanegigandet: It's a long chain reaction: in the previous attributes.t, the nutriments were in the old format, and not taken into account. In the new tests they are, and there's 25% sugars. But one of the ingredients is sugar, and that's one of the cases where I set the % max of that ingredient to 25% (instead of the 50% that was computed previously just from the order of ingredients). So in turn, when the algorithm picks the middle of the range, the ingredients percents are changed, and then we have changes to the forest footprint.

PR #12854, opened 2026-01-06
refactor: Nutrition data panel comparison column

Vic142:
This PR adds the comparison column of nutrient values of the products compared to the average of the products of their categories.

- In `lib/ProductOpener/Food.pm`, for the routine `compare_nutriments ($a_ref, $b_ref)`, I checked if `$a_ref` is a food or a petfood because I was not sure if there could be some other type of hash that could be compared and I didn't want to break the routine.
- Still for `compare_nutriments ($a_ref, $b_ref)`, since it wasn't clear for me if `b_ref` would always be the average values of the nutrients or if it could also be something else I chose not to update the average nutrient values structure, but I can do.
- **I haven't tried running the scripts to see if the comparison column actually shows on the web pages, which I will do as soon as I can.**
- If the average nutrient values hash for each product category is computed when running the script, it will probably not be correctly generated with the new nutrition schema so I will also look into this.

### DIFF
--- a/tests/unit/display.t
+++ b/tests/unit/display.t
@@ -353,7 +353,7 @@ my @display_tests = (
 		},
 		[
 			{
-				name => "en:Spreads",
+				name => "Spreads",
 				n => 1,
 				count => 100,
 				nutriments => {
--- a/tests/unit/expected_test_results/display/nutrition-facts-table-with-comparisons.json
+++ b/tests/unit/expected_test_results/display/nutrition-facts-table-with-comparisons.json
@@ -13,7 +13,7 @@
             {
                "col_id": "compare_0",
                "scope":"comparisons",
-				   "name": "Compared to: en:Spreads",
+				   "name": "Compared to: Spreads",
 				   "class": "compare_0"
             }
          ],
@@ -139,7 +139,7 @@
       {
          "col_id": "compare_0",
 		   "checked": 1,
-		   "name": "en:Spreads",
+		   "name": "Spreads",
 		   "link": "/facets/categories/en:spreads",
 		   "count": 100
 	   }


lib/ProductOpener/Food.pm
```diff
@@ -2289,7 +2305,49 @@ sub compare_nutriments ($a_ref, $b_ref) {
...
+			$nutriments{$nid} = $b_ref->{nutriments}{$nid};
+
+			if (    ($b_ref->{nutriments}{$nid} > 0)
+				and (defined $a_ref->{nutrition}{aggregated_set}{nutrients}{$nutrition_nid})
+				and (defined $a_ref->{nutrition}{aggregated_set}{nutrients}{$nutrition_nid}{value}))
+			{
+				# compute what percent the $a_ref value differs from the $b_ref value
+				# if the $a_ref value is given per serving, compute the value per 100g/ml before
+				my $a_value;
+				if ($a_ref->{nutrition}{aggregated_set}{per} eq "serving") {
```
stephanegigandet: In the spec we say that the aggregated_set is always for 100g or 100ml:  "aggregated_set (combined set of nutriments with the preferred sources, with normalized units, and for a normalized 100g or 100 ml quantity)"

lib/ProductOpener/Food.pm
```diff
@@ -2268,18 +2268,34 @@ sub compute_units_of_alcohol ($product_ref, $serving_size_in_ml) {
 sub compare_nutriments ($a_ref, $b_ref) {
 
 	# $a_ref can be a product, a category, ajr etc. -> needs {nutriments}{$nid} values
+	# $b_ref is the value references
+	my %nutriments = ();
+
+	if ($a_ref->{product_type} eq "food" || $a_ref->{product_type} eq "petfood") {
+		%nutriments = compare_nutrients_with_nutrition($a_ref, $b_ref);
+	}
+	else {
```
stephanegigandet: If it's not food or petfood, then we have no nutrition data.

lib/ProductOpener/Food.pm
```diff
@@ -2268,18 +2268,34 @@ sub compute_units_of_alcohol ($product_ref, $serving_size_in_ml) {
 sub compare_nutriments ($a_ref, $b_ref) {
```
stephanegigandet: Once a function has been converted to use the new nutrition format, we should rename "nutriment" to "nutrient" in function name and variables. That will help us see where we still have migration work to do.

stephanegigandet (approved):
Thanks a lot for the PR @Vic142 

It looks good. I made a few suggestions, I think we could merge it.

This PR handles the display part, we will also need to change the compute part so that the stats are computed on the aggregated data set.

Relevant code is in scripts/gen_top_tags_per_country.pl and in Display.pm in the functions add_product_nutriment_to_stats() and compute_stats_for_products()

In the 2nd PR, I think we should rename {nutriments} to {nutrients} in the stats format:

```
sub compute_stats_for_products ($stats_ref, $nutriments_ref, $count, $n, $min_products, $id) {

	#my $stats_ref        ->    where we will store the stats
	#my $nutriments_ref   ->    values for some nutriments
	#my $count            ->    total number of products (including products that have no values for the nutriments we are interested in)
	#my $n                ->    number of products with defined values for specified nutriments
	#my $min_products     ->    min number of products needed to compute stats
	#my $id               ->    id (e.g. category id)

	$stats_ref->{stats} = 1;
	$stats_ref->{nutriments} = {};
	$stats_ref->{id} = $id;
	$stats_ref->{count} = $count;
	$stats_ref->{n} = $n;
```

lib/ProductOpener/Food.pm
Vic142: ok thank you, I wasn't sure if it could be something else with the `nutriments` field. I think this also means that the function `compare_nutrients_with_nutriments` will not be used so I deleted it.

stephanegigandet (approved):
Thank you!

PR #12857, opened 2026-01-06
fix(api): return early with 400 on invalid barcode

### LINKED ISSUE #12848: Bug: Redundant "product_not_found" error returned after "invalid_code" check in API v3
### What
- When requesting a product with a structurally invalid barcode (e.g., `products/$` or containing special characters), the API v3 returns two contradictory errors simultaneously:
   1. `invalid_code` (Correct)
   2. `product_not_found` (Redundant and logically incorrect)
- The second error is triggered because the code falls through to the "product existence check" even after the barcode validation fails.

1. Open Postman or curl.
2. Make a GET request to: `https://world.openfoodfacts.org/api/v3/product/products/$`
3. Observe the `errors` array in the JSON response.
4. See that it contains both `invalid_code` and `product_not_found`.

### Expected behavior
- If the barcode validation fails (e.g. Regex check), the API should return immediately (Early Return) with a 400 Bad Request status.
- It should contain **only** the `invalid_code` error.
- It should not proceed to check if the product exists (which triggers the second error).

### Why
- **Developer Experience (DX):** It sends mixed signals to frontend clients (like the new SvelteKit Explorer), forcing them to write complex logic to prioritize errors.
- **Code Cleanliness:** The request processing should stop as soon as a fatal validation error is found.

Image

### Part of
- Related to openfoodfacts/openfoodfacts-explorer[#871](https://github.com/openfoodfacts/openfoodfacts-explorer/pull/871) (Where this behavior was observed and handled defensively)

### Additional context
- **Discovery:** I identified this issue while working on the frontend refactor (PR [#871](https://github.com/openfoodfacts/openfoodfacts-explorer/pull/871)) referenced above. The API returning duplicate errors made it difficult to determine the correct UI state (400 vs 404).
- **Technical Analysis:** I investigated `lib/ProductOpener/APIProductRead.pm` and found the potential cause.
- The `read_product_api` subroutine checks validation:
```perl
  if ($code !~ /^\d{4,24}$/) { ... }
```
- However, it lacks a return; statement inside that block, so execution falls through to:
```perl
if ((not defined $product_ref) ...) { ... }
```
- Since $product_ref is undefined (because the DB lookup was correctly skipped in the else block), the second error is added automatically.

### Type of device
REST-API

0 (API Logic Issue)

### Time per product
N/A

SoftSage07:
Fixes an issue where structurally invalid barcodes triggered both
`invalid_code` and `product_not_found`.

The API now returns early with a 400 Bad Request when barcode validation fails, preventing the product existence check and ensuring consistent error semantics for clients.

Fixes #12848


lib/ProductOpener/APIProductRead.pm
```diff
@@ -98,16 +98,20 @@ sub read_product_api ($request_ref) {
...
-			}
+				message => { id => "invalid_code" },
+				field   => { id => "code", value => $request_ref->{code} },
+				impact  => { id => "failure" },
+			},
+			400
 		);
-		$response_ref->{result} = {id => "product_not_found"};
+
+		return;
```
stephanegigandet, line 112: Thank you for your PR.

We want to keep the response result, it's different from the error.

also instead of returning here, it would be best to the rest of the function logic (starting at if ((not defined $product_ref) or (not defined $product_ref->{code})) {    ) inside the else clause where we have a valid product code.

Finally, to verify that the code change works, please run the test update locally (make update_tests_results) and commit the resulting changed test results.

Freso:
Will this still find [products that exist which have invalid barcodes](https://world.openfoodfacts.org/facets/codes/invalid-ean)? (Don’t mind that _most_ of the ones on that page are actually wrong and needs fixing; some of them are genuine products with non-validating barcodes.)

SoftSage07:
> Will this still find [products that exist which have invalid barcodes](https://world.openfoodfacts.org/facets/codes/invalid-ean)? (Don’t mind that _most_ of the ones on that page are actually wrong and needs fixing; some of them are genuine products with non-validating barcodes.)

Good point, I hadn’t considered existing products with non-validating barcodes.

Should this validation apply only on create/update, while still allowing reads for any barcode?
Or do you prefer expanding the regex to include these edge cases?

I want to avoid breaking legit products.

PR #12859, opened 2026-01-06
fix: return old nutrition schema fields in API v2

stephanegigandet:
Add support for ?fields=nutriments ?fields=nutriments.salt_100g etc.

+ new tests

### DIFF
--- a/tests/integration/expected_test_results/api_v2_product_write/get-product-nutrition-no_nutrition_data-empty.json
+++ b/tests/integration/expected_test_results/api_v2_product_write/get-product-nutrition-no_nutrition_data-empty.json
@@ -422,7 +422,6 @@
       ],
       "nutrient_levels" : {},
       "nutrient_levels_tags" : [],
-      "nutriments" : {},
       "nutriscore" : {
          "2021" : {
             "category_available" : 0,
--- a/tests/integration/expected_test_results/api_v2_product_write/get-product-nutrition-no_nutrition_data-on.json
+++ b/tests/integration/expected_test_results/api_v2_product_write/get-product-nutrition-no_nutrition_data-on.json
@@ -419,7 +419,6 @@
       ],
       "nutrient_levels" : {},
       "nutrient_levels_tags" : [],
-      "nutriments" : {},
       "nutriscore" : {
          "2021" : {
             "category_available" : 0,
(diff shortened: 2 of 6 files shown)


lib/ProductOpener/API.pm
```diff
@@ -693,6 +693,13 @@ sub api_compatibility_for_field ($field, $api_version) {
 		}
 	}
 
+	# Old nutrition schema with nutriments hash: we need to keep nutrition so that it can be converted back to nutriments
+	if ($api_version < 3.5) {
+		if ($field eq "nutriments") {
+			$field = "nutrition";
+		}
+	}
```
alexgarel: shan't we add a warning in the response about that use of a deprecated field (if not too hard to do) ?

lib/ProductOpener/API.pm
```diff
@@ -902,24 +912,19 @@ sub customize_response_for_product ($request_ref, $product_ref, $fields_comma_se
...
-					}
-					$customized_product_ref->{$return_hash}{$nutrient} = $product_ref->{nutriments}{$nutrient};
-				}
-				else {
-					$customized_product_ref->{$nutrient} = $product_ref->{nutriments}{$nutrient};
-				}
-			}
+			push @old_requested_nutrients, [$return_hash, $nutrient];
+			# Make sure we keep the nutrition field so that it can then be converted to the old nutriments hash
+			$customized_product_ref->{nutrition} = $product_ref->{nutrition};
```
alexgarel, line 927: there again might be great to add a warning to the response if we are in API v3 or disable it all together in API v3. Maintaining this kind of complex path is a bit tedious as time goes on.

lib/ProductOpener/API.pm
```diff
@@ -994,6 +999,28 @@ sub customize_response_for_product ($request_ref, $product_ref, $fields_comma_se
...
+					# return in nutriments / nutrients hash
+					deep_set($customized_product_ref, $return_hash, $nutrient, $full_nutriments_ref->{$nutrient});
+				}
+				else {
+					# return at top level
+					$customized_product_ref->{$nutrient} = $full_nutriments_ref->{$nutrient};
+				}
+			}
+		}
+	}
```
alexgarel, line 1022: great way to handle this !

lib/ProductOpener/API.pm
stephanegigandet: This will be triggered and will work only on API v2, those fields don't exist in API v3.

PR #12864, opened 2026-01-07
fix: test gen_top_tags_per_country.pl, add route for categories nutrient stats

stephanegigandet:
While looking at use of the "nutriments" structure, I found a broken route for a feature that was added more than 10 years ago: when grouping by categories, being able to see their nutrient stats for one specific nutrient. 
The route was something like /categories/sugars , but that doesn't work with recent changes to routes.

I changed the route to /categories?stats_nid=sugars

e.g. http://world.openfoodfacts.localhost/facets/categories/desserts/categories?stats_nid=sugars

image

We didn't have a test for it, so I added one.

But in fact we were not testing gen_top_tags_per_country.pl (which generates the categories stats), so I added an integration test for it in web_html.t

The category nutrition stats are loaded only once at startup, so in order to be able to test it, I also added an API route to reload the category stats (only possible on localhost server). It's very internal so not sure if it should be documented.

Then I also found that web_html.t had incorrect nutrition input data (nutriments_salt instead of nutriment_salt), so nutrient data was not loaded for those products, which in turn affected Nutri-Score, attributes, knowledge panels etc.

### DIFF
--- a/tests/integration/expected_test_results/web_html/world-edit-product.html
+++ b/tests/integration/expected_test_results/web_html/world-edit-product.html
@@ -2071,7 +2071,7 @@ You accept to be credited by re-users by a link to the product your are contribu
                             <!-- extra field suffixed with _displayed to allow detecting when a box is unchecked -->
                             <input type="hidden" name="nutrition_data_displayed" value="1" />
                             <input type="checkbox" id="nutrition_data" name="nutrition_data" checked="checked" />
-                            <label for="nutrition_data" class="checkbox_label"></label> &nbsp;
+                            <label for="nutrition_data" class="checkbox_label">Nutrition facts are specified for the product as sold.</label> &nbsp;
                             <input type="radio" id="nutrition_data_per_100g" value="100g" name="nutrition_data_per" checked="checked" />
                             <label for="nutrition_data_per_100g">for 100 g / 100 ml</label>
                             <input type="radio" id="nutrition_data_per_serving" value="serving" name="nutrition_data_per"  />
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/API.pm
```diff
@@ -469,7 +469,11 @@ my $dispatch_table = {
 		HEAD => \&external_sources_api,
 		OPTIONS => sub {return;},    # Just return CORS headers
 	},
-
+	# Used for testing purposes to reload categories stats without restarting the server
+	reload_categories_stats => {
+		GET => \&reload_categories_stats_api,
+		POST => \&reload_categories_stats_api,
+	},
```
alexgarel: I would prefix the route with __ maybe to make it clear it's private.

lib/ProductOpener/Food.pm
```diff
@@ -171,16 +172,20 @@ sub check_nutriscore_categories_exist_in_taxonomy() {
...
 # as well as in data quality checks and improvement opportunity detection
 
-if (opendir(my $dh, "$BASE_DIRS{PRIVATE_DATA}/categories_stats")) {
-	foreach my $file (readdir($dh)) {
-		if ($file =~ /categories_nutriments_per_country.(\w+).sto$/) {
-			my $country_cc = $1;
-			$categories_nutriments_per_country{$country_cc}
-				= retrieve(
-				"$BASE_DIRS{PRIVATE_DATA}/categories_stats/categories_nutriments_per_country.$country_cc.sto");
+sub load_categories_nutriments_per_country() {
```
alexgarel: I would by default only verify if the categories_nutriments_per_country is empty but add a force_reload parameter, that you can ask for your reload api.

This way there is less risk that we loose time reloading those values in scenario where it was not intended (while ensuring this data structure was loaded).

This would be more consistent with the other data initialization methods.

scripts/gen_top_tags_per_country.pl
```diff
@@ -445,17 +443,21 @@
 	# Category stats for nutriments
 
 	my $min_products = 10;
+	# On dev server we have less products, so use a smaller minimum
+	if ($server_domain =~ /localhost$/) {
+		$min_products = 2;
+	}
```
alexgarel, line 449: It would be better to have a config parameter "$dev_server" than to use this trick everywhere, or maybe a dev_server variable in request_ref.

stephanegigandet: That one I'm not too confortable with, I'm quite confused by the config parameters settings on the different platforms, with the env variables etc. and where I should set it (e.g. config_docker.pm which I guess is also used by staging, so would need an env var override). Let's do it in another PR.

alexgarel: setting it in request ref is a good alternative that said.

stephanegigandet:
This PR was incorporated (and changed for the testing) in this new-nutrition branch: https://github.com/openfoodfacts/openfoodfacts-server/pull/12871

So we probably won't merge this PR in the main branch directly, but instead wait for the new-nutrition-update branch to be merged in main.

PR #12865, opened 2026-01-07
fix: disable temporarily Google Vision extraction other than text

raphael0202:
As we run out of Google credits, let's disable non-critical Google Vision features.


raphael0202:
> Shan't we keep face detection and safe_search detection that are useful for moderation ? (or is the false positive ratio too high)

We just discussed the matter with Charles, safe search and face detection is half of the Google Vision cost, and this is something we can easily internalize (by using open models).

PR #12866, opened 2026-01-07
fix: glycemic index label from new nutrition data

stephanegigandet:
If we have glycemic index in nutrition data, we add a "Glycemic index" label.

There's also a label for carbon footprint, but in the future we'll probably remove the carbon footprint from nutrition data

Also added some comments that product edit rules probably don't work very well with API v3.

### DIFF
--- a/.test_groups_cache/unit_groups.json
+++ b/.test_groups_cache/unit_groups.json
@@ -2,6 +2,25 @@
   {
     "tests": [
       "tests/unit/additives.t",
+      "tests/unit/api.t",
+      "tests/unit/cursor.t",
+      "tests/unit/display.t",
+      "tests/unit/http.t",
+      "tests/unit/ingredients.t",
+      "tests/unit/ingredients_nutriscore.t",
+      "tests/unit/knowledge_panels.t",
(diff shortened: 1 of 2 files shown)


tests/unit/analyze_and_enrich_product_data.t
alexgarel, line 1: Thanks for adding a test.

lib/ProductOpener/FoodProducts.pm
```diff
@@ -116,6 +118,28 @@ sub specific_processes_for_food_product ($product_ref) {
 	# Determine packaging components in contact with food
 	determine_food_contact_of_packaging_components_service($product_ref);
 
+	# Add some labels from nutrition data (e.g. glycemic index, carbon footprint)
+	add_labels_from_nutrition_data($product_ref);
+
+	return;
+}
+
+sub add_labels_from_nutrition_data ($product_ref) {
```
alexgarel, line 127: Clean and neat separation of concern :tada:

alexgarel (approved):
Cool PR

PR #12870, opened 2026-01-08
fix: new nutrition schema: added sugars knowledge panel

stephanegigandet:
Use the new schema to get needed nutrients for added sugars knowledge panel, and alcoholic beverages food group.

### DIFF
--- a/lib/ProductOpener/FoodGroups.pm
+++ b/lib/ProductOpener/FoodGroups.pm
@@ -58,6 +58,8 @@ use ProductOpener::Food qw/is_beverage_for_nutrition_score_2021/;
 
 use Log::Any qw($log);
 
+use Data::DeepAccess qw(deep_get);
+
 # Note: the %pnns structure is a hash of sub-groups (aka "PNNS groups 2") to groups (aka "PNNN groups 1").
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/FoodGroups.pm
```diff
@@ -324,9 +324,10 @@ sub temporarily_change_categories_for_food_groups_computation ($product_ref) {
 	if (    ($product_ref->{nutrition_score_beverage})
 		and (not has_tag($product_ref, "categories", "en:instant-beverages")))
 	{
+		my $alcohol_100g = deep_get($product_ref, qw(nutrition aggregated_set nutrients alcohol value));
 
-		if (defined $product_ref->{nutriments}{"alcohol_100g"}) {
-			if ($product_ref->{nutriments}{"alcohol_100g"} < 1) {
+		if (defined $alcohol_100g) {
+			if ($alcohol_100g < 1) {
```
alexgarel, line 332: There seems to be no test change related to that ? Can we add a test ?

stephanegigandet: Good idea, I added a test, and indeed there was an issue!

PR #12871, opened 2026-01-08
refactor: category nutrition stats for new nutrition schema

stephanegigandet:
Includes PR https://github.com/openfoodfacts/openfoodfacts-server/pull/12854 from @Vic142 
+ https://github.com/openfoodfacts/openfoodfacts-server/pull/12864

We display nutrition facts tables in 3 places:

- product page (knowledge panel)
- category (nutrition stats for the category)
- search results for graphs (e.g. scatter plot with fat and sugar: we display stats for fat and sugars)

The last 2 were converted to use the same knowledge panel as the product nutrition facts.

This PR also updates the code to generate stats for categories, in order to support the new nutrition stats format.

For testing, instead of trying to generate stats on the fly and trying to get them reloaded by Apache+mod_perl, we now have a specific stats file in tests/data/categories_stats

Note: a lot of the code is existing code (sometimes moved to another module), it certainly can be improved a lot, but we can do it later.

Nutrition fact table with comparison to category:

image

(note that we have a lot of estimated nutrients, we could decide not to display them unless they are important nutrients)

Nutrition stats for category:

image

Nutrition stats for graphs:

image

### DIFF
--- a/lib/ProductOpener/Test.pm
+++ b/lib/ProductOpener/Test.pm
@@ -247,11 +247,9 @@ sub remove_all_products () {
 	if (@$err) {
 		confess("not able to remove some products directories: " . join(":", @$err));
 	}
-	# Also remove category stats
-	remove_tree($BASE_DIRS{PRIVATE_DATA} . "/categories_stats", {keep_root => 0, error => \$err});
-	if (@$err) {
-		confess("not able to remove some category stats directories: " . join(":", @$err));
-	}
+	# Note: we do not remove categories stats from PRIVATE_DATA and TEST_PRIVATE_DATA
+	# In integration tests, PRIVATE_DATA/categories_stats should not exist,
+	# and categories stats should be loaded from TEST_PRIVATE_DATA/categories_stats
 }
 
 =head2 remove_all_users ()
--- a/scripts/gen_top_tags_per_country.pl
+++ b/scripts/gen_top_tags_per_country.pl
@@ -50,13 +50,23 @@ use File::Path qw(mkpath);
 use JSON::MaybeXS;
 use Data::DeepAccess qw(deep_get);
 
-# Output will be in the $BASE_DIRS{PRIVATE_DATA} directory
+# By default, output will be in the $data_dir directory
 # data/index: data related to the Open Food Hunt operation (old): points for countries, users and ambassadors
 # data/categories_stats: statistics for the nutrients of categories, used to compare products to their categories
 
-ensure_dir_created_or_die("$BASE_DIRS{PRIVATE_DATA}/index");
-ensure_dir_created_or_die("$BASE_DIRS{PRIVATE_DATA}/categories_stats");
+# For integration tests, PRIVATE_DATA/categories_stats should not exist,
+# and we will load category stats from tests/data/category_stats/ (TEST_PRIVATE_DATA/categories_stats)
 
+# Allow to pass a different directory as first argument
+my $data_dir = $BASE_DIRS{PRIVATE_DATA};
+if (defined $ARGV[0]) {
+	$data_dir = $ARGV[0];
+}
+
+print STDERR "Generating top tags per country in $data_dir\n";
+
+ensure_dir_created_or_die("$data_dir/index");
+ensure_dir_created_or_die("$data_dir/categories_stats");
 # Generate a list of the top brands, categories, users, additives etc.
 
 my @fields = qw (
@@ -357,7 +367,7 @@ while (my $product_ref = $cursor->next) {
 # compute points
 # Read ambassadors.txt
 my %ambassadors = ();
-if (open(my $IN, q{<}, "$BASE_DIRS{PRIVATE_DATA}/ambassadors.txt")) {
+if (open(my $IN, q{<}, "$data_dir/ambassadors.txt")) {
 	while (<$IN>) {
 		chomp();
 		if (/\s+/) {
@@ -368,7 +378,7 @@ if (open(my $IN, q{<}, "$BASE_DIRS{PRIVATE_DATA}/ambassadors.txt")) {
 	}
 }
 else {
-	print STDERR "$BASE_DIRS{PRIVATE_DATA}/ambassadors.txt does not exist\n";
+	print STDERR "$data_dir/ambassadors.txt does not exist\n";
 }
 
 my %ambassadors_countries_points = (_all_ => {});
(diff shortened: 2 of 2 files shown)


lib/ProductOpener/API.pm
```diff
@@ -469,7 +469,12 @@ my $dispatch_table = {
 		HEAD => \&external_sources_api,
 		OPTIONS => sub {return;},    # Just return CORS headers
 	},
-
+	# Internal API used for testing purposes to reload categories stats without restarting the server
+	# Prefixed with _ to indicate that it is not intended for public use
+	_reload_categories_stats => {
+		GET => \&reload_categories_stats_api,
+		POST => \&reload_categories_stats_api,
+	},
```
alexgarel: Do we still need it ?

lib/ProductOpener/Stats.pm
```diff
@@ -0,0 +1,217 @@
...
+
+	# If already loaded, do not reload unless forced
+	return if (keys %categories_stats_per_country) and (not $force_reload);
+
+	%categories_stats_per_country = ();
+
+	my $dir = $BASE_DIRS{PRIVATE_DATA} . "/categories_stats";
+	if (!-d $dir) {
+		$dir = $BASE_DIRS{PRIVATE_DATA_TESTS} . "/categories_stats";
+	}
```
alexgarel, line 206: But this means we are using this folder also when we are in dev mode. I'm ok with it but why did you move it to tests ?

For me it was ok to keep it in data-default, and have it mounted in dev. Now that's more clear for tests, but less for devs who might ask themselves why they modified a test folder.

data-default was quite explicit, ain't it ?

lib/ProductOpener/Stats.pm
```diff
@@ -0,0 +1,217 @@
...
+use vars @EXPORT_OK;
+
+use ProductOpener::Config qw/:all/;
+use ProductOpener::Paths qw/%BASE_DIRS/;
+use ProductOpener::Store qw(retrieve);
+
+=head1 FUNCTIONS
+
+=head2 add_product_value_to_stats ($values_ref, $key, $value)
+
```
alexgarel, line 60:
```suggestion
=head2 add_product_value_to_stats ($values_ref, $key, $value)

This build a collection (C<$value_ref>) were for each nutrients (or other numeric attribute), C<$key>, we store the accumulate values for that nutrient (C<s>), number of instances <n> and each values C<array>).

We will then pass C<$values_ref> to C<compute_stats_for_products>.

```

lib/ProductOpener/Stats.pm
```diff
@@ -0,0 +1,217 @@
...
+=head3 Arguments
+
+=head4 Stats reference $stats_ref
+
+Where we will store the stats.
+
+=head4 Values reference $values_ref
+
+Values for some nutrients (or other numeric values).
+
```
alexgarel, line 97:
```suggestion
=head4 Values reference $values_ref

Values for some nutrients (or other numeric values),
with accumulated values and number of products considered
(it is different from C<$n> as some products might not have a value).
See C<add_product_value_to_stats>

```

lib/ProductOpener/Stats.pm
```diff
@@ -0,0 +1,217 @@
...
+		my @values = sort {$a <=> $b} @{$values_ref->{$key}{array}};
+		my $nb_values = $#values + 1;
+		my $kept_values = 0;
+		my $sum_of_kept_values = 0;
+
+		my $i = 0;
+		foreach my $value (@values) {
+			$i++;
+			next if ($i <= $nb_values * 0.05);
+			next if ($i >= $nb_values * 0.95);
```
alexgarel, line 136: This means you remove values only when you have at least 20 values ?

alexgarel: In the future we could consider more standard methods (easy to implement) like inter-quartile range or Z score ([ex in python](https://www.geeksforgeeks.org/data-science/detect-and-remove-the-outliers-using-python/))

lib/ProductOpener/Routing.pm
```diff
@@ -607,23 +606,11 @@ sub facets_route($request_ref) {
...
-		pop @components;
-		$log->debug(
-			"facets_route - request looks like a list of tags - categories with nutrients",
-			{groupby => $request_ref->{groupby_tagtype}, stats_nid => $request_ref->{stats_nid}}
-		) if $log->is_debug();
+	# categories group by can have a stats_nid parameter to add nutrition stats to list of categories
+	if (defined single_param("stats_nid")) {
+		$request_ref->{stats_nid} = single_param("stats_nid");
+		$log->debug("facets_route - got stats_nid parameter", {stats_nid => $request_ref->{stats_nid}})
+			if $log->is_debug();
```
alexgarel, line 613: Cool to see simplified code !

lib/ProductOpener/Test.pm
```diff
@@ -247,6 +247,11 @@ sub remove_all_products () {
 	if (@$err) {
 		confess("not able to remove some products directories: " . join(":", @$err));
 	}
+	# Also remove category stats
+	remove_tree($BASE_DIRS{PRIVATE_DATA} . "/categories_stats", {keep_root => 0, error => \$err});
+	if (@$err) {
+		confess("not able to remove some category stats directories: " . join(":", @$err));
+	}
```
alexgarel: Ok, so apache2 is using the TEST_PRIVATE_DATA, while gen_top_tags_per_country will use PRIVATE_DATA.

It might be confusing to developpers though…
At least we should add a comment here to explain this.

But IMO, it would be better, if you want to test gen_country_tags that you maybe able to specify a target directory as a parameter, this way this is more explicit.

templates/api/knowledge-panels/health/nutrition/nutrition_facts_table.tt.json
```diff
@@ -7,7 +7,8 @@
     "expand_for": "large",
     "evaluation": "unknown",
     "title_element": {
-        "title": "[% edq(lang('nutrition_data_table')) %]",
+        "title": "[% IF panel.title.defined %]edq([% panel.title %])[% ELSE %][% edq(lang('nutrition_data_table')) %][% END %]",
```
alexgarel:
```suggestion
        "title": "[% IF panel.title.defined %][% edq(panel.title) %)[% ELSE %][% edq(lang('nutrition_data_table')) %][% END %]",
```

tests/unit/expected_test_results/display/nutrition-facts-table-with-comparisons.json
alexgarel, line 1: There an unresolved merge conflict here !

lib/ProductOpener/Stats.pm
stephanegigandet: The data-default wasn't working for me for tests, I'm not sure why. So now in dev mode we can still use the data directory, and there's a separate dir that is used when the data directory does not have category stats.

lib/ProductOpener/API.pm
stephanegigandet: I removed it, it wasn't working well with multiple Apache processes.

lib/ProductOpener/Stats.pm
stephanegigandet: Yes, it definitely could be improved with more standard methods.

lib/ProductOpener/Test.pm
stephanegigandet: Good idea, I added a parameter to be able to specify a target directory (like tests/data/)

PR #12872, opened 2026-01-09
fix(site-layout): preserve selected country on language change

### LINKED ISSUE #12847: Country dropdown label resets when changing language, while country filter remains applied
On the products page, the Country filter and Language selector appear to be out of sync when changing the interface language.

While the applied country filter remains active (products do not change), the Country dropdown label resets, which makes the UI inconsistent and misleading.

**Steps to reproduce**

1. Go to the products page with no filters applied
2. Select Country = China
3. Products shown are from China
4. Country dropdown correctly shows “China” (in Chinese)
5. Change Language = English
6. Observe that:

- Products are still Chinese (filter still applied)
- The Country dropdown label resets to “Country” instead of “China”

**Expected behavior**

- Changing the interface language should not reset the selected country label
- The Country dropdown should continue to display China if the country filter is still applied
- UI state should always reflect the actual active filters

**Actual behavior**

- Country filter remains active internally
- Country dropdown label resets visually
- UI no longer reflects the real filtering state

**Why this is an issue**

- It breaks UI consistency
- It misleads users into thinking the country filter was removed
- It creates confusion between visual state and actual applied filters

**Additional notes**

This looks like a frontend state or re-rendering issue when switching locale (language), where the country filter value is preserved but the dropdown label is reinitialized.

Happy to investigate or propose a fix once the expected behavior is confirmed.

Image

Image

Image

sidhansu10:
### Fix country dropdown label reset on language change
When switching the interface language on the products page, the selected country filter remains active internally but the country dropdown label was reset, leading to a UI inconsistency. This PR rehydrates the country `<select>` server-side using the existing `country_code` variable so that the dropdown label always reflects the active filter after a language change.
1) No change to filtering logic
2) No JavaScript changes
3) Minimal server-side template fix
Fixes #12847


sidhansu10:
The SonarCloud Quality Gate failure appears to be a false positive. This PR only changes a Template::Toolkit HTML file and does not introduce any executable logic. SonarCloud also reports “No new alerts in code changed by this pull request”. Likely requires maintainer override.

alexgarel (commented):
thanks @sidhansu10 
This looks good but would need testing to be sure.
Did you tested it locally ? The best is to add a screenshot to you PR description.

alexgarel:
@sidhansu10 I tried it locally and it does not seems to work (I didn't debug).

image

teolemon:
@sidhansu10 any update adressing @alexgarel 's comment ?

PR #12873, opened 2026-01-10
fix: pH unit being incorrectly shown as grams

### LINKED ISSUE #12869: Editor uses the wrong unit for pH
### What
- The product editor shows grams as the unit for pH when it is a dimensionless quantity.
- Also it should be spelled with a lowercase 'p' and uppercase 'H'.

See https://en.wikipedia.org/wiki/PH#pH:
> The pH scale is logarithmic and therefore pH is a dimensionless quantity.

1. Log in to OFF.
2. Go to https://world.openfoodfacts.org/cgi/product.pl?type=edit&code=3256223010488
3. Go to the nutrition facts.
4. Look for the pH value. It is displayed as being 7.8 _g_ and the user can choose between g, mg and mcg.

### Expected behavior
- Show the pH with no unit (and preferably the correct case).

### Screenshot
Image

### Type of device
- Platform: Desktop
- Operating System: Debian 13
- Browser: Firefox 143.0.4

Hardik301002:
What -
Fixes pH being incorrectly displayed with mass units (g / mg / mcg) in the nutrition editor.

Why -
pH is a logarithmic, dimensionless quantity and must not have a unit.
Displaying grams is scientifically incorrect and confusing.

How -
- Prevented default unit fallback for pH
- Ensured pH remains unitless in nutrition forms and conversions

Steps to test -
1. Log in to Open Food Facts
2. Edit a product
3. Go to the Nutrition tab
4. Verify pH is displayed without any unit

Fixes #12869


lib/ProductOpener/Food.pm
```diff
@@ -304,8 +304,9 @@ sub default_unit_for_nid ($nid) {
 		return "%";
 	}
 	else {
-		return "g";
-	}
+        return undef if $nid eq 'ph';
+        return "g";
+    }
```
alexgarel: above we have hardcoded in the code because we match a regexp, but there you have a simple value so you should modify default_unit_for_nid_map instead to add an entry for ph.

lib/ProductOpener/Food.pm
```diff
@@ -3397,8 +3398,9 @@ sub get_nutrient_unit ($nid, $cc = undef) {
 		$unit = get_property("nutrients", "zz:$nid", "unit_$cc:en");
 		return $unit if $unit;
 	}
-	$unit = get_property("nutrients", "zz:$nid", "unit:en") // 'g';
-	return $unit;
+	$unit = get_property("nutrients", "zz:$nid", "unit:en");
+return undef if $nid eq 'ph';
+return $unit // 'g';
```
alexgarel, line 2656: There again we should add ph entry if it's not there to nutrients taxonomy and add a unit which is empty.

alexgarel (changes requested):
Thanks for the PR, see my comments.

PR #12876, opened 2026-01-11
fix: support carbohydrates-total in energy value computation check

anushkaaaaaaaa:
- [x] PR title is prefixed by one of the following: feat, fix, docs, style, refactor, test, build, ci, chore, revert, l10n, taxonomy
- [x] Code is well documented
- [x] If you have multiple commits please combine them into one commit by squashing them.
- [x] Read and understood the [contribution guidelines](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/CONTRIBUTING.md)

### What

Fix energy value computation check to support US-style `carbohydrates-total` nutrient field.

**Problem:** US products that correctly use "Total Carbohydrates" (`carbohydrates-total`) were getting false positive data quality errors:
- `en:energy-value-in-kcal-does-not-match-value-computed-from-other-nutrients`

**Root cause:** The `check_nutrition_data_energy_computation` function in `DataQualityFood.pm` only checked for `carbohydrates_value`, ignoring `carbohydrates-total_value` which is the default for US products.

**Fix:** 
1. Updated the condition to accept either `carbohydrates_value` OR `carbohydrates-total_value`
2. When computing energy from carbohydrates, use `carbohydrates-total_value` as fallback when `carbohydrates_value` is undefined

- Fixes the issue where US products with `carbohydrates-total` field trigger incorrect energy mismatch errors
- Examples: https://world.openfoodfacts.org/product/0850077995026/prebiotic-fiber-formula-loam and https://world.openfoodfacts.org/product/0055742550832/salted-caramel-frozen-dessert-compliments

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -902,7 +902,7 @@ sub check_nutrition_data_energy_computation ($product_ref) {
 		# We need at a minimum carbohydrates, fat and proteins to be defined to compute
 		# energy.
 		if (    (defined $specified_energy)
-			and (defined $nutriments_ref->{"carbohydrates_value"})
+			and ((defined $nutriments_ref->{"carbohydrates_value"}) or (defined $nutriments_ref->{"carbohydrates-total_value"}))
 			and (defined $nutriments_ref->{"fat_value"})
 			and (defined $nutriments_ref->{"proteins_value"}))
 		{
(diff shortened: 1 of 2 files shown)


jayaddison:
Hi @anushkaaaaaaaa - I am not a member of the team at OpenFoodFacts but I am a contributor to the project.  I work on fixing data-quality errors -- and a lot of those relate to nutritional energy values.

Can I check if my understanding is correct that:

 * In the United Kingdom, European Union (and maybe elsewhere): `Carbohydrates` on food product nutritional information labeling equals the `sum of all sugars, plus starch, minus polyols`
 * In the United States, (and again, maybe elsewhere): `Carbohydrates` on food product nutritional information labeling equals `sum of all sugars, plus starch, minus fibre, minus polyols`

In other words, it's generally ~~`polyols`~~ `fibre` that are the difference between the existing OpenFoodFacts `carbohydrates` datafield and the new `carbohydrates-total` datafield.

I think we should also attempt to identify the number of products already in the OpenFoodFacts database where we'll want to move/copy the data from one of the fields to the other.

Edit: fixup: fibre, not polyols, seems to be the difference?

jayaddison:
Sorry, I made a mistake - re-reading the calculations listed there, I think it's the `fibre` quantity that seems to be the difference between the two methods.  In my experience, missing polyol quantities do often cause energy calculation discrepancies too, but that's a separate topic.  I've edited my message to try to correct that.

PR #12898, opened 2026-01-17
fix: disable browser autocomplete on search input to prevent button o…

### LINKED ISSUE #12897: Fix browser autocomplete
## What
- Fixes browser autocomplete dropdown overlapping with the barcode scanner button in the search bar
- Added `autocomplete="off"` attribute to search input fields

## Why
When users type in the search bar and browser autocomplete suggestions appear from previous search history, the turquoise barcode scanner button overlaps with the dropdown, creating a visual obstruction and affecting usability.

## Changes
- Added `autocomplete="off"` to main navigation search input (line 230)
- Added `autocomplete="off"` to mobile navigation search input (line 270)

## Testing
1. Navigate to the homepage
2. Click the search input and type previously searched text
3. Verify no browser autocomplete appears
4. Confirm search functionality still works correctly

## Screenshot

Image

dhiraj-143r:
When users type in the search bar, browser autocomplete suggestions from previous search history can appear and overlap the turquoise barcode scanner button, causing a visual obstruction and reducing usability.

Changes:
Added autocomplete="off" to the main navigation search input (line 230)
Added autocomplete="off" to the mobile navigation search input (line 270)

Testing:
Go to the homepage
Click the search input and type previously searched text
Confirm browser autocomplete does not appear
Ensure search still works normally

Fixes #12897

### DIFF
--- a/templates/web/common/site_layout.tt.html
+++ b/templates/web/common/site_layout.tt.html
@@ -227,7 +227,7 @@
 								<div class="row"><div class="small-12">
 								<div class="row collapse postfix-round">
 									<div class="columns">
-									<input type="text" placeholder="[% edq(lang('search_a_product_placeholder')) %]" name="search_terms" value="[% search_terms %]" style="background-color:white">
+									<input type="text" placeholder="[% edq(lang('search_a_product_placeholder')) %]" name="search_terms" value="[% search_terms %]" style="background-color:white" autocomplete="off">
 									<input name="search_simple" value="1" type="hidden">
 									<input name="action" value="process" type="hidden">
 									</div>
@@ -267,7 +267,7 @@
 				<form action="/cgi/search.pl">
 					<div class="row collapse">
 						<div class="small-8 columns">
-							<input type="text" placeholder="[% edq(lang('search_a_product_placeholder')) %]" name="search_terms">
+							<input type="text" placeholder="[% edq(lang('search_a_product_placeholder')) %]" name="search_terms" autocomplete="off">
 							<input name="search_simple" value="1" type="hidden">
 							<input name="action" value="process" type="hidden">
 						</div>


john-gom:
Would it not be better to fix the CSS so that the button is not obscured rather than removing the auto complete functionality?

dhiraj-143r:
@john-gom 
I agree with you

but its length get increase continuously as it shows all the recent search , so I think having a clean search bar will be better

dhiraj-143r:
@john-gom @teolemon 
 I will Revert the autocomplete="off" changes, and Fix the overlapping via CSS

Let me know your thoughts

I just tried to keep the search bar clean

teolemon:
@dhiraj-143r have you run this locally ? could you provide a little video of the change ?

dhiraj-143r:
@teolemon 
yes I have run this locally 

 video before changes : 
https://github.com/user-attachments/assets/d2737546-3897-4e07-831e-47f3c66d3e00

 video after changes : 
https://github.com/user-attachments/assets/632a00a5-cc19-4a18-97a0-1933a8e55d57

dhiraj-143r:
now :
* Search functionality works normally without any issues and it looks clean as well 
* Both desktop and mobile search inputs behave correctly

dhiraj-143r:
@teolemon @stephanegigandet 

PR is ready for review 

let me know if any changes are required

john-gom:
I thought you were going to look at a CSS based solution?

john-gom:
@dhiraj-143r just looking at this some more. On your before video I can't see any evidence of anything being obstructed. I'm not sure what problem we are trying to solve here.

dhiraj-143r:
@john-gom 
thanks for re-checking this and for the follow‑up.

After looking at it again on my side, it seems the overlap I saw was likely due to my specific browser/setup (history length, viewport, etc.), because in the “before” video the button is not actually being obstructed as you pointed out.

Given that, this does not look like a real cross‑browser issue, so I’m happy to drop this change and close the PR so it doesn’t create noise.

PR #12899, opened 2026-01-17
fix: prevent taxonomy cache growth

### LINKED ISSUE #12893: Taxonomies GitHub cache keeps growing
### What
After some time builds start to fail with errors like this:

```
ERROR: write /opt/product-opener/build-cache/taxonomies/labels.a010e6b7976aa1a99e1fd076a1a1041bf5a2ddfb.result.json: no space left on device
```

This seems to be because the [taxonomies cache](https://github.com/openfoodfacts/openfoodfacts-server/actions/caches) (not to be confused with the build-cache GitHub repository) keeps growing as new taxonomies with different hashes are added to it.

Observe that the `Build backend dev image for tests` build step starts to fail if the taxonomy cache has not been cleared recently

### Expected behavior
Old taxonomy build hashes should be deleted from the GitHub cache when a new one is created.

### Why
Causes builds to fail which can affect deployment and also results in extra developer time as taxonomies are not saved in the build-cache repository so need to be re-built locally.

dhiraj-143r:
Fixes #12893

This PR prevents the taxonomy cache (build-cache/taxonomies) from growing forever and eventually breaking builds with “no space left on device”.
After writing new taxonomy cache files, we now automatically clean up older ones and keep only the 5 most recent cache sets per taxonomy (grouped by hash). This keeps recent caches available for ongoing PRs/builds while stopping unlimited cache growth.

Tested by validating the cleanup logic and running:


templates/web/common/site_layout.tt.html
```diff
@@ -227,7 +227,7 @@
 								<div class="row"><div class="small-12">
 								<div class="row collapse postfix-round">
 									<div class="columns">
-									<input type="text" placeholder="[% edq(lang('search_a_product_placeholder')) %]" name="search_terms" value="[% search_terms %]" style="background-color:white">
+									<input type="text" placeholder="[% edq(lang('search_a_product_placeholder')) %]" name="search_terms" value="[% search_terms %]" style="background-color:white" autocomplete="off">
```
john-gom: I think this should be in a different PR

john-gom:
You will need to rename the PR to "fix: ..." To comply with the semantic naming rules

dhiraj-143r:
@john-gom  PR is ready for review now,
Thanks for the note! I’ve renamed the PR title to follow the semantic naming convention

john-gom:
@dhiraj-143r you still need to remove the auto complete change from this PR

dhiraj-143r:
@john-gom Done! 
I have  removed the autocomplete changes from this PR. It now only contains the taxonomy cache cleanup fix.

dhiraj-143r:
@john-gom 

the PR is ready for review now

john-gom:
Thanks @dhiraj-143r , this is great. I have two small suggestions:

1. It might be good to also clean up the local cache when we fetch a taxonomy from github. Could do this by just adding another call to your function at the end of the `get_from_cache` sub.
2. Having the list of file types is a duplicating things a bit. Also, we recently moved from .sto files to .json so these older sto files won't be deleted. I would suggest just deleting all files that are prefixed with the older hashes.

dhiraj-143r:
@john-gom  
Thanks for the detailed review

1. That makes sense, I will add a call to cleanup_old_cache_files at the end of the get_from_cache sub so the local cache is also cleaned when fetching from GitHub.

2. Good point about the duplication and old .sto files. I will update the cleanup logic to delete all files whose names start with the older hashes, instead of relying on an explicit list of file types.

I will push an updated version of the PR shortly.

dhiraj-143r:
@john-gom 
Both suggestions implemented. Ready for review.
Thanks

dhiraj-143r:
@john-gom 
thank you so much

alexgarel:
thanks for this contribution @dhiraj-143r !

PR #12904, opened 2026-01-17
taxonomy: pickled vegetables

moon-rabbitOFF:
Pickles vegetables should not be in the category of their vegetable, because thew are not used in the same way. E.g one can use fresh mushrooms, canned mushrooms or frozen mushrooms to make the same recipe. Pickled mushrooms will have a very different taste and will not be adapted for the same recipes. Moreover, they can contain oil or a great quantity of salt which change substentially their nutrional values compared to fresh, canned or frozen products.

### DIFF
--- a/taxonomies/brands.txt
+++ b/taxonomies/brands.txt
@@ -1016,3 +1016,4 @@ xx: 1de Beste
 wikidata:en: Q136929195
 
 xx: 움트리
+
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -108105,8 +108105,8 @@ fr: Épis de maïs marinés, Maïs doux mariné
 en: Turnips pickles
 fr: Navets marinés
 
-< en:bamboo shoots
 < en:Vegetable pickles
+< en:bamboo shoots
 en: Bamboo shoots pickles, marinated bamboo shoots
 fr: Pousses de bambou marinées
 


Freso:
I don’t think I agree with this reasoning. All pickled [vegetables] _are_ [vegetables], but not all [vegetables] are pickled [vegetables]. If you want specifically to find fresh [vegetables], most (if not all) vegetables have a corresponding “fresh [vegetables]” category.

Additionally, vegetables often have subcategories that vary on those same criteria you specified. Using “mushrooms” as an example, the “mushrooms” category has subcategories for champignons, chanterelles, truffles, shiitakes, etc. as well as for “mixed mushrooms”. These all have different flavour and texture profiles as well as different nutritional values.

IMHO, if someone wants to use OFF categories for some kind of cooking recipe site or program, they should themselves specify the most precise ingredient and either allow for alternatives (e.g., “[Fresh mushrooms](https://world.openfoodfacts.org/facets/categories/fresh-mushrooms) (or [Frozen mushrooms](https://world.openfoodfacts.org/facets/categories/frozen-mushrooms) or [Canned mushrooms](https://world.openfoodfacts.org/facets/categories/canned-mushrooms))”) rather than giving the generic term (e.g., “[Mushrooms](https://world.openfoodfacts.org/facets/categories/mushrooms)”)—or just trust that the reader/user will know that “[Fresh mushrooms](https://world.openfoodfacts.org/facets/categories/fresh-mushrooms)” can be substituted with frozen or canned ones.

moon-rabbitOFF:
To me, pickled vegetable are no longer _vegetables_, they are _prepared vegetables_, since they contain other ingredients that affect substentially the taste and the nutritional values. 
All raw, canned or frozen mushrooms will have a Nutriscore of A. This is not the case of pickled mushrooms, most of them in the database have a nutriscore C.

PR #12500, opened 2025-10-14
feat: New nutrition schema (WIP) - Development branch that we merge new nutrition PRs to

stephanegigandet:
This branch is a develop branch for the new nutrition schema. A lot of changes are needed, so we are making smaller PRs to this branch first, and when everything is ready we will merge this branch into main.

This branch can also be used for early testing.

PR #12914, opened 2026-01-19
fix: failing unit and integration tests on new-nutrition-update branch

stephanegigandet:
This PR is to fix failing unit and integration tests on the new-nutrition-update branch.

### DIFF
--- a/tests/integration/expected_test_results/api_v2_product_write/get-product-nutrition-no_nutrition_data-on-api-v3-6.json
+++ b/tests/integration/expected_test_results/api_v2_product_write/get-product-nutrition-no_nutrition_data-on-api-v3-6.json
@@ -485,8 +485,6 @@
       ],
       "nutriscore_version" : "2023",
       "nutrition" : {
-         "aggregated_set" : {},
-         "input_sets" : [],
          "no_nutrition_data_on_packaging" : true
       },
       "nutrition_grade_fr" : "unknown",
--- a/tests/integration/expected_test_results/convert_and_import_excel_file/carrefour-images/products/3560070167470.json
+++ b/tests/integration/expected_test_results/convert_and_import_excel_file/carrefour-images/products/3560070167470.json
@@ -568,10 +568,7 @@
       "unknown"
    ],
    "nutriscore_version" : "2023",
-   "nutrition" : {
-      "aggregated_set" : {},
-      "input_sets" : []
-   },
+   "nutrition" : {},
    "nutrition_grade_fr" : "unknown",
    "nutrition_grades" : "unknown",
    "nutrition_grades_tags" : [
--- a/tests/integration/expected_test_results/convert_and_import_excel_file/carrefour-images/products/3560070815746.json
+++ b/tests/integration/expected_test_results/convert_and_import_excel_file/carrefour-images/products/3560070815746.json
@@ -604,10 +604,7 @@
       "unknown"
    ],
    "nutriscore_version" : "2023",
-   "nutrition" : {
-      "aggregated_set" : {},
-      "input_sets" : []
-   },
+   "nutrition" : {},
    "nutrition_grade_fr" : "unknown",
    "nutrition_grades" : "unknown",
    "nutrition_grades_tags" : [
--- a/tests/integration/expected_test_results/convert_and_import_excel_file/openfoodfacts_import_template_fr_test/products/5228857000838.json
+++ b/tests/integration/expected_test_results/convert_and_import_excel_file/openfoodfacts_import_template_fr_test/products/5228857000838.json
@@ -492,10 +492,7 @@
       "unknown"
    ],
    "nutriscore_version" : "2023",
-   "nutrition" : {
-      "aggregated_set" : {},
-      "input_sets" : []
-   },
+   "nutrition" : {},
    "nutrition_grade_fr" : "unknown",
    "nutrition_grades" : "unknown",
    "nutrition_grades_tags" : [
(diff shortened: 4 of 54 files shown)


lib/ProductOpener/DataQualityFood.pm
```diff
@@ -771,84 +772,34 @@ sub detect_categories ($product_ref) {
...
 Compares the nutrition score and nutrition grade (Nutri-Score) we have computed with
 the score and grade provided by manufacturers.
 
 =cut
 
-sub check_nutrition_grades ($product_ref) {
+sub check_nutriscore_grades ($product_ref) {
 
-	if ((defined $product_ref->{nutrition_grade_fr_producer}) and (defined $product_ref->{nutrition_grade_fr})) {
+	if ((defined $product_ref->{nutriscore_grade_producer}) and (defined $product_ref->{nutriscore_grade})) {
```
Vic142, line 784: why is not fr anymore ? At line 786 there is a check on the `nutriscore_grade_fr_producer`, would it need to be without the fr part too ?

tests/unit/process_product_edit_rules.t
```diff
@@ -98,24 +98,24 @@ my @tests = (
 		desc => "Remove edit on a salt nutrient",
 		edit_rules => [{name => "Disallow salt", actions => [["ignore_nutriments_salt"]]},],
 		product => {nutrients => {salt_100g => 1}},
-		form => {nutrients_salt => 2},
-		delete_param => ["nutriments_salt", "nutriments_salt_100g"],
+		form => {nutriments_salt => 2},
```
Vic142, line 101: why did you change `nutrients_salt` into `nutriments_salt` ? Aren't nutrient related items supposed to be used with _nutrient_ instead of _nutriment_ ?

lib/ProductOpener/DataQualityDimensions.pm
```diff
@@ -170,15 +170,15 @@ sub compute_completeness_score($product_ref) {
 		add_tag($product_ref, "data_quality_completeness", "en:categories-to-be-completed");
 	}
 	$completeness_nutrition_total++;
-	# 2-3- nutriments
+	# 2-3- nutrition data
 	if (   has_non_estimated_nutrition_data($product_ref)
 		or has_no_nutrition_data_on_packaging($product_ref))
 	{
-		add_tag($product_ref, "data_quality_completeness", "en:nutriments-completed");
+		add_tag($product_ref, "data_quality_completeness", "en:nutrition-completed");
```
alexgarel, line 177: don't you have to change it in the taxonomy ?
Maybe you should also advertise this change to quality team.

alexgarel: Ping @benbenben2 for any comment ?

lib/ProductOpener/DataQualityFood.pm
alexgarel: I think nutrition_grade_fr was the name for nutriscore (because it was not immediately name nutriscore !). Read it nutrition grade for france.

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -771,84 +772,34 @@ sub detect_categories ($product_ref) {
...
 
-		if ($product_ref->{nutrition_grade_fr_producer} eq $product_ref->{nutrition_grade_fr}) {
-			push @{$product_ref->{data_quality_info_tags}}, "en:nutrition-grade-fr-producer-same-ok";
+		if ($product_ref->{nutriscore_grade_fr_producer} eq $product_ref->{nutriscore_grade}) {
```
alexgarel: there you kept the "fr" ?

scripts/obsolete/fix_non_normalized_codes.pl
alexgarel, line 1: This script is currently called in en_feeds_daily.sh !

Also we think we do not generated non normalized codes, it's better to check.

alexgarel (approved):
Great PR. I imagine you still need to check results against originals (in branch main) in some way ?

lib/ProductOpener/DataQualityDimensions.pm
stephanegigandet: I'll change it in the taxonomy

lib/ProductOpener/DataQualityFood.pm
stephanegigandet: That's right, nutrition_grade_fr etc. are old names, now it's called nutriscore and it's not specific do fr anymore, so we have new fields as well (that are already there in fact, it's not related to the new nutrition schema). I'm just using the newer field names, as we'll probably deprecate the old ones at some point.

lib/ProductOpener/DataQualityFood.pm
stephanegigandet: Changing, thanks.

tests/unit/process_product_edit_rules.t
stephanegigandet: The edit rules work on values that are sent to the API v2, with field names like nutriments_salt. We are not changing API v2.

scripts/obsolete/fix_non_normalized_codes.pl
stephanegigandet: Ah my bad, I thought it was something we ran only before we merged the databases for all flavours. I'll put it back and fix the tests.

stephanegigandet:
> Great PR. I imagine you still need to check results against originals (in branch main) in some way ?

@alexgarel : Yes, we will do that from the new-nutrition-update -> main PR (once this PR is merged): https://github.com/openfoodfacts/openfoodfacts-server/pull/12500

PR #12915, opened 2026-01-19
build(deps): Unblock: upgrade to gulp v5.0.1 (continuation of PR #10065)

jayaddison:
### What

A previous attempt to upgrade `gulp` to v5 failed due to problems processing the [`icons`](https://github.com/openfoodfacts/openfoodfacts-server/blob/af54b9d274cd18e319a356b3e369049de972fbf5/gulpfile.ts#L65-L80) and [`attributesIcons`](https://github.com/openfoodfacts/openfoodfacts-server/blob/af54b9d274cd18e319a356b3e369049de972fbf5/gulpfile.ts#L82-L92) groups in the `gulpfile.ts` build script.

In particular: processing of the source file glob patterns (e.g. `src("*.svg", { cwd: "./icons" })`) failed and emitted `ENOENT` errors.

I believe that this is because some of the SVG icons were represented using relative-path symlinks that did not previously resolve to valid file paths after being [copied](https://github.com/openfoodfacts/openfoodfacts-server/blob/af54b9d274cd18e319a356b3e369049de972fbf5/Dockerfile.frontend#L26) into the frontend Docker container.  Earlier versions of `gulp` (in fact, earlier versions of one of its multi-level dependencies, `glob-stream`) silently ignored these unresolvable hyperlinks, but with v5 it becomes an error.

This changeset resolves the problem by converting affected SVG icon symlinks to use absolute paths that resolve to the intended SVG files in the ProductOpener container.

:warning: This isn't particularly intuitive/ergonomic for developers working with plain checkouts of the repository, because it won't be immediately obvious what these icons contain and what they evaluate to when the application runs, but it does fix the build.

It would be nice to figure out a better longer-term strategy to solve this, but I think that unblocking dependency upgrades is more important, in the short-term.

- Continues / resolves #10065.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,7 +49,7 @@
         "@typescript-eslint/parser": "^8.32.0",
         "dotenv": "^17.2.3",
         "eslint": "^8.57.1",
-        "gulp": "^4.0.2",
+        "gulp": "^5.0.1",
         "gulp-babel": "^8.0.0",
         "gulp-concat": "^2.6.1",
         "gulp-csso": "^4.0.1",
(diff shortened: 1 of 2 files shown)


jayaddison:
I think that the [`hashFiles` error message](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/21152652375/job/60831604612#step:6:1) that caused the GitHub Actions job failure for the frontend container build job here (at commit 61a04d204333a4762a4177cc2ee70802a80a4aa5) could be clearer; there is an existing bugreport for that at: https://github.com/actions/runner/issues/3015

Basically: the `hashFiles` utility in GitHub Actions seems incompatible in some way with invalid/broken symlinks within the glob paths.

It _might_ be possible to remove the `icons` input from the `hashFiles` arguments we're using, but that might be contrary to the purpose of the hashing.

jayaddison:
Additional discussion required to figure out how best to resolve-or-avoid the `hashFiles` broken-symlink issue; marking this pull request as draft until that is determined.

alexgarel (commented):
@jayaddison unfortunately, there is a problem: in production we don't use the `docker compose` but have a plain deploy and location is not /opt/productopener, so we can't use fullpath like that…

jayaddison:
Let's find a better way, and preferably one that doesn't require creation of symlinks that rely on deployment-environment specifics to resolve correctly.

Issue #12695, opened 2025-11-19
ENOENT: no such file or directory, open '/github/workspace/docs/api/ref/ingredient.yaml'


teolemon:
### Task
[Check failure on line 1 in docs/api/ref/schemas/product.yaml](https://github.com/openfoodfacts/openfoodfacts-server/pull/12507/files#annotation_41838088456) 

@github-actions
github-actions
/ Lint (pull_request)
invalid-ref

ENOENT: no such file or directory, open '/github/workspace/docs/api/ref/ingredient.yaml'

teolemon:
it's actually in docs/api/ref/ingredient.yaml

sidhansu10:
i have fixed this issue by correcting the relative $ref path to ingredient.yaml in the product_ingredients schema.
PR: #12919

jayaddison:
@teolemon where is this GitHub Actions lint job defined?  @sidhansu10's fix seems correct to me.

jayaddison:
I'm unable to find this error appearing in recent pull requests, so I think that this can be closed.  Please feel free to re-open it if it does occur again.

PR #12919, opened 2026-01-20
docs: fix ingredient.yaml reference in product_ingredients schema

sidhansu10:
This PR fixes the documentation linter failure: ENOENT: no such file or directory, open '/github/workspace/docs/api/ref/ingredient.yaml'
The issue was caused by product_ingredients.yaml referencing ingredient.yaml using an incorrect relative path:
  ./ingredient.yaml
Both files are located in:  docs/api/ref/schemas/
Updated the $ref path to:  ingredient.yaml#/components/schemas/...
This resolves the invalid-ref error in CI for issue #12695.

### DIFF
--- a/docs/api/ref/schemas/product_ingredients.yaml
+++ b/docs/api/ref/schemas/product_ingredients.yaml
@@ -22,7 +22,7 @@ properties:
       type: string
 
   ingredients:
-    $ref: "./ingredient.yaml#/components/schemas/Ingredients"
+    $ref: "ingredient.yaml#/components/schemas/Ingredients"
   ingredients_analysis:
     type: object
     properties:


docs/api/ref/schemas/product_ingredients.yaml
```diff
@@ -22,7 +22,7 @@ properties:
       type: string
 
   ingredients:
-    $ref: "./ingredient.yaml#/components/schemas/Ingredients"
+    $ref: "ingredient.yaml#/components/schemas/Ingredients"
```
jayaddison, line 25: Agreed, this is in the same directory, so hopefully this will fix the linting reported in #12695.  If this does not work, then an alternative to try could be:

```suggestion
    $ref: "./schemas/ingredient.yaml#/components/schemas/Ingredients"
```

(a few other places seem to follow that pattern; this may not be necessary, though)

jayaddison:
How can we test this to confirm that it has fixed the problem?

sidhansu10:
> How can we test this to confirm that it has fixed the problem?

I verified the fix locally by running the documentation linter the same way CI does: the previous path reproduced the enoent error, updating the reference to ingredient.yaml#/components/schemas/Ingredients resolves the  issue of missing file and the ci pipeline for this has also passed all validation checks, confirming it is  valid. I can try the alternate path of ./schemas/ingredient.yaml#… , but as both files are in the same directory the current fix is correct

jayaddison:
@sidhansu10 should the validation check failure appear in recent pull request #12922?  I don't find it visible there.

sidhansu10:
@jayaddison the failing checks you see belong to the translations pull request #12922. This #12919 doesn’t touch translations or per; code, so the validations don’t run here. All checks have passed.

jayaddison:
@sidhansu10 the bugreport #12695 was about a situation where an `invalid-ref` error was raised in a `GitHub Actions / Lint (pull_request)` check annotation, and that was a translation pull request (#12507).  So I expected to find the same error in #12922 - but I can't find it there.

sidhansu10:
@jayaddison the invalid-ref error only appear when the linter processes product_ingredients.yaml. pull request #12922 doesn’t touch that file, so it won’t reproduce the error from #12695. this is why the issue shows up here but not in the translation pull request

jayaddison:
Thanks @sidhansu10; I think there are three items I'd want to resolve in future before considering approving/merging this, even though the fix does seem correct:

- [ ] Providing instructions for how to run the lint job locally, so that code reviewers and others can confirm that the error occurs without the fix, and is resolved once the fix (or similar alternative fixes) are applied.
- [ ] Confirming whether linter processing of `product_ingredients.yaml` is required in the process of producing the error.  The bugreport described the error occurring in pull request #12507, and that pull request only changed `.po`-suffix files and `.html`-suffix files.
- [ ] Confirming that the bug still occurs for recent pull requests.  Even if item (2) in this list is not required, we should be able to open a pull request that does modify some of the `.yaml` files, and the error should reappear.

I'm unable to find the error reported in #12695 appearing in recent pull requests, so I think that this can be closed.  Please feel free to re-open it if it does occur again.

sidhansu10:
sure @jayaddison

PR #12934, opened 2026-01-23
chore: Move to OIDC Level 3 for staging, local development and integration tests

john-gom:

### What

Set the OIDC Implementation Level to 3 for local development, integration tests and the Staging environment.

This will now use the Keycloak UI for login and registration but still use the existing PO forms for account management.

### DIFF
--- a/lib/ProductOpener/API.pm
+++ b/lib/ProductOpener/API.pm
@@ -1097,7 +1097,7 @@ sub process_auth_header ($request_ref, $r) {
 	my $user_ref = retrieve_user_using_token($access_token, $request_ref);
 	unless (defined $user_ref) {
 		$log->info('User not found and not created') if $log->is_info();
-		display_error_and_exit($request_ref, 'Forbidden.', 403);
+		display_error_and_exit($request_ref, 'Authentication error', 401);
 	}
 	my $user_id = $user_ref->{userid};
 	$log->debug('user_id found', {user_id => $user_id}) if $log->is_debug();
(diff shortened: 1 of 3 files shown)


john-gom:
Hi @hangy . Since you approved this I've updated most of the code to create users directly in Keycloak and use an Authorization header to simulate being logged in. Did you want to take another look?

lib/ProductOpener/API.pm
```diff
@@ -1097,7 +1097,7 @@ sub process_auth_header ($request_ref, $r) {
 	my $user_ref = retrieve_user_using_token($access_token, $request_ref);
 	unless (defined $user_ref) {
 		$log->info('User not found and not created') if $log->is_info();
-		display_error_and_exit($request_ref, 'Internal error', 500);
+		display_error_and_exit($request_ref, 'Forbidden.', 403);
```
hangy: I think 403 is for the case that the authentication worked, but the user isn't allowed? In this case, wouldn't it rather be 401 (user has to sign in)?
One issue with 401 might be automated retries that may not happen (as much) with 500

hangy:
> Hi @hangy . Since you approved this I've updated most of the code to create users directly in Keycloak and use an Authorization header to simulate being logged in. Did you want to take another look?

Thanks for the ping! I think it looks good. 👍

lib/ProductOpener/API.pm
john-gom: I'll change to 'Authentication error' 401 to be consistent with https://github.com/openfoodfacts/openfoodfacts-server/blob/b1af9a0b17e3db046b088a4e04f8daebc2594772/lib/ProductOpener/Auth.pm#L197

PR #12935, opened 2026-01-23
fix: resave dangerous svg icon

stephanegigandet:
For some reason the dangerous SVG icon is making svgo fail on a Windows setup:

see @Vic142 message on Slack: https://openfoodfacts.slack.com/archives/C02LDQDDD/p1769023787624129

maybe a byte order mark or something...


jayaddison:
NB: the error message encountered for this was `SvgoParserError: <input>:1:1: Non-whitespace before first tag.`, and it was emitted by the `gulp-svgmin` plugin used in our icon/asset processing `gulpfile.ts` pipeline (this hyperlink to the most-recent commit before the problem report.. but I don't know if that is precisely the commit it occurred from): https://github.com/openfoodfacts/openfoodfacts-server/blob/2388ccd0026c03164af631e38d21eadd6f87a92f/gulpfile.ts#L65-L80

Issue #12936, opened 2026-01-23
Keycloak password reverts back to last one set in Product Opener when user preferences are updated


john-gom:
### What
When the User preference form is updated in Product Opener and the user does not set their password then PO sends the current value of the encrupted_password stored in the user.sto file to Keycloak.

This will overwrite any password change made directly in Keycloak via https://auth.openfoodfacts.org/realms/openfoodfacts/account

This would only be a problem for apps that expose this form, like off-explorer

### Steps to reproduce the behavior
1. Go to https://auth.openfoodfacts.org/realms/openfoodfacts/account and change your password
2. Go to https://world.openfoodfacts.org/cgi/user.pl?type=edit&userid={your user id} and change something (not the password)
3. Log out and try to log back in again
4. Only your previous password (before step 1) will be accepted

### Expected behavior
- The password in Keycloak should not be updated if it is not changed in PO

dhiraj-143r:
@john-gom 
I have implemented a fix that changes this logic to only send credentials to Keycloak if:

The user explicitly enters a new password in the form.
The user does not exist in Keycloak yet (to support migrating legacy accounts).
If the password field is empty and the user exists, we now send an empty credential list, which preserves the current password in Keycloak.

I have opened a PR with this fix.

PR #12937, opened 2026-01-23
fix: Prevent Keycloak password overwrite on user preference updates

dhiraj-143r:
# Fix for Issue #12936

## Solution
Modified `lib/ProductOpener/Keycloak.pm` to conditionally send credentials to Keycloak.

The changes in `create_or_update_user` now:
1. **Send credentials if `$password` is defined:** (User explicitly sets a new password in PO form) -> Update Keycloak.
2. **Send credentials if user does NOT exist in Keycloak:** (New user creation or first sync) -> Migrate the legacy `encrypted_password`.
3. **Send NO credentials if password field is empty AND user exists:** -> Preserve the existing password in Keycloak.

## Changes
- Updated `create_or_update_user` in `lib/ProductOpener/Keycloak.pm` to handle credential generation logic.
- Updated the `credentials` field in the API request payload to be an empty array `[]` when no credential updates are needed, preventing the overwrite.

### DIFF
--- a/lib/ProductOpener/Keycloak.pm
+++ b/lib/ProductOpener/Keycloak.pm
@@ -158,14 +158,22 @@ We create the user properties file locally before, and we create the user in key
 =cut
 
 sub create_or_update_user ($self, $user_ref, $password = undef) {
-	my $credential
-		= defined $password
-		? {
-		type => 'password',
-		temporary => $JSON::PP::false,
-		value => $password
-		}
-		: convert_scrypt_password_to_keycloak_credentials($user_ref->{'encrypted_password'});
+	my $credential = undef;
+
+	if (defined $password) {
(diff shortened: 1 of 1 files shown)


john-gom:
Hi @dhiraj-143r , I had already started this in pr #12938.

Also I don't think your solution would work as we still need to allow password change via the Perl forms as they are currently used like an API by some consumers, such as mobile (something we will be deprecating in the longer term)

dhiraj-143r:
@john-gom. 
I didn't verify if another PR was in progress my bad. 
Good point about the Perl forms being used by the mobile apps; I had not  accounted for that dependency.

alexgarel:
thanks @dhiraj-143r for the proposal.

PR #12939, opened 2026-01-24
fix: checkpoint resume logic and migration script safety

mskandashyambhat:
### What was broken
Several migration scripts skipped the last processed item when resuming from a checkpoint, which could prevent failed or interrupted items from being retried. Some scripts also used fatal die statements and unsafe update ordering that could lead to crashes or inconsistent state.

### What I changed
- Fixed checkpoint resume logic to re-process the last processed item
- Corrected string vs numeric comparisons for checkpoint values
- Replaced fatal die statements with controlled error handling
- Improved ordering of filesystem and MongoDB updates to avoid data loss
- Added missing checks for undef values and file operation failures

### Why this is safe
- Changes are limited to internal migration scripts
- No public APIs or data formats were modified
- Existing successful workflows remain unchanged


scripts/migrations/2025_05_convert_sto_to_json.pl
```diff
@@ -43,8 +43,11 @@
 	if (not $can_process) {
 		if ($path eq $last_processed_path) {
 			$can_process = 1;
+			# Don't skip - we want to re-process the last item in case it failed
```
john-gom: I don't see the need for this change as the checkpoint is only updated after the product has been processed

scripts/migrations/2025_05_compare_sto_with_json.pl
```diff
@@ -54,8 +54,11 @@ ($path)
...
 		if ($path eq $last_processed_path) {
 			$can_process = 1;
+			# Don't skip - re-process the last item in case it failed
+		}
+		else {
+			next;    # Skip items before the checkpoint
 		}
-		next;    # we don't want to process the product again
 	}
 	$checkpoint->update($path);
```
john-gom: Suggest moving this to after the processing rather than changing the skip logic

mskandashyambhat:
Thanks for the detailed feedback.

That makes sense — I agree that touching many older migration scripts increases risk, especially where they may no longer be testable or relevant.

I’ll narrow the scope of this PR to focus only on the 2025_05 scripts and revert the other changes. For those scripts, I’ll move the checkpoint update to after processing instead of modifying the skip logic, as you suggested.

I’ll push an updated version shortly.

mskandashyambhat:
I’ve reverted the earlier commit that touched older migration scripts and narrowed this PR to only the 2025_05 changes.

The checkpoint update is now moved to after processing, as suggested.

mskandashyambhat:
I’ve reverted the older migration changes and limited this PR to the 2025_05 scripts only.
The checkpoint update is now moved to after processing, as suggested.
Could you please take another look when you have time?

mskandashyambhat:
Thanks for the review and suggestions.
Appreciate the guidance.

PR #12949, opened 2026-01-26
chore: taxonomy additions

aleene:
### What
more additions


taxonomies/food/categories.txt
```diff
@@ -35891,6 +35891,10 @@ nl: Broodjes
 sv: Bullars
 intake24_category_code:en: BRDR
 
+< en:Bread rolls
+en: Soft bread rolls
+nl: Zachte broodjes, puntjes
+
```
moon-rabbitOFF, line 35897: What are soft bread roll? Does it contain an ingredient, like milk, that makes it soft?

aleene: Like this one: https://nl.openfoodfacts.org/product/8710400084730/witte-puntjes-albert-heijn There is some dairy in it, but I would not say milk.

aleene: These are less soft and less sweet than the pains au lait, which seem to have 10% milk

aleene: Pains au lait could be under this category. Just wanted to make the distinction for this typical dutch bread.

PR #12956, opened 2026-01-26
taxonomy(brands): add Ribena

Freso:
https://en.wikipedia.org/wiki/Ribena

Needed-for: https://world.openfoodfacts.org/facets/brands/ribena


alexgarel (approved):
Thanks.

PR #12957, opened 2026-01-26
feat: edit checks for new nutrition schema

Vic142:
This PR aims to edit the nutrient checks when editing products in order to work with the new nutrition schema.

When a number entered for a new nutrient value a question mark with the corresponding warning appears.
image
image

In this state, when a value of a nutrient is first entered and then the unit is changed, the checks don't run again for the nutrient. This allows values to not be flagged as invalid when they are as shown below. If needed I can change the code so that the checks run when the units are changed.
image


stephanegigandet:
Thank you @Vic142 ! I tested it locally, everything looks fine.

> If needed I can change the code so that the checks run when the units are changed.

Yes, I think that would be useful.

Can you also put back the dangerous icon files?

PR #12959, opened 2026-01-27
fix: incorrect UTM language codes in donation pages

### LINKED ISSUE #12950: Donation pages use incorrect utm_content language codes for analytics tracking
### What
  While reviewing donation HTML pages, I noticed that many non-English files
(e.g. pt_PT.html, yi.html, ha.html) use utm_content=donate-page-share-en
even though the page content is not English.

This likely causes incorrect analytics attribution and makes it difficult
to measure which language versions drive donations.

Example:
- pt_PT.html uses donate-page-share-en
- yi.html uses donate-page-share-en

----

1. Open a non-English donation HTML file (e.g. pt_PT.html).
2. Inspect the og:url meta tag.
3. Observe that utm_content is set to donate-page-share-en.

-----

### Expected behavior
- utm_content should reflect the language (or language variant) of the donation page
so analytics can correctly attribute donations by language.

----

### Why
- Incorrect utm_content values cause analytics and campaign attribution to be wrong,
making it difficult to understand which language versions drive donations.

----

### Additional context
- This issue was observed across multiple non-English donation pages.
Happy to prepare a small, scoped PR once the expected utm_content format
is confirmed.

arybhatt4533:
This PR corrects the hardcoded English UTM codes in non-English donation pages (PT, RO, RU, SK) for both meta tags and fallback donorbox links. Closes #12950.
- Fixes #[12950]


jayaddison:
This seems helpful.  This pull request doesn't appear to update all affected language pages, though - `xh` for example.

AryaBhatt45:
Thanks for pointing that out! I'll do a global search and update all the remaining non-English language pages (including xh) to ensure the UTM codes match their respective languages. I'll push the updates shortly."
@jayaddison

html/donate/ro.html
```diff
@@ -219,7 +219,7 @@ <h3>Unde se duce donația ta</h3>
       <div class="columns small-12 medium-6 large-5 xlarge-4 xxlarge-3 text-center">
         <div style="max-width:425px">
           <div class="loading">
-            <p>Formularul de donație este în curs de încărcare.<br>Îl poți accesa și <a href="https://donorbox.org/your-donation-to-open-food-facts">făcând clic aici</a>.</p>
+            <p>Formularul de donație este în curs de încărcare.<br>Îl poți accesa și <a href="https://donorbox.org/your-donation-to-open-food-facts?utm_source=donate-page-share&utm_campaign=donation-2025&utm_content=donate-page-share-ro">făcând clic aici</a>.</p>
```
jayaddison, line 222: @arybhatt4533 @alexgarel I didn't notice this previously, but there are some `utm` query-string params added to external hyperlinks, like this one -- was that intentional?

AryaBhatt45: Hi @jayaddison, yes, that was intentional. The goal was to ensure that even the fallback links have the correct language-specific UTM parameters (like utm_content=donate-page-share-ro for Romania) so that the donation sources can be accurately tracked across different localized pages

jayaddison: Ok, thanks @arybhatt4533.  The way I understand it, those UTM parameters are received by the (remote) website/server that hosts them, and also potentially read by the (local) user web browser, if Google Analytics (or Matamo, or any other analytics engine that reads from those parameters) script tags are included in the webpage.

If so, then this PR would send those tracking/campaign details to `donorbox.org`'s analytics service/account -- in other words: I don't think that OpenFoodFacts would automatically have access to the resulting analytics data (DonorBox would have to agree/device to share that with us).

Maybe we do want to intentionally provide DonorBox with campaign information about OpenFoodFacts users that we have agreed with them -- but I'm not aware of an agreement there, and I don't see any mention of a bugreport/feature-request to implement support as a result of that, so I'm not sure whether we should include it on external hyperlinks.

For hyperlinks to `openfoodfacts.org` websites, where ~~Google Analytics~~ Matamo (ref: #10931) is enabled, I'd expect `utm_source` and so on to be included, and that would seem normal/fine to me.

AryaBhatt45: Thanks for the detailed explanation, @jayaddison! You're right, I added those to maintain consistency with how tracking parameters are often used, but I wasn't aware of the specific data-sharing policy regarding external links like DonorBox.
Since the PR is already merged, should I create a follow-up PR to remove these UTM parameters from the external DonorBox links to ensure we're not sending unnecessary tracking data? I'm happy to fix it right away.

jayaddison: @arybhatt4533 yes please - thank you.

PR #12960, opened 2026-01-27
fix: fix API call to brevo for newsletter subscription

alexgarel:
API call was not working because listID needs to be an int.
We also update the attributes fields names and values (it was not matching the one we had on Brevo so they were dismissed).
Also added config for docker to be able to test Brevo API.

This PR also activate Redis for local development as this is needed by quite a lot of feature (including newsletter subscription).


alexgarel:
I took the opportunity to verify config on the different instances, it's ok.

teolemon (approved):
Thank you so much @alexgarel 🦉📬

PR #12970, opened 2026-01-29
fix(openapi): Require fields in product_meta schema

### LINKED ISSUE #12969: openapi: mark required params as such in product_meta.yaml
### What
https://github.com/openfoodfacts/openfoodfacts-server/blob/3577767c904d385ec58785c176fef4d29e6c0de6/docs/api/ref/schemas/product_meta.yaml#L6

This should be required (AFAIK). The automatic TS code generator otherwise inserts `number | undefined`

VaiTon:
I'm not entirely sure if these fields are always there or not.

Fixes #12969

### DIFF
--- a/docs/api/ref/schemas/product_meta.yaml
+++ b/docs/api/ref/schemas/product_meta.yaml
@@ -1,6 +1,13 @@
 type: object
 description: |
   Metadata of a product (author, editors, creation date, etc.)
+required:
+  - created_t
+  - creator
+  - last_modified_by
+  - last_modified_t
+  - last_updated_t
+  - rev
 
 properties:
   created_t:


stephanegigandet:
@VaiTon the fields probably are in most products, but I'm not sure I understand why it's required to mark them "required". Why would it cause an issue for the TS generator?

VaiTon:
> Why would it cause an issue for the TS generator?

Because if they are not marked as required the generator outputs something like

```
rev: number | undefined
```

which means that I have to check / assert `!= null` every time I want to use that field. In particular I was working on `created_t`, which I think is always defined (correct me if I'm wrong).

If none of these fields are always defined then we can drop this PR and I'll update explorer code to handle `null` cases.

VaiTon:
@stephanegigandet gentle reminder :)

PR #12975, opened 2026-01-29
fix: new nutrition: compute energy from other nutrients + salt from/to sodium

stephanegigandet:
Initially we said we could have a separate input set with a "computed" source for thinks like energy computed from macro nutrients, but in fact it's something we really want to have on all input sets (e.g. packaging, manufacturer, usda etc.).

One option was to have a separate nutrient like energy-kcal-computed, but I think it's much better to have a value_computed field, so that we don't duplicate nutrients and pollute the nutrients taxonomy.


alexgarel (approved):
Look goods, thank you !

PR #12977, opened 2026-01-29
taxonomy(product): add museum ticket category

Freso:
Also adds a new "Tickets" category to encompass both event and museum (and future other) tickets.

Request: https://github.com/openfoodfacts/open-prices-frontend/issues/1048#issuecomment-3658056403


raphodn:
oh that's cool !!
I only see this PR now.
I'll see how we can integrate this into Open Prices and fetch non food categories as well :heart:

PR #12981, opened 2026-01-30
fix(openapi): Require created_t in product_meta schema

sidhansu10:
this pull request  marks the `created_t` field as required in the product_meta openapi schema. The field is always present in the api , but it was not included in the `required` list of  schema. Because of this, code generators  were treating it as optional and producing `number | undefined`.
updating  the schema makes the openapi spec consistent with actual api behavior.
Fixes #12970.


alexgarel:
Thanks @sidhansu10, in the mean-time I merged https://github.com/openfoodfacts/openfoodfacts-server/pull/12970 that fixes it.

PR #12994, opened 2026-02-01
chore: migrate gulpfile to mjs

VaiTon:
Migrates Gulpfile to MJS. We were using TypeScript just for that (and Gulp doesn't even support it by default).

Supersedes https://github.com/openfoodfacts/openfoodfacts-server/pull/11506


package.json
hangy: These are probably now redundant, too

hangy (approved):

PR #12995, opened 2026-02-01
chore: Make Dockerfile.frontend usable on Docker rootless

VaiTon:
Standard Docker setups require matching the container's internal UID to the host's UID (e.g., 1000) to ensure write permissions on bind-mounted directories. However, in Rootless Mode, the host user is automatically mapped to UID 0 inside the container.

Attempting to force-assign UID 0 to a secondary user (like node) via usermod causes the build to fail because UID 0 is already reserved.

This PR introduces a conditional build flow: Only attempts to modify the users if the target USER_UID is not 0.

This allows to set `USER_UID=0` and still build and run the container in a rootless setting.

### DIFF
--- a/Dockerfile.frontend
+++ b/Dockerfile.frontend
@@ -6,27 +6,33 @@ ARG USER_GID=1000
 FROM node:22.22.0 AS builder
 ARG USER_UID
 ARG USER_GID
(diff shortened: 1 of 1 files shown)


VaiTon:
I have to think about this again, because the user inside docker still gets mapped to a different user

PR #13004, opened 2026-02-03
fix: use page-type 'edit' for logged in user

alexgarel:
on folksonomy component,
as new webcomponent does not enable edition in view mode.

Also add a button to invite to signin to edit properties.


po/common/common.pot
```diff
@@ -720,6 +720,10 @@ msgctxt "edit_product_page"
 msgid "Edit the page"
 msgstr ""
 
+msgctxt "signin_to_edit_fe_properties"
+msgid "Signin to add or edit properties"
```
stephanegigandet:
```suggestion
msgid "Sign-in to add or edit properties"
```

PR #13010, opened 2026-02-04
taxonomy: Chocolates

moon-rabbitOFF:
Created categories of chocolate for pastry


Freso (commented):
What’s the reason for the removal of the `Dark chocolate paste for cooking` category? Is it the same as drops and should be added as an alias/synonym for `Dark chocolate drops`?

moon-rabbitOFF:
` Dark chocolate paste for cooking` is the same as `Dark chocolate bar for cooking with more than 40% cocoa`. I didn't see the use of making it a synonym since the category is almost empty. I don't want to fill the taxonomy with plenty of synonyms from previous errors

stephanegigandet (approved):
Thank you!

PR #13015, opened 2026-02-05
chore: Ingredients latvian

aleene:
### What
first set of translations

### DIFF
--- a/cgi/display.pl
+++ b/cgi/display.pl
@@ -32,7 +32,7 @@ use ProductOpener::Display qw/:all/;
 use ProductOpener::HTTP qw/single_param redirect_to_url/;
 use ProductOpener::Users qw/$Owner_id init_user/;
 use ProductOpener::Lang qw/lang/;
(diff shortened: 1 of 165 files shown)


stephanegigandet:
perltidy has an error (the error seems to be in perltidy, not necessarily in the code changes in Ingredients.pm ... 

```
Run make check_perltidy
🥫 Checking with perltidy lib/ProductOpener/Ingredients.pm
test -z "lib/ProductOpener/Ingredients.pm" || COMPOSE_FILE="docker-compose.yml;docker/dev.yml" docker compose --env-file=.env  run --rm --no-deps backend perltidy --assert-tidy -opath=/tmp/ --standard-error-output lib/ProductOpener/Ingredients.pm
Use of strings with code points over 0xFF as arguments to bitwise xor (^) operator is not allowed at /opt/perl/local/lib/perl5/Perl/Tidy.pm line 3[4](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/21939671474/job/63362548465?pr=13015#step:8:5)61.
```

PR #13029, opened 2026-02-06
taxonomy(food): sichuan pepper edits

Freso:
Seems like Sichuan pepper was previously mixed in with chili peppers, depite them not being related. This expands on both the category and ingredient entries for Sichuan pepper and also adds one of the vegetable sources for Sichuan pepper: Japanese pepper, which is still not a pepper.

Source: https://en.wikipedia.org/wiki/Sichuan_pepper
Source: https://en.wikipedia.org/wiki/Zanthoxylum_piperitum
Needed-for: https://se.openfoodfacts.org/product/7350116923995/sichuanpeppar-eaglobe

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -76003,14 +76003,6 @@ fr: Pulla
 fr: Serrano
 
 < en:Chili peppers
-en: Sichuan pepper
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Thanks

PR #13031, opened 2026-02-07
fix: set default aluminium weight to light #12696

### LINKED ISSUE #12696: Default weight for aluminium pack is heavy --> should be light

arybhatt4533:
As per the issue discussion, I've updated the logic to handle en:unknown shapes and set the default fallback weight for aluminium to light. This ensures the worst-case scenario is correctly applied. This is a clean PR following a branch reset. 
Fix[#12696]
@teolemon

### DIFF
--- a/lib/ProductOpener/EnvironmentalScore.pm
+++ b/lib/ProductOpener/EnvironmentalScore.pm
@@ -1710,14 +1710,18 @@ sub compute_environmental_score_packaging_adjustment ($product_ref) {
 
 					my $weight = "thin";
 
-					if (defined $packaging_ref->{shape}) {
-						$weight = get_inherited_property("packaging_shapes", $packaging_ref->{shape}, "weight:en");
-						$log->debug("aluminium", {weight => $weight, shape => $packaging_ref->{shape}})
-							if $log->is_debug();
-						if (not defined $weight) {
-							$weight = "heavy";
-						}
-					}
+					# 1. Unknown shape handle karo
+               if (not $packaging_shape or $packaging_shape eq 'en:unknown') {
+                  $weight = get_inherited_property("packaging_shapes", $packaging_ref->{shape}, "weight:en");
+                  $log->debug("aluminium", {weight => $weight, shape => $packaging_ref->{shape}})
+               if $log->is_debug();
+                }
+
+               # 2. Aluminium fallback ko alag rakho (taaki ye hamesha apply ho sake)
+               # # 2. Aluminium fallback ko alag rakho
+               if ($packaging_ref->{material} eq 'en:aluminium') {
+                  $weight_fallback = 'light';
+               }
 
 					if ($weight eq "heavy") {
 						$packaging_ref->{material} = "en:heavy-aluminium";


AryaBhatt45:
Hi @teolemon, I have force-pushed a clean version of this fix after resetting the branch history. The logic now correctly handles en:unknown shapes and applies the light fallback for aluminium. I noticed some Perl test failures, but they seem related to existing upstream issues as SonarCloud shows no new issues in my changes. Ready for review!

lib/ProductOpener/EnvironmentalScore.pm
```diff
@@ -1710,14 +1710,18 @@ sub compute_environmental_score_packaging_adjustment ($product_ref) {
...
 
-					if (defined $packaging_ref->{shape}) {
-						$weight = get_inherited_property("packaging_shapes", $packaging_ref->{shape}, "weight:en");
-						$log->debug("aluminium", {weight => $weight, shape => $packaging_ref->{shape}})
-							if $log->is_debug();
-						if (not defined $weight) {
-							$weight = "heavy";
-						}
-					}
+					# 1. Unknown shape handle karo
```
Freso, line 1713: I don’t think this comment is needed. `not $packaging_shape or $packaging_shape eq 'en:unknown'` already says the same thing. (Also, “karo”?)

```suggestion
```

lib/ProductOpener/EnvironmentalScore.pm
```diff
@@ -1710,14 +1710,18 @@ sub compute_environmental_score_packaging_adjustment ($product_ref) {
...
+               if (not $packaging_shape or $packaging_shape eq 'en:unknown') {
+                  $weight = get_inherited_property("packaging_shapes", $packaging_ref->{shape}, "weight:en");
+                  $log->debug("aluminium", {weight => $weight, shape => $packaging_ref->{shape}})
+               if $log->is_debug();
+                }
+
+               # 2. Aluminium fallback ko alag rakho (taaki ye hamesha apply ho sake)
+               # # 2. Aluminium fallback ko alag rakho
```
Freso, line 1721: With very few exceptions, code comments should be in English.

lib/ProductOpener/EnvironmentalScore.pm
```diff
@@ -1710,14 +1710,18 @@ sub compute_environmental_score_packaging_adjustment ($product_ref) {
...
+               if ($packaging_ref->{material} eq 'en:aluminium') {
```
Freso, line 1722: We’re already inside an `if ($packaging_ref->{material} eq "en:aluminium")` clause, so this is a redundant `if` statement.

lib/ProductOpener/EnvironmentalScore.pm
```diff
@@ -1710,14 +1710,18 @@ sub compute_environmental_score_packaging_adjustment ($product_ref) {
...
+                  $weight_fallback = 'light';
```
Freso, line 1723: Is `$weight_fallback` used anywhere else?

Freso (changes requested):
Thank you for taking this on! Have you tested this on your/a running instance to verify that it works? I think it would be good to also include a test case for this to ensure both that this works, but also that future changes doesn’t cause a regression to this behaviour.

teolemon:
- Dupe of https://github.com/openfoodfacts/openfoodfacts-server/pull/12809

Issue #13034, opened 2026-02-08
Bug: Application crash on signout due to debug `die` statement in Auth.pm


dhiraj-143r:
### Description

I was analyzing the OIDC authentication flow in `lib/ProductOpener/Auth.pm` and noticed a critical issue in the `start_signout` function.

There appears to be a debug statement leftover in the production code that will cause the application to crash whenever a user signs out with a `return_url` parameter.

### Expected Behavior

The application should continue the signout process:
1.  Read the `return_url`.
2.  Validate it against the allowed domains.
3.  Redirect the user to the OIDC end session endpoint.

john-gom:
@dhiraj-143r , It would be good to see the steps to reproduce the problem and the observed symptoms so we can check it is fixed.

dhiraj-143r:
@john-gom ,

here are the Steps to Reproduce:
1. Ensure you have a running instance of the openfoodfacts-server (local or staging)
2. Log in with valid OIDC credentials
3. Trigger a sign-out by navigating to the logout endpoint with a `return_url` parameter
   -Example URL: `/cgi/logout.pl?return_url=https://world.openfoodfacts.org/`
   -Or any signout flow that includes the return URL parameter

dhiraj-143r:
here are the Observed Symptoms due to this :
1) Browser Behavior: The application will terminate immediately, resulting in an **HTTP 500 Internal Server Error** displayed to the user
2) User Impact: Users cannot complete the sign-out process successfully and are not redirected to their intended destination
3) Server Logs: The error log shows a crash originating from `lib/ProductOpener/Auth.pm` at line 421, with output containing the `return_url` value

LakshuCodes:
Hi,
I’d like to work on this issue.

Please let me know if it’s okay to proceed and assign it to me — once confirmed, I’ll start working on the issue and preparing a fix. Thanks!

dhiraj-143r:
@LakshuCodes ,
 I’m currently waiting for confirmation from the maintainer and I’m working on it.
In the meantime, you can choose any issue labeled “good first issue” from this list to work on:
https://github.com/openfoodfacts/openfoodfacts-server/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22

john-gom:
[...]

PR #13035, opened 2026-02-08
fix: remove debug die statement causing crash on signout

dhiraj-143r:
### What

Removed a debug `die` statement from `lib/ProductOpener/Auth.pm` that was accidentally causing the application to crash during the signout process.

#13034


lib/ProductOpener/APIProductRead.pm
```diff
@@ -55,8 +55,6 @@ use ProductOpener::URL qw(format_subdomain);
 use ProductOpener::Images qw(add_images_urls_to_product);
 use ProductOpener::Store qw/retrieve_object/;
 
-my $cc;
```
john-gom, line 58: This is still being referenced later in the file.

dhiraj-143r: @john-gom 
I have restored it

PR #13036, opened 2026-02-08
taxonomy(food): improving frozen food categories

Freso:
A number of frozen food categories are currently not children of “Frozen foods”, and thus not counted/considered for the current Open Prices community challenge: https://prices.openfoodfacts.org/challenges/10

I tried doing all the dead animal categories too, but after doing a few of them, I started feeling ill and upset, so I skipped through those, leaving those for someone else to fix.

This includes new “Frozen fruit and vegetable juices” and “Frozen fruit juices” categories separate from “Frozen foods”.

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -8217,6 +8217,7 @@ hr: Ohlađeni sokovi od cijeđene naranče
 lt: Šaltai laikomos spaustos apelsinų sultys
 nl: Gekoelde verse sinaasappelsappen, Gekoelde verse sinaasappelsap
 
+< en:Frozen fruit juices
 < en:Orange juices
 en: Frozen orange juices
 fr: Jus d'orange surgelé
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -122704,11 +122780,20 @@ ciqual_food_code:en: 2048
 ciqual_food_name:en: Fruit juice, mixed - apple base, standard
 ciqual_food_name:fr: Jus multifruit - base pomme, standard
 
-< en:fruit-juices
+#< en:Frozen foods
+< en:Fruit and vegetable juices
+en: Frozen fruit and vegetable juices, Frozen juices
```
Freso, line 122793: Not sure if this should be a child of `en:Frozen foods` or not. It’s not strictly a _food_, but it could be used in cooking, and we don’t have a “Frozen foods and beverages” (though I considered making one, similar to this category).

Freso:
I don’t see how `ERROR - circular_parent - 117449 - en:frozen-cereals is a parent of itself`. `en:frozen-cereals` is a child of `en:Frozen cereals and their products` which in turn is a child of `< en:Cereals and their products` and `< en:Frozen plant-based foods`, neither of which have any parents that are children of `en:frozen-cereals` as best as I can tell.

jayaddison:
I couldn't find any obvious reasons for the detected circular dependency either.  I wondered whether the GPC category codes could somehow be used to override child/parent category mappings -- but I don't think they do.  And I also wondered whether the fact that two similarly-named frozen vegetable product categories exist, one with the suffix `mixed`, could somehow be resulting in an anomaly, but I'm not sure how/why that would happen either.

PR #13038, opened 2026-02-08
fix: Remove unused Action::CircuitBreaker and Action::Retry dependencies

hangy:
### What

`Action::Retry` and `Action::CircuitBreaker` were essentially unused due to `max_retries_number` being set to `0`. We should just remove them.

### Screenshot

N/A

N/A


alexgarel (approved):
Thanks a lot !

PR #13051, opened 2026-02-10
fix: clarify search map wording to use EMB traceability codes

### LINKED ISSUE #21: "Products search" page generate maps using "EMB code" but use the term "production place"
It is confusing to use "production place" instead of "EMB code".
"production place" will likely be confuse with "Manufacturing or processing places".

Mounika-max-348:
### What

This PR fixes misleading wording in the search map section.

The map is generated using EMB (traceability/packager) codes, but the UI text referred to them as "production place", which could be confused with manufacturing or processing locations.  
The wording has been updated to correctly reflect that the map is based on EMB traceability codes.

### Screenshot
Not applicable (text-only change).

- Fixes #21

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -111,8 +111,6 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt \
         #
         # cpan dependencies that can be satisfied by apt even if the package itself can't:
         #
-        # Action::Retry
-        libmath-fibonacci-perl \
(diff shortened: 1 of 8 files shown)


Mounika-max-348:
Hi, I’ve updated the search map text to correctly reference EMB (traceability) codes instead of production place.  
Please let me know if any changes are needed. Thanks!

PR #13055, opened 2026-02-11
fix: Issues running pro platform locally in docker

john-gom:

### What

Unable to follow pro platform instructions since postgres was added as a dependency to backend (it shares the instance with off). Removed this dependency and re-added explict start in the Makefile.

Also fixed an issue where images were not transferred when exporting from Pro platform to the main database because the product_images volume was not shared.

### DIFF
--- a/env/env.off-pro
+++ b/env/env.off-pro
@@ -17,3 +17,4 @@ APACHE_ENVVARS=/etc/apache2/off-envvars
 # app configurations
 PRODUCT_OPENER_FLAVOR=openfoodfacts
 PRODUCT_OPENER_FLAVOR_SHORT=off
+QUERY_URL=


docker/dev.yml
```diff
@@ -138,6 +137,8 @@ volumes:
     name: ${PO_COMMON_PREFIX:-}po_orgs
   export_files:
     name: ${PO_COMMON_PREFIX:-}po_export_files
+  product_images:
+    name: ${PO_COMMON_PREFIX:-}po_off_product_images
```
alexgarel, line 141: For product_images, it's not that simple… normally public platform must be able to access pro images, but not in the same volume as it's own images.

This way it will work still as folder organization is a bit different on both platform. But that could not be a production setup.

john-gom: I think it works because the pro platform puts the images in a sub folder for each producer, so there is no clash

PR #13056, opened 2026-02-11
chore: Reduce log volume

john-gom:

### What

In development the current debug logs are massive which makes them difficult to use. They also contain potentially sensitive data.

This PR removes potentially large and sensitive items from log files and comments out log statements that generate many log lines

### DIFF
--- a/.test_groups_cache/unit_groups.json
+++ b/.test_groups_cache/unit_groups.json
@@ -12,10 +12,11 @@
       "tests/unit/nutrition.t",
       "tests/unit/packaging_stats.t",
       "tests/unit/products.t",
-      "tests/unit/store.t",
-      "tests/unit/templates.t"
+      "tests/unit/spam_user.t",
+      "tests/unit/taxonomy_suggestions.t",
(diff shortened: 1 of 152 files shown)


lib/ProductOpener/API.pm
```diff
@@ -203,7 +242,8 @@ sub read_request_body ($request_ref) {
 	}
 	$request_ref->{body} = $content;
 
-	$log->debug("read_request_body - end", {request => $request_ref}) if $log->is_debug();
+	# Don't log payload as could be huge
+	# $log->debug("read_request_body - end", {request_ref => $request_ref}) if $log->is_debug();
```
hangy: What do you think about having a config toggle for these commented out log statements? That way, it would still get checked by `make check_perl`

john-gom: I'm not sure if that's making things overcomplicated. I could just delete the logging rather than comment it out as I suspect it was only useful while the code was originally being written

alexgarel: I totally agree.

Issue #13041, opened 2026-02-09
Restrict display of map for categories with wikidata property without an OSM property


teolemon:
### What
- Facets that shouldn't display a map do display a map
- https://world.openproductsfacts.org/facets/categories/smartphones
<img width="966" height="926" alt="Image" src="https://github.com/user-attachments/assets/7fab6c66-95a1-4db4-b8c0-12ad061706da" />

dhiraj-143r:
@teolemon ,
I'd like to work on this issue. Could you clarify what should be the rule for hiding the map? Should it be hidden for all Open Products Facts facets, only for specific categories without geographic relevance, or only when products lack geographic data?

stephanegigandet:
This happens because we display a map for tags that have a wikidata property. At runtime, display-tag.js makes a request:

```
async function getOpenStreetMapFromWikidata(id) {
  const endpointUrl = 'https://query.wikidata.org/sparql';
  const sparqlQuery = `SELECT ?OpenStreetMap_Relations_ID WHERE {
    wd:${id} wdt:P402 ?OpenStreetMap_Relations_ID.

```
And then if there are OSM relations, they are added to the map.

e.g. https://world.openfoodfacts.org/facets/categories/fr:fitou

This makes sense for food categories that have geographic indication of origins, it's less useful for smartphones.

One way to solve this would be to hide the map until we actually get OSM properties.

It could be a good thing to also restrict the feature to specific tags. (e.g. to start with, only when product_type is "food" and only for the tag type is "categories"). There's no point in having browsers send useless queries to wikidata.org

PR #13060, opened 2026-02-11
fix: only display map for geo-relevant facets (countries, origins, manufacturing_places, emb_codes)

sidhansu10:
### Summary
This pull request fixes issue #13041 where non-geographic facets  were incorrectly displaying a world map.

### Cause
The map was rendered whenever `countries_map_data` contained entries, even for non-geographic facets.

### Fix
Introduced a whitelist of geo-related facets:
- countries 
- origins 
- manufacturing_places 
- emb_codes

Maps now only render when:
1. The facet is geo-relevant, and 
3. map data exists.

### DIFF
--- a/docs/api/ref/schemas/product_meta.yaml
+++ b/docs/api/ref/schemas/product_meta.yaml
@@ -2,6 +2,9 @@ type: object
 description: |
   Metadata of a product (author, editors, creation date, etc.)
 
+required:
+  - created_t
+
 properties:
   created_t:
     type: integer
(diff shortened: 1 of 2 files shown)


stephanegigandet:
That's not where the issue is. Before solving an issue, it's really important to reproduce it first, otherwise there is absolutely no way to know if something will fix it. I added more details in the bug: https://github.com/openfoodfacts/openfoodfacts-server/issues/13041#issuecomment-3895932976

teolemon:
fixed already it seems. closing as a result

PR #13062, opened 2026-02-11
taxonomy: tajine

moon-rabbitOFF:
Created a new category "vegetarian tajine with side dishes" and added "chicken tajine" to "tajine".


Freso (approved):
Looks good to me!

PR #13069, opened 2026-02-12
feat: folksonomy autocomplete suggestions (percent labels)

pranjal2004838:
This PR extracts the frontend enhancement for the Folksonomy autocomplete: it computes frequency-based percentages for suggested values and returns autocomplete items as `{label,value}` (e.g. `yes (67%)`).

This change is a new feature and is intentionally opened as a separate PR on its own feature branch.

Files changed: `html/js/folksonomy.js`.

Please review only the frontend suggestion behavior; no backend changes are included in this PR.

### DIFF
--- a/html/js/folksonomy.js
+++ b/html/js/folksonomy.js
@@ -277,19 +277,26 @@ function displayFolksonomyPropertyValues() {
                     ]
                 */
                 const list = findOcc(json, "v");
+                if (list.length === 0) {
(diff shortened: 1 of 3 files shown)


TaciteOFF (changes requested):
This is now outdated. 
It uses the former jQuery form. 
We now use the folksonomy-editor component. 

The PR also modifies the producer platform page for no reason?

PR #13073, opened 2026-02-13
fix: restrict map display to food product categories

### LINKED ISSUE #13041: Restrict display of map for categories with wikidata property without an OSM property

dhiraj-143r:
I have Added a simple guard in display_tag() in Display.pm  that checks two conditions before adding a Wikidata object to the map data:
1) $options{product_type} eq 'food' — maps are only relevant for food products
2) $tagtype eq 'categories' — categories like wine regions or cheese appellations can have geographic boundaries, but brands, ingredients, etc. generally don't

If both conditions aren't met, we just skip adding the wikidata object, so the frontend never makes the SPARQL query and the map section stays hidden.

Fixes #13041

### DIFF
--- a/.test_groups_cache/unit_groups.json
+++ b/.test_groups_cache/unit_groups.json
@@ -9,12 +9,12 @@
       "tests/unit/ingredients_analysis.t",
       "tests/unit/ingredients_parsing_todo.t",
       "tests/unit/lang.t",
-      "tests/unit/nutrition.t",
-      "tests/unit/packaging_stats.t",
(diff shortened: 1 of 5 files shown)


stephanegigandet:
@dhiraj-143r can you run "make update_tests_results" to update the test results?

stephanegigandet (approved):
Thank you!

PR #13075, opened 2026-02-13
fix: rename 'Suffering Fingerprint' to 'Suffering Footprint'

### LINKED ISSUE #12690: Rename Suffering Fingerprint to Suffering Footprint

jayaddison:
### What
Renames the `Suffering Fingerprint` label in our translation string resources to `Suffering Footprint`.

This was done using the following command-line:

```
$ grep -rwl 'Suffering Fingerprint' | xargs sed -e 's#Suffering Fingerprint#Suffering Footprint#g' -i {}
```

(expanded: `grep` with options `--recursive`, `--word-regexp` and returning `--files-with-matches` (only the filenames, not the lines containing the pattern-matches) -- and then that list of files redirected to `sed`, a text stream editor, with a single command `s` pattern-matching the text `Suffering Footprint`, as indicated at the commandline by the delimiter `#` characters, globally within each file as configured by the terminating `g` character.  sed usually produces the resulting output at the command-line -- but using the `-i` flag, short for `--in-place`, means that it edits the input file on the filesystem instead of displaying the results).

### Screenshot
N/A

- Fixes #12690.

Edit: update the checklist.
Edit: fixup for a typo in the explanation.

### DIFF
--- a/po/common/aa.po
+++ b/po/common/aa.po
@@ -7628,7 +7628,7 @@ msgid "Those preferences are based on Open Food Facts' understanding of the ingr
 msgstr ""
 
 msgctxt "external_sources_empreinte_souffrance_name"
-msgid "Suffering Fingerprint"
+msgid "Suffering Footprint"
 msgstr ""
 
 msgctxt "external_sources_empreinte_souffrance_description"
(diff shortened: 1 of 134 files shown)


alexgarel (approved):

jayaddison:
Thanks @alexgarel!

PR #13078, opened 2026-02-14
test: add integration test for taxonomy file structural validation

Kunal241207:
### What

This PR adds an integration test that validates taxonomy files to prevent syntax regressions introduced through manual edits.

The test performs lightweight structural validation (duplicate properties, malformed lines, unexpected line formats) and is intentionally independent of the taxonomy parsing logic. Its goal is early detection of common errors during CI rather than full parser validation.

This change affects testing only and does not modify runtime behavior.

### DIFF
--- a/tests/integration/taxonomy_lint.t
+++ b/tests/integration/taxonomy_lint.t
@@ -4,73 +4,53 @@ use warnings;
 use Test::More;
 use File::Find;
 use File::Basename;
+use File::Spec;
+use FindBin;
+use IPC::Open3;
+use IO::Select;
+use Symbol qw(gensym);
 
 # ------------------------------------------------------------
-# Lightweight taxonomy validation
+# Taxonomy validation
 #
-# This test performs minimal structural checks to prevent
-# obvious syntax regressions in taxonomy files.
-#
-# It intentionally does NOT reimplement the full taxonomy
-# parser and is independent from developer tooling.
+# This test runs the canonical taxonomy linter in --check mode
+# to prevent syntax regressions in taxonomy files.
 # ------------------------------------------------------------
 
-sub lint_file {
-	my ($file) = @_;
-
-	open my $fh, '<:encoding(UTF-8)', $file
-		or return ["Could not open file: $!"];
-
-	my @errors;
-	my %seen_props;
-	my $line_num = 0;
-
-	# language prefix (lightweight check only)
-	my $lang_re = qr/(?:[a-zA-Z]{2,3}(?:[-_][a-zA-Z0-9]{2,8})*|xx)/;
-
-	while (my $line = <$fh>) {
-		$line_num++;
-		chomp $line;
-		$line =~ s/\s+#.*$//;    # strip inline comments only when preceded by whitespace
-		$line =~ s/^\s+|\s+$//g;
-
-		next if $line eq '' || $line =~ /^#/;
-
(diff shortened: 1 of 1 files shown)


john-gom:
I thought we already did some validation as part of make lint?

Kunal241207:
Yes @john-gom  - `make lint` does run taxonomy validation tooling, but this PR adds a lightweight, self-contained test that runs under the test suite/CI.

It is intentionally independent of the lint scripts and focuses on a minimal structural sanity check for taxonomy files. The goal is early detection of common syntax regressions during pull requests, rather than full parser or tooling validation

john-gom:
I still don't see he benefit of adding another workflow. Make lint runs quickly to me

Kunal241207:
@john-gom Thanks for the feedback. The intent is to ensure basic taxonomy checks run automatically in CI on every PR, rather than relying on contributors running make lint locally. It’s meant as a lightweight guardrail in the test suite, not a replacement for existing lint tooling.

john-gom:
In that case it would be better to link to the existing linting code from CI rather than duplicating the logic

Kunal241207:
@john-gom I updated the test to invoke the existing linting code (lint_taxonomy.pl --check) rather than duplicating logic. This keeps the behavior consistent with the current lint tooling while running as part of the test suite in CI.

Kunal241207:
@john-gom @teolemon Hi, Could anyone review this whenever they have time.

Freso:
How is this different from [`make check_taxonomies`](https://github.com/openfoodfacts/openfoodfacts-server/blob/dccfe737ba30fc0a5675c7c8d3c1280e6dfc5440/Makefile#L454-L457) that is [currently being run on PRs](https://github.com/openfoodfacts/openfoodfacts-server/blob/dccfe737ba30fc0a5675c7c8d3c1280e6dfc5440/.github/workflows/pull_request.yml#L388-L389)? If the goal of this is to run checks on taxonomy files during CI, we already do that (and you’re also not changing any CI files to actually run the code you’re adding).

Is this supposed to allow for running taxonomy linting locally without having to pull in docker, redis, mongodb, …? Because if so, your previous comments are not making that very obvious. :)

Kunal241207:
Yes, the goal is to catch obvious structural issues in taxonomy files early during local development without requiring the full Docker/Perl setup. This test is intentionally independent and lightweight, while make check_taxonomies provides deep full parsing in CI.

Freso:
`make check_taxonomy` literally runs `lint_taxonomy.pl --check` just the same as your script, as far as I can tell? In the CI, the docker is already set up, so running these checks through that shouldn’t incur too much overhead there.

Changes to taxonomy files may break Perl unit/integration tests, so the CI needs to run those too, it can’t only do syntax linting.

And also, again, you have not changed any CI files to show how/where/when you would be running this code in the CI workflow.

-----

If your goal is to make CI return results faster, another approach is looking into simply running fewer things that are not affected by taxonomy-only changes. I did some initial work on this in https://github.com/openfoodfacts/openfoodfacts-server/pull/13209 (and follow-up fix to make it actually work as intended in https://github.com/openfoodfacts/openfoodfacts-server/pull/13220 ).

Kunal241207:
Fair points. Closing this PR. I'll look into the CI optimization work you mentioned instead. Thanks for guidance!

Issue #13039, opened 2026-02-09
Some fixable warnings from production


alexgarel:
I just use `tail -n 100000  modperl_error_log |cut -d "]" -f 2| sed -e 's/Wide character [^ ]\+/Wide character .../g'|sort |uniq -c|sort` in `/mnt/off/logs/off` on off server to get some errors that could be avoided.

Here they are (fixing the latest have a highest impact):
```bash
tail -n 100000  modperl_error_log |cut -d "]" -f 2| sed -e 's/Wide character [^ ]\+/Wide character .../g'|sort |uniq -c|sort
      1  -e: Argument "id" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 147.
      1  -e: Argument "id" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 91.
      1  -e: Argument "percent_estimate" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 147.
      1  -e: Argument "percent_estimate" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 91.
      1  -e: Argument "rank" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 147.
      1  -e: Argument "rank" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 91.
      1  -e: Argument "text" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 147.
      1  -e: Argument "text" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 91.
      1  -e: Argument "text_en" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 147.
      1  -e: Argument "text_en" isn't numeric in array or hash lookup at /usr/local/share/perl/5.32.1/Data/DeepAccess.pm line 91.
      1  -e: Use of uninitialized value $value in string ne at /srv/off/cgi/product.pl line 1245.
      1  -e: Use of uninitialized value in addition (+) at /srv/off/lib/ProductOpener/Images.pm line 2327.
[...]

PR #13092, opened 2026-02-19
fix: Add null guards for remaining high-priority warnings (#13039)

dhiraj-143r:
This PR addresses the remaining high-priority production warnings identified in issue #13039


lib/ProductOpener/APIProductRead.pm
```diff
@@ -55,7 +55,7 @@ use ProductOpener::URL qw(format_subdomain);
 use ProductOpener::Images qw(add_images_urls_to_product);
 use ProductOpener::Store qw/retrieve_object/;
 
-my $cc;
```
stephanegigandet, line 58: I don't think this variable is used anywhere in this file, we should remove it completly.

dhiraj-143r: I am done with it

PR #13099, opened 2026-02-20
fix: skip already processed dirs in product iter in export scans

stephanegigandet:
This PR is to avoid going through each directory recursively when we want to resume exporting scans  data in scripts/export_scans_to_query.pl

### DIFF
--- a/lib/ProductOpener/Store.pm
+++ b/lib/ProductOpener/Store.pm
@@ -597,7 +597,7 @@ sub remove_object($path) {
 	return;
 }
 
-=head2 object_iter($initial_path, $name_pattern = undef, $exclude_path_pattern = undef, $skip_until_path = undef)
+=head2 object_iter($base_path, $name_pattern = undef, $exclude_path_pattern = undef, $skip_until_path = undef)
 
 Iterates over the path returning a cursor that can return object paths whose
 name matches the $name_pattern regex and whose path does not match the $exclude_path_pattern.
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Store.pm
```diff
@@ -597,14 +597,15 @@ sub remove_object($path) {
...
 
 Iterates over the path returning a cursor that can return object paths whose
-name matches the $name_pattern regex and whose path does not match the $exclude_path_pattern
+name matches the $name_pattern regex and whose path does not match the $exclude_path_pattern.
+If $skip_until_path is provided, skips all object paths that are lexicographically less than $skip_until_path.
 
 =cut
 
-sub object_iter($initial_path, $name_pattern = undef, $exclude_path_pattern = undef) {
+sub object_iter($initial_path, $name_pattern = undef, $exclude_path_pattern = undef, $skip_until_path = undef) {
```
alexgarel: Nice that you were able to implement that !

john-gom:
@stephanegigandet , @alexgarel although I approved and merged this, when I try it out on the current ingredients branch I'm working on I realised it doesn't work for directory structures where the files are at different depths. This is because the current object_iter function will start returning files as soon as it finds a directory that contains files, so if we have the following:

000/0001/0001/1.json
234/4567/2.json

Then 2.json will be returned first. If we resume from this point then the whole 000 directory will be missed. This isn't a problem as long as all the directories are the same depths, which should be OK for products but not for special folders (or in my case ingredient products).

I'm going to see if I can re-work the logic so that it returns the deepest item from the first directory first, but this is just a heads up in case you try and use the skip_until_path before then.

PR #13108, opened 2026-02-21
taxonomy(food): category updates

Freso:
- Adds (mostly Danish and Swedish) translations.
- Standardises some Danish and Swedish that was currently translated differently at different places.
- Updates/fixes some taxonomy tree structuring.

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -7570,7 +7570,7 @@ nl: Smoothie kiwi appel ananas
 < en:Non-alcoholic beverages
 < en:Plant-based beverages
 en: Fruit and vegetable shots
-da: Frugt- og grønsagsshots
+da: Frugt- og grøntsagsshots
 sv: Frukt- och grönsaksshots
 description:en: Beverages of (usually perceived-to-be healthy) juices meant to be drunk in small doses, e.g. like a shot.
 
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -81148,6 +81155,7 @@ ciqual_food_name:en: Couscous with meat or chicken, light
 ciqual_food_name:fr: Couscous à la viande ou au poulet, allégé
 
 < en:Cooked vegetables
+< en:Mixed vegetables
 < en:Prepared couscous
 en: Cooked mixed vegetables for couscous
```
moon-rabbitOFF, line 81169:
```suggestion
< en:Cooked vegetables
< en:Prepared vegetables
en: Cooked mixed vegetables for couscous
```
I think this category would go better in "prepared vegetables". The products I have seen in the database contain vinegar, oil, salt, sugar or concentrated tomato paste. It can also be removed from "prepared couscous", if it contains only prepared begetables without semolina, it is not a couscous.

moon-rabbitOFF (changes requested):

taxonomies/food/categories.txt
Freso: If you think my suggested change makes the situation worse, I’ll remove it from this PR. Thanks!

Freso:
@moon-rabbitOFF Can you re-review this please so it can move forward?

taxonomies/food/categories.txt
moon-rabbitOFF: Yes I think it would be better to put this category in "prepared vegetables"

Freso: This PR isn’t touching this category at all now. Since this change wasn’t obvious (or as obvious as the other ones), I took it out. It can be done in another PR with more explanation/justification.

Issue #3959, opened 2020-07-29
[ingredient-analysis] ingredient group in parentheses has a colon before the opening parenthesis, not parsed correctly


aleksejrs:
**Describe the bug**
`ingredient group: (element1, element2, element3)` needs to be parsed as `ingredient group (element1, element2, element3)`.

**To Reproduce**
1. Go to https://world.openfoodfacts.org/product/4604075020176/%D1%81%D0%BE%D1%83%D1%81-mr-ricco-%D1%81%D1%8B%D1%80%D0%BD%D1%8B%D0%B9#ingredients_list
2. Click "Details of the analysis of the ingredients »"
3. Look at the items under "комплексная пищевая добавка".

**Expected behavior**
"порошок сыра гауда" to "эмульгирующая соль Е 339" must be under "комплексная пищевая добавка".

svensven:
For clarity, it's currently parsing it into `ingredient group (element1), element2, element3`.

PR #13133, opened 2026-02-23
fix: treat ':' before opening bracket as nested ingredients (#3959)

pranjal2004838:
Minimal parser fix to treat a colon immediately followed by an opening bracket as the bracket separator, so constructs like 'meat: (beef, pork, lamb)' are parsed as nested ingredients.

- References: #3959

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -2092,6 +2092,16 @@ Text to analyze
 
 			if ($sep =~ /(:|\[|\{|\(|\N{U+FF08})/i) {
 
+				# If the separator is a colon but is immediately
+				# followed by an opening bracket, treat the
+				# bracket as the separator so the parenthesis
+				# matching logic below will run on the group.
+				# Example: "meat: (beef, pork)"
+				if ($sep eq ':' and $after =~ /^\s*(\(|\[|\{|\N{U+FF08})/i) {
+					$sep = $1;
+					$after =~ s/^\s*(?:\(|\[|\{|\N{U+FF08})\s*//;
+				}
+
 				# Single separators like commas and dashes
 				my $match = '.*?';    # non greedy match
 				my $ending = $last_separator;
--- a/tests/unit/ingredients_nesting.t
+++ b/tests/unit/ingredients_nesting.t
@@ -431,6 +431,35 @@ my @tests = (
 		]
 	],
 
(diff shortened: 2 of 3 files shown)


lib/ProductOpener/Ingredients.pm
```diff
@@ -2092,6 +2092,16 @@ Text to analyze
 
 			if ($sep =~ /(:|\[|\{|\(|\N{U+FF08})/i) {
 
+				# If the separator is a colon but is immediately
```
john-gom, line 2095: Should add a test for this

html/js/folksonomy.js
```diff
@@ -418,8 +418,8 @@ function delPropertyValue(_this) {
  * @param {string} _k  The property to add
  * @param {string} _v  The value to add
  * @param {string} _owner The owner or "" if it's a public declaration
- * @returns {string} Returns values from http POST
- * @todo returns {string} Returns if yes (0) or no (1) the new declaration has been created ?
+ * @returns {undefined}
```
john-gom: This looks like it should be in a different PR

lib/ProductOpener/Ingredients.pm
```diff
@@ -1,7 +1,7 @@
 # This file is part of Product Opener.
 #
 # Product Opener
-# Copyright (C) 2011-2026 Association Open Food Facts
+# Copyright (C) 2011-2023 Association Open Food Facts
```
john-gom: Why this change?

teolemon: I think it's because you used a LLM @pranjal2004838 ?

pranjal2004838:
It was accidental and Ill revert it.

lib/ProductOpener/Ingredients.pm
pranjal2004838: Hi — sorry about the noise in the previous push. While preparing the parser fix I accidentally included two unrelated, tiny edits (a copyright year line and a brief JSDoc change).
I’ve reverted those unrelated edits in a single cleanup commit (4932f106f8) and pushed it to this branch, so the PR now contains only the intended parser fix.
I’m sorry for the extra review work this caused. Please re-check when convenient and let me know if you’d like me to rebase/split anything further — I’ll follow whatever workflow you prefer.

html/js/folksonomy.js
pranjal2004838: It was accidental and I have reverted it to a clean single commit. Kindly review it again and let me know if anything else is required

PR #13145, opened 2026-02-25
fix: docker add log rotation to prevent disk saturation

suvammpatra:
## Summary
This PR adds Docker json file log rotation to all services in docker-compose.yml.

## Changes
Configured max-size: 20m
Configured max-file: 5
Applied to backend, frontend, postgres, memcached and related services

## Impact
Prevents uncontrolled log growth and disk saturation in production Docker deployments.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,9 +87,21 @@
     - ./conf/apache2-docker/apache-mpm_prefork.conf:/etc/apache2/mods-enabled/mpm_prefork.conf
     - ./conf/apache-2.4/modperl.conf:/etc/apache2/conf-enabled/modperl.conf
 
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
teolemon, line 1: we try to avoid dual purposes PRs

teolemon:
@suvammpatra was there an issue related for that ? did anyone ask for it in the Slack ? was it bothering you ?

taxonomies/food/categories.txt
suvammpatra: I will separate the taxonomy changes from the Docker fix and submit a clean PR focused only on log rotation.

PR #13150, opened 2026-02-25
fix: Don't bother fixing the product_type for deleted products

john-gom:

### What
Script was previously correcting the product type for deleted products. This creates unnecessary updates

### DIFF
--- a/scripts/fix_mongodb_collections.pl
+++ b/scripts/fix_mongodb_collections.pl
@@ -99,7 +99,8 @@ while (my $path = $next->()) {
 		}
 	}
 	my $product_type = $product_ref->{product_type};
-	if (not defined $product_type) {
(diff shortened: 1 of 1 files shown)


teolemon:
I'm not sure if this is a good idea since @CharlesNepote has put a script to delete and complete products and those products get recreated after a while by third-party apps.

john-gom:
> I'm not sure if this is a good idea since @CharlesNepote has put a script to delete and complete products and those products get recreated after a while by third-party apps.

So you think we should add the product_type even for deleted products?

john-gom:
> > I'm not sure if this is a good idea since @CharlesNepote has put a script to delete and complete products and those products get recreated after a while by third-party apps.
> 
> So you think we should add the product_type even for deleted products?

I think if the product was re-created the code would add the product_type in anyway

PR #13155, opened 2026-02-25
fix: correct nutrient filtering for API search (numeric parsing and Mongo operators)

### LINKED ISSUE #12175: Wrong search results when filtering for nutriments
### What
- When filtering for nutriments, wrong results appear not following the defined filters.

1) Go to the wiki: https://wiki.openfoodfacts.org/Open_Food_Facts_Search_API_Version_2
2) On the wiki, there is an example to get sugars_serving > 10
3) Click on the given link: https://world.openfoodfacts.org/api/v2/search?categories_tags=en:sodas&sugars_serving%3E10g&fields=code,product_name,sugars_serving
4) The search results show products with less than 10g of sugar per serving

### Expected behavior
- The result does not show product with more less than or equal to 10g of sugar per serving.

### Why
- As this is a very crucial functionality in searching for products

```{
  "count": 554,
  "page": 1,
  "page_count": 20,
  "page_size": 20,
  "products": [
    {
      "code": "54491472",
      "product_name": "coca cola 500ml",
      "sugars_serving": 53
    },
    {
      "code": "5449000131836",
      "product_name": "Zero Sugar",
      "sugars_serving": 0
    }]}```

### Part of
- #525 <!-- Add the issue number preceded by # (that also allows searching by name for similar issues) --> 

### Type of device
- Browser: Chrome
- Postman
- Python requests

ALL

### Time per product
A lot

Mounika-max-348:
### What

Fix incorrect results when filtering by nutriments in the API search.

The API previously returned products that did not respect numeric filters
(e.g. `sugars_serving>10` returning products with lower values).

### Root cause

The nutrient filter:

- did not properly parse numeric values when units were present (e.g. `10g`)
- did not consistently build MongoDB numeric comparison operators
- could overwrite existing conditions instead of merging them

### Fix

- Strip non-numeric characters from nutrient values before comparison
- Correctly map `<, <=, >, >=` to MongoDB `$lt, $lte, $gt, $gte`
- Ensure multiple conditions on the same nutrient are merged safely
- Always treat values as numeric (`+0.0`)

### Impact

API nutrient filters now return correct results and respect the defined thresholds.

### Related issue

Fixes #12175

### DIFF
--- a/gulpfile.mjs
+++ b/gulpfile.mjs
@@ -24,6 +24,7 @@ const jsSrc = [
   "./html/js/barcode-scanner*.js",
   "./html/js/rewind-browser.js",
   "./html/js/external-knowledge-panels.js",
+  "./html/js/matomo-events.js",
 ];
 
 const sassSrc = "./scss/**/*.scss";
--- a/html/js/matomo-events.js
+++ b/html/js/matomo-events.js
@@ -1,0 +1,20 @@
+/*exported trackMatomoEvent*/
+/*global _paq*/
+
+// Safely track event to Matomo queue, ignoring if blocked.
+function trackMatomoEvent(category, action, name, value) {
+    if (typeof _paq === 'undefined') {
+        return;
+    }
+
+    const eventData = ['trackEvent', category, action];
+
+    if (name) {
+        eventData.push(name);
+        if (typeof value === 'number') {
+            eventData.push(value);
+        }
+    }
+
+    _paq.push(eventData);
+}
--- a/html/js/product-multilingual.js
+++ b/html/js/product-multilingual.js
@@ -20,7 +20,7 @@
 
 /*eslint dot-location: "off"*/
 /*eslint no-console: "off"*/
-/*global lang admin initializeTagifyInput other_nutrients:writable*/ // we change other_nutrients to remove nutrients when they are added
+/*global lang admin initializeTagifyInput other_nutrients:writable trackMatomoEvent*/ // we change other_nutrients to remove nutrients when they are added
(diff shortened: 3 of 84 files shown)


Mounika-max-348:
Hi  , I have opened a PR that fixes the incorrect nutrient filtering in the API:

https://github.com/openfoodfacts/openfoodfacts-server/pull/13155

The fix ensures numeric parsing and proper MongoDB comparison operators.
Would appreciate feedback.

john-gom:
Please verify that any changes to the MongoDB filter expression style work with off-query

Mounika-max-348:
Hi @john-gom,

Updated the nutrient filter logic to be off-query compatible and to correctly merge multiple conditions without overwriting existing filters.  
PR now only contains the intended changes.

Please review.

stephanegigandet:
Hi @Mounika-max-348 ,
Thanks for the PR.
Can you add unit tests for this function that show what didn't work before and what works now?
In practice I believe we were already generating queries with $lt etc.
What's new would be the merging of conditions on the same field I think.
Also I didn't see where was the stripping of the unit which was causing the problem in the first place.

Have you run product opener locally to verify it solves the bug you reference?

Thank you!

Mounika-max-348:
Hi @stephanegigandet ! I added unit tests for `add_params_to_query` that cover:
• merging of multiple conditions on the same nutrient
• handling of the `_100g_serving` suffix
• stripping of units (e.g. "10g" → 10)

I tried running the test locally, but it requires the full Product Opener Perl environment, which is only available in the Docker dev setup.
The tests should run correctly in CI where all dependencies are installed.

PR #13167, opened 2026-02-26
fix: fallback nutrition_data_per from aggregated nutrition set

### LINKED ISSUE #13126: Nutrition edit: "for 100 g" checkbox can be unchecked and saved but state is not persisted; no feedback to user
### What
On the product edit page (Nutrition tab), under "Nutrition facts are specified for the product as sold," the "for 100 g" checkbox can be unchecked and the form saved successfully. When the product is opened for edit again, "for 100 g" is ticked again — the unchecked state is not persisted. The frontend does not indicate that this change was not saved (no validation, no error, no reflection of actual persisted state after save).

### Steps to reproduce the behavior:
1. Go to a product edit page (e.g. world.openfoodfacts.org, product with nutrition data).
2. Open the **Nutrition** tab.
3. Under "Nutrition facts are specified for the product as sold," uncheck **"for 100 g"** so only "for 100 ml" is checked.
4. Save the product.
5. Re-open the same product for edit and go to the Nutrition tab.
6. See that "for 100 g" is ticked again (state was not persisted).

### Expected behavior
Either the unchecked state is persisted and remains unchecked when re-opening, or the UI prevents saving that state and clearly informs the user (e.g. validation error or disabled option with explanation), so the user is not led to believe the change was saved.

### Screenshots

Image

### Platform (Desktop, Mobile, Hunger Games)
 - OS: Ubuntu
 - Platform: Desktop (world.openfoodfacts.org in browser)

nithyaV-dev:
### What
Fix missing nutrition_data_per in API response when top-level field is absent but available in normalized nutrition schema.

### Why
During schema conversion, nutrition_data_per may not exist at top-level but is present in nutrition.aggregated_set.per.  
API previously skipped this field, causing incomplete response.

### Fix
Add safe fallback in customize_response_for_product to populate nutrition_data_per from normalized nutrition.aggregated_set.per when preparation = "as_sold".

### Impact
- No schema change
- No DB change
- Backward compatible
- Only fills missing field when data exists

- Fixes #13126


stephanegigandet:
Hi @nithyaV-dev , have you reproduced the bug https://github.com/openfoodfacts/openfoodfacts-server/issues/13126 locally?
It's not related to the API.

nithyaV-dev:
Hi @stephanegigandet, thanks for checking!

I haven’t fully reproduced it locally yet — I investigated starting from the API response side because I noticed nutrition_data_per could be missing depending on schema conversion.

If the issue is not API-related, I’ll try to reproduce it locally and trace where the checkbox state is actually being saved (or lost) during product edit.

Thanks for the guidance — I’ll dig deeper and update here.

Issue #8191, opened 2023-03-13
Log key events in Matomo


teolemon:
- [ ] Ensure that Matomo event names for Mobile and the Web are the same, whenever possible

```[tasklist]
### Tasks
- [ ] Instrument the "Signup successful page" in Matomo
- [ ] Add Matomo event for User clicking on account deletion - we can try to see if we can guess what caused it (eg: missing Scores, attempt to do something that failed)
- [ ] Add Matomo event for Completed sign-up - ensure the signup form is working well
- [ ] Add Matomo event for "Share product"
- [ ] Actual interaction with food preferences (we can anonymize the values, or even the attributes, but my question is: do people tweak the scoring, if so for which reasons)
- [ ] Add events for each section of the Edit page in Matomo, Clicking save as well, to see how many people give up
- [ ] Add events for knowledge panel interaction (unfolding…)
- [ ] Add events to visits to key pages (Nutri-Score, NOVA, Eco-Score, Contribute, Discover)
- [ ] Explore by button
- [ ] Sorting by button
- [ ] Clicking on PlayStore/AppStore button
- [ ] Hovering on black bar menus (user menu, tools menu, language/country menu)
- [ ] Enabling the personal ranking button
- [ ] Producer platform events (Import images, Import files…)
```
### Part of
- Better understanding users

teolemon:
done

LakshuCodes:
Hi ,
I’d like to work on a small, focused part of this issue.

I propose to implement the Matomo event for the “Signup successful page” (P0).

Does this sound like a good first step and is it still a priority?
Thanks.

devsk18:
Hi @teolemon , I'd like to work on the **Personal Search** section,

 specifically: 
- adding a Matomo event for the personal ranking toggle button.

I've added the tracking call - `trackMatomoEvent()` inside `activate_preferences_switch_buttons()` in `product-preferences.js`,
- category: `personal_search`
- action: `enabled` / `disabled`

Tested locally and everything is working as expected. Can I raise a PR for this?

PR #13174, opened 2026-02-27
fix: track Nutri-Score, Eco-Score, and NOVA Matomo events on product

dhiraj-143r:
This PR tracks Matomo events for Nutri-Score, Eco-Score, and NOVA when a user views a product page. 

Since these scores are generated dynamically by Knowledge Panels, I added a small DOM script that waits for the page to load, checks which score images are being displayed, and pushes the exact grades to Matomo.

Fixes part of #8191.

### Events Tracked:
* `has_nutriscore` (e.g,  a, b, c, d, e)
* `has_ecoscore` (e.g,  a, b, c, d, e)
* `has_nova` (e.g,  1, 2, 3, 4)

I tested this locally in the browser by mocking the DOM, and all three events are successfully firing and pushing the correct values to the `_paq` array.

https://github.com/user-attachments/assets/27f04c1f-1ff8-4b6a-82e6-d317c7034e66

### DIFF
--- a/html/js/matomo-events.js
+++ b/html/js/matomo-events.js
@@ -18,3 +18,35 @@ function trackMatomoEvent(category, action, name, value) {
 
     _paq.push(eventData);
 }
+
+// Track product page scores (Nutri-Score, Eco-Score, NOVA) dynamically
+document.addEventListener('DOMContentLoaded', function () {
+    if (document.body.classList.contains('product_page')) {
+        // Nutri-Score
+        var nsImg = document.querySelector('img[src*="nutriscore-"]');
+        if (nsImg) {
+            var nsMatch = nsImg.src.match(/nutriscore-([a-e])/);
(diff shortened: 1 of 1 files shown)


templates/web/pages/product/product_page.tt.html
```diff
@@ -449,4 +449,39 @@ <h2 id="data_sources" class="text-medium">[% lang('data_sources_p') FILTER ucfir
         </div>
 </section>
 
+
+<script>
```
alexgarel: Can we move all matomo tracking stuff to a matomo-events.js ?

dhiraj-143r: Sure, I will move it  into matomo-events.js

dhiraj-143r:
@alexgarel ,
I have removed the tracking script from the HTML template and moved it to matomo-events.js

PR #13182, opened 2026-02-28
taxonomy: new categories for verrines

moon-rabbitOFF:
I created two new categories, one for savory and the other one for sweet verrines.


Freso (approved):
Seems good! Thanks!

PR #13184, opened 2026-02-28
taxonomy: modification savory cakes 

moon-rabbitOFF:
I renamed "salted savory cake" in "savory cakes". I moved it under appetizers. I changed the sub-categories, which didn't make so much sense. E.g. I deleted "Savoury cake with cheese" because savory cakes rarely contain only cheese and a distinction between the one that contain cheese and the one that don't doesn't make sense. I made two separated categories "cakes with meat" and "vegetarian cakes", which is a more useful distinction.

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -118387,7 +118387,7 @@ en: Vegetarian savory cakes, Savoury cake with vegetables
 fr: Cakes salés végétariens, Cake salé aux légumes
 
 < en:Savory cakes
-en: Savory cakes, Savoury cake with meat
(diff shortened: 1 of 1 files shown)


Freso:
My understanding is that categories follow to some extent how products are marketed, and it seems like there are indeed [cakes sold as “with cheese”](https://world.openfoodfacts.org/facets/categories/savoury-cake-with-cheese).

moon-rabbitOFF:
Only two of them have only cheese, the other are have a vegetable or a fruit plus cheese. Labelling this category as "with cheese" makes it overlap with the former "savoury cakes with vegetables"; while the new category "vegetarian savory cakes" encompass them all.

PR #13198, opened 2026-03-01
taxonomy: basque-style chicken

moon-rabbitOFF:
Created two sub-categories for basque-style chicken: with and without side dishes


stephanegigandet (approved):
Thanks

PR #13205, opened 2026-03-02
feat: Add GET /api/v3/current-user/permissions endpoint for user role

swastikCommits:
### What

Added a new API endpoint `GET /api/v3/users/me` that returns information about the currently authenticated user, whether they are `moderator` or `admin`

Clients like the OFF-Explorer use Keycloak for authentication via PKCE. While Keycloak provides basic user info, the authoritative source for user roles (moderator, admin) remains in the (`.sto` files). 
Currently, there is no OIDC-compatible API endpoint to retrieve these flags. This new endpoint allows the Explorer (and other clients) to fetch authenticated user data using a Bearer token, enabling the UI to correctly display moderator/admin badges and gate features accordingly.

Response: 
```json
{
  "status": "success",
  "result": { "id": "user_found" },
  "user": {
    "userid": "username",
    "name": "Full Name",
    "moderator": 1,
    "admin": 0
  }
}
```

- Part of https://github.com/openfoodfacts/openfoodfacts-js/issues/615 & https://github.com/openfoodfacts/openfoodfacts-explorer/issues/65


lib/ProductOpener/Routing.pm
```diff
@@ -369,6 +369,13 @@ sub api_route($request_ref) {
 		param("tagid", $components[4]);
 		$request_ref->{tagid} = $components[4];
 	}
+	elsif ($api_action eq "users") {    # api/v3/users/[sub_action]
+		# Currently only 'me' is supported: GET /api/v3/users/me
```
stephanegigandet: there is a user with the userid "me", so maybe a different endpoint would be better/api/v3/users/current-user maybe? I'm guessing we may want to use /api/v3/users/[userid] in the future.

lib/ProductOpener/APIUserRead.pm
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+
+On error (not authenticated): HTTP 401 with error details
+
+=cut
+
+sub read_user_api ($request_ref) {
+
+	$log->debug("read_user_api - start", {request => $request_ref}) if $log->is_debug();
```
stephanegigandet: We should return an error if we don't have the expected userid (/me or /current-user )

swastikCommits:
@alexgarel @stephanegigandet can anyone of your review this PR?

stephanegigandet:
@swastikCommits I made some comments yesterday. Also could you add an integration test? There are examples in tests/integration/api_*.t

lib/ProductOpener/API.pm
```diff
@@ -507,6 +510,11 @@ my $dispatch_table = {
 		HEAD => \&external_sources_api,
 		OPTIONS => sub {return;},    # Just return CORS headers
 	},
+	# User info: GET /api/v3/users/me
```
john-gom: I don't think it should be a path under users. We probably won't ever have an API to get users in general but if we did this might clash with a real user id

swastikCommits: If that's the case then, we can change the route to `/api/v3/auth/me` ?

stephanegigandet: @john-gom well wouldn't product opener user data be available someday from /users/[userid] ? e.g. product opener specific badges or leaderboard for instance

@swastikCommits Or for admin / moderator status, maybe /api/v3/permissions would be better, we could add extra permissions later, and that endpoint would be specific to the currently logged in user

stephanegigandet:
@swastikCommits sorry my comments were in "pending" for the last few days and I didn't notice they were not sent...

lib/ProductOpener/API.pm
john-gom: I like the idea of putting it under an Auth endpoint as that would give a good clue not to cache. I personally prefer "self" to "me" although I can't say why so either is fine 😁

john-gom: Hi @stephanegigandet I wasn't ruling out a users endpoint in the future, but a "me" endpoint would potentially return more personal data than a general users API

swastikCommits: I had it updated and working locally as `/api/v3/permissions` based on Stephane's suggestion, but `api/v3/auth/self` also makes sense for caching reasons. 

Can both of you (@stephanegigandet and @john-gom) give a final call which one shall we go with so that I can push the commit along with updated test integrations?

lib/ProductOpener/APICurrentUser.pm
```diff
@@ -0,0 +1,148 @@
...
+	$log->debug("read_current_user_permissions_api - start", {request => $request_ref}) if $log->is_debug();
+
+	my $response_ref = $request_ref->{api_response};
+
+	# Validate the sub-action: only "permissions" is currently supported.
+	my $sub_action = $request_ref->{current_user_sub_action} // '';
+	if ($sub_action ne 'permissions') {
+		$log->info("read_current_user_permissions_api - invalid sub-action", {sub_action => $sub_action})
+			if $log->is_info();
+		add_error(
```
hangy, line 101: Might want to make this a 404 error until other sub actions get implemented

john-gom: I made this consistent with API.pm. I guess the reason it isn't a 404 is that its not a variable URL, like product/xxx

PR #13207, opened 2026-03-02
fix: add docker log rotation to prevent disk saturation

### LINKED ISSUE #13081: Handle logs size of docker deployments
@john-gom saw the disk of scaleway-docker-prod saturated by docker logs.

We should limit system wide the logs, augment a bit the disk size also (to support docker images and logs)

Modify role to:
- add a global log rotation / deletion for docker daemon (see https://tecadmin.net/rotate-docker-logs/)
- issue a warning when VM disk size is too small (maybe at least 100G) considering we have to store images and logs

suvammpatra:
Summary
This PR adds Docker json file log rotation to all services in docker-compose.yml.

Changes
Configured max-size: 20m
Configured max-file: 5
Applied to backend, frontend, postgres, memcached and related services

Impact
Prevents uncontrolled log growth and disk saturation in production Docker deployments.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,9 +87,21 @@
     - ./conf/apache2-docker/apache-mpm_prefork.conf:/etc/apache2/mods-enabled/mpm_prefork.conf
     - ./conf/apache-2.4/modperl.conf:/etc/apache2/conf-enabled/modperl.conf
 
+  logging:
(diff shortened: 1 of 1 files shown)


teolemon:
@alexgarel is it interesting ?
@suvammpatra there's a merge conflict.

If not relevant, please close.

PR #13208, opened 2026-03-02
fix: option to export nutrition aggregated set in CSV

stephanegigandet:
This PR:
- removes the "estimate" input nutrition set from CSV exports, except if computed fields are requested
- adds a new option to also export the aggregated set


Vic142 (approved):
There are failing integration tests

PR #13214, opened 2026-03-03
fix: Safari nutrition image layout rounding issue

### LINKED ISSUE #13202: Nutrition image layout issue
Image

### What
When adding new nutrients the nutrition image covers the units of the newly entered nutritients,

1. while editing https://nl.openfoodfacts.org/product/8711521319800/margarine-met-olijfolie-your-organic-nature

### Expected behavior
Image next to the table

### Why
Hides units

### Type of device
- Browser: Safari

nithyaV-dev:
### Description

Fixes nutrition image copy misalignment in Safari.

Safari was calculating element widths with fractional pixel values,
which caused incorrect positioning. This change uses
getBoundingClientRect() along with Math.ceil / Math.floor
to ensure consistent width calculations.

Fixes #13202

### Type of change
- Bug fix

### Tested on
- Safari
- Chrome

### DIFF
--- a/html/js/product-multilingual.js
+++ b/html/js/product-multilingual.js
@@ -1093,6 +1093,12 @@ $(function () {
             // clear the values: inputs with class nutrient_value that are inside a cell with the input_set_class
             $('.' + input_set_class + ' input.nutrient_value').val('');
         }
+        
+        
+        // Recalculate nutrition image position after table resize
+        setTimeout(update_nutrition_image_copy, 50);
+        
+        
     });
 
     $('#no_nutrition_data').on('change', function() {


nithyaV-dev:
It looks like the "Generate Docker SBOM" check failed due to
`backend-trivy.sarif` not being generated.

All other checks are passing, and this change only affects a frontend JS file.

Could this be a CI infrastructure issue? I’d be happy to rebase or rerun if needed.

stephanegigandet:
@nithyaV-dev Thanks for the PR, but that's not the issue from the bug. It's not restricted to Safari, it happens in all browsers.

To replicate the issue:

Log in and edit one product:

image

Add a nutrition column by adding a checkbox:

image

stephanegigandet:
About the rounding issue: can you show screenshots without your change and after your change? I'm not sure why Safari would have a problem with fractional pixels?

stephanegigandet:
To fix the issue with the overlapping image when adding column, you can just call update_nutrition_image_copy(); whenever a checkbox for the nutrition input set is changed, near this code:

```
    // Nutrition input set checkboxes
    // For each element with the class nutrition_input_set,
    // we use the id of the checkbox nutrition_input_sets_${preparation}_${per}_shown
    // to show or hide the input set column cells with the class nutrition_input_sets_${preparation}_${per}
```

nithyaV-dev:
### Fix nutrition image layout overlap when nutrition columns change

This patch fixes an issue where the nutrition image could overlap the nutrition table (particularly the units column) when the table width changed after toggling nutrition input set checkboxes.

The issue was caused by inconsistent width calculations when determining whether there is enough space to display the nutrition image next to the table.

### Changes

* Replaced jQuery `.width()` calculations with `getBoundingClientRect().width` combined with `Math.ceil` / `Math.floor` to avoid rounding inconsistencies that could cause layout overlap.
* Triggered `update_nutrition_image_copy()` when nutrition input set checkboxes change so the layout recalculates after the table width changes.

### Result

* The nutrition image is positioned correctly next to the nutrition table.
* The layout updates correctly when nutrition columns are toggled.
* The image hides automatically when the table becomes too wide for the available space.

### Browser testing

Tested in:

* Chrome

The layout behaves correctly when nutrition columns are toggled.

### Screenshots

**Before fix – image overlapping the units column**

Screenshot (1088)

**After fix – image positioned correctly**

Screenshot 2026-03-05 001049


**Table expanded – image hides when there is not enough space**

Screenshot 2026-03-04 223424

PR #13224, opened 2026-03-04
feat: Add --since param

CharlesNepote:


CharlesNepote:
It's the script I have used to extract emails. Works well.

PR #13239, opened 2026-03-05
feat: restructure tags schema and API to separate sources and parents

stephanegigandet:
Tags refactor: https://docs.google.com/document/d/151UlwHUjxiE3UXde4BB2aXWZdSx6HSE3ot0iqhdGDow/edit?tab=t.0#heading=h.ffj0m6gyxp9

Note: there are many changes in test results, including in some results of the API, but only in fields that were not documented/should not be used.

There are some visible changes for fields that are not taxonomized, and for unrecognized entries of taxonomized fields:
- we currently save the original value (capitalized, accented etc.) in tags_sources, but not in fields like brands_tags, store_tags, categories_tags etc. which contain a normalized version.
- when we display those fields (e.g. in product page), we use brands_tags to display the fields, so we display unaccented / lowercased values (except for taxonomy entries that exist in the taxonomy).
- an alternative is to store values as-is in categories_tags etc. but it has other issues, especially for the associated URLs that can then contain any character, spaces etc.

Changes are documented here: https://docs.google.com/document/d/151UlwHUjxiE3UXde4BB2aXWZdSx6HSE3ot0iqhdGDow/edit?tab=t.0#heading=h.memy8vfc3a4s


lib/ProductOpener/Tags.pm
```diff
@@ -5199,6 +5201,48 @@ sub get_taxonomy_tag_path ($tagtype, $tagid) {
...
+
+A reference to a list of tagids.
+
+=head3 Return value
+
+A list of tagids that are not parents of any other tagid in the input list.
+
+=cut
+
+sub get_minimal_tags_subset ($tagtype, $tags_ref) {
```
Freso, line 5063: Would this be the function to drop tags in too, or should there be a separate function for this? I reckon the most simple would be to have a list of undesireables (per `$tagtype`) and check if the current `$tagid` is in there, and if it is, just drop it and move to the next `$tagid`?

I think it’d probably be good to do this in a separate function, since `get_minimal_tags_subset()` seems generally useful, where `drop_undesired_tags()` can probably get dropped after the transition/conversion.

stephanegigandet: Right, we would remove unwanted tags separately.

lib/ProductOpener/Display.pm
```diff
@@ -3156,7 +3156,9 @@ sub canonicalize_request_tags_and_redirect_to_canonical_url ($request_ref) {
 		else {
 			$display_tag = canonicalize_tag2($tagtype, $tagid);
 			# Use "no_language" normalization for tags types without a taxonomy
-			$new_tagid = get_string_id_for_lang("no_language", $display_tag);
+			# Tag refactor: we don't normalize tags
+			#$new_tagid = get_string_id_for_lang("no_language", $display_tag);
+			$new_tagid = $display_tag;
```
alexgarel: Shan't we verify we have no "," inside ?

lib/ProductOpener/Import.pm
```diff
@@ -1099,6 +1024,120 @@ sub compare_old_and_new_objects (
...
+
+	# Make a deep copy
+	my $old_tags_sources = dclone($product_ref->{tags_sources} || {});
+
+	# If an import_lc was passed as a parameter, assume the imported values are in the import_lc language
+	my $tag_lc = $imported_product_ref->{lc} // $args_ref->{import_lc} // "en";
+
+	# Go through each field to find fields of the form tags_sources.[tag field].[source].tags|last_updated_t
+	foreach my $field (sort keys %$imported_product_ref) {
+		if ($field =~ /^tags_sources\.([^\.]+)\.([^\.]+).tags$/) {
```
alexgarel, line 1049: Note: Those kind of "convention" based argument passing is a bit brittle (quiet impossible to find where it is used in the codebase for sure).

Not to change now, but this kind of things is better if isolated in a piece of functions written nearby that handle getting the field name and parsing it. (get_tag_source_csv_name, parse_tag_source_csv_name, something along this line).

lib/ProductOpener/Import.pm
```diff
@@ -1099,6 +1024,120 @@ sub compare_old_and_new_objects (
...
+						source => $source,
+						tags_field => $tags_field,
+						value => $imported_product_ref->{$field}
+					}
+				) if $log->is_error();
+				next;
+			}
+
+			# Check the source is "packaging", "manufacturer" or starts with "label" or "database"
+			if ($source !~ /^packaging|manufacturer|label.*|database.*$/) {
```
alexgarel, line 1067: Do we have this constraint written in the OpenAPI ?

In `docs/api/ref/schemas/product_tags_v3.yaml` I just see:
```yaml
      - type: object
        properties:
          tags_sources:
            type: object
            description: |
              All the relevant data on the sources of the different tags of the product.
              The data is sorted by the sources from which the tags are obtained, e.g. packaging, producer, predicted.
            patternProperties:
              "^[a-z]+(_[a-z]+)*$":
                type: object
```
So maybe the `patternProperties` should be changed to something more strict ?

alexgarel: Also why is this constraint not internalized in `set_field_input_tags_for_source` ?

lib/ProductOpener/IngredientsStrings.pm
alexgarel, line 1: Really cool that you took the time to refactor in a separate module !

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -444,10 +454,11 @@ sub convert_schema_1002_to_1003_refactor_product_nutrition_schema ($product_ref)
...
 
 		# Generates the nutrition sets,
-		# which, for old data, are all from source "packaging" if we are on the public platform,
-		# and "manufacturer" if we are on the pro platform and the product has an 'owner' field
-		my $source
-			= ($server_options{private_products} && defined $product_ref->{owner}) ? "manufacturer" : "packaging";
+		# If we are on the producers platform, we set the data source to "manufacturer" for all organizations with ids that start with org-
+		# except organizations with ids that start with org-database- or org-label- (e.g. "org-database-usda")
+		# otherwise, we set the source to "packaging" (for organizations that start with user-
+		# when the pro platform is used by individual users to load data in bulk, e.g. from scan parties)
```
alexgarel: It's a bit strange to have this comment here, it belongs to get_source_for_site_and_org (here it might stay the say while implementation may change and so it will be inacurate)
Here I would just say that there is a whole logic to set it.

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -902,3 +913,134 @@ sub _compute_nutrition_data_per_100g_and_per_serving_for_old_nutrition_schema ($
...
+=cut
+
+sub convert_schema_1003_to_1004_refactor_tags ($product_ref) {
+
+	$product_ref->{tags_sources} = {};
+
+	# If we are on the producers platform, we set the data source to "manufacturer" for all organizations with ids that start with org-
+	# except organizations with ids that start with org-database- or org-label- (e.g. "org-database-usda")
+	# otherwise, we set the source to "packaging" (for organizations that start with user-
+	# when the pro platform is used by individual users to load data in bulk, e.g. from scan parties)
```
alexgarel: there again don't repeat an explanation that really belongs to get_source_for_site_and_org.

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -902,3 +913,134 @@ sub _compute_nutrition_data_per_100g_and_per_serving_for_old_nutrition_schema ($
...
+	my $source = get_source_for_site_and_org();
+
+	# TODO: special case for ingredients_tags
```
alexgarel: Why special ? Is it done ?

lib/ProductOpener/ProductSchemaChanges.pm
```diff
@@ -902,3 +913,134 @@ sub _compute_nutrition_data_per_100g_and_per_serving_for_old_nutrition_schema ($
...
+		if (defined $tags_hierarchy_or_tags_field) {
+
+			if (scalar @{$tags_hierarchy_or_tags_field} > 0) {
+
+				$product_ref->{$tagtype . "_tags"} = $tags_hierarchy_or_tags_field;
+
+				# we create a tags_source.categories with the minimal tags subset to generate categories_tags
+
+				my $source_ref = {
+					last_updated_t => time() + 0,
```
alexgarel: why don't you use the last_updated_t of the product, that would seems more logical to me.

lib/ProductOpener/Tags.pm
```diff
@@ -2950,17 +2959,38 @@ sub gen_tags_hierarchy_taxonomy ($tag_lc, $tagtype, $tags_list) {
...
+=head4 reference to a list of tags $tags_ref
+
+=cut
+
+sub gen_tags_list_with_parents($tag_lc, $tagtype, $tags_ref) {
+
+	# If the tagtype is not a taxonomy, we just unique the tags and return them in the same order
+	if (not defined $all_parents{$tagtype}) {
+		my %seen = ();
+		return grep {$_ ne ''} grep {!$seen{$_}++} @$tags_ref;
```
alexgarel: Waow this is a bit a tricky one-liner. But it's ok.

lib/ProductOpener/Tags.pm
```diff
@@ -3586,6 +3632,9 @@ sub canonicalize_tag2 ($tagtype, $tag) {
 
 sub get_taxonomyid ($tag_lc, $tagid) {
 
+	# 20260413 tag refactor: do not normalize tags in URLs (but taxonomized entries are already canonicalized)
+	return $tagid;
```
alexgarel: The code below is unreachable ?

lib/ProductOpener/Tags.pm
```diff
@@ -3597,6 +3646,9 @@ sub get_taxonomyid ($tag_lc, $tagid) {
 
 sub get_taxonomyurl ($tag_lc, $tagid) {
 
+	# 20260413 tag refactor: do not normalize tags in URLs (but taxonomized entries are already canonicalized)
+	return $tagid;
```
alexgarel: There again, code below is unreachable ?

lib/ProductOpener/Tags.pm
```diff
@@ -4663,6 +4715,304 @@ sub add_tags_to_field ($product_ref, $tag_lc, $field, $additional_fields) {
...
+		}
+	}
+
+	# If we did not recognize the allergen, return the allergen id
+	return $allergenid;
+}
+
+=head2 set_field_input_tags_for_source ($product_ref, $tag_lc, $field, $source, $input_tags)
+
+New function to set the input tags for a field and a source. (e.g. categories for the manufacturer source)
```
alexgarel: New functions are not new forever ;-)

lib/ProductOpener/Tags.pm
```diff
@@ -4663,6 +4715,304 @@ sub add_tags_to_field ($product_ref, $tag_lc, $field, $additional_fields) {
...
+
+The input tags to set for the field and source
+
+=head4 $add
+
+Optional flag to indicate whether to add the input tags to existing tags (default is 0, which replaces existing tags)
+
+=cut
+
+sub set_field_input_tags_for_source ($product_ref, $tag_lc, $field, $source, $input_tags, $add = 0,
```
alexgarel: I find it strange to make this function belongs to Tags.pm, it's more a Products.pm sub IMHO.
(this is the only product_ref instance in this module, it makes the dependency with product circular)

If we need to use the Tags.pm internal structure, it means there is a lack of separation of concern.

lib/ProductOpener/Tags.pm
```diff
@@ -4663,6 +4715,304 @@ sub add_tags_to_field ($product_ref, $tag_lc, $field, $additional_fields) {
...
+}
+
+=head2 generate_field_tags_from_all_sources ($product_ref, $tagtype)
+
+This function gathers all the input tags for a field from all sources, and generates the final tags for the field,
+including parents for taxonomy fields.
+
+=cut
+
+sub generate_field_tags_from_all_sources ($product_ref, $tagtype, $normalize = 0) {
```
alexgarel: Same here, this is more a Products.pm sub IMO

Maybe we need a ProductsTags.pm module ?

lib/ProductOpener/Test.pm
```diff
@@ -774,6 +773,24 @@ sub normalize_object_for_test_comparison ($object_ref, $specification_ref) {
 					}
 					# Note: * on a scalar means nothing
 				}
+				# If the final key contains a * (with something else), treat the * as a wildcard for any character
+				# and apply the transformation to all fields that match the pattern
+				elsif ($final_key =~ /\*/) {
```
alexgarel, line 786: 💪

tests/integration/api_v2_product_write.t
```diff
@@ -518,6 +519,82 @@ my $tests_ref = [
 		path => '/api/v3.5/product/2234567890002?fields=raw'
 		,    # we need to specify fields=raw so that we are not redirected to the petfood site
 	},
+	# Facets
```
alexgarel: Why do you use the term "facets" instead of tags ?

```suggestion
	# Facets (aka tags)
```

tests/integration/api_v3_product_read.t
```diff
@@ -338,6 +338,13 @@ my $tests_ref = [
 		path => '/api/v3.5/product/200000000035',
 		expected_status_code => 200,
 	},
+	# v3.6 new tags schema
+	{
+		test_case => 'get-existing-product-new-tags-schema-api-v3-6',
+		method => 'GET',
+		path => '/api/v3.6/product/4260392550101',
+		expected_status_code => 200,
+	},
```
alexgarel, line 347: In the result we have:

```json
     "tags_sources" : {
         "allergens" : {
            "ingredients" : {
               "last_updated_t" : "--ignore--",
               "tags" : []
            }
         },
         "categories" : {
            "packaging" : {
               "last_updated_t" : "--ignore--",
               "tags" : [
                  "en:snacks"
               ]
            }
         },
         "traces" : {
            "ingredients" : {
               "last_updated_t" : "--ignore--",
               "tags" : []
            }
         }
      },

```
Is it worth it having the `allergens` and `traces` sources if they are empty ?

tests/integration/api_v3_product_write.t
```diff
@@ -707,24 +710,63 @@ my $tests_ref = [
...
+		path => '/api/v3.6/product/3234567890100',
 		body => '{
 			"fields" : "updated",
 			"product": { 
+				"brands_tags_add": ["Another brand"],
 				"categories_tags_add": ["en:tea"],
 				"stores_tags_add": ["Carrefour", "Mon Ptit magasin"],
 				"countries_tags_fr_add": ["Italie", "en:spain"],
-				"labels_tags_fr": ["végétarien", "Something unrecognized in French"]
+				"labels_tags_fr": ["végétarien", "Something unrecognized in French"],
```
alexgarel, line 768: in `tests/integration/expected_test_results/api_v3_product_write/patch-tags-fields-add-v3-6.json`
I have the "Something unrecognized in French" normalized :
```json
      "labels_tags" : [ 
         "en:vegetarian",
         "fr:something-unrecognized-in-french"
      ],       
      "labels_tags_fr" : [
         "Végétarien",
         "Something-unrecognized-in-french"
      ],
```
is this expected ?
In tags source, I indeed have:
```json
         "labels" : {
            "packaging" : {
               "last_updated_t" : "--ignore--",
               "tags" : [
                  "en:vegetarian",
                  "fr:Something unrecognized in French"
               ]
            }
         },
```

alexgarel (commented):
Not finished my review yet, but this is a chunk of it ;-)

lib/ProductOpener/Import.pm
alexgarel: I also see that docs/api/ref/schemas/product_tags_v3.yaml seems to lack a level !

it should have the tag name under "tags_sources" and then an object inside, where we have one entry per source name

lib/ProductOpener/Display.pm
stephanegigandet: I think we can make the assumption that tags that we already have in products don't have commas. We need to make sure that it's not possible to add one with a comma. But currently I think we always split on commas whenever we write tags.

lib/ProductOpener/Import.pm
stephanegigandet: Good catch, adding the missing level.

stephanegigandet: The constraint on source name is not in OpenAPI as we may add more sources. The constraint is for writing, but currently the source is not sent through OpenAPI when we write, it is inferred.

lib/ProductOpener/ProductSchemaChanges.pm
stephanegigandet: Good idea

lib/ProductOpener/ProductSchemaChanges.pm
stephanegigandet: I actually don't remember what I wanted to do, removing the comment.  ingredients_tags are not directly writable, so they are not in tags_sources anyway.

lib/ProductOpener/ProductSchemaChanges.pm
stephanegigandet: Good idea

lib/ProductOpener/Tags.pm
stephanegigandet: Yes, putting back 6 lines instead. :) I think it was auto-completed by Copilot, I prefer verbose code anyway.

lib/ProductOpener/Tags.pm
stephanegigandet: Right, I was changing things according to what we discussed on Monday in the call and I wasn't sure yet of how to do it.

lib/ProductOpener/Tags.pm
stephanegigandet: Yes, it's completely removed now.

lib/ProductOpener/Tags.pm
stephanegigandet: Right, I also renamed an old canonicalize_tag2() function to canonicalize_tag() :)

lib/ProductOpener/Tags.pm
stephanegigandet: Agreed, but there are already 83 $product_ref instances in Tags.pm in fact.

tests/integration/api_v3_product_read.t
stephanegigandet: A good question. It shows we have ingredients and that we didn't find allergens in them.

lib/ProductOpener/Tags.pm
stephanegigandet: Good idea, I'm creating a ProductsTags.pm module

alexgarel (approved):
I'll let you fix the tests, but for me, it's good to merge.

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -51,6 +51,7 @@ BEGIN {
 use ProductOpener::Config qw(:all);
 use ProductOpener::Store qw(get_string_id_for_lang);
 use ProductOpener::Tags qw(:all);
+use ProductOpener::ProductsTags qw/:all/;
```
Freso: what’s the difference between `qw/:all/` and `qw(:all)`?

lib/ProductOpener/Export.pm
```diff
@@ -585,6 +617,28 @@ sub export_csv ($args_ref) {
...
+				# Tags sources fields
+				elsif ($field =~ /^tags_sources\.(.*)\.(.*)\.last_updated_t$/) {
+					my ($tags_field, $source) = ($1, $2);
+					my $last_updated_t
+						= deep_get($product_ref, ("tags_sources", $tags_field, $source, "last_updated_t"));
+					if (defined $last_updated_t) {
+						$value = int($last_updated_t);
+					}
+				}
+				# $populated_fields{"tags_sources.${field}.${source}.tags"} = $field_sort_key . "_tags:${source}";
```
Freso: What’s the reason for leaving this in?

lib/ProductOpener/IngredientsStrings.pm
Freso, line 1: I agree that it’s great that this got refactored out to its own module… I just wish it was also its own PR, since I have no idea what things (if any) have changed here compared to before/current. 😅

Freso (commented):
This is way too much and way too Perl for me to feel confident with giving a thumbs up, but I at least didn’t spot anything that felt/looked _wrong_. 😅 It would have been nice if the big refactors moving code out into their own modules had been their own PR(s) so this one would only focus on the code that actually changed, rather than whole files that may or may not be identical to removed code elsewhere, but alas.

lib/ProductOpener/DataQualityFood.pm
stephanegigandet: It's the same, qw basically says quote the words and the character after specifies the start and end separator. I'll put ( ) evverywhere

lib/ProductOpener/IngredientsStrings.pm
stephanegigandet: Right, the thing is the refactors happened in the middle of the implementation, they were not planned initially, so it wasn't practical to do so unfortunately.

stephanegigandet:
> This is way too much and way too Perl for me to feel confident with giving a thumbs up, but I at least didn’t spot anything that felt/looked _wrong_. 😅 

Thank you @Freso !

lib/ProductOpener/Export.pm
stephanegigandet: Good catch, I removed it

PR #13240, opened 2026-03-05
fix: Miscellaneous fixes in preparation for importing Ingredient Products

john-gom:

### What
Fixes and other changes identified to support the import of products as nutrients. Includes the following:

 - ignoring estimated nutrients or those with a less than modifier in data quality checks
 - Move the logic for validating product codes into one place
 - Add water as a standard nutrient
 - API documentation fix
 - Some error logging and supporting wide characters in the minion log

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -1097,7 +1097,11 @@ sub nutrient_total_less_than_parts($nutrients_ref, $total_nid, @parts_nids) {
 		$parts += min_nutrient_value($nutrients_ref, $part);
 	}
 	# increased threshold from 0.001 to 0.01 (see issue #10491)
-	if (sprintf("%.2f", $parts) > sprintf("%.2f", $total + 0.01)) {
+	my $tolerance = 0.01;
+	if ($total >= 10) {
+		$tolerance = 0.1;
+	}
+	if (sprintf("%.2f", $parts) > sprintf("%.2f", $total + $tolerance)) {
 		return 1;
 	}
 	return 0;
--- a/tests/unit/dataqualityfood.t
+++ b/tests/unit/dataqualityfood.t
@@ -2019,6 +2019,32 @@ ok(
 	'sum of sugars and starch less than minimal traces of carbohydrates'
 ) or diag Dumper $product_ref;
 
+## Don't go more than 3 significant figures for larger values
+$product_ref = {
+	nutrition => {
+		input_sets => [
+			{
+				source => "producer",
+				preparation => "as_sold",
+				per => "100g",
+				nutrients => {
+					"carbohydrates" => {value => 10, unit => "g"},
+					"sugars" => {value => 9, unit => "g"},
+					"starch" => {value => 1.09, unit => "g"},
+				}
+			}
+		]
+	}
+};
+ProductOpener::DataQuality::check_quality($product_ref);
+ok(
+	!has_tag(
+		$product_ref, 'data_quality',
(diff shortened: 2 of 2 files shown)


lib/ProductOpener/DataQualityCommon.pm
```diff
@@ -101,7 +101,7 @@ sub check_bug_code_missing ($product_ref) {
 	elsif ($product_ref->{code} eq '') {
 		push @{$product_ref->{data_quality_bugs_tags}}, "en:code-empty";
 	}
-	elsif ($product_ref->{code} == 0) {
+	elsif ($product_ref->{code} eq 0) {
```
stephanegigandet: I'm guessing the original intention was to match 0, 00, 000000 etc.

"eq 0" would convert 0 to "0" before doing a string match

```suggestion
	elsif ($product_ref->{code} =~ /^0*$/) {
```

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -1050,6 +1050,36 @@ sub check_energy_for_input_set ($product_ref, $nutrition_set_ref, $set_id, $data
 	return;
 }
 
+sub min_nutrient_value ($nutrients_ref, $nid) {
+	# Treat component nutrient as having zero value f it is estimated or has a < modifier
+	my $modifier = deep_get($nutrients_ref, $nid, "modifier") // "";
+	if ($modifier eq "~" or $modifier eq "<") {
```
stephanegigandet: The modifier can be \N{U+2264} (the <= char) too

lib/ProductOpener/DataQualityFood.pm
```diff
@@ -1064,134 +1094,40 @@ sub check_specific_nutrients_for_input_set ($product_ref, $nutrition_set_ref, $s
...
-			(
-				(((defined $sugars_modifier) and ($sugars_modifier eq "<")) and ($sugars > ($carbohydrates) + 0.001))
-				or (    ((defined $starch_modifier) and ($starch_modifier eq "<"))
-					and ($starch > ($carbohydrates) + 0.001))
-			)
-		)
-		)
-	{
-
+	if (nutrient_total_less_than_parts($nutrients_ref, "carbohydrates", "sugars", "starch")) {
```
stephanegigandet, line 1126: Great!

lib/ProductOpener/Products.pm
```diff
@@ -527,6 +540,11 @@ sub split_code ($code) {
 		return "invalid";
 	}
 
+	if ($code =~ /^ingredient/) {
+		# Replace dashes with slashes
```
stephanegigandet: Is that just to put ingredients in different directories? I think we could just put them all in one dir if it's just a couple of thousands.

One thing we don't want is having ingredient:a-b directory inside the ingredient:a directory.

It's something that made a lot of mess previously with short codes, that was one of the reason to prepend them with 0s.  (e.g. if we had a product code "333", it had products "333/4567" in it, and then if we changed the code of 333 it would change all the paths under it...)

tests/unit/products.t
```diff
@@ -277,6 +283,7 @@ is(preprocess_product_field('labels', 'email@example.com, Green Dot'), ', Green
 is(preprocess_product_field('stores', 'Carrefour, abc@gmail.com'), 'Carrefour, ');
 
 is(split_code("26153689"), "000/002/615/3689");
+is(split_code("ingredient-en-test-ingredient"), "ingredient/en/test/ingredient");
```
stephanegigandet: I don't think we should do that.

john-gom: Agreed

PR #13247, opened 2026-03-06
fix: product revisions get saved even when there are no meaningful changes

bobcorn:
## Description

This PR fixes an issue where any attempt to edit a product would always result in a new noisy revision being created, even if no real changes were made to the product.

#### `lib/ProductOpener/Store.pm`
  - Exported the shared canonical JSON encoding so product comparison can consistently reuse the same serialization logic.

#### `lib/ProductOpener/Products.pm`
  -  Created an extensible list of field names whose changes are not meaningful, so they are ignored when evaluating a revision.
  - Introduced a deterministic product comparison mechanism based on a canonical JSON representation of two products.
  - The whole mechanism was implemented following a safety-first approach: we always save a new revision if any field not in the ignore list changes, if the revision is triggered by an internal structural migration, or if anything goes wrong during the normalization process.

#### `tests/integration/expected_test_results/api_v3_product_images_selected/get-product-image-crop.json`
  - Updated the `rev` count as a natural consequence of introducing the "no-op" skip mechanism. Some revisions of this product are no-ops, therefore the rev count needed to change accordingly, otherwise the test will always fail. This is expected to happen.

#### `tests/integration/expected_test_results/api_v3_product_write/patch-request-fields-all.json`
  - Same as above.

#### `tests/integration/products_comparison.t`
  - Integration tests for the product comparison mechanism.

#### `tests/unit/products_comparison.t`
  - Unit tests for the product comparison mechanism.

## Related issue(s)

Fixes [Stop Foodless Saving Pages without Contributing](https://github.com/openfoodfacts/contributor-quality-issues/issues/16)

Note: in the end, computing a SHA-256 hash of the normalized JSON was unnecessary, since comparing the canonical JSON directly gives the same result and avoids an extra step.

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -232,7 +232,14 @@ my @ignored_fields_for_revision_check = (
 	"last_modified_by", "last_modified_by_client",
 	"last_modified_t", "last_updated_t",
 	"rev", "editors_tags",
-	"last_editor"
+	"last_editor", "_keywords",
+	"allergens_from_user", "allergens_lc",
+	"categories", "categories_lc",
+	"countries", "countries_lc",
+	"labels", "labels_lc",
+	"origins", "origins_lc",
+	"traces", "traces_from_user",
+	"traces_lc"
 );
 
 # Make a clean copy before comparing products.


stephanegigandet:
Looks great, thank you @bobcorn !

bobcorn:
> Looks great, thank you @bobcorn !

Thank you, @stephanegigandet! I was just finishing up some final tests locally before publishing the PR for review. It's published now, feel free to take a look when you have time.

bobcorn:
Temporarily converted back to draft following suggestion from @john-gom to investigate whether we can reuse some common logic from [Store.pm](https://github.com/openfoodfacts/openfoodfacts-server/blob/bcf386de75554961a3b8a69fe22d0de7ec9089d1/lib/ProductOpener/Store.pm)

bobcorn:
Slightly refactored the PR to reuse the serialization logic from [Store.pm](https://github.com/openfoodfacts/openfoodfacts-server/blob/bcf386de75554961a3b8a69fe22d0de7ec9089d1/lib/ProductOpener/Store.pm). The implementation should now be consistent with the `Store` module, while the overall mechanism remains unchanged.

Thanks @john-gom for the great suggestion! Feel free to take a look and let me know if this matches your expectations.

The PR is now ready for review again.

lib/ProductOpener/Products.pm
```diff
@@ -222,6 +223,168 @@ sub make_sure_numbers_are_stored_as_numbers ($product_ref) {
...
+	# If either side cannot be turned into JSON, keep the save.
+	if ((not defined $current_comparison_json) or (not defined $latest_comparison_json)) {
+		return 0;
+	}
+
+	return ($current_comparison_json eq $latest_comparison_json);
+}
+
+# After a skipped save, put the stored product back into the product ref.
+sub _restore_product_state_from_latest_product ($product_ref, $latest_product_ref) {
```
john-gom, line 365: Why do we need to make a clone here as this would be an expensive operation? We have just loaded the latest product ref from a file so I'm not sure that there would be any side-effects from not cloning

bobcorn:
Temporarily converting back to draft while I'm working on a performance optimization. Thanks a lot @john-gom for spotting this!

lib/ProductOpener/Store.pm
```diff
@@ -315,6 +322,21 @@ sub write_json($file_path, $ref) {
...
 
+=head2 encode_canonical_json($ref, $strip_nuls = 0)
+
+Return a canonical JSON string for a reference.
+Set $strip_nuls to a true value to remove nul characters from the JSON text.
+
+=cut
+
+sub encode_canonical_json($ref, $strip_nuls = 0) {
+	my $json = $json_for_config->encode($ref);
```
john-gom, line 333: Note that json_for_config also adds indentation, which is unnecessary for your purposes. However, on benchmarking it didn't make a huge difference so happy if we keep it the same but maybe with a comment to explain

lib/ProductOpener/Products.pm
```diff
@@ -222,6 +223,168 @@ sub make_sure_numbers_are_stored_as_numbers ($product_ref) {
...
+	}
+
+	# Drop fields that should not create a revision.
+	my %prepared_product = %{$product_ref};
+	foreach my $field (@ignored_fields_for_revision_check) {
+		delete $prepared_product{$field};
+	}
+
+	my $normalization_failed = 0;
+	my $normalized_product_ref = _normalize_product_for_comparison(\%prepared_product, \$normalization_failed);
```
john-gom, line 312: _normalize_product_for_comparison already deep clones the object so maybe do this first. Could potentially add the fields to ignore to the normalize method (not passed any deeper than to top level) so that the ignored fields aren't created in the first place, so no need to delete them

john-gom: OK, benchmarked this and it didn't make a huge difference

lib/ProductOpener/Products.pm
bobcorn: Great point, the `dclone` was not really necessary so I removed it. Thank you!

lib/ProductOpener/Store.pm
bobcorn: Added a [comment](https://github.com/openfoodfacts/openfoodfacts-server/pull/13247/changes/a2616295bf0341071db1945442746ce15ef6dca3#diff-d6b9a7ba08e9ecb060d8284b1bd62bcd2861eccd7d3f65f479830b915c679761R322) in the code for this.

bobcorn:
I spent some time looking into whether the serialization step could be made faster, but that turned out to be quite difficult in practice. `Cpanel::JSON::XS` already seems to do a very good job here, and I could not find an optimization that felt both meaningful and safe without overcomplicating the PR.

One idea that came to mind, and that might be worth exploring, could be to compare digests of the normalized payloads instead of comparing the payloads directly, and to persist the digest of the normalized comparison payload in metadata stored with the product. That would not change much for the cold path, but in steady state it should remove roughly half of the serialization work, since we would only need to compute the digest for the current incoming revision.

That said, I did not feel super comfortable trying to introduce this optimization in the current PR. I have already tested the current implementation fairly extensively, and it seems quite stable at this point, so I would prefer not to risk introducing regressions here. It might also make sense to discuss the digest approach first in a dedicated issue, especially to align on where the digest should be stored and what the right mechanism would be. If it still seems worthwhile after that, we could handle it in a separate PR.

@stephanegigandet @john-gom feel free to let me know your thought on this! In the meantime, I am republishing the PR for review, so it can be evaluated again and eventually merged if you think it is in good shape.

bobcorn:
@stephanegigandet added you as a reviewer too following the dev meeting discussion, in case you'd like to take another look before merging as this touches a delicate area.

PR #13250, opened 2026-03-06
ci: make release please early fail and add assets

alexgarel:
* In case of a release, fail if we can't build docker images
* Also add more assets to a release: compiled taxonomies, and compiled translations

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,7 +1,5 @@
 name: 🚀 Adding the commit to the next release
 on:
-  # FIXME: temporary test
-  pull_request:
   push:
     branches:
       - main
@@ -12,14 +10,12 @@ jobs:
       # we try to know if it's a release before triggering RELEASE_PLEASE
       # To wait container build and eventually be able to fail early
       IS_RELEASE: "${{ startsWith(github.event.head_commit.message, 'chore(main): release') }}"
-      # FIXME: temporary, remove also all occurrence before merge
-      IS_RELEASE_TEST: true
       RELEASE_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
     steps:
 
       - name: Wait for frontend container build workflow
         # only if a new release was created
-        if: ${{ env.IS_RELEASE || env.IS_RELEASE_TEST }}
+        if: ${{ env.IS_RELEASE }}
         id: wait-build
         uses: tomchv/wait-my-workflow@v1.1.0
         with:
@@ -31,7 +27,7 @@ jobs:
 
       - name: Wait for backend container build workflow
         # only if a new release was created
-        if: ${{ env.IS_RELEASE || env.IS_RELEASE_TEST }}
+        if: ${{ env.IS_RELEASE }}
         uses: tomchv/wait-my-workflow@v1.1.0
         id: wait-build2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -63,7 +76,7 @@ jobs:
             if [[ $waited -gt 360 ]]; then
               echo "Timeout waiting for image $image_name"
               # Abort the job
-              exit 1
+              exit 1      - name: Get JS dist files to add to the release
```
Freso: Wrong copy-paste?

.github/workflows/release-please.yml
```diff
@@ -74,15 +87,75 @@ jobs:
...
+#        run: |
+#          TAG_NAME="${{ steps.release.outputs.tag_name }}"
+#          # remove eventual initial v
+#          STRIPPED_TAG_NAME=${TAG_NAME/v/}
+#          echo "TAG_NAME=$STRIPPED_TAG_NAME" >> $GITHUB_ENV
+      - name: Add the assets to the release
         # only if a new release was created
         if: ${{ steps.release.outputs.release_created }}
-        id: add-frontend-dist-to-release
+        id: add-assests-to-release
```
Freso:
```suggestion
        id: add-assets-to-release
```

alexgarel: thanks @Freso !

.github/workflows/release-please.yml
alexgarel: Indeed !
```suggestion
              exit 1
```

.github/workflows/release-please.yml
```diff
@@ -74,15 +87,75 @@ jobs:
...
+          docker cp -a po_off-backend-1:/mnt/podata/build-cache/taxonomies-result/ /tmp/release_data/food-taxonomies-result
+          # get compiled languages
+          docker cp po_off-backend-1:/mnt/podata/data/Lang.openfoodfacts.localhost.sto /tmp/release_data/lang/Lang.openfoodfacts.org.sto
+          docker cp po_off-backend-1:/mnt/podata/data/Lang_tags.openfoodfacts.localhost.sto /tmp/release_data/lang/Lang_tags.openfoodfacts.org.sto
+          # make archives
+          tar czf /tmp/release-assets/food-taxonomies-json.tgz -C /tmp/release_data/food-taxonomies-json .
+          tar czf /tmp/release-assets/food-taxonomies-results.tgz -C /tmp/release_data/food-taxonomies-results .
+          tar czf /tmp/release-assets/openfoodfacts-org-lang.tgz -C /tmp/release_data/lang .
+
+      - name: Add commit to release or release !
```
Freso:
```suggestion
      - name: Add commit to release or release!
```

`fr` vs. `en` punctuation; _very_ minor, feel free to ignore 🤭

.github/workflows/release-please.yml
```diff
@@ -1,59 +1,72 @@
...
       - main
 jobs:
   release-please:
     runs-on: ubuntu-latest
+    env:
+      # we try to know if it's a release before triggering RELEASE_PLEASE
+      # To wait container build and eventually be able to fail early
+      IS_RELEASE: "${{ startsWith(github.event.head_commit.message, 'chore(main): release') }}"
+      # FIXME: temporary, remove also all occurrence before merge
+      IS_RELEASE_TEST: true
```
Freso:
```suggestion
      # FIXME: uncomment the condition
      IS_RELEASE: true  # "${{ startsWith(github.event.head_commit.message, 'chore(main): release') }}"
```

Genuine question: Would it not be simpler to do something like this? Then you’d only have to fix it here before getting it ready for production and not have to deal with all the `||`s? Or is the actual `IS_RELEASE` condition also being tested here and I’m just not seeing it? Or is there some other reason that makes the two variables approach better? (Ultimately it doesn’t matter, since it’s only here for testing while developing it, I’m just curious. :))

alexgarel: Yes but I wanted to skip the real release_please action, that's why I prefered to do this like that.

PR #13261, opened 2026-03-08
feat: add Matomo event when deleting a product page (#13218)

### LINKED ISSUE #13218: (P0) Add Matomo event for deleting a product page

nithyaV-dev:
## What

This PR adds a Matomo tracking event when a product page is deleted.

The event is triggered when the delete confirmation form is submitted in the product edit interface.

## Why

Tracking this action helps us better understand how often product pages are deleted and provides more visibility into user interactions on the platform.

## Changes

- Added a Matomo `trackEvent` call when the delete product form is submitted.
- The event is triggered from the `product_edit_form_display_user-moderator.tt.html` template.

Fixes #13218

### DIFF
--- a/templates/web/pages/product_edit/product_edit_form_display_user-moderator.tt.html
+++ b/templates/web/pages/product_edit/product_edit_form_display_user-moderator.tt.html
@@ -16,4 +16,12 @@
 
 </form>
 
+<script>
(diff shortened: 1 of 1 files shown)


Freso (approved):
seems like a simple and sensible addition

PR #13273, opened 2026-03-10
taxonomy(food): some protein (bar) related edits

Freso:
- This adds `wikidata` entry for the ingredient `en:protein` and links it to the WD item for proteins in general.
- It also changes the `wikidata` property for the nutrient `zz:protein` to point to the WD item specifically about proteins as a human nutrient.
  - Also fixes `wikipedia` property which was previously a dead link.
- Adds ingredients etc. specifically needed for https://se.openfoodfacts.org/product/7350068295416/n%C3%B6t-cr%C3%A9me-propud

Needed-for: https://se.openfoodfacts.org/product/7350068295416/n%C3%B6t-cr%C3%A9me-propud


stephanegigandet (approved):
Thank you

Issue #13276, opened 2026-03-11
For some products, `aggregated_set.nutrients.[nutrient_name].source_per is `100kj`


raphael0202:
This shouldn't be allowed according to the schema.

Example product: https://world.openfoodfacts.org/api/v3.5/product/00433556?fields=nutrition

PR #13278, opened 2026-03-11
fix: Validate nutrition per values to prevent invalid source_per in aggregated sets

harsh-791:
## What
Added validation for **nutrition `per` field values** to prevent invalid `source_per` values (e.g., `"100kj"`, `"100kcal"`) from appearing in `aggregated_set.nutrients.[nutrient_name].source_per`.

## Why This Was Needed
- The API schema only allows the following `source_per` values: `["100g", "100ml", "serving"]`.
- Invalid values like `"100kj"` were being copied directly from input without validation.
- This caused schema violations, visible in products such as:  
  https://world.openfoodfacts.org/api/v3.5/product/00433556?fields=nutrition

## Changes Made
- Created a **`normalize_and_validate_per()`** function to validate values against allowed values.
- Applied validation at two entry points:
  - **API v3 requests:** Returns a proper error response for invalid values.
  - **Legacy field parsing:** Silently falls back to a default when extracting from old field names.
- Added **14 new unit tests**.
- Exported the validation function for better testability.

## Impact
- Prevents new invalid `source_per` values from being created.
- Existing products with invalid values will require **database cleanup** (e.g., regenerating aggregated sets).

### Verification
- **204 unit tests pass** in `nutrition.t`
- No compilation or lint errors
- Tested in Docker development environment

## Related Issue(s)
Fixes: **#13276**

### DIFF
--- a/lib/ProductOpener/Nutrition.pm
+++ b/lib/ProductOpener/Nutrition.pm
@@ -51,6 +51,7 @@ BEGIN {
 		&get_preparations_for_product_type
 		&get_pers_for_product_type
 		&get_default_per_for_product
+		&normalize_and_validate_per
 		&get_unit_options_for_nutrient
 		&normalize_nutrient_value_string_and_modifier
 		&assign_nutrient_modifier_value_string_and_unit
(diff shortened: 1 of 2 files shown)


harsh-791:
@raphael0202 can you please review this PR and suggest any changes if required

PR #13283, opened 2026-03-12
taxonomy(food): potato category edits

Freso:
The main purpose of this is to include a few more potato-based categories under `en:Potatoes and their products`, since that seems like it might be what the next Open Prices challenge centers around, based on the current votes on the post in #prices on Slack.

In addition to that, this also:
- Adds some Danish and Swedish translations.
- Adds some `wikidata` properties


taxonomies/food/categories.txt
```diff
@@ -81196,6 +81227,7 @@ hr: Jela od bulgura
 lt: Bulgur patiekalai
 
 < en:Meals
+< en:Potatoes and their products
```
moon-rabbitOFF:
```suggestion
```

The categories "(something) and their products" is for products derived from this product. E.g. fries, mashed potatoes, A meal contains a lot of ingredients: a potato dish is not derived from potato, it contains potatoes.

taxonomies/food/categories.txt
```diff
@@ -84279,6 +84311,7 @@ ciqual_food_name:fr: Gratin de pâtes
 #wikidata:en:
 
 < en:Gratins
+< en:Potatoes and their products
```
moon-rabbitOFF:
```suggestion
```

stephanegigandet:
Personal (but not very strong) opinion, for "X and their products", I would put only products that are described only with the X ingredient. If the product name contains another significant ingredient (not just a flavor or spice), like "Potatoes with chicken" or "Chicken with potatoes".

taxonomies/food/categories.txt
Freso: Potato gratins are 80+%, if not 90+%, potato, so I think they still qualify here. I’ll remove the potato meals from this PR though.

Freso: Upon actually reviewing some of the products in https://world.openfoodfacts.org/facets/categories/potato-gratin it seems like I was mistaken in how much potato they contain. I’ve changed this to instead be `< en:Potato dishes` now.

Freso:
@moon-rabbitOFF If you’re happy with the current state of the PR, feel free to approve so it can get merged. :)

PR #13287, opened 2026-03-14
taxonomy(food): fish analogue categories synonyms

Freso:
Adds some synonyms for the fish analogue categories that are already in use, and also some Danish and Swedish translations + synonyms.

Needed-for: https://world.openfoodfacts.org/facets/categories/vegan-fish
Needed-for: https://world.openfoodfacts.org/facets/categories/vegan-salmon
Needed-for: https://world.openfoodfacts.org/facets/categories/vegan-tuna

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -120264,16 +120264,22 @@ ciqual_food_name:fr: Saucisse végétale au tofu (convient aux véganes ou vég�
 
 #Meat and fish categories are separated while Fish analogues is under meat analogue. This difference is intended, since fishes are animals and therefore considered as meat by vegetarians
 < en:Meat analogues
-en: Fish analogues, Plant-based fish, Plantbased fish
+en: Fish analogues, Plant-based fish, Plantbased fish, Vegan fish
+da: Plantebaserede fisk, Veganske fisk
 hr: Analozi ribe
 nl: Plantaardige vissen
 pl: Zastępniki ryb
+sv: Växtbaserade fiskar, Veganska fiskar
 
 < en:Fish analogues
-en: Salmon analogues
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -120262,16 +120262,22 @@ ciqual_food_name:fr: Saucisse végétale au tofu (convient aux véganes ou vég
 
 #Meat and fish categories are separated while Fish analogues is under meat analogue. This difference is intended, since fishes are animals and therefore considered as meat by vegetarians
 < en:Meat analogues
-en: Fish analogues, Plantbased fish
+en: Fish analogues, Plantbased fish, Vegan fish
+da: Plantebaserede fisk, Veganske fisk
```
moon-rabbitOFF, line 120268: I think its great to put plantbased fish (with a hyphen maybe btw plant and based?) as a synonym. It will be more easy for people to find the category. But it might better to find another term for "Plantebaserede fisk" as a first entry in Danish, since by definition a fish cannot be plant-based

Freso: I agree in principle, and when I think of a suitable replacement, I will switch to that. For the time being though, I think _a_ Danish (and Swedish, which has the same issue) translation is better than _no_ Danish translation. It might even prompt some other Danish speaker/reader to suggest a better translation. :)

Freso: The lack of hyphen is not introduced by this PR, but I agree, and am fixing it in https://github.com/openfoodfacts/openfoodfacts-server/pull/13297

moon-rabbitOFF: Okay for me then, I agree better an approximate translation than no translation at all

PR #13289, opened 2026-03-14
fix: bypass cache for brand facet links so newly added products appear immediately

### LINKED ISSUE #13257: Recently added product does not appear immediately in brand facet
### What
- After adding a new product with brand "Bolas", the product does not appear in the brand facet page (`/brands/bolas`) when clicking the brand link from the product page.
- This creates inconsistency: the brand link suggests "see all products from this brand", but omits the just-added one.

1. Go to an existing product or add a new one: e.g. https://world.openfoodfacts.org/product/8906142774408/black-chia-seeds-bolas
2. On the product page, locate the "Brand" field (shows "Bolas").
3. Click the brand link "Bolas" (leads to https://world.openfoodfacts.org/brands/bolas or similar).
4. Observe that the newly added/edited product ("Black Chia Seeds") is missing from the list of products under that brand, even though other Bolas products are shown.

### Expected behavior
- When clicking on the Brand link (e.g. "Bolas") from a product page, the facet page should list **all** products for that brand, **including** the one the user just came from (or recently added/edited).
- Ideally, show the product immediately, or display a message like "This product was recently added — it may take a few minutes to appear in lists" if there's unavoidable indexing delay.

### Why
- Improves user experience and trust in the platform: contributors expect their edits to be visible right away in related views.
- Prevents confusion ("Did my edit save? Why isn't it here?").
- Common navigation flow (product → brand) should feel consistent and complete.

- Screenshot of product page showing brand "Bolas" link.

Image

- Screenshot of /brands/bolas page showing other products but missing "Black Chia Seeds".

Image

### Part of
- (If you find a related issue later, add e.g. #525 here; none matched exactly in current searches)

### Additional context
- Product added recently (check edit history on the product page for exact time).
- Brand page shows other Bolas products correctly, so it's not a complete failure to recognize the brand tag.
- Tested in Firefox 148 on Windows 10.

### Type of device
- Browser:  Firefox 148
- Operating System: Windows
- Other: Web interface (world.openfoodfacts.org)

- At least 1 (the newly added one); potentially many per day if contributors notice and get confused.
- Impacts all recently added/edited products until facets refresh.

### Time per product
- Minor confusion/annoyance per contributor (~1–5 min wondering if edit failed).
- If widespread, cumulative time loss for the community could be noticeable.

sidhansu10:
## Problem
When navigating from a product page to its brand facet page (e.g. `/facets/brands/<brand>`), recently added or edited products may not appear immediately because facet queries can be served from cache. This creates confusion for contributors who expect to see the product they just added.

## Solution
Add `no_cache=1` to brand facet links generated from product pages so the query bypasses cache and reflects the latest data. The change only applies to the `brands` tag type to avoid disabling caching for other facets.

- Fixes #13257


teolemon (changes requested):
You're tackling 2 unrelated topics in this PR. 
cc @john-gom  for the Cache part

Freso (commented):
> Add `no_cache=1` to brand facet links generated from product pages so the query bypasses cache and reflects the latest data. The change only applies to the `brands` tag type to avoid disabling caching for other facets.

Won’t users also expect the newly added product to show on category and other facet pages?

Maybe the `no_cache` query parameter could be added (for all facet links) on a product page when the product has just been added/edited (I don’t know if we have a good way to check if we come from the editing page or not) but not otherwise?

Ultimately, it would be even better if edited products could invalidate caches of facets they’re added to or removed from, instead of injecting a query parameter that probably shouldn’t be exposed in most circumstances. (I also feel like this is much less of an issue now than it was 2 months ago, but that doesn’t mean it won’t become an issue again.)

PR #13295, opened 2026-03-15
taxonomy: deleted croissants with almonds and pistachio

moon-rabbitOFF:
Deleted the category "croissants with almonds and pistachio", I cannot see any product in the databse which correspond to this description

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -119068,10 +119068,6 @@ sales_format:en: package, item
 en: Croissant with almonds, Almond croissant
 fr: Croissants aux amandes
 
-< en:Croissants
-en: Croissants with almonds and pistachio
-fr: Croissants aux amandes et pistache
-
 < en:Viennoiseries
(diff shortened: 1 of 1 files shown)


Freso:
Looked in the Slack, and it does indeed seem that this is a request from the Prices folks: https://openfoodfacts.slack.com/archives/C02Q6KM7F/p1717741906462959?thread_ts=1717135469.141019&cid=C02Q6KM7F

@raphodn can you confirm? It doesn’t seem like there are any prices for this category, but maybe I’m missing it?

moon-rabbitOFF:
Thank you @Freso for checking who made the former pull request and who asked for it. I am not familiar enough with github to check these kind of thing

raphodn:
hi ! thanks @Freso for the ping !
indeed i'm in favor of adding real-life categories that don't necessarily (yet) correspond to existing barcode products :innocent: 
mainly for open prices and allowing users to add prices for any kind of product 

but i do see some useless/duplicate categories like "* with bakery", but you're the experts if this info is useful ^^

PR #13303, opened 2026-03-17
fix: handle redis failures in listen_to_redis_stream.pl - not working

stephanegigandet:
See Slack thread: https://openfoodfacts.slack.com/archives/C02LDQDDD/p1773741118086119?thread_ts=1773681988.567029&cid=C02LDQDDD

In production listen_to_redis_stream.pl takes 100% CPU if redis_url is not configured, or if at some point there is an error reading from Redis.

This PR is an attempt to change the loop so that it does not happen, but it does not work,  xread somehow is not blocking for 5 seconds as it should.

### DIFF
--- a/lib/ProductOpener/Redis.pm
+++ b/lib/ProductOpener/Redis.pm
@@ -126,6 +126,7 @@ sub init_redis() {
 
 Connects to Redis and processes any events received. Blocks until there is an error or the application terminates.
 Returns on error or when receiving a terminate signal from the OS
+
 =cut
 
 sub subscribe_to_redis_streams () {


lib/ProductOpener/Redis.pm
```diff
@@ -123,72 +123,82 @@ sub init_redis() {
...
 
 	if (!defined $redis_client) {
 		$log->warn("Can't connect to Redis") if $log->is_warn();
-		return;
-	}
-
-	if (get_oidc_implementation_level() >= 2) {
-		# Read Keycloak events to process actions following user creation / deletion
-		_read_user_streams('$');
+		return 0;
```
stephanegigandet: So if for some reason redis becomes unreachable for some time, we just exit and never retry? Or is the idea that systemd will restart on failure after a while?

john-gom: We could retry indefinitetely if you like

stephanegigandet:
Tested locally, looks good. Thanks @john-gom !

PR #13313, opened 2026-03-19
fix: Logos for partners

teolemon:
fix: Logos for partners
Keeping duplicates in misc, but bringing a unified place to find them


stephanegigandet:
Thanks

PR #13317, opened 2026-03-20
fix: put back donation banner instead of Divinfood survey

stephanegigandet:
This is to put back the donation banner instead of the Divinfood survey:

image


templates/web/common/includes/donate_banner.tt.html
```diff
@@ -2,7 +2,7 @@
 
 [% banner_main = BLOCK %]
 
-[% IF product_type == "food" %]
+[% IF "divinfood_2026" == "divinfood_deactivated" && product_type == "food" %]
```
Freso, line 5: Shouldn’t this check for `!=` deactivated or `==` activated? It feels weird that we need to “deactivate” divinfood in order to show it, no?

PR #13320, opened 2026-03-20
feat: FDC import script

Vic142:
This PR adds a Python script to import products data from the [USDA'S FoodData Central](https://fdc.nal.usda.gov/).

The _BrandedFoods_ json export is used to create a CSV file to load onto the pro platform.

Most values are kept as is (nutrient values, FDC categories, ingredients, ...) in order not to lose any precision for nutrient values and to allow the Open Food Facts backend deal with logic concerning these fields.

Below is an extract from the CSV output from the current FDC json export file.
image

The script is currently run using *uv*. This can be changed if necessary.


scripts/imports/fdc/data/categories_mapping_fdc_off.json
john-gom, line 1: It might be nice to incorporate this into the off categories taxonomy by having a USDA category entry against the off category. That way the mapping can be seen and updated by more general taxonomy maintainers. Same for nutrient_map

scripts/imports/fdc/main.py
```diff
@@ -0,0 +1,63 @@
+import logging
```
john-gom, line 7: You could add [inline script metadata](https://packaging.python.org/en/latest/specifications/inline-script-metadata/#inline-script-metadata) here to avoid the need for a pyproject.toml or manual download of dependencies.

scripts/imports/fdc/fdc_import/uv.lock
john-gom, line 1: Maybe don't need a lock file as this is not part of the day to day production service

scripts/imports/fdc/script/mappers.py
```diff
@@ -0,0 +1,182 @@
...
+    quantity = normalize_empty_string(product.get("packageWeight"))
+
+    row = {
+        "code": code,
+        "name": normalize_empty_string(product.get("description")),
+        "last_updated_t": convert_to_seconds(
+            normalize_empty_string(product.get("modifiedDate"))
+        ),
+        "categories": normalize_empty_string(fdc_category_without_commas),
+        "countries": ["United States"],
```
stephanegigandet: countries should just be a string, with comma separated values if some day we get more than one market.

```suggestion
        "countries": "United States",
```

stephanegigandet (approved):
Thank you!

PR #13338, opened 2026-03-23
chore: Retry Redis errors indefinitely every minute

john-gom:
### What
If there is a failure reading from redis then retry every minute indefinitely


stephanegigandet (approved):
Thank you

PR #13345, opened 2026-03-24
chore(gen_feeds_daily): export MongoDB dump & JSONL first

raphael0202:
Currently, the CSV and RDF files are exported before the MongoDB and JSONL files.

There seems to be a bug in the `export_database.pl`, when we sort the CSV by code: the following call never ends, and the systemd service gets killed after 20h:

`system("(head -1 $csv_filename.temp && (tail -n +2 $csv_filename.temp | sort)) > $csv_filename.temp2");`

We need to investigate why this fails, but in the meantime, we can still have the JSONL and MongoDB dumps, which are the most important ones. The JSONL dump is notably used for the Parquet export.

In this PR, I change the run order and run MongoDB and JSONL export before the CSV/RDF export.

I also fixed a warning in remove_empty_products.pl script:

```
remove_empty_products.pl: Use of uninitialized value in numeric eq (==) at ./remove_empty_products.pl line 51.
```

### DIFF
--- a/scripts/gen_feeds_daily.sh
+++ b/scripts/gen_feeds_daily.sh
@@ -52,6 +52,9 @@ fi
 ./remove_empty_products.pl || report_error $? "remove_empty_products.pl"
 ./gen_top_tags_per_country.pl  || report_error $? "gen_top_tags_per_country.pl"
 
+# Generate the MongoDB dumps and jsonl export
+./mongodb_dump.sh $OFF_PUBLIC_DATA_DIR $PRODUCT_OPENER_FLAVOR $MONGODB_HOST $PRODUCT_OPENER_FLAVOR_SHORT || report_error $? "mongodb_dump.sh"
+
 # Generate the CSV and RDF exports
 ./export_database.pl || report_error $? "export_database.pl"
 
@@ -75,10 +78,7 @@ mc cp \
     s3/$PRODUCT_OPENER_FLAVOR-ds \
     || report_error $? "mc.cp.csv"
 
-# Generate the MongoDB dumps and jsonl export
 cd $OFF_SCRIPTS_DIR
-
-./mongodb_dump.sh $OFF_PUBLIC_DATA_DIR $PRODUCT_OPENER_FLAVOR $MONGODB_HOST $PRODUCT_OPENER_FLAVOR_SHORT || report_error $? "mongodb_dump.sh"
 
 # Small products data and images export for Docker dev environments
 # for about 1/100000th of the products contained in production.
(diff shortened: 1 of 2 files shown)


stephanegigandet:
> the following call never ends, and the systemd service gets killed after 20h:
> 
> `system("(head -1 $csv_filename.temp && (tail -n +2 $csv_filename.temp | sort)) > $csv_filename.temp2");`

That's strange! But this oneliner is a bit convoluted, maybe we could just output the header line and the body separately instead if we want to sort them with sort...

scripts/gen_feeds_daily.sh
```diff
@@ -52,6 +52,11 @@ fi
 ./remove_empty_products.pl || report_error $? "remove_empty_products.pl"
 ./gen_top_tags_per_country.pl  || report_error $? "gen_top_tags_per_country.pl"
 
+# Generate the MongoDB dumps and jsonl export
+./mongodb_dump.sh $OFF_PUBLIC_DATA_DIR $PRODUCT_OPENER_FLAVOR $MONGODB_HOST $PRODUCT_OPENER_FLAVOR_SHORT || report_error $? "mongodb_dump.sh"
+
+cd $OFF_SCRIPTS_DIR
```
Freso: Why is the `cd` after `./mongodb_dump.sh` now? In the removed section it `cd`s first, which suggests that the script is located in `$OFF_SCRIPTS_DIR`, and I don’t see anything moving the script. If the `cwd` at this point is already `$OFF_SCRIPTS_DIR`, then we shouldn’t need to `cd` at all… or am I missing something?
```suggestion
cd $OFF_SCRIPTS_DIR

./mongodb_dump.sh $OFF_PUBLIC_DATA_DIR $PRODUCT_OPENER_FLAVOR $MONGODB_HOST $PRODUCT_OPENER_FLAVOR_SHORT || report_error $? "mongodb_dump.sh"
```


```suggestion
./mongodb_dump.sh $OFF_PUBLIC_DATA_DIR $PRODUCT_OPENER_FLAVOR $MONGODB_HOST $PRODUCT_OPENER_FLAVOR_SHORT || report_error $? "mongodb_dump.sh"
```

raphael0202: Well this `cd` is actually not needed at all, indeed. We're already in the scripts dir.

edit: the `cd` was actually needed after compressing CSV exports, as we changed dir to `$OFF_PUBLIC_DATA_DIR`! I added the `cd` back. In the end, we only move the `./mongodb_dump.sh` call.

PR #13348, opened 2026-03-25
fix: CORS: allow X-User-Agent header

stephanegigandet:
This is to allow web app to send us a X-User-Agent header as they cannot modify the User-Agent header.


alexgarel (approved):
LGTM

PR #13349, opened 2026-03-25
refactor: remove CDN imports for jquery-ui and font-awesome in folksonomy.js

bhaskar21-7:
This PR removes external CDN imports for jquery-ui and font-awesome in folksonomy.js.

- jquery-ui is already bundled in the project
- font-awesome appears to be deprecated (Material Icons used instead)

The jqGrid CSS import is kept as it may still be required.

This change reduces dependency on third-party CDNs and improves reliability.


html/js/folksonomy.js
```diff
@@ -62,9 +62,10 @@ function folskonomy_engine_init() {
  * (Use //ajax... so that https or http is selected as appropriate to avoid "mixed content".)
  */
 
-@import url("https://ajax.googleapis.com/ajax/libs/jqueryui/1.11.4/themes/redmond/jquery-ui.css");
+/* Removed CDN imports for jquery-ui and font-awesome as they are already bundled or deprecated */
```
Freso: No need for a comment for every change made to the file, that’s what the `git log` is for. It might still be relevant to note that jquery-ui is bundled so it doesn’t need (external) `@import`ing, but there’s no need to mention every deprecated library that’s no longer in use. If it _is_ still in use anywhere in the `.js`, then those usages should be removed/fixed at the same time as the `@import`.

html/js/folksonomy.js
```diff
@@ -61,9 +61,10 @@ function folskonomy_engine_init() {
...
```
teolemon:
```suggestion
```

Freso (approved):

munzzyy:
Confirmed on current `main` that this is still needed. `html/js/folksonomy.js` still pulls jquery-ui from googleapis, jqgrid from rawgit, and font-awesome from bootstrapcdn.

Also checked the jquery-ui removal here against how the rest of the app loads it. `templates/web/common/site_layout.tt.html` already includes the gulp-built `css/dist/jqueryui/themes/base/jquery-ui.css` sitewide, and the only jQuery UI widget folksonomy.js actually uses is `.autocomplete()` (the `.dialog()` calls in this file are all commented out already), so dropping the CDN `@import` here shouldn't lose any styling since the base theme from the layout covers it. Font-awesome isn't referenced by any icon class in this file either, so that removal looks clean too.

Only thing left is teolemon's suggestion above (dropping the leftover comment line) — looks like a one-line fixup away from being mergeable. jqGrid/rawgit is still open after this lands since there's no vendored copy in `package.json` yet, probably worth its own follow-up issue rather than blocking this.

PR #13366, opened 2026-03-28
taxonomy: New category bhaji

moon-rabbitOFF:


taxonomies/food/categories.txt
```diff
@@ -123556,16 +123556,21 @@ ciqual_food_name:fr: Beignet de légumes
 intake24_category_code:en: 19VFRIT
 intake24_category_code:fr: 19VFRIT
 
-< en:Salty snacks
+< en:Salty fritters
+en: Bhaji, baji, bhajee
+xx: Bhaji, baji, bhajee
+Wikidata:en: Q680242
```
```suggestion
wikidata:en: Q680242
```

taxonomies/food/categories.txt
```diff
@@ -123556,16 +123556,21 @@ ciqual_food_name:fr: Beignet de légumes
 intake24_category_code:en: 19VFRIT
 intake24_category_code:fr: 19VFRIT
 
-< en:Salty snacks
+< en:Salty fritters
+en: Bhaji, baji, bhajee
+xx: Bhaji, baji, bhajee
```
Freso, line 123561: Would it make sense to preemptively make the primary/canonical synonym something that disambiguates it from <https://www.wikidata.org/wiki/Q13119441>?

taxonomies/food/categories.txt
```diff
@@ -123556,16 +123556,21 @@ ciqual_food_name:fr: Beignet de légumes
 intake24_category_code:en: 19VFRIT
 intake24_category_code:fr: 19VFRIT
 
-< en:Salty snacks
+< en:Salty fritters
```
Freso: Wikipedia says that this <q>is made from spicy hot vegetables,</q> would it make sense to put it under Vegetable fritters instead? (I don’t know Indian cuisine half as well as I would like, so I have no idea how correct WP is here.)
```suggestion
< en:Vegetable fritters
```

taxonomies/food/categories.txt
moon-rabbitOFF: I think the second one is really not common, so it shouldn't be an issue. We can create a category "Bhaji curry" later one when we see some corresponding products

taxonomies/food/categories.txt
moon-rabbitOFF: I put it in salty fritters in case there are exception. Pakora are also described as vegetable fritters in Wikipedia and yet, the majority of pakora products in OFF are chicken pakora 😆
I moved it in vegetable fritters, we can still change it again if some non-vegetable bhaji are appearing

PR #13367, opened 2026-03-28
feat: handle parser artifacts in single-ingredient checks (#8353)

### LINKED ISSUE #8353: Data quality error: category and computed Nutri-Score is not coherent
The Nutri-Score of some categories is predictable. Eg:
* Extra-virgin olive oils => Nutri-Score C
* Apples => A
* etc.

If an extra-virgin olive oils is not Nutri-Score C, then:
* either the category is wrong
* either the nutrition facts are wrong.
In both cases, it's an error, there can't be false positives. It should lead to a data-quality-error, and not just a warning.

To let it scale, I would be in favor to add this data in the taxonomy. Eg. (look at the last line):
```
<en:Virgin olive oils
en:Extra-virgin olive oils, Extra virgin olive oil
bg:Зехтин екстра върджин
de:Native Olivenöle Extra, Natives Olivenöl Extra
fr:Huiles d'olive vierges extra, Huile d'olive vierge extra
agribalyse_food_code:en:17270
ciqual_food_code:en:17270
ciqual_food_name:en:Olive oil, extra virgin
ciqual_food_name:fr:Huile d'olive vierge extra
nutri-score:c
```

adityav27:
- fixes #8353
### What
This PR resolves the 1000% spike in false positives for the `ingredients-single-ingredient-from-category-does-not-match-actual-ingredients` data quality check. 

**Why it was made:**
Previously, if the NLP parser extracted a correct ingredient but failed to parse additional descriptive marketing text (e.g., parsing both `"en:olive-oil"` and `"en:unknown-manually-harvested"`), `DataQualityFood.pm` would blindly throw a strict error simply because the array length was `> 1`. It failed perfectly valid products without checking if the correct ingredient was actually present.

**What was changed:**
I updated the validation block to scan the parsed ingredients array in $O(M)$ time:
1. **Maintained Strict Errors:** If the expected taxonomy ingredient is completely missing from the array (e.g., finding "sugar" when expecting "olive oil"), it pushes to `data_quality_errors_tags`. 
2. **Downgraded to Warning:** If the expected ingredient *is* found, but the count is `> 1` (due to unparsed marketing artifacts), it now pushes to `data_quality_warnings_tags`. This safely flags the text for human review without invalidating the product.

I used GitHub Copilot Chat to analyze the repository structure and simulate edge cases based on the existing `dataqualityfood.t` test suite.


adityav27:
@teolemon could you please take a look when you have time?

PR #13375, opened 2026-03-30
fix: speed up CSV export generation

stephanegigandet:
@raphael0202 noted that on off2, the sort of the CSV can be blocked for hours until it is killed by systemd.

I'm wondering if the tail + sort combination somehow makes things worse on very large files.

On my dev box I have timings that suggest extra disk operations (5 minutes total for tail -n + 2 on the OFF csv).

```
$ time tail -n +2 en.openfoodfacts.org.products.csv > t

real	1m38,033s
user	0m1,063s
sys	0m21,306s

$ time tail -n +2 en.openfoodfacts.org.products.csv | sort > t2

real	5m7,352s
user	0m18,017s
sys	1m24,379s

$ time (head -1 en.openfoodfacts.org.products.csv && (tail -n +2 en.openfoodfacts.org.products.csv | sort)) > t3

real	3m5,218s
user	0m18,160s
sys	1m27,329s
```

I also request less fields in the projection.

### DIFF
--- a/scripts/export_database.pl
+++ b/scripts/export_database.pl
@@ -59,6 +59,9 @@ use Encode;
 use POSIX qw(strftime);
 use Data::DeepAccess qw(deep_get);
 
+# 2023/03/30: we temporarily remove RDF exports as they take a lot of time to generate and do not seem to be in current use.
+my $export_rdf = 0;
+
 init_emb_codes();
 
 sub xml_escape_NFC($) {
@@ -132,7 +135,7 @@ foreach my $field (@export_fields) {
 $fields_ref->{empty} = 1;
 $fields_ref->{"nutrition.aggregated_set"} = 1;
 $fields_ref->{ingredients} = 1;
-$fields_ref->{"images.selected"} = 1;
+$fields_ref->{"images"} = 1;
 $fields_ref->{lc} = 1;
 
 # Current date, used for RDF dcterms:modified: 2019-02-07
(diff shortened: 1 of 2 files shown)


scripts/import_more_sample_data.sh
```diff
@@ -7,17 +7,17 @@ cd /tmp
 echo "\033[32m------------------ 1/ Retrieve products -----------------\033[0m";
 # explicitly specify the wget output file name so that wget does not append .1 if already present
 # e.g. if the tar command failed and the script was stopped
-wget -O products.tar.gz https://static.openfoodfacts.org/exports/products.random-modulo-1000.tar.gz 2>&1
```
raphael0202, line 10: What's this change about?

stephanegigandet: ah that was just for testing, i'll revert that part

scripts/export_database.pl
```diff
@@ -534,9 +529,11 @@ sub sanitize_field_content {
 	# only overwrite previous dump if the new one is bigger, to reduce failed runs breaking the dump.
 	my $csv_size_old = (-s $csv_filename) // 0;
 
-	print "Sort lines by code, except header line: $csv_filename.temp to $csv_filename.temp2\n";
-	system("(head -1 $csv_filename.temp && (tail -n +2 $csv_filename.temp | sort)) > $csv_filename.temp2");
+	print
+		"Sort lines by code, and add header line: add $csv_filename.temp.header + $csv_filename.temp to $csv_filename.temp2\n";
```
alexgarel, line 545: Why don't we ask MongoDB to sort items for us so that we don't need this  step ?

stephanegigandet: @alexgarel It's what we had before, but it didn't work as the database grew, there is another comment from earlier:

```
		# Removed sort({code => 1} in order to speed up the MongoDB query and not run into the error
		# "MongoDB::DatabaseError: Executor error during find command :: caused by :: Sort exceeded memory limit of 104857600 bytes, but did not opt in to external sorting."
```

PR #13379, opened 2026-03-31
ci: rotate log on staging deployment

alexgarel:
This is to avoid logs accumulation (which leads to high disk usage)


stephanegigandet (approved):
Thanks

hangy (approved):

PR #13381, opened 2026-04-01
 fix: Parse "* Svensk råvara" as Swedish origin indicator

### LINKED ISSUE #13258: Handle “* Svensk råvara”
### What
- Products using “* Svensk råvara” to denote that something’s `origins` is Sweden are not currently recognised.

1. Go to https://world-sv.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=Vetemj%C3%B6l*%2C+vatten%2C+surdeg+av+r%C3%A5g*+%2818%25%29%2C+vetegluten%2C+durumvetemj%C3%B6l+%284%25%29%2C+r%C3%A5gsk%C3%A5llning%2C+r%C3%A5gmj%C3%B6l*+%283%2C5%25%29%2C+salt%2C+j%C3%A4st%2C+m%C3%A4ltat+korn.+*+Svensk+r%C3%A5vara&type=add&lc=sv&action=process&estimator=product_opener&fat=&saturated-fat=&carbohydrates=&sugars=&fiber=&proteins=&salt=&submit=Send+foresp%C3%B8rgsel
2. Note that “Svensk råvara” is considered an ingredient rather than an `origin` denominator

### Expected behavior
- For the ingredients marked with “*” to have `origins` set to `en:sweden`.

[At least 7.](https://world.openfoodfacts.org/facets/ingredients/sv:svensk%20r%C3%A5vara)

dhiraj-143r:
- Fixes #13258.

Changes:
* Added Svensk råvara to taxonomies/labels.txt with origin en:sweden.
* Added en:swedish-raw-material to @labels in Ingredients.pm so the footnote resolver processes it.
* Added a preparsing test case using the exact ingredient text from the issue.

Why: Swedish products use * Svensk råvara to mark ingredients of Swedish origin. The parser previously treated this footnote as an unrecognized ingredient. This fix leverages the existing symbol-to-label resolution pipeline to automatically set the geographic origin of all starred ingredients to Sweden.


tests/unit/expected_test_results/ingredients_preparsing/231.json
Freso, line 1: Would probably be better to have a test that checks that the `origins` gets picked up.

taxonomies/labels.txt
```diff
@@ -4122,6 +4122,14 @@ ingredients:en: en:potato
 label_categories:en: en:Geographic label
 origins:en: en:sweden
 
+# "* Svensk råvara" is used as a footnote on Swedish products
+# to indicate that starred ingredients are of Swedish origin.
+# See https://github.com/openfoodfacts/openfoodfacts-server/issues/13258
+en: Swedish raw material
```
Freso, line 4128:
```suggestion
en: Swedish ingredient
```

lib/ProductOpener/Ingredients.pm
```diff
@@ -651,6 +651,7 @@ my @labels = (
 	"en:fed-without-gmos", "fr:crc",
 	"en:without-gluten", "en:sustainable-farming",
 	"en:krav", "en:whole-grain",
+	"en:swedish-raw-material",
```
Freso, line 654: This is why I didn’t want to go this route. If we have to add “en:danish-ingredient”, “en:indian-ingredient”, etc., this is going to be quite long. I was hoping to figure out a way to parse it similar to how we currently figure out when an ingredient is annotated with `from Palestine` or similar country/origin designation.

Freso (commented):
While this could work as a short-term stop-gap measure to handle this specific case, it would be better if this was handling generic “Xish ingredient” markers, so it could also pick up on “Syrian ingredient” or “Tanzanian ingredient” without further code changes. I think this happens in one of the functions in `Ingredients.pm` (`analyze_ingredient_function()`, `parse_origins_from_text()`, `parse_specific_ingredients_from_text()` might be good places to look to start, maybe).

Additionally, the issue states that the expected outcome is for associated ingredients to have their `origins` set appropriately, yet this isn’t actually tested here as far as I can tell.

lib/ProductOpener/Ingredients.pm
dhiraj-143r: @Freso 
I'll rework it with a generic parser instead. One thing though: "Svensk" (the adjective) isn't in the origins taxonomy, only "Sverige" is. Should I add adjective forms there, or handle it differently? Also, where should the logic go preparse_text() or elsewhere?

tests/unit/expected_test_results/ingredients_preparsing/231.json
dhiraj-143r: @Freso 
Got it, I'll add a test that verifies origins is correctly set on the starred ingredients.

lib/ProductOpener/Ingredients.pm
Freso: I don’t know. :) Discussion on https://github.com/openfoodfacts/openfoodfacts-server/issues/13258 regarding those questions should probably have been resolved before a PR was made.

You might want to ask for feedback in #productopener on Slack or similar channels.

Freso:
Going to close this for now, unless we decide in https://github.com/openfoodfacts/openfoodfacts-server/issues/13258 that this is the approach we want to take. Thank you for trying to tackle this though!

PR #13403, opened 2026-04-06
fix: keep double-quotes in product fields instead of encoding as &quot;

### LINKED ISSUE #12772: ingredient_spellcheck HTML encodes original text incorrectly
### What
It looks like Robotoff incorrectly encodes HTML entities in the JSON response (or potentially before sending them to the spellcheck API in the first place). This should be done in the Hunger Games frontend.

1. Go to https://hunger.openfoodfacts.org/ingredient-spellcheck and randomly land on an insight for a product with potential HTML entities in the text, ie. 4025500283148
2. See that the `"` in the ingredients list is incorrectly encoded
3. Check the JSON response https://robotoff.openfoodfacts.org/api/v1/insights?lc=de&insight_types=ingredient_spellcheck&annotated=false and see that `""` is already encoded as `&quot;` in the JSON:
4. ```json
    {
      "id": "5dccae13-c442-4f29-8a77-7f85a18d3dca",
      "barcode": "4025500283148",
      "type": "ingredient_spellcheck",
      "data": {
        "lang": "de",
        "original": "Buttermilch, Zucker, Wasser, 3% Heidelbeersaft&quot; aus Heidelbeersaftkonzentrat, 1,8% Airbende Pflanzenkonzentrate (Karotte, Aronia), Bananenpüree, Stärke, Zitronensaftkonzentrat, natürliches Aroma (enthält Milch), Laktase. 4,8 % Fruchtanteil im Endprodukt",
        "correction": "Buttermilch, Zucker, Wasser, 3% Heidelbeersaft\" aus Heidelbeersaftkonzentrat, 1,8% Apfelpflanzenkonzentrate (Karotte, Aronia), Bananenpüree, Stärke, Zitronensaftkonzentrat, natürliches Aroma (enthält Milch), Laktase. 4,8 % Fruchtanteil im Endprodukt",
        "lang_confidence": 0.88093853
      },
      "timestamp": "2025-06-11T09:48:32.684779",
      "completed_at": null,
      "annotation": null,
      "annotated_result": null,
      "n_votes": 0,
      "username": null,
      "countries": ["en:germany"],
      "brands": ["xx:muller"],
      "process_after": null,
      "value_tag": "de",
      "value": null,
      "source_image": null,
      "automatic_processing": false,
      "server_type": "off",
      "unique_scans_n": 1,
      "reserved_barcode": false,
      "predictor": "fine-tuned-mistral-7b",
      "predictor_version": "llm-v1-20241227172352",
      "campaign": [],
      "confidence": null,
      "bounding_box": null,
      "lc": ["de"]
    }
    ```

### Expected behavior
The product text should only be encoded by the website, making parts of the suggested correction unnecessary.

### Screenshots
<img width="819" height="402" alt="Image" src="https://github.com/user-attachments/assets/03009df5-3ac2-4175-92ba-53dd4e0a3075" />

### Platform (Desktop, Mobile, Hunger Games)
 - OS: Windows 10.0.22621.5624, Firefox 140.0.4
 - Platform: Hunger Games

areebahmeddd:
### What

* Fixed an issue where quotation marks (`"`) in product text were being converted into `&quot;` before being saved, causing some content (like ingredients) to display incorrectly
* This encoding should happen when displaying content, not when storing it, so it has been removed at the storage stage, while keeping encoding for `<` and `>` to avoid potential security issues

Used GitHub Copilot (Claude Sonnet 4.6) to understand the issue and help generate the Perl fix. Code was self-reviewed and tested locally with `make unit_test`

### Screenshot or video

image

- Fixes: #12772


lib/ProductOpener/Text.pm
```diff
@@ -307,7 +307,9 @@ sub remove_tags_and_quote ($s) {
 	$s =~ s/<(([^>]|\n)*)>//g;
 	$s =~ s/</&lt;/g;
 	$s =~ s/>/&gt;/g;
-	$s =~ s/"/&quot;/g;
```
hangy, line 310: I don't know precisely where this function is used, but the sub name explicitly mentions quote, so changing the behaviour in this way might cause unexpected side-effects. I think we probably might not want to replace `<`/`>` by `&lt;`/`&gt;` in ingredient texts, either. The texts are explicitly not stored as HTML and shouldn't be rendered as HTML without some sort of escaping on the consumer's side, IMHO

areebahmeddd: got it, i’ve restored the original behavior of remove_tags_and_quote. i added a new func that only strips tags without encoding and used it in preprocess_product_field. I think that should keep the existing callers unaffected

areebahmeddd:
closing this pr after discussing in product openerer call with stephane, this change might have a regression issue

PR #13411, opened 2026-04-08
taxonomy: Small Portuguese category PR

teolemon:
taxonomy: Small Portuguese category PR


Freso (approved):
My Portuguese isn’t the best, but these at least look reasonable to the tiny bit I do know.

PR #13417, opened 2026-04-08
taxonomy(food): coconut milk and cream edits

Freso:
- Separates out coconut milks and coconut creams from the “Coconut milks and creams” category.
- Adds `wikidata` properties specifically to milks and creams.
- Adds missing brands.
- Adds missing ingredient synonym.

Needed-for: https://se.openfoodfacts.org/product/7611612121347/coconut-milk-taste-from-asia


stephanegigandet (approved):
Thanks!

PR #13429, opened 2026-04-09
feat: recognise sv:“ursprungsland”

Freso:
Needed-for: https://se.openfoodfacts.org/product/7340083475603/fast-potatis-garant


benbenben2:
Would it be possible to add a screenshot? If you can run it locally. To be sure that it catch the regex.

Freso:
I’m not currently able to run it locally, no. I could add a test for it, but I’m also confident enough in my regex-fu that this should work (or at the very least not break anything).

stephanegigandet (approved):
Thanks

PR #13430, opened 2026-04-10
taxonomy: Place synonyms for nutrients in Japanese

Naruyoko:
### What
* Put the combined synonym first because synonyms are not usable on the web editor.
* Small miscellaneous fixes.


benbenben2 (approved):
Thank you @Naruyoko

PR #13438, opened 2026-04-11
fix: Remove unused $singular variable in Ingredients.pm

### LINKED ISSUE #13437: Clean up unused `$singular` variable in Ingredients.pm regexp builders
While reviewing `lib/ProductOpener/Ingredients.pm`, I noticed an unused variable `$singular` lingering in two of the regexp initialization functions (`init_allergens_regexps` and `init_labels_regexps`).

In both places, the pattern looks exactly like this:

```perl
my $singular = $suffix;     # <-- assigned here
$suffix =~ s/s$//;
$allergens_regexps{$allergens_lc} .= '|' . $suffix . '|' . $suffix . 's';

### PR DESCRIPTION
**Pull request by dhiraj-143r**
Fixes #13437 

While reviewing `lib/ProductOpener/Ingredients.pm`, I noticed an unused variable `$singular` in two regexp builder functions.

### What changed

Removed the dead `my $singular = ...` assignment from:
- `init_allergens_regexps()` (was at line 314)
- `init_labels_regexps()` (was at line 721)

In both cases, `$singular` was assigned the loop variable's value right before the loop variable was modified in-place with `s/s$//`, but `$singular` was never read after that.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -312,11 +312,11 @@ sub init_allergens_regexps() {
 		foreach my $suffix (sort {length($b) <=> length($a)} @allergenssuffixes) {
 			# simple singulars and plurals
 			my $singular = $suffix;
-			$suffix =~ s/s$//;
-			$allergens_regexps{$allergens_lc} .= '|' . $suffix . '|' . $suffix . 's';
+			$singular =~ s/s$//;
+			$allergens_regexps{$allergens_lc} .= '|' . $singular . '|' . $singular . 's';
 
-			my $unaccented_suffix = unac_string_perl($suffix);
-			if ($unaccented_suffix ne $suffix) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/ProductOpener/Ingredients.pm [hunk 1]
```diff
@@ -311,7 +311,6 @@ sub init_allergens_regexps() {
 
 		foreach my $suffix (sort {length($b) <=> length($a)} @allergenssuffixes) {
 			# simple singulars and plurals
-			my $singular = $suffix;
```
stephanegigandet, line 314: It might be that we wanted to use $singular in the regexp after, in order not to modify the entries in @allergenssuffixes 
So maybe it would be better to keep $singular but use instead of $suffix below.

dhiraj-143r:
@stephanegigandet 
I've updated the PR now $singular is kept as a safe copy and used in place of $suffix/$synonym below, so the original arrays stay unmodified

stephanegigandet (approved):
Thank you!

PR #13439, opened 2026-04-11
taxonomy(food): add rose hip soup category

Freso:
https://en.wikipedia.org/wiki/Rose_hip_soup

Needed-for: https://se.openfoodfacts.org/product/7350000550597/nypon-proviva


alexgarel (approved):
LGTM !

PR #13443, opened 2026-04-12
taxonomy: new stop word German

moon-rabbitOFF:
Added "enthalten sein" to deal with this kind of products https://world.openfoodfacts.org/product/4006040367826/vollmilch-kuverture-rapunzel


stephanegigandet (approved):
Thank you!

PR #13448, opened 2026-04-14
taxonomy(food): add dill and chive crisp category

Freso:
Needed-for: https://se.openfoodfacts.org/product/7340005403530/dill-gr%C3%A4sl%C3%B6k-olw


alexgarel (approved):
Thanks

PR #13449, opened 2026-04-14
taxonomy(food): add ranch crisp category

Freso:
Needed-for: https://se.openfoodfacts.org/product/7340005403691/ranch-olw

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -45205,6 +45205,12 @@ da: Dild og purløgschips
 sv: Dill och gräslökschips
 
 < en:Flavoured potato crisps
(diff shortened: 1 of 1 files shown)


moon-rabbitOFF:
Does "Ranch crisps" refers to a particular recipe followed by many brands or is it just a name that makes the product sell better? If it is a particular recipe, you can add the specificity in description maybe; if not I wouldn't create a category for it

Freso:
I don’t know whether they follow a particular recipe, but I believe that they aim for a taste akin to that of [ranch dressing](https://en.wikipedia.org/wiki/Ranch_dressing).

teolemon (approved):

PR #13459, opened 2026-04-16
taxonomy: Pickled vegetables and gherkins

moon-rabbitOFF:
* Moved "pickled vegetables" from "canned vegetables" to "Prepared vegetables". "Canned vegetables" is in "Vegetables" and "Pickled vegetables" are not vegetables anymore, but prepared ones. Same as for the meat which is cured or smoked and is then not in meat but in prepared meat
* Removed accordingly "Pickled gherkins" from "Gherkins" since gherkins is in vegetables
* put sweet and sour gherkins in pickled gherkins


taxonomies/food/categories.txt
```diff
@@ -109092,7 +109092,7 @@ ru: Маринованные растительные продукты
 sv: Inlagda växter
 intake24_category_code:en: PION
 
-< en:Canned vegetables
+< en:Prepared vegetables
```
Freso: Why not move Canned vegetables to be a child of Prepared vegetables instead? I don’t think there are any canned vegetables that are not in some kind of brine or other “soup” to help preserve them.

taxonomies/food/categories.txt
```diff
@@ -109185,7 +109185,6 @@ sv: Smågurkor
 en: Gherkins
 comment:en: “Gherkins” can refer to both non-pickled and pickled small cucumbers
 
-< en:Gherkins
 < en:Pickled cucumbers
 en: Pickled gherkins, Gherkins (pickled), Pickled small cucumbers
```
Freso, line 109754: I still don’t agree with this. If I’m browsing https://world.openfoodfacts.org/facets/categories/ I’d expect to be able to navigate from a vegetable to its pickled/canned version and vice versa. 🤷

taxonomies/food/categories.txt
moon-rabbitOFF: Most canned vegetables just have a small quantity of salt, not enough to change their A Nutriscore, so I think it's fine to let them in vegetables

taxonomies/food/categories.txt
moon-rabbitOFF: That's why we have categories "[something] and their products", wa can create a category "gherkins and their products" if you think it would be useful

taxonomies/food/categories.txt
```diff
@@ -109624,7 +109624,6 @@ sv: Smågurkor
 en: Gherkins
 comment:en: “Gherkins” can refer to both non-pickled and pickled small cucumbers
 
-< en:Gherkins
```
Freso: If you don’t want “pickled gherkins” to be under cucumbers, I think it’d be better to move “gherkins” out of it, as pickled products may still get added to “gherkins” due to its ambiguity.

moon-rabbitOFF: I'm fine with that option, I put gherkin under "Vegetables based foods". It doesn't follow the logical we apply elsewhere, but at least with don't have pickles under vegetables and we don't need to create multiple intermediary categories

Freso (approved):
I looked a bit more into this and I think it’d actually be better yet to just merge “gherkins” into “pickled gherkins”… but I’ll make a new PR for that once this has been merged,

Freso:
Follow-up PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/13568

PR #13486, opened 2026-04-20
taxonomy(food): mini and gluten-free baguettes

Freso:
Adds new categories:
- Gluten-free baguettes
- Mini baguettes
- Gluten-free mini baguettes

Needed-for: https://se.openfoodfacts.org/product/8008698048849/minibaguetter-sch%C3%A4r

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -36374,6 +36374,24 @@ ciqual_food_code:en: 7002
 ciqual_food_name:en: Bread, French bread (baguette or ball), with yeast
 ciqual_food_name:fr: Pain, baguette ou boule, au levain
 
+< en:Baguettes
+< en:Gluten-free breads
+en: Gluten-free baguettes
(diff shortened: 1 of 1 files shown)


moon-rabbitOFF:
I see only 1 gluten-free mini baguette products and 4 mini baguette products in total, it seems a bit too little to justify a category for me. When the word helps categorization (like the category taboon you created) it makes sense. But here users will put the category "baguettes" for those products, they don't need a finer category to be well categorized.

Freso:
https://world.openfoodfacts.org/cgi/search.pl?search_terms=mini+baguette returns 103 products. Not all of them would go under the "mini baguette" category, but there are definitely way more than 4 of them that would. And at least a handful of them seem to be gluten-free too.

teolemon (approved):

Freso:
74 products in https://world.openfoodfacts.org/facets/categories/mini-baguettes right now :) (15 of them in the gluten-free variant too)

PR #13502, opened 2026-04-24
ci: allow merge-conflict-autolabel to fail

Freso:
This action will sometimes gets cancelled on PRs, presumably because something else triggers it and it cancels itself to not run multiple times/to avoid race conditions.

However, this job cancellation counts as a failure and the PR will subsequently get a red “×” even if everything else passes.

This makes it so that the `triage` job’s step failing (e.g., getting cancelled) won’t mark the whole job as a failure.


teolemon (approved):

teolemon:
thanks @Freso

PR #13515, opened 2026-04-28
ci: only generate SBOMs when manifests change

Freso:
The SBOM generations seem very prone to breaking due to upstream connectivity/availability issues, leaving PRs with nasty red ×’s.

This adds some `*_sbom` filters on manifest files so we only generate SBOMs when any manifests actually change.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -27,6 +27,9 @@ jobs:
     outputs:
       code_modified: ${{ steps.filter.outputs.code }}
       taxonomy_only: ${{ steps.filter.outputs.any_non_taxonomy == 'false' }}
+      perl_sbom: ${{ steps.filter-some.outputs.perl_sbom }}
+      docker_sbom: ${{ steps.filter-some.outputs.docker_sbom }}
+      frontend_sbom: ${{ steps.filter-some.outputs.frontend_sbom }}
     steps:
       - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
       - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -44,6 +47,22 @@ jobs:
...
+            perl_sbom:
+            - 'cpanfile'
+            docker_sbom:
+            - 'docker/**'
+            - 'Dockerfile'
+            - 'docker-compose.yml'
+            frontend_sbom:
+            - 'docker/**'
+            - 'Dockerfile.frontend'
+            - 'docker-compose.yml'
```
Freso, line 62: I think these are all the relevant files, but I would really like a 👍 that I haven’t missed any.

teolemon: are the requirements.txt files in the script folders relevant ?

Freso: Not as far as I can tell. None of the SBOM files contain anything Python per se. The Perl SBOM only contained Perl things, and the Docker ones seem to only contain (Debian) packages.

alexgarel:
@hangy can you look at it ? I'm not sure if the JS libs (that are present on the frontend image) are taken into account by the SBOM generator.

.github/workflows/pull_request.yml
```diff
@@ -44,6 +47,22 @@ jobs:
...
           predicate-quantifier: 'every'
+      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
+        id: filter-some
+        with:
+          # FIXME: since this uses 'some' it isn’t able to exclude files (e.g., docker/*.md)
+          filters: |
+            perl_sbom:
+            - 'cpanfile'
+            docker_sbom:
+            - 'docker/**'
```
hangy: I think we can exclude `docker/**` for SBOM. Only the `Dockerfile` and `Dockerfile.frontend` should matter.

hangy:
> @hangy can you look at it ? I'm not sure if the JS libs (that are present on the frontend image) are taken into account by the SBOM generator.

Me neither. Maybe we should explicitly generate an SBOM from `package-lock.json`?

.github/workflows/pull_request.yml
Freso: So also not `docker-compose.yml` I guess?

teolemon (approved):

Freso:
Merging for now. We can adjust the files being in/excluded in the filter if it turns out this doesn't catch enough or catches too much, or if SBOMs are otherwise not getting generated when/where they need to be.

PR #13524, opened 2026-04-29
taxonomy(food): Portuguese+Brazilian brands

Freso:
Adds a selection of the most common and currently unrecognised brands on Portuguese and Brazilian OFF.

Reference: https://br.openfoodfacts.org/facets/brands
Reference: https://pt.openfoodfacts.org/facets/brands


stephanegigandet (approved):
Thank you!

PR #13525, opened 2026-04-29
taxonomy(food): add somun categories

Freso:
There’s both a Turkish bread and a Balkan (flat)bread going by the name of “somun”, so this adds both with a disambiguation parenthesis.

Needed-for: https://se.openfoodfacts.org/product/3870004009579/somun-br%C3%B6d-plivit
Source: https://www.tasteatlas.com/somun


teolemon (approved):
mmh, not quite sure about this one way to differentiate, but it's a limited impact

PR #13528, opened 2026-04-30
fix: rename func in test

teolemon:
_This PR applies 5/5 suggestions from code quality [AI findings](https://github.com/openfoodfacts/openfoodfacts-server/security/quality/ai-findings)._

### DIFF
--- a/lib/ProductOpener/Test.pm
+++ b/lib/ProductOpener/Test.pm
@@ -936,7 +936,7 @@ sub normalize_user_for_test_comparison ($user_ref) {
 
 Normalize a org to be able to compare them across tests runs.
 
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Test.pm
```diff
@@ -361,6 +361,11 @@ sub capture_ouputs ($meth) {
 	return ($out, $err, @result);
 }
 
+# Backward-compatible alias for historical typo.
```
stephanegigandet, line 364: This makes no sense, just replace capture_ouputs to capture_outputs everywhere in Test.pm and tests/integration/*.t

PR #13556, opened 2026-05-04
fix: send CORS headers for /cgi/search.pl API

stephanegigandet:
This PR adds CORS headers for API requests made to /cgi/search.pl queries (with a JSON response).


alexgarel (approved):
Thanks !

PR #13557, opened 2026-05-04
fix: change 'ouput' to 'output' in function names etc.

stephanegigandet:
Replace https://github.com/openfoodfacts/openfoodfacts-server/pull/13528

`find cgi tests lib scripts -o -type f -print0 | xargs -0 sed -i 's/ouput/output/g'`


alexgarel (approved):
A good ouput ;-)

PR #13559, opened 2026-05-04
feat: Move to Keycloak Level 5 (full implementation)

john-gom:
### What
Move to OIDC Implementation Level 5 for local development and in staging.

Also includes:
- fixes to maintain compatibility with the Dart SDK
- consistent replacement of "create an account" links in static pages
- ensuring the locale from the domain is passed through in the reset password link

None


hangy (commented):
The link "Account parameters" always forces me to sign in again, before I'm allowed to access the Keycloak account console. Do you know if this is some change we did intentionally, or just some oddity on my local setup?


Editing "Account parameters": Unable to save, even without any changes. "You need to enter a name or nickname." https://github.com/openfoodfacts/openfoodfacts-server/blob/ecbd201d82798a366e47f740a1058106b1e4c4a1/lib/ProductOpener/Users.pm#L606-L611 probably needs to be moved into the `if (get_oidc_implementation_level() < 5) {` guard

john-gom:
> The link "Account parameters" always forces me to sign in again, before I'm allowed to access the Keycloak account console. Do you know if this is some change we did intentionally, or just some oddity on my local setup?

The Keycloak account screen is seen as a different client, so it will start a separate session. It may silently redirect past the Keycloak login if you have logged in recently but otherwise will re-prompt.

hangy:
> > The link "Account parameters" always forces me to sign in again, before I'm allowed to access the Keycloak account console. Do you know if this is some change we did intentionally, or just some oddity on my local setup?
> 
> The Keycloak account screen is seen as a different client, so it will start a separate session. It may silently redirect past the Keycloak login if you have logged in recently but otherwise will re-prompt.

IIRC, I didn't have to sign in again to access the account console when it was initially implemented (especially when I had just signed into ProductOpener). Maybe it used to use some info from the Keycloak session cookie, but it's not that important. It's probably not a bad from a security perspective to have a new signin when editing those parameters.

john-gom:
> probably needs to be moved into the `if (get_oidc_implementation_level() < 5) {` guard

Done. Thnaks @hangy

tests/integration/create_user.t
```diff
@@ -70,4 +70,22 @@ $user = retrieve_user('bobnocountry');
 is($user->{email}, 'bobnocountry@example.com', "User created");
 is($user->{country}, "", "User created with no country");
 
+# Check legacy method sets locale and country from domain. This is how the Dart SDK currently does it
```
alexgarel, line 73: 👍

alexgarel (approved):
Great !

john-gom:
Looks like there is still an issue with editing preferences. I've created issue https://github.com/openfoodfacts/openfoodfacts-server/issues/13743. I'll try and fix as soon as possible if I have time. cc @hangy

john-gom:
> Looks like there is still an issue with editing preferences. I've created issue #13743. I'll try and fix as soon as possible if I have time. cc @hangy

Fixed here: [#13747](https://github.com/openfoodfacts/openfoodfacts-server/pull/13747). cc. @alexgarel @hangy

PR #13562, opened 2026-05-05
fix: allow input of sodium in Japanese

### LINKED ISSUE #13518: Unable to (explicitly) add sodium to nutrients on jp.OFF
### What
- The editor's default nutrient list in Japanese only has the salt-equivalent value, so it does not have the sodium row. This is correct to follow the usual list for Japanese labels. However, there is a bug where the sodium row can't be added explicitly. Sodium is shown in the dropdown, but selecting it does nothing.

1. Go to OFF Japan site
2. Add/edit a product
3. Try to explicitly add a sodium row

### Expected behavior
- Selecting sodium should add it to the nutrients list.

### Additional context

Reported by @Naruyoko on Slack: https://openfoodfacts.slack.com/archives/C031V3X7A/p1777442989524559

stephanegigandet:
Allow input of sodium in Japanese. fixes #13518

Additionally there was an issue with \ before Japanese commas that broke the JSON of nutrition tables with sodium.
The \ is only needed before normal commas.

Also do not display sodium in EU unless explicitly specified.


Freso (approved):
LGTM

Issue #13489, opened 2026-04-21
Organisation detected as existing during signup whereas it's not existing


manon-corneille:
When creating an account for a new organisation on the Open Food Facts Pro platform, the system incorrectly reports that the organisation already exists.

The problem was reported by the organisation "Andros" and I was able to reproduce the bug:

- Creating an account where org = "Andros" → system says the organisation already exists and my account creation is being moderated by an Open Food Facts admin 
- When I log in as an Open Food Facts admin, the organisation does not exist

<img width="1001" height="366" alt="Image" src="https://github.com/user-attachments/assets/9151129b-82f7-406d-a0d2-a301f098b40f" />

manon-corneille:
The system message when creating a new account is the following. 
The org-name doesn't seem to be picked up (there is a blank between "name" and ".") 


<img width="1262" height="440" alt="Image" src="https://github.com/user-attachments/assets/f87b5c8b-10b8-46b9-858b-352977242fa0" />

john-gom:
Alternative approach would be: https://github.com/openfoodfacts/openfoodfacts-server/issues/11867

Issue #11867, opened 2025-01-07
Remove producer from general user creation flow and support separate flow for producers


john-gom:
Proposed flow is documented [here](https://docs.google.com/document/d/1VBp7bNQ5QMs5okDx0taQqr-Lj45YaKy-jEIWKnK3hVY/edit?tab=t.0#heading=h.il3l2ciig0u1)

TODO commit here: https://github.com/openfoodfacts/openfoodfacts-server/commit/eb4d10af92ad2e80d9c796f032546faa5f9d5980

- [ ] Remove request to join producer from Keycloak user forms
- [ ] Allow any authenticated user to create an organization (with honeypot bot protection)
- [ ] Ability for authenticated user to request to join an existing org (possibly with bot protection)
- [ ] Move Odoo CRM sync to asynchronous for non-admin users
- [ ] Update invitation process to create pending users who are linked to org automatically when they create an account
- [ ] Update integration tests

teolemon:
@john-gom fyi, there's an Open PR attempting to solve this

john-gom:
Hi @teolemon I don't think this will solve the entire issue

PR #13563, opened 2026-05-05
fix: user and org creation on pro platform with OIDC >= 2

stephanegigandet:
At OIDC >= 2, the org creation / request when a user creates an account on the pro platform is not done synchronously anymore, instead Keycloak generates a REDIS event which is then processed by Minion.

As a result, we cannot tell the user right away after user creation if the org has been created or not.

Related to: https://github.com/openfoodfacts/openfoodfacts-server/issues/13489

### DIFF
--- a/cgi/user.pl
+++ b/cgi/user.pl
@@ -442,9 +442,13 @@ elsif ($action eq 'process') {
 
 		$template_data_ref->{user_requested_org} = $user_ref->{requested_org};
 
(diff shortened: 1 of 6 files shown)


john-gom:
@stephanegigandet this is the alternative workflow previously discussed: https://github.com/openfoodfacts/openfoodfacts-server/issues/11867

john-gom (approved):

PR #13565, opened 2026-05-05
fix: Catch errors when processing redis events

john-gom:
### What
Make sure any bad messages don't block the rest. Bad message details are logged in data/bad-messages.log

### DIFF
--- a/lib/ProductOpener/Redis.pm
+++ b/lib/ProductOpener/Redis.pm
@@ -260,14 +260,7 @@ sub process_xread_stream_reply($reply_ref) {
 			my @outer = @{$outer_ref};
 			$message_id = $outer[0];
 			eval {
-				my @values = @{$outer[1]};
-
-				my %message_hash;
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Redis.pm
```diff
@@ -255,36 +254,49 @@ sub process_xread_stream_reply($reply_ref) {
...
-	}
+		my $stream_name = $stream[0];
+
+		foreach my $outer_ref (@{$stream[1]}) {
+			my @outer = @{$outer_ref};
+			$message_id = $outer[0];
+			eval {
+				my @values = @{$outer[1]};
+
+				my %message_hash;
```
stephanegigandet: I think you can just do this: my %message_hash = @values;

stephanegigandet (approved):
Looks good, thank you!

lib/ProductOpener/Redis.pm
john-gom: Thanks, I think that worked...

PR #13567, opened 2026-05-06
fix: Fix MongoDB script should send an event for all product_types if affected

john-gom:
### What
Was previously only generating a Redis event if a food product was affected

None

Data fix for https://github.com/openfoodfacts/openfoodfacts-query/issues/260


Freso (approved):
Seems sensible to me.

PR #13571, opened 2026-05-07
chore: capacity increment

alexgarel:
* Give more capacity to apache servers now that we are on a bigger server at scaleway.
* fix some anomalous production log config making logs accumulate and slowing down server with disks writes


stephanegigandet (approved):
Thank you

teolemon (approved):
🎉

PR #13572, opened 2026-05-07
fix: Exclude other-flavors-codes when iterating over products

john-gom:
### What
product_iter was inlcuding the other-flavors-codes folder by default which it shouldn't have


stephanegigandet (approved):
Thank you

PR #13573, opened 2026-05-07
fix: handle CORS preflight before authentication (fixes /contributors)

### LINKED ISSUE #13555: CORS Preflight (OPTIONS) fails with 503 when authentication is required
### What
The server currently returns a 503 Service Unavailable for CORS preflight (OPTIONS) requests when the endpoint requires authentication. This occurs because the authentication layer expects a login cookie, which browsers do not include in preflight requests.

Because the 503 response lacks the necessary Access-Control-Allow-Origin headers, the browser blocks the subsequent actual request, even if the user has a valid session.

1. Go to https://openfoodfacts-explorer.vercel.app/
2. Login
3. Try to go to your account page (top right button)

### Why
- Web based API usage is broken as of right now.

### Type of device
- Browser: Firefox 150.0

TheSanjBot:
### What

Handle CORS preflight requests before authentication runs.

Browsers dont send authentication cookies with preflight `OPTIONS` requests, so routes could fail before CORS headers were added. This caused the block

This keeps the change narrow by returning early only for real preflight requests: `OPTIONS` requests with both `Origin` and `Access-Control-Request-Method` headers.

Added a regression test for a facet preflight request.

Used ChatGPT to understand the CORS code paths

- Fixes: #13555


alexgarel (approved):
LGTM, thanks !

alexgarel:
Thanks @TheSanjBot ! Really cool.

teolemon:
- @VaiTon @alexgarel @stephanegigandet merged. This blocks Explorer, so deployment would be nice. 
- We also created an EPIC with all the CORS issues blocking us: https://github.com/openfoodfacts/openfoodfacts-explorer/issues/1489

PR #13575, opened 2026-05-07
fix: always use prepared nutrition data to compute Nutri-Score, fixes case where we only have prepared data, with category that does not require it

stephanegigandet:
There was an issue with products with only prepared nutrition data but in categories that don't require nutrition data. We determined we had nutrition data to compute the Nutri-Score, but then later we were looking for values for 100g which we didn't have, and so we computed Nutri-Score with undef values, which gave incorrect result.

This PR makes it so that we always use prepared nutrition data for Nutri-Score, even if the category is not something like "product to be rehydrated" etc.


alexgarel (approved):
LGTM ! Thanks

PR #13588, opened 2026-05-10
taxonomy: add fdc categories as parent categories

Vic142:
This pull request aims to add some more FoodData Central's (FDC) categories into the OFF food category taxonomy to help during the FDC import.
In this PR, the FDC categories are added as parents of existing OFF categories in order to extract as much information as possible from the FDC categories of the future imported products.


taxonomies/food/categories.txt
```diff
@@ -77369,6 +77418,11 @@ pnns_group_2:en: Dressings and sauces
 who_id:en: 17
 wikidata:en: Q178359
 
+< en:Sauces
+en: Salad dressings and mayonnaises
+fr: Sauces salades et mayonnaises
+fdc_name:en: Salad Dressing & Mayonnaise
```
Freso: Does this _only_ contain mayonaises? No mustard, ketchup, etc.? That’s an interesting grouping. 😅

Freso (approved):
This looks good to me, at a glance. Might be good to get some others' opinions on new parent categories as well though.

taxonomies/food/categories.txt
```diff
@@ -52129,9 +52162,14 @@ pl: Masło kozie
 #wikidata:en:
 incompatible_with:en: categories:en:Sheep butters, Butters
 
+en: Butters and nut butters
```
aleene: I think it should be en: Butters and butter substitutes

Nut butters are not a substitute for butters

taxonomies/food/categories.txt
```diff
@@ -99496,6 +99557,7 @@ de: Maultaschen, Maultasche
 wikidata:en: Q20055
 
 < en:Biscuits and cakes
+< en:Desserts and toppings
```
aleene, line 99664: Not sure that pastries are viewed as a dessert.

taxonomies/food/categories.txt
```diff
@@ -106446,6 +106519,7 @@ tr: Ekmeğe sürülen ürünler, Sürülebilir ürünler
 zh: 塗抺食品
 
 < en:Spreads
+< en:Butters and nut butters
```
aleene: This should then be Butters and butter substitutes

taxonomies/food/categories.txt
```diff
@@ -124114,6 +124200,7 @@ ciqual_food_name:fr: Aligot (purée de pomme de terre à la tomme fraîche)
 wikidata:en: Q2724885
 
 < fr:Charcuteries diverses
+< en:Dips and pâté
```
aleene, line 124349: Maybe check what the FDC actually means by pate. I noticed that different countries mean different things.

aleene (changes requested):
OK great. Do the FDC categories also have an ID (number?), like used by GPC.

Freso:
I just went and got the “Full Download of All Data Types” from https://fdc.nal.usda.gov/download-datasets and the two category tables, `food_category` and `wweia_food_category`, seem to not really map that closely to the FDC categories you’ve been adding here. Can you say what your source(s) for FDC categories is/are?

taxonomies/food/categories.txt
```diff
@@ -35653,8 +35654,13 @@ gpc_category_code:en: 10000165
 gpc_category_description:en: Definition: Includes any products that can be described/observed as a leavened or unleavened, dough-based baked product. These products do not typically contain eggs and tend to have lower fat content than other dough products such as pastry. These products have been treated or packaged in such a way as to extend their consumable life. Includes such products as naan, chapatti, loaves, rolls, bagels, focaccia, ciabatta, baguette. Also includes rye bread, bread made from rice flour and part baked bread products. Definition Excludes: Excludes products such as Frozen and Perishable Bread Products, Dried Breads, Shelf-Stable Baking Mixes and Baking Supplies, and Pies and Pastries.
 gpc_category_name:en: Bread (Shelf Stable)
 
-< en:Breads
 < en:Frozen foods
+en: Frozen breads and frozen dough
```
```suggestion
en: Frozen breads and frozen doughs
```

taxonomies/food/categories.txt
```diff
@@ -38258,7 +38264,7 @@ gpc_category_description:en: Definition: Includes any products that can be descr
 gpc_category_name:en: Baking/Cooking Mixes (Perishable)
 
 < en:Baking Mixes
-< en:Frozen foods
+< en:Frozen breads and frozen dough
```
```suggestion
< en:Frozen breads and frozen doughs
```

taxonomies/food/categories.txt
```diff
@@ -52087,6 +52114,12 @@ nl: Plantaardige speltromen
 pt: Natas à base de espelta para cozinhar
 sv: Speltbaserade grädden för matlagning
 
+en: Eggs and egg substitutes
+fr: Œufs et substituts des œufs
+fdc_name:en: Eggs/Eggs Substitutes
+fdc_name_1:en: Eggs & Egg Substitutes
```
Freso: Did you ever read my comments on your other PR? I don’t like the idea of having `fdc_name_N` properties. It’d be better to do something like `fdc_name: First Name; Second Name`.

taxonomies/food/categories.txt
Freso:
```suggestion
en: Butters and butter substitutes
```

taxonomies/food/categories.txt
```diff
@@ -52129,7 +52162,12 @@ pl: Masło kozie
 #wikidata:en:
 incompatible_with:en: categories:en:Sheep butters, Butters
 
+en: Butters and nut butters
+fr: Beurres et beurres de fruits à coques
+fdc_name:en: Butter/Butter Substitutes
+
 < en:Animal fats
+< en:Butters and nut butters
```
```suggestion
< en:Butters and butter substitutes
```

taxonomies/food/categories.txt
```diff
@@ -89481,6 +89537,11 @@ agribalyse_food_code:en: 25089
 ciqual_food_code:en: 25089
 ciqual_food_name:en: Veal escalope cordon bleu (topped with a ham slice and Gruyere sauce)
 
+en: Fresh meats and fresh seafood
```
Freso, line 89620: maybe a `comment` that "poultry" is under meat in OFF, so no need to explicitly include it here?

taxonomies/food/categories.txt
```diff
@@ -103086,6 +103148,11 @@ intake24_category_code:en: RWFH
 plant_alternative:en: en:fish-analogues
 wikidata:en: Q600396
 
+< en: Fresh meats and fresh seafood
+< en:Fishes
```
```suggestion
< en:Fishes
< en:Fresh meats and fresh seafood
```

taxonomies/food/categories.txt
```diff
@@ -106384,6 +106452,11 @@ fr: Salades de fruits de mer
 en: Marinated seafood
 fr: Fruits de mer marinés
 
+en: Dips and pâté
+fr: Trempettes et pâté
+fdc_name:en: Dips/Hummus/Pate
```
Freso: again, it might be good with a `comment` that OFF’s "hummus" under "dips", so there’s no need to explicitly mention/include it

taxonomies/food/categories.txt
```diff
@@ -106445,6 +106518,7 @@ sv: Bredbara pålägg, Bredbara smörgåspålägg, Smörgåspålägg
 tr: Ekmeğe sürülen ürünler, Sürülebilir ürünler
 zh: 塗抺食品
 
+< en:Butters and nut butters
```
```suggestion
< en:Butters and butter substitutes
```

Freso (changes requested):
Looking more in-depth at this, I actually have some more comments too 😅

taxonomies/food/categories.txt
Vic142: I saw ranch, balsamic vinegar, vinaigrettes, but no ketchup (there is one item called "mayo-ketchup" which is [this item](https://world.openfoodfacts.net/product/0041331038713/mayo-ketchup-goya). There are mustard items though so I should probably link think category to the mustards category.

But I totally agree that this category is weird! Unfortunately it's not the only one that has weird groupings, but I didn't know what else to do that would make sense. If you think that this should be removed I will remove it and other weird categories!

Vic142:
> I just went and got the “Full Download of All Data Types” from https://fdc.nal.usda.gov/download-datasets and the two category tables, `food_category` and `wweia_food_category`, seem to not really map that closely to the FDC categories you’ve been adding here. Can you say what your source(s) for FDC categories is/are?

I am only adding branded foods from the Branded dataset. There are different categories for each dataset. I got the list of the categories from a script I did that checked all the items in the Branded food json download file and extracted each category in the field `brandedFoodCategory` in a set. I didn't find any documentation on the values of the branded food categories so this is the best I was able to do. If you go on the [FDC branded foods browsing page](https://fdc.nal.usda.gov/food-search?type=Branded) you can also see the categories in the `Branded Food Category` columns of the table.

Vic142:
> OK great. Do the FDC categories also have an ID (number?), like used by GPC.

I did not find any field mentioning a category id in the documentation or when checking the fields on products so I don't think branded food categories have an ID

taxonomies/food/categories.txt
Vic142: yes I did, this was probably some work I did before checking the comments on the other pr.

Is the separator `;` or `,` ? Because if it is a comma, I remember that I talked about it with Stéphane because some FDC categories have a comma inside their names so it would be a problem and that he said that I should keep the numbers for now because of it. But if it is a `;` that should be used to separate different categories inside one field then I can change it!

taxonomies/food/categories.txt
Vic142: that's done, but out of curiosity why change the order ?

taxonomies/food/categories.txt
Freso: Oh, this wasn’t a critique of the grouping being in here, just of the grouping existing in the first place. :)

taxonomies/food/categories.txt
Freso: Without the space between `: F`, `i` sorts before `r`. :)

taxonomies/food/categories.txt
Freso: `;` is my suggestion for a separator (which I’ve also mentioned before for maybe using as a separator for OBF’s taxonomy, since _a lot_ of OBF ingredients have commas in them). It is a standard ASCII/ANSI character, it is used _mostly_ in the same way a comma is (so its semantic meaning isn’t getting too overloaded), just much less frequently.

Another option could be the pipe character (`|`), but that one can be a bit harder to find on some keyboards (esp. mobile ones) and can also possibly get confused with `1`, `l`, `ı`, `I`, etc. depending on font and user’s familiarity with that symbol. Other options yet could be `~`, `^`, `@`, …

For the purpose of the brand new property, the separator can be whatever you want (as long as you don’t choose something that gets everyone up in arms about it :sweat_smile: ), just do it consistently, and preferably add a note on the wiki and/or in the comments at the top of the file.

Freso:
> I am only adding branded foods from the Branded dataset.

Would it maybe be better to use `fdc_branded_category:en:` instead? that feels like a better descriptor, and we could use `fdc_category_(name|code):en:` for the `food_category` if we ever want to crossreference with that, too. (And `wweia_category_(name|code)` could be an interesting one to have too, at some point.)

taxonomies/food/categories.txt
```diff
@@ -51594,6 +51615,11 @@ ciqual_food_code:en: 19433
 ciqual_food_name:en: cream, light, 4 to 8% fat, liquid or thick
 ciqual_food_name:fr: Spécialité à base de crème légère 8% MG, fluide ou épaisse
 
+en: Creams and cream substitutes
+fr: Crèmes et crèmes végétales
+fdc_name:en: Cream/Cream Substitutes
+
+< en: Creams and cream substitutes
```
```suggestion
< en:Creams and cream substitutes
```

taxonomies/food/categories.txt
```diff
@@ -89481,6 +89537,12 @@ agribalyse_food_code:en: 25089
 ciqual_food_code:en: 25089
 ciqual_food_name:en: Veal escalope cordon bleu (topped with a ham slice and Gruyere sauce)
 
+en: Fresh meats and fresh seafood
+fr: Viandes et produits de la mer frais
+fdc_name:en: Fresh Meat, Poultry and Seafood
+# Poultry is not explicitly included in the canonic category name because the poultry category is under meat in OFF
```
```suggestion
comment:en: Poultry is not explicitly included in the canonic category name because the poultry category is under meat in OFF
fdc_name:en: Fresh Meat, Poultry and Seafood
```

A comment attribute would be better than a `#` comment. :)

taxonomies/food/categories.txt
```diff
@@ -106384,6 +106453,12 @@ fr: Salades de fruits de mer
 en: Marinated seafood
 fr: Fruits de mer marinés
 
+en: Dips and pâté
+fr: Trempettes et pâté
+fdc_name:en: Dips/Hummus/Pate
+# Hummus is not explicitly included in the canonic category name because the hummus category is under dips in OFF
```
```suggestion
comment:en: Hummus is not explicitly included in the canonic category name because the hummus category is under dips in OFF
fdc_name:en: Dips/Hummus/Pate
```

Comment attribute is probably better than `#` comment here. :)

Vic142:
> > I am only adding branded foods from the Branded dataset.
> 
> Would it maybe be better to use `fdc_branded_category:en:` instead? that feels like a better descriptor, and we could use `fdc_category_(name|code):en:` for the `food_category` if we ever want to crossreference with that, too. (And `wweia_category_(name|code)` could be an interesting one to have too, at some point.)

That's a good idea, I will change it to this!

taxonomies/food/categories.txt
Vic142: ok thanks! I will use `;` since I haven't seen any FDC category with them and as you said it is common

PR #13591, opened 2026-05-12
fix: flush memcached in remove_all_results in tests

stephanegigandet:
Different test files make search queries (including searches with no filters which return all products in the database). Search results get cached in Memcached and depending on test order during runs, the tests can get stale cached search results.

This hopefully will fix the issues we have seen in the last few days with the cors.t test


teolemon (approved):
Nice catch !

alexgarel (approved):
👏

hangy (approved):

PR #13605, opened 2026-05-16
taxonomy(food): add brand Polarbröd

Freso:
Needed-for: https://world.openfoodfacts.org/facets/brands/polarbröd

### DIFF
--- a/taxonomies/brands.txt
+++ b/taxonomies/brands.txt
@@ -915,6 +915,9 @@ wikidata:en: Q136875384
 xx: Poiesz
 wikidata:en: Q137701150
 
+xx: Polarbröd
+wikidata:en: Q4567800
+
 #< xx:Intermarché
 xx: PorSi
 wikidata:en: Q139587415


teolemon (commented):
Isn't that a generic term @Freso ? We risk an issue with Robotoff.

Freso:
What do you mean by “generic term”? I don’t think I’ve seen or heard it used for anything other than a brand. The first page of https://duckduckgo.com/?q=Polarbr%C3%B6d is also exclusively about the brand.

teolemon:
Pain polaire or polar bread is a generic term in France and in the UK. So i assumed it might be the same elsewhere

Freso:
From a glance, it looks like the French (and probably British) term is derived from this very brand name and their commercialising of a traditional Swedish(/Sami?) bread type. In Sweden it’s known as tunnbröd and looks like most of the world (based on wikipedia/wikidata, so, questionable) also refers to it as such. Related: https://github.com/openfoodfacts/openfoodfacts-server/pull/13609

stephanegigandet (approved):
Thanks

PR #13606, opened 2026-05-17
refactor: Replace Time::Monotonic with Time::HiRes

hangy:
### What

Replace the `Time::Monotonic` dependency with `Time::HiRes` to provide monotonic clock functionality using `clock_gettime(CLOCK_MONOTONIC)`.

Reasoning: While testing OFF with Debian Trixie, I got Perl compiler issues while building `Time::Monotonic`. Turns out the Perl module hasn't been updated in [11 years](https://github.com/caldwell/Time-Monotonic/tree/0.9.8) and the underlying C library hasn't been updated in [7 years](https://github.com/ThomasHabets/monotonic_clock/tree/20b16818951d6a22a8904bf2a8f490b09ec5335b).
The replacement `Time::HiRes` also provides an option to get monotonic time. While it hasn't been updated since [2020](https://metacpan.org/dist/Time-HiRes/changes) either, it is a standard Perl module ([repo](https://github.com/Perl/perl5/tree/blead/dist/Time-HiRes), [list](https://perldoc.perl.org/modules#Standard-Modules)) and stands better chances of being kept up-to-date.

### Large Language Models usage disclosure

None.

### Screenshot or video

Not applicable.

### Related issue(s) and discussion

None.

PR #13621, opened 2026-05-18
build(docker): Update backend image to use Debian Trixie

hangy:
### What

Update base image from Debian Bullseye to Debian Trixie.

OpenCode Zen (Big Pickle Free) used to research correct Debian packages for `Imager::File::HEIF` migration, and to research change JSON serialization behaviour.

### Screenshot or video

Not applicable.

- Private Slack discussion: https://openfoodfacts.slack.com/archives/G039LDNF3/p1779020414602719
- Includes #13606  to resolve incompatibility with trixie.


tests/unit/expected_test_results/ingredients_contents/individual_ingredients.json
```diff
@@ -122,14 +122,14 @@
...
-      "is_fruits_vegetables_nuts_olive_walnut_rapeseed_oils" : 1,
+      "is_fruits_vegetables_nuts_olive_walnut_rapeseed_oils" : true,
       "is_milk" : 0
    },
    {
       "ingredient" : "chestnuts",
       "ingredient_id" : "en:chestnut",
       "is_fruits_vegetables_legumes" : 0,
-      "is_fruits_vegetables_nuts_olive_walnut_rapeseed_oils" : 1,
+      "is_fruits_vegetables_nuts_olive_walnut_rapeseed_oils" : true,
```
hangy, line 132: Different handling of truthy values due to Perl/Debian-provided JSON library?

tests/integration/expected_test_results/api_v3_product_read/get-existing-product-gs1-ai-data-str.json.metadata
```diff
@@ -1,9 +1,9 @@
 {
    "api_call" : {
-      "parameters" : {},
+      "body" : null,
```
hangy, line 3: JSON lib with different field sorting? Looks like alphabetical sorting now. (Which might not be ideal due to unnecessary performance impact.)

tests/integration/expected_test_results/api_v2_product_write/get-product-nutrition-new-fields-nutrition_data_per-100g-v3.json.metadata
```diff
@@ -4,8 +4,6 @@
       "content-type" : null,
       "method" : "GET",
       "parameters" : {},
-      "content-type" : null,
-      "method" : "GET",
```
hangy: Actually looks correct, as duplicate fields in JSON don't make any sense. Interesting behavioural change, though

tests/integration/expected_test_results/api_v3_taxonomy/canonicalize-ingredients-no-local-tags-list.json.metadata
```diff
@@ -3,7 +3,6 @@
       "body" : null,
       "content-type" : null,
       "method" : "GET",
-      "path" : "/api/v3/taxonomy_canonicalize_tags",
```
hangy: Field deleted?

tests/integration/expected_test_results/web_html/fr-categories.html
```diff
@@ -1651,7 +1651,7 @@ <h3 class="title-5 text-medium">Découvrez le projet</h3>
                   "grade":"e",
                   "icon_url":"//static.openfoodfacts.localhost/images/attributes/dist/nutrient-level-sugars-high.svg",
                   "id":"low_sugars",
-                  "match":8,
+                  "match":8.0,
```
hangy, line 1658: Change in accuracy of the underlying type?

tests/unit/expected_test_results/ingredients_contents/individual_ingredients.json
hangy: According to OpenCode Big Pickle:

> The change is in **Perl itself** (5.32 → 5.40). Since Perl 5.36, operators like `defined`, `eq`, `not`, and `and` return scalars with the internal `SvIsBOOL` flag set. `Data::Dumper` now shows these as `!!1`/`!!0`. The JSON encoder (`Cpanel::JSON::XS` or `JSON::PP`) in Trixie respects this flag and encodes those values as JSON `true`/`false` instead of `1`/`0`.
> You can see this in the function at `lib/ProductOpener/Ingredients.pm:8411` — its return value is a chain of (`defined ... and ... eq ... and not ... and not ...) or 0`. When the condition is true, the result is a flagged boolean (`!!1`) rather than a plain integer `1`. Functions like `is_milk` that use `$found = 1` (plain integer assignment) are unchanged.

For the unit test, this doesn't really matter. In general, it might be better to use `!!0` and `!!1` for truthy values to utilize `SvIsBOOL`.

tests/integration/expected_test_results/api_v3_taxonomy/canonicalize-ingredients-no-local-tags-list.json.metadata
hangy: Actually no - duplicate JSON property in the old version.

tests/integration/expected_test_results/web_html/fr-categories.html
hangy: Should not be an issue for consumers, because we already have other values like `"match":33.6666666666667,` in the same array.

Interesting analysis from OpenCode Zen (Big Pickle Free):

> It's a combination, but the dominant factor is Cpanel::JSON::XS becoming stricter about Perl's internal IV/NOK flags.
> The `match` value 8 comes from `lib/ProductOpener/Attributes.pm:1287`:
> 
> ```perl
> $match = 20 * ($high * 2 - $value) / $high;   # HIGH range, no integer addition
> ```
> 
> This is pure division — the result is stored as a float (NV-only flag). The LOW and MODERATE formulas have `80 +` > or `20 +` prefixes that cause Perl to also set the integer (IV) flag when the result is whole. Cpanel::JSON::XS 4.37+ checks the IV flag: if only NV is set, it serializes as `8.0` to preserve the float semantics. Older versions (Bullseye's 4.26) just wrote `8`.
> The fix would be `+ 0` to force the IV flag:
> 
> ```perl
> $match = 0 + 20 * ($high * 2 - $value) / $high;
> ```

(I think this would actually be wrong, because we don't want it to be forced into an int and don't want to lose precision.)

tests/unit/expected_test_results/ingredients_contents/individual_ingredients.json
hangy: (Not for this PR, though, because we might still have a setup with Debian Bullseye when completing this.)

tests/unit/expected_test_results/attributes/en-no-ingredients.json
```diff
@@ -877,7 +877,7 @@
             "fiber" : 0,
             "fruits_vegetables_nuts_colza_walnut_olive_oils" : 0,
             "is_beverage" : 0,
-            "is_cheese" : 1,
+            "is_cheese" : true,
```
hangy, line 876: `SvIsBOOL`

tests/integration/expected_test_results/api_v2_product_read/get-fields-all-knowledge-panels.json
```diff
@@ -3712,14 +3712,14 @@
          {
             "food_contact" : 0,
             "material" : "en:wood",
-            "number_of_units" : "1",
+            "number_of_units" : 1,
```
hangy, line 3721: via OpenCode Zen (Big Pickle Free):

> Yes, it's the same class of issue — **Perl/JSON library version change** — but through a different internal mechanism.
> 
> `number_of_units` is set via `$1 + 0` (e.g. `Packaging.pm:258`), which creates a **dualvar** — a Perl scalar that internally holds both an integer value (IV) and a string representation (PV from the regex capture `$1`). The regex check at line 419 (`=~ /^\d+$/`) also sets the string flag.
> 
> - **Bullseye** (Cpanel::JSON::XS ~4.27): when both IV and PV flags are set, the encoder heuristically prefers the string encoding, producing `"1"` in JSON.
> - **Trixie** (Cpanel::JSON::XS ~4.38): improved dualvar detection recognizes that the integer value and string value match, and now correctly encodes it as a JSON number: `1`.
> 
> So while it's also a library behavior change, it's about **dualvar handling** in Cpanel::JSON::XS rather than SvIsBOOL (booleans) or NOK-only (floats). The result is actually more correct — `number_of_units` is an integer field.

I agree with the conclusion: `number_of_units` is supposed to be an integer according to the schema:

https://github.com/openfoodfacts/openfoodfacts-server/blob/2c9dca61f8d938a02572caaec2a9e48c99b00d33/docs/api/ref/schemas/packagings/packaging_component.yaml#L28-L31

tests/integration/expected_test_results/api_v3_product_read/get-existing-product-gs1-ai-data-str.json.metadata
hangy: Perf impact is not really relevant for the integration test. If the library now canonicalizes the JSON by default, it's not a big issue in here.

teolemon:
small merge conflict @hangy

lib/ProductOpener/Test.pm
```diff
@@ -441,7 +441,12 @@ sub compare_to_expected_results ($object_ref, $expected_results_file, $update_ex
 				$title = $test_ref->{desc} // $test_ref->{test_case} // $test_ref->{id};
 				$title = undef unless $title;
 			}
-			is($object_ref, $expected_object_ref, $title) or diag(Dumper($test_ref), Dumper($object_ref));
+			# Normalize both objects through JSON round-trip for consistent boolean types
+			# (JSON::PP::Boolean vs Cpanel::JSON::XS::Boolean mismatch)
+			my $normalized_object_ref = $json->decode($json->encode($object_ref));
+			my $normalized_expected_object_ref = $json->decode($json->encode($expected_object_ref));
+			is($normalized_object_ref, $normalized_expected_object_ref, $title)
+				or diag(Dumper($test_ref), Dumper($object_ref));
```
alexgarel, line 449: Good idea.

alexgarel (approved):
Great PR.

alexgarel:
Kudos @hangy :tada:  I'm eager to also update production !

PR #13619, opened 2026-05-18
feat: Update donation form integration with Donorbox

teolemon:
Removed the loading message and old Donorbox widget code, added a new script for the Donorbox widget. migrate to OmniGive Form, as recommanded by Donorbox

"OmniGive offers faster loading speeds, over 20 localized payment methods for global donors, and an optimized checkout to boost conversion rates."
feat: Update donation form integration with Donorbox

https://donorbox.org/nonprofit-blog/donorbox-omnigive-donation-form

it says here that we should enable more payment methods using Stripe, including Revolut, Coins… (20 methods in total)

PR #13623, opened 2026-05-19
feat: harmonize to donorbox

teolemon:
- We had decided to harmonize to donorbox to simplify reporting, and since they were supporting fiscal receipts
feat: harmonize to donorbox
- Blocked on: https://github.com/openfoodfacts/openfoodfacts-server/pull/13619 will update the leftover from helloasso in fr.html to the new omnigive form from donorbox


alexgarel (approved):
Go !

PR #13630, opened 2026-05-20
fix: other is a valid image type for upload

alexgarel:
currently some images sent by our mobile app are blocked because having type like 'other_fr'

Screenshot_20260520-181219 OpenFoodFacts

### DIFF
--- a/lib/ProductOpener/Images.pm
+++ b/lib/ProductOpener/Images.pm
@@ -186,7 +186,8 @@ e.g. food, pet food and beauty products have an "ingredients" image type.
 
 %valid_image_types = (
 	front => 1,
-	packaging => 1
+	packaging => 1,
+	other => 1,
 );
 
 if (feature_enabled("ingredients")) {


teolemon (approved):

alexgarel (commented):
Fix might be incorrect…

alexgarel:
We have to understand why smoothie call crop image with other_fr as type.

lib/ProductOpener/Images.pm
```diff
@@ -186,7 +186,8 @@ e.g. food, pet food and beauty products have an "ingredients" image type.
 
 %valid_image_types = (
 	front => 1,
-	packaging => 1
+	packaging => 1,
+	other => 1,
```
alexgarel, line 190: This break tests, and I think this means this is the wrong fix.

stephanegigandet:
This error is returned only by process_image_crop.pl I think, so it should happen only for image selection, not upload.
image type "other" means the image is not selected, there's no "other_fr" etc.

teolemon:
@monsieurtanuki @alexgarel @stephanegigandet

monsieurtanuki:
> This error is returned only by process_image_crop.pl I think, so it should happen only for image selection, not upload. image type "other" means the image is not selected, there's no "other_fr" etc.

In Smoothie we use `process_image_crop.pl` only in two cases:
1. crop an existing image - typical use case
2. as a fallback when we cannot upload an image because it already exists - in that case we say "hey, this existing image can be considered as `$field_$language`"

Image the user uploads a picture as front_it.
If the user uploads the same picture as "other_it", it will fail because the picture exists, and then as a fallback it will be set as "other_it". Which isn't accepted by the server.

I've just managed to reproduce the issue in TEST env this morning with Smoothie:
* I've added an "other" picture to a product - for the record as "other_fr", which was successful
* I've added the same picture to the same product (as "other_fr")
  * it was rejected in a first approach (the very same picture is already uploaded)
  * as a fallback the code tried to set the existing picture as "other_fr", which is not accepted by the server as we saw earlier

A solution on the Smoothie (and off-dart) side would be:
* if the picture is already uploaded, don't play the fallback "set as" code IF THE FIELD IS "OTHER" - anyway we already have the picture uploaded, which is just what we need for OTHER

That would be quick to implement.

Are you guys OK with that fix?

alexgarel:
@monsieurtanuki thanks for testing it and for the analysis.
Yes your fix seems perfectly valid to me.

monsieurtanuki:
Thank you @alexgarel for your feedback!

The fix is already merged for Smoothie, and PR'ed waiting to be approved in off-dart (https://github.com/openfoodfacts/openfoodfacts-dart/pull/1240).

The question is now: does it make sense to PR on the server too?

alexgarel:
@monsieurtanuki nope we will close this PR, as the problem was on the smoothie app.

I will just verify the openapi docs is consistent (but in another PR).

teolemon:
@alexgarel @monsieurtanuki we won't be able to ship an updated version of smoothie before long. So a server side stopgap fix made sense to me

alexgarel:
https://github.com/openfoodfacts/openfoodfacts-server/pull/13883

@teolemon the problem is that this PR breaks things… and  @monsieurtanuki analysis shows that we are not loosing data, as it's images that already exists on the server.

PR #13632, opened 2026-05-20
chore(docker): Use cpm for CPAN installs

VaiTon:
### What

Replace cpanm calls with parallel cpm install. This should significantly reduce the backend image build time.

### Tests

With patch:

```
Executed in  251.91 secs    fish           external
   usr time    3.40 secs    0.00 millis    3.40 secs
   sys time    2.69 secs    9.17 millis    2.69 secs
```

Without patch
```
Executed in   15.93 mins    fish           external
   usr time    5.66 secs    0.40 millis    5.66 secs
   sys time    4.76 secs    8.47 millis    4.75 secs
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 ARG USER_UID=1000
 ARG USER_GID=1000
 # options for cpan installs
-ARG CPANMOPTS=
+ARG CPANMOPTS=""
 
 ######################
 # Base modperl image stage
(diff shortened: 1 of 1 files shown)


hangy (approved):
For some reason, we moved away from `cpm` when moving from alpine (which regularly caused build issues with Perl modules) to minideb in 29563f42dd81fd7eef06f5af3039315fe5909820. Not sure why. I liked `cpm`. 😆

PR #13633, opened 2026-05-21
chore: Add parallel Perl syntax checker script

VaiTon:
Replace make-based compilation with `scripts/perlcheck.pl`.

Add `check_perl_args` and introduce `CHECK_PERL_EXCLUDES` to control excluded files.


scripts/perlcheck.pl
alexgarel, line 1: @copilot this file is not following the conventions we normally use to develop in perl (named arguments for example).

See lib/ProductOpener/Test.pm as an example.

Please also start with `use ProductOpener::PerlStandards;` instead of all perl options you added.

VaiTon:
@alexgarel I'm still not sure about this approach, I want to try with another make-based approach, so I'm going to set this as a draft

teolemon:
@VaiTon @alexgarel small conflict, otherwise it looks good to merge ?

PR #13634, opened 2026-05-21
refactor: modernize taxonomy building with parallelization and Log::Any

VaiTon:
### What
This PR modernizes the taxonomy building process by:
- Replacing manual `print` and `warn` calls with `Log::Any` in `ProductOpener::Tags` to provide structured logging.
- Enabling parallelization in `build_tags_taxonomy.pl` using `fork`, which significantly speeds up the building of all taxonomies.
- Adding support for parallel jobs in the `Makefile` via the `jobs` parameter (defaulting to the number of CPUs).
- Improving script output with colored logs and a real-time progress indicator for interactive terminal sessions.
- Ensuring the build script exits with a non-zero status code on failure for better CI integration.

I used Gemini CLI (agentic) to review the commits and generate this PR description.

- Fixes: #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -494,24 +494,32 @@ build_packager_codes: create_folders
 	@echo "🥫 build packager codes"
 	${DOCKER_COMPOSE_BUILD} run --no-deps --rm backend /opt/product-opener/scripts/update_packager_codes.pl
 
+build_taxonomies: name ?= *
+build_taxonomies: jobs ?= $(CPU_COUNT)
 build_taxonomies: create_folders
 	$(MAKE) MOUNT_FOLDER=build-cache MOUNT_VOLUME=build_cache _bind_local
 	@echo "🥫 build taxonomies"
 # GITHUB_TOKEN might be empty, but if it's a valid token it enables pushing taxonomies to build cache repository
-	${DOCKER_COMPOSE_BUILD} run --no-deps --rm -e GITHUB_TOKEN=${GITHUB_TOKEN} backend /opt/product-opener/scripts/taxonomies/build_tags_taxonomy.pl ${name}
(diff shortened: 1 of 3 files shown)


stephanegigandet:
Thanks for the PR @VaiTon !

Did you try it? If so could you post some metrics of time saved? I think in practice ingredients.txt and categories.txt take most of the time, all others are much quicker to build.

You can force the rebuild by changing this in Tags.pm:

```
# Add a version string to the taxonomy data
# Change this version string if you want to force the taxonomies to be rebuilt
# e.g. if the taxonomy building algorithm or configuration has changed
# This needs to be done also when the unaccenting parameters for languages set in Config.pm are changed

my $BUILD_TAGS_VERSION = "20260413 - do not capitalize the first letter of all entries names and synonyms";
```

Maybe you could also check that we do get the same built taxonomies output.

One thing to watch as well is memory: the building might be very memory hungry on developer machines.

VaiTon:
@stephanegigandet the real time improvement is in #13643, but here there are timings for this PR:

Before:
```
❯ time make rebuild_taxonomies

Executed in   19.65 mins    fish           external
   usr time    1.79 secs    0.51 millis    1.79 secs
   sys time    1.01 secs    2.29 millis    1.01 secs

```

After:
```
❯ time make rebuild_taxonomies

Executed in  627.38 secs    fish           external
   usr time    1.73 secs    0.00 millis    1.73 secs
   sys time    1.13 secs    3.07 millis    1.12 secs

```

> Maybe you could also check that we do get the same built taxonomies output.

I checked, and will re-check, but this PR should not modify the algorithm, as the process is forked and the memory space is cloned so no shared vars whatsoever.

stephanegigandet (approved):
Thank you!

alexgarel:
Fantastic @VaiTon !

PR #13643, opened 2026-05-23
refactor: Improve tags taxonomy building performance 

VaiTon:
## Summary

Replace the `@smaller_synonyms` array with a Trie-based lookup for synonym substitution. Add a local memoization cache for `get_string_id_for_lang`. These two changes reduce the theoretical complexity of the synonym expansion phase from O(N²) to ~O(N).

### Trie-based synonym matching

The previous approach stored all candidate synonyms in a flat array and iterated over every candidate for each tag, using `index`/`rindex` plus prefix/suffix checks to detect where a synonym appeared. The new implementation builds a Trie keyed by hyphen-split words of each synonym. To find matches, it walks the Trie from each word position in the current tag, collecting matches in a single pass. The priority ordering (middle → end → start) is preserved via the same sequential fallback the old `if/elsif` chain used.

### Regex escaping with `\Q...\E`

Two literal-string comparisons (`$tagid =~ /\${tagid2_s}/` and `$tagid2_s =~ /\$tagid/`) previously passed user-derived strings directly into regex patterns. This was a latent ReDoS vector — a tag containing `(`, `*`, or nested quantifiers could cause catastrophic backtracking or silently match wrong substrings. Wrapping both sides with `\Q...\E` turns these into exact literal substring checks.

### Shallow copy instead of `clone()`

`$synonym_contains_synonyms` values are flat hashes (string keys → `1`). The old code called `clone()` (a deep copy) on them, which is unnecessary for one-level-deep structures. Replacing it with `{%{$hash}}` (a hash slice copy) avoids the overhead of walking nested structures that don't exist here.

### `splice`-based word substitution

The old code constructed a regex replacement dynamically — building patterns like `-tag2-`, `^tag2-`, or `-tag2$` and substituting via `s/$replace/$replaceby/e`. The new code splits the tag on `-`, then uses `splice` to swap the matched word range with the replacement synonym, and re-joins with `-`. This is equivalent semantically but avoids regex construction and evaluation entirely. The old approach also carried a stale `# TODO: why do we need /e here?` comment — that question is now moot.

## Performance

Build time for all taxonomies:

```
Before:
Executed in  642.86 secs    fish           external
   usr time    2.00 secs    1.08 millis    2.00 secs
   sys time    0.61 secs    1.03 millis    0.61 secs

After:
Executed in   33.97 secs    fish           external
   usr time    1.95 secs    0.26 millis    1.95 secs
   sys time    0.56 secs    2.08 millis    0.56 secs
```

### DIFF
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -203,6 +203,7 @@ use GraphViz2;
 use JSON::MaybeXS;
 
 use Data::DeepAccess qw(deep_get deep_exists deep_set);
+use Time::HiRes qw(gettimeofday tv_interval);
 
 binmode STDERR, ":encoding(UTF-8)";
 
@@ -1201,6 +1202,7 @@ Like "categories", "ingredients"
 =cut
 
 sub build_tags_taxonomy ($tagtype, $publish) {
+	my $t_start = [gettimeofday];
 	binmode STDERR, ":encoding(UTF-8)";
 	binmode STDIN, ":encoding(UTF-8)";
 	binmode STDOUT, ":encoding(UTF-8)";
@@ -1565,6 +1567,12 @@ sub build_tags_taxonomy ($tagtype, $publish) {
 
 		close($IN);
 
+		my $t_phase1 = [gettimeofday];
+		$log->info("Taxonomy Phase 1 (Read) completed", { 
+			tagtype => $tagtype, 
+			duration => tv_interval($t_start, $t_phase1) 
+		});
+
 		if (scalar @taxonomy_errors) {
 
 			$log->error("Errors in the $tagtype taxonomy definition:");
@@ -1796,6 +1804,12 @@ sub build_tags_taxonomy ($tagtype, $publish) {
 				}
 			}
 		}
+
+		my $t_phase2 = [gettimeofday];
+		$log->info("Taxonomy Phase 2 (Synonyms) completed", { 
+			tagtype => $tagtype, 
+			duration => tv_interval($t_phase1, $t_phase2) 
+		});
 
 		# 3rd phase: compute the hierarchy
 		# there we will associate each tags with its parent
@@ -2305,6 +2319,12 @@ sub build_tags_taxonomy ($tagtype, $publish) {
 			store_config("$result_dir/$tagtype.errors", {errors => \@taxonomy_errors});
 			put_to_cache($tagtype, $cache_prefix);
 		}
+
+		my $t_end = [gettimeofday];
+		$log->info("Taxonomy build finished", { 
+			tagtype => $tagtype, 
+			total_duration => tv_interval($t_start, $t_end) 
+		});
 	}
 
 	return @taxonomy_errors;
--- a/lib/ProductOpener/Tags.pm
+++ b/lib/ProductOpener/Tags.pm
@@ -1350,6 +1350,13 @@ sub build_tags_taxonomy ($tagtype, $publish) {
 	# $tagtype -> $canon_tagid -> "$property:$lc" stores the value for property
 	$properties{$tagtype} = {};
 
+	# Local cache for string normalization to avoid expensive Unicode NFC
+	my %id_cache;
+	my $get_id = sub ($l, $s) {
+		return "" if not defined $s or $s eq "";
+		return $id_cache{"$l|$s"} //= get_string_id_for_lang($l, $s);
+	};
+
 	my @taxonomy_errors = ();
 
 	if (open(my $IN, "<:encoding(UTF-8)", $file_path)) {
@@ -1398,7 +1405,7 @@ sub build_tags_taxonomy ($tagtype, $publish) {
 				$line =~ s/^\s+//;    # normalize spaces
 				my @tags = split(/\s*,\s*/, $line);    # split on comma
 				foreach my $tag (@tags) {
-					my $tagid = get_string_id_for_lang($lc, $tag);
+					my $tagid = &$get_id($lc, $tag);
 					next if $tagid eq '';
 					defined $stopwords{$tagtype}{$lc} or $stopwords{$tagtype}{$lc} = [];
 					defined $stopwords{$tagtype}{$lc . ".strings"} or $stopwords{$tagtype}{$lc . ".strings"} = [];
@@ -1424,7 +1431,7 @@ sub build_tags_taxonomy ($tagtype, $publish) {
 
 				# first entry gives id of tag
 				$lc_tag = $tags[0];
-				$lc_tagid = get_string_id_for_lang($lc, $lc_tag);
+				$lc_tagid = &$get_id($lc, $lc_tag);
 
 				# check if we already have an entry listed for one of the synonyms
 				# this is useful for taxonomies that need to be merged, and that are concatenated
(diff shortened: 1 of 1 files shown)


lib/ProductOpener/Tags.pm
```diff
@@ -1655,24 +1673,34 @@ sub build_tags_taxonomy ($tagtype, $publish) {
...
 					if (scalar @{$synonyms_for{$tagtype}{$lc}{$lc_tagid1}} > 1) {
 						# limit length of synonyms for performance
 						if (length($tagid) < (30 / $pass)) {
-							push @smaller_synonyms, $tagid;
-							#print "$tagid (canon: $lc_tagid1) has other synonyms\n";
+							my $curr = $substitutable_trie;
+							foreach my $w (split('-', $tagid)) {
+								$curr = $curr->{$w} //= {};
+							}
+							$curr->{_ID} = $tagid;
```
alexgarel: the code is already complex, can we abstract the trie behaviour in a few functions instead of having this raw access in the middle of the code ?

lib/ProductOpener/Tags.pm
```diff
@@ -1655,24 +1673,34 @@ sub build_tags_taxonomy ($tagtype, $publish) {
...
+					# Add to Trie if it qualifies as a substitutor
 					# Does $tagid have other synonyms?
 					if (scalar @{$synonyms_for{$tagtype}{$lc}{$lc_tagid1}} > 1) {
 						# limit length of synonyms for performance
 						if (length($tagid) < (30 / $pass)) {
-							push @smaller_synonyms, $tagid;
-							#print "$tagid (canon: $lc_tagid1) has other synonyms\n";
+							my $curr = $substitutable_trie;
+							foreach my $w (split('-', $tagid)) {
+								$curr = $curr->{$w} //= {};
```
alexgarel: IMO it's better to decompose this kind of statement.
Many of our contributors are not seasoned Perl developpers and 
will bug a bit on such a statement.
Two lines is better (there are other places like that in this PR).

lib/ProductOpener/Tags.pm
alexgarel: Shan't we define this _ID as a variable, say $trie_leaf ? It would be more readable (~~I don't even understand how it works this way…~~ oups of course this is equivalent to "_ID"… my perl practice isn't much those day :grimacing: )

alexgarel (commented):
@VaiTon thanks a lot for the PR, the idea is great.
I really would like the trie functions to be isolated before having another review.

Also did you test that the taxonomies remains the same before / after ? Maybe we should add a test that compile the test.txt taxonomy, adding some entries to ensure therse is some expansion (then use the json exports to compare, in this case test.extended.json, as well as the others)

teolemon:
@VaiTon 
test -z "lib/ProductOpener/Tags.pm" || COMPOSE_FILE="docker-compose.yml;docker/dev.yml" docker compose --env-file=.env  run --rm --no-deps backend perlcritic lib/ProductOpener/Tags.pm
Subroutine "trie_insert" does not end with "return" at line 1192, column 1.  See page 197 of PBP.  (Severity: 4)

lib/ProductOpener/Tags.pm
VaiTon: Done!

lib/ProductOpener/Tags.pm
VaiTon: Done!

VaiTon:
> Also did you test that the taxonomies remains the same before / after ?

I did manually, but I don't know how to test it in CI or anything. I will create a script and post it here.

VaiTon:
I'm attaching a script that I used to check that the before and after match

```
#!/bin/bash
set -e

# Script to verify that the refactored taxonomy builder yields identical outputs.
# Usage: ./scripts/taxonomies/compare_taxonomy_builds.sh [taxonomy_name]
# Example: ./scripts/taxonomies/compare_taxonomy_builds.sh categories

TAXONOMY="${1:-categories}"
TEMP_DIR="compare_temp"
mkdir -p "$TEMP_DIR"

CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
DOCKER_COMPOSE_CMD="env COMPOSE_FILE=docker-compose.yml;docker/dev.yml docker compose --env-file=.env"

echo "=== Step 1: Building taxonomy '$TAXONOMY' with the NEW implementation ==="
make rebuild_taxonomies name="$TAXONOMY" jobs=1

echo "Copying new result to $TEMP_DIR/$TAXONOMY.new.txt..."
$DOCKER_COMPOSE_CMD run -T --no-deps --rm backend cat "/mnt/podata/build-cache/taxonomies-result/$TAXONOMY.result.txt" > "$TEMP_DIR/$TAXONOMY.new.txt"

echo "=== Step 2: Stashing changes and checking out parent commit (old implementation) ==="
git stash
git checkout main

echo "=== Step 3: Building taxonomy '$TAXONOMY' with the OLD implementation ==="
# Clean old result inside the container first
$DOCKER_COMPOSE_CMD run -T --no-deps --rm backend rm -f "/mnt/podata/build-cache/taxonomies-result/$TAXONOMY.result.txt"
make rebuild_taxonomies name="$TAXONOMY" jobs=1

echo "Copying old result to $TEMP_DIR/$TAXONOMY.old.txt..."
$DOCKER_COMPOSE_CMD run -T --no-deps --rm backend cat "/mnt/podata/build-cache/taxonomies-result/$TAXONOMY.result.txt" > "$TEMP_DIR/$TAXONOMY.old.txt"

echo "=== Step 4: Restoring original branch ==="
git checkout "$CURRENT_BRANCH"
git stash pop || true

echo "=== Step 5: Comparing the outputs ==="
if diff -q "$TEMP_DIR/$TAXONOMY.old.txt" "$TEMP_DIR/$TAXONOMY.new.txt" > /dev/null; then
    echo "--------------------------------------------------------"
    echo "SUCCESS: The generated taxonomy results are IDENTICAL!"
    echo "--------------------------------------------------------"
    rm -rf "$TEMP_DIR"
else
    echo "--------------------------------------------------------"
    echo "FAILURE: The generated taxonomy results differ!"
    echo "--------------------------------------------------------"
    diff -u "$TEMP_DIR/$TAXONOMY.old.txt" "$TEMP_DIR/$TAXONOMY.new.txt" | head -n 50
    exit 1
fi

```

alexgarel:
Hum @VaiTon you compare the result.txt file but I don't think it is really meaningful.

I think you have to compare the exported json files (which are in a docker volume, so docker cp might be needed).

VaiTon:
@alexgarel you're right... and I found a mismatch :) I'll post another comment in a bit

VaiTon:
```diff
❯ diff -u compare_temp/categories.old.json  compare_temp/categories.new.json
--- compare_temp/categories.old.json	2026-06-27 00:15:02.755298811 +0200
+++ compare_temp/categories.new.json	2026-06-27 00:02:14.977010926 +0200
@@ -433379,7 +433379,6 @@
    "farina-di-farro":"farine-di-farro",
    "farina-di-farro-dicocco":"farine-di-farro",
    "farina-di-farro-medio":"farine-di-farro",
-   "farina-di-farro-medio-dicocco":"farine-di-farro",
    "farina-di-grano":"farine-di-grano",
    "farina-di-grano-duro":"farine-di-grano-duro",
    "farina-di-grano-saraceno":"farine-di-grano-saraceno",
@@ -933018,7 +933017,6 @@
     "farina-di-farro":1,
     "farina-di-farro-dicocco":1,
     "farina-di-farro-medio":1,
-    "farina-di-farro-medio-dicocco":1,
     "farine-di-farro":1,
     "farine-di-farro-dicocco":1,
     "farine-di-farro-medio":1
```

PR #13645, opened 2026-05-23
taxonomy(food): add French label synonyms

Freso:
Also some related English, Danish, and Swedish `synonyms`.

Source: https://eur-lex.europa.eu/eli/reg/2006/1924
Source: https://www.lexology.com/library/detail.aspx?g=59a9ec86-f345-4c3c-8209-7947c0ed40f1
Needed-for: https://se.openfoodfacts.org/product/5202390007934/savour-original-bloc-violife

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -83,10 +83,10 @@ synonyms:fi: merenantimet, merenelävät
 synonyms:ru: дары моря, морепродукты
 
 # https://www.lexology.com/library/detail.aspx?g=59a9ec86-f345-4c3c-8209-7947c0ed40f1
-synonyms:en: contains, with, source of
-synonyms:da: indeholder, med, kilde til
-synonyms:fr: contient, source de, avec du, avec
-synonyms:sv: innehåller, med, källa till
+synonyms:en: contains, source of
+synonyms:da: indeholder, kilde til
+synonyms:fr: contient, source de
+synonyms:sv: innehåller, källa till
 
 synonyms:en: without, 0%, free, contains no
 synonyms:da: uden, ingen, intet, nul, 0%, fri for
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -82,6 +82,12 @@ synonyms:bg: морска храна
 synonyms:fi: merenantimet, merenelävät
 synonyms:ru: дары моря, морепродукты
 
+# https://www.lexology.com/library/detail.aspx?g=59a9ec86-f345-4c3c-8209-7947c0ed40f1
+synonyms:en: contains, with, source of
+synonyms:da: indeholder, med, kilde til
+synonyms:fr: contient, source de, avec du, avec
+synonyms:sv: innehåller, med, källa till
+
 synonyms:en: without, 0%, free, contains no
 synonyms:da: uden, ingen, intet, nul, 0%, fri for
 synonyms:de: ohne, 0%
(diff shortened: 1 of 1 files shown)


taxonomies/labels.txt
```diff
@@ -2290,6 +2290,7 @@ sv: Berikad med vitamin D
 
 < en:Enriched with vitamins
 en: Enriched with vitamin B12
+fr: Et vitamine B12
```
Freso: English “with vitamin B12” is a synonym of “source of vitamin B12”, but does “with [micronutrient]” carry the same legal weight/implications that “source of […]/[…] source” does? And if so, is this the case for this French form too?

teolemon:
```suggestion
fr: Enrichi à la vitamine B12
```

Freso: Yeah, “et” is en:“and”, isn’t it? Thanks!

Freso:
I can’t determine what error this PR is causing in https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/26416279549/job/77761729582?pr=13645 :\ I think it’s something in `tests/unit/taxonomies_enhancer.t`, but I can’t figure out what exactly. Also, based on the tests getting updated by `/update_tests_results`, it could just be the order of some output that’s changed?

stephanegigandet (approved):
Thanks!

stephanegigandet:
Maybe @benbenben2 can help with the taxonomy_enhancer.t test

taxonomies/labels.txt
```diff
@@ -2290,6 +2296,7 @@ sv: Berikad med vitamin D
 
 < en:Enriched with vitamins
 en: Enriched with vitamin B12
+fr: Enrichi à la vitamine B12
```
fgouget: This should be "Enrichi en vitamine B12" just like it's "Enrichi en xxx" everywhere else in this file.

taxonomies/labels.txt
```diff
@@ -2122,7 +2128,7 @@ da: Kilde til vitamin B12
 de: Vitamin B12-Quelle
 es: Fuente de vitamina B12
 fi: B12-vitamiinin lähde
-fr: Source de vitamine B12
+fr: Source de vitamine B12, À vitamine B12
```
fgouget: "À vitamine B12" is not valid in French. Even "À la vitamine B12" does not feel right as it would normally be used to indicate a product's flavor like "Chewing-gum à la fraise" not the presence of a micronutrient. If the packaging wants to brag about the vitamin B12 content it would say "Source de vitamine B12", "Riche en vitamine B12" or "Enrichi en vitamine B12".

fgouget (commented):
The French translations for vitamin B12 are wrong.

taxonomies/labels.txt
Freso: > If the packaging wants to brag about the vitamin B12 content it would say […].

Okay. But it doesn’t. :shrug: I made this PR because of a product I came across, not just because I felt like adding a French phrase/word/synonym. You can see the product this PR is made for in the PR’s description, but I’ll include the relevant part of its front image here for easier reference: 
cropped front image, saying “BLOC”, “à l’huile de noix de coco”, “ET VITAMINE B12”, and “OUVREZ ICI”

Freso: (This is also why I initially added it as “et vitamine B12” because I didn't read the various bits of text as belonging together, and I forgot for a moment that “et” means “and”.)

fgouget: The way it's written, if you want to remove "à l'huile de noix de coco" then the only part to keep would be "vitamine B12". It's really like for flavors after all. One would normally say "À la fraise" as an abbreviation for "Glace parfumée à la fraise", but most often on the packaging all that's written is just "Fraise".

Here what they really mean is "Enrichi en Vitamine B12" (based on "Vitamine B12" being listed as a standalone ingredient), but all they wrote on the packaging is "Vitamine B12".

PR #13647, opened 2026-05-23
fix: install self-contained cpm

VaiTon:
Previously cpm was installed before the local lib env vars were set, causing it and its dependencies (notably YAML::PP) to be installed globally instead of into PO_LIB_DIR. 

Since only PO_LIB_DIR is copied to the prod image, YAML::PP was missing at runtime, breaking `make check_critic`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -102,6 +102,7 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt \
       apt-get update || true \
     ) && \
     apt-get install -y --no-install-recommends \
+        curl \
         #
         # cpan dependencies that can be satisfied by apt even if the package itself can't:
         #
@@ -187,6 +188,11 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt \
         libwebp-dev \
         libx265-dev
 
+RUN curl -fsSL https://raw.githubusercontent.com/skaji/cpm/main/cpm -o /tmp/cpm && \
+    mv /tmp/cpm /usr/bin/cpm && \
+    chmod +x /usr/bin/cpm && \
+    /usr/bin/cpm --version
+
 # Run www-data user AS host user 'off' or developper uid
 ARG USER_UID
 ARG USER_GID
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -214,17 +217,21 @@ RUN --mount=type=cache,id=apt-cache,target=/var/cache/apt \
...
-    export PERL5LIB="/tmp/local/lib/perl5/" && \
     # first install some dependencies that are not well handled
     cpm install --show-build-log-on-failure -w $(nproc) -g "Apache::Bootstrap" && \
     cpm install $CPANMOPTS --show-build-log-on-failure -w $(nproc) -g && \
     # Install the JUnit renderer separately so tests can keep using --renderer=JUnit
     # without adding an unresolved dependency back into cpanfile.
-    cpm install --show-build-log-on-failure -w $(nproc) -g "Test2::Harness::Renderer::JUnit"
+    cpm install --show-build-log-on-failure -w $(nproc) -g "Test2::Harness::Renderer::JUnit" && \
+    # delete cpm
+    rm $PO_LIB_DIR/bin/cpm
```
hangy: I think there are more leftovers from `cpm` itself:

```suggestion
    rm -rf $PO_LIB_DIR/bin/cpm $PO_LIB_DIR/lib/perl5/App/cpm.pm lib/perl5/App/cpm/
```

Additionally, `App::cpm` may install way more deps that aren't actually required to run ProductOpener? https://github.com/skaji/cpm/blob/96f9fb0bcb0a33c9b36358f78f6911cc1b0852ca/dist.pl#L17-L37 Maybe installing a [self-contained cpm](https://metacpan.org/release/SKAJI/App-cpm-v1.1.0/view/lib/App/cpm/Tutorial.pm#How-to-install-cpm) could be useful to avoid mixing cpm deps (at least if it gets removed afterwards) and ProductOpener runtime deps

VaiTon: I tried doing that, but then I have to modify the PERL5LIB env var to make CPM work, and with that done the same issue can reappear, i.e. libraries that are installed for CPM are not installed for PO

VaiTon: We can try using a self-contained CPM

VaiTon:
@hangy I pushed another version using the self-contained cpm

hangy (approved):

PR #13649, opened 2026-05-23
chore(taxonomy): remove old/ taxonomy files

Freso:
These files have only been touched in the last couple of years for syntax fixing etc., but several of them still have a fair number of issues, and it gets annoying to keep jumping through hoops to ignore them with doing `grep`s and such.

The files will still be available in the git repository’s history if anyone wants/needs to reference anything in them, they just won’t be present in the working directory and getting in the way of present-day operations. :)

Also, this leaves the unused/ files in place, as those are files that are not currently used, but that might eventually possibly be, so those make more sense to still maintain to some degree.


stephanegigandet:
Sounds good to me, let's share the PR in #taxonomies too

taxonomies/old/beauty/special_ingredients.txt
teolemon, line 1: this one was used for the Paraben/Phenoxyethanol detector. It's currently broken/disabled but was used

taxonomies/old/food/additives/additives.wikipedia.txt
teolemon, line 1: This was an extraction done by @stephanegigandet (or perhaps by me) to be able to show Wikipedia abstract on Additives knowledge panels. It worked at some point. AFAIK the script was nowhere to be found. We haven't done a real replacement. It showed initial content in many languages. Nowadays our value proposition is more complete, and having wikipedia extracts is probably less important

taxonomies/old/food/additives_classes/32008R1333.tmx
teolemon, line 1: *.tmx Those are Translation Memories from the European Union. Hopefully they have been leveraged into taxonomies and we don't need them anymore.

taxonomies/old/beauty/special_ingredients.txt
Freso: it’ll still be available at e.g. https://github.com/openfoodfacts/openfoodfacts-server/blob/8ccfe79371e377f5130ec02b8b910216019ea13f/taxonomies/old/beauty/special_ingredients.txt if we ever need to bring it back.

Freso:
This has been announced in #taxonomies and sitting for review for almost a week. Unless any objections are raised over the weekend, I’ll enable the auto-merge on Monday before the PO meeting.

stephanegigandet (approved):
Thank you!

PR #13653, opened 2026-05-25
fix: Iterating over `keys %$stats_ref` while deleting keys from the same hash inside the loop (lines 78-79) is unsafe in Perl

teolemon:
From Copilot:
```Iterating over `keys %$stats_ref` while deleting keys from the same hash inside the loop (lines 78-79) is unsafe in Perl and can lead to unpredictable behavior or missed iterations. You should collect all matching keys first (e.g., using `grep`) and then iterate over the collected list to safely delete from the hash.```
_This PR applies 1/1 suggestions from code quality [AI findings](https://github.com/openfoodfacts/openfoodfacts-server/security/quality/ai-findings)._


stephanegigandet:
Reported problem does not exist as "keys" is executed only once, and the "fix" just adds redundant code.

PR #13657, opened 2026-05-25
taxonomy: pizza pocket edit

moon-rabbitOFF:
I previously forgot to put pizza pockets in meals


teolemon (approved):
is pizza pies and quiche also in meals ?

stephanegigandet (approved):
Thanks!

PR #13660, opened 2026-05-27
fix: keep original unnormalized input values when upgrading non taxonomized fields

stephanegigandet:
This is a fix for the tags refactor, to upgrade existing tags field that are not taxonomies, such as "stores" and "purchase_places".
Previously, "stores" contained a comma separated list of the input values, and "stores_tags" contained an array of normalized values.
In the new tags refactor, we don't normalize (unaccent, lowercase) fields that are not taxonomized. This PR changes the product schema upgrade to take "stores" as input, instead of "stores_tags" which is normalized in the old schema (and not normalized in the new one).


alexgarel (approved):
Great !

PR #13662, opened 2026-05-27
taxonomy(food): edits for Community Co almond milk

Freso:
Did various clean-ups in related/surrounding taxons too.

Needed-for: https://world.openfoodfacts.org/product/9310246048546/unsweetened-almond-milk-community-co

### DIFF
--- a/taxonomies/additives.txt
+++ b/taxonomies/additives.txt
@@ -8029,8 +8029,6 @@ vegan:en: yes
 vegetarian:en: yes
 wikidata:en: Q221945
 
-# comment:en:Put the plural form in E322 and the specific singular forms as i and ii. Any origin soy or sunflower can be added to the ingredients.
-
 en: E322, Lecithins, Phosphatides
 xx: E322
 ar: E322, ليسيثين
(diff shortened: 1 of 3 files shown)


taxonomies/additives.txt
```diff
@@ -8029,9 +8029,7 @@ vegan:en: yes
 vegetarian:en: yes
 wikidata:en: Q221945
 
-# comment:en:Put the plural form in E322 and the specific singular forms as i and ii. Any origin soy or sunflower can be added to the ingredients.
-
-en: E322, Lecithins, Phosphatides
+en: E322, Lecithins, Phosphatides, 322
```
Freso: Will the system handle number-only ingredients okay? Should this instead be done in the parser, something like "this ingredient consists only of a number, let's try to prepend an `E` to it and see if that resolves"?

taxonomies/additives.txt
```diff
@@ -11544,6 +11543,7 @@ vegetarian:en: yes
 wikidata:en: Q4783458
 
 
+#< en:tree gum
```
Freso, line 11546: we can’t inherit from `ingredients.txt` in `additives.txt`, right? only the other way around?

Freso (commented):
A couple of questions regarding my changes. :)

taxonomies/additives.txt
stephanegigandet: This should be done by the parser if at all. There's a huge chance for false positives with some numbers (especially 100 -> e100).

taxonomies/additives.txt
stephanegigandet: Right

taxonomies/additives.txt
Freso: Dropped the non-E E numbers from the PR.

stephanegigandet (approved):

PR #13664, opened 2026-05-27
taxonomy: Fajitas sauces

moon-rabbitOFF:
* created "fajitas sauces"
* deleted "other sauces"

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -77578,15 +77578,12 @@ pl: Sosy wegetariańskie, wegetariańskie sosy
 sv: Vegetariska såser
 
 < en:Sauces
-en: Other sauces
-
-< en:Other sauces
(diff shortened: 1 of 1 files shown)


Freso (changes requested):
`en:Other sauces` is a parent of other categories, so if you want to remove that, you have to fix up its children (“[Refrigerated other sauces](https://world.openfoodfacts.org/facets/categories/Refrigerated%20other%20sauces)” and “[Shelf-stable other sauces](https://world.openfoodfacts.org/facets/categories/Shelf-stable%20other%20sauces)”, both of which are GPC categories).

(This is also why the test is failing.)

moon-rabbitOFF:
Ah right! Thx
We will have to solve this naming, but for now, I move them directly to sauces

Freso (approved):

PR #13671, opened 2026-05-29
taxonomy: week23 changes

aleene:
### What
small taxonomy updates, mainly adding Ciquals


taxonomies/food/ingredients.txt
```diff
@@ -47175,6 +47108,8 @@ wikidata:en: Q10978
 # wuu:葡萄
 # xmf:ჸურზენი
 # yue:提子
+ciqual_food_code:en: 13046
+ciqual_food_name:en: Raisin
```
Freso: This item already has a ciqual code associated with it. (This is one of the reasons I’ve been moving the commented out translation lists up below the translations rather than for some reason being stuck in the middle of the properties.)

(The double ciqual is one of the failing tests.)

Freso (commented):
Other than replacing grapes’ ciqual code with that of raisins (and the various issues the linter and tests have found), this PR seems fine.

taxonomies/food/ingredients.txt
```diff
@@ -32485,6 +32517,8 @@ fr: sirop d’érable canadien
 hr: kanadski javorov sirup
 hu: kanadai juharszirup
 it: sciroppo d'acero canadese
+ciqual_food_code:en: 31034
+ciqual_food_name:en: maple syrup
```
Freso: This should already be inherited via `< en:maple syrup`, no?

aleene: Good observation Maybe it should, but at the moment this does not seem to be the case. Ciqual only calls it Maple syrup, but no mention of Canada nor pure. I will rename those as proxies.

taxonomies/food/ingredients.txt
aleene: my mistake, removed

stephanegigandet (approved):
Thank you!

PR #13703, opened 2026-06-02
fix: reset to-be-exported on pro platform

stephanegigandet:
There was an issue with the mount mp8: /zfs-hdd/podata/pro_export_files,mp=/mnt/off-pro/cache/export_files on the off contained that resulted in products being marked as exported on the pro platform, while they were not actually imported on the public platform.

This fix is to reset the en:exported -> en:to-be-exported state when last_updated_t is within a specific range.

Discussion: https://openfoodfacts.slack.com/archives/CN25JTE6N/p1780392784039839


alexgarel (approved):
LGTM

PR #13712, opened 2026-06-03
feat: allow searching by product name or generic name only

### LINKED ISSUE #485: Allow to search only by name or generic name
### What

Allow to search only by name or generic name

"name" contains
"generic name" contains

### Code pointers

You can find some [code pointers here](https://github.com/openfoodfacts/openfoodfacts-server/blob/main/.github/labeler.yml). If some are missing, feel free to add them yourself in this file, or to ask them.

### Part of 
- #5492

CodeIsFire:
## What
Allow users to filter product searches by specific fields instead of searching across all fields. Added a dropdown selector in the search form to choose between:
- All fields (default) - searches product name, generic name, brands, categories, origins, and labels
- Product name only - searches only product name keywords
- Generic name only - searches only generic name keywords

## Changes Made
- Add separate keyword fields `_keywords_product_name` and `_keywords_generic_name` in `compute_keywords()`
- Support `search_type` parameter (values: all, name, generic_name) in search.pl
- Add dropdown filter in search form template to select search scope
- Add English language strings for new search type options
- Default to searching all fields when search_type is not specified

## Related issue(s)
- Fixes #485

## Testing
This feature has been tested to:
- ✅ Search across all fields (default behavior)
- ✅ Search by product name only
- ✅ Search by generic name only
- ✅ Preserve filter selection during pagination
- ✅ Maintain backward compatibility

### DIFF
--- a/cgi/search.pl
+++ b/cgi/search.pl
@@ -355,6 +355,7 @@ foreach my $series (@search_series, "nutrition_grades") {
 if ($action eq 'display') {
 
 	$template_data_ref->{search_terms} = $search_terms;
+	$template_data_ref->{search_type} = $search_type;
 
 	my $active_list = 'active';
 	my $active_map = '';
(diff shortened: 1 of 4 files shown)


john-gom:
@stephanegigandet , @alexgarel , @raphael0202 , @teolemon should we be making further changes to standard OFF searching now or be moving more to searchalicious?

If we do want to take this route then a corresponding change in off-query would be required.

raphael0202:
@john-gom I don't think we should be making further changes to the standard MongoDB OFF search, as it's going to be replaced by search-a-licious.

raphael0202:
As discussed above, I'm closing this PR. The associated issue was created in 2016, it's quite outdated...

PR #13713, opened 2026-06-04
taxonomy(nutrients): separate vit D₃ out from vit D

Freso:
“Vitamin D” specified on packaging can be one of several compounds, but we’re currently just assuming that it’s cholecalciferol (D₃), even if ergocalciferol (D₂) is also reasonably common, with other variants being more rare.

This separates out cholecalciferol/vitamin D₃ from the current `vitamin-d` nutrients entry into a new `vitamin-d3` block.

Translations are taken/adapted from Wikidata and/or current vitamin D ones.

This may (eventually, as products get updated) introduce some issues for data quality calculations that check for `vitamin-d` specifically, e.g. for “source of vitamin D” labels. I’m not sure what the best way to handle this would/will be.

Needed-for: https://world.openfoodfacts.org/product/8723300710739/vitamine-d-10-innovit
Needed-for: https://dk.openfoodfacts.org/product/5711657050021/vitamin-d3-bodylab


stephanegigandet (approved):
Thank you

PR #13730, opened 2026-06-07
taxonomy(food): move synonyms to categories

Freso:
As the file itself says:

> don't add synonyms if they are already the name of an individual
> category, in that case add the synonym to the category

Additionally, this:
- adds translations (mostly from Wikidata).
- normalises towards capitalised category name.
- does some normalising toward pluralised category names.


stephanegigandet (approved):
Thanks

PR #13731, opened 2026-06-07
taxonomy: triangle sandwiches

moon-rabbitOFF:
I renamed "Sandwich made with loaf bread" to "Triangle sandwiches". This is not an ideal name, because it's an expression that doesn't exist in English, but there is no expression for this kind of sandwiches anyway. "Sandwich made with loaf bread" is long and not really understandable. "Triangle sandwiches" is at least really explicite, even if the expression doesn't exist, people will understand what it refers to.
Club sandwich is also a close synonym, but from what I understand on wikipedia, in English it refers to a sandwich with three slices of breads, which is not the case of all "triangle sandwiches". https://en.wikipedia.org/wiki/Club_sandwich

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -100857,11 +100857,11 @@ gpc_category_name:en: Sandwiches/Filled Rolls/Wraps (Frozen)
 
 < en:Sandwiches
 en: Triangle sandwiches, Sandwich made with loaf bread, Club sandwich
-es: sándwich de pan de molde, sándwich club
+es: Sándwich de pan de molde, sándwich club
 fr: Sandwichs triangles, sandwich club, sandwich triangle, sandwich pain de mie
 hr: Sendvič od pogače
-it: tramezzino, sandwich club
-nl: club sandwich, sandwich met witbrood
+it: Tramezzino, sandwich club
+nl: Club sandwich, sandwich met witbrood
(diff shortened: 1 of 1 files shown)


Freso (approved):
Looks good! Would probably make sense to separate out the club sandwiches from the triangle sandwiches, and maybe even have them both be children of “sandwiches made with loaf bread”?

Freso:
I _think_ the failing check is because of the `synonyms` lines I’m removing in https://github.com/openfoodfacts/openfoodfacts-server/pull/13730

moon-rabbitOFF:
In French, and maybe someother languages too I guess, club sandwich is a strict synonym of triangle sandwich. We also use this word for sandwiches with only two slices of bread. E.g. https://world.openfoodfacts.org/product/3350033961562/le-club-poulet-facon-caesar-monoprix
https://world.openfoodfacts.org/product/3367651000016/the-classic-club-sandwich-tuna-and-eggs-daunat

moon-rabbitOFF:
I think we can keep club sandwich as the main entry. Other language also seem to use club sandwich 
eg in Norway https://world.openfoodfacts.org/product/7032069104552/klassisk-club-sandwich-rema-1000
in Austria https://world.openfoodfacts.org/product/9010437056526/club-sandwich-billa

PR #13733, opened 2026-06-08
chore: USDA ingredient references

aleene:
### What
Add USDA references to ingredients that already exist in the taxonomy.

I used Claude to offer me suggestions. These have all been reviewed and added by hand (so no copy/paste).


aleene:
A word on the approach. USDA has a lot of ingredients with adjectives, like raw. Usually these specific ingredients are not the taxonomy. These will be added as new entries. If plausible these will be added as USDA proxies to their parent.

aleene:
If a node is marked as (ciqual-/usda)-proxy there might be one or more descendants that are applicable as well. The proxy will be the most likely, but one of the other descendants is possible as well. By calculating the variance of each applicable proxy the best one can be selected.

stephanegigandet (approved):
Thank you!

PR #13734, opened 2026-06-08
fix: recipe-estimator config for docker

stephanegigandet:
I mistakenly put my local dev changes in the recipe-estimator docker config, this PR removes them.


lib/ProductOpener/Config2_docker.pm
```diff
@@ -158,13 +158,13 @@ $folksonomy_url = $ENV{FOLKSONOMY_URL};
 $recipe_estimator_url = $ENV{RECIPE_ESTIMATOR_URL};
 # To test a locally running recipe-estimator with Product Opener in a docker dev environment:
 # run recipe-estimator with `uvicorn recipe_estimator.main:app --reload --host 0.0.0.0`
-$recipe_estimator_url = "http://host.docker.internal:5521/api/v3/estimate_recipe";
+#$recipe_estimator_url = "http://host.docker.internal:5521/api/v3/estimate_recipe";
```
john-gom: Should we just get rid of these lines as we are already setting from the environment?

stephanegigandet: OK, I'll remove them.

john-gom (approved):

hangy (approved):

PR #13744, opened 2026-06-10
chore: usda match

aleene:
### What
Added USDA links for entries that match.

Used Claude Sonnet 4.6 to do the matches.


aleene:
why do I get the conflict message. I do not see anything in the code.

PR #13747, opened 2026-06-10
fix: Updating preferences on staging is wiping the user name in keycloak

### LINKED ISSUE #13743: Updating Preferences on Staging is wiping the user name in Keycloak
### What
- If preferences are saved the user name gets wiped

1. Log in to staging
2. Click on the Preferences menu item
3. Save a change
4. Observe that the profile icon no longer contains your name

### Expected behavior
- Editing preferences should not update Keycloak

john-gom:
### What
Only update the user name, language and country in Keycloak Level 5 if the form has been submitted via an API (e.g. the DART SDK). This is detected by the presence of the name field in the payload.

None

- Fixes: https://github.com/openfoodfacts/openfoodfacts-server/issues/13743

### DIFF
--- a/lib/ProductOpener/Users.pm
+++ b/lib/ProductOpener/Users.pm
@@ -483,12 +483,18 @@ sub check_user_form ($request_ref, $type, $user_ref, $errors_ref) {
 	# Assigning 'userid' to 0 -- if userid is not defined
 	$user_ref->{userid} = remove_tags_and_quote(single_param('userid'));
 
-	# Check for spam
 	my $is_spam = undef;
-	# e.g. name with "Lydia want to meet you! Click here:" + an url or + a .com / .ru
-	if (is_suspicious_name($user_ref->{name})) {
-		$is_spam = 1;
+	# Allow for sending the 'name' & 'email' as a form parameter instead of a HTTP header, as web based apps may not be able to change the header sent by the browser
+	if (defined single_param('name')) {
+		$user_ref->{name} = remove_tags_and_quote(decode utf8 => single_param('name'));
+
+		# Check for spam
+		# e.g. name with "Lydia want to meet you! Click here:" + an url or + a .com / .ru
+		if (is_suspicious_name($user_ref->{name})) {
+			$is_spam = 1;
+		}
 	}
+
 	# check for spam, that may have filled the honeypot faxnumber field
 	my $faxnumber = single_param('faxnumber');
 	if ((defined $faxnumber) and ($faxnumber ne "")) {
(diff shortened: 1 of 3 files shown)


lib/ProductOpener/Users.pm
```diff
@@ -532,10 +535,13 @@ sub check_user_form ($request_ref, $type, $user_ref, $errors_ref) {
...
-	$user_ref->{preferred_language} = remove_tags_and_quote(single_param('preferred_language')) || $request_ref->{lc};
-	$user_ref->{country} = remove_tags_and_quote(single_param('country')) || $request_ref->{country};
-
+	# Still allow these to be sent for legacy API support, but don't update if supplied by the preferences form
+	# Which we deduce by the absence of a name field
+	if (defined single_param('name')) {
+		$user_ref->{preferred_language}
+			= remove_tags_and_quote(single_param('preferred_language')) || $request_ref->{lc};
+		$user_ref->{country} = remove_tags_and_quote(single_param('country')) || $request_ref->{country};
+	}
```
Freso, line 544: Maybe I don’t understand what the legacy API support is for, but why check on `name` rather than `preferred_language`/`country`?

Freso (approved):
I have a question about one part, but the logic seems fine.

lib/ProductOpener/Users.pm
hangy: The legacy API support is for eg. the Flutter mobile app, which uses one of the cgi scripts to update the user profile. It has it's own user preferences dialogs, and we don't want to break it before updating the app(s).

With Keycloak integration level 5, the [user parameters](https://world.openfoodfacts.org/cgi/user.pl?type=edit) will only allow editing preferred language, and country, so the idea is to match on the fields that will only be sent by use of the legacy API.

That being said, don't like the idea of matching on that field, only. It'll work, but the implicit behaviour is not super obvious. What do you think about having an hidden form field that _explicitly_ opts out from editing `name` & `email`?

hangy (approved):

lib/ProductOpener/Users.pm
john-gom: This is mainly a legacy feature to cope with the mobile app. Longer term the app should use a web view directly into Keycloak for user registration and this code will be completely removed.

I couldn't use the presence of language or country as the Dart SDK doesn't use these fields and instead uses the domain prefix to indicate both. so name was the easiest field to use as it is mandatory

john-gom: @hangy to clarify, at level 5 the user preferences just does producer flags and display preferences. Language and country can only be changed in Account Parameters (Keycloak UI)

PR #13750, opened 2026-06-11
fix: log request stats in HTML pages comment

stephanegigandet:
This is to put the time taken by requesting mongodb or off-query in a comment of HTML pages.


tests/integration/expected_test_results/api_v2_product_read/get-auth-bad-user-password.html
```diff
@@ -692,4 +692,14 @@ <h3 class="title-5 text-medium">Discover the project</h3>
 
 
 
+<!-- request.stats : 
+
+	request_duration => 0.0400689319999401
+
+	pid => 36
+
+	request_t => 1781180612
```
hangy: The values probably need to be removed as they will vary on each run?

stephanegigandet: yes, I'll remove them in the tests

teolemon:
@stephanegigandet by essence, the tests on this PR can never validate since the timing is variable.
How should we go about it ?
Have the test ignore commented out HTML ?

PR #13751, opened 2026-06-11
fix: use JSON allow_nonref

stephanegigandet:
Should fix this error reported on Slack: https://openfoodfacts.slack.com/archives/C02KVRT2C/p1781092853221419

```
I'm having trouble editing some products on the website, for instance 0744096655408, 3660992012756 and 0731509930092. When I click Save I get the following error:

-----
Software error:

encountered object '1', but neither allow_blessed, convert_blessed nor allow_tags settings are enabled (or TO_JSON/FREEZE method missing) at /srv/off/lib/ProductOpener/APIProductServices.pm line 161.

For help, please send mail to the webmaster ([contact@openfoodfacts.org](mailto:contact@openfoodfacts.org)), giving this error message and the time and date of the error.
```

### DIFF
--- a/lib/ProductOpener/APIProductServices.pm
+++ b/lib/ProductOpener/APIProductServices.pm
@@ -98,6 +98,7 @@ use ProductOpener::EnvironmentalImpact;
 use ProductOpener::HTTP qw/create_user_agent/;
 use ProductOpener::Store qw/$json_for_objects/;
 
+use Cpanel::JSON::XS;
 use Storable qw/dclone/;
 use Encode;
 use Types::Serialiser;
--- a/lib/ProductOpener/APIProductServices.pm
+++ b/lib/ProductOpener/APIProductServices.pm
@@ -98,7 +98,6 @@ use ProductOpener::EnvironmentalImpact;
 use ProductOpener::HTTP qw/create_user_agent/;
 use ProductOpener::Store qw/$json_for_objects/;
 
-use Cpanel::JSON::XS;
 use Storable qw/dclone/;
 use Encode;
 use Types::Serialiser;
--- a/lib/ProductOpener/APIProductServices.pm
+++ b/lib/ProductOpener/APIProductServices.pm
@@ -97,10 +97,12 @@ use ProductOpener::API qw/add_error customize_response_for_product init_api_resp
 use ProductOpener::EnvironmentalImpact;
 use ProductOpener::HTTP qw/create_user_agent/;
 
-use JSON qw(decode_json encode_json);
+use JSON::MaybeXS;
 use Storable qw/dclone/;
 use Encode;
 use Types::Serialiser;
+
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/APIProductServices.pm
```diff
@@ -97,11 +97,13 @@ use ProductOpener::API qw/add_error customize_response_for_product init_api_resp
 use ProductOpener::EnvironmentalImpact;
 use ProductOpener::HTTP qw/create_user_agent/;
 
-use JSON qw(decode_json encode_json);
+use JSON::MaybeXS;
 use Storable qw/dclone/;
 use Encode;
 use Types::Serialiser;
 
+my $json_utf8 = JSON::MaybeXS->new->convert_blessed->utf8(1)->allow_nonref->canonical;
```
john-gom: Should we maybe share code with Store to keep things in sync? https://github.com/openfoodfacts/openfoodfacts-server/blob/57e9cb5f22f2c1ccb99b346385f1876c2057b134/lib/ProductOpener/Store.pm#L76

lib/ProductOpener/APIProductServices.pm
```diff
@@ -96,8 +96,9 @@ use ProductOpener::Products qw/:all/;
 use ProductOpener::API qw/add_error customize_response_for_product init_api_response/;
 use ProductOpener::EnvironmentalImpact;
 use ProductOpener::HTTP qw/create_user_agent/;
+use ProductOpener::Store qw/$json_for_objects/;
 
-use JSON qw(decode_json encode_json);
+use Cpanel::JSON::XS;
```
john-gom: Is this needed now?

john-gom (approved):

lib/ProductOpener/APIProductServices.pm
stephanegigandet: I thought it was but it looks like it's working without it, so removing it, thanks!

stephanegigandet: In fact we do need it:

```
(  DIAG  )  job  4    Response status line: 500 Internal Server Error
(  DIAG  )  job  4    external-sources-en - GET http://world.openfoodfacts.localhost/cgi/display.pl?/api/v3/external_sources request got a response that is not valid JSON: malformed JSON string, neither tag, array, object, number, string or atom, at character offset 0 (before "<h1>Software error:<...") at /opt/product-opener/lib/ProductOpener/APITest.pm line 934.
(  DIAG  )  job  4    Response content: <h1>Software error:</h1>
(  DIAG  )  job  4    <pre>Undefined subroutine &amp;ProductOpener::APIProductServices::decode_json called at /opt/product-opener/lib/ProductOpener/APIProductServices.pm line 438.
(  DIAG  )  job  4    </pre>
```

PR #13765, opened 2026-06-16
fix: Add last_updated_t index as this is needed by off_query

john-gom:
### What
During off_query's incremental import the data is searched using the last_updated_t field which is not currently indexed, resulting in a slow query and excessive load on the MongoDB server.

The teams index was dropped to make space for this

None


stephanegigandet (approved):
Thanks

PR #13784, opened 2026-06-18
fix: add --query-codes-from-file to update_all_products.pl

stephanegigandet:
This is needed to selectively update products needed for the FOODTURE project, and it could be useful for other users too.


alexgarel (approved):
Great

PR #13791, opened 2026-06-19
fix: Pagination issues

john-gom:
### What
Perl errors were being logged because teh test for page > 10 for unauthenticated users was occurring before request_ref->{page} had been defined.

Also found an issue where the sort by was being appended twice to links

None

### Fixes
 - #9392

### DIFF
--- a/tests/integration/expected_test_results/web_html/fr-search-results-cached.html
+++ b/tests/integration/expected_test_results/web_html/fr-search-results-cached.html
@@ -10,7 +10,7 @@
 	<meta property="fb:app_id" content="219331381518041">
     <meta property="og:type" content="food">
     <meta property="og:title" content="Résultats de la recherche - France">
-    <meta property="og:url" content="//fr.openfoodfacts.localhost/cgi/search.pl?action=process&search_terms=tarte&sort_by=unique_scans_n&page_size=50">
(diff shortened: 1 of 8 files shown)


tests/integration/expected_test_results/web_html/fr-search-results-cached.html
```diff
@@ -10,7 +10,7 @@
 	<meta property="fb:app_id" content="219331381518041">
     <meta property="og:type" content="food">
     <meta property="og:title" content="Résultats de la recherche - France">
-    <meta property="og:url" content="//fr.openfoodfacts.localhost/cgi/search.pl?action=process&search_terms=tarte&sort_by=unique_scans_n&page_size=50">
+    <meta property="og:url" content="//fr.openfoodfacts.localhost/cgi/search.pl?action=process&search_terms=tarte&page_size=50">
```
stephanegigandet, line 13: In the tests there are many instances where the sort_by has been removed, even when there was only one, is that wanted?

john-gom: This is where the sort order is the default. I could get it to persist the default in follow up links but that would make it more complicated as we'd need to strip it out again in the sort links

stephanegigandet (approved):
Thank you!

PR #13828, opened 2026-06-26
chore: USDA part 8

aleene:
### What
More USDA ingredients added


teolemon (approved):
LGTM, thanks a lot @aleene 👍

PR #13833, opened 2026-06-26
chore: FoodEx2 entries - part 1

aleene:
### What
Adding links between ingredients and equivalent FoodEx2 entries.

Based on https://github.com/openefsa/efsa-catalogues/raw/master/ecf_catalogues/FOODEX.ecf


aleene:
@stephanegigandet have a look at this possible FoodEx2 integration

teolemon (commented):
- Are you matching by hand ? 
- Can we already pivot using CIQUAL, USDA or Wikidata + a deterministic script like the one for wikidata entries ?
- Alternatively Could LLM-assisted matches be relevant ?

stephanegigandet:
> Based on https://github.com/openefsa/efsa-catalogues/raw/master/ecf_catalogues/FOODEX.ecf

@aleene I think those are old FOODEX codes, they don't look like FOODEX2 codes. e.g. only cereals products should start with A.

aleene:
> > Based on https://github.com/openefsa/efsa-catalogues/raw/master/ecf_catalogues/FOODEX.ecf
> 
> @aleene I think those are old FOODEX codes, they don't look like FOODEX2 codes. e.g. only cereals products should start with A.

Yes if it is eaten by humans. If it is used in the food industry it starts with a G. Thus G.1 (cereal grains) is animal fodder.

aleene:
The G.* ones seem interesting as they are used by the industry in recipes. Unfortunately we do not have the associated nutritional values.

alexgarel (commented):
Hi @aleene is it still draft or ready to merge ?

PR #13835, opened 2026-06-26
taxonomy(food): chutney(+relish) edits

Freso:
- Imports translations from Wikidata.
- Adds Danish/Swedish/Norwegian translations.
- Adds `wikidata` properties.

Needed-for: https://se.openfoodfacts.org/product/7340083433603/mango-chutney-sweet-garant

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -56027,12 +56027,74 @@ zh: 咖啡甜品, 咖啡甜点
 < en:Plant-based spreads
 en: Relish
 xx: Relish
+ar: المقبّلات
+da: Relisher
+eo: Reliŝoj
+fa: گزک (چاشنی)
+fi: Kurkkusalaatti
+hy: Ռելիշ
+ja: レリッシュ
+ki: Gĩtoero
+ko: 렐리시
+pl: Relisz
+ru: Релиш
+sl: Reliš
+sw: Kitoweo
+uk: Реліш
+zh: 開味小菜
 intake24_category_code:en: RLSH
+wikidata:en: Q766777
 
 < en:Plant-based spreads
 en: Chutneys
 xx: Chutneys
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -56047,41 +56109,56 @@ gpc_category_description:en: Definition: Includes any products that can be descr
 gpc_category_name:en: Chutneys/Relishes (Perishable)
 
 < en:Chutneys
+< en:Fruits based foods
 en: Fruit chutneys
```
moon-rabbitOFF, line 56112: I don't know if we should put "fruit chutneys" in "fruit based foods", when it can contain really little fruit. This one as dried mango 15% (= mango + sugar) https://world.openfoodfacts.org/product/5020332111309/shaws-yorkshire-chunky-mango-chutney

aleene: At what percentage will it be fruit-based?

moon-rabbitOFF: We can discuss about that 😃 Two threshold that would make sense to me are either:
* More than half
* The first ingredient of the list whatever the percentage is
For a product with multiple fruits, the sum of fruit should then be either > 50% or on top of the list. 
What do you think?

Freso: I’d say it is still fruits _based_, given that fruit (dried mango) is the primary actual ingredient. If you think that particular chutney isn’t fruit-based, then perhaps it shouldn’t be in fruit chutneys at all.

Anyway, I’ll take this out of this PR and make a new one with just this change where we can discuss further.

Freso: https://github.com/openfoodfacts/openfoodfacts-server/pull/13846

moon-rabbitOFF (approved):

PR #13839, opened 2026-06-27
taxonomy: Brittany-style biscuits

moon-rabbitOFF:
* Renamed "Shortbread cookies from Brittany" --> "Brittany-style galette biscuits". Those biscuits are not always made in Brittany, saying they are "from Brittany" is then incorrect. Moreover, other biscuits originated in Brittany, such as the palets. I replaced shortbread cookies by biscuits, because this first expression is already made to distinguish "real" shortbreads from Scottland from "shortbread-style cookies". "Galette shortbread cookies" would mean "cookies in the style of shortbread, in the style of palets", which is redundant.
* Translated "fr:palets" as "en:Brittany-style palet biscuits"

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -35270,10 +35270,16 @@ ciqual_food_name:fr: Sablé pâtissier
 wikidata:en: Q2915670
 
 < en:Shortbread cookies
-en: Shortbread cookies from Brittany
+en: Shortbread
+xx: Shortbread
+wikidata:en: Q652695
+
+< en:Shortbread cookies
+en: Brittany-style galette biscuits, Shortbread cookies from Brittany
 fr: Galettes bretonnes au beurre, galettes bretonnes
 
 < en:Shortbread cookies
+en: Brittany-style palet biscuits
 fr: Palets, Palet, Palets sablés, Palet sablé, Palet breton, Palets bretons
 
 < en:Shortbread cookies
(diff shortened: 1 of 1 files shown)


Freso (approved):

taxonomies/food/categories.txt
```diff
@@ -35352,11 +35358,8 @@ ciqual_food_name:en: Biscuit shortbread, with fruits
 ciqual_food_name:fr: Biscuit sec, sablé, galette ou palet, aux fruits
 
 < en:Shortbread cookies
-fr: Sablés au beurre
-
-< en:Biscuits
-en: Shortbread biscuit with butter
-fr: Biscuit sec au beurre
+en: Plain butter shortbreads, Shortbread biscuit with butter
+fr: Sablés nature au beurre, Sablés au beurre
```
Freso, line 35362:
```suggestion
fr: Sablés nature au beurre, Sablés au beurre, Biscuit sec au beurre
```
maybe? (might not make sense, idk enough french to determine, just in case anyone has links lying around pointing to `fr:Biscuit sec au beurre` would be nice if we didn’t linkrot them if we can help it)

stephanegigandet (approved):
Thank you!

taxonomies/food/categories.txt
moon-rabbitOFF: I prefer to linkrote a few products than putting a synonym that will add more confusion. It is an open door for contributors to add any biscuits with butter in this category 😄 I will correct the wrong links manually

PR #13860, opened 2026-06-30
fix: "Explore by" button to point directly to /facets URLs

### LINKED ISSUE #13779: The "Explore by" button should link directly to /facets/* URLs to avoid unecessary queries
### What
- The "Explore by" button should link directly to /facets/* URLs to avoid unecessary queries

oxidn:
This PR fixes the "Explore by" links on the homepage so they point directly to `/facets/` URLs, avoiding  unnecessary query.

## Changes

- Generate `/facets/countries`, `/facets/categories`, etc. directly.
- Preserve the producer platform prefix using `get_owner_pretty_path($Owner_id)`.

## Testing

- Verified the links now point directly to `/facets/` URLs.
- Confirmed in the browser's Network tab that the extra redirect no longer occurs.

Fixes #13779

### DIFF
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -5648,7 +5648,7 @@ sub search_and_display_products ($request_ref, $query_ref, $sort_by, $limit, $pa
 
 					push @{$template_data_ref->{current_drilldown_fields}},
 						{
-						current_link => get_owner_pretty_path($Owner_id) . "/facets",
+						current_link => get_owner_pretty_path($Owner_id) . $request_ref->{current_link},
 						tag_type_plural => $tag_type_plural{$newtagtype}{$lc},
 						nofollow => $nofollow,
 						tagtype => $newtagtype,
--- a/lib/ProductOpener/Display.pm
+++ b/lib/ProductOpener/Display.pm
@@ -5648,7 +5648,7 @@ sub search_and_display_products ($request_ref, $query_ref, $sort_by, $limit, $pa
 
 					push @{$template_data_ref->{current_drilldown_fields}},
 						{
-						current_link => get_owner_pretty_path($Owner_id) . $request_ref->{current_link},
+						current_link => get_owner_pretty_path($Owner_id) . "/facets",
 						tag_type_plural => $tag_type_plural{$newtagtype}{$lc},
 						nofollow => $nofollow,
 						tagtype => $newtagtype,
--- a/lib/ProductOpener/Routing.pm
+++ b/lib/ProductOpener/Routing.pm
@@ -250,7 +250,7 @@ sub index_route($request_ref) {
 	# Root, ex: https://world.openfoodfacts.org/
 	$request_ref->{text} = 'index';
 	$request_ref->{current_link} = '';
-	
+
 	# Root + page number, ex: https://world.openfoodfacts.org/2
 	if (exists $request_ref->{param}{page}) {
 		$request_ref->{page} = $request_ref->{param}{page};
--- a/lib/ProductOpener/Routing.pm
+++ b/lib/ProductOpener/Routing.pm
@@ -249,8 +249,8 @@ sub index_route($request_ref) {
 
 	# Root, ex: https://world.openfoodfacts.org/
 	$request_ref->{text} = 'index';
-	$request_ref->{current_link} = '';
-
+	$request_ref->{current_link} = '/facets';
+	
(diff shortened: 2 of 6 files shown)


lib/ProductOpener/Display.pm
```diff
@@ -5648,7 +5648,7 @@ sub search_and_display_products ($request_ref, $query_ref, $sort_by, $limit, $pa
 
 					push @{$template_data_ref->{current_drilldown_fields}},
 						{
-						current_link => get_owner_pretty_path($Owner_id) . $request_ref->{current_link},
+						current_link => get_owner_pretty_path($Owner_id) . "/facets",
```
stephanegigandet: The right solution is to fix the generation of $request_ref->{current_link} to add "/facets" at the beginning for facets URLs.
Otherwise this breaks second level facets where current link is  something like /facets/categories/cakes and where we want to link to /facets/categories/cakes/labels

image

stephanegigandet (changes requested):
Need to update $request_ref->{current_link} instead

lib/ProductOpener/Routing.pm
```diff
@@ -249,8 +249,8 @@ sub index_route($request_ref) {
 
 	# Root, ex: https://world.openfoodfacts.org/
 	$request_ref->{text} = 'index';
-	$request_ref->{current_link} = '';
-
+	$request_ref->{current_link} = '/facets';
+
```
Freso:
```suggestion

```

lib/ProductOpener/Routing.pm
Freso, line 248: Is this the right place to change this? It feels like this might change for things other than facets URLs, but I’m not sure. Have you tested this locally making sure non-`facets` pages continue to work correctly?

oxidn: You were right about the `index_route` approach. I tested it locally and while it fixes the second-level facet navigation, it also changes the homepage "Sort by" links to `/facets?sort_by=...`, so it introduces a regression. I'll look for a more targeted place where `current_link` is generated for facet URLs only.

oxidn: I traced the request flow to understand where `current_link` is used. The homepage starts with an empty `current_link` and no `$request_ref->{tags}`, so the existing tag canonicalization logic isn't directly reusable there. I also found that `current_link` is used not only for the "Explore by" links, but also for the homepage sorting links (e.g. Most scanned products, Products with the best Nutri-Score, etc.). Initializing `/facets` earlier breaks their links.

Because of that, my solution is to handle only the case where `current_link` is empty immediately before generating the "Explore by" dropdown links. By that point, the sorting links and the other search-related links have already been generated using the original `current_link`, so they are unaffected.

I tested this locally on the homepage, facet pages, and second-level facet pages, and it appears to work correctly in all of those cases.

Does this approach sound reasonable or would you prefer a different approach?

Freso: @stephanegigandet

stephanegigandet:
Thinking back a bit more about this, I think my earlier suggestion to add /facets to the current_link is the wrong one. (sorry!)

The current_link is intended to contain the URL of the page currently displayed, we use it for different things, we should not change it to something that is not the URL of the current page. If we want to change the links of the explore by dropdown, then the right place, is probably in the template that generate the links: web/common/includes/list_of_products.tt.html

If the current link is empty, we add /facets at the beginning of the links, otherwise it should already contain something like /facets/categories/something

oxidn:
> Thinking back a bit more about this, I think my earlier suggestion to add /facets to the current_link is the wrong one. (sorry!)
> 
> The current_link is intended to contain the URL of the page currently displayed, we use it for different things, we should not change it to something that is not the URL of the current page. If we want to change the links of the explore by dropdown, then the right place, is probably in the template that generate the links: web/common/includes/list_of_products.tt.html
> 
> If the current link is empty, we add /facets at the beginning of the links, otherwise it should already contain something like /facets/categories/something

Thanks for the clarification. I updated the solution based on your suggestion and tested it locally. The homepage sorting links are unaffected, and the "Explore by" links now work correctly without unnecessary redirects. Facet pages, and second-level facet pages also work correctly.

stephanegigandet (approved):
Perfect, thank you very much @oxidn

stephanegigandet:
@oxidn could you run "make update_tests_results" and push the changes?

oxidn:
@stephanegigandet

teolemon:
thanks @oxidn merging

PR #13889, opened 2026-07-02
chore: Add USDA ingredients - part 2-2

aleene:
### What
The second part of part 2


taxonomies/food/ingredients.txt
```diff
@@ -41987,11 +41987,18 @@ ciqual_food_code:en: 13082
...
 ecobalyse:en: 299941e9-577c-4618-968c-6901ffa99a8d
+usda_ndb_proxy_code:en: 9433
+usda_ndb_proxy_name:en: Clementines, raw
 wikidata:en: Q460517
 wikipedia:en: https://en.wikipedia.org/wiki/Clementine
 # openfoodfacts:https://world.openfoodfacts.org/ingredient/fr:clémentines
 # 81 products @2018-10-01
 
+< en:clementine
+en: raw clementine
```
stephanegigandet: I think we should avoid creating "raw XYZ" entries, and by default consider that all ingredients are raw.

And possibly we could have properties like usda_ndb_code_processing_cooked:en: and sda_ndb_code_processing_frozen:en:  if we have multiple usda or ciqual value

aleene: The idea is that we will be able to compare recipes with raw/frozen/cooked/with skin/dried ingredient, so we can determine which one fits best. I started adding the raw ones, as there were small differences between ciqual/usda/ifct, which required me to correct each time.

aleene: OK, I will slowly revert

teolemon:

aleene:
Documentation: https://wiki.openfoodfacts.org/Recipe/Ingredient_taxonomy/Ingredient_linking

alexgarel (approved):
LGTM

PR #13899, opened 2026-07-03
chore: repairing sugar hierarchy

aleene:
### What
The sugar hierarchy is repaired. `en:added-sugar` is parent of all sugars.


stephanegigandet (approved):
Thanks

PR #13916, opened 2026-07-04
feat: add salt/sodium consistency data quality facet

### LINKED ISSUE #13531: Facet for salt:sodium ratio being out of alignment
### Problem

With sodium no longer being calculated from salt and salt from sodium at the point of user input, there’s a chance that users might update one without updating (or removing) the other, causing the values to drift apart.

### Proposed solution

Adding a data quality facet to find products that have salt and sodium set but where the values don’t match up.

### Additional context
- Slack discussion: https://openfoodfacts.slack.com/archives/C031V3X7A/p1777469589708269

### Part of
-

oxidn:
## What

Add a new data quality tag/facet that detects when the declared salt value is inconsistent with the declared sodium value.

When both nutrients are present for nutrition data per 100 g or 100 ml, the check compares the declared salt value with the salt value calculated from sodium (salt = sodium × 2.5). If the difference is greater than 0.1 g, the following tag is added:

`en:<set_id>-salt-does-not-match-sodium`

## Why

Detecting inconsistencies will help identify potential data entry errors and improves nutrition data quality.

## Changes

- Added the `salt-does-not-match-sodium` data quality tag.
- Added the corresponding consistency check in `DataQualityFood.pm`.
- Added unit tests.
- Updated expected test results where the new tag is expected.

## Notes

- The check only runs for nutrition values per 100 g or 100 ml.
- A tolerance of 0.1 g is used to account for rounding on nutrition labels.

Fixes: #13531

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -1160,33 +1160,33 @@ sub check_specific_nutrients_for_input_set ($product_ref, $nutrition_set_ref, $s
 
 	# Too small salt value? (e.g. g entered in mg)
 	# warning for salt < 0.1 was removed because it was leading to too much false positives (see #9346)
-my $per = deep_get($nutrition_set_ref, "per");
-if ((defined $per) and (($per eq "100g") or ($per eq "100ml"))) {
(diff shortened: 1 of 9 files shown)


stephanegigandet (approved):
Looks good, thank you

PR #13931, opened 2026-07-06
taxonomy: Banh dau xanh

moon-rabbitOFF:


taxonomies/food/categories.txt
```diff
@@ -39785,6 +39809,12 @@ food_groups:en: en:sweets
 intake24_category_code:en: CONF
 pnns_group_2:en: Sweets
 
+< en: Confectioneries
+xx: Banh dau xanh, Bánh đậu xanh
```
Freso: Any reason to not do this?
```suggestion
xx: Bánh đậu xanh, Banh dau xanh
```

Seems like if we use the Vietnamese name for the `xx` entry, we might as well use the actually Vietnamese name, and not some simplification to make it easier for English speakers. The simplified version can still be used for input, but, IMHO, we should display the name written properly.

Freso (approved):

taxonomies/food/categories.txt
moon-rabbitOFF: I'm fine with it, I changed it

PR #13932, opened 2026-07-06
taxonomy: Dried sweet potatoes

moon-rabbitOFF:
* Created a category for dried sweet potatoes
* Removed sweet potato categories from vegetable categories, they are not vegetables.

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -110182,7 +110182,6 @@ wikidata:en: Q1072192
 
 < en: Plant-based spreads
 < en: Salted spreads
-< en: Sweet potato dishes
 en: Sweet potato spreads
 de: Süßkartoffel-Aufstriche, Süßkartoffel-Aufstrich
 fr: Tartinades de patates douces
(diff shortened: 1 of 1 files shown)


Freso:
> Removed sweet potato categories from vegetable categories, they are not vegetables.

What makes you say that?

Wikidata disagrees:
billede

And the [Wikipedia article](https://en.wikipedia.org/wiki/Sweet_potato#Culinary) refers to sweet potato as a vegetable in multiple places.

moon-rabbitOFF:
Similarly to potatoes, they have to much starch to be considered a vegetable. The WHO excludes them from their list of vegetables, in their recommendation of eating 400 g of fruits and vegetables per day https://www.who.int/docs/default-source/healthy-diet/who-nmh-nhd-18-12-eng.pdf
Many countries are following those guidelines and are not considering them as vegetables. They also don't count as fruits and vegetables for the Nutriscore. I think it is then misleading for the consumers to classify them as vegetables.
We could create a category "starchy roots", which would allow them to be better classified without confusing the users.

The role of Wikipedia is to document informations, almost as many as possible. Our role is a bit different I think. We could display the UK "traffic light" system and the Australian "Health star rating" along with the Nutriscore. Similarly we could display the "Planet score" along to the "Green score". But we decided in each case to keep a single indicator and advocate for the one we think is the most relevant. For categories also, I think we should aim to keep one coherant system which match the indicators we are displaying.

PR #13936, opened 2026-07-07
taxonomy: Add translations for product type errors in categories

teolemon:
taxonomy: Add translations for product type errors in categories
(reapply old PR)


stephanegigandet (approved):
Thanks

PR #13937, opened 2026-07-07
build(docker): Use Node.js 22.x floating version

hangy:
We currently don't update the image version regularly, and don't pin any other base image versions in this repository. Simply using the floating 22.x major tag makes more sense and is more consistent here, for now.


alexgarel (approved):
Perfect

PR #13938, opened 2026-07-07
chore: Replace remaining `smartmatch` usages

### LINKED ISSUE #13658: Replace Perl "smartmatch" usages
### Task

The experimental smartmatch feature from Perl 5.10 has been [deprecated since 2022](https://www.nntp.perl.org/group/perl.perl5.porters/2022/12/msg265180.html) and will be removed in Perl 5.42.

Find all places where [smartmatch is being used](https://github.com/search?q=repo%3Aopenfoodfacts%2Fopenfoodfacts-server%20smartmatch&type=code) and replace the usage accordingly. Note that some places might have a performance impact, so a solution needs to be chosen carefully. (Be wary of brute-forcing the solution.)

hangy:
### What

Replace smartmatch due to it being dropped from future Perl versions.

Closes #13658

### DIFF
--- a/lib/ProductOpener/CRM.pm
+++ b/lib/ProductOpener/CRM.pm
@@ -47,7 +47,6 @@ package ProductOpener::CRM;
 
 use ProductOpener::PerlStandards;
 use Exporter qw< import >;
-use experimental 'smartmatch';
 
 BEGIN {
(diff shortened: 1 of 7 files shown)


alexgarel (approved):
Thanks !

PR #14015, opened 2026-07-16
taxonomy(food): fair trade edits

Freso:
Imports translations from Wikidata and moves some around, with the assumption that label:en:fair-trade is a generic term used for organisations/certifications other than only Fairtrade International/Max Havelaar.

E.g., `nl`’s name in Wikidata matched the first term in `description:nl:`, so I opted to use that as the primary Dutch synonym.

Also cleans up “Fair trade” and “Fairtrade” used as non-primary synonyms, since they’re both (now) in the `xx` list.

Needed-for: https://se.openfoodfacts.org/product/7311043023752/ekologisk-kokosolja-kallpressad-virgin-garant

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -1179,17 +1179,60 @@ label_categories:en: en:Social responsibility
 
 en: Fair trade
 xx: Fair trade
+af: regverdige handel
+ar: تجارة عادلة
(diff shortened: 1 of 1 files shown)


teolemon (approved):
Really cool and impactful PR @Freso 👍

Issue #3086, opened 2020-03-23
Handling geo-adjectives in ingredient lists


teolemon:
`cuisse de porc italienne -> fr:cuisse de porc italienne - percent_min: 50 - percent_max: 100`

To multiply by all language having feminine and plural forms of adjectives

fr:singular:masculine:français
fr:singular:feminine:française
fr:plural:masculine:français
fr:plural:feminine:françaises

italien
italienne
italiens
italiennes

espagnole
allemande
polonaise
…

stephanegigandet:
We can do something like that, but it needs to be done one by one for each adjective in each language:

1. First do queries like https://fr.openfoodfacts.org/ingredients?filter=italien https://fr.openfoodfacts.org/ingredients?filter=suedois etc. 
2. Find false positives ("pain suédois", "pain allemand", "french toast" etc.) and add them to the taxonomy 
3. Add the adjective.

CharlesNepote:
Another example: "Pitted california prunes":
https://world.openfoodfacts.org/product/0802763331700/ones-amazin-prunes-sunsweet

teolemon:
I have generated a multilingual list of adjectives for countries using Wikidata: w.wiki/QjV

aleene:
And add them to the processes/adjectives taxonomy?

Ban3:
> And add them to the processes/adjectives taxonomy?

I'd suggest a new demonym taxonomy.

teolemon:
Given the volatility of Wikidata (more demonyms are added overtime, in more languages), I would rather have a conversion pipe rather than a fixed taxonomy.
French regions: https://w.wiki/RyT
French communes: https://w.wiki/RyX

PR #14027, opened 2026-07-19
taxonomy(food): add Swedish strawberry ingredient

Freso:
Needed-for: https://se.openfoodfacts.org/product/7311043022151/svenska-jordgubbar-garant

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -49357,6 +49357,12 @@ wikidata:en: Q104219130
 wikipedia:en: https://en.wikipedia.org/wiki/Fragaria_vesca
 # ingredient/fr:fraises-des-bois has 23 products in 4 languages @2020-12-23
 
+< en: strawberries
+en: Swedish strawberry, Swedish strawberries
+da: svenskt jordbær, svenske jordbær
+sv: svensk jordgubbe, svenska jordgubbar
+origin:en: en:sweden
+
 ###################################################################################################
 #
 #                                Sea-Buckthorn


moon-rabbitOFF:
I thought that countries were handled in another file and recognized in ingredient list? I don't think adding every country to every ingredient would be the easiest way to deal with it

Freso:
They’re not handled if they’re used as an adjective (cf. “Sweden” vs. “Swedish”). If the ingredient was “strawberries (Sweden)” or “strawberries from Sweden” the parser would pick up on it just fine, but “Swedish strawberries” is just an unknown ingredient to it.

I agree that this is not the best way to deal with it, but it’s what we currently do. Hopefully at some point we’ll figure out a better way to handle it, and we’ll be able to drop a bunch of ingredients.

taxonomies/food/ingredients.txt
```diff
@@ -49303,6 +49303,12 @@ wikidata:en: Q104219130
 wikipedia:en: https://en.wikipedia.org/wiki/Fragaria_vesca
 # ingredient/fr:fraises-des-bois has 23 products in 4 languages @2020-12-23
 
+< en: strawberries
+en: Swedish strawberry
```
stephanegigandet: It's useful to add the plural as it's not done automatically for y / ies

```suggestion
en: Swedish strawberry, Swedish strawberries
```

stephanegigandet:
Related discussion: https://github.com/openfoodfacts/openfoodfacts-server/issues/3086

Freso:
Closing this in favour of https://github.com/openfoodfacts/openfoodfacts-server/pull/14047

PR #14055, opened 2026-07-24
fix: round ugly nutritional data on web (#14035)

### LINKED ISSUE #14035: Ugly nutritional data on web
### What
When a product has no per 100g information, it will be calculated from the per serving data. This results in numbers with 10 digits after the comma.

Image

1. Go to  https://world.openfoodfacts.org/product/7891000106921/sorvete-sensacao-nestle-pote-1-5l
2. Check out nutritional data

### Expected behavior
Round the info to single digits

### Type of device
Web-interface

All US type products

pranavchoudhary-tech:
Fixes #14035

When a product has no per-100g information, the system calculates it from the per-serving data. Due to floating-point calculations, this often results in numbers with many decimal places (e.g., 16.6666666667), creating "ugly" nutritional data on the web.

This PR addresses the issue by safely rounding formatted calculated numeric values to a single decimal place in `lib/ProductOpener/Display.pm` before they are rendered in the HTML table.

By fixing this at the presentation layer, we prevent data loss in the raw API/JSON while fixing the visual bug reported by the user.

## Summary by CodeRabbit

* **Improvements**
  * Nutrition facts now display calculated values with concise, rounded decimal precision instead of lengthy fractional values.
  * Rounding is applied consistently across product pages, knowledge panels, and API responses, including energy, macronutrients, salt, sodium, and fruit/vegetable/legume percentages.
  * Full-precision numeric content remains available where applicable for accessibility and structured data.
  * Very small values retain meaningful precision while unnecessary trailing zeros are removed.


Naruyoko:
Does your code handle tiny values well, such as 0.0012?

pranavchoudhary-tech:
@Naruyoko Great catch! You are totally right, rounding everything strictly to 1 decimal place broke the display for tiny values like `0.0012` (and also caused the unit tests to fail). 

I just pushed a fix to dynamically calculate the required decimal precision for numbers . It now preserves 1-2 significant digits for tiny values, so `0.0012` stays `0.0012`, while calculated values like `16.6666667` will correctly be displayed as `16.7`. This should make all the failing unit tests pass now too! Thanks for pointing it out!

stephanegigandet:
Thanks for your PR @pranavchoudhary-tech 

Could you add an integration test with a product with nutrition values with a lot of numbers, and values like 0.0000123 to show that it works (and to make sure we don't break it in the future)

pranavchoudhary-tech:
@stephanegigandet I've just added a test case in `tests/unit/display.t` with a product containing many decimals (including `0.0000123` for calcium) to verify the rounding logic works as expected for the nutrition facts table!

Since I don't have the docker environment running locally to run `make update_unit_tests_results`, the newly added test will fail in CI because the `nutrition-facts-table-many-decimals.json` expected result file is missing. 

Could you please approve the workflows to run? Once they fail, I can either grab the expected JSON diff from the logs and push it, or you can just run `make update_unit_tests_results` before merging. Thanks!

pranavchoudhary-tech:
@stephanegigandet I have successfully updated the expected HTML snapshots to reflect the dynamically rounded nutrition values for the UI. The underlying JSON payload data remains unrounded and intact. The tests should pass flawlessly now!

aleene:
@stephanegigandet this can be accepted?

aleene (commented):
Can not judge this

pranavchoudhary-tech:
Hi @aleene,

To provide some context: this PR addresses #14035 by formatting/rounding ugly long floating-point decimal numbers in the web nutritional tables (e.g. values like `~ 14.0666666666667 g` are now cleanly displayed as `~ 14.1 g`, and `~ 1.71833333333333 g` as `~ 1.7 g`). 

The changes ensure standard number formatting is applied consistently so nutritional panels look clean and readable to end users. You can also see this reflected in the updated test fixture outputs across the knowledge panel tests.

Let me know if you'd like any specific test cases or further details!

lib/ProductOpener/Display.pm
```diff
@@ -9555,9 +9555,29 @@ CSS
...
+								my $decimals = 1;
+								if (abs($formatted_value) < 1 && abs($formatted_value) > 0) {
+									if ($formatted_value =~ /\.(0+)/) {
+										$decimals = length($1) + 2;
+									}
+									else {
+										$decimals = 2;
+									}
+								}
+								require ProductOpener::Numbers;
```
hangy, line 9577: We usually add required modules at the top of the file, ie. `use ProductOpener::Numbers qw/round_to_max_decimal_places/;`. Is there a specific reason to use `require` in this case?

lib/ProductOpener/Display.pm
```diff
@@ -9555,9 +9555,29 @@ CSS
...
+							if (($formatted_value . ' ') =~ /e/i) {
+								# Normalize scientific notation without %f's fixed 6-decimal truncation
+								$formatted_value = sprintf("%.15f", $value);
+								$formatted_value =~ s/0+$//;
+								$formatted_value =~ s/\.$//;
+							}
+							
+							# Round numeric values to 1 decimal place to avoid
+							# ugly calculated values like 16.6666666667 (Issue #14035)
+							if ($formatted_value =~ /^-?\d+\.\d{3,}$/) {
```
hangy, line 9567: A lot of the changes in `Display.pm` assume `.` as the decimal point. We use CLDR to use locale-aware formatting in a lot of places of `ProductOpener`. That means assuming `.` is the decimal separator can be incorrect.

lib/ProductOpener/Display.pm
```diff
@@ -9555,9 +9555,29 @@ CSS
...
+								if (abs($formatted_value) < 1 && abs($formatted_value) > 0) {
+									if ($formatted_value =~ /\.(0+)/) {
+										$decimals = length($1) + 2;
+									}
+									else {
+										$decimals = 2;
+									}
+								}
+								require ProductOpener::Numbers;
+								my $rounded = ProductOpener::Numbers::round_to_max_decimal_places($formatted_value,
```
hangy, line 9578: At this point `$formatted_value` might be a string instead of a number. Passing it to `round_to_max_decimal_places` might incur an unwanted conversion, which might especially incorrect with other decimal separators than `.`?

tests/integration/expected_test_results/web_html/fr-product.html
```diff
@@ -2467,7 +2467,7 @@ <h3 class="panel_title_card"></h3>
          <span
          
          >
-            ~ 22.2222222222222 %
+            ~ 22.2 %
```
hangy, line 2470: Shouldn't `,` be used as a decimal separator for French? cc @stephanegigandet & @teolemon

hangy (commented):
I'd really like to see additional test cases for different locales.

For example: Does it work correctly with the Arabic locale?

<sub>Commented in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/openfoodfacts/openfoodfacts-server/pull/14055)</sub>

lib/ProductOpener/Display.pm
pranavchoudhary-tech: I've moved the import to the top of the file using `use` as requested. I initially used `require` inline just to avoid loading the module if it wasn't needed, but I agree it's cleaner at the top.

lib/ProductOpener/Display.pm
pranavchoudhary-tech: The regex runs before any CLDR formatting is applied. The `$value` here comes directly from the database, meaning it is still a raw Perl float (like `16.66666667`). Because it hasn't been localized yet, it is guaranteed to use `.` as the decimal separator. This ensures `round_to_max_decimal_places` gets a standard numeric string and won't break on other locales.

I can definitely add an Arabic locale test case if you'd like to double-check this end-to-end.

tests/integration/expected_test_results/web_html/fr-product.html
pranavchoudhary-tech: Good catch. My PR just changed the rounding length (from `22.2222222222222` to `22.2`), which is why this line showed up in the diff. However, the existing code was already outputting the dot instead of a comma for French. It seems the underlying CLDR formatter isn't applying the separator properly here. Let me know if you want me to look into fixing this localization bug as part of this PR, or if we should track it in a separate issue.

pranavchoudhary-tech:
Hi @hangy, thanks for the review! I've replied to the specific threads, but to summarize here:

1. I've moved the `ProductOpener::Numbers` import to the top of the file as requested.
2. Regarding the decimal separators (like Arabic or French) and the string conversion: the rounding logic runs *before* any CLDR localization is applied. The value at this stage is fetched directly from the DB, meaning it is still a raw Perl float using a standard dot (`.`). It safely rounds the standard number before any locale-specific commas or Arabic numerals are introduced.
3. For the French HTML snapshot, the existing code was already outputting a dot instead of a comma for the percentage. Let me know if you want me to look into fixing that pre-existing CLDR formatter issue separately.

If you'd still like me to add an explicit Arabic test case in `display.t` to verify the end-to-end behavior, just let me know and I'll add it!

PR #14093, opened 2026-07-27
taxonomy: translate periods_after_opening.txt

teolemon:
taxonomy: translate periods_after_opening.txt
Will cleanup the few untaxonomized entries, and ensure people get autocomplete in the future: https://world.openbeautyfacts.org/facets/periods-after-opening


Freso:
instead of essentially duplicating every language entry, would it make sense to have the parser force/ensure a space between number and non-number? (similar to [how organic is handled for finnish](https://github.com/openfoodfacts/openfoodfacts-server/blob/a2661af658e062441b9eb79a2f4659e00de8ee69/lib/ProductOpener/Ingredients.pm#L6881-L6886))

teolemon:
@Freso it would, it was kind of a quick fix as the data wasn't clean in prod.
https://world.openbeautyfacts.org/facets/periods-after-opening

PR #14096, opened 2026-07-27
taxonomy(food_ingredients): Add Swedish ingredients missing from patented herb salt

ginger51011:
### What

Källfräne is the plant, vattenkrasse is the vegetable. They are the same thing according to https://sv.wikipedia.org/wiki/K%C3%A4llfr%C3%A4ne, so I added both

No LLM plz.
### Screenshot or video

N/A

N/A

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -52553,7 +52553,7 @@ nl: waterkers
 pt: agrião
 ru: Брункресс
 sr: Поточарка
-sv: källfräne, vattenkrasse
+sv: vattenkrasse, källfräne
 tr: Su teresi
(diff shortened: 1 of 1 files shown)


taxonomies/food/ingredients.txt
```diff
@@ -52552,6 +52553,7 @@ nl: waterkers
 pt: agrião
 ru: Брункресс
 sr: Поточарка
+sv: källfräne, vattenkrasse
```
Freso:
```suggestion
sv: vattenkrasse, källfräne
```

“vattenkrasse”, i think, is the more common word for the vegetable, so probably the better primary synonym for Swedish

Freso (approved):

PR #14099, opened 2026-07-28
fix: make ingredient list more legible

### LINKED ISSUE #14083: Make the ingredient list more legible
### Task
Make the ingredient list more legible

### Before

Image

### After
Image

pranavchoudhary-tech:
Fixes #14083

This PR improves the legibility of the ingredients list in the Knowledge Panel by wrapping it in a `.panel_ingredients_text` class and slightly increasing the `line-height` and `font-size`.

This creates a more comfortable reading experience for users, especially on smaller screens or when the ingredient list is very dense.

### DIFF
--- a/scss/_off.scss
+++ b/scss/_off.scss
@@ -938,6 +938,12 @@ html[dir="rtl"] {
   font-size: 0.8em;
 }
 
+/* increased line-height for better legibility of ingredients list */
+.panel_ingredients_text {
+  line-height: 1.6;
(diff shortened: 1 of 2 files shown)


stephanegigandet:
Hi @pranavchoudhary-tech , thanks for your PR, but I don't think that it is what @teolemon wanted. Please wait for the issue to be decided upon before continuing the implementation, as it's not yet clear what should be done.

PR #14101, opened 2026-07-28
chore: usda ingredients part 16

aleene:
### What
Adding and checking usda ingredients


taxonomies/food/ingredients.txt
```diff
@@ -253,7 +253,7 @@ en: caramel chocolate preparation
 hr: karamelno čokoladni pripravak
 it: preparazione di cioccolato al caramello
 
-#< en: compound
+# <en:compound
```
teolemon: @aleene could you fix those (they were modified by @Freso to be ready in case we wanted to uncomment them)

aleene: As it is written is OK: spaces between <, language and name

taxonomies/food/ingredients.txt
```diff
@@ -198,7 +199,7 @@ it: fritto, frittura
 #
 # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
 
-# < en:syrup
+# <en:syrup
```
Freso: why are you removing this space?

taxonomies/food/ingredients.txt
```diff
@@ -186,6 +186,7 @@ description:en: Imazalil is a fungicide widely used in agriculture, particularly
 # azb:انیل‌کونازول
 wikidata:en: Q421576
 wikipedia:en: https://en.wikipedia.org/wiki/Enilconazole
+# ingredient/fr:imazalil has 23 products @2019-05-29
```
Freso: why are you readding these lines?

taxonomies/food/ingredients.txt
Freso: yes, but your PR is removing those spaces in favour of only a space between `#` and `<`.

Freso (changes requested):
i'm having trouble finding the actually relevant changes amidst all the reverts happening here. did you botch resolving a merge conflict?

taxonomies/food/ingredients.txt
aleene: You should ask github. I think we should not do a lot of changes at the same time. The merger gets to confused.

taxonomies/food/ingredients.txt
aleene: Ask the merger, I will add those in.

aleene:
@Freso let me finish my USDA stuff before you do systemic changes to the taxonomy. I have another request in the queue. The merger/copilot/github client no longer knows what to do.

taxonomies/food/ingredients.txt
Freso: If GitHub can’t figure it out automatically, then handle the conflicts by hand/manually—or ask for help to do so.

There are always going to be changes happening at the same time, which is exactly what the version control helps handle. If you’re not doing “systemic” changes yourself, it shouldn’t be difficult to incorporate changes done in upstream—esp. in parts of the codes not touched by you at all, like the part that this is a comment on. I have had to resolve multiple conflicts on some of my larger PRs exactly because they’re done against a constantly moving codebase.

I suspect GitHub is more confused by the dozens of merges and commits in this branch’s history (109 commits is a lot for a PR!), a lot of them coinciding with the target branch (`main`), making it difficult to discern where what changes come from.

aleene:
I will redo this update. I cannot put the @freso changes back in.

PR #14113, opened 2026-07-28
taxonomy(ingredients): Add Swedish fava bean variant

ginger51011:
### What

[According to Swedish Wikipedia](https://sv.wikipedia.org/wiki/Bondb%C3%B6na), "favaböna" is an alternative name.

No.
### Screenshot or video

N/A

N/A


Freso (approved):
Tack!

PR #14116, opened 2026-07-29
feat: add facets index page to fix blank page error

### LINKED ISSUE #14102: Going to https://world.openfoodfacts.org/facets currently returns a blank page
### What
- Going to https://world.openfoodfacts.org/facets currently returns a blank page. I would expect it to return a list of available facets, and I have this (probably false) memory that we had this at some point.
- That's what we're doing in Explorer: https://beta.openfoodfacts.net/facets

pranavchoudhary-tech:
Fixes #14102

When visiting the `/facets` route without any search parameters, the router previously classified the route as `facets_products`, but `cgi/display.pl` exited silently due to a lack of tags, which resulted in a blank page (or 502/503 error) for end users.

This PR introduces a `facets_index` handler which automatically generates and returns a comprehensive list of all supported facets extracted from `%ProductOpener::Lang::tag_type_plural`. This brings the behavior in line with what is implemented on the beta site and provides a much better user experience than a broken page.

### Screenshots

**Before**
Screenshot 2026-07-29 at 10 20 08 AM

**After**
Screenshot 2026-07-29 at 10 21 17 AM


lib/ProductOpener/Display.pm
```diff
@@ -7547,6 +7548,26 @@ sub search_permalink ($request_ref) {
 	return $html;
 }
 
+sub display_facets_index ($request_ref) {
+	my $template_data_ref = { facets => [] };
+	foreach my $tagtype (sort keys %ProductOpener::Lang::tag_type_plural) {
```
stephanegigandet: We should use values from the drilldown_fields array: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Config_off.pm#L840
Otherwise this will list facets like nutriscore that are not available on all site flavors (like Open Beauty Facts)

po/common/am.po
```diff
@@ -7622,7 +7622,7 @@ msgstr ""
 # Please change get-it-on-en.svg to get-it-on-XX.svg. check the url https://static.openfoodfacts.org/images/misc/f-droid/svg/get-it-on-XX.svg
 msgctxt "f_droid_app_icon_url"
 msgid "/images/misc/f-droid/svg/get-it-on-en.svg"
-msgstr ""
+msgstr "/images/misc/f-droid/svg/get-it-on-am.svg"
```
stephanegigandet, line 7625: Those changes to .po files are unrelated, can you remove them?

templates/api/knowledge-panels/health/ingredients/ingredients.tt.json
```diff
@@ -17,7 +17,7 @@
         {
             "element_type": "text",
             "text_element": {
-                "html": `[% panel.ingredients_text_with_allergens %]`,
+                "html": `<div class="panel_ingredients_text">[% panel.ingredients_text_with_allergens %]</div>`,
```
stephanegigandet: seems unrelated

tests/integration/facets.t
```diff
@@ -199,6 +199,13 @@ edit_product(
 # Note: expected results are stored in json files, see execute_api_tests
 # We use the API with .json to test facets, in order to easily get the products that are returned
 my $tests_ref = [
+	{
+		test_case => 'facets_index',
```
stephanegigandet, line 203: the PR is missing the corresponding expected result file, you can generate it with "make update_tests_results"

PR #14119, opened 2026-07-29
feat: add density to ingredients for volume quantity conversions to g - WIP

### LINKED ISSUE #14087: Taxonomy: add density to ingredients frequently expressed as volumes (liquids, creamy, etc.)
### Problem

In recipes, some ingredients are expressed as ml/cl or some kind volumes (table spoon, cups, ...) while to compute environmental score contribution of a single ingredient we need a measure in grams.

### Proposed solution
Add density of ingredients in the taxonomy (expressed as kg/l)

For example for water:
```
density_kg_per_l:en: 1
```

### Additional context

See also https://github.com/openfoodfacts/score-my-recipe/issues/30

This information might be part of USDA data.

Also an issue propose computing it based on data we have in OFF (might be very partial): #545

stephanegigandet:
Needed in particular to analyze recipe ingredients that can be listed by volume (e.g. 500ml of milk).

Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/14087

WIP, still need to add some tests, and to add support for common units in food recipes

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -14734,11 +14734,9 @@ zh: 油
 # yue:油
 # zh-hans:油
 # zh-hant:油
+density_g_per_ml:en: 0.92
 wikidata:en: Q42962
 wikipedia:en: https://en.wikipedia.org/wiki/Cooking_oil
-
-# description:en:Hydrogenation converts liquid vegetable oils into solid or semi-solid fats, such as those present in margarine. Changing the degree of saturation of the fat changes some important physical properties, such as the melting range, which is why liquid oils become semi-solid.
-
 
 # nova:en:hydrogenated-oil
 # processing:en: en:hydrogenated
@@ -14754,6 +14752,7 @@ hu: hidrogénezett olaj
 it: olio idrogenato
 nb: hydrogenert olje
 nl: Gehydrogeneerde olie
+description:en: Hydrogenation converts liquid vegetable oils into solid or semi-solid fats, such as those present in margarine. Changing the degree of saturation of the fat changes some important physical properties, such as the melting range, which is why liquid oils become semi-solid.
 wikidata:en: Q57656844
 wikipedia:en: https://en.wikipedia.org/wiki/Hydrogenation#Food_industry
 # ingredient/hydrogenated-oil has 283 products in 9 languages @2019-07-10
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -19,7 +19,6 @@
 # vegetarian:en     can have values yes, no, maybe - in the case of maybe, if the ingredient has sub-ingredients, then only its sub-ingredients are considered
 # from_palm_oil:en  can have values yes, no, maybe - in the case of maybe, if the ingredient has sub-ingredients, then only its sub-ingredients are considered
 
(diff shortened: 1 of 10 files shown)


taxonomies/food/ingredients.txt
```diff
@@ -15288,10 +15294,11 @@ tr: Bitkisel yağ
 uk: Рослинні жири й олії
 vi: Dầu thực vật
 zh: 植物油
+# ingredient/vegetable oil has 103341 products @2019-07-10
+density_g_per_ml:en: 0.92
 # roa-rup:Untulemnu
 # sco:Vegetable ile
 wikipedia:en: https://en.wikipedia.org/wiki/Vegetable_oil
```
Freso: Why are you moving the ingredient counts around?
```suggestion
# roa-rup:Untulemnu
# sco:Vegetable ile
density_g_per_ml:en: 0.92
wikipedia:en: https://en.wikipedia.org/wiki/Vegetable_oil
# ingredient/vegetable oil has 103341 products @2019-07-10
```

taxonomies/units.txt
```diff
@@ -626,6 +626,16 @@ it: cucchiaino metrico
 conversion_factor:en: 5
 standard_unit:en: ml
 
+# US legal cup (used in nutrition facts) is 240ml, different from the metric cup 250ml used in Canada, Australia, New-Zealand etc.
+# and from the traditional US cup
+# for "cups", we default to US legal cups.
```
Freso: Could this be a `description:en:` or `comment:en:` property?

tests/unit/expected_test_results/taxonomies/eurocodes.csv
```diff
@@ -70,7 +70,7 @@ eurocode_2_group_2	eurocode_2_group_3	ingredient
 8.15	8.15.20	en:cabbage
 8.15	8.15.20	en:savoy-cabbage
 8.15	8.15.24	en:red-cabbage
-8.15	8.15.28	en:napa-cabbage
```
Freso: This doesn’t seem like a related change, unless there’s something changing with how ids are determined. Or is this just a remnant that no other `/update_tests_results` run has caught? (Though, do those still update CSV files? Maybe they don’t because if how much noise that was generating.)

taxonomies/food/ingredients.txt
stephanegigandet: I just added density_g_per_ml:en: 0.92 at the end, but then the lint applied the last comment to density I guess

taxonomies/units.txt
stephanegigandet: Good idea

tests/unit/expected_test_results/taxonomies/eurocodes.csv
stephanegigandet: yes, just a test update thing I guess

alexgarel (approved):
Thanks @stephanegigandet 

As @Freso noted there is the eurocodes difference to fix.
And adding some test is a good idea :-)

taxonomies/food/ingredients.txt
Freso: Made a motion to drop the counts entirely: https://github.com/openfoodfacts/openfoodfacts-server/pull/14126 :)

PR #14126, opened 2026-07-30
chore(taxonomy): drop product counts

Freso:
These counts are mostly quite old (though there are a couple from 2025), and while the historical data might be interesting, it would probably be better to extract that from the database as needed, rather than storing it as comments in the taxonomy files.

I mostly consider these when they appear as noise in PRs, being moved around, but maybe there is some use for them that I’m not aware of.

This also reduces the size of the taxonomy files (if only marginally), which may help some editors better deal with them.

The changes were made using `sed --follow-symlinks -iE '/[pattern]/d' taxonomies/**.txt` with the following patterns:
- `^# openfoodfacts:`
- `# [[:digit:]]\+ products`
- `# [[:digit:]]\+ entry`
- `# [[:digit:]]\+ entries`
- `# [[:digit:]]\+ in [[:digit:]]\+ @20`
- `# .* has [[:digit:]]\+ products\?`

### DIFF
--- a/taxonomies/additives.txt
+++ b/taxonomies/additives.txt
@@ -336,7 +336,6 @@ zh: E102, :柠檬黄
 additives_classes:en: en:colour
 colour_index:en: CI 19140
 description:en: TARTRAZINE is a synthetic lemon yellow azo dye primarily used as a food coloring.
-# ingredient/fr:tartrazine has 196 products in 3 languages @2019-02-09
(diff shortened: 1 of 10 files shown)


aleene:
Yes, I put them in when starting with the taxonomy. OK to remove them.

stephanegigandet (approved):
Thanks

PR #14137, opened 2026-07-30
feat: add a script to convert product schema

alexgarel:
This script convert a dump of jsonl to a specific version schema (whatever the initial schema of each product).
So you can apply it to a mongodb dump to get products in a uniform schema. 

I need it to import initial data in searchalicious.
It will be also of use for open food facts exports (to be deployed in the future).

Disclaimer: code generated by glm2 after a process of validation, reviewed by me.

### DIFF
--- a/scripts/convert_product_schema_dump.pl
+++ b/scripts/convert_product_schema_dump.pl
@@ -1,0 +1,247 @@
+#!/usr/bin/perl
+
+# This file is part of Product Opener.
+#
+# Product Opener
+# Copyright (C) 2011-2026 Association Open Food Facts
+# Contact: contact@openfoodfacts.org
+# Address: 21 rue des Iles, 94100 Saint-Maur des Fossés, France
+#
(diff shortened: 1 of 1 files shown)


scripts/convert_product_schema_dump.pl
```diff
@@ -0,0 +1,247 @@
...
+}
+
+$target_version //= $current_schema_version;
+validate_target_version($target_version);
+
+my $opened_input = defined $ARGV[0];
+my $in = open_input();
+$out = open_output($output_path);
+
+while (my $line = <$in>) {
```
stephanegigandet, line 244: I think this hangs is the script is called without any parameter at all.

stephanegigandet (approved):
Looks good, one minor comment

scripts/convert_product_schema_dump.pl
alexgarel: I made it explicit, by passing "-" as filename.

PR #14142, opened 2026-07-31
taxonomy: add missing es/pt/it synonyms to heart of palm

lucaspretti:
## What

Adds the missing Spanish, Portuguese and Italian synonyms to the existing `heart of palm` entry, so that ingredient lists naming it in those languages are actually recognised. Also adds a section header and an English description, and the `eurocode_2_group_3` value the file's own header comments already used palm heart as the example for.

## Why

Found while building a small vegan-scoring app on the OFF API. A real product, [Palmito Natural Entero (8480000862723)](https://world.openfoodfacts.org/product/8480000862723), lists only `Corazones de palmito, agua, sal y ácido cítrico E330`. Water, salt and citric acid all resolve as vegan, but `corazones de palmito` matched nothing (`is_in_taxonomy: 0`), because the entry carried only `es: Palmito`. One unrecognised ingredient is enough to drag the whole product to `vegan-status-unknown` / `vegetarian-status-unknown`, so a product whose every ingredient is unambiguously plant-derived came back undetermined.

Verified this was a taxonomy gap and not stale analysis: resubmitting the identical `ingredients_text` via the write API on staging left the tags unchanged, since there was no node to match against.

With the synonyms in place the ingredient resolves, and as @Freso notes it then inherits `vegan`/`vegetarian` from `en:vegetable` via `en:shoot-vegetable`, so no explicit properties are needed on the entry.

## Note on this PR's history

The first version of this PR wrongly added a second, duplicate entry, because I concluded no entry existed from a `grep` truncated at 20 results, where the real one fell below the cut. @Freso caught it and redirected the change to where it belonged. The diff now touches only the existing entry.

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -53819,24 +53819,6 @@ ciqual_food_name:fr: Artichaut, fond, surgelé, cru
 
 ###################################################################################################
 #
-#                                Palm heart
-#
-###################################################################################################
-
-#category/palm-hearts
-< en: shoot vegetable
-en: Palm heart, heart of palm, hearts of palm
-es: Corazón de palmito, corazones de palmito, palmito
-pt: Palmito, coração de palmeira
-fr: cœur de palmier, coeurs de palmier, chou palmiste
-de: Palmherzen
-it: Cuore di palma
-vegan:en: yes
-vegetarian:en: yes
-description:en: The heart of palm is the inner core of certain palm trees, harvested as a vegetable and typically canned or jarred in brine.
-
-###################################################################################################
-#
 #                                Asparagus
 #
 ###################################################################################################
@@ -55017,13 +54999,19 @@ usda_ndb_name:en: Cabbage, red, raw
 wikidata:en: q263203
 wikipedia:en: https://en.wikipedia.org/wiki/Red_cabbage
 
+###################################################################################################
+#
+#                                Palm heart
+#
+###################################################################################################
+
 # <en:category:en:Fresh plant-based foods
 < en: shoot vegetable
 en: heart of palm, Hearts of palm, palm heart
 ar: جمار
 de: Palmherzen, Palmherz
 eo: Palmomedolo
-es: Palmito
+es: palmito, corazón de palmito, corazones de palmito
 fa: پنیر نخل
 fi: Palmun sydän
 fr: Cœur de palmier, cœurs de palmiers, cœurs de palmier, coeur de palmier, coeurs de palmiers, coeurs de palmier, choux palmistes, chou palmiste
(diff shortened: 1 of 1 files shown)


taxonomies/food/ingredients.txt
```diff
@@ -53817,6 +53817,24 @@ ciqual_food_code:en: 20232
...
+en: Palm heart, heart of palm, hearts of palm
+es: Corazón de palmito, corazones de palmito, palmito
+pt: Palmito, coração de palmeira
+fr: cœur de palmier, coeurs de palmier, chou palmiste
+de: Palmherzen
+it: Cuore di palma
+vegan:en: yes
+vegetarian:en: yes
+description:en: The heart of palm is the inner core of certain palm trees, harvested as a vegetable and typically canned or jarred in brine.
+
```
Freso:
```suggestion
```

taxonomies/food/ingredients.txt
Freso, line 55008:
```suggestion
###################################################################################################
#
#                                Palm heart
#
###################################################################################################

# <en:category:en:Fresh plant-based foods
```

Freso:
```suggestion
es: palmito, corazón de palmito, corazones de palmito
```

Freso:
```suggestion
pt: palmito, coração de palmeira
```

Freso:
```suggestion
it: palmito, cuore di palma, cuori di palma
```

Freso, line 55034:
```suggestion
description:en: The heart of palm is the inner core of certain palm trees, harvested as a vegetable and typically canned or jarred in brine.
eurocode_2_group_3:en: 8.25.50
```

Freso (changes requested):
Heart of palm _is_ already in the taxonomy, but it doesn’t have “corazones de palmito” as a synonym for Spanish.

It _should_ inherit the `vegan`/`vegetarian` properties from `en:vegetable` (via `en:shoot-vegetable`), once it’s recognised. :)

lucaspretti:
Thanks @Freso, all six suggestions applied in def0b03, and thank you for redirecting this rather than just closing it as a duplicate.

**Correcting my own PR description**: I claimed no entry existed for palm heart. That was wrong. It does exist, and it already had the `shoot vegetable` parent, twenty translations and the eurocode. I reached the wrong conclusion from a `grep` truncated at 20 results, where the palm heart entry fell below the cut. Sorry for the noise.

The underlying gap is real, but narrower and more precise than I described. The entry carried `es: Palmito`, while the product that led me here lists **"Corazones de palmito"** in its ingredients ([8480000862723](https://world.openfoodfacts.org/product/8480000862723)), which doesn't match that synonym, so the ingredient stayed unrecognised (`is_in_taxonomy: 0`) and dragged the whole product to `vegan-status-unknown` even though water, salt and citric acid all resolved as vegan. Your suggestion #3 is exactly the fix.

The PR now proposes only improvements to the existing entry: section header, the missing es/pt/it synonyms with lowercase-first ordering, and the description above the eurocode line.

Freso (approved):
Looks great! Thank you! :)

Freso:
@lucaspretti Do you want to actually update the PR description? Then I can use that as the commit message when merging this. :)

lucaspretti:
Done, title and description rewritten to describe what the PR actually does, so it should work as a commit message now. Thanks for the review and for the pointer about inheriting `vegan`/`vegetarian` from `en:vegetable` via `en:shoot-vegetable`, that's noted on my side too.

Freso:
Thank you (again)! :)

lucaspretti:
Merged, thank you! And sorry about the trouble: the duplicate entry, then the description rewrite. You could reasonably have closed this as a duplicate and instead you showed me where the change actually belonged, with ready-to-apply suggestions. That's a good first experience of contributing here.

Two things I've written down for next time: search the taxonomy properly before claiming something is missing (mine was a truncated grep, and the real gap was a missing `es` synonym, not a missing entry), and let the properties inherit from the parent rather than setting `vegan`/`vegetarian` explicitly.

Thanks again for the patience.

PR #14149, opened 2026-08-01
taxonomy(food): mirin/rice wine edits

Freso:
- Splits out rice wine from sake.
  - All sake is rice wine, not all rice wine is sake.
- Moves “Mirin” out of alcoholic beverages and into condiments, and adds three “mirin” types as children.
  - Regardless of the type of “mirin”, its primary use is as a condiment/seasoning.
  - Two of them are actual rice wines and have been set as children of the new rice wine category.
- Adds/fixes some `agribalyse`/`ciqual` and `wikidata` properties.
- Imports some translations from Wikidata.
- Normalises towards sentence-cased plural forms.

Source: https://en.wikipedia.org/w/index.php?title=Mirin&oldid=1345671002

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -1560,7 +1560,6 @@ uk: Ферментовані продукти
 vi: Thực phẩm lên men
 zh: 发酵食品
 
-
 < en: Fermented vegetables
 en: Kimchi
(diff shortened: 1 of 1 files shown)


teolemon (approved):
Looks great @Freso

PR #14170, opened 2026-08-01
fix: OCRed ingredients with "allergy advice" or "nutrition:" still in the text

svensven:
### What
As I've been editing product ingredients, lots still have "allergy advice" or "nutrition" text still on the end. The OCR already correctly trims "for allergens", but not the text that generally precedes it, at least in the UK.
https://uk.openfoodfacts.org/facets/ingredients/allergy-advice

"Nutrition" is often followed by a colon, which the script doesn't handle.
https://uk.openfoodfacts.org/facets/ingredients/nutrition was much worse, but I've been fixing them one at a time.

image

- Fixes:


stephanegigandet (approved):
Thanks

Issue #6688, opened 2022-04-25
Add support for language variants (like Simplified/Traditional Chinese and PT and BR Portuguese)


teolemon:

TaciteOFF:
Could we split this into two steps?
1. Support language variants in the interface first, including a fallback when a translation is missing.
2. Then support separate product text and images for each variant.
Adding translation files alone is not enough: the current loader skips files like pt_BR.po, and other parts of the server still expect two-letter codes.

A few things to watch out for:
- Keep existing pt and zh data and API clients working.
- Make sure saving one variant does not overwrite another.
- Choose fallbacks carefully, especially between Traditional and Simplified Chinese.
- Prefer standard language codes over temporary codes like xt, to avoid another migration later.

Tests should cover loading translations, missing translations, and saving both variants on the same product.

teolemon:
I'm adding an old related bug about the API part

HugoCebrian:
Hi! I'm starting a project that mirrors and contributes Brazilian data to OFF. I'd like to add the **taxonomy / ingredient-parsing** side to this discussion, since so far it (and #14471, nice work @TaciteOFF) focuses on UI catalogs and API codes.

### Variants matter for taxonomies too

Brazilian products are stored with `lc: pt`, so their ingredient lists are parsed with the shared `pt` taxonomy entries, which mix both variants. Some words refer to different foods in each country:

| Word | Portugal | Brazil | Current mapping |
|---|---|---|---|
| `cuscuz` | couscous (durum wheat) | steamed corn dish, gluten-free | `en:couscous` → child of `en:wheat` → allergen `en:gluten` |
| `requeijão` | fresh whey cheese | processed spreadable cheese | `en:ricotta` |

These can't be fixed with `pt:` synonyms without breaking Portugal. To be transparent: I haven't found a live product affected yet (the 3 Brazilian products tagged `en:couscous` are real wheat couscous), so this is latent, but it's the clearest case for a `pt_BR` override.

Display names are the other case: Brazilian users see `Natas` (en:creams), `Pequenos alomoços` (en:breakfasts) or `Emulsionante` (en:emulsifier), where Brazil says *creme de leite*, *café da manhã*, *emulsificante*. Synonyms make those match, but can't change what is displayed.

### Most of the Brazilian gap does *not* need variants

[...]

PR #14173, opened 2026-08-01
l10n: possibly pointless attempt to fix english vs american

svensven:
### What
Possibly pointless, as I noticed uk.OFF runs as "en" and so gets american defaults for things, e.g.
image
image

units.txt says it takes pairs of values, so rather than add more to en, I added en_GB language entries.

I'm assuming nutrients.txt is what provides the dropdown entries in the screenshots? I can't find them anywhere else.


teolemon (commented):
- The blocker here is : https://github.com/openfoodfacts/openfoodfacts-server/issues/6688
- The variants won't be used until we fix it.

PR #14176, opened 2026-08-02
l10n(ja): Replace translation of saturated fat with 飽和脂肪酸

Naruyoko:
Although both are used, 飽和脂肪酸 'saturated fatty acid' is more common.


Freso (commented):
Looks good to me, but I need to double check that it’s okay to edit the PO files directly rather than through Weblate.

Freso:
I’m going to merge this for the taxonomy updates, but note that there’s a good chance that your changes to the PO files will get overwritten on the next Crowdin sync.

Freso (approved):

PR #14177, opened 2026-08-02
l10n(ja): Replace some instance of 消費 to 摂取

Naruyoko:
Although both are translatable to "consume", the former is closer to the meaning in economics and alternately "expend". The latter fits better for nutrition take in.


Freso (commented):
Seems good to me; just need to confirm that it’s fine to edit the PO files directly rather than through Weblate. :)

Freso (approved):
Just a heads up that the title mentions taxonomies, but no taxonomy edits are in this PR.

Going to go ahead and merge, as it seems like the PO changes _might_ survive the Crowdin sync.

PR #14180, opened 2026-08-02
fix(api): normalize blessed product values before serialization

### LINKED ISSUE #13752: Product edit fails with "encountered object '1', but neither allow_blessed, convert_blessed nor allow_tags settings are enabled"
## Bug Report

**Reported by:** François Gouget on OFF Slack (#general channel)

**Description:**
Editing certain products fails with the following error when clicking Save:
Software error: encountered object '1', but neither allow_blessed,
convert_blessed nor allow_tags settings are enabled (or TO_JSON/FREEZE
method missing) at /srv/off/lib/ProductOpener/APIProductServices.pm line 161.

**Affected products (confirmed reproducible):**
- 0744096655408
- 3660992012756
- 0731509930092

**Steps to reproduce:**
1. Go to any of the affected product pages
2. Click Edit
3. Make any change
4. Click Save
5. Error appears

**Notes:**
- The error is consistent and reproducible on the same products
- Other products edit fine — only specific products trigger this
- The error originates in `APIProductServices.pm` line 161
- Stéphane is aware and investigating (Slack thread Jun 12 2026)

**Environment:** production (world.openfoodfacts.org)

subhwastaken:
## Summary
- normalize blessed scalar, array, and hash values before sending product-service payloads
- preserve custom `TO_JSON` conversions while ensuring unsupported blessed values become plain JSON data
- add a regression test for blessed scalar values

## Why
Some products fail to save because a non-plain blessed value reaches the product-services JSON encoder. Normalizing the request payload prevents the serialization exception while keeping strict JSON encoding.

Fixes #13752


lib/ProductOpener/APIProductServices.pm
```diff
@@ -161,7 +162,7 @@ sub add_product_data_from_external_service ($request_ref, $product_ref, $url, $s
 
 	my $response = $ua->post(
 		$url,
-		Content => $json_for_objects->encode($body_ref),
+		Content => $json_for_objects->encode(_plain_data_for_json($body_ref)),
```
stephanegigandet, line 165: $json_for_objects should already convert blessed values:

in Store.pm:

$json_for_objects = Cpanel::JSON::XS->new->allow_nonref->convert_blessed->utf8;

subhwastaken: You're right that `$json_for_objects` already enables `convert_blessed`. My reasoning was that the value triggering #13752 still wasn't serializable with the current configuration. `_plain_data_for_json()` preserves existing `TO_JSON` conversions when available, while recursively normalizing unsupported blessed scalar/array/hash references into plain Perl data before encoding, which is why I introduced it.

If there's a simpler way to handle this with the existing serializer configuration, I'm happy to adjust the implementation.

subhwastaken:
Updated the branch with the latest `main` and confirmed that all required checks are now green, including `triage`. The PR remains open and conflict-free; please review when convenient. Thank you!

PR #14181, opened 2026-08-02
fix(build): upgrade npm to 11.18.0

### LINKED ISSUE #14139: Upgrade our version of npm
### Task
- #16 0.934 npm notice New major version of npm available! 10.9.4 -> 11.18.0
- not sure why it doesn't show up in dependabot, probably linked to wider node upgrade ?

subhwastaken:
## Summary
- upgrade the frontend build image to npm 11.18.0
- install npm with lifecycle scripts disabled before dependency installation

## Why
The repository's frontend build reports that npm 10.9.4 is outdated and should be upgraded to 11.18.0.

Fixes #14139


hangy:
Thanks for the suggestion @subhwastaken! I think this gets us halfway there. IMHO, we should probably set the `packageManager` property in `package.json`, so that GitHub Actions also [use the configured version](https://github.com/actions/setup-node/blob/e51e5fe84fc33b4c73ebe40526b2694712b5b858/README.md#caching-global-packages-data). Potentially, this could also be used in `Dockerfile.frontend`, so that there's one centrally configured `npm` version. Do you want to take a stab at this?

subhwastaken:
Thanks for the suggestion! I'll look into centralizing the npm version via the `packageManager` field and wiring it up for both GitHub Actions and the Dockerfile.

subhwastaken:
@hangy I've updated the PR to add the packageManager field to package.json to declare the npm version in a central location.

I also looked through the current GitHub Actions workflows and Dockerfile.frontend. The workflows already use actions/setup-node, while the Docker build still installs npm@11.18.0 explicitly. I couldn't find an existing mechanism in the repository to reuse the packageManager value from package.json during the Docker build. If you have a preferred approach for that part, I'd be happy to update the PR accordingly.

PR #14182, opened 2026-08-02
taxonomy: misc

svensven:
### What
From this weekend's ingredients editing.

(I'm wondering if there's actually a standard for the descriptions? I just copied the neighbour's style in each case...)

- Fixes:

### DIFF
--- a/taxonomies/ingredients_processing.txt
+++ b/taxonomies/ingredients_processing.txt
@@ -451,6 +451,9 @@ uk: тертий, терта, терте, терті
 # sv:parse:riven $
 # en:wiktionary:grated
 
+# en:description:Generally herbs, dried then rubbed against each other and the stems/stalks discarded
+en: rubbed
+
 en: shred, shredded
 pt: retalhado, retalhada, retalhados, retalhadas
 
@@ -1295,6 +1298,9 @@ sv: från koncentrat, delvis från koncentrat
 uk: з концентрату
 zh: 浓缩还原
 description:en: Water has been added to concentrate
+
+en:from comminute
+description:en:A comminute is a puree that is derived from whole citrus fruits.
 
 #en:description:Dehydration by artificial or natural processes
 # anhydrous: No water, dehydrated: Water removed, dried: Less water.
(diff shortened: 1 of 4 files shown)


taxonomies/ingredients_processing.txt
```diff
@@ -451,6 +451,9 @@ uk: тертий, терта, терте, терті
 # sv:parse:riven $
 # en:wiktionary:grated
 
+# en:description:Generally herbs, dried then rubbed against each other and the stems/stalks discarded
+en: rubbed
```
Freso:
```suggestion
en: rubbed
description:en: Generally herbs, dried then rubbed against each other and the stems/stalks discarded
```

Might be good to also include that it does not _always_ refer to this, which is why it isn’t a synonym or child of `dried ground`. (Could be a `comment:en:` rather than part of the `description` though.)

taxonomies/ingredients_processing.txt
```diff
@@ -1296,6 +1299,9 @@ uk: з концентрату
 zh: 浓缩还原
 description:en: Water has been added to concentrate
 
+en:from comminute
+description:en:A comminute is a puree that is derived from whole citrus fruits.
```
Freso:
```suggestion
en: from comminute
description:en: A comminute is a puree that is derived from whole citrus fruits.
```

taxonomies/ingredients_processing.txt
```diff
@@ -2877,6 +2883,9 @@ uk: без лактози, безлактозний, безлактозна, б
 description:en: for ingredients, “lactose-free” is not a processing per se, but involves a chemical process, use of enzyme lactase to break down lactose into simpler sugars (glucose and galactose), which are easier to digest for those with lactose intolerance.
 wikidata:en: Q11874508
 
+en: lactose-reduced, lactose reduced
+description:en:Not lactose free, the ingredient still shows as containing milk as an allergen
```
Freso:
```suggestion
description:en: Not lactose free, the ingredient still shows as containing milk as an allergen
```

Freso (changes requested):
Some minor formatting changes/fixes, but otherwise seems good. :)

The standard is `description:XX: `, with XX being the language code relevant for the description. (E.g., some taxons have descriptions in French (`fr`), Dutch (`nl`), or other languages.)

Also note that all keys end with colon-space (`: `) and then the value, so: `key: value`. (e.g., to define the key `description:en` you use `description:en: some description in english (en)`.)

Freso (approved):

PR #14183, opened 2026-08-02
fix: nutrient unit change causing UI freeze (#14117)

### LINKED ISSUE #14117: web: changing unit blocks the interface
### What
when changing the units, it takes some 10 seconds before the change is accepted and the interface is usable again.

https://github.com/user-attachments/assets/7c39d32e-7bd3-47d3-bfa8-228c163f0026

1. Go to any product, like: https://world.openfoodfacts.org/product/5214000075097/greek-honey-with-red-hot-chili-pepper-vasilissa
2. Click on edit
3. change any unit in the nutritional table

### Expected behavior
- no lag, direct response

### Why
- annoying, stops workflow

### Type of device
- Browser: name  Safari

any edit for which a nutrient unit must be changed

ganga-300:
## What

Fixes #14117 - changing a nutrient's unit on the product edit page's nutrition table caused the entire UI to freeze for approximately 10 seconds before becoming responsive again.

## Root cause

In `html/js/product-multilingual.js`, the shared `.nutrient_unit` change handler was scoped to re-validate every nutrient field on the page, not just the one that changed:

```javascript
$('.nutrient_unit').on('change', function () {
    $('.nutrient_value').trigger('input');
});
```

This is expensive on its own, since `check_nutrient()` performs multiple DOM lookups and CSS writes (for warning highlighting) per field. It gets significantly worse because of how unit changes cascade:

When a global unit selector changes (`.global_nutrient_unit`), it updates every matching per-cell unit selector for that nutrient - typically 4 (as-sold/100g, as-sold/serving, prepared/100g, prepared/serving) - and explicitly triggers a `change` event on each of them individually:

```javascript
$('.global_nutrient_unit').on('change', function() {
    ...
    $('.nutrient_' + nutrient_id + '_unit').trigger('change');
});
```

Since each of those 4 triggers fires the shared handler above, a single unit change ends up re-validating every nutrient field on the page, 4 times over - hundreds of redundant `check_nutrient()` calls, each doing DOM reads/writes, on one user action. That volume of repeated DOM work (including forced style recalculation from `show_warning()`'s `.css()` calls) is what produces the multi-second freeze.

## Fix

Scoped the `.nutrient_unit` change handler to only re-validate the nutrient(s) actually affected by the change, instead of every nutrient on the page:

- If the changed nutrient is one of the cross-validated nutrients (`fat`, `carbohydrates`, `sugars`, `saturated-fat`), re-validate all four, since their warnings depend on each other.
- Otherwise, only re-validate the single nutrient that changed.

This preserves the existing cross-validation behavior (sugars > carbohydrates, saturated fat > fat) while cutting the work from "every field on the page, up to 4x" down to at most 4 fields per change.

## Testing

Tested locally on the product edit page's Nutrition tab, on a product with 10+ nutrients added (wide table):

- [x] Changing a nutrient's unit is now instant - no perceptible delay (previously approximately 10s freeze)
- [x] Sugars greater than carbohydrates warning still triggers correctly when both fields share the same unit
- [x] Saturated fat greater than fat warning still triggers correctly when both fields share the same unit
- [x] Changing the unit on an unrelated nutrient (outside the fat/carbs/sugars/saturated-fat group) does not affect other fields' warnings
- [x] Changing a global unit selector (which cascades to all "per"/"preparation" variants) is also fast and correct
- [x] No new errors in the browser console

## Notes for reviewers

This is a minimal, targeted fix - no changes to validation logic itself (`check_nutrient()`, `get_nutrient_value()`, `show_warning()` are untouched), only to which fields get re-validated on a unit change.


stephanegigandet (approved):
Looks good, thank you

PR #14185, opened 2026-08-03
fix: tags levels computation when building taxonomies

stephanegigandet:
This PR:
- fixes a bug in the computation of levels for parents in taxonomies (for the forest footprint I added a get_tag_parent() function in Tags.pm and didn't get the results ordered as I expected).
- fixes to the Forest Footprint that were in this PR have been moved to a separate PR: https://github.com/openfoodfacts/openfoodfacts-server/pull/14210

Note that the levels will probably change in many taxonomies, which may cause a lot of test results to  change.


lib/ProductOpener/Tags.pm
```diff
@@ -5031,6 +5079,8 @@ Returns the path of the tag in the taxonomy (from the root to the tag, included)
 
 If there are multiple parents for the tag (or one of its parents), we take the first parent.
 
+See also:get_tags_parents() if you need a list with all parents.
```
Freso:
```suggestion
See also: get_tags_parents() if you need a list with all parents.
```
Or is the lack of space intentional?

Freso (commented):
Not seeing anything obviously wrong, but I’m not confident enough to approve. :)

Freso:
I fixed the taxonomy errors from https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/30884806782/job/91913556097?pr=14185 here: https://github.com/openfoodfacts/openfoodfacts-server/pull/14199

Freso (approved):
Okay, I’ve looked it over again, and I’m not seeing anything that could irrevocably harm the data or cause other issues that wouldn’t be manageable/fixable, so I’m going to go ahead and approve so it can move along. 🙏

john-gom:
@stephanegigandet I created a test locally. Created one tag called "en:z" and another called "en:z-yogurt" with parents of "en:z" and "en:yogurt". When I call get_tag_with_parents for "en:z-yogurt" I would expect to see "en:yogurt" listed before "en:z" because it is alphabetically sooner, but it comes after as it is at a higher level for other tags. This wouldn't be immediately obvious to someone looking at the taxonomies.

john-gom:
Also, when I've been looking at storing taxonomies in ontologies, I've been designing it so that SPARQL queries can fetch a property form parents. We can do it using the preference rule above but having a global level won't work so well

stephanegigandet:
Thanks for the review @john-gom , in fact get_tag_with_parents can just return the stored parents in %all_parents, so I removed the sorting by level in that function. The sorting by global level is only useful when generating a list of parents for multiple tags (e.g. when a product is in several categories).

john-gom:
> Thanks for the review @john-gom , in fact get_tag_with_parents can just return the stored parents in %all_parents, so I removed the sorting by level in that function. The sorting by global level is only useful when generating a list of parents for multiple tags (e.g. when a product is in several categories).

Is it maybe safer just to revert the changes in `build_tags_taxonomy` and the taxonomy version number for now given that the $level is not being used for forest footprint?

stephanegigandet:
> Is it maybe safer just to revert the changes in `build_tags_taxonomy` and the taxonomy version number for now given that the $level is not being used for forest footprint?

@john-gom well the level is used indirectly, as the order of parents is incorrect, but it won't really affect the forest footprint indeed.

I made a separate PR with just the forest footprint fix: https://github.com/openfoodfacts/openfoodfacts-server/pull/14210

Once the forest footprint is merged, then we can make this PR only about the parent ordering fix.

stephanegigandet:
@john-gom can we merge this PR? Looking at the test results, the level fix make things more logical. e.g. "Strawberries, Fruits, Berries" becomes "Strawberries, Berries, Fruits".

john-gom:
I'd still like to see more documentation on the purpose of this field so that we can validate behaviour

PR #14188, opened 2026-08-03
taxonomy(ingredients): Add red and green chili paste

ginger51011:
Sometimes appear on Swedish sauces for example.


Freso (approved):
It might be better to uncomment lines 1259–1270 of `ingredients_processing.txt`. There’s a note from 5 years ago by @aleene that there’s an issue with `en:cocoa paste`, but not clear what that issue is/was, or if it’s been resolved since.

Anyway. These entries can be removed again if that ends up being done.

PR #14191, opened 2026-08-03
taxonomy(labels): Fix "råvara"

ginger51011:
This should include the variants I found here: https://se.openfoodfacts.org/facets/ingredients/r%C3%A5vara


Freso (approved):

Freso:
You might be interested in https://github.com/openfoodfacts/openfoodfacts-server/issues/13258 too :)

PR #14192, opened 2026-08-03
taxonomy: Add ingredients for popular Swedish energy/protein drink

ginger51011:
Peach juice from concentrate in Swedish is commented out, any reason why?


Freso (approved):
> Peach juice from concentrate in Swedish is commented out, any reason why?

Might be because it should be recognised from the `ingredients_processing` taxonomy, but it seems like it’s not. Ask in Slack, currently running an analysis on `ingredients.txt` to figure out where the change is from, but the file is _large_ and the repository is old, so it may take a little while. :)

PR #14194, opened 2026-08-03
taxonomy(ingredients): Add Swedish variant for pea

ginger51011:
Often found in vegan alternatives, such as here:

- https://se.openfoodfacts.org/product/7350037193477/chorizo-peas-of-heaven

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -63150,7 +63150,7 @@ sl: Grah
 sq: Bizelja
 sr: грашак
 su: Kacang polong
-sv: ärt, ärter, ärtor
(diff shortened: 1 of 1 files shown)


Freso (approved):
Might be good to add `ärtaprotein` to `en:pea protein` as well, as there are some cases where (as I understand it) the code will combine the “group” with the “members” so e.g., “oils (palm, sunflower)” will get analysed as “oils (palm oil, sunflower oil)”—but it requires these to actually exist in the taxonomy, otherwise it will fall back to the base ingredient.

I’m not sure how good the support for this is for Swedish currently, but may as well be prepared for when it is better. :)

PR #14202, opened 2026-08-04
taxonomy(labels): put 'no <substance>' before '<substance>-free'

jayaddison:
### What
There is some [guidance](https://github.com/openfoodfacts/openfoodfacts-server/blob/41177330b827160af552dc5efdcc80d1dc9bca8d/taxonomies/labels.txt#L3) at the top of the `labels.txt` taxonomy file that indicates that for multi-valued labels that indicate the _absence_ of a substance, the form `No <substance>` should appear before `<subtance>-free`.

This is the case for most of the existing cases; for example:

https://github.com/openfoodfacts/openfoodfacts-server/blob/41177330b827160af552dc5efdcc80d1dc9bca8d/taxonomies/labels.txt#L1520

(English label names: `No wheat, without wheat, Wheat-free`)

I did not use any LLMs for this pull request, neither during its preparation, nor to submit it.

### Screenshot or video
N/A

None (strictly speaking I know that an issue should be created first; I can do this if needed)

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -5266,8 +5266,8 @@ zh: 低饱和脂肪含量
 description:en: In EU, a claim that a food is low in saturated fat may only be made if the sum of saturated fatty acids and trans-fatty acids in the product does not exceed 1,5 g per 100 g for solids or 0,75 g/100 ml for liquids and in either case the sum of saturated fatty acids and trans-fatty acids must not provide more than 10 % of energy. (see: https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX%3A02006R1924-20141213)
 
 < en: Low or no saturated fat
-en: saturated fat-free
-xx: saturated fat-free
+en: No saturated fat, saturated fat-free
+xx: No saturated fat, saturated fat-free
 bg: без наситени мазнини
 ca: sense greixos saturats
 de: ohne gesättigte Fettsäuren
@@ -23737,7 +23737,7 @@ bg: Без остатъци от пестициди
 fr: Sans résidus de pesticides, sans résidu de pesticide, zéro résidus de pesticides, zéro résidu de pesticide
 
 # obligation of means: cultivated without pesticides (but can be a specific year, or after flowers bloom)
-en: Pesticide-free, without pesticide, no pesticide
+en: No pesticide, pesticide-free, without pesticide
 bg: Без пестициди
 fr: Sans pesticides, cultivé sans pesticides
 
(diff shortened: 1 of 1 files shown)


jayaddison:
As for the reason _why_ this change is being made: I think that the reason that the guidance about ordering of these label names exists is likely that `No <x>`, particularly in the situation of a translator reading the original English value before translating it, minimises ambiguity about the meaning of the label.

taxonomies/labels.txt
```diff
@@ -23737,7 +23737,7 @@ bg: Без остатъци от пестициди
 fr: Sans résidus de pesticides, sans résidu de pesticide, zéro résidus de pesticides, zéro résidu de pesticide
 
 # obligation of means: cultivated without pesticides (but can be a specific year, or after flowers bloom)
-en: Pesticide-free, without pesticide, no pesticide
+en: No pesticide, pesticide-free, without pesticide
```
Freso: “No pesticide” feels a bit awkward here. Maybe “No pesticides” would be a better phrasing? Or maybe “no pesticide” sounds perfectly fine to native English speakers.

```suggestion
en: no pesticides, pesticide-free, without pesticide, no pesticide
```

Freso (approved):
> I think that the reason that the guidance about ordering of these label names exists […]

I think it’s more related to how the order affects which is used as the canonical id, and several of the `en:no-…` are used directly in checks in the product opener code—so if the canonical identifier was changed, things are likely to break.

taxonomies/labels.txt
jayaddison: To my mind, `no pesticides` does seem slightly more typical (idiomatic?) - I think because it's indicating the absence of a range of pesticides.  But yes, no-pesticide seems acceptable too, and pesticide-free would use the singular (as you suggest).

jayaddison:
> > I think that the reason that the guidance about ordering of these label names exists […]
> 
> I think it’s more related to how the order affects which is used as the canonical id, and several of the `en:no-…` are used directly in checks in the product opener code—so if the canonical identifier was changed, things are likely to break.

Ok; would this only affect identifier references from the code, or should we worry about persistent datastructures/URLs that might also be invalidated?

Freso:
> should we worry about persistent datastructures/URLs that might also be invalidated?

This has been a topic at a couple of the recent Monday Product Opener meetings. :)

So far, the consensus seems to be that we should limit how much we break canonical ids for _no_ reason, but as long as old ids are stored as synonyms, it’s okay to do it when needed (e.g., for standardisation/normalisation, fixing things, or anything else you can make a reasonable argument for).

stephanegigandet (approved):
Thank you!

jayaddison:
Thanks @stephanegigandet @Freso!

PR #14206, opened 2026-08-05
feat(webmanifest): add F-Droid related application

Freso:
Reference: https://github.com/w3c/manifest/wiki/Platforms


stephanegigandet (approved):
Thanks

PR #14210, opened 2026-08-06
fix: also match parent ingredients for Forest Footprint 2026

stephanegigandet:
This PR contains only the Forest Footprint 2026 related changes from https://github.com/openfoodfacts/openfoodfacts-server/pull/14185 and does not contain the fix to the computation of parents..


Freso (approved):
Looks reasonable to me and, based on the updated test results, seems to work as intended. :)

PR #14216, opened 2026-08-06
fix: manifest value for prefer_related_applications: true

stephanegigandet:
https://world.openfoodfacts.org/cgi/manifest.pl currently returns:

`prefer_related_applications: null,`

### DIFF
--- a/cgi/manifest.pl
+++ b/cgi/manifest.pl
@@ -36,7 +36,7 @@ use CGI qw/:cgi :form escapeHTML/;
 use URI::Escape::XS;
 use Storable qw/dclone/;
 use Encode;
-use JSON::MaybeXS;
+use boolean;
 
 my $request_ref = ProductOpener::Display::init_request();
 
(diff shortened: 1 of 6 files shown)


cgi/manifest.pl
```diff
@@ -36,7 +36,7 @@
 use URI::Escape::XS;
 use Storable qw/dclone/;
 use Encode;
-use JSON::MaybeXS;
+use boolean;
```
Freso, line 39: Is this needed to use `true`? Or for the `->canonical->encode()` method?

Freso (approved):

Freso:
You may have to `touch` the new test file before `/update_tests_results` will actually update the file. It doesn’t add new files. (Probably an issue with it not doing any `git add`’ing so its `git commit` doesn’t pick up on new files.)

Edit: Hopefully fixed with https://github.com/openfoodfacts/openfoodfacts-server/pull/14217 :)

cgi/manifest.pl
stephanegigandet: It's for "true", canonical is to get the JSON keys sorted, to always have the same output

PR #14217, opened 2026-08-06
ci(pr_actions): `git add` new test result files

Freso:
Currently `/update_tests_results` will only update test results for already existing files. If you make a test that expects/requires a new `expected_test_results` file, it won’t get added by this command alone.

This makes the action use `git add --all` as well as limiting the file pattern to only match files that are actually under `expected_test_results` directories.


stephanegigandet (approved):
Cool, thanks!

PR #14219, opened 2026-08-06
build: reduce frontend Docker context

VaiTon:
## Summary
- add a frontend-specific Docker ignore file
- keep only frontend build inputs in the frontend context
- preserve existing exclusions for generated, runtime, and oversized legacy assets

## Validation
- frontend image build completed successfully
- BuildKit context: 180.08 MB
- baseline and slim builds produced the same image digest
- final image: 361,755,571 bytes


stephanegigandet (approved):
Thank you!

Issue #10807, opened 2024-09-18
The "Get the app" links are not working: `https://world.openproductsfacts.org/?utm_source=off&utf_medium=web&utm_campaign=search_and_links_promo_en`


teolemon:
The "Get the app" links are not working: `https://world.openproductsfacts.org/?utm_source=off&utf_medium=web&utm_campaign=search_and_links_promo_en`

_Originally posted by @teolemon in https://github.com/openfoodfacts/openfoodfacts-server/issues/10757#issuecomment-2349309480_

PR #14221, opened 2026-08-07
fix: utm_medium typo, missing heartbeat, event category casing, and a duplicated query string

UrbanFercec:
### What

Four small, independent fixes to what Matomo records for the web. **11 files, +28 / -21 hand-written, one commit per fix**, plus a fifth commit of regenerated expected results written by your own action. None of them adds an event or a dimension, and none changes anything a visitor sees.

**1. `utf_medium` is a typo for `utm_medium`** (`Config_off.pm` and the three flavour twins, `templates/web/common/site_layout.tt.html`)

The Play Store, APK, App Store and Facebook links are built with `utm_source=off&utf_medium=web`. Nothing reads `utf_medium`, so every visit arriving through those links has no medium attribution at all. One character each, and it makes the existing `utm_source` actually usable.

**2. Event category casing is inconsistent** (`html/js/display.js`, `html/js/product-multilingual.js`, `templates/web/pages/user_form/user_form.tt.js`)

`'Product'` and `'product'`, `'User'` and `'user'` are being sent for the same things. Matomo is case sensitive on category, so each pair is reported as two separate rows today and neither one is the real total.

Normalised to lower case, because that is what the rest of your web code already sends (`personal_search`, `has_nutriscore`, `signup_submit`, `delete_account`) and it is the casing the app's own categories use (`scanning`, `share`, `user management`). #8191 opens by asking that mobile and web event names match, so this is a step toward that rather than a new convention. **Only the casing changed, no event was renamed.**

⚠️ **This is the one part here with a real trade-off**, so it is a **separate commit** on purpose. Old rows stay under the old casing, so reports spanning the change need both. If you would rather standardise upward, or not at all, say so and I will drop that commit. Parts 1 and 3 do not depend on it.

**3. The heartbeat timer is missing from the site-wide snippet** (the four flavour configs, `madenearme/*.html`)

Same one-line fix as #14220, which covers only the standalone donation page. Without it, a visit that views a single page records 0 seconds, and a single product page reached from a search engine is the shape of most of your web traffic.

⚠️ **This is the part to weigh carefully, and my first estimate of its cost was too optimistic. The corrected version:** the bound is **one extra request per 15 seconds of page lifetime**, fired on every `blur` / `visibilitychange` / `unload`, not one per pageview. `setUpHeartBeat` registers those handlers and `heartBeatPingIfActivity` sends whenever 15s have passed since the last request, with no per-pageview cap. So the median pageview costs +1 request, but a tab left open and returned to repeatedly, or a phone being switched away from and back, costs more. Visits shorter than 15s and crawlers cost nothing. A realistic range is **+50% to +80% tracker requests, with +100% as a realistic ceiling rather than a hard bound**.

That matters because this covers more than one site: `$analytics` is rendered on every HTML page, so it is world OFF, the producers platform, obf, opf, opff, and madenearme.

**Your infrastructure, your call, and I cannot see the numbers that decide it** - peak tracker requests/sec, whether QueuedTracking is enabled, and whether the analytics host shares capacity with anything visitor-facing. The reassuring part is that the tracker is a separate async-loaded host, so a saturated Matomo degrades analytics without slowing the site. **If the answer is "not at that volume", say so and I will drop this part**, or narrow it to the donation and campaign pages where the 0-second problem actually costs you something.

**4. A duplicated query string on the non-`off` APK links** (`Config_obf/opf/opff.pm`)

Found while fixing part 1. `site_layout.tt.html` appends `?utm_source=...&utm_medium=web&utm_campaign=...` to `options.android_apk_app_link`. On `off` that value carries no query so the result is fine; on the other three it already ended in `?utm_source=<flavour>&utm_medium=web`, so the rendered URL had **two `?` and two `utm_medium`**, the first valued `web?utm_source=off`.

**Being precise about what this is: cosmetic, not a behaviour fix.** Matomo reads campaign parameters from the page URL, not from an outbound link's target, so neither version was ever parsed - I checked `piwik.js` rather than assume. It is malformed output on a line this PR already touches, so it is cleaned up here. No expected results change; `off` is untouched. The mirror-image fix (give `off` the same query and change the template's `?` to `&`) would be more consistent with how `android_app_link` and `ios_app_link` are built, but it changes what `off` renders, so I left that call to you.

**Commit structure, so dropping a part is a real offer and not a figure of speech:**

| Commit | Part |
|---|---|
| `fix: utm_medium typo in the app, APK and Facebook links` | 1 |
| `fix: enable the Matomo heartbeat timer on the shared snippet` | 3 |
| `fix: normalise Matomo event casing on the web` | 2 |
| `chore: drop a duplicated query string on the non-off APK links` | 4 |
| `test: Update tests results` | your bot |

Parts 2 and 4 revert cleanly on their own. Part 3 also reverts on its own, but the regenerated fixtures carry both part 1 and part 3, so dropping it means one more `/update_tests_results` round. **Say which and I will do it** - I would rather split this than have it sit.

Three scoping notes:
- The **docs site is deliberately left alone**. `docs/assets/api-rapidoc.html` and `overrides/main.html` both carry a Matomo snippet on site 16, but those two are written differently from the product ones (no `setDoNotTrack`, no `disableCookies`), so they look like their own decision rather than the same bug. Happy to do them if you want them consistent.
- The **156 translated donate pages** stay untouched on purpose. `html/donate/en.html` is the Crowdin source (fixed in #14220), so they pick the heartbeat up on the next translation sync rather than by hand.
- The `openfoodfacts-web` repo has its own copies of the `utf_medium` links. Not touched here, since it is a separate repo. Say the word and I will do that one too.

### Test results

⚠️ **11 hand-written changes, the rest regenerated.** The `utm_medium` links and the Matomo snippet both appear in the rendered page, so they are asserted by the stored expected results. Rather than hand-edit those, I used your own `/lint+update_tests_results` action so the regeneration is done by your tooling with no manual edits. Please review that bot commit as machine output rather than as my work.

Written with Claude Opus 5 (Anthropic), used agentically through Claude Code.

Being straight about what I did and did not run:

- **The heartbeat behaviour is verified on my machine.** For #14220 I served the page locally with `setTrackerUrl` repointed at a local server, so nothing was written to your Matomo, and captured the tracker requests before and after. The image below is that capture. It is the same one-line change as part 3 here.
- **The other two are string fixes** with no runtime behaviour to exercise: a parameter name, and the case of an event category.
- **I did not run your stack locally**, so I have not run the integration suite on my device. The expected results were regenerated by your own `/update_tests_results` action rather than hand-edited. ⚠️ **Be aware that this means they have not been validated by anything** - the run that wrote them is the same run that would have checked them, and `Pull Request checks` and `CodeQL` are both sitting at `action_required` waiting for a maintainer to approve workflow runs. Approving them is the real check. If you would rather I stand the stack up locally and run it properly first, say so and I will.

### Screenshot or video

Captured tracker requests before and after

- Part 2 moves toward the first line of #8191, "Ensure that Matomo event names for Mobile and the Web are the same, whenever possible".
- Part 3 is the site-wide half of #14220.
- Part 1 is adjacent to #12950, which fixed the `utm_content` language codes on the donation pages. This is the same family of bug in a different parameter.
- **Related to #10807** (deliberately not `Fixes:`). That issue reports the OPF "Get the app" link as broken, and its URL shows **three** separate defects: it resolves to the site root rather than the app page, it carries a hardcoded `utm_source=off` on an Open Products Facts page, and it misspells `utf_medium`. **This PR fixes only the third.** Closing the issue automatically would have shut a P1 that is still broken, so I have not used a closing keyword.
- The hardcoded `utm_source=off` is in `templates/web/common/site_layout.tt.html`, which stamps it on every flavour. That is a real bug and part of what #10807 is about, but fixing it changes what the `off` flavour renders and so would drag the 69 expected results through another regeneration. Happy to do it as a follow-up if you want it.
- Part of #8191.
- This came out of a wider review of what Matomo currently records on web and on the app, which I have been discussing with Pierre.

### DIFF
--- a/html/js/display.js
+++ b/html/js/display.js
@@ -40,7 +40,7 @@ function doWebShare(e) {
             console.info('Successfully sent share');
 
             if (typeof _paq !== 'undefined') {
-                _paq.push(['trackEvent', 'Product', 'Share', 'Product Page']);
+                _paq.push(['trackEvent', 'product', 'share', 'product page']);
             }
         },
         (error) => console.error('Error sharing: ' + error)
--- a/html/js/product-multilingual.js
+++ b/html/js/product-multilingual.js
@@ -654,7 +654,7 @@ const maximumRecentEntriesPerTag = 10;
 
                                     $('#' + imagefield + '_' + data.result.image.imgid).addClass("ui-selected").siblings().removeClass("ui-selected");
                                     change_image(imagefield, data.result.image.imgid);
-                                    trackMatomoEvent('Product', 'Image Upload', imagefield);
+                                    trackMatomoEvent('product', 'image upload', imagefield);
                                 }
 
                                 if (data.result.error) {
--- a/lib/ProductOpener/Config_obf.pm
+++ b/lib/ProductOpener/Config_obf.pm
@@ -188,11 +188,11 @@ $flavor = 'obf';
 	site_name => "Open Beauty Facts",
 	product_type => "beauty",
 	og_image_url => "https://static.openbeautyfacts.org/images/logos/obf-logo-vertical-white-social-media-preview.png",
-	android_apk_app_link => "https://github.com/openfoodfacts/smooth-app/releases?utm_source=obf&utf_medium=web",
+	android_apk_app_link => "https://github.com/openfoodfacts/smooth-app/releases?utm_source=obf&utm_medium=web",
 	android_app_link =>
-		"https://play.google.com/store/apps/details?id=org.openfoodfacts.scanner&utm_source=obf&utf_medium=web",
-	ios_app_link => "https://apps.apple.com/app/open-food-facts-product-scan/id588797948?utm_source=obf&utf_medium=web",
-	facebook_page_url => "https://www.facebook.com/openfoodfacts?utm_source=obf&utf_medium=web",
+		"https://play.google.com/store/apps/details?id=org.openfoodfacts.scanner&utm_source=obf&utm_medium=web",
+	ios_app_link => "https://apps.apple.com/app/open-food-facts-product-scan/id588797948?utm_source=obf&utm_medium=web",
+	facebook_page_url => "https://www.facebook.com/openfoodfacts?utm_source=obf&utm_medium=web",
 	x_account => "OpenFoodFacts",
 	# favicon HTML and images generated with https://realfavicongenerator.net/ using the SVG icon
 	favicons => <<HTML
--- a/lib/ProductOpener/Config_obf.pm
+++ b/lib/ProductOpener/Config_obf.pm
@@ -188,7 +188,7 @@ $flavor = 'obf';
 	site_name => "Open Beauty Facts",
 	product_type => "beauty",
 	og_image_url => "https://static.openbeautyfacts.org/images/logos/obf-logo-vertical-white-social-media-preview.png",
-	android_apk_app_link => "https://github.com/openfoodfacts/smooth-app/releases?utm_source=obf&utm_medium=web",
+	android_apk_app_link => "https://github.com/openfoodfacts/smooth-app/releases",
 	android_app_link =>
 		"https://play.google.com/store/apps/details?id=org.openfoodfacts.scanner&utm_source=obf&utm_medium=web",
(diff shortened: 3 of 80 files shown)


stephanegigandet (approved):
Thanks

PR #14224, opened 2026-08-07
feat: Improve producer column mapping build output

VaiTon:
## Summary
- add taxonomy-style progress, color, and success/error output to the producer column mapping generator
- fail clearly when language metadata has not been generated

## Validation
- Docker Perl syntax check passed
- targeted Docker run exposed the missing-language prerequisite and the new guard handles it clearly


stephanegigandet (approved):
Thanks

PR #14227, opened 2026-08-08
chore: added dry-residue

aleene:
### What
Added dry residue to the nutrients table for mineral waters.


stephanegigandet (approved):
Thanks

PR #14229, opened 2026-08-09
fix: move serving size before nutrition facts table (#14084)

### LINKED ISSUE #14084: The Serving size looks out of place in the nutrition knowledge panel (it could appear before the table)
### Task
The Serving size looks out of place in the nutrition knowledge panel (it could appear before the table instead)
Image

ganga-300:
## What
Fixes #14084 - Serving size appeared after the nutrition facts table instead of before it, in the nutrition knowledge panel.

## Fix
Reordered the `panel_ids` array in `templates/api/knowledge-panels/health/nutrition_panels.tt.json` so `serving_size` appears before `nutrition_facts_table`. New order: physical_activities -> serving_size -> nutrition_facts_table.

## Testing
- [x] Verified on a real product with serving size data (e.g. "Serving size: 1 scoop (34 g)") that it now displays above the nutrition facts table
- [x] Verified products without serving size data are unaffected (panel simply doesn't render, same as before)
- [x] Updated affected unit and integration test fixtures to reflect the new panel order, carefully isolating only the relevant `panel_ids` changes from unrelated staleness in those same fixture files (see note below)

## Before
Before

## After
Screenshot 2026-08-09 at 8 43 11 AM

## Notes for reviewers
While updating test fixtures, I found some pre-existing failures unrelated to this change, confirmed via `git stash` comparison against `main` (they fail identically without this fix applied):
- `tests/integration/api_v3_product_read.t`: 4 failures related to a separate `nutrition-value-very-high-for-category-*` data quality feature - not touched by this PR
- `tests/integration/data_quality_knowledge_panel.t`: 1 failure, same root cause as above
- `tests/integration/web_html.t`: failures were a local Robotoff URL environment config mismatch on my machine, not present in CI

I also found (and fixed) an unrelated "Compared to" comparison-column feature that appeared in some fixture files after running `--update-expected-results` locally - this data seems to be present in my local dev environment's output but absent in CI's, likely due to differing category stats data availability between environments. I isolated and removed it from every affected fixture so this PR only touches the actual `panel_ids` reorder.


tests/integration/expected_test_results/api_v3_product_read/get-fields-attribute-groups-all-knowledge-panels.json
```diff
@@ -2140,6 +2140,12 @@
                            "text" : "As sold<br>for 100 g / 100 ml",
                            "text_for_small_screens" : "100g",
                            "type" : "text"
+                        },
+                        {
```
stephanegigandet: I don't understand why there is an extra column with the per serving nutrition facts, your change to nutrition_panels.tt.json should not cause that

ganga-300: @stephanegigandet 
Good catch, thanks! That extra column was leftover from running `--update-expected-results` locally, which captured an unrelated "Compared to" comparison-column feature that happens to be present in my local dev environment's output but not in CI's (likely due to differing category stats data availability between environments). It had nothing to do with this fix - I've now gone through and removed it from all affected fixture files, isolating only the actual `panel_ids` reorder in each. Pushed the correction.

ganga-300:
@stephanegigandet , just a  follow-up in case my earlier reply got missed  happy to make any changes needed.

Also, the 5 failing checks all trace back to one cause unrelated to this PR: `tests/integration/api_v3_product_read.t` has 4 pre-existing sub-test failures (confirmed via `git stash` that they fail the same way on `main` too). The "Perl unit tests" check just goes red because of those, not a separate issue.

PR #14230, opened 2026-08-09
fix: vector bio image for hunger-games

teolemon:
fix: vector bio image for hunger-games
The Bio image looked ugly

image


hangy (approved):

teolemon:
thank you @hangy

PR #14239, opened 2026-08-10
fix: Forest Footprint 2026, origins with no data should get the en:unknown origin footprint value

stephanegigandet:
This PR applies the value of the en:unknown origin footprint when we have an origin for an ingredient, but no corresponding forest footprint data. (e.g. if we have a value like en:africa while we only have country values). Previously it resulted in a 0 footprint.

Also removed label_id: null when there is no label.


Freso (approved):
Seems simple and straightforward to me.

raphael0202 (commented):
Looks good to me!

PR #14240, opened 2026-08-10
ci(pr_actions): deduplicate the git commit step(s)

Freso:
It’s basically the same git commit step used for all three `/` actions, so this utilises YAML anchors and merge keys to simplify this.


stephanegigandet (approved):
Looks good, thank you

PR #14242, opened 2026-08-10
fix: preserve edit type when redirecting products

### LINKED ISSUE #14232: `cgi/product.pl?type=edit` doesn’t properly redirect between product type sites
### What
- I go to edit a product on OPF to move said product to OBF.
- I save the change.
- I click the “Edit the product again” button (`https://se.openproductsfacts.org/cgi/product.pl?type=edit&code=8710464116446`).
- I get taken to the “Add a product” page on OBF (`https://se.openbeautyfacts.org/cgi/product.pl?code=8710464116446`).

Note that “See the product page” (`https://se.openproductsfacts.org/product/8710464116446/sensodyne-haleon`) properly redirects to the OBF page (`https://se.openbeautyfacts.org/product/8710464116446/sensodyne-haleon`), so it _has_ been moved to OBF.

1. Go to any OxF site
2. Edit any product
3. Change the product type to move it to a different OxF site.
4. Click the “Edit the product again” button

### Expected behavior
- I expect to be taken to edit the product on OBF (`https://se.openbeautyfacts.org/cgi/product.pl?type=edit&code=8710464116446`).

### Why
- The current behaviour interrupts editing flow, risks confusing editors (though I guess not everyone has permission to move product types, so I guess that mitigates that somewhat), and I don’t know what happens if you try to add a product that already exists.

### Additional context
- It could probably also be argued that `https://se.openbeautyfacts.org/cgi/product.pl?code=8710464116446` giving a HTTP 200 is a bug to begin with, and should either redirect to the `/product/` page or the edit page.

- At least in the 1000s; all the products that are currently on the wrong site (someone just posted an approximation of ~5500 products on OFF alone)

### Time per product
- If further editing is needed, this adds _at least_ one more click and one more pageload before being able to edit the product again—and that’s if you know to not click the “Edit the product again” button directly.

bhavyaKhatri2703:
### What
Preserve type=edit when redirecting a product to its correct OxF site so editors are taken back to the edit page instead of the “Add a product” page.

- Fixes: #14232


cgi/product_multilingual.pl
```diff
@@ -362,9 +362,11 @@ ($product_ref)
 				and ($product_ref->{product_type} ne $options{product_type}))
 			{
 				redirect_to_url($request_ref, 302,
-						  format_subdomain($request_ref->{subdomain}, $product_ref->{product_type})
-						. '/cgi/product.pl?code='
-						. $code);
+          		          format_subdomain($request_ref->{subdomain}, $product_ref->{product_type})
+        				. '/cgi/product.pl?type='
+      					. $type
```
Freso: Do we know that `$type` is always set to something meaningful here? Have you tested this on a local test install?

Freso (commented):
Seems reasonable to me. Thank you! :) One small question/doubt.

cgi/product_multilingual.pl
bhavyaKhatri2703: `$type` can take values other than `search_or_add here, such as edit, add, delete, and show` (I’m not sure if there are other possible values).

So if someone tries to edit a product on OFF that actually belongs to OPF, it should redirect them to the edit URL on OPF. Similarly, if someone tries to delete a product on OFF that belongs to OPF, I think it should redirect them to the delete URL on OPF. That’s why I chose to preserve $type rather than hardcoding edit.

For local testing, I did:

`curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?type=edit&code=5000295157873"`

and got a 302 redirect to:

`http://world.openproductsfacts.org/cgi/product.pl?type=edit&code=5000295157873`

So the type=edit is correctly preserved by my local change.

Freso: Yes, but are we certain `$type` is always set here? Could `$type` be `null` (not `"null"`) or otherwise undefined? It seems to be set here via `local $log->context->{type} = $type;` (line 205), but I don’t know if `$log->context->{type}` is guaranteed to be set to something meaningful.

-----

What happens if you do `curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?code=5000295157873"`, `curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?type=&code=5000295157873"`, `curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?type&code=5000295157873"`, or `curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?type=thisisnotarealtype&code=5000295157873"`?

bhavyaKhatri2703: 1. No `type` : 200,  `$type` becomes `search_or_add` at `my $type = single_param('type') || 'search_or_add';` (line 190)

```bash
curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?code=2537851010837"
```

2. `type=` : 200, same as above

```bash
curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?type=&code=2537851010837"
```

3. `type` : 200, same as above

```bash
curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?type&code=2537851010837"
```

4. `type=edit` : 302, and the redirect preserves `type=edit`.

```bash
curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?type=edit&code=2537851010837"
```

5. `type=thisisnotatype` : 302, and the redirect preserves `type=thisisnotatype`.

```bash
curl -I "http://world.openfoodfacts.localhost/cgi/product.pl?type=thisisnotatype&code=2537851010837"
```

Freso (approved):

stephanegigandet (approved):
Thank you!

PR #14244, opened 2026-08-10
feat: handle “*/” in ingredient lists

Freso:
“/” can be used to denote that what follows is the ‘definition’ or ‘explanation’ of the ‘footnote’ denoted by “*”, “**”, or similar (ie., what’s in `$symbols_regexp`).

Needed-for: https://se.openfoodfacts.org/product/7350056848709/%C3%B6rtsalt-original-spicemaster


stephanegigandet (approved):
Thank you!

PR #14245, opened 2026-08-10
fix: move admins to Config2.pm

stephanegigandet:
Changing admins should not require a git change.
One consequence is that admins may be different on each flavour. Which is ok I think.


Freso (approved):
Seems simple and reasonable to me. 👍 to making the admin-map a per-instance configuration too.

teolemon (changes requested):
- Can we document admins publicly somewhere, for each flavour, before we do this ?
- Can we commit to keeping the documentation in sync with what's actually on the servers ?

Freso:
> - Can we document admins publicly somewhere, for each flavour, before we do this ?
> - Can we commit to keeping the documentation in sync with what's actually on the servers ?

Could maybe make sense to have some automated list somewhere with this information? Similar to https://musicbrainz.org/privileged or https://en.wikipedia.org/w/index.php?title=Special:ListUsers&group=sysop or similar

lib/ProductOpener/TestDefaults.pm
```diff
@@ -84,7 +84,7 @@ a user which is an admin
 %admin_user_form = (
 	%{clone(\%default_user_form)},
 	email => 'admin@openfoodfacts.org',
-	userid => 'stephane',    # has to be part of %admins
+	userid => 'admin',    # has to be part of %admins
```
Freso, line 87: Is there a `Config.pm` for the tests? Does the user need adding to `%admins` there too?

hangy:
> /update_tests_results

I don't know the snapshots should change based on the changes. Rather, we need to update the tests, as they depended on (an) specific admin user(s) being present

lib/ProductOpener/TestDefaults.pm
hangy: My idea was to add the expected user to `%admins` specifically for the test. Maybe even use random admin user names to avoid accidentally reusing a user with too many permissions in a test

PR #14247, opened 2026-08-11
taxonomy: ukrainian

svensven:
### What
A quick dive into the ukrainian unknowns.

Apparently accented vowels are optional stress markers, not different letters.

- Fixes:


stephanegigandet (approved):
Thank you!

PR #14248, opened 2026-08-11
fix: small fixes for */ and **/

stephanegigandet:
Some changes so that we don't match * in ** when the label for ** is not recognized.

### DIFF
--- a/tests/unit/expected_test_results/ingredients/en-asterisk-slash.json
+++ b/tests/unit/expected_test_results/ingredients/en-asterisk-slash.json
@@ -87,6 +87,7 @@
       },
       {
          "ciqual_proxy_food_code" : "11000",
+         "ecobalyse_code" : "12f8ffd4-cbc0-5d7b-ba0e-9a0acb3e6918",
          "id" : "en:garlic",
          "is_in_taxonomy" : 1,
          "labels" : "en:organic, en:organic",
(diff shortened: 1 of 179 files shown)


tests/unit/expected_test_results/ingredients/en-asterisk-slash.json
```diff
@@ -0,0 +1,171 @@
...
+   "ingredients_without_ecobalyse_ids_n" : 7,
+   "known_ingredients_n" : 8,
+   "labels" : "en:fair-trade",
+   "labels_hierarchy" : [
+      "en:fair-trade"
+   ],
+   "labels_lc" : "en",
+   "labels_tags" : [
+      "en:fair-trade"
+   ],
```
Freso, line 164: This should ideally be applied to the ingredient rather than the product, no?

stephanegigandet: Ideally yes, but currently we do not parse something like "Ingredient (67%) Fair Trade" correctly. It could be fixed later.

Freso (approved):

Freso:
merged via https://github.com/openfoodfacts/openfoodfacts-server/pull/14259

see https://github.com/openfoodfacts/openfoodfacts-server/pull/14259#pullrequestreview-4929526035 for what was going on here.

PR #14255, opened 2026-08-12
chore: usda part 20

aleene:
### What
reviewing an checking entries


aleene:
this is the last one for the usda legacy entries. Some small edits might be left.

stephanegigandet (approved):
Thank you

PR #14256, opened 2026-08-12
chore: remove .Rhistory file and add to .gitignore

Freso:
@aleene [You added this file](https://github.com/openfoodfacts/openfoodfacts-server/pull/13733) but two months later, it’s still empty. Do you have a reason for keeping it around?


aleene:
I do not where that comes from, either a github client- or mac-relic.

aleene (approved):

PR #14270, opened 2026-08-15
fix: Hide nutrition section from Open Beauty Facts product edit form

### LINKED ISSUE #11226: We are showing full nutrition facts with food nutrients on open beauty facts
### What
- We are showing full nutrition facts with food nutrients on open beauty facts
- We should only show a table with the ability to add fluoride, or SPF index (see related wiki page: https://wiki.openfoodfacts.org/OBF:Numerical_Values)
- Example: https://world.openbeautyfacts.org/cgi/product.pl?type=edit&code=8886467010293

Image

TaciteOFF:
- Fixes #11226

### What

- Prevent the nutrition facts section from being rendered on Open Beauty Facts product edit forms (unless nutrition facts have already been added, when transferred from OFF for example)
- Remove the "Nutrition facts to be completed", "Nutrition photo to be selected" tag for OBF
- Photos to be validated mustn't require nutrition photo
- Keep the nutrition section unchanged for Open Food Facts.
- Add unit tests covering both the food and beauty flavors.

I successfully ran:
- `make test-unit test=product_edit_form.t`
- `make checks`

### DIFF
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -1035,6 +1035,10 @@ CSS
 
 	$template_data_ref_display->{nutrition_checked} = $checked;
 
+	# The nutrition feature is computed for the product and not for the site, as the producers
+	# platform can host products of different types
+	$template_data_ref_display->{nutrition_feature_enabled} = feature_enabled("nutrition", $product_ref) ? 1 : 0;
+
 	# For product types for which the nutrition feature is disabled (e.g. beauty products),
 	# the nutrition section is normally hidden in the edit form.
 	# We still display it if the product already has nutrition facts (or the "no nutrition data
--- a/cgi/product_multilingual.pl
+++ b/cgi/product_multilingual.pl
@@ -1034,6 +1034,15 @@ CSS
 	}
 
 	$template_data_ref_display->{nutrition_checked} = $checked;
+
+	# For product types for which the nutrition feature is disabled (e.g. beauty products),
+	# the nutrition section is normally hidden in the edit form.
+	# We still display it if the product already has nutrition facts (or the "no nutrition data
+	# on packaging" checkbox is set), so that contributors can delete them.
+	$template_data_ref_display->{product_has_nutrition_data}
+		= ((has_no_nutrition_data_on_packaging($product_ref)) or (has_non_estimated_nutrition_data($product_ref)))
+		? 1
+		: 0;
 	$template_data_ref_display->{display_tab_ingredients_image}
 		= display_input_tabs($product_ref, "ingredients_image", $product_ref->{sorted_langs},
 		\%Langs, \@ingredients_fields, $request_ref);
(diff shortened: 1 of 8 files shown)


TaciteOFF:
Nutrition facts already added to current products could be mass erased.

TaciteOFF:
The migration to nutrition.input_sets is skipped for OBF (ProductSchemaChanges.pm), so older data stays in the nutriments field. This PR does not detect it, so it cannot be deleted from the edit form.

Numbers from a local copy of the OBF database:

┌───────────────────────────────────────────┬──────────┐
│                                           │ products │
├───────────────────────────────────────────┼──────────┤
│ nutriments without nutrition.input_sets   │ 53,971   │
├───────────────────────────────────────────┼──────────┤
│ … with real values                        │ 1,026    │
├───────────────────────────────────────────┼──────────┤
│ … with only computed or estimated values  │ 52,945   │
├───────────────────────────────────────────┼──────────┤
│ nutrition.input_sets (handled by this PR) │ 1,592    │
└───────────────────────────────────────────┴──────────┘

Most of those products only have computed values (mostly fruits-vegetables-nuts-estimate-from-ingredients_100g), so the detection must ignore computed values. Otherwise the nutrition section would be shown on ~53,000 products that have nothing to delete.

This old data is almost never displayed anymore, but it is still in the API and in the data exports.

Possible follow-up: detect real values in nutriments and delete them together with the input sets. A one-off script would also be needed, as nobody is going to edit 1,026 products by hand.

Freso:
Some OBF (and OPF) products could be O(P)FF products that have been misplaced, so I’m a bit wary about just cleaning them out. How about a data quality facet for “nutrition information on non-(pet )food product”? This would probably require that products with non-calculated nutrition info still have this editable on OB/PF though, so that editors could delete this information if its not a misplaced product.

TaciteOFF:
> Some OBF (and OPF) products could be O(P)FF products that have been misplaced, so I’m a bit wary about just cleaning them out

This PR already goes in the direction you describe: on product types without the nutrition feature, the nutrition section is hidden unless the product has nutrition data that is not estimated. In that case the section is shown, and the editor can delete the data (by checking "no nutrition data"). Nothing is deleted automatically, and estimated values (computed from the ingredients) are ignored, so they never trigger the section.

Some numbers from the OBF MongoDB production dump (15 August, 73,093 products, obsolete products not included), which I think support your idea:
| | products |
|---|---|
| with `nutrition.input_sets` (detected by this PR) | 1,592 |
| with real values in the old `nutriments` field only | 1,026 |
| with only computed or estimated values in `nutriments` | 52,945 |

Of the 1,026 products with real old values: 118 have food-like categories (possibly misplaced products), 640 have no category at all, and the rest look like real cosmetics (hand soap, hair colour, gloss…). So there is definitely something to review, but it is a small set, and it is a mix.

A data quality facet sounds like a good way to make them visible. There is already a similar check for en:cosmetic-product in DataQualityCommon.pm, which runs for all flavors, so a check like "nutrition data on a product type without nutrition" should be easy to add. It would need to look at both the old and the new schema.

stephanegigandet (approved):
Looks good, thank you

alexgarel:
Thanks @TaciteOFF !

PR #14271, opened 2026-08-15
fix(docker): support multi-instance deployment for OBF and custom flavors

### LINKED ISSUE #14269: OBF multi-deployment guide uses the wrong host port variable and an unavailable Compose profile
The Open Beauty Facts example in `docs/dev/how-to-develop-using-docker.md` does not match the current Docker Compose configuration.

There seem to be two separate issues:

1. The guide changes `PRODUCT_OPENER_PORT` from 80 to 81, but
   `docker/dev.yml` also publishes `PRODUCT_OPENER_HOST_PORT`, which remains
   set to 80 in the copied `.env` file.

2. The guide sets `COMPOSE_PROFILES=obf`, but `docker compose config --profiles`
   currently only reports the `off` profile. `postgres` and `redis-listener`
   are assigned to that `off` profile.

3. The default Docker network name is hard-coded as `product-opener` in
   `docker-compose.yml`. As a result, the OFF and OBF Compose projects join
   the same internal network even though they have different
   `COMPOSE_PROJECT_NAME` values. Both projects then register identical
   network aliases such as `backend`, `frontend`, `postgres` and `memcached`,
   making service-name resolution ambiguous between the two deployments.

4. The Docker configuration still selects the `off` MongoDB database for an
   OBF flavor. `Config2_docker.pm` sets:
   ```perl
   $mongodb = $producers_platform ? "off-pro" : "off";
and Config_obf.pm reuses that value instead of selecting the obf
   database. Therefore, an instance configured with
   PRODUCT_OPENER_FLAVOR=openbeautyfacts and
   PRODUCT_OPENER_FLAVOR_SHORT=obf still queries MongoDB database off.

The only available Compose profile is off.

postgres and redis-listener are not included with COMPOSE_PROFILES=obf.

The frontend configuration contains both:
published: "81"
published: "80"

Starting OBF alongside OFF therefore still attempts to bind host port 80.

**Suggested changes :** 
1. Add PRODUCT_OPENER_HOST_PORT=81 to the documented OBF configuration.
2. Clarify whether OBF should keep COMPOSE_PROFILES=off, or add obf to the profiles of services required by every public flavor.
3. Consider updating env/env.obf so it also overrides the domain, host port and Minion queue required to run OBF alongside OFF.
4. Make the default Product Opener network name configurable or project-scoped, for example:
  ```yaml
  networks:
    default:
      name: ${PRODUCT_OPENER_NETWORK:-product-opener}
```
5. Select the MongoDB database from the configured flavor, or introduce an explicit environment variable such as `MONGODB_DATABASE`. For example, `PRODUCT_OPENER_FLAVOR_SHORT=obf` should result in `mongodb=obf`.

### PR DESCRIPTION
**Pull request by pranavchoudhary-tech**
Fixes #14269.

### Summary
This PR resolves issues when running Open Beauty Facts (OBF) or custom product opener flavors alongside Open Food Facts (OFF) using Docker Compose:

1. **Docker Network Isolation**: Updated default network name in `docker-compose.yml` to `${PRODUCT_OPENER_NETWORK:-product-opener}` to prevent container alias collisions across multiple deployments.
2. **MongoDB Database Selection**: Updated `lib/ProductOpener/Config2_docker.pm` to dynamically select `$mongodb` based on `MONGODB_DATABASE` or `PRODUCT_OPENER_FLAVOR_SHORT` (e.g. `obf`).
3. **Documentation**: Updated `docs/dev/how-to-develop-using-docker.md` OBF multi-deployment example to include `PRODUCT_OPENER_HOST_PORT=81` and `PRODUCT_OPENER_NETWORK=po_obf_network`.

### CODE REVIEW

**Comment by TaciteOFF** [comment]
Thanks for this PR! It fixes the incorrect Compose profile, ports, main network name, and MongoDB database selection.

However, two issues remain before OFF and OBF can safely run together:

Only the default network is isolated. Both instances still share:
off_shared_network
minion_db
Product image, user, organization, and export volumes

The shared-services network should also be different for each instance. MongoDB, Redis, and Keycloak must then be connected to both networks. Otherwise, both stacks may expose a service named backend on the same network, and OBF could reach the OFF backend.

MONGODB_DATABASE is read by Config2_docker.pm, but Docker Compose does not pass it to the containers.

**Comment by pranavchoudhary-tech** [comment]
Thanks @TaciteOFF! I have updated docker-compose.yml to pass MONGODB_DATABASE down to the containers, and updated the OBF .env.obf documentation example to include MONGODB_DATABASE=obf.

File: lib/ProductOpener/Config2_docker.pm [hunk 1]
```diff
@@ -107,7 +107,8 @@ $sftp_root = "/mnt/podata/sftp";
 $geolite2_path = $ENV{GEOLITE2_PATH};
 
 $mongodb_host = $ENV{MONGODB_HOST} || "mongodb";
-$mongodb = $producers_platform ? "off-pro" : "off";
+my $flavor_short = $ENV{PRODUCT_OPENER_FLAVOR_SHORT} || "off";
+$mongodb = $ENV{MONGODB_DATABASE} || ($producers_platform ? "${flavor_short}-pro" : $flavor_short);
```
stephanegigandet, line 112: Note that in production we currently have only one producter platform, for off-pro. We probably should add a comment about that.

pranavchoudhary-tech: Done, I've added a comment to clarify this.

alexgarel (approved):
hi @pranavchoudhary-tech waow thank you so much for helping making progress on that !

I'll approve this PR for it goes in the right direction. I'm not sure this is the end of it though.

For example one next step might be to be able to download example data based on the flavor (see the Makefile).

alexgarel:
@TaciteOFF if you have the time to test that it works, we could close the issue https://github.com/openfoodfacts/openfoodfacts-server/issues/14269

TaciteOFF:
Thanks, I agree that this PR is a good step, but some follow-up work is still needed.

When a product is changed from “food” to “beauty”, the same images and history must remain available.

The important parts to keep separate are:
- the OFF and OBF frontend/backend services;
- the MongoDB databases (off and obf);
- the Minion queues;
- Docker names such as backend and postgres, so they do not resolve to the wrong container.

So I think separate follow-up issues would be useful for service isolation, Minion queues, and flavor-specific sample data (OBF).

PR #14272, opened 2026-08-15
fix: #1587 week and month name double utf8 encoding

### LINKED ISSUE #1587: https://ru.openfoodfacts.org/products has a string in the wrong encoding
1. Go to https://ru.openfoodfacts.org/product-count
2. Hover over the graph.
In the tooltip, the first line with a date over "Продукты: 1234" is prefixed with a string in an incorrect encoding, that appears a little differently sometimes.

Image

svensven:
### What
The product count weekday names were being double-utf8-encoded.

https://ie-ga.openfoodfacts.org/product-count
https://ru.openfoodfacts.org/product-count
https://jp.openfoodfacts.org/product-count

VSCode Copilot using Claude Sonnet 4.6.
Before running `make build_lang_test`, the new tests in `make test-unit test=lang.t` fail. Afterwards, they pass.

### Screenshot or video

image
image

image
image

- Fixes: #1587

### DIFF
--- a/lib/ProductOpener/Lang.pm
+++ b/lib/ProductOpener/Lang.pm
@@ -479,7 +479,7 @@ sub build_lang ($Languages_ref) {
 				DateTime->new(year => 2000, time_zone => 'UTC', month => $month, locale => $locale)->month_name;
 		}
 
-		$Lang{months}{$l} = encode_json(\@months);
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Thank you!

PR #14275, opened 2026-08-16
fix: don't add "Missing serving size" tag after deleting per-serving values

svensven:
### What

When unticking the "per serving" box on a record that incorrectly had it ticked and per-serving nutrition values, the data structure could (according to copilot) still have an empty set for the per-serving data, and incorrectly cause the nutrition-data-per-serving-missing-serving-size quality error.

https://uk.openfoodfacts.org/product/5000157156853/heinz-garlic-mayo

VSCode Copilot auto model mode - GPT-5.3-Codex, GPT-5.4.

### Screenshot or video

image

image

- Fixes:

### DIFF
--- a/lib/ProductOpener/DataQualityFood.pm
+++ b/lib/ProductOpener/DataQualityFood.pm
@@ -926,7 +926,8 @@ sub check_nutrition_data ($product_ref) {
 				}
 			}
 
-			if ($per eq "serving") {
+			# Only consider per-serving nutrition present when the set still has nutrient values.
+			if (($per eq "serving") and (defined $set_ref->{nutrients}) and (scalar keys %{$set_ref->{nutrients}})) {
 				$has_nutrition_per_serving = 1;
 			}
 		}
--- a/tests/unit/dataqualityfood.t
+++ b/tests/unit/dataqualityfood.t
@@ -832,6 +832,65 @@ check_quality_and_test_product_has_quality_tag(
 	'serving size cannot be parsed', 0
 );
 
+# stale per-serving set without nutrient values should not trigger missing serving size
+# and per-100g nutrient values should still be validated
+$product_ref = {
+	nutrition => {
+		input_sets => [
+			{
+				source => "producer",
+				preparation => "as_sold",
+				per => "serving",
+				nutrients => {}
+			},
+			{
+				source => "producer",
+				preparation => "as_sold",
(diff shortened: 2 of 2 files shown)


TaciteOFF:
The fix works. I would merge it. But it only fixes part of the problem, and there are a few small things to clean up.
I ran the unit tests on my machine.

Nutrition values are stored in "input sets". One set is "per 100g", another is "per serving", and so on. There is already code that deletes empty sets (`Nutrition.pm:800`). But it keeps a set if it has `unspecified_nutrients` in it. That is the leftover set this PR is dealing with. So far so good.

But in `Nutrition.pm:1311`, when a user saves the form, the values are only written into the set for the option they picked right now. **The old set is never deleted.** It just stays there forever.

1. Start with a product that has fat = 3 per serving, and no serving size.
2. The user switches the setting to "per 100g" and types fat = 10. They do not clear the old fields.
3. Result:

```
input_sets = [ {per: 100g, fat: 10}, {per: serving, fat: 3} ]   <-- old set still here
error tags  = [en:nutrition-data-per-serving-missing-serving-size]   <-- still there
```

So with this PR merged, that user still gets the wrong error.

This PR fixes the case where the leftover set is empty. It does not fix the case where the leftover set still has old values in it. That second case comes from the older app/API path (`product_jqm_multilingual.pl`). The newer web form is fine, because it sends empty values for every option.

There is another piece of code, `add_misc_tags_for_input_nutrition_data_pers` (`Nutrition.pm:2889`), that does not check whether a set is empty. So the leftover empty per-serving set still adds the tag `en:nutrition-data-per-serving` to the product.

After this PR, the same product will say:

- "this product has per-serving data" (in the misc tags and facets)
- "this product does not have per-serving data" (in the data quality check)

**The most useful test is missing.** The first new test uses `nutrients => {}` (an empty list). That shape is never actually saved to the database, because `remove_empty_nutrient_values_and_set_unspecified_nutrients` removes the key. The shape that really happens is a set with only `unspecified_nutrients` and no `nutrients` key at all. That is the one this fix is really for, and it is not covered by a test.

svensven:
I don't know how closely the api json reflects the underlying data, but I also noticed there's still a _serving value for nova-group. could that be part of the problem?
https://uk.openfoodfacts.org/api/v2/product/5000157156853.json
image

PR #14288, opened 2026-08-18
fix: prompt for category before secondhand options

### LINKED ISSUE #11110: We should not display the Donation / 2nd Hand options on empty category
### What
- We should not display the Donation / 2nd Hand options on empty category

Image
Image
Image
Image

seanwessmith:
## What

- show an `Add a category` action in the Secondhand card when a product has no categories
- keep Donation and Used product options hidden until a category is available
- cover missing, empty, and populated `categories_tags` paths with a regression test

## Why

Donation and Used product links need a category to build a relevant search. The action explains why those options are unavailable and takes contributors directly to the Categories field.

## Validation

- `make test-unit test=secondhand_knowledge_panels.t`
- `make check_perltidy TO_CHECK='lib/ProductOpener/KnowledgePanels.pm tests/unit/secondhand_knowledge_panels.t'`
- `make check_critic TO_CHECK='lib/ProductOpener/KnowledgePanels.pm tests/unit/secondhand_knowledge_panels.t'`
- `make check_translations`
- `git diff --check`

Fixes #11110


teolemon:
@seanwessmith the extra mile would be to add a nudge to add a category to get 2nd hand options. Something like "Add a category to discover 2nd hand options", using the Knowledge Panel Actions syntax present in other panels.

PR #14293, opened 2026-08-18
taxonomy: grated carrots

moon-rabbitOFF:
The categories "unseasoned grated carrots" and "seasoned grated carrots" were created both under "grated carrots" to avoid people putting grated carrots with sauce under carrots. Someone removed "seasoned grated carrots" from "grated carrots" and put "grated carrots" directly under carrots. With this new configuration, "unseasoned grated carrots" became a synonym of "grated carrots" and "seasoned grated carrots" of "carrot salads".

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -86134,39 +86134,32 @@ en: Prepared sweet potatoes
 nl: Geprepareerde zoete aardappelen
 
 < en: Carrots
-en: Grated carrots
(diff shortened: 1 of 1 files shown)


Freso (commented):
I’m not sure why it’s better to do things this way, rather than going back to how it was before… ?

PR looks fine otherwise, just not sure it’s better to double down on this approach rather than going back to the previous one. :)

teolemon:
image

PR #14302, opened 2026-08-20
feat: Add more columns to the Foodture extract

john-gom:
### What

Adds scan data to the Foodture extract.

Fetches product data directly from folders rather than use the API to avoid rate limiting and so we can also fetch scan data, so removed the need for caching

### DIFF
--- a/projects/foodture/extract_data_for_representative_products.pl
+++ b/projects/foodture/extract_data_for_representative_products.pl
@@ -34,7 +34,6 @@ use LWP::UserAgent;
 use JSON::MaybeXS;
 use Text::CSV;
 
-
 # This script:
 # - reads a list of categories (and other columns) from projects/foodture/foodture_foodex2_categories_mapping_to_off.csv
 # - reads a list of most scanned products by country and category from projects/foodture/ranked_products_202602231414.csv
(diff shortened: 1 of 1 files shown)


projects/foodture/extract_data_for_representative_products.pl
```diff
@@ -396,21 +390,27 @@ sub collect_ingredients {
...
-				$l1, $l2,
-				$l7, $segmentation,
-				$category_tag, $batch1,
-				$countries_names{$country_tag} // '', $country_tag,
-				$exists_in_taxonomy || 0, $categories_agb{$category_tag} || '',
-				$categories_agb_proxy{$category_tag} || ''
-			);
-			push @row, ('') x (scalar(@hdr) - scalar(@row));
-			$csv_out->print($OUT, \@row);
+			# my @row = (
```
stephanegigandet, line 404: Why is it commented out?

stephanegigandet (approved):
Thank you! Just one question about not outputting rows for which we didn't find products. The idea was to give back exactly the same rows we got from Foodture.

projects/foodture/extract_data_for_representative_products.pl
john-gom: Just for testing on local data so I can easily see the rows that have matches to the sample products

PR #14306, opened 2026-08-20
feat: use Ciqual table 2025 version in addition to 2020

stephanegigandet:
We currently use the 2020 Ciqual table in Product Opener. There is an updated 2025 table.

This PR adds the 2025 table, and we still use the 2020 table as fallback. This is because some Ciqual codes from 2020 do not exist in the Ciqual table 2025.

Total entries in 2020: 3185
Total entries in 2025: 3484

Only in 2020: 182
Only in 2025: 481
In both versions: 3003


external-data/ciqual/ciqual/CIQUAL2020_FR_2020_07_07.xls
Freso, line 1: Do we need to keep both the source .xls and derived .csv’s around? If so, maybe we could store the .xls files in [git-lfs storage](https://git-lfs.com/) rather than directly in the git repository itself? (The repository is already rather big even without adding (more) MB-size binary blobs to it…)

stephanegigandet: We can remove the xls files

alexgarel (approved):
Great ! Thanks

PR #14309, opened 2026-08-21
fix: fixes for labels in Forest Footprint 2026

stephanegigandet:
This PR adds labels that need to be taken into account for the new Forest Footprint.

### DIFF
--- a/taxonomies/labels.txt
+++ b/taxonomies/labels.txt
@@ -1243,8 +1243,12 @@ fr: Non issu du commerce équitable
 incompatible_with:en: labels:en:fair-trade
 opposite:en: en:fair-trade
 
-# Should not be translated, "Fairtrade International" is a label
+# Should not be translated
+< en: Fair trade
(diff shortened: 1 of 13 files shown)


taxonomies/labels.txt
```diff
@@ -6335,6 +6339,10 @@ xx: NPOP-NAB-0041
 < en: organic
 fr: Bio équitable en France
 
+# "Équitable en France" is a logo that seems to be only on Biocoop products
+< en: fair trade
+fr: Équitable en France
```
Freso, line 6344: are these two related? should `Bio équitable en France` be a child of `Équitable en France`?

Freso (approved):
i've mostly looked at `labels.txt`, but that part looks fine :)

taxonomies/labels.txt
stephanegigandet: I don't think they are related

PR #14311, opened 2026-08-21
fix: support allergen markup in compound words

### LINKED ISSUE #179: Text formatting for allergens doesn't work properly
In some languages, what we target as allergen is included in a full word. 

For example, in dutch "lécithine de soja" (written lécithine de _soja_ in OFF) is written "sojalecithine".
However, our current system doesn't put in bold & doesn't flag allergens when there's no space before and after the underscore character. 

```
_soja_lecithine will not work
_soja_ lecithine will work
```

You can find a live example here : http://world.openfoodfacts.org/product/8718265523944/chocchip-cookies-nougatelli-albert-heijn

TaciteOFF:
## Summary

- support underscore allergen markup **at the start of** compound words in every language
- remove the German-only workaround
- add focused regression tests for Dutch `_soja_lecithine` and Swedish `_vete_mjöl`

Fixes #179

## Testing

- `yath test tests/unit/allergens.t` (42 assertions)
- `perltidy --assert-tidy` on the two changed Perl files
- `git diff --check`


stephanegigandet (approved):
Thank you!

PR #14313, opened 2026-08-21
feat: add misc tags for the number of uploaded images

### LINKED ISSUE #302: A new filter idea: pictures count
It's already possible to filter the database with the number of ingredients or additives.

Doing the same for pictures could be useful to track new products that can be _easily_ completed (ex: 3 pictures sent + state "pictures to be validated")

TaciteOFF:
### What

- Add searchable `misc` tags to filter products by the number of uploaded photos, e.g. `/misc/number-of-uploaded-images-1-to-3`.
- One exclusive range tag per product, computed in `compute_completeness_and_missing_tags()`: `1-to-3`, `4-to-6`, `7-to-10`, `11-to-50`, `51-to-100`, `more-than-100`. Previous `en:number-of-uploaded-images-*` tags are removed before adding the new one, so the tag stays correct when images are added or deleted.
- Add `en`/`fr` entries for the 6 tags to the misc taxonomy (other languages to come via Crowdin).
- Unit tests cover all bucket boundaries, removal of stale tags, and preservation of unrelated misc tags.

_Existing products get the tag when they are next saved. To backfill all products, update_all_products.pl must be run with --compute-history; --analyze-and-enrich-product-data alone does not compute this tag._

- Fixes: #302

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -1836,13 +1836,43 @@ sub compute_completeness_and_missing_tags ($product_ref, $current_ref, $previous
 	my @states_tags = ();
 
 	# Images
+	my $uploaded_images_n
+		= defined $current_ref->{uploaded_images} ? scalar keys %{$current_ref->{uploaded_images}} : 0;
+
+	# Add misc tags for ranges of uploaded images. Unlike states tags, misc tags are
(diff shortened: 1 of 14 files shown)


stephanegigandet (approved):
Looks good, thanks

PR #14315, opened 2026-08-21
taxonomy(additives): Add Swedish variant of E425(i)

ginger51011:
I think this producer adds a binding 's' here for some reason.

Needed-for: https://se.openfoodfacts.org/product/7350037192548/chorizo-peas-of-heaven


TaciteOFF (approved):

Freso (approved):
The -s- form as actually the more correct one [in this case](https://svenska.se/?q=konjak&exactMatch=true&activeTab=alla) :)

PR #14317, opened 2026-08-22
feat(lint_taxonomy): catch “comment” typos

Freso:
I’ve seen `comment` taxonomy properties being typo’d a few times by now, sometimes even making it past review. It would be nice to catch them when linting. :)

This looks at property names between 6 and 8 characters (in case a character was added (“commentt”) or missed (“coment”) by mistake) and if it consists of the same letters as “comment” but doesn’t equal `comment`, then push a warning to notify about this.

As of this commit, there are 61 lines matching the regular expression, 0 of which are not “comment”:
```sh
$ rg -e '^[coment]{6,8}:' taxonomies/*.txt|wc -l
61
$ rg -I -e '^[coment]{6,8}:' taxonomies/*.txt|rg -v -e '^comment:' -|wc -l
0
```

### DIFF
--- a/scripts/taxonomies/lint_taxonomy.pl
+++ b/scripts/taxonomies/lint_taxonomy.pl
@@ -210,6 +210,22 @@ sub iter_taxonomy_entries ($lines_iter) {
 						}
 					);
 				}
+				# detect “comment” typos
+				if ($prop =~ /^[coment]{6,8}$/ && $prop ne "comment") {
+					push(
+						@errors,
+						{
+							severity => "Warning",
+							type => "Correctness",
+							line => $line_num,
+							message => (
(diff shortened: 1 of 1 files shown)


stephanegigandet (approved):

alexgarel:
Thanks @Freso !

Naruyoko:
Interestingly simple approach. How does it compare to more complex algorithms, such as [String::Approx](https://metacpan.org/pod/String::Approx) and [Text::Levenshtein::Damerau](https://metacpan.org/pod/Text::Levenshtein::Damerau)?

Freso:
No idea, but definitely something that could be useful to improve/expand on it! I want to expand this to catch more typos of known properties, and the current simple regex approach is not really scalable for that. Something like those two modules might be! :)

PR #14322, opened 2026-08-22
fix: render EAN barcodes with the correct format

TaciteOFF:
## Summary

- render 13-digit product codes as EAN-13;
- render 8-digit product codes as EAN-8;
- keep Code 128 as the fallback for non-standard or internal product codes;
- update the affected HTML integration snapshots.

## Context

The product page currently uses JsBarcode with the `auto` format. JsBarcode does not detect EAN formats automatically: when Code 128 is available, `auto` selects it. As a result, a product such as [3600542656733](https://fr.openbeautyfacts.org/produit/3600542656733) is labelled as EAN-13 but displayed as Code 128.

The barcode is generated as an SVG in the browser, so this change does not require any product recomputation or data migration.

## Tests

- `git diff --check`
- verified that JsBarcode 3.12.3 accepts `3600542656733` as EAN-13 and produces the expected 95-module encoding;
- updated the expected HTML integration results.

### DIFF
--- a/templates/web/pages/product/product_page.tt.html
+++ b/templates/web/pages/product/product_page.tt.html
@@ -180,7 +180,7 @@
                                 <!-- Display with SVG if not mobile -->
                                 <div id="barcode_div_svg">
                                     <svg id="barcode_svg"
-                                        jsbarcode-format="auto"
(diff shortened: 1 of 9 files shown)


Freso (approved):
Looks good to me. Thanks!

PR #14325, opened 2026-08-23
chore(nutrition): add some czech translations

paper42:
I have checked some similar PRs and they don't include a link to an issue, if it's needed, I can add it, but for a simple translation, I don't see much point.

### What

I scanned some vegan products in the app in Czechia and noticed some ingredients are not being detected, they exist in the file, just translations were missing.

none

### Screenshot or video

-

- Fixes:

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -819,7 +819,7 @@ fr: mono- et diglycérides d'acides gras et stéaroyl-2-lactylate de sodium
 en: nigari, bittern
 bg: Нигари
 ca: Nigari
-cs: nigari
+cs: nigari, mořská jedlá sůl nigari, mořská sůl nigari
 de: Nigari
 es: Nigari
 fi: nigari
@@ -13902,7 +13902,7 @@ vegetarian:en: yes
 # <en:rice
 en: rice starch
 bg: оризово нишесте
-cs: Rýžová škrob
+cs: rýžový škrob
 da: risstivelse
 de: Reisstärke
 es: almidon de arroz, amido de arroz
@@ -17111,6 +17111,7 @@ wikipedia:en: https://en.wikipedia.org/wiki/Shea_butter
 # < en:shea why is this here?
 < en: vegetable oil
 en: shea oil, sheanut oil
+cs: bambucký olej
 da: sheaolie
 de: Sheaöl
 hr: shea ulje
@@ -18158,7 +18159,7 @@ it: aceto bianco distillato
 en: wine vinegar
 bg: винен оцет
 ca: Vinagre de vi
-cs: vinný ocet
+cs: vinný ocet, ocet kvasný vinný
 da: Vineddike
 de: Weinessig
 es: vinagre de vino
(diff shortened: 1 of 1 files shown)


taxonomies/food/ingredients.txt
```diff
@@ -13902,7 +13902,7 @@ vegetarian:en: yes
 # <en:rice
 en: rice starch
 bg: оризово нишесте
-cs: Rýžová škrob
+cs: rýžový škrob
```
paper42, line 13905: I replaced the previous formulation here because this form is never valid gramatically

taxonomies/food/ingredients.txt
```diff
@@ -41846,7 +41850,7 @@ it: estratto naturale di limone, estratto naturale di limoni
 en: concentrated lemon juice, Lemon juice concentrate
 bg: концентрат от лимонов сок, концентриран сок от лимони, концентриран лимонов сок
 ca: Suc de llimona concentrat, Concentrat de suc de llimona
-cs: koncentrovaná citrónová šťáva, citronová šťáva z koncentrátu
```
paper42, line 41849: I removed citronová šťáva z koncentrátu because that's lemon juice from concentrate - the following item, not this the concentrate itself

taxonomies/food/ingredients.txt
```diff
@@ -17118,6 +17118,7 @@ it: olio di karité
 pl: olej shea, olej roślinny shea
 ro: ulei de shea
 sv: sheaolja
+cd: bambucký olej
```
Freso: languages should generally be sorted alphabetically (`en`, `xx` are handled differently, and _sometimes_ other languages), so `cs` should go before `da` :)

Freso (approved):
Looks good, thank you!

taxonomies/food/ingredients.txt
Freso: Never valid on its own, or never-never? Would it be valid if other words were before and/or after it (e.g., “[organic] rýžová škrob [from somewhere]”)? Is it a matter of plural form where only the noun and not the adjective has been pluralised?

taxonomies/food/ingredients.txt
paper42: thanks, fixed :)

taxonomies/food/ingredients.txt
paper42: these two words together would never be valid in any context, it would be valid only if we changed the second word and added something - eg. `rýžová škrobová mouka` - `rice starch flour`, it would then be an entirely different ingredient and I don't think such an ingredient exists + we are changing the second word from noun to adjective

paper42: it's about the grammatical gender of the word - škrob (starch) is masculine in czech, so the ending of the adjective has to be in a masculine form -ový, but -ová is a feminine ending

paper42:
I had one more added line in the wrong order, should be fixed now

stephanegigandet (approved):
Thank you very much @paper42 !

PR #14326, opened 2026-08-23
taxonomy(food): chervil/cicely edits

Freso:
English “chervil” may more-or-less exclusively refer to garden chervils, but it seems like other languages use “chervil” interchangeably for multiple different plants/herbs/spices, so trying to make this clear here with some reorganising of the categories (and ingredients).

- Splits out some of the combined chervil categories into discrete ones.
- Adds new cicely and cicely powder categories.
- Adds translations (incl. from Wikidata).
- Updates some `ciqual` properties.
  - Looks like it previously used “fresh” where it now uses “raw”, so moved the non-proxy up a level.
- Cleans up some comments.
- Removes “fresh chervil” ingredient.
  - As far as I can tell, there are 0 products with it.
- Splits out “garden chervil” ingredient.

Needed-for: https://se.openfoodfacts.org/product/7310394002362/k%C3%B6rvel-kockens

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -74554,7 +74554,7 @@ wikidata:en: Q139678670
 en: Chervils and their products
 da: Kørvel og kørvelprodukter
 es: Productos del perifollo
-fr: Produits de le cerfeuil
+fr: Produits de cerfeuil
 nl: Kervelproducten
 sv: Körvel och körvelprodukter
 agribalyse_proxy_food_code:en: 11002
@@ -74606,7 +74606,7 @@ nl: Kervelbladeren
 
 < en: Chervils and their products
 en: Garden chervils and their products, Garden chervil products
-fr: Produits de le cerfeuil commun
+fr: Produits de cerfeuil commun
 
 < en: Chervils
 < en: Garden chervils and their products
@@ -74714,7 +74714,7 @@ de: Süßdolde
 et: Mesiputk
 fa: چتردار شیرین
 fi: Saksankirveli
-fr: Cerfeuil musque, Cerfeuil anisé, Cerfeuil d'Espagne
+fr: Cerfeuil musqué, Cerfeuil anisé, Cerfeuil d'Espagne
 ga: Lus áinleoige
 hr: Čehulja
 id: Seledri melur
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -74551,46 +74551,198 @@ ciqual_proxy_food_name:en: Caraway, seed
 wikidata:en: Q139678670
 
 < en: Plant-based foods
-en: Fresh chervil, Garden chervil products, Garden chervil, Chervil
-bg: Керевиз
-de: Kerbel
-es: Productos del perifollo, Perifollo, Cerefolio, Perifolio
-fr: Cerfeuil, Cerfeuil frais, Produits de le cerfeuil commun, Cerfeuil commun
-hr: Prave krasuljice, čerfoli
-ja: チャービル
-la: Anthriscus cerefolium
+en: Chervils and their products
+da: Kørvel og kørvelprodukter
+es: Productos del perifollo
+fr: Produits de le cerfeuil
 nl: Kervelproducten
-pl: Trybula ogrodowa
+sv: Körvel och körvelprodukter
+agribalyse_proxy_food_code:en: 11002
+ciqual_proxy_food_code:en: 11002
+ciqual_proxy_food_name:en: Chervil, raw
+comment:en: While “chervil” in English pretty much exclusively refers to the garden chervil variety, other languages have multiple different kinds of “chervils”.
+
+< en: Chervils and their products
+en: Chervils, Chervil
+bg: Керевиз
+da: Kørvel
+de: Kerbel
+es: Perifollo, Cerefolio, Perifolio
+fr: Cerfeuil
+hr: Krasuljica, Čerfoli
+ja: チャービル
+pl: Trybuli, Trybula
+sv: Körvel
 agribalyse_food_code:en: 11002
 ciqual_food_code:en: 11002
-ciqual_food_name:en: Chervil, fresh
-ciqual_food_name:fr: Cerfeuil, frais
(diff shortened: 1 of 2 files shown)


taxonomies/food/categories.txt
```diff
@@ -74545,46 +74545,198 @@ ciqual_proxy_food_name:en: Caraway, seed
 wikidata:en: Q139678670
 
 < en: Plant-based foods
-en: Fresh chervil, Garden chervil products, Garden chervil, Chervil
+en: Chervils and their products
+da: Kørvel og kørvelprodukter
+es: Productos del perifollo
+fr: Produits de le cerfeuil
```
stephanegigandet:
```suggestion
fr: Produits du cerfeuil
```

taxonomies/food/categories.txt
```diff
@@ -74545,46 +74545,198 @@ ciqual_proxy_food_name:en: Caraway, seed
...
 es: Hojas de perifollo, Hojas de cerefolio, Hojas de perifolio
-fr: Feuilles de cerfeuil, Feuilles de cerfeuil commun
+fr: Feuilles de cerfeuil
 hr: Lišće vrtnog češnjevka
 nl: Kervelbladeren
 
-< en: Garden chervil products
+< en: Chervils and their products
+en: Garden chervils and their products, Garden chervil products
+fr: Produits de le cerfeuil commun
```
stephanegigandet:
```suggestion
fr: Produits du cerfeuil commun
```

taxonomies/food/categories.txt
```diff
@@ -74545,46 +74545,198 @@ ciqual_proxy_food_name:en: Caraway, seed
...
+bg: Смирна
+ca: Mirris
+cs: Čechřice vonná
+cy: Creithig bêr
+da: Spanske kørvel, Spansk kørvel, Sødskærme, Anis-kål
+de: Süßdolde
+et: Mesiputk
+fa: چتردار شیرین
+fi: Saksankirveli
+fr: Cerfeuil musque, Cerfeuil anisé, Cerfeuil d'Espagne
```
stephanegigandet:
```suggestion
fr: Cerfeuil musqué, Cerfeuil anisé, Cerfeuil d'Espagne
```

stephanegigandet (changes requested):
I think that chervil is the same as garden chervil, and bulbous chervil and cicely are 3 different plants, the later 2 should not be under chervil. Just like we don't put sweet potato under potato.

Freso:
> I think that chervil is the same as garden chervil

As I said at the beginning, in English “chervil” may be used exclusively for garden chervil, but in other languages “chervil” seems to be more generic. E.g., https://se.openfoodfacts.org/product/7310394002362/k%C3%B6rvel-kockens is Spanish chervil (or cicely), which I was only able to deduce by going to the producer’s website, <https://www.kockens.se/produkter/korvel/>, to check. If I had gone _only_ by the product name, I would have filed it under “chervil”, which would have been wrong if chervil==garden chervil, but will be okay if chervil acts as a parent category.

stephanegigandet:
> As I said at the beginning, in English “chervil” may be used exclusively for garden chervil, but in other languages “chervil” seems to be more generic. E.g., https://se.openfoodfacts.org/product/7310394002362/k%C3%B6rvel-kockens is Spanish chervil (or cicely), which I was only able to deduce by going to the producer’s website, https://www.kockens.se/produkter/korvel/, to check. If I had gone _only_ by the product name, I would have filed it under “chervil”, which would have been wrong if chervil==garden chervil, but will be okay if chervil acts as a parent category.

Just looking at the changes to the taxonomy, I'm not sure which entries / languages are causing issues. Is it in Swedish, where one word (which one?) can refer to common chervil, bulbous chervil or cicely? and there are no other words in those languages that can be used more specifically? To me it makes more sense to keep the taxonomy hierarchy based on the "is a" relationship, and to try to find at least one specific translation for each entry. Bulbous chervil is not chervil.

Freso:
Okay, disregarding the bulbous one, these were more or less the questions I asked myself working on this. If you come upon a different approach/solution/conclusion than I did, I’m happy to hear it. :)

- How would you handle `sv:körvel` (lit. “chervil”) in a way that encompasses both garden chervils and spanish chervils(/cicelies) both for ingredients and categories?
  - Make a new Nordic-language only parent?
    - But how do you know which other languages use the generic “chervil” for other types of “chervil” than garden chervil?
    - And knowing that other languages use “chervil” in a more generic manner, can you be sure that every single instance of “chervil” in English is, indeed, garden chervil across all English-speaking diasporas and also where ingredient lists may have been translated from other languages?

_For me_, this led me to the conclusion that I can’t trust that “chervil” on the packaging/ingredient list means that it’s garden chervil in the product, which then animated these changes. Spanish chervil and garden chervil _are_ both “chervils” in that they’re colloquially referred to as “chervil”, at least in some linguistic contexts, even if they’re not biologically related. (This is also why I included bulbous chervil, since if “chervil” can refer to either of those two, then I wouldn’t be surprised to find it referring to that third one as well… but I also haven’t come across any direct evidence of this being the case, so I’m okay with leaving that one out.)

stephanegigandet:
My suggestion:

For categories, we don't even need to have a category mapped to "Körvel" if it is ambiguous, we could just use the more specific names. If in most (e.g. 90% of the cases) it's the common Chervil, then we could add it as a synonym to en:chervils

en: Chervils, garden chervils etc.
sv: Trädgårdskörvel, Dansk körvel 

en: Bulbous chervils
sv: Rotkörvel

en: Cicelies
sv: Spanska körvel, Spansk körvel

Worse case (if we put "Körvel" as the synonym of en: Chervils, but we don't have to): we get a few products miscategorized in en: Chervils instead of en: Cicelies.

For ingredients, I would do the same and put Körvel as a synonym for chervil (if it's the most probable) so that we do parse it. If the name is ambiguous, it's very likely that most products would need to use an unambiguous name in their official ingredient list.

I understand that there might a few Swedish products where this would be an issue. But if we use the approach to have en:chervil as the parent, then all products in all languages that contain cicely or bulbous chervils will be listed for the "Chervils" ingredient even though they are not or don't contain chervil...

PR #14327, opened 2026-08-23
feat: Add more ingredient information to the Foodture export

john-gom:
### What

Only inlude child (leaf) ingredients when calculating the top ingredients by quantity

The first 10 root ingredients (i.e. without recursing into compund ingredients) are included separately.

Also includes the full ingredients and packaging data structures

Added CIQUAL data to the ingredients_sum.csv and removed parent ingredients from the main file.

Fixed CSV parsing of the ranked products file.


stephanegigandet (approved):
Perfect, thank you @john-gom !

PR #14335, opened 2026-08-24
fix: treat newlines / line feeds as ingredients separators sometime

### LINKED ISSUE #2867: Treat carriage returns as a comma for ingredient parsing
Many users are putting carriage returns to separate ingredients. For the sake of quality analysis (at least), treat carriage returns as a comma for ingredient parsing.

### Part of
- #5706

stephanegigandet:
This is in particular common in cooking recipes ingredients:

```
1 kg of potatoes
500 g of tomatoes
```

Fixes https://github.com/openfoodfacts/openfoodfacts-server/issues/2867

The approach is that if an ingredient list contains new lines, we try to convert them to commas to see if it results in a better ingredients recognition rate.

Code done with the assistance of AI (Kilo code VSCode extension with free models)


tests/unit/expected_test_results/ingredients/en-origin-field-with-not-taxonomized-entries.json
```diff
@@ -90,6 +90,24 @@
       "en:estimated-ingredients-with-product-opener"
    ],
    "specific_ingredients" : [
+      {
```
john-gom: Why are the specific ingredients doubling up here? Is the second parse maybe appending?

stephanegigandet: @john-gom good catch, I think it's fixed now

john-gom (approved):
Good stuff

PR #13614, opened 2026-05-18
taxonomy: Add translations for product type errors in categories

teolemon:
### What
taxonomy: Add translations for product type errors in categories

PR #14336, opened 2026-08-24
taxonomy: add some ZH-language food category translations

jayaddison:
### What
This is a batch import of Chinese translations that have been prepared by my friend and colleague [Mido](https://www.midoleeproductions.com), under informal paid contract work for OpenCulinary C.I.C. (my company).

To support the translations I have developed a pair of AGPL-licensed utility scripts that export the contents of the OpenFoodFacts `food/categories.txt` file for translation, and subsequently import the translated results (both using CSV format suitable for import into Excel, Google Sheets, or other software during the manual translation entry phase).

The (work-in-progress!) version of scripts used for this iteration of the import can be found at: https://codeberg.org/openculinary/openfoodfacts-data-utils/src/commit/c81940aa752b44ab8fc0d8451610aef04fdedd4c

I have not used and LLMs during the preparation of this pull request.  ~~I will confirm with Mido whether she has used any AI/LLMs for the translations; I do not believe that she has.~~ Mido uses AI translation for a first-pass draft, and then manually reviews the results because those have limited (~50%) accuracy.

### Screenshot or video
N/A

N/A

cc @Freso @moon-rabbitOFF as taxonomy champions

Edit: update AI/LLM usage disclosure

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -272,8 +272,8 @@ nl: Gepasteuriseerde producten
 pl: Produkty pasteryzowane
 pt: Produtos pasteurizados
 sv: Pastöriserade produkter
-incompatible_with:en: categories:en:unpasteurised-products
 zh: 低温杀菌产品
+incompatible_with:en: categories:en:unpasteurised-products
 
 en: Unpasteurised products
 bg: Непатьоризирани продукти
@@ -287,8 +287,8 @@ nl: Ongepasteuriseerde producten
 pl: Produkty niepasteryzowane
 pt: Produtos não pasteurizados
 sv: Opastöriserade produkter
-incompatible_with:en: categories:en:pasteurised-products
 zh: 未低温杀菌产品
+incompatible_with:en: categories:en:pasteurised-products
 
 en: Baby foods
 bg: Бебешки храни
@@ -906,8 +906,8 @@ fr: Lait anti-reflux
 hr: Mlijeko protiv refluksa
 it: Latte antirigurgito, Latte antireflusso
 pt: Leite anti-refluxo
-wikidata:en: Q16034323
 zh: 防反流奶粉
+wikidata:en: Q16034323
 
 < en: Baby milks
 en: Hypoallergenic milks
@@ -1449,9 +1449,9 @@ hu: Italkészítmények
 it: Preparati per bevande
 nl: Drankbereidingen
 pt: Preparações de bebidas
+zh: 饮料调制品
 default_nutrition_prepared_per:en: 100ml
 description:en: Any product which purpose it is to create a beverage which uses water (hor or cold) in some way (diluting, infusing, dissolving, etc).
-zh: 饮料调制品
 
 < en: Beverages and beverages preparations
 en: Beverages, Drinks
@@ -1767,10 +1767,10 @@ hr: Pribor za pravljenje alkohola
 it: Kit per la produzione di alcolici
 nl: Alcohol maak kits
 pt: Kits para fazer álcool
+zh: 酿酒套件
 gpc_category_code:en: 10000142
 gpc_category_description:en: Definition: Includes any products that can be described/observed as an alcoholic drink making kit. These products are total units that include most of the necessary ingredients and equipment required to produce various flavours of beer, wine or spirits. Definition Excludes: Excludes products such as Alcohol Making Accessories sold separately, Non-Alcoholic Drink Making Kits, Alcohol Flavouring Kits and Pre-Made Alcohol.
 gpc_category_name:en: Alcohol Making Kits
-zh: 酿酒套件
 
 < en: Alcoholic beverages
 < en: Variety packs
@@ -1985,9 +1985,9 @@ it: Assenzio
 nl: Absint
 pl: Absynt
 pt: Absinto
+zh: 苦艾酒
 google_product_taxonomy_id:en: 505761
 wikidata:en: Q170210
-zh: 苦艾酒
 
 < en: Absinthe
 fr: Absinthe de Pontarlier
@@ -9636,8 +9636,8 @@ it: Senna
 la: Senna alexandrina, Cassia angustifolia, Senna angustifolia
 nl: Sennapeul
 pl: Senes ostrolistny, Kasja ostrolistna
-wikidata:en: Q132675
 zh: 番泻叶
+wikidata:en: Q132675
 
 < en: Herbal teas
 en: Maca, Maca powder, Peruvian ginseng
@@ -10195,8 +10195,8 @@ nl: Olijfboomproducten
 pl: Produkty drzewa oliwnego
 pt: Produtos de oliveira
 ru: Продукты из оливок
-wikidata:en: Q37083
 zh: 橄榄类制品
+wikidata:en: Q37083
 
 < en: Herbal teas
 < en: Olive tree products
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -149,6 +149,7 @@ tr: Yanlış ürün tipi
 uk: Неправильний тип продукту
 vi: Sai loại sản phẩm
 zh: 错误的产品类型
+zh: 错误产品类型
```
jayaddison, line 151: NB: The import created a duplicate `zh: ...` entry here.  I think that pull request #13936 added this entry while our scripts and translations were being prepared.

```suggestion
zh: 错误产品类型, 错误的产品类型
```

taxonomies/food/categories.txt
```diff
@@ -272,6 +274,7 @@ pl: Produkty pasteryzowane
 pt: Produtos pasteurizados
 sv: Pastöriserade produkter
 incompatible_with:en: categories:en:unpasteurised-products
+zh: 低温杀菌产品
```
```suggestion
zh: 低温杀菌产品
incompatible_with:en: categories:en:unpasteurised-products
```

Freso (changes requested):
Thank you!! Is the script just inserting `zh:` after the last line of the block? It should go above any property lines (and also before any commented-out 3-letter language-code lines). :)

Freso:
I just looked at the import script, and I’m _wondering_ if it would be easier to insert the new entries using [Tree-sitter queries](https://tree-sitter.github.io/tree-sitter/using-parsers/queries/index.html) using [tree-sitter-off_taxonomy](https://codeberg.org/Freso/tree-sitter-off_taxonomy) (also AGPL :)). I haven’t messed around much with Tree-sitter queries outside of for syntax highlighting purposes, so not sure if this would actually be easier or not, but thought I’d at least bring it to your attention!

taxonomies/food/categories.txt
Freso: `zh: 错误的产品类型` seems to originate from https://github.com/openfoodfacts/openfoodfacts-server/pull/13614 and @teolemon didn’t note what their source(s) was/were on neither that one or in https://github.com/openfoodfacts/openfoodfacts-server/pull/13936 so I’m inclined to trust your Certified Human™ (probably) translation over the one with unknown origin. (But still keep the other one as a synonym, probably, unless a Chinese reader/writer chimes in and says that it’s flat-out wrong.)

jayaddison:
@Freso 
> Thank you!! Is the script just inserting `zh:` after the last line of the block? It should go above any property lines (and also before any commented-out 3-letter language-code lines). :)

And thank you for the review!

So far, that is what the script does, yes - but I should be able to update it to factor in those updated rules.

_Really_, I should refactor it to allow writing test coverage of some example situations such as those too (and doing that would help with readability/comprehensibility in future too).

taxonomies/food/categories.txt
jayaddison: This makes sense, yep -- I'll add a check that appends the additional value to the end of the list of existing synonyms -- albeit with an existing-presence-check, so that duplicates do not get added.

(again, test coverage of this logic would make sense)

jayaddison: _Generally_ this would make sense, although since this seems to be an OFF-internal category (I think?), maybe we should aim to have a single value.  Visually the two values look very similar; perhaps I can check with Mido about the difference in interpretation/meaning of them, and whether either is more sensible/appropriate.

jayaddison:
@Freso 
> I just looked at the import script, and I’m _wondering_ if it would be easier to insert the new entries using [Tree-sitter queries](https://tree-sitter.github.io/tree-sitter/using-parsers/queries/index.html) using [tree-sitter-off_taxonomy](https://codeberg.org/Freso/tree-sitter-off_taxonomy) (also AGPL :)). I haven’t messed around much with Tree-sitter queries outside of for syntax highlighting purposes, so not sure if this would actually be easier or not, but thought I’d at least bring it to your attention!

Thanks :)  I'm vaguely aware of tree-sitter, but have not used it much myself.  Because I'll be the maintainer/user of this code at the moment, I'll probably stick with what I know at the moment (even if the resulting code is somewhat inelegant).

jayaddison:
> ### Large Language Models usage disclosure
> 
> I have not used and LLMs during the preparation of this pull request. I will confirm with Mido whether she has used any AI/LLMs for the translations; I do not believe that she has.

Mido begins with a single-pass AI machine translation, to produce sample/candidate translations.  This has a low accuracy rate (estimated ~50%), so all the translations are manually reviewed, and corrected/replaced where necessary.

Freso:
```
ERROR - duplicate_synonym - 39609 - zh:糖果 already is associated to en:bonbons (categories) - zh:糖果 cannot be mapped to entry en:candies
[categories] ERROR - duplicate_synonym - 40628 - zh:杏仁膏 already is associated to en:marzipan (categories) - zh:杏仁膏 cannot be mapped to entry en:almond-paste
[categories] ERROR - duplicate_synonym - 41941 - zh:谷物 already is a synonym of zh:谷物 for entry en:cereal-grains (categories) - zh:谷物 cannot be mapped to entry en:seeds / zh:种子
```

`zh:糖果` is mapped to both `en:bonbons` and `en:candies`,
`zh:杏仁膏` is mapped to both `en:marzipan` and `en:almond-paste`, and
`zh:谷物` is mapped to both `en:cereal-grains` and `en:seeds`.

Meowdolee:
@jayaddison Here are my thoughts on how we could differentiate these similar items in Chinese.

zh: 精致糖果／夹心糖／巧克力糖  en:bonbons (categories) 
zh: 糖果  en:candies

zh: 杏仁蛋白软糖  en:marzipan (categories)  
zh: 杏仁膏  en:almond-paste


zh: 麦片谷物  en:cereal-grains (categories) - 
zh: 种子/谷物  en:seeds

jayaddison:
Thank you @Meowdolee 
> zh: 精致糖果／夹心糖／巧克力糖 en:bonbons (categories)

I think including the notion that they are chocolate makes sense.  What I know them as in the United Kingdom can be purely sugar-based (and often oval-shaped) -- but in other cultures, and more importantly, in the OpenFoodFacts taxonomy, `en:bonbons` is within the category of `en:Chocolate candies` -- so let's go with `巧克力糖`?

taxonomies/food/categories.txt
jayaddison: After discussing these with @Meowdolee: the shorter version is preferable.  I should have asked explicitly whether we should remove the longer synonym version; I suppose it's harmless to keep it in place too.

I'm going to apply your suggestion.

jayaddison: Eugh, except then I had already forgotten that this is in the context of an OFF-internal category, so... yep, let's just keep the shorter version.

taxonomies/food/categories.txt
jayaddison: I've applied this change -- regrouping the translations to appear within the block of all language translations, before properties -- across the translations in the file.

...and in the meantime I've begun refactoring the import script, but do not have this logic implemented there yet.  I promise I will do before any subsequent import PR uses it.

Meowdolee:
If I have to pick one of these three, I would choose **zh: 精致糖果 / en: bonbons** for the category. @jayaddison 
Again, this word can have different meanings depending on whether you’re referring to British, American, French, or German usage, so it’s difficult to pinpoint which translation makes the most sense.

jayaddison:
> If I have to pick one of these three, I would choose **zh: 精致糖果 / en: bonbons** for the category. @jayaddison Again, this word can have different meanings depending on whether you’re referring to British, American, French, or German usage, so it’s difficult to pinpoint which translation makes the most sense.

Ok, thank you @Meowdolee.  I'll use that as the primary/first translation, and list the others as synonyms/secondary.  Enjoy your weekend!

Freso (approved):
Thanks!

jayaddison:
You're welcome - thank you!

jayaddison:
There was a bug in my category-translations import script that meant that only categories that had a single primary name -- such as `en: foo` -- _without_ any synonyms, as for example `en: foo, bar` -- would be imported.

So: this pull request was fairly incomplete.  I should have a second round of translations provided by @Meowdolee that I'll open as a pull request soon.  I'm doing some more tidyup and cleanup for the script before then.

PR #14337, opened 2026-08-24
taxonomy(food): allspice edits

Freso:
Allspices seem to not actually be biologically related to peppers, but maybe they are even if they’re not part of the Piper genus? They seem to colloquially be referred to as such anyway, and do have some visual resemblance to peppers.

- Adds allspice categories.
- Adds translations for allspice ingredient.
  - da/sv/nb/nn + from Wikidata.
- Normalises case and singular/plural forms.

Needed-for: https://se.openfoodfacts.org/product/7311041034293/hel-kryddpeppar-eldorado
Needed-for: https://world.openfoodfacts.org/facets/categories/allspice


stephanegigandet (approved):
Thanks

PR #14343, opened 2026-08-26
(DRAFT) taxonomy: Propose affixing an issue to a nutrition_field

teolemon:
### What
- We have added inline warning for nutrition editing in Explorer. Those are hard-coded, and rely on simple heuristics
- We show in view mode the DQ Knowledge Panel
- This proposal is about adding a property: 'nutrition_field' to every data quality issue, so that it's not just about opening nutrition (which we can already do in explorer), but showing the issue directly inside/next to the field

### DIFF
--- a/taxonomies/data_quality.txt
+++ b/taxonomies/data_quality.txt
@@ -12654,6 +12654,8 @@ description:ru: Категория «детское питание» была а
 en: Detected category from brand - Cigarettes
 description:en: Those are likely to be cigarettes, based on the brand. There might be some false positives. All cigarettes should be categorized as non-food products, and moved to Open Products Facts by inputting OPF in the Categories field.
 fix_action:en: add_categories
+# fix_action:en: change_product_type: product
+# fix_action:en: add_category: en:cigarettes not syntaxically correct
 # show_to:en: moderators
 
 < en: Categories warnings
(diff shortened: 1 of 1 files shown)


taxonomies/data_quality.txt
```diff
@@ -12654,6 +12654,8 @@ description:ru: Категория «детское питание» была а
 en: Detected category from brand - Cigarettes
 description:en: Those are likely to be cigarettes, based on the brand. There might be some false positives. All cigarettes should be categorized as non-food products, and moved to Open Products Facts by inputting OPF in the Categories field.
 fix_action:en: add_categories
+# fix_action:en: change_product_type: product
+# fix_action:en: add_category: en:cigarettes not syntaxically correct
```
teolemon:
```suggestion
```

taxonomies/data_quality.txt
```diff
@@ -12692,6 +12694,7 @@ description:ru: Этот продукт относится одновремен
 en: Detected category from brand - Pet Foods
 description:en: Those are likely to be product for Open Pet Food Facts, based on the brand. There might be some false positives. All Pet Food products should be moved to Open Pet Food Facts by inputting OPFF in the Categories field.
 fix_action:en: add_categories
+# fix_action:en: change_product_type: pet_food
```
teolemon:
```suggestion
```

PR #14350, opened 2026-08-26
chore:  ifct entries part 3

aleene:
### What
Some more IFCT entries


Freso (commented):
```
Warning(Correctness): on en: semolina (line 24269)
"coment" might be a typo of "comment":
- 
- coment:en: Be careful that in some languages semolina is also used for coarse grinded rice, maize, etc. Wikipedia is not very consistent.
```
and
```
Warning(Correctness): on en: wheat flour (line 27354)
"commnet" might be a typo of "comment":
- 
- commnet:en: IFCT classifies by processing description, while the French system is a legally defined ash-content standard — so the mapping is functional/nutritional rather than an exact technical equivalence.
```
from https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/32973135493/job/98195912745?pr=14350

aleene:
@freso. OK thanks. DIdn't that used to result in an error?

stephanegigandet (approved):

Freso:
@aleene It’s a brand new check (less than a week old? about a week?) that I set to a warning-level in case there _is_ some property that matches the check, so we are free to ignore it… but I think I might raise the level to error and it can be decreased to warning if we ever introduce a property that would conflict with it.

PR #14353, opened 2026-08-26
taxonomy: Update nutrient_levels.txt by removing English stored as other languages entries

teolemon:
taxonomy: Update nutrient_levels.txt by removing English stored as other languages entries


stephanegigandet (approved):
Thanks

PR #14355, opened 2026-08-26
taxonomy: translate HBT additive

teolemon:
taxonomy: translate HBT additive
does not have e E number, only 3 products with it: https://world.openfoodfacts.org/facets/ingredients/hydroxymethylbutyrate
current additive parsing does not work well : https://world.openfoodfacts.org/facets/additives/hydroxymethylbutyrate%20-%20HMB

### DIFF
--- a/taxonomies/additives.txt
+++ b/taxonomies/additives.txt
@@ -26136,41 +26136,6 @@ sv: metylsulfonylmetan
 wikidata:en: Q423842
 wikipedia:en: https://en.wikipedia.org/wiki/Methylsulfonylmethane
 
-en: hydroxymethylbutyrate, HMB
-xx: hydroxymethylbutyrate, HMB
-bg: хидроксиметилбутирова киселина, HMB
-ca: hidroximetilbutirat, HMB
-cs: hydroxymethylbutyrát, HMB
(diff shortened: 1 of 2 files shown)


Freso (approved):
is it needed to have “HMB” for all the languages in addition to it being in `xx`?

taxonomies/additives.txt
```diff
@@ -26138,6 +26138,35 @@ wikipedia:en: https://en.wikipedia.org/wiki/Methylsulfonylmethane
 
 en: hydroxymethylbutyrate, HMB
 xx: hydroxymethylbutyrate, HMB
+bg: хидроксиметилбутирова киселина, HMB
```
stephanegigandet: No need for HMB everywhere

stephanegigandet (changes requested):
> does not have e E number, only 3 products with it: https://world.openfoodfacts.org/facets/ingredients/hydroxymethylbutyrate

It doesn't sound like it's an additive, it should be moved to ingredients.txt I think

teolemon:
fixed @stephanegigandet @Freso

Freso (approved):

PR #14358, opened 2026-08-26
taxonomy: support for care instructions - textile

teolemon:
I had it on my drive for a while
taxonomy: support for care instructions - textile


stephanegigandet (approved):
Images should be renamed to 90x90.svg or they won't be displayed

teolemon:
@stephanegigandet done and i've added a README for good measure

PR #14366, opened 2026-08-27
feat: ingredient parsing for 1 egg, 2 small carrots

### LINKED ISSUE #12161: Specified amounts/counts prevent recognition of ingredients
### What
- If an ingredient list has entries like “6 pressed apples”, that entry will not get recognised by the ingredient parser.

1. Go to https://world.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=6+pressed+apples&type=add&lc=en&action=process&estimator=product_opener
2. See ingredient not being recognised

### Expected behavior
- should give same result as https://world.openfoodfacts.org/cgi/test_ingredients_analysis.pl?ingredients_text=pressed+apples&type=add&lc=en&action=process&estimator=product_opener

At least two:

- https://se.openfoodfacts.org/product/5038862134293/bl%C3%A5b%C3%A4r-svarta-vinb%C3%A4r-tranb%C3%A4r-%C3%A4pple-juice-innocent
- https://se.openfoodfacts.org/product/5038862138376/innocent-apple-kiwi-cucumber

(and I’m guessing it’s same for any [other juices from the same brand](https://se.openfoodfacts.org/facets/brands/innocent/categories/fruktsafter).)

### LINKED ISSUE #14056: Handle single quantity in parsing
### Problem
Score my recipe use the OFF parser to parse recipes,
but in recipes you might find quantities like 1 egg, 3 tomatos, 1 chicken wing etc.
We would like to be able to parse it.

See https://github.com/openfoodfacts/score-my-recipe/issues/29

### Proposed solution

Ingredient parser already recognize quantities with units (like 100g) and percents, make it also recognize numerals.

We already have an (undocumented) `quantity_g` field. Either we introduce a quantity: { amount unit} field, either we add a quantity_unit field (a bit ugly IMO).

### LINKED ISSUE #14057: Find the way to get "median" weight for using taxonomies
### Problem

In score-my-recipe project, we need the weight of each component/ingredient of a recipe.
But usually some components are expressed as units: 1 egg, 3 tomatoes, 2 chicken wing

See: https://github.com/openfoodfacts/score-my-recipe/issues/29

### Proposed solution

Can we introduce some data about median weight (or something like that, maybe with variance  or quartiles) in the taxonomy ?

On source can be the USDA which propose different weights, eg. https://fdc.nal.usda.gov/food-details/171689/measures, but their might be other data sources.

On caveats might also be the existence of difference classes (like for melons, ranging from baby to big, [see here for example](https://guiderhd.ctifl.fr/fiche/fruits/melon))
(fun fact, some of this info can be found in french legal portal (digging into documents), [for example for melon](https://www.legifrance.gouv.fr/search?typeRecherche=date&searchField=ALL&query=calibrage+melon&searchProximity=&searchType=ALL&typePagination=DEFAUT&pageSize=25&page=1&isAdvancedResult=false)

Of course we can also decide to just use the USDA link (and add more on ingredients) to get this information (and other database later)

### Additional context

Could we eventually "crowd source this data" like letting people input how much an egg / a tomato they have at home weight ?

I don't know if Open Prices would have some kind of use for that (eg. to compute price/kg for items sold by units, as it's frequently the case for melon or kiwi) ?

stephanegigandet:
Needed in particular to parse cooking recipes like "1 egg, 2 carrots, 3 small carrots" or "3 carottes de taille moyenne".
Needed by the Score my recipe project: https://github.com/openfoodfacts/score-my-recipe/issues/29

Fixes #14056 (parsing)
Fixes #12161 (parsing)
Fixes #14057 (average weight in ingredients taxonomy)

The PR adds new properties in the ingredients taxonomy:

```
# weight value based (rounded) on https://www.fda.gov/food/nutrition-food-labeling-and-critical-foods/raw-fruits-poster-text-version-accessible-version
average_weight_per_unit:en: 200
average_weight_per_unit_large:en: 240
```

We need at least average_weight_per_unit:en: We can add sizes for "small" and "large" variants, but if not present, we will just use a conversion factor.

Recognized size names are in a new sizes.txt taxonomy. There are stopwords so that we can recognize things like "3 tomates de taille moyenne".

In the ingredients structure, we record "size" if we recognized one, and the quantity_g property is computed if we have an average_weight_per_unit property for the ingredient.


tests/integration/expected_test_results/export/export_database/rows/0850032917148.json
```diff
@@ -88,7 +88,7 @@
    "image_small_url" : "",
    "image_url" : "",
    "ingredients_analysis_tags" : "en:palm-oil-content-unknown,en:vegan,en:vegetarian",
-   "ingredients_tags" : "en:water,en:aloe-vera,en:vegetable,en:sugar,en:added-sugar,en:disaccharide,en:e330,en:sodium-citrate,en:minerals,en:sodium,en:Aloe Artifical Flavor,en:e327,en:Strawberry Artificial Flavor,en:e300,en:e418,en:EDTA,en:e955,en:FD&C Red No,en:40",
+   "ingredients_tags" : "en:water,en:aloe-vera,en:vegetable,en:sugar,en:added-sugar,en:disaccharide,en:e330,en:sodium-citrate,en:minerals,en:sodium,en:Aloe Artifical Flavor,en:e327,en:Strawberry Artificial Flavor,en:e300,en:e418,en:EDTA,en:e955,en:FD&C Red No",
```
stephanegigandet, line 91: This one is a false positive.

lib/ProductOpener/Ingredients.pm
```diff
@@ -1495,16 +1495,57 @@ if ($ingredient =~ /\s$percent_or_quantity_regexp$/i) {
...
+	# % unit
 	if ($percent_or_quantity_unit =~ /\%/) {
 		$percent = $percent_or_quantity_value;
 	}
+	# Empty unit
+	elsif ($percent_or_quantity_unit eq "") {
+		$quantity = $percent_or_quantity_value;
+		# Check if the ingredient has properties like:
+		# average_weight_per_unit:en: 70
+		# average_weight_per_unit_large:en: 100
```
teolemon, line 1512: the usage of the average_weight_per_unit_large will come in a later PR @stephanegigandet ?

teolemon (commented):
Had a go at it with Gemini 3.1 Pro in Antigravity, which only proposed false positives

teolemon (approved):

PR #14378, opened 2026-08-28
taxonomy: IFCT ingredients part 4

aleene:
### What
More IFCT entries added


taxonomies/food/ingredients.txt
sqwrrl: Please add "Moth Dal" & "Moth" as synonyms

mvark, line 1: Please add "Rajma" as a synonym as that spelling variant is common in India

aleene: ok, done

teolemon (approved):

PR #14382, opened 2026-08-28
feat: Add support for Amazon FNSKU barcode

### LINKED ISSUE #11038: Add support for Amazon barcodes
### Problem
- a while ago we tightened barcode validation 
- this reduces bad barcodes for food but in the case of non-food products, Amazon is actually producing its own barcode symbiology with letters and numbers for products that don't typically have a barcode. 
- we end up with a lot of products that we cannot upload to Open Products Facts as a result 
- Add support for Amazon barcodes

oxidn:
### What
Adds support for Amazon FNSKU barcodes such as X001UVTA2F.

### Changes
Recognize Amazon FNSKUs using the X00 alphanumeric format and Prevent invalid barcodes like  X101UVTA2F from being converted into numeric barcodes.

### Testing
Manually tested and numeric barcodes continue to work.

Fixes: #11038

### DIFF
--- a/lib/ProductOpener/Products.pm
+++ b/lib/ProductOpener/Products.pm
@@ -581,6 +581,10 @@ sub normalize_code_with_gs1_ai ($code) {
 
 	my $ai_data_str;
 	if (defined $code) {
+		# Keep Amazon FNSKUs unchanged
+        return ($code, undef) if $code =~ /^X00[A-Z0-9]{7}$/;
+		return ($code, undef) if $code =~ /^X/ && $code !~ /^X00[A-Z0-9]{7}$/;
+
 		my ($gs1_code, $gs1_ai_data_str) = &_try_normalize_code_gs1($code);
 		if ($gs1_code and $gs1_ai_data_str) {
 			$code = $gs1_code;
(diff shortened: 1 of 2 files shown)


TaciteOFF:
Thanks for this PR. The idea is useful, but I think a few points need to be fixed before merging:
- The storage path is not reversible. X001UVTA2F is stored as 000/X00/1UV/TA2F, then some maintenance scripts read it back as 000X001UVTA2F.
- Lowercase codes or codes with a leading space can still be converted into an unrelated numeric barcode.
- This change affects all flavors (OFF, OBF, OPF, OPFF...). Not sure it's intended?
- Image scanning and some import tools might reject alphanumeric codes.
- Please add tests for normalization and the storage-path round trip.

See:
https://github.com/openfoodfacts/openfoodfacts-server/blob/4e803ef57c0981eb65ddba398527f95d4e780dbd/lib/ProductOpener/Images.pm#L414-L465
https://github.com/openfoodfacts/openfoodfacts-server/blob/4e803ef57c0981eb65ddba398527f95d4e780dbd/lib/ProductOpener/ImportConvert.pm#L146-L156
+ inputmode="numeric" & pattern="\d*" on the web scanner. 

Please note that Amazon say that the FNSKU is linked to the seller. (https://sellercentral-europe.amazon.com/seller-forums/discussions/t/47c3e272-b3e2-4965-8073-483f3b25a863)
Which means many different products from different sellers could share the same FNSKU.

PR #14383, opened 2026-08-28
fix: count individual polyols (isomalt, maltitol, sorbitol) in energy computation

### LINKED ISSUE #14344: isomalt should be considered as a polyol
### What
It seems that isomalt is not considered as a polyol in the energy calculation.

1. Edit this product (which raises an NF incoherence error): https://world.openfoodfacts.org/product/4002590140940/pulmoll-eukalyptus-menthol-zuckerfrei
2. Replace the isomalt with polyols, save, and see that it fixes the problem

### Expected behavior
- isomalt and maltitol should be calculated as polyols because they are sugar alcohols

### Additional context
- mentioned for the 1st time here: https://openfoodfacts.slack.com/archives/C03H290LF/p1785199581513629
- Stéphane says it's in this function of Nutrition.pm: https://github.com/openfoodfacts/openfoodfacts-server/blob/main/lib/ProductOpener/Nutrition.pm#L2619

Eduardo-Vilas-Boas:
## What
Products that enter individual polyols (isomalt, maltitol, sorbitol) instead of the
aggregate "polyols" nutrient had those grams counted as regular carbohydrate
(4 kcal/g) instead of the polyols rate (2.4 kcal/g) when computing energy from
nutrients. This produced an inflated computed energy value and a false
"energy does not match nutrients" data quality warning, as reported for
[Pulmoll Eukalyptus Menthol Zuckerfrei](https://world.openfoodfacts.org/product/4002590140940/pulmoll-eukalyptus-menthol-zuckerfrei),
which lists isomalt but no aggregate polyols value.

## How
- `compute_energy_from_nutrients_for_nutrients_set` in `Nutrition.pm` only ever
  fell back to `erythritol` when the aggregate `polyols` value was missing.
- When `polyols` isn't set, sum whichever individual polyols
  (`erythritol`, `isomalt`, `maltitol`, `sorbitol`) are present and use that sum
  as the effective polyols value in the energy formula, instead of only checking
  erythritol.
- `erythritol` keeps contributing 0 kcal/g as before; the other individual polyols
  now correctly contribute at the general polyols rate (2.4 kcal/g / 10 kJ/g)
  instead of being silently treated as plain carbohydrate.

## Testing
- Added two unit tests to `tests/unit/dataqualityfood.t`: isomalt alone, and
  isomalt + maltitol summed, both without an aggregate `polyols` value. Verified
  they fail against the old code (computed energy off by 700 kJ) and pass against
  the fix.
- `make test-unit test="dataqualityfood.t"` passes (196 assertions).
- Ran related suites (`nutrition.t`, `nutriscore.t`, `nutrition_estimation.t`,
  `nutrient_levels.t`, `nutrition_ciqual.t`, `ingredients_nutriscore.t`) — all pass.
- `make checks` (perltidy, perlcritic, syntax, taxonomy lint) — clean.

Fixes #14344

### DIFF
--- a/lib/ProductOpener/Nutrition.pm
+++ b/lib/ProductOpener/Nutrition.pm
@@ -2584,6 +2584,11 @@ sub convert_salt_to_sodium ($salt_value) {
 	},
 );
 
+# Polyols that can be entered individually in the nutrition facts, without a value for the
+# aggregate "polyols" nutrient. Erythritol is listed here too, but it is handled separately
+# above as it does not contribute any energy, unlike the other polyols.
+my @individual_polyols_nids = ("erythritol", "isomalt", "maltitol", "sorbitol");
+
 =head2 compute_energy_from_nutrients_for_nutrients_set ( $nutrients_ref, $unit )
 
 Computes the energy from other nutrients for a given input set.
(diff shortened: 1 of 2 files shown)


stephanegigandet (approved):
Thank you!

PR #14402, opened 2026-08-31
fix: add CSRF tokens

stephanegigandet:
Requires POST method on destructive actions.
Add CSRF token.


lib/ProductOpener/Users.pm
```diff
@@ -1660,6 +1666,11 @@ sub init_user ($request_ref) {
 							if $session_ref->{refresh_expires_at};
 						$request_ref->{id_token} = $session_ref->{id_token} if $session_ref->{id_token};
 					}
+					if (not defined $user_ref->{csrf_token}) {
+						$user_ref->{csrf_token} = generate_token(32);
+						store_user_preferences($user_ref);
+					}
+					$request_ref->{csrf_token} = $user_ref->{csrf_token};
```
alexgarel, line 1676: I find it a bit strange to store CSRF token with user preferences.
It will put a bit more strain on our disks.

My approach would be to store the CSRF / user in memcached. We may loose CSRF on memcached restart, but it's a rare event.

Also ideally CSRF should be unique to a URL (but it might be ok to have one per user).

A method that requires no caching is to compute a HMAC with `day + userid + uri` using an secret defined in configuration. From midnight to 3 am you might also test for the previous day HMAC.

PR #14435, opened 2026-09-02
taxonomy(food): deduplication of name lists in categories.txt

jayaddison:
### What
I'm developing a process to batch-import translations from a CSV file provided by a translator into OpenFoodFacts taxonomies, beginning with the food `categories` taxonomy.

To tolerate changes that occur while translation is taking place, I've added logic to run an order-preserving deduplication on product names, and also to sort the resulting translations on a per-product basis.

During this process, the script has uncovered that a small number of duplicated product names exist, and that a similar small number of translations were not sorted by language code.

This pull request applies those changes _without_ merging in any new translations.  In other words: this should not produce any user-visible effect, but does tidy up the `categories.txt` taxonomy file.

:warning: **NB**: It seems that I do not yet understand the sorting rules for names in the taxonomy files -- so I have reverted the changes in this branch related to that aspect.

I have not used any LLMs, neither to develop the import script that produces these cleanups, nor to open this pull request.

Follow-up to #14336 

Edit: add note that only deduplication is applied in the latest revision of the changeset.

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -2035,8 +2035,8 @@ wikidata:en: Q1142986
 
 < en: Rice wines
 xx: Sake
-bg: Саке
 en: Sake
+bg: Саке
 fr: Saké
 hu: Szaké
 it: Sake, Sakè
@@ -34981,8 +34981,8 @@ wikidata:en: Q940999
 
 < en: Breads
 en: Sourdough breads
-Fr: Pains au levain
 da: Surdejsbrød
+Fr: Pains au levain
 sv: Surdegsbröd
 agribalyse_food_code:en: 7002
 ciqual_food_code:en: 7002
@@ -36463,20 +36463,20 @@ gpc_category_name:en: Cakes - Sweet (Perishable)
 
 < en: Cakes
 xx: Bienenstich
-de: Bienenstich
 en: Bienenstich
+de: Bienenstich
 wikidata:en: Q338205
 
 < en: Cakes
 xx: Donauwelle
-de: Donauwelle
 en: Donauwelle
+de: Donauwelle
 wikidata:en: Q829243
 
 < en: Cakes
 xx: Streuselkuchen
-de: Streuselkuchen
 en: Streuselkuchen
+de: Streuselkuchen
 wikidata:en: Q451645
 
 < en: Streuselkuchen
@@ -36493,8 +36493,8 @@ de: Pflaumen Streuselkuchen
 
 < en: Cakes
 xx: Zupfkuchen
-de: Zupfkuchen
 en: Zupfkuchen
+de: Zupfkuchen
 
 < en: Cakes
 en: Banana breads
@@ -80909,8 +80909,8 @@ pl: Polskie produkty mięsne
 
 < en: Polish meat products
 pl: Kiełbasa krakowska sucha staropolska
-de: Krakauer
 en: Krakowska
+de: Krakauer
 fr: Karkauer
 origins:en: en:poland
 protected_name_file_number:en: TSG-PL-2145
@@ -81310,8 +81310,8 @@ pl: Słoweńskie produkty mięsne
 
 < en: Slovenian meat products
 sl: Kranjska klobasa
-de: Krainer Wurst
 en: Carniolan sausage
+de: Krainer Wurst
 origins:en: en:slovenia
 protected_name_file_number:en: PGI-SI-0764
 protected_name_type:en: pgi
@@ -93068,9 +93068,9 @@ wikidata:en: Q1265899
 # don't confused the "French" style and the "German" style liver sausages. The French ones are cured and the German ones are precooked and usually spreadable
 < en: Prepared meats
 xx: Liverwurst
+en: Liverwurst
 bg: Колбаси от черен дроб
 de: Leberwürste, Leberwurst
-en: Liverwurst
 fi: Maksamakkarat
 hr: Jetrene kobasice
 hu: Májkolbászok
@@ -100479,8 +100479,8 @@ nl: Vruchtenbiscuits
 
 < en: Biscuits and cakes
 fr: Biscuits fourrés aux fruits
-de: Fruchtgefüllte Kekse
 en: Fruit-filled biscuits
+de: Fruchtgefüllte Kekse
 es: Galletas rellenas de fruta
 hr: Keksići punjeni voćem
 nl: Gevulde koekjes met fruit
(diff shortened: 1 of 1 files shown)


taxonomies/food/categories.txt
```diff
@@ -118152,7 +118152,6 @@ pt: Manteigas de soja
 < en: Legumes and their products
 < en: Meat alternatives
 xx: Tofu
-en: Tofu, Tofus, bean curds
```
jayaddison, line 118155: The `xx` (language-less) language code is the first-listed / primary language code here; is it safe to move the `en` language from second-place to much later in the list?  For example: would this cause `bean-curds` not to be a valid part of the URL (path component) to refer to this category?

jayaddison: Answer (courtesy of the [`Check Perl` tests](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/33642009381/job/100289637109?pr=14435)): no, even when `xx` is the first-listed translation, we should not move the subsequent translation line (e.g. `en` or `fr`).  It is simply a special case where there is _also_ a declaration that the product has a language-agnostic name.

taxonomies/food/categories.txt
```diff
@@ -2035,8 +2035,8 @@ wikidata:en: Q1142986
 
 < en: Rice wines
 xx: Sake
-en: Sake
 bg: Саке
+en: Sake
```
```suggestion
en: Sake
bg: Саке
```
`en` should always be the first “real” language (except when not, but it should never be sorted with the other languages)

taxonomies/food/categories.txt
```diff
@@ -34981,8 +34981,8 @@ wikidata:en: Q940999
 
 < en: Breads
 en: Sourdough breads
-da: Surdejsbrød
 Fr: Pains au levain
+da: Surdejsbrød
```
```suggestion
da: Surdejsbrød
fr: Pains au levain
```
`fr` doesn’t get special treatment like `en` does

taxonomies/food/categories.txt
```diff
@@ -36463,20 +36463,20 @@ gpc_category_name:en: Cakes - Sweet (Perishable)
 
 < en: Cakes
 xx: Bienenstich
-en: Bienenstich
 de: Bienenstich
+en: Bienenstich
```
```suggestion
en: Bienenstich
de: Bienenstich
```
though given that the English name is just the German name, and it’s already in `xx`, i think a reasonable argument could also be made for

```suggestion
de: Bienenstich
```

taxonomies/food/categories.txt
```diff
@@ -36463,20 +36463,20 @@ gpc_category_name:en: Cakes - Sweet (Perishable)
...
 wikidata:en: Q338205
 
 < en: Cakes
 xx: Donauwelle
-en: Donauwelle
 de: Donauwelle
```
Freso, line 36473: as with Bienenstich

taxonomies/food/categories.txt
```diff
@@ -36463,20 +36463,20 @@ gpc_category_name:en: Cakes - Sweet (Perishable)
...
+en: Donauwelle
 wikidata:en: Q829243
 
 < en: Cakes
 xx: Streuselkuchen
-en: Streuselkuchen
 de: Streuselkuchen
+en: Streuselkuchen
```
Freso, line 36479: see Bienenstich

taxonomies/food/categories.txt
```diff
@@ -36493,13 +36493,13 @@ de: Pflaumen Streuselkuchen
 
 < en: Cakes
 xx: Zupfkuchen
-en: Zupfkuchen
 de: Zupfkuchen
+en: Zupfkuchen
```
Freso, line 36497: see Bienenstich

taxonomies/food/categories.txt
```diff
@@ -64015,7 +64015,7 @@ ciqual_food_name:en: Buckwheat flour
 ciqual_food_name:fr: Farine de sarrasin
 
 < en: Buckwheat flours
-fr: Farine de blé noir de Bretagne, Farine de blé noir de Bretagne
+fr: Farine de blé noir de Bretagne
```
Freso, line 64018: `fr` is first here because it’s the native language of the product term :) (and it doesn’t have any `xx` or `en` entries)

taxonomies/food/categories.txt
```diff
@@ -80909,8 +80909,8 @@ pl: Polskie produkty mięsne
 
 < en: Polish meat products
 pl: Kiełbasa krakowska sucha staropolska
-en: Krakowska
 de: Krakauer
+en: Krakowska
```
Freso, line 80913:
```suggestion
en: Krakowska
de: Krakauer
```
`pl` is first because it’s the native language. `en` should be first among other languages.

taxonomies/food/categories.txt
jayaddison: I've [reverted](https://github.com/openfoodfacts/openfoodfacts-server/pull/14435/commits/104d1f75183d714067cacacc5b48f6886be46025) the attempt to sort the names lists.  I need to do more work to understand the logic behind the sort order for those.

taxonomies/food/categories.txt
```diff
@@ -81310,8 +81310,8 @@ pl: Słoweńskie produkty mięsne
 
 < en: Slovenian meat products
 sl: Kranjska klobasa
-en: Carniolan sausage
 de: Krainer Wurst
+en: Carniolan sausage
```
Freso, line 81314: https://github.com/openfoodfacts/openfoodfacts-server/pull/14435/changes#r3915854424

taxonomies/food/categories.txt
```diff
@@ -93068,9 +93068,9 @@ wikidata:en: Q1265899
 # don't confused the "French" style and the "German" style liver sausages. The French ones are cured and the German ones are precooked and usually spreadable
 < en: Prepared meats
 xx: Liverwurst
-en: Liverwurst
 bg: Колбаси от черен дроб
 de: Leberwürste, Leberwurst
+en: Liverwurst
```
Freso, line 93073: `en` always “first”

Freso (commented):
i see there’s been an update here, so i’ll just submit what i’ve already commented—a lot of it may be obsolete now :)

Freso (approved):
without resorting, this looks great ;) feel free to reach out in slack or somewhere with questions about the sorting!

jayaddison:
> without resorting, this looks great ;) feel free to reach out in slack or somewhere with questions about the sorting!

Thanks @Freso - yep, reconnecting on Slack sounds like a good idea.  I'll get around to that Soon :tm:.

I'd like to generalise my script so that it can add names in a sensible position within the existing translations list (`ZH` translations are an artifically simple case, because adding them to the end of a product name list generally fits with alphabetic sorting).

I suppose a language-agnostic approach could be for the script to scan backwards from the end of each product's translations, and to merge the names at the point where `prev < value-to-insert < next` -- doing that would mean that I could ignore any special-cases around the first one or two names defined per-product.

...I'll think about it more (and perhaps get back onto Slack to discuss exactly that) within the next few days/weeks.

PR #14438, opened 2026-09-02
taxonomy: standardise lowercasing of language codes

jayaddison:
### What
I think it is at least an inconsistency, or perhaps a bug, that some of the taxonomy files have entries where the language codes contain uppercase letters.

Only a small number of entries are affected, in `ingredients.txt` and `categories.txt`, and it always seems to be the first of the two characters in the two-letter language codes that are unexpectedly uppercase.

The following `vim` search-and-replace can fix this within a file: `:%s/^\([A-Z][a-z]:\) /\L\1 /g`

(`%s/` = 'global search and replace', `\(...\)` captures and saves a temporary result for each match, `[A-Z][a-z]` matches precisely one uppercase latin character followed by precisely one lowercase, `/` divides the search (on the left) with the replacement (on the right), `\L` in the replacement transforms a saved item from uppercase to lowercase, and `\1` indicates that the saved result to insert is the match from the search)

I haven't used any LLMs to prepare this changeset or to open the pull request.

### Screenshot or video
N/A

Observed during follow-up work from #14336.

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -34982,7 +34982,7 @@ wikidata:en: Q940999
 < en: Breads
 en: Sourdough breads
 da: Surdejsbrød
-Fr: Pains au levain
+fr: Pains au levain
 sv: Surdegsbröd
 agribalyse_food_code:en: 7002
 ciqual_food_code:en: 7002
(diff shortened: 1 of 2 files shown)


Freso (approved):

Freso:
Thank you!

jayaddison:
You're welcome :)  thank you!

PR #14442, opened 2026-09-02
taxonomy: clean up vigno bean confusion

paper42:
* Vigna radiata is the plant producing green mung beans (green gram)
  * mung dal is mung bean split
  * yellow mung dal is split and hulled mung bean which gives it a yellow color
* Vigna mungo is a relative of Vigna radiata which produces black gram
  * ivory white lentils is hulled and split black gra which removes the black skin and gives it white colorm, urad dal seems to be a name used for both black gram and ivory white lentils

### DIFF
--- a/taxonomies/food/ingredients.txt
+++ b/taxonomies/food/ingredients.txt
@@ -62263,7 +62263,7 @@ ifct_food_process:en: en:dried
 ###################################################################################################
 
 < en: pulse
-en: mung bean, mungbean, Mung dal, Moong dal, green gram
+en: mung bean, mungbean, green gram
 ar: بقلة الماش
 bg: птичи боб, бебриджа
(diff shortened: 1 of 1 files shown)


paper42:
now, two items have Urad Dal as an english identifier because that term seems to be used for both, is that ok or will that cause issues?

Freso:
> now, two items have Urad Dal as an english identifier because that term seems to be used for both, is that ok or will that cause issues?

It will cause issues. It looks like one of them is a child of the other, so it would probably make sense to remove it from the child entry.

aleene:
remove the urad dal synonym where it does not apply.

paper42:
I removed it from black gram, thanks :)

stephanegigandet (approved):
Thanks

PR #14443, opened 2026-09-03
fix: escape ampersands in facet links Fixes #14433

### LINKED ISSUE #14433: Ampersands (`&`s) aren’t escaped in (some?) facet links
### What
- Brand link on https://se.openfoodfacts.org/product/6006507008840/korean-style-bbq-rub-cape-herb-spice for `https://se.openfoodfacts.org/facets/brands/cape%20herb%20&%20spice` leads to `https://se.openfoodfacts.org/facets/brands/cape%20herb?%20spice`

1. Go to https://se.openfoodfacts.org/product/6006507008840/korean-style-bbq-rub-cape-herb-spice
2. Click on “cape herb & spice” (the brand)
3. Land on an error page

### Expected behavior
- Expected to land on the page for “cape herb & spice” so I could see how many products it had.

### Additional context
- This might be browser specific behaviour, trying to translate the non-escaped `&` to a GET request parameter (with `?`).
- I haven’t tested whether the behaviour is the same with items that exist in the taxonomy with an `&` or if it only happens if they’re not known in the taxonomy.

### Type of device
- Browser: Firefox 154.0.1
- Operating System: Linux

SurbhiAgarwal1:
### What
Facet links for tags containing ampersands (e.g. brand `cape herb & spice`) were rendering raw `&` characters in the URL path (`/facets/brands/cape herb & spice`), causing browsers and servers to misinterpret `&` as a query parameter separator (`?`).

This PR updates `display_taxonomy_tag_link` and `canonicalize_taxonomy_tag_link` in `lib/ProductOpener/Tags.pm` to pass tags through `get_tag_url_id`, ensuring special characters like `&` are properly URI-encoded as `%26`. Added unit tests in `tests/unit/tags.t` to verify URL encoding for facet links containing `&`.

N/A

### Screenshot or video
N/A

- Fixes: #14433
- Related to: openfoodfacts/openfoodfacts-query#362

## Summary by CodeRabbit

- **Bug Fixes**
  - Facet, category, brand, label, country, and ingredient links now safely encode spaces, accents, colons, ampersands, and other special characters.
  - Improved normalization of city-related taxonomy links.
  - Link text and displayed product information remain unchanged.

- **Tests**
  - Added coverage verifying URL encoding for special characters in taxonomy links.
  - Updated integration and unit test expectations for encoded URLs.


hangy (approved):

lib/ProductOpener/Tags.pm
```diff
@@ -3095,6 +3095,7 @@ sub canonicalize_taxonomy_tag_link ($target_lc, $tagtype, $tag, $tag_prefix = un
 
 	$target_lc =~ s/_.*//;
 	$tag = display_taxonomy_tag($target_lc, $tagtype, $tag);
+	$tag =~ s/&/%26/g;
```
Freso: is there no more general purpose function that could be used for this? this would both have the logic be only one place (instead of repeating again on line 3115), and could also catch [cases other than `&` that might have issues](https://en.wikipedia.org/wiki/Percent-encoding#Reserved_characters). i don't know if we have a library or something we already use for this purpose, or if there’s already a function somewhere in the codebase.

Freso (commented):
Looks fine, but it would be good to have a more general solution than targeting just that one single character. :)

lib/ProductOpener/Tags.pm
SurbhiAgarwal1: Hi @Freso, thanks for the suggestion!

Refactored `display_taxonomy_tag_link` and `canonicalize_taxonomy_tag_link` in `lib/ProductOpener/Tags.pm` to use `get_tag_url_id($tagtype, $tag)`, the central helper for tag URL formatting in Open Food Facts. 

This replaces ad-hoc regex escaping with `encodeURIComponent` from `URI::Escape::XS`, ensuring all special characters (including `&` and spaces) are consistently URI-encoded across all taxonomy facet links.

Freso (approved):
thanks! looks good to me! i'll leave someone else to merge it though, just in case there's some issue i'm not seeing :)

SurbhiAgarwal1:
Hi @hangy ,@Freso 

Thanks again for the reviews! I wanted to flag that the 🐪 Perl Integration Tests (5) check is still failing (exit code 2) on the latest commit, even though the unit tests in `tests/unit/tags.t` are passing now.

Looking at the failing group, it covers `api_current_user_permissions.t`, `api_v3_product_services_external.t`, `convert_and_import_excel_file.t`, `external_sources.t`, `page_crawler.t`, and `search_v1.t`, none of which seem directly related to the `display_taxonomy_tag_link` / `canonicalize_taxonomy_tag_link` changes in this PR.

Could someone help me understand a few things?

1. Is this integration test group known to be flaky or unstable, or does it usually pass reliably?
2. Is there a way for me to see the actual test failure output, or is that only visible to maintainers or org members?
3. If it's not flaky, could this be an indirect effect of URI-encoding the tag in `get_tag_url_id`, where some other code path might be expecting an unencoded tag string?

I really appreciate you both taking the time to review this already. Happy to dig further once I can see what's actually failing, just want to make sure I'm not missing something in the diff. Thanks so much for your patience!

lib/ProductOpener/Tags.pm
```diff
@@ -3095,9 +3095,11 @@ sub canonicalize_taxonomy_tag_link ($target_lc, $tagtype, $tag, $tag_prefix = un
 
 	$target_lc =~ s/_.*//;
 	$tag = display_taxonomy_tag($target_lc, $tagtype, $tag);
-
+	my $tagurl = $tag;
+	$tagurl =~ s/ /%20/g;
+	$tagurl =~ s/&/%26/g;
```
Freso: please don’t listen to the cruddy “ai”, this is just yet another case of “ai” not understanding an issue. 🙄 if `get_tag_url_id()` encodes _too much_, then that’s something that needs fixing in `get_tag_url_id()`, not something that should be “fixed” by not using `get_tag_url_id()` as `get_tag_url_id()` is meant to be used. (and i’m not convinced by this either way. sometimes code changes mean tests need to be updates, and that’s okay. we have semi-automated ways of doing this, which i can set in motion when you've undone this “ai” “fix”.)

Freso (changes requested):
most of the failing tests should be possible to easily regenerate using `/update_tests_results` in a comment. “ai” doesn’t know what it’s talking about, it’s just making up stuff that sounds plausible.

SurbhiAgarwal1:
Hi @Freso, I've undone the previous AI fix and restored get_tag_url_id($tagtype, $tag) in canonicalize_taxonomy_tag_link as requested. Could you please trigger /update_tests_results when convenient? Thanks!

lib/ProductOpener/Tags.pm
SurbhiAgarwal1: Hi @Freso, I've undone the previous AI fix and restored get_tag_url_id($tagtype, $tag) in canonicalize_taxonomy_tag_link as requested. Could you please trigger /update_tests_results when convenient? Thanks!

Freso (approved):
Looks good to me now! I’ll wait for someone else’s review before merging though. Thank you!

hangy (approved):

PR #14450, opened 2026-09-03
taxonomy(food): add sötstark mustards category

Freso:
It doesn’t seem like this type of mustard has an established English name/translation, but the ones used here seem to be some of the ones in use, according to some quick searching.

Needed-for: https://se.openfoodfacts.org/product/7310155600516/s%C3%B6tstark-senap-johnny-s
Needed-for: https://world-sv.openfoodfacts.org/cgi/search.pl?search_terms=S%C3%B6tstark
Needed-for: https://world.openfoodfacts.org/cgi/search.pl?search_terms=S%C3%B6tstark

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -79920,6 +79920,11 @@ nl: Zoete mosterds
 pl: Musztardy słodkie
 
 < en: Mustards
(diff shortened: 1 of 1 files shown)


teolemon (commented):
Thank you @Freso 👍

teolemon (approved):
thanks @Freso

PR #14453, opened 2026-09-03
taxonomy(food): add sv “no nuts” label synonym

Freso:
Needed-for: https://se.openfoodfacts.org/product/7340083479434/n%C3%B6tfri-pesto-di-basilico-garant


teolemon (approved):
thanks @Freso

PR #14470, opened 2026-09-04
taxonomy: add more ZH-language food category translations

jayaddison:
### What
This is a follow-up to an import provided in pull request #14336; a [bug in the import script](https://codeberg.org/openculinary/openfoodfacts-data-utils/issues/2) used to create that PR meant that some translations were omitted (specifically: cases where the target product had a list of synonyms in the identifier/primary language).

This is done to improve the translation coverage of categories in Chinese, and to complete the work started in #14336.

The [latest version](https://codeberg.org/openculinary/openfoodfacts-data-utils/src/commit/82ebfaa405cf28ac083af3a72de662712638a239/category-taxonomy-translation) of the import script, used to merge the translations, includes improvements on the language-code and property sorting behaviour compared to #14336.

I did not use any LLMs, neither to develop the import script, nor to identify/fix the cause of the bug in the script, nor to create this pull request.

@meowdolee used machine translation as an initial batch/draft process on the food categories presented here, and then followed that by manually reviewing each entry (due to a ~%50 accuracy in those automated machine translations).

Follow-up to #14336

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -1263,6 +1263,7 @@ pt: Produtos des abelhas, produtos de colmeia
 ro: Produse apicole, Produse de apicultură, Produse apicultură, Produse de apicultura, Produse apicultura
 ru: Продукция пчеловодства, Продукты пчеловодства
 tr: Arı ürünleri
+zh: 蜂产品
 wikidata_category:en: Q7189914
 
 < en: Bee products
@@ -1436,6 +1437,7 @@ it: Bevande e preparati per bevande
 nl: Dranken en drankbereidingen
 pl: Napoje i składniki do napojów
 pt: Bebidas e preparações para bebidas
+zh: 饮料及饮料备制品
 description:en: This mixes ready to drink beverages and not-ready to drink beverages
 incompatible_with:en: categories:en:meals
 
@@ -1669,6 +1671,7 @@ ro: Băuturi dulci, Băuturi cu zahăr, Bauturi dulci, Bauturi cu zahar
 ru: Напитки с добавлением сахара
 th: น้ำหวาน
 tr: Tatlı içecekler, Şekerli içecekler
+zh: 含糖饮料
 food_groups:en: en:sweetened-beverages
 pnns_group_2:en: Sweetened beverages
 
@@ -1691,6 +1694,7 @@ pl: Napoje sztucznie słodzone, Napoje ze sztucznymi słodzikami
 pt: Bebidas artificialmente doçeadas
 ro: Băuturi îndulcite artificial, Băuturi cu îndulcitori artificiali, Bauturi indulcite artificial, Bauturi cu indulcitori artificiali
 th: เครื่องดื่มผสมน้ำตาลเทียม
+zh: 人工甜味剂饮料
 food_groups:en: en:artificially-sweetened-beverages
 pnns_group_2:en: Artificially sweetened beverages
 
@@ -1713,6 +1717,7 @@ pt: Bebidas não adoçadas, Bebidas não açucaradas, bebidas sem açucar adicio
 ro: Băuturi fără zahăr, Băuturi fără zahăr adăugat, Bauturi fara zahar, Bauturi fara zahar adaugat
 ru: Напитки без добавления сахара
 th: เครื่องดื่มไม่เติมน้ำตาล
+zh: 无糖饮料
 food_groups:en: en:unsweetened-beverages
 pnns_group_2:en: Unsweetened beverages
 
@@ -2240,6 +2245,7 @@ description:en: Beers that are packed for example: 6x330ml
 en: Ales, Ale
 xx: Ales, Ale
 lt: Elis
+zh: 艾尔啤酒
 wikidata:en: Q208385
 
 < en: Ales
@@ -2267,6 +2273,7 @@ en: American-style IPA, American IPA
 xx: American-style IPA, American IPA
 lt: Amerikietiškas IPA
 pt: IPA estilo americano
+zh: 美式IPA啤酒
 
 < en: India Pale Ale (IPA)
 en: Cornish IPA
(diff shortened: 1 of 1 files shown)


jayaddison:
There is one ambiguity/duplicate reported by the automated tests: https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/33886504088/job/101067469515#step:15:61

> `[categories] [2026/09/04 14:58:58] [ERROR] ERROR - duplicate_synonym - 122584 - zh:肉类替代品 already is associated to en:meat-alternatives (categories) - zh:肉类替代品 cannot be mapped to entry en:meat-analogues`

After reading the comments in `categories.txt`, I think that:

* `Meat alternatives` are situations where meat in existing recipes can be substituted with a non-meat ingredient that _may or may not_ attempt to look/taste similar to the original.
* `Meat analogues` are similar, except that the substitute ingredient _does_ explicitly attempt to be similar to the original.

So if I cook stir-fry with `tofu` in a recipe that describes `chicken`, I am using a `meat alternative`.

On the other hand, when I ask for `vegetarian sausages` in my breakfast bread roll, those sausages are a `meat analogue` (I know that it can be contentious in some cases whether it is sensible/appropriate to use the same term to describe them -- however, this is an example of the genuine terms used in a few situations I've encountered, and have benefitted from a simple way to choose a non-meat option).

jayaddison:
Ah, one more contextual detail that could help for disambiguation: `meat analogues` are a subcategory (more precise definition) of `meat alternatives`.

So: every `meat analogue` (`vegetarian burger`) is a `meat alternative`, but not every `meat alternative` (`falafel`) is a `meat analogue`.

Freso:
I think that reasoning is sound, yeah. @moon-rabbitOFF might have some input too.

taxonomies/food/categories.txt
```diff
@@ -59535,6 +59576,7 @@ ciqual_food_name:fr: Matière grasse végétale ou margarine, 80% MG, doux
 < en: Unsalted margarines
 en: 70% unsalted vegetable fat, 70% fat salted margarine
 fr: Matière grasse végétale à 70% doux, Margarine douce à 70% MG
+zh: 70% 无盐植物脂肪
```
Freso, line 59579: Sometimes there’s a space between “%” and whatever comes after, and sometimes there’s not. Is this intentional, or could/should this be standardised/normalised? And should the other variant be added as a synonym maybe?

Freso (approved):
FWIW, not being a Chinese speaker/reader/writer, this looks good to me. a question about percentage signs and spacing, but this is still an improvement as it stands either way.

Freso:
@jayaddison (or maybe @Meowdolee?), any chance of a response to https://github.com/openfoodfacts/openfoodfacts-server/pull/14470#discussion_r3937631229 ? :)

taxonomies/food/categories.txt
Meowdolee: Hi @Freso ,

In Chinese writing, we put spaces around foreign words, English terms, and numbers due to Chinese typography and mixed-language typesetting, rather than Chinese grammar. This makes the text easier for native Chinese speakers to read.
Thus, there should be a space between the % and 无盐植物脂肪.

For example, “There are 5,000 users.” we would write:
有 5000 個使用者。
We put spaces around the number.

Another example would be:
"I use Photoshop to process photos."
我用 Photoshop 處理照片。

The space around “Photoshop” helps distinguish the English term from the surrounding Chinese characters.
Let me know if this makes sense!

Meowdolee:
Hi team,

Sorry for the late reply! I was deep in the mountains on a backpacking trip for my birthday last week, so I was mostly offline. I’ve been trying to work my way through all the emails that accumulated while I was away, and I finally got to yours.

Thanks to @jayaddison for the detailed explanation. In this case, I would translate them as follows:
meat analogue: 植物肉
meat alternative: 肉類替代物

taxonomies/food/categories.txt
Freso: That makes sense, but shouldn’t there then also be a space between “80%” and “无盐植物性脂肪”/“含盐植物性脂肪” in “80%无盐植物性脂肪” and “80%含盐植物性脂肪”? Or is that different somehow?

Some other ones that might need adjusting: “Rosacker产区阿尔萨斯特级葡萄酒”, “美式淡色艾尔啤酒（APA）” (though maybe this one is okay because of the parentheses?), “美式IPA啤酒”, “含葵花籽的60%无盐植物脂肪”, “30-40%脂肪量 无盐植物脂肪”, and “2-3%脂肪量 小瑞士奶酪”.

jayaddison:
Thanks again @Meowdolee - I have added an update here including your suggested translations

taxonomies/food/categories.txt
Meowdolee: Hi @Freso ,

Sorry for the mix-up here! As @jayaddison mentioned, we should just keep the spacing around English text, numbers, and symbols. I’ll make sure to include the appropriate spacing in my future inputs.
Thanks for the clarification, and sorry again for the confusion.

taxonomies/food/categories.txt
```diff
@@ -122475,7 +122563,7 @@ fr: Alternatives à la viande
 hr: Mesne alternative
 nl: Vleesvervangers
 pl: Alternatywy mięsa
-zh: 肉类替代品
+zh: 肉類替代物
```
Freso, line 122566: if 肉类替代品 is not completely off the mark, it might be good to keep it as a synonym. (i see that at least 肉 and 替代 are the same (and 类 seems to be a component of 類?), but i don’t know how much weight that holds.)
```suggestion
zh: 肉類替代物, 肉类替代品
```

Meowdolee: Hi @Freso,

I believe there may have been a mix-up between Traditional Chinese and Simplified Chinese during my input.
For this one, let’s just use 肉类替代品.

Thanks!

teolemon: @Freso can you apply the suggestion. Otherwise, can we merge ?

Freso: I think @jayaddison has not allowed maintainers to make changes to the PR’s branch, so I can’t. 🫤 I guess we can merge and I can make a follow-up for making “肉类替代品” the primary.

Freso: Follow-up: https://github.com/openfoodfacts/openfoodfacts-server/pull/14575

jayaddison: Ah, sorry about that: it is because the pull request was made from a GitHub organisation account (`openculinary`).  No such option exists to allow edits from maintainers in that case.

Thanks for handling the fixup @Freso, and thanks @Meowdolee for the clarification.

PR #14471, opened 2026-09-04
fix: Prepare language codes and catalogs for variants

TaciteOFF:
### What

This prepares the shared code for language variants :) Language codes and catalog fallbacks now have one implementation in `I18N.pm`. No new locale is enabled.

`$language_code_re` defines the syntax used by the normalizer, PO loader and future taxonomy changes. It supports two- or three-letter languages, scripts and regions. Catalog filenames must use the normalized spelling, such as `pt_BR.po` or `zh_Hant_TW.po`.

The shared helpers normalize codes, format BCP-47 tags, return the root language and look up values through the fallback chain. For example, `zh_Hant_TW` tries `zh_Hant` before `zh`. English fallback stays with the caller, since ingredient matching and interface translations have different rules.

Catalog compilation and JSON export use those helpers. JSON directories use BCP-47 names. Calendar locales keep the Chinese script mappings, and string normalization inherits the base language's rules.

Tag catalogs can omit `:langname` and `:langtag`, as the real catalogs do. `split_tags` leaves missing keys out, and `build_lang_tags` skips undefined entries. A test covers this case so the shared fallback code doesn't break `build_lang.pl`.

This changes production code in `I18N.pm`, `Lang.pm` and `Store.pm`. The current registry, request handling, language selector and templates are unchanged. No locale is activated.

### Tests

`scripts/build_lang.pl` finishes with the real catalogs and writes both stored catalogs and the JSON exports. No `pt-BR` directory is generated on this branch.

Four focused test files, 340 assertions: all passed (`language_codes.t`, `i18n_language_variants.t`, `lang_tags.t`, `lang.t`). They cover the shared helpers, strict filenames, script fallback, missing language-name keys, catalog compilation, JSON export and fresh-process reload.

The translation and calendar data were also compared with the original preparation commit: all 189 registered languages, their names, translations and calendars are identical. There are no regional languages in that registry.

All 306 tracked PO paths were checked with the shared normalizer: 153 in `po/common` and 153 in `po/tags`, including the `nl.po` symlink. Every filename already has the canonical spelling. None is rejected by the filename filter, and no normalization collisions were found.

Perltidy, Perlcritic and `git diff --check` passed.

Related to #6688, which stays open.

### DIFF
--- a/lib/ProductOpener/I18N.pm
+++ b/lib/ProductOpener/I18N.pm
@@ -39,6 +39,12 @@ The functions used in this module take the directory to look for the .po files a
 package ProductOpener::I18N;
 
 use ProductOpener::PerlStandards;
+use Exporter qw/import/;
+
+our @EXPORT_OK = qw/$language_code_re normalize_language_code language_tag base_language
+	language_fallbacks lookup_with_language_fallback/;
+our %EXPORT_TAGS = (all => \@EXPORT_OK);
+our $language_code_re = qr/[a-z]{2,3}(?:[-_][a-z]{4})?(?:[-_](?:[a-z]{2}|[0-9]{3}))?/iaa;
 
 use File::Basename;
 use File::Find::Rule;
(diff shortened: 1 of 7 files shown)


lib/ProductOpener/I18N.pm
```diff
@@ -79,24 +85,144 @@ my @metadata_fields = qw<
...
+
+The internal spelling (C<pt_BR>, C<zh_Hant_TW>), or undef for unsupported syntax.
+This does not register or enable a language.
+
+=cut
+
+sub normalize_language_code ($code) {
+	return if not defined $code or $code !~ /\A$language_code_re\z/;
+	my ($language, @subtags) = split /[-_]/, $code;
+	return join('_', lc($language), map {length($_) == 4 ? ucfirst(lc($_)) : uc($_)} @subtags);
```
alexgarel, line 108: this kind of line is very Perl idiomatic, but hard to follow for beginners in Perl. I always prefer if they are commented.


```suggestion
	# join main language code, and normalization of subtags by '_'
	return join('_', lc($language), map {length($_) == 4 ? ucfirst(lc($_)) : uc($_)} @subtags);
```

lib/ProductOpener/I18N.pm
```diff
@@ -79,24 +85,144 @@ my @metadata_fields = qw<
...
+=head3 Return values
+
+The first defined value in the language's fallback chain, or undef. Zero and
+empty strings are values too. The hash is not modified, and English fallback is
+left to the caller.
+
+=cut
+
+sub lookup_with_language_fallback ($hash_ref, $code, $matched_code_ref = undef) {
+	$$matched_code_ref = undef if defined $matched_code_ref;
```
alexgarel, line 190: There again a comment to help non Perl mongers

```suggestion
	# $$matche_code_ref is the value behind $matched_code_ref pointer
	$$matched_code_ref = undef if defined $matched_code_ref;
```

lib/ProductOpener/I18N.pm
```diff
@@ -111,16 +237,22 @@ sub read_po_files ($dir) {
...
-		local $log->context->{file} = basename($file);
+		my $filename = basename($file);
+		local $log->context->{file} = $filename;
 		$log->debug("Reading po file");
 
 		my $lc;
-
-		if ($file =~ /\/(\w\w).po/) {
-			$lc = $1;
+		if ($filename =~ /\A($language_code_re)\.po\z/) {
```
alexgarel, line 248: I think

```suggestion
		if ($filename =~ /^($language_code_re)\.po$/) {
```
is less precise but more readable to most devs

lib/ProductOpener/I18N.pm
```diff
@@ -111,16 +237,22 @@ sub read_po_files ($dir) {
...
+			$lc = normalize_language_code($1);
 		}
 		else {
-			$log->debug("Skipping file (not in [2-letter code].po format)");
+			$log->debug("Skipping file (not in language.po format)");
+			next;
+		}
+		next if $filename ne "$lc.po";
```
alexgarel: it's strange that there we skip silently… while we do not above.
For me it's even a warning !


```suggestion
		if ($filename ne "$lc.po") {
			$log->warning("Skipping file (not in lc.po format)");
			next;
		}
```

alexgarel (approved):
LGTM @TaciteOFF, only some style comments

That's really great.

lib/ProductOpener/I18N.pm
TaciteOFF: Thanks. I will put this comment:
`# language lowercase (pt), 4-letter script in Title case (Hant), region uppercase (BR / 419), joined with '_'.`

lib/ProductOpener/I18N.pm
TaciteOFF: There's a typo in your suggested change (matche / matched). 
I will put this comment:
`# reset the caller's out-parameter (scalar reference) before the lookup.`

lib/ProductOpener/I18N.pm
TaciteOFF: The same `\A$language_code_re\z` anchoring is used at line 105 in normalize_language_code. There it actually matters: $ allows a trailing newline, so normalize_language_code("fr\n") would return "fr\n" as a "canonical" code. That function is meant to handle taxonomy prefixes and, eventually, user-controlled input (lc=, Accept-Language), so I'd rather keep the strict anchors there.

On line 245 the looser form would indeed be harmless — line 252 (next if $filename ne "$lc.po") rejects "fr.po\n" anyway. But having ^...$ in one place and \A...\z in the other in the same file seems worse than either option applied consistently.

I will add this comment:
` # \A and \z anchor the whole string; unlike $, \z does not allow a trailing newline`

lib/ProductOpener/I18N.pm
TaciteOFF: Agreed, this one is a real gap. 
 I will use warn rather than warning to match the convention in the rest of the repo.

PR #14473, opened 2026-09-05
chore: migrate from unmaintained tomchv/wait-my-workflow to lewagon/wait-on-check-action

### LINKED ISSUE #14421: Migrate github actions away from wait-my-workflow
### What

- Migrate github actions away from wait-my-workflow
- We [have asked for an update](https://github.com/TomChv/wait-my-workflow/issues/423). No answer.
- [Robotoff has already moved away](https://github.com/openfoodfacts/robotoff/pull/1580)
- https://github.com/openfoodfacts/openfoodfacts-server/blob/main/.github/workflows/release-please.yml

release-pleaseProcess completed with exit code 1.
--
release-pleaseNode.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: tomchv/wait-my-workflow@v1.1.0. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
release-pleaseThe `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
release-pleaseThe `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

[release-please](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/33509373876/job/99861057982#step:8:46)
Process completed with exit code 1.
[release-please](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/33509373876/job/99861057982#step:13:2)
Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: tomchv/wait-my-workflow@v1.1.0. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
[release-please](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/33509373876/job/99861057982#step:3:19)
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
[release-please](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/33509373876/job/99861057982#step:2:19)
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

ganga-300:
## What

Fixes #14421 - migrates away from `tomchv/wait-my-workflow`, an unmaintained GitHub Action that is currently generating deprecation warnings and has received no response to a maintainer update request ([[TomChv/wait-my-workflow#423](https://github.com/TomChv/wait-my-workflow/issues/423)](https://github.com/TomChv/wait-my-workflow/issues/423)).

Robotoff (OFF's sister project) already migrated away from this same action ([[openfoodfacts/robotoff#1580](https://github.com/openfoodfacts/robotoff/pull/1580)](https://github.com/openfoodfacts/robotoff/pull/1580)), confirming this is a known, shared problem across OFF's repos.

## Root cause

`tomchv/wait-my-workflow@v1.1.0` is used to poll for the completion of two named GitHub check runs (`build (frontend)`, `build (backend)`) before proceeding with release/deploy steps. The action is unmaintained and internally still targets Node.js 20 (now deprecated on GitHub's runners) and uses the deprecated `set-output` command syntax instead of modern `$GITHUB_OUTPUT` environment files.

Confirmed directly against the actual failing run linked in the issue ([[run 33509373876](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/33509373876/job/99861057982)](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/33509373876/job/99861057982)):
- **"Complete job"** annotation: `Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: tomchv/wait-my-workflow@v1.1.0`
- **"Wait for frontend/backend container build workflow"** annotations (x2): `The set-output command is deprecated`

**Important scoping note:** that same run also shows a genuine job **failure** (`exit code 1`) at a *different*, later step - `"Get JS dist files to add to the release"` (a `docker pull`/extraction step). This failure is **unrelated** to `wait-my-workflow` and is **not addressed by this PR**. I'm flagging this explicitly so it isn't assumed fixed - it looks like a separate, pre-existing problem (possibly a timing/registry issue) that would need its own investigation.

## Fix

Replaced `tomchv/wait-my-workflow@v1.1.0` with [`[lewagon/wait-on-check-action@v1.9.1](https://github.com/lewagon/wait-on-check-action)`](https://github.com/lewagon/wait-on-check-action), an actively maintained action supporting the same core use case (poll for a named check run on a given ref).

**Files changed:**
- `.github/workflows/release-please.yml`
- `.github/workflows/container-deploy.yml` (found while investigating - same unmaintained action, same problem, not mentioned in the original issue but clearly in scope)

### Why this specific replacement, and what changed structurally

Unlike `wait-my-workflow`, `lewagon/wait-on-check-action` fails its own step directly when the check doesn't succeed, rather than requiring the caller to branch manually on an output value. This let me simplify `release-please.yml`: the old three explicit `if: steps.wait-build.outputs.conclusion == '...'` branch steps (for `does not exist` / `failure` / `timed_out`) are no longer needed, since GitHub Actions implicitly skips all subsequent steps once a prior step fails (steps' `if:` conditions carry an implicit `&& success()` unless overridden).

### Inputs used, and why

```yaml
uses: lewagon/wait-on-check-action@v1.9.1
timeout-minutes: 20   # 10 for container-deploy.yml, matching its original 600s timeout
with:
  ref: ${{ env.RELEASE_SHA }}
  check-name: build (frontend)   # or build (backend)
  repo-token: ${{ secrets.GITHUB_TOKEN }}
  wait-interval: 10
  checks-discovery-timeout: 300
  allowed-conclusions: success
```

- `checks-discovery-timeout: 300` - **not present in the original**, added deliberately. Since this job and the container-build workflow both trigger on the same push event, there's a real race where this step could start polling before the corresponding build check has even been created. This input tells the action to keep waiting (up to 5 minutes) for the check to *appear* at all, instead of immediately failing with a "check never ran" error - a documented gotcha with this action if left at its default.
- `timeout-minutes` on the step itself replicates the original's overall timeout (`timeoutSeconds: 1200` / `600`), since the action doesn't have its own single "give up entirely" input in the same way the old one did.

## Behavior changes for reviewers to sanity-check

- **`release-please.yml`**: behavior is preserved as-is - missing/failed/timed-out check still fails the job, same as before.
- **`container-deploy.yml`**: **one intentional behavior change**. The original explicitly treated a *missing* check as a non-fatal warning (`echo job does not exist && true` - note `true`, not `false`), only failing on an actual `failure` or `timed_out` conclusion. With the new action, a missing check now blocks deployment the same way a failed check would. I believe this is the more correct behavior - deploying without ever confirming the build succeeded seems riskier than the original's "warn and continue" - but it's a real, deliberate change in outcome, not a transparent drop-in swap, so I wanted to call it out rather than have it slip through unnoticed.

## Testing

The affected code paths only run under specific conditions that can't be exercised by this PR's own CI:
- `release-please.yml`'s wait logic only runs when `IS_RELEASE` is true (i.e. on an actual release-please-generated commit).
- `container-deploy.yml` only runs on a real push to `main`.

To verify the replacement action itself works correctly against this repo's real check data, I temporarily added a manual `workflow_dispatch` test workflow (since removed, not part of this PR) and ran it against real commits on my fork:

- [x] Verified it correctly finds and confirms an already-passing `build (frontend)` check (tested against two separate real commits)
- [x] Verified it correctly finds and confirms an already-passing `build (backend)` check
- [x] Confirmed the action's own verbose logging correctly lists and filters all in-progress/completed checks on a ref before resolving the named one, e.g.:
  ```
  Checks completed: build (backend)
  build (backend): completed (success)
  ```

I was **not** able to directly test the failure path (couldn't find a real failed `build (frontend)`/`build (backend)` check on a commit to test against) - noting this as a known verification gap rather than claiming full coverage. The action is otherwise widely used in production across many other repos, and its pass/fail logic is straightforward and transparent in its own logs, as shown above.

## Checklist
- [x] Confirmed no remaining references to `tomchv/wait-my-workflow` anywhere in the codebase (`grep -rn "wait-my-workflow" .github/workflows/*.yml` returns nothing)
- [x] Verified the replacement action against real check data on two independent commits
- [x] Documented the one real behavior change (`container-deploy.yml`'s "missing check" handling) rather than letting it pass silently
- [x] Clarified which part of the original bug report this PR does and does not address


teolemon (commented):
good. @hangy are you happy ?

hangy:
> good. @hangy are you happy ?

Looks okay on the surface. However,I don't really know this workflow.

john-gom:
Looks like Robotoff switched to this workflow on this PR: https://github.com/openfoodfacts/robotoff/pull/1580

@raphael0202 has it been working OK for you?

Looks like lots of other places should be fixed if we want to be consistent: https://github.com/search?q=org%3Aopenfoodfacts+wait-my-workflow&type=code

alexgarel (approved):
Thanks @ganga-300 ! This is precious.

PR #14474, opened 2026-09-05
ci: Pin GitHub Actions versions

hangy:
### What

OpenSSF Scorecard recommends using pinned dependencies. We can do that for GitHub Actions, and do so for other `openfoodfacts/*` projects.

None.

### Screenshot or video

None.

None.


teolemon (approved):

jayaddison:
@hangy @teolemon I think this is OK to do, but please be aware that there is one downside that I'm aware of: pinning GitHub Actions versions means that GitHub Dependabot Alerts will not detect/alarm about vulnerabilities in those dependencies (because it uses semver range comparison to identify whether an action in a workflow file is vulnerable).

It's a documented limitation in GitHub's Dependabot documentation at: https://docs.github.com/en/code-security/concepts/supply-chain-security/dependabot-alerts#limitations

> "For GitHub Actions, alerts are only generated for actions that use semantic versioning, not SHA versioning."

There is a GitHub community discussion thread to request support for GitHub Dependabot Alerts functionality for SHA-pinned actions (thread title: `Please make Dependabot Alerts support SHA-hash-versioned GitHub Actions`) at: https://github.com/orgs/community/discussions/154189

PR #14478, opened 2026-09-06
taxonomy(food): de:Labaustauschstoff

aleene:
### What
This is the non-animal variant of Lab. Question from Nutri-Patrol


Freso (approved):
thanks!

PR #14485, opened 2026-09-07
taxonomy(food): soy ice cream categories

Freso:
- Adds vanilla and strawberry soy ice cream categories.
- Removes “Chocolate soy desserts” parent from chocolate soy ice cream.

Needed-for: https://se.openfoodfacts.org/product/8714100083352/vanilla-flavoured-soya-ice-cream-choice
Needed-for: https://se.openfoodfacts.org/product/8714100079249/refreshing-strawberry-choice

### DIFF
--- a/taxonomies/food/categories.txt
+++ b/taxonomies/food/categories.txt
@@ -67101,14 +67101,21 @@ nl: Plantaardige ijswafels
 
 < en: Plant-based ice creams
 < en: Soy desserts
-en: Soy-based ice creams
+en: Soy-based ice creams, Soy ice creams
 sv: Sojaglasser
 wikidata:en: Q133892366
 
-< en: Chocolate soy desserts
 < en: Soy-based ice creams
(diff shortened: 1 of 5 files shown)


moon-rabbitOFF:
Flavoured soy dessert was created for creamy pudding-like dessert. It is not really clear, I will change the category name. So the two new ice cream categories would go only under soy-based ice cream and not flavoured soy dessert.

In France (and maybe in EU?) the word "crème" (cream) can only be used for dairy product, which is why those puddings are called "dessert au soja" and not "crème dessert au soja".

Update: I made a pull request https://github.com/openfoodfacts/openfoodfacts-server/pull/14488

teolemon (approved):
thanks @Freso

Freso:
@moon-rabbitOFF I removed the parent from the chocolate one and didn't make one for the strawberry one. did i understand your comment correctly by doing this?

PR #14495, opened 2026-09-07
build: add cache busting mechanism to Docker builds

hangy:
### What

We noticed that using the `buildcache` can be relatively old, causing updates from `trixie-security` not being picked up by the `backend` image. This PR adds a deliberate cache busting argument, which enables us to get a completely fresh build with an updated base image and `apt` updates once in a while.

Right now the idea is to use the ISO week, so that each first build in a business week invalidates the cache and picks up security & package updates, without the rest of the layers needing to be changed.

None

### Screenshot or video

None

- https://openfoodfacts.slack.com/archives/C02LDQDDD/p1788777840462189

## Summary by CodeRabbit

* **Chores**
  * Container builds now refresh their build cache on a regular weekly schedule.
  * Builds pull the latest available base images, helping ensure container updates include current image changes.
  * Frontend and backend container builds support cache-refresh triggers for more reliable rebuilds.


alexgarel (approved):
Thanks, it's a good idea indeed

PR #14501, opened 2026-09-08
taxonomy: modifications of some sauces

moon-rabbitOFF:
* created whole tahini
* put maafe/peanut stew under peanut sauces


Freso (approved):
What is "whole tahini"? :)

PR #14502, opened 2026-09-09
fix: warn and skip edit on producer protected fields (#14492)

### LINKED ISSUE #14492: Several issues regarding data provided by the producer
### What

I have a product where the logo that values were provided by the producer and as such I might not be able to modify it. From there, there's both a UX issue and a data integrity issue (it might be necessary to split this issue in two, if you feel so).

First I'll address the UX problem: aside from this logo and it's explanation, there's nothing different from products where we won't have data coming from the producer. Application allows me to eg. change the ingredients list, and even to submit it. From there , it looks like I could change them, but if I go back at modifying ingredients, then the old list is back, my change is gone, and I had zero feedback from the UX about it: no error message, information, message telling me it have to be check by a moderator/admin or anything.

Now, you'll tell me if producer provided data, then I have no reasons to modify those. Well… it looks like some of them provided them in the past, but didn't kept them up-to-date, and for example the ingredients list isn't exactly the same it was. Worse: I could update the photo, so now the photo shows a list different than OFF data and I can't do anything about it.

Finally a little digression on the topic: it's not the first product I encounter where recipe was changed and data are obsolete, and as such, and because there isn't a centralized mandatory database, I don't get how a change in recipe doesn't mandate a barcode change too? One could argue this is a different version of the product. I guess they need authorization from an authority to be sell those products; a different recipe should result in a different authorization and as such a different barcode…?
The result is that user consult obsolete information and have no obvious way to know those are incorrect.

N/A , description above should be enough, but if it isn't I'll gladly provide what's additionally requested.

### Expected behavior

- Be able to change obsolete data
- Have a UX feedback if changes aren't allowed/cause an error
- Better: don't even let the user try if the changes are going to be rejected anyway (it's frustrating to lose your time that way)
- If user won't be able to change ingredients or nutritional values, then don't let them change the corresponding picture.

### Why

UX and data integrity issue.

### Part of
- openfoodfacts/smooth-app#525  

### Smartphone model
 - Device: Moto G84
 - OS: Android 15
 - App Version: 4.23.0

SurbhiAgarwal1:
### What
When product data contains fields provided by the producer (recorded in `owner_fields` or under organization data protection), non-moderator user edits to those fields are skipped with explicit warning feedback returned in the API response.

Specifically:
- Updated `is_owner_field` in `lib/ProductOpener/Products.pm` to resolve localized field keys (e.g., `ingredients_text_fr`) to their base field names (`ingredients_text`) when checking `$product_ref->{owner_fields}`.
- Enforced `skip_protected_field` in `update_product_fields` (API v3) for non-moderators.
- Added structured API warning responses (`message => { id => "field_protected_by_producer" }`, `field => { id => $field }`, `impact => { id => "field_ignored" }`) to both API v3 and API v2/CGI write pathways.
- Added unit tests in `tests/unit/products.t` covering base and localized `is_owner_field` checks as well as moderator bypass rules.

- **LLM Used**: Claude 3.7 Sonnet (Agentic Pair Programmer)
- **Usage**: Used for root cause analysis of API field protection flow, Perl code updates, and unit test implementation.
- **Verification**: Verified code syntax and unit test coverage locally.

### Screenshot or video
N/A (Backend API response and product protection rule logic).

- Fixes #14492

### DIFF
--- a/cgi/product_jqm_multilingual.pl
+++ b/cgi/product_jqm_multilingual.pl
@@ -58,7 +58,7 @@ use ProductOpener::EnvironmentalScore qw/:all/;
 use ProductOpener::Packaging qw/:all/;
 use ProductOpener::ForestFootprint qw/:all/;
 use ProductOpener::Text qw/remove_tags_and_quote/;
-use ProductOpener::API qw/get_initialized_response check_user_permission/;
+use ProductOpener::API qw/add_warning get_initialized_response check_user_permission/;
 use ProductOpener::APIProductWrite
 	qw/process_change_product_type_request_if_we_have_one process_change_product_code_request_if_we_have_one skip_protected_field update_product_field_api_v2_and_cgi/;
 
@@ -406,7 +406,7 @@ else {
 			$add_tags = 1;
 		}
 
-		update_product_field_api_v2_and_cgi($product_ref, $lc, $field, single_param($field), $source, $add_tags);
+		update_product_field_api_v2_and_cgi($product_ref, $lc, $field, single_param($field), $source, $add_tags, $response_ref);
 
 		if (defined $language_fields{$field}) {
 
@@ -416,6 +416,14 @@ else {
 
 					# Only moderators can update values for fields sent by the producer
 					if (skip_protected_field($product_ref, $field_lc, $User{moderator})) {
+						add_warning(
+							$response_ref,
+							{
+								message => {id => "field_protected_by_producer"},
+								field => {id => $field_lc},
+								impact => {id => "field_ignored"},
+							}
+						);
 						next;
 					}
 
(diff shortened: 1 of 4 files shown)


SurbhiAgarwal1:
Thanks for the review! I have updated the PR to address the feedback:

1. **Rebased onto latest `main`**: Cleaned up the branch to remove unrelated taxonomy tag commits (`Tags.pm` & `tags.t`). The PR now contains a single, focused commit.
2. **CGI Localized Fields Warning Feedback**: Updated `cgi/product_jqm_multilingual.pl` so the `$param_lang` loop also populates warning feedback (`add_warning($response_ref, ...)`) when protected localized fields (e.g., `ingredients_text_fr`) are skipped.
3. **POD Documentation**: Updated the POD documentation in `lib/ProductOpener/APIProductWrite.pm` for `update_product_field_api_v2_and_cgi`.

PR #14504, opened 2026-09-09
fix: wrong Google Play badges

teolemon:
fix: wrong Google Play badges


alexgarel (approved):
The validate translations file will be very useful.

That said I hope validation won't make the server too much traffic…

It would be better to only run validation on line changed (maybe by loading the po from the commit before).

teolemon:
I disabled it on main

PR #14509, opened 2026-09-09
taxonomy: translation boost for the states taxonomy

teolemon:
taxonomy: translation boost for the states taxonomy
Disclaimer: Antigravity with Gemini Pro 3.1


TaciteOFF (approved):
Good for me.
NB: Correctly changes Turkish “Photos uploaded” from “will be uploaded” to “uploaded.”

PR #14511, opened 2026-09-09
fix: cleanup translations

teolemon:
fix: cleanup translations

- Some brands (OFF, Nutri-Score) were translated
- We had an old url which was refactored away, but still in the translations


Freso (approved):
seems good to me (with the caveat that i only looked at a few of the 2000+ changes)

teolemon:
thank you @Freso

PR #14518, opened 2026-09-09
fix: some additional stuff spotted on PO spawning

teolemon:
fix: some additional stuff spotted on PO spawning


Freso (approved):
seems ok

PR #14521, opened 2026-09-09
fix: move the donation to webcomponent

teolemon:
fix: move the donation to webcomponent
blocked on https://github.com/openfoodfacts/openfoodfacts-webcomponents/pull/618/changes being merged, and then the webcomponents being released, and upgraded here.

image

## Summary by CodeRabbit

* **New Features**
  * Added reusable donation banners for page headers and footers, with content loaded dynamically.
  * Added a tagline feed and asset proxy with caching and CORS support.
  * Donation banners retain cookie-based dismissal and close controls.
* **Style**
  * Updated visible banner layout and positioning for consistent rendering across pages.
* **Tests**
  * Updated integration expectations to reflect the new donation banner presentation.

### DIFF
--- a/templates/web/common/includes/donate_banner.tt.html
+++ b/templates/web/common/includes/donate_banner.tt.html
@@ -357,66 +357,9 @@
 <!-- Open Food Facts Days end -->
 [% ELSE %]
 <!-- Donation banner TOP start -->
-<section id="donation-banner-top" class="donation-banner row">
-
-  <div class="donation-banner__left-aside">
-    <div class="donation-banner__hook-section">
-      <p>[% lang("donation_hook_2026") %]</p>
-    </div>
-    <img src="/images/misc/donation-banners/donation-banner-group-photo.png" alt="group photo donation 2024" />
-  </div>
-  <div>
-    <div class="donation-banner__aside">
-      <div class="donation-banner__main-section">
-        <img
-          width="50"
-          height="50"
-          src="https://world.openfoodfacts.org/images/logos/logo-variants/CMJN-ICON_WHITE_BG_OFF.svg"
-          alt="Open Food Facts logo"
-        />
-        <h3 class="donation-banner__main-title">[% lang("donation_title_2026") %]</h3>
-      </div>
-      <div style="padding:1rem;">
-        <p>[% lang("donation_list_2026_main") %]</p>
-        <ul>
-          <li>
-            [% lang("donation_list_2026_first") %]
-            <ul>
-              <li>[% lang("donation_list_2026_first_sub") %]</li>
-            </ul>
-          </li>
-          <li>
-            <p>[% lang("donation_list_2026_second") %]</p>
-          </li>
-          <li>
-            <p>[% lang("donation_list_2026_third") %]</p>
-          </li>
-          <li>
-            <p>[% lang("donation_list_2026_fourth") %]</p>
-          </li>
-        </ul>
-      </div>
-    </div>
-    <div class="donation-banner__actions-section">
-      <div class="donation-banner__actions-section__financial">
-        <p>
-          [% IF cc == "fr" %]
-          (déductible à 66 % de vos impôts: ex. un don de 20€ vous reviendra à 6,80€)
-          [% ELSE %]
-          [% lang("donation_financial_2026_text") %]
-          [% END %]
-        </p>
-      </div>
-      <div class="donation-banner__actions-section__donate-button">
-        <a href="[% link %]?utm_source=off&utm_medium=web&utm_campaign=donate-2026-a&utm_term=en-text-button">
-          <button>[% lang("donation_button_2026_text") %]</button>
-        </a>
-      </div>
-    </div>
-  </div>
-[% END %]
-
-  <div class="donation-banner__close">
+<section id="donation-banner-top" style="position: relative;">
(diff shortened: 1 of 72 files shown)


templates/web/common/includes/donate_banner.tt.html
```diff
@@ -455,66 +398,18 @@ <h3 class="donation-banner__main-title">[% lang("donation_title_2026") %]</h3>
...
-      </div>
-      <div class="donation-banner-footer__actions-section__donate-button">
-        <a href="[% link %]?utm_source=off&utm_medium=web&utm_campaign=donate-2026-a&utm_term=en-text-button">
-          <button>[% lang("donation_button_2026_text") %]</button>
-        </a>
-      </div>
-    </div>
-  </div>
-</section>
+<donation-banner donate-url="[% link %]" news-url="https://raw.githubusercontent.com/openfoodfacts/smooth-app_assets/refs/heads/main/prod/tagline/web/main.json"></donation-banner>
```
hangy: Is it okay, policy-wise, to include data from GitHub directly?

It might be against GitHub policy, and embedding data directly from githubusercontent.com could allow tracking and violate GDPR

hangy (commented):
We should also look into deleting old translations and the images then

templates/web/common/includes/donate_banner.tt.html
teolemon: no, we should do something cleaner

UrbanFercec:
@hangy is right, and it is fixable without moving the feed. The fetch happens in the visitor's browser, so GitHub gets an IP, a User-Agent and a Referer per page view, and their ToS says raw is not a CDN. Both go away if the same file is served from an OFF origin:

```nginx
location = /data/tagline/web/main.json {
  proxy_pass https://raw.githubusercontent.com/openfoodfacts/smooth-app_assets/refs/heads/main/prod/tagline/web/main.json;
  proxy_cache tagline;
  proxy_cache_valid 200 10m;
  proxy_cache_use_stale error timeout updating;
}
```

and `news-url` points at `/data/tagline/web/main.json`. Editing stays in `smooth-app_assets`, every visitor collapses into one upstream request per cache window so the rate limits stop mattering, and the banner survives GitHub being down.

The web component needs no change for this: its feed URL is an attribute with no default, so it is one line here in the template. If we would rather the browser fetch nothing at all for the figures, the server could read the JSON itself and pass `raised` / `goal` / `currency` in as attributes, which the component already accepts; the news items would still want a URL.

Happy to open the template change once someone confirms the nginx side and where the path should live.

UrbanFercec:
Correction to my own proposal: proxying `main.json` alone is not enough. The feed's content carries absolute GitHub URLs of its own, for example the donation campaign's image:

```json
"image": { "url": "https://raw.githubusercontent.com/openfoodfacts/smooth-app_assets/.../prod/tagline/web/assets/donation_campaign/donate-heart.png" }
```

That is rendered as an ``, so the visitor's browser still ends up requesting it from GitHub, with the same IP, User-Agent and Referer. Same issue, one hop later.

So the proxy wants to be a prefix rather than a single file:

```nginx
location /data/tagline/ {
  proxy_pass https://raw.githubusercontent.com/openfoodfacts/smooth-app_assets/refs/heads/main/prod/tagline/;
  proxy_cache tagline;
  proxy_cache_valid 200 10m;
  proxy_cache_use_stale error timeout updating;
}
```

and the image URLs in the feed then point at `https://world.openfoodfacts.org/data/tagline/web/assets/...`. Editing the file stays where it is, and since the app reads the same feed, it stops calling GitHub from users' devices too.

teolemon:
@hangy @alexgarel wdyt?

teolemon:
@alexgarel @hangy  I have added @UrbanFercec 's suggestion

teolemon:
@coderabbitai full review

UrbanFercec:
+1 on `proxy_ssl_verify on` - the snippet I posted skipped it. For the stock nginx image the bundle is Debian's:

```nginx
proxy_ssl_verify on;
proxy_ssl_verify_depth 2;
proxy_ssl_trusted_certificate /etc/ssl/certs/ca-certificates.crt;
```

Without it a DNS or MITM answer for `raw.githubusercontent.com` would be cached and served under the OFF origin for up to 10 minutes.

teolemon:
@copilot resolve the merge conflicts in this pull request  and update openfoodfacts-webcomponents and apply 
+1 on proxy_ssl_verify on - the snippet I posted skipped it. For the stock nginx image the bundle is Debian's:

proxy_ssl_verify on;
proxy_ssl_verify_depth 2;
proxy_ssl_trusted_certificate /etc/ssl/certs/ca-certificates.crt;
Without it a DNS or MITM answer for raw.githubusercontent.com would be cached and served under the OFF origin for up to 10 minutes.

teolemon:
@alexgarel @stephanegigandet @UrbanFercec the tests are now passing

stephanegigandet:
Tested locally, it works but most translations have &amp; instead of & :
(and "Veuillez faire un don" sounds bad)

image

stephanegigandet (approved):
Approved, but before merging it would be good to fix the translation issue with &amp; and verify that the title sounds good in key languages.

teolemon:
@stephanegigandet I released the webcomponent with fixes and I bumped.

PR #14491, opened 2026-09-07
fix(ingredients): parse mg/kg and activity units without splitting on slash 6132

stephanegigandet:
Same PR as https://github.com/openfoodfacts/openfoodfacts-server/pull/14163 from @nightwinkle , just to try updating the tests with /update_tests_results

PR #14522, opened 2026-09-09
fix: preserve additive-class parsing with compound units

NightWinkle:
### What

Fixes the six failing ingredient and taxonomy-enhancer assertions in #14491. Ordinary unknown or misspelled additive children are flattened again, keeping ingredient lists aligned across languages. The child-recognition check is now limited to protected compound-unit cases such as `mg/kg`, where it also handles processing and localized conjunctions.

The class is still kept for unrecognized unit text such as `mg/kg 1b306(i)`. Quantity removal requires an explicit unit, preserving additive names such as `Natural Red 4`; one regression case covers this.

The four taxonomy-enhancer failures also reproduced after merging `main` at `baaa4e0e1` into the original correction branch. They require the parser fix here.

One export snapshot is restored to its value on `main`: #14491's test-result update had recorded the regression for `colorants : lutéine et anthocyanes`. The restored ingredient tags match the actual CI export exactly; no other snapshots change.

### Testing

- On this branch at `bd422c02e3` and on its local merge with `main`: `ingredients_processing.t` 86/86, `ingredients_tags.t` 90/90, `ingredients.t` 132/132, `taxonomies_enhancer.t` 48/48.
- Full unit suite on that local merge: 90 files, 7,446 assertions passed. The sparse checkout was expanded to include the tracked logo and label assets required by the tests.
- The new `Natural Red 4` assertion failed before the quantity fix and passed afterward.
- Perl compilation, perltidy, perlcritic, taxonomy checks, and `git diff --check` passed.
- [Final CI at `48af4e0cc7`](https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/34418608897) passed all six unit groups, all nine integration groups (including `export.t`), Perl checks, frontend checks, and `make dev`. The restored snapshot is valid JSON, identical to `main`, and matches the generated export.

OpenAI Codex (GPT-6) was used agentically for investigation, implementation, testing, and review. Local checks ran in Docker; remote checks ran in GitHub Actions.

- Follow-up to #14491; targets `fix/ingredients-mg-per-kg-6132b`.
- Related to #6132.

### DIFF
--- a/lib/ProductOpener/Ingredients.pm
+++ b/lib/ProductOpener/Ingredients.pm
@@ -3045,14 +3045,11 @@ Text to analyze
 					# ingredients tags that are too long (greater than 1024, mongodb max index key size)
 					# will cause issues for the mongodb ingredients_tags index, just drop them
 					if (length($ingredient{id}) < 500) {
-						# Only flatten an additive class when at least one of its children is
-						# recognizable (#6132 follow-up): protecting the solidus of compound
-						# units removed the accidental '/' splits that used to create list
-						# boundaries, so e.g. "Antioxygènes : Avec antioxydant naturel :
-						# mg/kg 1b306(i)" would otherwise be flattened into unknown junk
-						# children and the class tag would be lost entirely.
-						my $flatten_children = 0;
-						if ($is_flattenable_additive_class && $between ne "") {
+						# Only require recognizable children when protected compound-unit slashes
+						# could flatten the class into unknown text, e.g. "mg/kg 1b306(i)".
+						my $flatten_children = $is_flattenable_additive_class && $between ne '';
+						if ($flatten_children && $between =~ /\N{U+2044}/) {
+							$flatten_children = 0;
 							foreach my $raw_chunk (split(/$separators/, $between)) {
 								# $separators contains capturing groups, so split
 								# also yields undef delimiter slots.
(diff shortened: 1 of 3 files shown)


NightWinkle:
I had an LLM take a look at the failing tests on #14491 and it came back with a fix (this PR #14522, stacked on this branch). I reviewed the changes, here is a quick summary :

My mg/kg change made additive-class flattening conditional on having at least one recognized child. That protected the `mg/kg 1b306(i)` junk case, but it applied to every additive class, so stuff like `colour (Natural Red 4)` stopped flattening. That's what broke the six assertions and the export snapshot.

The fix goes back to flatten-by-default and only keeps the "recognized child required" rule when the text actually contains a compound unit (fraction slash). The child check also got a bit smarter: it splits on the localized "and", only strips a trailing quantity when it has a unit (so `Natural Red 4` doesn't lose its name), and accepts processing phrases too. On top of that, the export snapshot is restored to main's version and there's a new regression test for the Natural Red 4 case. The new unit check reads the quantity regexp's $2 group, same as the rest of the file already does, so it's consistent.

All the failing suites pass now and CI is green.
It looks okay to me, what do you think @stephanegigandet ? 

Merge order should be this one then #14491 to keep main state green.

stephanegigandet (approved):
Looks good, thank you

PR #14525, opened 2026-09-10
chore: Script to allow replay of redis events that have not been processed

john-gom:

### What
Adds a `redis_retry_stream.pl` script to allow redis events to be replayed in the event that some were not processed

None

### DIFF
--- a/lib/ProductOpener/Redis.pm
+++ b/lib/ProductOpener/Redis.pm
@@ -126,10 +126,26 @@ sub init_redis() {
 	return;
 }
 
-=head2 subscribe_to_redis_streams ()
+=head2 subscribe_to_redis_streams ($search_from = undef, $search_to = undef, @streams)
 
 Connects to Redis and processes any events received. Blocks until there is an error or the application terminates.
 Returns on error or when receiving a terminate signal from the OS
+
+=head3 Parameters
+
+=head4 $search_from (input)
+
+The starting point for reading from the Redis stream. Starts from the last processed message if not provided.
+
+=head4 $search_to (input)
+
+The ending point for reading from the Redis stream. Stays active indefinitely if not provided.
+
+=head4 @streams (input)
+
+The list of Redis streams to subscribe to. Defaults to all streams if not provided.
+
+=cut
 
 =cut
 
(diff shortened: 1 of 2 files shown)


lib/ProductOpener/Redis.pm
```diff
@@ -198,15 +232,14 @@ sub _read_user_streams() {
 
 		# Listen for user-deleted events so that we can redact product contributions for this flavor
 		# This will block for up to 5 seconds waiting for messages and return a maximum of 1000
-		my @streams = (
-			'COUNT', 1000, 'BLOCK', 5000, 'STREAMS', 'user-deleted',
-			'user-registered', 'user-updated', $search_from, $search_from, $search_from
-		);
+		my @params = ('COUNT', 1000, 'BLOCK', 5000, 'STREAMS');
+		push @params, @streams;
+		push @params, map {$search_from} @streams;
```
alexgarel, line 237: I think this is worth a comment otherwise it's a bit cryptic.

```suggestion
		# push as many search_from as we have streams
		push @params, map {$search_from} @streams;
```

lib/ProductOpener/Redis.pm
```diff
@@ -217,13 +250,15 @@ sub _read_user_streams() {
...
+					# 	if $log->is_info();
 					# Process any received messages
-					my $last_processed_message_id = process_xread_stream_reply($reply_ref);
+					my $last_processed_message_id = process_xread_stream_reply($reply_ref, $search_to);
 					if ($last_processed_message_id) {
 						$search_from = $last_processed_message_id;
-						store_object("$BASE_DIRS{PRIVATE_DATA}/last-processed-id", $search_from);
+						if (not defined $search_to) {
+							store_object("$BASE_DIRS{PRIVATE_DATA}/last-processed-id", $search_from);
+						}
```
alexgarel, line 261: Is it a good idea ? You should at least comment to explain why.

I may run a job to replay last messages up to now, while product opener runs on it's own and this may interfere (or I am wrong ?)

I would better define a `$store_last_processed_id = 1 if ($search_from or $search_to)` at the start and use it there.

or simply a "$persist_id" parameter !

alexgarel: Reading the script I see that you made the search_to mandatory, but I think it's not fool proof this way (counting on a behaviour defined elsewhere as a constraint to this sub)

alexgarel (approved):
Thanks @john-gom 
I'm approving but I have some comments that I would prefer you to see.

lib/ProductOpener/Redis.pm
john-gom: I have made it a bit more explicit in the code. I didn't want to add another flag as I think it would be redundant

PR #14531, opened 2026-09-11
taxonomy: translate part of ingredients_processing into cs

paper42:
### What

processing items translated to czech

none


Freso (approved):
thanks!

PR #14533, opened 2026-09-11
taxonomy: translate various ingredients into cs

paper42:
### What

Various unidentifiable ingredients I found while scanning and adding new items.

none

- Fixes:


Freso (approved):
thanks!

PR #14536, opened 2026-09-12
refactor(perf): sort remaining taxonomy entries alphabetically in suggestions once

hangy:
### What

Moves sorting of tags before the loop to avoid redundant execution. This should speed up `/api/v3/taxonomy_suggestions` for larger taxonomies.

None

### Screenshot or video

Tested `tests/unit/taxonomy_suggestions.t` with `categories` instead of the `tests` taxonomy to actually get a few iterations in. 

#### `Devel::NYTProf` before

> Profile of tests/unit/taxonomy_suggestions.t for 5.01s (of 7.09s), executing 6906569 statements and 1617356 subroutine calls in 725 source files and 340 string evals.

NYTProf flame graph before optimization

#### `Devel::NYTProf` after

> Profile of tests/unit/taxonomy_suggestions.t for 3.51s (of 3.96s), executing 1155926 statements and 456627 subroutine calls in 725 source files and 341 string evals.

NYTProf flame graph after optimization

None.

### DIFF
--- a/lib/ProductOpener/TaxonomySuggestions.pm
+++ b/lib/ProductOpener/TaxonomySuggestions.pm
@@ -187,12 +187,12 @@ sub generate_sorted_list_of_taxonomy_entries ($country, $tagtype, $search_lc, $c
 		@tags = generate_popular_suggestions_according_to_context($country, $tagtype, $search_lc, $context_ref,
 			\%seen_tags);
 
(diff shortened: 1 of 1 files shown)


TaciteOFF (approved):
Awesome! I tried locally after setting up Modern::Perl and observed similar results.

PR #14537, opened 2026-09-12
fix(taxonomy): resolve duplicate synonyms in three-letter languages

TaciteOFF:
### What

Eight names in `countries.txt` and `languages.txt` are each mapped to two different entries. The taxonomy builder reports them as `duplicate_synonym` errors.

**countries**

- `got: 𐌰𐌽𐍄𐌰𐍂𐌺𐍄𐌹𐌺𐌰` is on both `en:Antarctic` and `en:Antarctica`. The Gothic name transliterates *Antarktika*, so it stays on Antarctica. German already distinguishes the two entries the same way, with `de: Antarktis` and `de: Antarktika`.
- `bxr: Гренада` is on both `en:Grenada` and `en:Honduras`. It means Grenada, so it is removed from Honduras.
- `crh: Kongo` is on both `en:Democratic Republic of the Congo` and `en:Republic of the Congo`. Turkish, the closest relative in this taxonomy, uses `Demokratik Kongo Cumhuriyeti` and `Kongo Cumhuriyeti`, so plain *Kongo* is the Republic of the Congo and is removed from the DRC.
- `jbo: as` is an artifact of splitting the Lojban names `bylgAri,as` (Bulgaria) and `slovEni,as` (Slovenia) at their internal comma. Restore the complete names as `bylgArias` and `slovEnias`, following the review suggestions. In Lojban, this comma is an optional syllable separator; omitting it preserves the name and avoids the taxonomy synonym delimiter. See [Bulgaria](https://www.wikidata.org/wiki/Q219), [Slovenia](https://www.wikidata.org/wiki/Q215), and the [Lojban reference grammar](https://www.lojban.org/publications/cll.before-20160606/cll_v1.1_xhtml-section-chunks/section-lojban-characters.html).

**languages**

- `pms: Lenga inuktitut` is on both `en:Greenlandic` and `en:Inuktitut`. It means Inuktitut, so it is removed from Greenlandic.
- `pms: Lenga norvegèisa` is on both `en:Bokmål` and `en:Nynorsk`. It names the Norwegian macrolanguage, so it moves to `en:Norwegian`, where the same Romance pattern already appears as `lij: Lengua norvegeise` and `lmo: Lengua nurvegesa`.
- `mdf: Норвегонь кяль` is on both `en:Norwegian` and `en:Nynorsk`, and likewise stays on Norwegian.
- `gag: Uygur dili` is on both `en:Hungarian` and `en:Uyghur`. It means Uyghur, so it is removed from Hungarian.

Most of these look like Wikidata imports that landed on a neighbouring entry.

### Impact

None today. The builder only reads two-letter language prefixes, so every line above is currently ignored and no entry changes. The errors appear as soon as three-letter prefixes are read, which makes the affected taxonomies fail to build.

### Tests

All 35 taxonomies build with no error, against a builder that reads three-letter prefixes. The eight fixes are sufficient: no other taxonomy file has a hidden duplicate of this kind.

### DIFF
--- a/taxonomies/countries.txt
+++ b/taxonomies/countries.txt
@@ -6783,7 +6783,7 @@ io: Bulgaria
 is: Búlgaría, Bulgaria
 it: Bulgaria
 ja: ブルガリア, ブルガリア国民議会, 勃牙利, ブルガリアの下院, ブルガリヤ, ブルガリア共和国, 勃
-jbo: bylgAri, bylgarias, bulgar, balgarias, Bulgaria, bulgarias
+jbo: bylgArias, bulgar, balgarias, Bulgaria, bulgarias
 jv: Bulgaria
 ka: ბულგარეთი
 kaa: Bolgariya
@@ -38200,7 +38200,7 @@ io: Slovenia
 is: Slóvenía
 it: Slovenia
 ja: スロベニア, スロヴェニア
-jbo: slovEni
+jbo: slovEnias
 jv: Slovénia
 ka: სლოვენია
 kaa: Sloveniya
--- a/taxonomies/countries.txt
+++ b/taxonomies/countries.txt
@@ -1729,7 +1729,6 @@ fa: سرزمین جنوبی
 fiu-vro: Antarktiga
 fr: Région antarctique
 gd: An Antartaig
-got: 𐌰𐌽𐍄𐌰𐍂𐌺𐍄𐌹𐌺𐌰
 gu: એન્ટાર્કટીકા
 hsb: Antarktis
 hu: Antarktisz
@@ -6784,7 +6783,7 @@ io: Bulgaria
 is: Búlgaría, Bulgaria
 it: Bulgaria
 ja: ブルガリア, ブルガリア国民議会, 勃牙利, ブルガリアの下院, ブルガリヤ, ブルガリア共和国, 勃
-jbo: bylgAri, as, bylgarias, bulgar, balgarias, Bulgaria, bulgarias
+jbo: bylgAri, bylgarias, bulgar, balgarias, Bulgaria, bulgarias
 jv: Bulgaria
 ka: ბულგარეთი
 kaa: Bolgariya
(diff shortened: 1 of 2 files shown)


taxonomies/countries.txt
```diff
@@ -6784,7 +6783,7 @@ io: Bulgaria
 is: Búlgaría, Bulgaria
 it: Bulgaria
 ja: ブルガリア, ブルガリア国民議会, 勃牙利, ブルガリアの下院, ブルガリヤ, ブルガリア共和国, 勃
-jbo: bylgAri, as, bylgarias, bulgar, balgarias, Bulgaria, bulgarias
+jbo: bylgAri, bylgarias, bulgar, balgarias, Bulgaria, bulgarias
```
svensven:
```suggestion
jbo: bylgArias, bulgar, balgarias, Bulgaria, bulgarias
```
the jbo (lojban) `,as` appears to be part of the preceding words that got reformatted to have a space in front. e.g. 
https://jbo.wikipedia.org/wiki/bylgAri,as 
https://jbo.wikipedia.org/wiki/slovEni,as
so assuming the code can't handle entries containing a comma, perhaps it's best just to remove them? (or maybe we need some special handling for lojban taxonomy entries, not that anyone likely cares)

taxonomies/countries.txt
```diff
@@ -38203,7 +38200,7 @@ io: Slovenia
 is: Slóvenía
 it: Slovenia
 ja: スロベニア, スロヴェニア
-jbo: slovEni, as
+jbo: slovEni
```
svensven:
```suggestion
jbo: slovEnias
```

taxonomies/countries.txt
svensven: per chatgpt:
> It depends on the exact word, but in your example slovEni,as → slovEnias, removing the comma changes the pronunciation, not necessarily the intended word.
>  
> The comma (,) in Lojban is called an apostrophe-like phonetic separator and indicates that the vowels on either side should be pronounced as separate syllabic units.
> 
> So roughly:
> 
>     slovEni,as → slo-VE-ni-AS (a clear break between i and a)
>     slovEnias → the ia is treated as a diphthong, something like slo-VE-nyas
> 
> This can matter because Lojban has rules about how vowel sequences are interpreted. So the comma can prevent an i + vowel sequence from being interpreted as a single diphthong.
> 
> In short: removing the comma can make two adjacent vowels merge into a diphthong, potentially changing how the word is parsed/pronounced.

TaciteOFF: You're right, thank you!

Freso (approved):
might be good to keep `bylgAri\,as` around, but this is still an improvement, and the comma version can be added back later if wanted. thanks!

PR #14539, opened 2026-09-13
taxonomy: riz blanc long grain basmati

aleene:
### What
Added riz blanc long grain basmati due to Nutri-Patrol


moon-rabbitOFF (commented):
basmati rice is always long grain, I think this can be put as a synonym of "white basmati rice".

Freso (approved):

stephanegigandet (approved):

PR #14543, opened 2026-09-14
fix: Revise footer links and add new sections

teolemon:
- Updated social media links and added new footer items.
- footer-open-prices opff opf data


teolemon:
@copilot can you add the missing strings in common.pot and en.po ?

alexgarel (approved):

PR #14544, opened 2026-09-14
fix: Add missing Ecobalyse token exports

TaciteOFF:
`scripts/build_lang.pl` crashes for Open Beauty Facts, Open Products Facts and Open Pet Food Facts. Their configs do not export `$ecobalyse_api_token`, which `EnvironmentalImpact.pm` needs to load.

This adds the missing export and reads the value from `Config2`, just like Open Food Facts. The code can load even when no token is set.

The new test loads the module for all four product types, with and without a token. It reproduces the bug on `main` and passes with this fix. The new test and the existing environmental impact test pass all 10 checks.

The full build script now finishes for all four product types without a token and writes the translation files. Perltidy, Perlcritic and `git diff --check` also pass.


stephanegigandet (approved):
Thanks

alexgarel:
@TaciteOFF thanks for this PR !

PR #14545, opened 2026-09-14
fix(Makefile): quote `$DEPS_DIR`

Freso:
Commands like `git clone … $DEPS_DIR/…` would break if the off-server checkout had any parent directories with spaces in them, since they would get split up into multiple arguments from the shell’s PoV.

None used.


TaciteOFF (approved):
Thanks!

hangy:
We may need to update this in a few more places: https://github.com/search?q=org%3Aopenfoodfacts+DEPS_DIR&type=code

Freso:
I don’t understand why the whole path gets duplicated at https://github.com/openfoodfacts/openfoodfacts-server/actions/runs/34831472719/job/103936785712?pr=14545#step:13:74 :face_with_diagonal_mouth: 
```
open /home/runner/work/openfoodfacts-server/openfoodfacts-server/"/home/runner/work/openfoodfacts-server/openfoodfacts-server/deps"/openfoodfacts-shared-services/docker-compose.yml: no such file or directory
```

Freso:
@hangy The idea of quoting it where it’s set is that that should alleviate the need for quoting it when it’s used, but maybe it isn’t actually working like that, and it instead needs to be quoted every time it’s referenced. :face_with_diagonal_mouth:

john-gom:
Investigating...

john-gom:
I think the problem is this like as the variable is already quoted so the extra quotes confuse things:
`DOCKER_COMPOSE_TEST=COMPOSE_FILE="${COMPOSE_FILE_BUILD};${DEPS_DIR}/openfoodfacts-shared-services/docker-compose.yml"`

Freso:
i was hoping to avoid having to quote every instance of referencing it, but ah well. thanks @john-gom!

PR #14554, opened 2026-09-14
taxonomy(food): Scandinavian free-range eggs

Freso:
Needed-for: https://world.openfoodfacts.org/product/8710522882276/same-real-mayonnaise-hellmann-s


teolemon (approved):
thanks @Freso

PR #14557, opened 2026-09-14
taxonomy(food): gifflar edits

Freso:
Includes some additional translations from Wikidata and some normalisation of ingredient names.

Needed-for: https://world.openfoodfacts.org/product/7311070006209/hvedeboller-med-kanel-pagen

PR #14558, opened 2026-09-15
taxonomy(food): fix it:armellina ingredient

Freso:
As far as I can tell, it:armellina is apricot kernels, not bitter almonds.

Follow-up-to: https://github.com/openfoodfacts/openfoodfacts-server/pull/14557
Follow-up-to: bd30f338035c66c221a2598aa1337dcdd0acbea2


teolemon (approved):
thanks for the followup

PR #14563, opened 2026-09-15
feat: force products images to be taken from images sub-domain

alexgarel:
So that they are downloaded from the right server.


stephanegigandet (approved):
Thanks

hangy (approved):

PR #14564, opened 2026-09-16
refactor: Split backend image into `dev` and `prod` variant

hangy:
### What

Reduces the number of libraries/packages distributed with the production image greatly. This means fewer CVEs, and a smaller attack surface in general.

None.

### Screenshot or video

None.

None.


Freso:
Would it make sense to have the dev image build on top of the prod one?

hangy:
> Would it make sense to have the dev image build on top of the prod one?

If we were to keep the packages installed in both layers, of course. My idea for this stack is to eliminate the unnecessary packages, though: https://github.com/openfoodfacts/openfoodfacts-server/pull/14567/changes#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557 With that approach, they share as much as they can from the `runnable` stage/layer, and the `prod` ends up just being some kind of label

PR #14571, opened 2026-09-16
refactor: replace CLDR::Number with Locale::Unicode::Data in Text.pm

hangy:

- Fixes: 

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


stephanegigandet:
@hangy there seems to be a lot of added code for fallbacks etc. that I guess covers functionality that was provided by CLDR::Number ? Why change?

PR #14575, opened 2026-09-16
taxonomy(food): primary zh for meat alternatives

Freso:
@Meowdolee wrote on the previous PR:
> I believe there may have been a mix-up between Traditional Chinese
> and Simplified Chinese during my input.
> For this one, let’s just use 肉类替代品.

So adding “肉类替代品” as the primary `zh` synonym for the meat alternatives category.

Follow-up-to: https://github.com/openfoodfacts/openfoodfacts-server/pull/14470
Follow-up-to: 6cc2dbd23b34c878910c6fac68bfe27d755820f5
Reference: https://github.com/openfoodfacts/openfoodfacts-server/pull/14470/changes#r4017596701


teolemon (approved):
thanks @Freso 👍

PR #14583, opened 2026-09-17
chore: Archive non-eu packager codes script

hangy:
### What

Updates scrapy, updates the scrapy calling logic, then _archives_ the whole script, because the source is not available any more. 😊

None.

### Screenshot or video

None.

- https://github.com/openfoodfacts/openfoodfacts-server/security/dependabot/89
- https://github.com/openfoodfacts/openfoodfacts-server/security/dependabot/112
- https://github.com/openfoodfacts/openfoodfacts-server/security/dependabot/132
- https://github.com/openfoodfacts/openfoodfacts-server/security/dependabot/175


Freso (commented):
why are we moving them to obsolete rather than just removing them? they’ll still be in git’s history if we need to pull them out again in the future

other than that, seems good to me!

hangy:
> why are we moving them to obsolete rather than just removing them? they’ll still be in git’s history if we need to pull them out again in the future

Personally, I'd delete them. Moving old scripts to the `obsolete` path just seems to be common practice in this repo, which is why I followed it.

+1 for Git history 😊

PR #14587, opened 2026-09-17
taxonomy: gellan gum

aleene:
### What
Added description and nova for gellan gum


Freso (approved):
are you writing the descriptions yourself? if not, it would be good if you would note where you get them from.

PR #14598, opened 2026-09-17
fix: Ensure product_updates are pushed to Redis when a user is deleted

### LINKED ISSUE #14596: Deleting a user does not generate a Redis event
### What
When a user is deleted their id is replaced with an anonymous placeholder in all products that they edited. However, this does not trigger a product_updates event so the data in systems like off-query is not updated.

1. Delete a user with contributions
2. Check off-query to see if that user can still be found with a script like this:

```
curl -X 'POST' \
  'https://query.openfoodfacts.org/find?obsolete=false&product_type=food' \
  -H 'accept: */*' \
  -H 'Content-Type: application/json' \
  -d '{
  "filter": {
    "editors_tags": "<deleted isuer id>"
  },
  "projection": {
    "code": true
   }
}'
```

john-gom:

### What
Generate a product_updatres event when user information is redacted from products on user deletion

None

- Fixes: #14596


alexgarel (approved):
Good catch !

PR #14472, opened 2026-09-04
taxonomy(food): add Caesar dressing category

Freso:
Also merges Caesar sauces into Caesar dressings.

Needed-for: https://se.openfoodfacts.org/product/7350126089032/caesardressing-felix
Needed-for: https://se.openfoodfacts.org/product/7313310015620/caesardressing-fresh-food

PR #14615, opened 2026-09-19
fix(taxonomy): restore environmental-score match for Caesar salad dressings

### LINKED ISSUE #14547: not sure why the category name change removes the environmental score… ? is it because it now has an `agribalyse` food?
not sure why the category name change removes the environmental score… ? is it because it now has an `agribalyse` food?

_Originally posted by @Freso in https://github.com/openfoodfacts/openfoodfacts-server/pull/14472#discussion_r3940649211_

Eduardo-Vilas-Boas:
Fixes #14547

## Status before the #14472 merge

Product `8722700472575` (Amora Caesar dressing) was categorized under `en:caesar-sauces`, which had no `agribalyse_food_code` of its own. The environmental-score matching logic (`compute_environmental_score_agribalyse()`) walked up to the parent category `en:salad-dressings`, which has `agribalyse_food_code:en: 11187` — a code with real data in the shipped Agribalyse LCA dataset (`AGRIBALYSE_vf.csv.3`).

Result: `environmental_score_grade: "d"`, `environmental_score_score: "38"`, `agribalyse.code: "11187"`.

## Status after

#14472 merged `en:caesar-sauces` into a new, more specific `en:caesar-salad-dressings`, and gave it its own `agribalyse_food_code:en: 11248` (Ciqual code for "Salad dressing, Caesar style, prepacked" — a food newly added in Ciqual's 2025 edition).

Problem: `11248` has **no entry** in the Agribalyse LCA dataset we ship — Agribalyse only covers ~2,517 products and hasn't studied this specific one. The matching loop commits to the first category with *any* `agribalyse_food_code` and stops (`last if $agb_match`), without checking whether that code actually resolves to real data. So it picks `11248`, finds nothing, and never falls back to the parent's valid `11187`.

Result: `environmental_score_grade: "unknown"`, `environmental_score_score: ""`, `agribalyse.code: ""`.

## Proposed solution

Replace the false `agribalyse_food_code:en: 11248` with `agribalyse_proxy_food_code:en: 11187` on `en:caesar-salad-dressings` — the field that exists specifically for "no dedicated LCA study exists, use the closest category's data as a proxy." `ciqual_food_code:en: 11248` is left untouched since that nutrition data is accurate and unrelated to the environmental-score bug.

Verified with the actual `export.t` integration test: `environmental_score_grade`/`score`/`agribalyse.code` for product `8722700472575` are now byte-identical to their pre-#14472 values (`d` / `38` / `11187`). `make check_taxonomies` also passes on the modified file.

**Out of scope, noted for a follow-up:** `compute_environmental_score_agribalyse()` itself never validates that a matched `agribalyse_food_code` resolves to real data before committing to it — a broader bug that could recur for any other category with the same kind of mismatch. Not fixed here to keep this PR scoped to the taxonomy data.


taxonomies/food/categories.txt
```diff
@@ -79083,7 +79083,7 @@ hr: Cezar umaci
 it: Salse caesar, salsa caesar
 lt: Cezario padažai
 sv: Caesardressingar
-agribalyse_food_code:en: 11248
+agribalyse_proxy_food_code:en: 11187
```
Freso: How does this behave if we just don’t have a code here?

Freso (commented):
The PR description looks a lot like “AI”. Please remember to disclose any LLM-usage when submitting PRs.

taxonomies/food/categories.txt
Eduardo-Vilas-Boas: Regarding the AI question — you're right, I am using AI (Claude Code) to help with this contribution, but I review all code changes myself and check all explanations for factual accuracy before posting. I'll disclose this more clearly going forward, as expected.

On the code question: tested it directly — identical result either way (grade d, score 38, agribalyse code 11187). The loop in compute_environmental_score_agribalyse() just walks up to the parent en:salad-dressings when nothing's declared here.

So functionally it's a wash today. For explicit vs. none:
- No code: simpler diff, nothing to keep in sync if the parent's code changes.
- Explicit proxy: survives future re-parenting — this category has already moved once (en:sauces → en:salad-dressings, in #14472), so a future move to a branch with no Agribalyse data would silently break the score again, with nothing in this category's own block to explain why. It also matches existing convention (many other categories already use agribalyse_proxy_food_code this way), and doesn't rely on compute_environmental_score_agribalyse()'s own fallback loop.

I'm not sure what the right convention is here — which would you recommend?

Freso: Please add the expected disclosure to this PR as well.

I would _personally_ do something like https://github.com/openfoodfacts/openfoodfacts-server/pull/14616/changes/610e5ecb30832ace469f30de4a87044be4ab0527 which would make it clear to future editors _why_ there’s no agribalyse code for it. if it is reparented in the future in a way that the inherited proxy changes… then it’s probably good that the inherited proxy changed. if not, the proxy can be specifically overridden then.

Eduardo-Vilas-Boas: Thanks for your patience and guidance, Freso — updated as you suggested.

Freso (approved):
> mirroring the existing precedent in `taxonomies/additives.txt`

There’s no precedent (that I can readily recall at least). I [reverted that commit](https://github.com/openfoodfacts/openfoodfacts-server/pull/14616/commits/7810966b1a6a14affe085b386863d91cd8f250eb) for that PR ([which also isn’t merged anyway](https://github.com/openfoodfacts/openfoodfacts-server/pull/14616)).

Other than that, this looks good to me now.

stephanegigandet (approved):
Thanks

